PR #7, opened 2018-01-20
Really fix bindings link in README.md

johannhof:
Sorry, the previous commit hid the reference correctly but had the wrong link.

Turns out that you still have to use square brackets for reference style links, but the reference can't be part of the block.

Markdown is a mess.


alexcrichton:
Heh I know what you mean by sometimes it takes a few tries...

Issue #10, opened 2018-02-01
unsafe `into_inner()`


bringking:
@alexcrichton thanks for this tool, it looks amazing. I was trying to get it running from master and I am running into an unsafe call error. I tried wrapping the module in `unsafe` but that didn't seem to work.

_code_
```
#![feature(proc_macro)]

extern crate wasm_bindgen;

use wasm_bindgen::prelude::*;

wasm_bindgen! {
    pub fn greet(name: &str) -> String {
      format!("Hello, {}!", name)
    }
}
```

```
error[E0133]: call to unsafe function requires unsafe function or block
   --> /Users/charlesking/.cargo/git/checkouts/wasm-bindgen-aa905f4284ca7fbe/322f520/src/lib.rs:341:13
    |
341 |             self.value.into_inner()
    |             ^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function

error: aborting due to previous error

error: Could not compile `wasm-bindgen`.

```

Seems to be related to changes made here - https://github.com/alexcrichton/wasm-bindgen/commit/c51a342cb3ef27d176d5713084d05ac04963b6db. If I specify an earlier revision, it compiles without error.

bringking:
Nightly was out of date! Sorry

PR #11, opened 2018-02-01
WasmRefCell into_inner with 'unsafe'

bringking:
This change wraps the `into_inner` call with `unsafe`. `UnsafeCell into_inner` is unsafe, thus it seems the `into_inner` in WasmRefCell needs to be unsafe as well. Relates to my issue here - #10

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -337,8 +337,8 @@ pub mod __rt {
             }
         }
 
-        pub unsafe fn into_inner(self) -> T {
-            self.value.into_inner()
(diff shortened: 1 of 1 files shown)


alexcrichton:
Ah thanks for this! I think though that this was a recent change in nightly, if you upgrade nightly does the error go away?

bringking:
@alexcrichton Let me check, I am running Mac, and I thought it was up to date..

bringking:
@alexcrichton my bad, I was just slightly out of date! Updating fixed it

PR #20, opened 2018-02-12
Don't panic on unsupported exports

dflemstr:
It's most likely safe for us to skip exports that we don't yet support.  Doing nothing will at least not hurt in those cases.

This is needed because recent `rustc` has started exporting a `__web_table` table (not sure what it's used for though)


alexcrichton:
Sounds like a good idea to me, thanks!

PR #34, opened 2018-02-18
Ensure Memory Accesses are Exposed Before Accessing

rylev:
The following program leads to a runtime error:

```rust 
#![feature(proc_macro)]
extern crate wasm_bindgen;
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
#[no_mangle]
pub extern fn foo() -> Vec<i16> {
    vec![1,2,3]
}
```

At run time this leads to: `getUint16Memory is not defined`. 

This change ensures that the memory accessors are all available.


alexcrichton:
Oops, thanks!

PR #36, opened 2018-02-22
C-Style Enums

### LINKED ISSUE #31: Support for C-Style Enums
Supporting the full power of Rust enums is probably a lot of work, but C-Style enums seem more doable. Is this something worth while tackling? If so, I would volunteer to help out (although I might need a bit of coaching).

rylev:
Fixes #31. This is a very basic version of C-Style enum support. 

Some limitations are:
* Since they are C-Style, they are only represented as numbers in JS and cannot have associated data
* The number representation is hardcoded to be in order starting at 0.
*  If functions that take these enums as arguments are not given a number that can be translated to the enum, the JS throws an error. 

All these limitations should be fairly easy to overcome, but I wanted to get this reviewed sooner rather than later since I'm new to the codebase. If this is merged, I can work on fixing the limitations (especially the first two).

I wasn't sure if this was the best way to structure the code, so feedback is very welcome :-D

### DIFF
--- a/crates/wasm-bindgen-cli-support/src/js.rs
+++ b/crates/wasm-bindgen-cli-support/src/js.rs
@@ -978,6 +978,9 @@ impl<'a, 'b> SubContext<'a, 'b> {
         for f in self.program.imports.iter() {
             self.generate_import(f);
         }
+        for e in self.program.enums.iter() {
+            self.generate_enum(e);
+        }
     }
 
     pub fn generate_export(&mut self, export: &shared::Export) {
@@ -1423,6 +1426,19 @@ impl<'a, 'b> SubContext<'a, 'b> {
         self.cx.globals.push_str("export ");
         self.cx.globals.push_str(&dst);
         self.cx.globals.push_str("\n");
+    }
+
+    pub fn generate_enum(&mut self, enum_: &shared::Enum) {
+        let mut variants = String::new();
+
+        let mut value = 0;
+        for variant in enum_.variants.iter() {
+            variants.push_str(&format!("{}:{},", variant.name, value));
+            value = value + 1;
+        }
+        self.cx.globals.push_str(&format!("export const {} = {{", enum_.name));
+        self.cx.globals.push_str(&variants);
+        self.cx.globals.push_str("}\n");
     }
 }
 
--- a/crates/wasm-bindgen-cli-support/src/js.rs
+++ b/crates/wasm-bindgen-cli-support/src/js.rs
@@ -1051,7 +1051,8 @@ impl<'a, 'b> SubContext<'a, 'b> {
                 passed_args.push_str(arg);
             };
             match *arg {
-                shared::TYPE_NUMBER => {
+                shared::TYPE_ENUM | shared::TYPE_NUMBER => {
+                    // TODO: TS for Enum
                     dst_ts.push_str(": number");
                     if self.cx.config.debug {
                         self.cx.expose_assert_num();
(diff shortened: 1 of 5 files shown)


alexcrichton:
Awesome, thanks! Would it be possible to add some tests for this as well? 

Additionally, for the JS bindings side of things, perhaps the names of the variants could be exported as well?

rylev:
I will add tests for sure! 

For the variant exports did you mean the following?
```javascript
export const MyEnum = {Red: 1, Blue: 2};
export const Red = "Red";
export const Blue= "Blue";
```

dbkaplun:
IMO for signature parity, the translated enum should be:
```ts
type MyEnum =
  | { A: any }
  | { B: { '0': number } }
  | { C: { '0': number, '1': number } }
  | { D: { x: number, y: number } }
  ;
```

RReverser:
@dbkaplun I guess you meant to comment on the other issue? This PR covers only C-style enums.

RReverser:
@rylev As far as I understand from code, you're currently generating entries as constant object? Maybe it would make sense to use TypeScript's native enum for clarity like following?

```ts
enum Direction {
    Up,
    Down,
    Left,
    Right,
}
```

rylev:
@RReverser the generated code is pure JavaScript so it's not possible to use TypeScript enums. Or were you referring to when we generate the appropriate stuff for the .d.ts file?

RReverser:
@rylev Oh, sorry, I misunderstood that it's pure JS generation. I guess "Generates Typescript for now instead of JS (although that may come later)" from README should be probably excluded now?

rylev:
@RReverser hmm interesting. I didn't notice that in the README. The output is definitely just JavaScript (i.e. no type annotations). I wonder what @alexcrichton meant by that...

crates/wasm-bindgen-cli-support/src/js.rs
```diff
@@ -1054,15 +1054,7 @@ impl<'a, 'b> SubContext<'a, 'b> {
...
-                shared::TYPE_ENUM => {
-                    dst_ts.push_str(&format!(": {}", "any"));
-                    if self.cx.config.debug {
-                        self.cx.expose_assert_num();
-                        arg_conversions.push_str(&format!("_assertNum({});\n", name));
-                    }
-                    pass(&name)
-                }
-                shared::TYPE_NUMBER => {
+                shared::TYPE_ENUM | shared::TYPE_NUMBER => {
```
RReverser: Is it not possible to expose actual enum name instead of generic `number`?

rylev: It is. It requires a bit of a refactoring of how things are handled though. Specifically we lose all information about enums inside of functions right now, and we'll have to refactor how we handle "custom types" to make it work. I would prefer to save it for the next PR where I'll handle non-C-style enums. The changes necessary for that will sure this up.

rylev:
@alexcrichton This is ready to go. I will start looking into supporting enums with associated data. This might be a good time to look into refactoring how custom types are handled. Right now all "custom types" are assumed to be represented as classes in JS. This breaks down with enums. Having a more flexible definition of custom types will help us generate better JS. If you have any time to chat about this, let me know.

alexcrichton:
This looks great to me, thanks so much @rylev!

Also yeah the README may be a bit outdated, this used to generate typescript by default but has since switched to js. I'll update that!

Urhengulas:
@rylev Did further work on Enum support happen after this PR?

PR #38, opened 2018-02-23
Add section on node dependency in README

rylev:
The tests were failing to run for me because my version of node was old (7.2.1). Version 8.9.4 is the latest LTS version of nodejs.org. I verified the tests run with this version and version 9.6.1 - the latest version.


alexcrichton:
Thanks!

PR #40, opened 2018-02-28
Enable returning a vector of js values

### LINKED ISSUE #33: Support for returning Vec<JsValue>
Returning a `JsValue` works and returning `Vec` or `String` works, but returning a `Vec<JsValue>` does not, failing with: `the trait wasm_bindgen::convert::WasmBoundary is not implemented for std::vec::Vec<wasm_bindgen::JsValue>`. I don't see a reason why returning a `Vec<JsValue>` shouldn't work.

I would also be interested in implementing this, but I'm getting lost in the code generation.

rylev:
Fixes #33 

@alexcrichton This continues with special casing vectors. Hopefully next we can find a general solution to vectors.

### DIFF
--- a/crates/wasm-bindgen-cli-support/src/js.rs
+++ b/crates/wasm-bindgen-cli-support/src/js.rs
@@ -660,12 +660,12 @@ impl<'a> Context<'a> {
         if !self.exposed_globals.insert("get_array_js_value_from_wasm") {
             return
         }
-        self.expose_get_array_u8_from_wasm();
+        self.expose_get_array_u32_from_wasm();
         self.expose_get_object();
         self.globals.push_str(&format!("
                 function getArrayJsValueFromWasm(ptr, len) {{
-                    const mem = getUint8Memory();
-                    const slice = mem.slice(ptr, ptr + len);
+                    const mem = getUint32Memory();
+                    const slice = mem.slice(ptr / 4, ptr / 4 + len);
                     const result = []
                     for (ptr in slice) {{
                         result.push(getObject(ptr))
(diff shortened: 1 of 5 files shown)


crates/wasm-bindgen-cli-support/src/js.rs
```diff
@@ -1540,6 +1569,7 @@ impl VectorType {
             VectorKind::U32 => "Uint32Array",
             VectorKind::F32 => "Float32Array",
             VectorKind::F64 => "Float64Array",
+            VectorKind::JsValue => "object[]",
```
alexcrichton: I think this should be `any[]`?

crates/wasm-bindgen-cli-support/src/js.rs
```diff
@@ -656,6 +656,25 @@ impl<'a> Context<'a> {
...
 
+    fn expose_get_array_js_value_from_wasm(&mut self) {
+        if !self.exposed_globals.insert("get_array_js_value_from_wasm") {
+            return
+        }
+        self.expose_get_array_u8_from_wasm();
+        self.expose_get_object();
+        self.globals.push_str(&format!("
+                function getArrayJsValueFromWasm(ptr, len) {{
+                    const mem = getUint8Memory();
```
alexcrichton: This should be a u32 array, right?

rylev: Whoops! Got lucky this worked every time I tested it. :-D

alexcrichton:
Thanks!

PR #48, opened 2018-03-02
Don't run `npm install` during tests

alexcrichton:
Instead add a `package.json` and use `yarn run`.


crates/test-support/src/lib.rs
```diff
@@ -219,10 +161,21 @@ impl Project {
...
-            .arg("-c");
+        let mut cmd = if cfg!(windows) {
+            let mut c = Command::new("cmd");
+            c.arg("/c");
+            c.arg("yarn");
+            c
+        } else {
+            Command::new("yarn")
+        };
+        cmd.arg("run")
```
sendilkumarn: this `run` is not mandatory

alexcrichton:
Ok turned out invoking `rollup` concurrently was racy, so I figured I'd give `webpack` a spin given that it [now supports](https://github.com/alexcrichton/wasm-bindgen/issues/39#issuecomment-370072160) wasm natively (and I figured out how to use it!). Turns out `webpack` was just as racy (in how I was invoking it) so I fixed that bug elsewhere but now the tests are using webpack to boot!

Hurray for less base64 encoded wasm!

PR #51, opened 2018-03-03
Add workaround for hello_world running in Chrome

### LINKED ISSUE #50: hello_world example gives "WebAssembly.Instance is disallowed" in Chrome
Heya! I just tried to run the hello_world example on my machine (rustc 1.26.0-nightly (9cb18a92a 2018-03-02), wasm-bindgen master today 98b9bee876, Mac Chrome) and it throws the error:

> RangeError: WebAssembly.Instance is disallowed on the main thread, if the buffer size is larger than 4KB. Use WebAssembly.instantiate.

which appears to directly result from the generated `import * as wasm from './hello_world_wasm';` line. I found https://github.com/webpack/webpack/issues/6475 implying this is a browser limitation of sync wasm instantiation and https://github.com/alexcrichton/wasm-bindgen/issues/39 where you briefly discuss this issue, but from your comments it sounds like *async* import causes this error, not the *sync* import which is what I'm seeing.

Any chance you can point me in the right direction? Not sure if it's user error. :)

sophiebits:
Fixes #50.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,13 +221,15 @@ $ npm run serve
 ```
 
 If you open https://localhost:8080 in a browser you should see a `Hello, world!`
-dialog pop up!
+dialog pop up! This works in Firefox out of the box but not in Chrome due to a
+webpack issue. See [the hello_world README][hello-readme] for a workaround.
 
 If that was all a bit much, no worries! You can [follow along
 online][hello-tree] to see all the files necessary as well as a script to set it
 all up.
 
 [hello-tree]: https://github.com/alexcrichton/wasm-bindgen/tree/master/examples/hello_world
+[hello-readme]: https://github.com/alexcrichton/wasm-bindgen/tree/master/examples/hello_world/README.md
 
 ## What just happened?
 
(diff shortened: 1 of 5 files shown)


sophiebits:
Not sure if I overcomplicated this, let me know. (Seems in my testing that webpack prefers .wasm over .js if it finds both.)

alexcrichton:
Thanks! Since this is hopefully a bug that'll be fixed in the near future w/ webpack 4.0, I'm wondering if we could perhaps add a separate set of scripts/JS to work in chrome? Something like:

* Adding `examples/hello_world/chrome` with the shell script/JS necessary
* Updating `examples/hello_world/README.md` to mention `chrome` and when it should be used

And that way hopefully when the bug in webpack is fixed (afaik it's to use `instantiateStreaming` everywhere or `instantiate` otherwise) we can just delete some code!

sophiebits:
This diff is basically entirely green so it's deleting code either way. ;) But yeah I can look at making a separate thing!

alexcrichton:
Ok I tweaked things a bit but otherwise kept most of this, thanks again for the PR!

PR #54, opened 2018-03-05
feat: rename generated file to have bg instead of wasm

### LINKED ISSUE #52: question: why name the wasm file `<inflected_crate_name>_wasm.wasm`
the wasm part seems redundant. was just curious why it's named this way ;)

sendilkumarn:
fixes #52


alexcrichton:
Awesome, thanks!

PR #55, opened 2018-03-05
Remove no_mangle and extern wherever applicable

### LINKED ISSUE #46: Should #[no_mangle] and extern be required?
Right now when you want to export a function to JS it looks like:

```rust
#[wasm_bindgen]
#[no_mangle]
pub extern fn foo() {
    // ...
}
```

but should the `#[no_mangle]` and `extern` be required? The original thinking was that one day the `#[wasm_bindgen]` attribute could be removed and this'd just reflect what rustc does under the hood already, but I'm getting less and less certain that'll be the case (aka custom types and `&str` probably "won't ever work" in terms of being the right ABI).

It'd certainly be a little easier on the eyes to remove those two components!

sendilkumarn:
fixes #46


alexcrichton:
Awesome, thanks @sendilkumarn!

Issue #69, opened 2018-03-15
Passing Rust struct ownership into JS


Healthire:
Currently you can't pass ownership of a rust struct from rust to JS through an import, like this:
```rust
#[wasm_bindgen]
struct Foo();

#[wasm_bindgen]
extern {
    fn take_struct(foo: Foo);
}
```
but you can return it from an exported function:
```rust
#[wasm_bindgen]
fn create_foo() -> Foo {
    Foo()
}
```
Is this a conscious style choice or is passing in structs by value just not implemented yet?

In both cases js takes ownership of the object and becomes responsible for freeing it if it's dropped. But I can see that being clearer when the object is a return from a rust export, than when receiving an object as an argument from rust.

alexcrichton:
Oh oops I think that's an oversight! I think this definitely makes sense to support

alexcrichton:
Added in https://github.com/alexcrichton/wasm-bindgen/pull/70

PR #70, opened 2018-03-16
Support passing custom types by value into JS imports

Healthire:
As mentioned in #69

### DIFF
--- a/crates/wasm-bindgen-cli-support/src/js.rs
+++ b/crates/wasm-bindgen-cli-support/src/js.rs
@@ -1402,6 +1402,17 @@ impl<'a, 'b> SubContext<'a, 'b> {
                     invoc_args.push(format!("getObject(arg{})", i));
                     abi_args.push(format!("arg{}", i));
                 }
+                custom if (custom as u32) & shared::TYPE_CUSTOM_REF_FLAG == 0 => {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Awesome, thanks!

nkoehring:
I still didn't really get this to work. I can export the struct but don't see anything in the wasm exports. I tried to create a static instance with lazy_static and use that in an exported function, which compiles but throws an error (unreachable used). What am I doing wrong? Does the struct and its implementations (like ::new) need to be public?

alexcrichton:
@nkoehring can you share the code you're with? I may be able to help out if I can poke around!

nkoehring:
@alexcrichton thanks for your reply! In the meantime I found out that it is the Rand crate that I used in `::new`. So my question actually changed to "how to make crates available".

alexcrichton:
@nkoehring hm so you mean making existing crates available to JS? Currently that requires wrapping them in `#[wasm_bindgen]` to have an interface that JS can use as well

nkoehring:
So the crate itself would have to be changed, I guess? Anyways, I solved the problem by using Javascripts random function.

alexcrichton:
Yeah at this time it'd require upstream changes or a local wrapper

PR #76, opened 2018-03-25
Fix some JS == → === and != → !== occurrences

dflemstr:
These are the ones my linter complained about in particular


alexcrichton:
Thanks!

Issue #75, opened 2018-03-24
Mapping a futures::Future to an ES6 Promise?


dflemstr:
(This issue is mostly opened for "discussion" purposes, trying to reach the right target audience)

A lot of people will probably have the use-case of mapping a Rust `futures::Future` to an ES6 `Promise`.  My solution right now has been to do this:

```rust
fn foo() -> impl futures::Future<Item = String, Error = String> {
    // ...
}

#[wasm_bindgen]
struct JsApi { /* ... */ }

#[wasm_bindgen]
impl JsApi {
    pub fn foo(&self, resolve: JsValue, reject: JsValue) {
        let future = foo()
            .map(move |s| resolveString(resolve, s))
            .map_err(move |e| rejectString(reject, e));
        self.executor.spawn(future);
    }
}

#[wasm_bindgen(module = "../ffi")]
extern "C" {
    #[allow(non_snake_case)]
    fn resolveString(resolve: JsValue, string: String);
    #[allow(non_snake_case)]
    fn rejectString(reject: JsValue, string: String);
}
```

Then, in `ffi.ts`:

```typescript
export function resolveString(resolve: (value: string) => void, value: string) {
  console.debug("resolved:", value);
  resolve(value);
}
export function rejectString(reject: (value: string) => void, value: string) {
  console.warn("rejected:", value);
  reject(value);
}
```

...and finally:

```typescript
export function foo() {
  return new Promise<string>((resolve, reject) => wasm.foo(resolve, reject));
}
```

This is quite verbose and lots of boilerplate.  Some ideas on what might help:

  * Being able to call `JsValue`s from Rust.
  * Supporting generics + monomorphized extern JS functions
  * Having native support for `Promise`/`Future` in particular since it might be a common use-case.

Any other ideas on how to improve this?

alexcrichton:
I think this'd be a fantastic idea! I unfortunately haven't figure out much of a way to work with closures and wasm-bindgen today, but I'm hopeful that we can figure that out soon to start binding APIs like promises!

fitzgen:
I have integrated ES Promises and Futures (and tokio) once before, and I wrote a bit about the gotchas that will be encountered: https://github.com/starlingjs/starling/blob/master/src/promise_future_glue.rs#L1-L67

Pauan:
[...]

PR #77, opened 2018-03-27
Add support for constructing JsValue instances generically

dflemstr:
This is one step in the direction of fixing #75.  The idea is that it should be possible to construct `JsValue`s from any Rust type so that we are able to eventually call generic callbacks such as `Promise.resolve` without having to round-trip through Javascript helper functions.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -88,6 +88,7 @@ impl ToTokens for ast::Program {
 impl ToTokens for ast::Struct {
     fn to_tokens(&self, tokens: &mut Tokens) {
         let name = &self.name;
+        let new_fn = syn::Ident::from(shared::new_function(self.name.as_ref()));
         let free_fn = syn::Ident::from(shared::free_function(self.name.as_ref()));
         let c = shared::name_to_descriptor(name.as_ref());
         let descriptor = Literal::byte_string(format!("{:4}", c).as_bytes());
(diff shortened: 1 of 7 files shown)


dflemstr:
Should this rather be done as `impl From<Foo> for JsValue`?

dflemstr:
I believe the build is failing due to flakiness/a dependency change. I can't see how it's related to my changes.

alexcrichton:
Thanks for the PR! Sorry this has been a crazy week so I've been a bit delayed!

Suffices to say though this looks great! My only comment would be if we can avoid `__wbindgen_classes`? We may be able to generate a shim that doesn't go through a `JsValue` method (aka a method generated by the `wasm-bindgen` CLI tool) perhaps?

I think the custom shim approach may also optimize a bit better with engines as well

dflemstr:
@alexcrichton I switched to the approach that you suggested. Sadly `master` doesn't compile so I can't test it, but does the code at least look reasonable or is there something I should change about the approach before trying to make things build?

alexcrichton:
Egad sorry about that! I rebased and pushed a few commits as well to address some recent changes, and I think tests should pass now. 

I think the only final thing to do is to slim down the JS if possible. Could the generation of the JS shim be avoided if the wasm doesn't actually require it?

alexcrichton:
Awesome, thanks!

PR #81, opened 2018-03-28
Cli: Fix browser flag

killercup:


killercup:
I have no clue why your Ci didn't catch this. Also this wouldn't have happend if you used structopt.

PR #98, opened 2018-03-30
Fix import TextDecoder in output for nodejs config

corbinu:
Quick typo fix. When compiling for --nodejs the output is incorrect for importing TextDecoder from Node util.


alexcrichton:
Oh dear, sorry about that!

PR #100, opened 2018-04-03
Reduce the JS file size and optimizing encoder and decoder functions

sendilkumarn:
Based on : https://jsperf.com/funccallstocheck

yay 🎉 100th


alexcrichton:
Thanks!

PR #101, opened 2018-04-04
Initial support for closures

alexcrichton:
This commit starts wasm-bindgen down the path of supporting closures. We
discussed this at the recent Rust All-Hands but I ended up needing to pretty
significantly scale back the ambitions of what closures are supported. This
commit is just the initial support and provides only a small amount of support
but will hopefully provide a good basis for future implementations.

Specifically this commit adds support for passing `&Fn(...)` to an *imported
function*, but nothing elese. The `&Fn` type can have any lifetime and the JS
object is invalidated as soon as the import returns. The arguments and return
value of `Fn` must currently implement the `WasmAbi` trait, aka they can't
require any conversions like strings/types/etc.

I'd like to soon expand this to `&mut FnMut` as well as `'static` closures that
can be passed around for a long time in JS, but for now I'm putting that off
until later. I'm not currently sure how to implement richer argument types, but
hopefully that can be figured out at some point!


crates/backend/src/ast.rs
```diff
@@ -81,10 +81,25 @@ pub struct Variant {
...
-    ByMutRef(syn::Type),
-    ByValue(syn::Type),
+pub struct Type {
+    pub ty: syn::Type,
+    pub kind: TypeKind,
+    pub loc: TypeLocation,
+}
+
+#[derive(Copy, Clone)]
+pub enum TypeKind {
```
fitzgen: `OwnershipKind`? When I saw `TypeKind` I was expecting something like scalar vs struct vs enum vs function...

alexcrichton:
Ok! I've pushed up some more ambitious changes which support long-lived `'static` closures as well. There's associated documentation/examples as well.

@fitzgen mind giving this another look? I'm specifically curious for your take on `src/closure.rs`, which took me *way* too long to figure out and I'm still not particularly happy with unfortunately.

PR #106, opened 2018-04-06
Enabling amd builds 

csharad:
Before I describe what I have done here. Here's why I want it to be included?

My usecase is that I wanted to  create a cli without any dependency on webpack. For that I could use `--browser` but it required `type="module"` which is behind flags in firefox and testing in chrome is weird. 

Seeing what `stdweb` did, I wanted to emulate the same here using `--umd` builds. The implementation is very simple. Only proper js is accessible in exports.

You could directly import and use as is.
```
<script src="sample.js"></script>
<script>
sample.then(m => m.run());
</script>
```

As you can tell, this commit is incomplete. I don't know how to handle `SubContext::import_name`? As far as I know, it handles the `js_namespace` values and would I need to skip through the imports?

Also, where are custom js library imports handled? Because resolving their dependency tree would require `webpack`. And I would not want to use that, so panicking when js libs are used would be best in this usecase.

### DIFF
--- a/crates/cli-support/src/js.rs
+++ b/crates/cli-support/src/js.rs
@@ -1622,6 +1622,10 @@ impl<'a, 'b> SubContext<'a, 'b> {
 
     fn import_name(&mut self, import: &shared::Import, item: &str) -> String {
         if let Some(ref module) = import.module {
+            if self.cx.config.no_modules {
+                panic!("import from `{}` module not allowed in `--no-modules`. use `--nodejs` or `--browser` instead", module);
+            }
+
             let name = import.js_namespace.as_ref().map(|s| &**s).unwrap_or(item);
 
             if self.cx.imported_names.insert(name.to_string()) {
--- a/crates/cli-support/src/js.rs
+++ b/crates/cli-support/src/js.rs
@@ -47,7 +47,7 @@ impl<'a> Context<'a> {
         let contents = contents.trim();
         let global = if self.config.nodejs {
             format!("module.exports.{} = {};\n", name, contents)
-        } else if self.config.umd {
+        } else if self.config.amd {
             format!("__exports.{} = {}\n", name, contents)
         } else {
             if contents.starts_with("function") {
@@ -66,7 +66,7 @@ impl<'a> Context<'a> {
         {
             let mut bind = |name: &str, f: &Fn(&mut Self) -> String| {
                 if !self.wasm_import_needed(name) {
-                    return
+                    return;
                 }
                 let contents = f(self);
                 let contents = contents.trim();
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR!

The current thinking behind the current system is that if no flags are passed the output uses ES modules as the structure. The `--nodejs` flag recently was tweaked to use CommonJS imports/exports as well as the ouput is tailored for node anyway. In that sense I'd be totally down for supporting other module systems!

I know right now it's indeed painful to use the raw output of `wasm-bindgen` in a browser, it currently requires a bundler like Webpack 4.0 to come along and make it usable for the browser. I think adding a flag here is a good idea and makes sense to strive to make the output usable-by-default. I'm not personally too privvy with JS conventions though so I don't know how this would best be done, but I'm ok trusting you as well :)

As for your questions do those functions/umd map well to ES modules? I can answer with that context but I'm otherwise not sure of how I'd answer specifically with umd

corbinu:
@alexcriton just a note the companion plugin I mentioned on the prettier issue would also handle this by using the TypeScript compiler to emit whatever module format is needed.

Not to say I don’t like it @cshard just wanted to say we may be duplicating some work here

csharad:
@corbinu I don't think it would be duplication, as I am targetting a build tool-less path for using this lib. Though I could remove `--umd` (universal) in favour of [`--amd`](https://github.com/umdjs/umd/blob/master/templates/amdWeb.js) (browser only), as `--nodejs` needs are already fullfilled.

@alexcrichton `#[wasm_bindgen(module = "./index")]` refers to some external dependency, possibly an npm one, so mapping them to umd would require a full-on bundler which is not feasible and also not required for accessing browser js api from rust.
So, I was thinking if you could point me to the right direction, if I could panic whenever an external module is used when building with umd.

I also wanted to know what the following lines did as they are invoking `require|import` which are not available in umd.

https://github.com/rustwasm/wasm-bindgen/blob/9723fdd69bbcd26bed7c61d1904ca21ffbf5fdf7/crates/cli-support/src/js.rs#L1593

csharad:
I have dropped support for node in umd and changed it to `--amd`.

alexcrichton:
Thanks again! There's a few things I'm not entirely sure about here, but just to confirm the motivation here is to make the output usable by default without a preprocessor, right? If so, can you detail a bit about why a preprocessor isn't being used but a module system is still being used?

csharad:
@alexcrichton Because this is similar to the technique webpack uses to bundle the project into one single file. It creates wrappers around the dependencies and make it accessible to the app without exposing it as global.
This is not a module system per se, but just a way to make it behave like one(without imports).

alexcrichton:
@csharad I think this makes sense if wasm-bindgen is the only player, but does this integrate with existing JS code and such? I'm still trying to sort of see how this'd be used and how it'd integrate into a larger framework

csharad:
@alexcrichton This won't integrate with existing JS code, for that webpack would be best. The only use case for me would be to use every thing from Rust with plain bindings in JS. 
I see you are reluctant to see its worth, but from my perspective it could provide the basic binding as provided by `stdweb`.

alexcrichton:
Oh no sorry I'm mostly just trying to explore the use case here and better understand it to figure out if this is the best fit. If this is explicitly not intended for interoperating with existing JS code then I wonder if the command line switch should be a bit different? Maybe something like `--standalone` or `--no-es-modules` or something like that?

csharad:
`--standalone` seems fair enough because it only concerns with a single module i.e. itself.

alexcrichton:
@corbinu a good point! @csharad is the module system necessary if it's not intended for general consumption?

alexcrichton:
Ok! I think then in that case we may want a few things:

* We'll probably want to call this command line option something like `--no-modules`
* We'll want to verify that the wasm doesn't import from anything other than `./name` (aka JS shims generated by wasm-bindgen). Nothing else can be imported anyway in this mode!
* I think we'll want a sketch of how this hooks up into an application, for example how will we know when the wasm is downloaded and finished instantiating?

Does that sound ok?

csharad:
1. I'll do it with `--no-modules` then. 
2. Panicking out with appropriate message could be done when a module is used.
3. I don't understand correctly what you meant by the third one. Did you mean, to invoke `main()` when every thing has built or user be able to invoke any exposed code in JS?
For the second question user has to wait on the project global:

> ```
> <script src="sample.js"></script>
> <script>
> sample.then(m => m.run());
> </script>
> ```

crates/cli-support/src/js.rs
```diff
@@ -227,23 +229,59 @@ impl<'a> Context<'a> {
...
+                        root.{module} = factory();
+                    }}
+                }}(typeof self !== 'undefined' ? self : this, function() {{
+                    let wasm;
+                    const __js_exports = {{}};
+                    const __exports = {{}};
+                    {globals}
+                    __js_exports['./{module}'] = __exports;
+                    {import_wasm}
+                }}))
```
alexcrichton: I think this preamble can be simplified quite a bit, right? If we're only working with one local module I dont think we need the extra function scopes and such?

alexcrichton:

crates/cli-support/src/js.rs
csharad: This preamble would be useful if someone else imports it as a library in js but then again they should rather use other builds.

csharad:
@alexcrichton How would you test it? It requires `window` and am not able to run it on node. Are browser tests possible?

alexcrichton:
Ah good point, we may have to go without tests on this one. Thinking some more about this, though. what do you think about something like:

* By default all the internal helpers of wasm-bindgen are hidden in a `(function() { ... })()` scope
* All of the wasm-bindgen exports are located at `window.wasmBindgen`, aka `window.wasmBindgen.greet` for a hello world example
* `window.wasmBindgen.init` is a generated function which takes a path to the wasm file and runs initialization/returns a promise

I think that way usage would look like:

```html
<script src="wasm-bindgen-generated.js"></script>
<script>
  wasmBindgen.init('/path/to/wasm_bg.wasm').then(() => wasmBindgen.greet('Hello!'));
</script>
```

How's that sound?

corbinu:
@alexcriton if you want I can add a browser testing set for all the features in a separate PR?

alexcrichton:
@corbinu that'd be awesome!

csharad:
@alexcrichton Sounds fine. I see that it will trim down the size for multiple imports.

csharad:
What if the user exposes `fn init()`?

alexcrichton:
Ok I went ahead and folded that in, thanks again @csharad!

Issue #123, opened 2018-04-13
FnMut stack closures aren't supported


alexcrichton:
A bummer! They definitely should be as they're often the most ergonomic to use.

Issue #111, opened 2018-04-09
Allow returning Vec<T>


rookboom:
It would really be great to be able to return a vector of structs or tuples:

For example. I have the following type:
```
#[wasm_bindgen]
struct Range {
    offset: u32,
    length: u32
}
```

that I want to return in my function

```
#[wasm_bindgen]
pub fn token_ranges(text: &str) -> Vec<Range> 
```

I am getting this error:

```
   Compiling picl_wasm_runtime v0.1.0 (file:///A:/Repos/picl_native_runtime/bindings/typescript)
error[E0277]: the trait bound `std::boxed::Box<[Range]>: wasm_bindgen::convert::WasmBoundary` is not satisfied
  --> src\lib.rs:15:1
   |
15 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::WasmBoundary` is not implemented for `std::boxed::Box<[Range]>`
   |
   = help: the following implementations were found:
             <std::boxed::Box<[u16]> as wasm_bindgen::convert::WasmBoundary>
             <std::boxed::Box<[i16]> as wasm_bindgen::convert::WasmBoundary>
             <std::boxed::Box<[f32]> as wasm_bindgen::convert::WasmBoundary>
             <std::boxed::Box<[i32]> as wasm_bindgen::convert::WasmBoundary>
           and 5 others
   = note: required because of the requirements on the impl of `wasm_bindgen::convert::WasmBoundary` for `std::vec::Vec<Range>`
```

My workaround is to flatten my data and return a Vec<u32> and then splice my token ranges on the JS side. This is unfortunate...

How can I add a WasmBoundary trait for a custom type?
Is there a better way?

alexcrichton:
Thanks for the report!

Right now there's not a great way to do this in wasm-bindgen itself in terms of the conversion here has to be also accompanied with a conversion on the JS side which may be somewhat tricky. It should certainly be possible though given enough support!

Hywan:
Same issue here :-).

What kind of support do you need?

alexcrichton:
@rookboom @Hywan do y'all basically need `Vec<T>` where `T` has `#[wasm_bindgen]` on it?

Hywan:
@alexcrichton Exactly, that's precisely my usecase.

rookboom:
Yes, that would be fantastic.

alexcrichton:
[...]

Issue #104, opened 2018-04-05
Support types like strings in closures


alexcrichton:
The first support for closures is being added in https://github.com/alexcrichton/wasm-bindgen/pull/101, but it's very bare-bones. Currently it doesn't allow working with arguments/return values that are any more interesting than the four fundamental wasm types themselves (and even then u64 won't work). 

Let's take a look at an example:

```rust
#[wasm_bindgen]
extern {
    fn foo(a: &Fn(&str));
}
```

Here we're calling a JS function, `foo`, which takes a closure. This closure will have a stack lifetime (not persistent beyond this one function call). The closure also receives one argument, a string. 

First up we need to generate a bunch of shims. The ones in Rust need to look like:

```rust
fn foo(a: &Fn(&str)) {
    extern fn shim(data: *const u8, a: u32) {
        let a = extract_string(a); // a bit more fancy than this, but mostly the same
        (*(data as *const &Fn(&str)))(a)
    }

   extern {
        fn foo_import(data: *const u8, shim: u32);
    }
    unsafe {
        foo_import(&a as *const &Fn(&str) as *const u8, shim as u32)
    }
}
```

This part is relatively straightforward and easy to do today with the construction of wasm-bindgen, heavily relying on the trait system. In other words all this functionality can be wrapped up in trait implementations that are easier to read/write/document than generated code.

The problem here comes when we want to generate the JS shims, and unfortunately comes back to the trait system. Somehow we need to instruct JS to generate a shim that takes the arguments passed here *and* generate a custom JS wrapper which wraps the rust function pointer passed to JS here. 

Currently the way wasm-bindgen works is that it emits a JSON AST effectively for the `wasm-bindgen` CLI tool to later come, parse, and generate bindings for. This JSON AST is constructed at compile time by the `#[wasm_bindgen]` attribute. This is done by making a giant `const` and slapping `#[wasm_custom_section]` on that value.

The way this all works is a bit janky though and ends up preventing the closure case from working well. An import like above will have a list of arguments associated with it. We rely on the trait system to use an associated type projection to figure out what the actual argument is here. In other words we're building this giant JSON literal, and somewhere in the middle of it we're splicing in some content from a projection of a particular type (aka `<&Fn(&str) as ToRefWasmBoundary>::DESCRIPTOR`).

[...]

PR #124, opened 2018-04-13
Overhaul how type information gets to the CLI

alexcrichton:
This commit is a complete overhaul of how the `#[wasm_bindgen]` macro
communicates type information to the CLI tool, and it's done in a somewhat...
unconventional fashion.

Today we've got a problem where the generated JS needs to understand the types
of each function exported or imported. This understanding is what enables it to
generate the appropriate JS wrappers and such. We want to, however, be quite
flexible and extensible in types that are supported across the boundary, which
means that internally we rely on the trait system to resolve what's what.

Communicating the type information historically was done by creating a four byte
"descriptor" and using associated type projections to communicate that to the
CLI tool. Unfortunately four bytes isn't a lot of space to cram information like
arguments to a generic function, tuple types, etc. In general this just wasn't
flexible enough and the way custom references were treated was also already a
bit of a hack.

This commit takes a radical step of creating a **descriptor function** for each
function imported/exported. The really crazy part is that the `wasm-bindgen` CLI
tool now embeds a wasm interpreter and executes these functions when the CLI
tool is invoked. By allowing arbitrary functions to get executed it's now *much*
easier to inform `wasm-bindgen` about complicated structures of types. Rest
assured though that all these descriptor functions are automatically unexported
and gc'd away, so this should not have any impact on binary sizes

A new internal trait, `WasmDescribe`, is added to represent a description of all
types, sort of like a serialization of the structure of a type that
`wasm-bindgen` can understand. This works by calling a special exported function
with a `u32` value a bunch of times. This means that when we run a descriptor we
effectively get a `Vec<u32>` in the `wasm-bindgen` CLI tool. This list of
integers can then be parsed into a rich `enum` for the JS generation to work
with.

This commit currently only retains feature parity with the previous
implementation. I hope to soon solve issues like #123, #104, and #111 with this
support.


fitzgen:
> This commit takes a radical step of creating a descriptor function for each
> function imported/exported. The really crazy part is that the wasm-bindgen CLI
> tool now embeds a wasm interpreter and executes these functions when the CLI
> tool is invoked. By allowing arbitrary functions to get executed it's now much
> easier to inform wasm-bindgen about complicated structures of types. Rest
> assured though that all these descriptor functions are automatically unexported
> and gc'd away, so this should not have any impact on binary sizes
> 
> A new internal trait, WasmDescribe, is added to represent a description of all
> types, sort of like a serialization of the structure of a type that
> wasm-bindgen can understand. This works by calling a special exported function
> with a u32 value a bunch of times. This means that when we run a descriptor we
> effectively get a Vec<u32> in the wasm-bindgen CLI tool. This list of
> integers can then be parsed into a rich enum for the JS generation to work
> with.

This seems like a good comment to have at the top of some module -- maybe `describe.rs`?

PR #127, opened 2018-04-14
Support `new Foo(...)` to fix #115

### LINKED ISSUE #115: Supporting `new Foo(...)` instead of `Foo.new(...)`
Given the otherwise pleasantly natural mapping between rust and javascript types, the behavior of the javascript new is counterintuitive and an easy pitfall. Therefore I'd like to change the constructor to map to what `new` is in rust.

So instead of the currently generated code

```javascript
module.exports.MyClass = class MyClass {

    constructor(ptr) {
        this.ptr = ptr;
    }

    static new(arg0, arg1) {
        const ret = wasm.myclass_new(arg0, arg1);
        return new MyClass(ret);
    }
    // ...
}
```

I'd suggest something like the following: 

```javascript
class _ConstructorToken {
    constructor(ptr) {
        this.ptr = ptr
    }
}

module.exports.MyClass = class MyClass {
    constructor(...args) {
        if (args.length === 1 && args[0] instanceof _ConstructorToken) {
            this.ptr = args[0].ptr;
        }

        this.ptr = wasm.myclass_new(...args);
    }

    // ...
}
```

An alternative approach would be passing a token the way it's currently done with --debug. I don't have enough background knowledge though to decide which is better.

It also needs to be figured out whether a static new returning an instance should be used for the  constructor by convention or if it should be specially annotated (e.g. with `#[wasm_bindgen(constructor)]` with the extern functions).

konstin:
When there is a function is annotated `#[wasm_bindgen(constructor)]` in an impl block for `Foo`, you can  use `new Foo()` in the javascript directly. The `constructors::classes` test should give a good impression on how this looks in practice.

~~A nice side effect is that constructors are now independent of debug.~~ I haven't tested for perfomance implications though as mentioned in https://github.com/rustwasm/wasm-bindgen/issues/115#issuecomment-380119274. I also refrained from replacing `Foo.new()` with `new Foo()` in the docs and examples.

[The user facing constructor variant](https://github.com/rustwasm/wasm-bindgen/compare/describe...konstin:new?expand=1#diff-8e501067c51f03c3db3c9d1286f71da5R335) is currently a hack. I couldn't think of a better way to do that, so suggestions are welcome.

This pull request is based on the describe branch from #124 to avoid merge conflicts.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -203,23 +203,13 @@ impl Program {
             panic!("can only bindgen safe functions");
         }
 
-        let mut opts = BindgenAttrs::find(&mut method.attrs);
+        let opts = BindgenAttrs::find(&mut method.attrs);
         let is_constructor = opts.constructor();
         let constructor = if is_constructor {
             Some(method.sig.ident.to_string())
         } else {
             None
         };
-
-        if is_constructor {
-            let pos = opts.attrs
-                .iter()
-                .enumerate()
-                .find(|(_, a)| **a == BindgenAttr::Constructor)
-                .unwrap()
-                .0;
-            opts.attrs.remove(pos);
-        }
 
         let (function, mutable) = Function::from_decl(
             method.sig.ident,
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -14,7 +14,7 @@ pub struct Export {
     pub class: Option<syn::Ident>,
     pub method: bool,
     pub mutable: bool,
-    pub constructor: bool,
+    pub constructor: Option<String>,
     pub function: Function,
 }
 
@@ -123,7 +123,7 @@ impl Program {
                     class: None,
                     method: false,
                     mutable: false,
-                    constructor: false,
+                    constructor: None,
                     function: Function::from(f, opts),
                 });
             }
@@ -204,12 +204,14 @@ impl Program {
         }
 
         let mut opts = BindgenAttrs::find(&mut method.attrs);
-        let constructor = opts.constructor();
-        if constructor {
-            if method.sig.ident != syn::Ident::from("new") {
-                panic!("The constructor must be called 'new' for now")
-            }
+        let is_constructor = opts.constructor();
+        let constructor = if is_constructor {
+            Some(method.sig.ident.to_string())
+        } else {
+            None
+        };
 
+        if is_constructor {
             let pos = opts.attrs
                 .iter()
                 .enumerate()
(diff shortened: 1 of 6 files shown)


crates/backend/src/ast.rs
```diff
@@ -201,21 +203,39 @@ impl Program {
...
+        };
+
+        if is_constructor {
+            let pos = opts.attrs
+                .iter()
+                .enumerate()
+                .find(|(_, a)| **a == BindgenAttr::Constructor)
+                .unwrap()
+                .0;
+            opts.attrs.remove(pos);
```
alexcrichton: How come this was necessary to remove the attribute from our list?

crates/cli-support/src/js.rs
```diff
@@ -316,42 +317,42 @@ impl<'a> Context<'a> {
...
-                dst.push_str(&format!("
-                    constructor(ptr, sym) {{
-                        _checkToken(sym);
-                        this.ptr = ptr;
+
+            self.expose_constructor_token();
+
+            dst.push_str(&format!("
+                constructor(...args) {{
+                    if (args.length === 1 && args[0] instanceof ConstructorToken) {{
```
alexcrichton: Could this binding code be generated only if the constructor is present? If it's not present I think the previous construction should work, right?

crates/cli-support/src/js.rs
```diff
@@ -765,6 +753,27 @@ impl<'a> Context<'a> {
...
+        self.globals.push_str("
+            class ConstructorToken {
+                constructor(ptr) {
+                    this.ptr = ptr;
+                }
+            }
+        ");
+
+        self.typescript.push_str("
+            class ConstructorToken {
```
alexcrichton: I think we probably don't want to expose this in the typescript, right?

in that it's not intended for public consuption

crates/backend/src/ast.rs
konstin: This was a failed attempt to remove the `#[wasm_bindgen(constructor)]` from the function; It's removed now.

crates/cli-support/src/js.rs
konstin: I've removed the token and the constructor from the typings, as the constructor needs the token for its signature. I'd argue that in this case we should also remove the `public ptr: number;` from the typings.

crates/cli-support/src/js.rs
konstin: Yes, I've changed it so that Token version is only used when either debug or a constructor are set. Direct construction is possible through a tiny `__construct` shim. This also allows functions to directly create an instance even if they don't know whether the struct has a constructor or not. (I can change the name if you've got a better one)

alexcrichton:
This looks fantastic, thanks so much @konstin!

PR #129, opened 2018-04-14
Overhaul the conversion traits

alexcrichton:
This commit overhauls the conversion traits used for types crossing the Rust/JS
boundary. Previously there were a few ad-hoc traits but now there's only two.
One is for converting JS values to Rust values, `FromWasmAbi`, and the other is
for converting Rust values into JS values, `IntoWasmAbi`. These two traits are
now uniformly used for all operations such as creating references and such.

The `IntoWasmAbi` trait is pretty straightforward but `FromWasmAbi` is much
trickier. We need to be able to create references which necessitates a good deal
of funkiness with the API, but it in general should work out alright.

The goal of this commit is to reduce the number of traits in play so it's more
straightforward which traits are used for closures.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -57,8 +57,8 @@ pub struct ImportType {
 
 pub struct Function {
     pub name: syn::Ident,
-    pub arguments: Vec<Type>,
-    pub ret: Option<Type>,
+    pub arguments: Vec<syn::Type>,
+    pub ret: Option<syn::Type>,
     pub opts: BindgenAttrs,
     pub rust_attrs: Vec<syn::Attribute>,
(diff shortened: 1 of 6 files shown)


alexcrichton:
This is a necessary prerequisite for https://github.com/rustwasm/wasm-bindgen/issues/104

alexcrichton:
Oh dear I ended up merging this in https://github.com/rustwasm/wasm-bindgen/commit/0e6325d833c72971f041dbe3c7026661cd9659fb rather than pushing to this commit by accident.

In any case I realized right after posting this that these traits as written aren't sound, so I had to scale back and be a bit more conservative in the referenced commit (namely having multiple traits for creating Rust values from JS ones.

PR #135, opened 2018-04-16
Update DESIGN.md on recent closure/conversion changes

alexcrichton:
Hopefully covering a lot of the groundwork done for closures in case anyone's
curious down the road!

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -385,8 +385,9 @@ happening:
   free the space we allocated to pass the string argument once the function call
   is done.
 
-At this point it may be predictable, but let's take a look at the Rust side of
-things as well
+Next let's take a look at the Rust side of things as well. Here we'll be looking
+at a mostly abbreviated and/or "simplified" in the sense of this is what it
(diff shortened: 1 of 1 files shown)


alexcrichton:
@fitzgen as requested in https://github.com/rustwasm/wasm-bindgen/pull/124!

DESIGN.md
```diff
@@ -1042,48 +1044,181 @@ possibilities!
...
+The `Stack` argument to the conversion traits above looks like:
+
+```rust
+pub trait Stack {
+    fn push(&mut self, bits: u32);
+    fn pop(&mut self) -> u32;
+}
+```
+
+A trait is used here to facilitate testing but typically the calls don't end up
```
fitzgen: "typically"? Is there ever such a case?

alexcrichton: If you don't enable optimizations it'll probably stick around as virtual, but with optimizations any trivial amount of inlining will make the dispatch go away.

PR #153, opened 2018-04-20
include fetch arg for wasm2es6js

FreeMasen:
This update to `wasm2es6js` would add the new flag `fetch` to the command line flags, in addition to the existing `--base64` flag. The project would build the same file as `--base64` with 2 key changes.

1. The `bytes` variable would not be declared or parsed into a `Buffer`/`Uint8Array`
2. The `export const booted` variable to be assigned a call to `fetch` who's response would be passed to `WebAssembly.Instantiate`

The reason for adding this is to allow for the chrome/webpack work-around to not increase the overall file size. With the `--base64` option the hello_world example build a `.js` file that is 322kb, a ~33% increase.


alexcrichton:
Thanks! I wonder though if the flag could be something like `--fetch /path/to/foo.wasm` instead of inferring it to be the same path on the filesytem? That may also provide a bit more flexibility for this in terms of deployment perhaps?

FreeMasen:
That shouldn’t be a problem. Should it be optional?

crates/cli-support/src/wasm2es6js.rs
```diff
@@ -180,35 +191,51 @@ impl Output {
...
+                    let bytes;
+                    const base64 = \"{base64}\";
+                    if (typeof Buffer === 'undefined') {{
+                        bytes = Uint8Array.from(atob(base64), c => c.charCodeAt(0));
+                    }} else {{
+                        bytes = Buffer.from(base64, 'base64');
+                    }}", base64 = base64::encode(&wasm)),
+                inst
+            )
+        } else if self.fetch_path.is_some() {
```
alexcrichton: Could this use `if let` to avoid the `unwrap` below?

examples/hello_world/chrome/build_fetch.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -ex
+
+# This is the same as the directory above this.
```
alexcrichton: Could the comments here be tweaked slightly to document what `--fetch` is and what the difference is from `./build.sh`?

PR #156, opened 2018-04-22
Fix Typescript definition of constructor arguments

jsheard:
TS uses `type[]` to denote arrays rather than Rusts `[type]`


alexcrichton:
Oops, thanks!

PR #161, opened 2018-04-25
Add support for version specifications

alexcrichton:
This commit adds a `#[wasm_bindgen(version = "...")]` attribute support. This
information is eventually written into a `__wasm_pack_unstable` section.
Currently this is a strawman for the proposal in ashleygwilliams/wasm-pack#101

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -886,6 +886,34 @@ controlling precisely how imports are imported and what they map to in JS. This
 section is intended to hopefully be an exhaustive reference of the
 possibilities!
 
+* `module` and `version` - we've seen `module` so far indicating where we can
+  import items from but `version` is also allowed:
+
+  ```rust
+  #[wasm_bindgen(module = "moment", version = "2.0.0")]
+  extern {
+      type Moment;
+      fn moment() -> Moment;
+      #[wasm_bindgen(method)]
+      fn format(this: &Moment) -> String;
+  }
+  ```
+
+  The `module` key is used to configure the module that each item is imported
+  from. The `version` key does not affect the generated wasm itself but rather
+  it's an informative directive for tools like [wasm-pack]. Tools like wasm-pack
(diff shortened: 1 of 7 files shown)


alexcrichton:
@ashleygwilliams ok I've updated with our discussion this morning:

* Imports that have a module name which starts with `./` cannot have `version` listed
* All other imports *must* have a version listed
* The custom section generated for wasm-pack only lists *package* names (in theory). The `module = "..."` directives may not be package names due to how imports sometimes work in JS. As a result, some postprocessing of the string specified to `module = "..."` is done to ensure that wasm-pack only sees package names, not erroneous names like `foo/bar/baz`
  * Names that don't start with `@` assume that everything up to the first `/` character (if any) is the package name
  * Names that start with `@` assume that everything up to the *second* `/` character (if any) is the package name

Pauan:
The example should use `#[wasm_bindgen(module = "moment", version = "^2.0.0")]`

Unlike Rust, npm does not default to semver (the string `"2.0.0"` matches the *exact* version `2.0.0`).

So if you want to use semver semantics you need to add a `^` at the start (which then causes it to match anything `>=2.0.0` and `<3.0.0`)

alexcrichton:
@Pauan ah indeed! That'll happen automatically here, but I can update example usage of this to recommend that. Currently wasm-bindgen doesn't interpret these strings at all, it just ferries them along

ashleygwilliams:
@alexcrichton @Pauan it's defnitely an interesting question what type of "semver defaults" we want to enforce given that npm and cargo have different ones. seems like a decision for `wasm-pack` to make, and i'm vaguely leaning towards keeping to the rust defaults in rust and translate those to npm one's just when we move the values to the `package.json` in the interest of keeping rust people in rust ecosystem defaults. would be curious what others thoughts are here and i'll probably open an issue on wasm pack about it

PR #171, opened 2018-04-27
Add JsValue::{from_serde, into_serde}

### LINKED ISSUE #96: Allow Serialize/Deserialize into a `JsValue`
We should have an option to convert an arbitrary object to a `JsValue` via `Serialize` an `Deserialize`. It may not be quick but it's often quite useful!

alexcrichton:
These functions are activated with the `serde-serialization` feature of the
`wasm-bindgen` crate. When activated they will allow passing any arbitrary value
into JS that implements the `Serialize` trait and receiving any value from JS
using the `Deserialize` trait. The interchange between JS and Rust is JSON.

Closes #96

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,12 @@ doctest = false
 default = ["spans", "std"]
 spans = ["wasm-bindgen-macro/spans"]
 std = []
+serde-serialize = ["serde", "serde_json", "std"]
 
 [dependencies]
 wasm-bindgen-macro = { path = "crates/macro", version = "=0.2.6" }
+serde = { version = "1.0", optional = true }
+serde_json = { version = "1.0", optional = true }
 
 [dev-dependencies]
 wasm-bindgen-cli-support = { path = "crates/cli-support", version = '=0.2.6' }
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -83,213 +83,230 @@ impl<'a> Context<'a> {
             }
         }
 
-        bail!("\n\nthe exported function `{}` is required to generate bindings \
+        bail!("the exported function `{}` is required to generate bindings \
                but it was not found in the wasm file, perhaps the `std` feature \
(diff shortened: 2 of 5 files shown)


udoprog:
Oh. This looks _really good_. So we serialize incoming objects and pass the string in, vice-versa for passing them out?

I think this effectively minimises the number of context switches and is the most pragmatic solution right now.

A potential future improvement could be to encode objects in a way which has a more efficient representation and supports skipping properties when decoding. But it will be hard to compete with the efficiency of the built-in JSON support.

alexcrichton:
@udoprog precisely! You basically got all the intentions there :)

David-OConnor:
Can someone provide an example of how to enable this? 
    wasm-bindgen = {version = "^0.2", features = [“serde-serialization”,]}

in Cargo.toml, and 
    #[cfg(feature = "serde-serialize")]
    extern crate serde;

both lead to this:
    error[E0433]: failed to resolve. Maybe a missing `extern crate wasm_bindgen;`?

alexcrichton:
@David-OConnor oh so after you enable the features in `Cargo.toml` you don't need the `#[cfg]` in your own code for serde, you should just be able to use `wasm-bindgen` as usual. Do you have an example I could poke around though?

alexcrichton:
Oops, thanks @jsheard!

David-OConnor:
Thank you for the help. I think I'm just not understanding how to enable the feature; I've attempted the above two, plus two additional attempts, eg]

    `[dependencies.wasm-bindgen]
version = "^0.2"
features = ["serde-serialize"],
(invalid TOML)
and 
`#![feature(proc_macro, wasm_custom_section, wasm_import_module,
   wasm_serde_serialize)]`
(Doesn't break anything like the other attempts, but doesn't expose from_serde.

alexcrichton:
@David-OConnor oh the trailing comma there is invalid syntax, but if you remove that you should be good!

David-OConnor:
Thank you! That worked.

PR #175, opened 2018-04-27
Enable `cargo test` where possible

### LINKED ISSUE #114: Adding `#[wasm_bindgen]` means you can't run `cargo bench` anymore
You get link errors and things like that.

Ideally, we should make all the wasm-bindgen generated code guarded by `#[cfg(target_arch = wasm32)]` or something like that.

The

```toml
[lib]
crate-type = ["cdylib"]
```

stuff is a little harder. It can be commented out just when running `cargo bench`, but that is clunky. Ideally that would not be necessary either.

alexcrichton:
Currently `#[wasm_bindgen]` generates a bunch of references to symbols that
don't actually exist on non-wasm targets, making it more difficult to get a
crate working across multiple platforms. This commit updates the symbol
references to be dummy ones that panic on non-wasm targets to allow simple
testing/benchmarking to work on native targets.

While this isn't a perfect solution for #114 it's probably as good as we can do
for now pending upstream Cargo features, so I'm gonna say that it...

Closes #114

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -155,6 +155,7 @@ impl ToTokens for ast::Struct {
             impl ::wasm_bindgen::__rt::core::convert::From<#name> for
                 ::wasm_bindgen::JsValue
             {
+                #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
                 fn from(value: #name) -> Self {
                     let ptr = ::wasm_bindgen::convert::IntoWasmAbi::into_abi(
                         value,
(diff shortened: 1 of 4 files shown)


alexcrichton:
cc @fitzgen, I'm curious what you think of this approach.

The tl;dr; is that wasm-bindgen should be fine to mix in with cross-platform code, but if you invoke any imported functionality (aka anything that would otherwise call out to JS) then it panics at runtime.

alexcrichton:
@fitzgen I think you'll still need to do something like `crate-type = ["cdylib", "rlib"]`, but otherwise it should be at least a big step towards making it work!

PR #177, opened 2018-04-29
update typescript option

TimBest:


alexcrichton:
Thanks for the PR! The behavior here changed in https://github.com/rustwasm/wasm-bindgen/pull/173 which actually reflects the behavior of master, but sounds like I should publish a new version!

PR #179, opened 2018-04-30
Add an example of `wasm2asm` and wasm-bindgen

alexcrichton:
This commit adds an example of executing the `wasm2asm` tool to generate asm.js
output instead of WebAssembly. This is often useful when supporting older
browsers, such as IE 11, that doesn't have native support for WebAssembly.


alexcrichton:
cc @sendilkumarn

sendilkumarn:
@alexcrichton seems good to me 👍

PR #183, opened 2018-05-01
Add support for mutable slices

### LINKED ISSUE #53: Passing mutable slice from JS to Rust
I am trying to reduce allocation overhead and avoid the use of `Vec`, which leads me to a C-style interface like this

```rust
#[wasm-bindgen]
pub fn expensive_work(input: &[u8], output: &mut [u8]) {
  // do expensive work
}
```

now currently this throws when trying to compile with a missmatched types error. 
Would it be possible to support something like this?

alexcrichton:
This commit adds support for mutable slices to pass the boundary between JS and
Rust. While mutable slices cannot be used as return values they can be listed as
arguments to both exported functions as well as imported functions.

When passing a mutable slice into a Rust function (aka having it as an argument
to an exported Rust function) then like before with a normal slice it's copied
into the wasm memory. Afterwards, however, the updates in the wasm memory will
be reflected back into the original slice. This does require a lot of copying
and probably isn't the most efficient, but it should at least work for the time
being.

The real nifty part happens when Rust passes a mutable slice out to JS. When
doing this it's a very cheap operation that just gets a subarray of the main
wasm memory. Now the wasm memory's buffer can change over time which can produce
surprising results where memory is modified in JS but it may not be reflected
back into Rust. To accomodate this when a JS imported function returns any
updates to the buffer are copied back to Rust if Rust's memory buffer has
changed in the meantime.

Along the way this fixes usage of `slice` to instead use `subarray` as that's
what we really want, no copying. All methods have been updated to use `subarray`
accessors instead of `slice` or constructing new arrays.

Closes #53

### DIFF
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -187,6 +187,12 @@ impl Descriptor {
                     _ => return None,
                 }
             }
+            Descriptor::RefMut(ref d) => {
+                match **d {
+                    Descriptor::Slice(ref d) => &**d,
+                    _ => return None,
+                }
+            }
             _ => return None,
         };
         match *inner {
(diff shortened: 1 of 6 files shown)


dignifiedquire:
Thank you

PR #191, opened 2018-05-06
Fix generated binding for functions returning structs.

### LINKED ISSUE #190: Functions returning structs are not working in no-modules mode
Smallest example that reproduces this bug:
```rust
#[wasm_bindgen]
pub struct S;

#[wasm_bindgen]
pub fn create_s() -> S {
    S
}
```

The generated JS file (with `--no-modules`) contains something like this:
```js
__exports.create_s = function() {
    return S.__construct(wasm.create_s());
}

__exports.S = class S {
    // ...
}
```

When I call the `create_s` function it fails with `ReferenceError: S is not defined`.
If I modify the JS file either to call `__exports.S.__construct` or export the class as `let S = null; __exports.S = S = class S { ... }` then it works as supposed.

The second way seems like a really easy fix, I'll create a pull request.

ngg:
This only affects --no-modules and --nodejs modes.

Fixes #190.


crates/cli-support/src/js/mod.rs
```diff
@@ -56,9 +56,17 @@ impl<'a> Context<'a> {
         let contents = deindent(contents);
         let contents = contents.trim();
         let global = if self.config.nodejs {
-            format!("module.exports.{} = {};\n", name, contents)
+            if contents.starts_with("class") {
+                format!("let {0} = null;\nmodule.exports.{0} = {0} = {1};\n", name, contents)
```
alexcrichton: To avoid the null could this perhaps instead be `let {0} = {1};\nmodule.exports.{0} = {0};\n`?

alexcrichton:
Oops sorry about that!

crates/cli-support/src/js/mod.rs
ngg: Even simpler solution: "{1}\nmodule.exports.{0} = {0};\n" (it seems to work perfectly)

Issue #204, opened 2018-05-21
Bug in exporting classes with --no-modules


alippai:
With --no-modules wasm-bindgen generates constructs like:
```
var __exports = {};
__exports.b = class X {}
function c() {
  return X.__construct...
}
```
Where calling `c()` you will get an _`X` is not defined_ error.
It should be:
```
function c() {
  return __exports.X.__construct...
}
```

PR #205, opened 2018-05-21
Add __exports. to exported class references when using --no-modules

alippai:
Fixing #204

### DIFF
--- a/crates/cli-support/src/js/js2rust.rs
+++ b/crates/cli-support/src/js/js2rust.rs
@@ -265,7 +265,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
 
         if let Some(name) = ty.rust_struct() {
             self.ret_ty = name.to_string();
-            self.ret_expr = format!("return {name}.__construct(RET);", name = name);
+            self.ret_expr = format!("return {export}{name}.__construct(RET);", name = name, export = if self.cx.config.no_modules { "__exports." } else { "" });
             return Ok(self)
         }
 


alippai:
This introduces a bug if not all classes are exported.

Note: I have no experience writing Rust or using wasm-bindgen, any feedback is appreciated :)

alippai:
The rust2js change is not needed, I'll update this PR soon

alexcrichton:
Thanks for the PR and the report! Recently though https://github.com/rustwasm/wasm-bindgen/pull/191 was targeted at fixing this, although it may not have fixed everything! Do you have an example I could build locally and poke around with to see what's going on?

alippai:
Oh, that was my colleague, I'm working on the same piece of code.
```
enum DigestImpl {
    Sha1(sha1::Sha1),

...

#[wasm_bindgen]
pub struct Digest(Option<DigestImpl>);

#[wasm_bindgen]
impl Digest {

...

#[wasm_bindgen]
pub fn sha1() -> Digest {
    Digest(Some(DigestImpl::Sha1(Default::default())))
}
```

alippai:
Generating code like this could work also:
```
class X {}
__exports.X = X;
function c() {
  return X.__construct();
}
```

alippai:
It was my bad wasm-bindgen-cli version :(

alippai:
Also I have to check how cargo works, I didn't expect it to be outdated (it was version 0.2.8 instead of 0.2.10).

alexcrichton:
Ah ok, glad to hear that it's working now!

alippai:
@alexcrichton does it make sense for the wasm-bindgen-cli to check the Cargo.lock file (if available) for the correct wasm-bindgen version? It could help with the newbie mistakes like mine.

alippai:
It is, thanks!

alexcrichton:
Hm yeah that could work! We could do something like try to learn what the `wasm-bindgen` version is in the background, and if successful print a warning if that looks to be newer than the local version

Issue #198, opened 2018-05-17
the trait `wasm_bindgen::convert::FromWasmAbi` is not implemented for `char`


kimtuck:
```
#[derive(Debug)]
#[wasm_bindgen]
pub struct Counter {
    key: char,
    count: i32
}
#[wasm_bindgen]
impl Counter {
    pub fn new(key: char, count: i32) -> Counter {
        Counter { key: key, count: count }
    }
}
```

Yields the message the trait `wasm_bindgen::convert::FromWasmAbi` is not implemented for `char`

alexcrichton:
Ah yes excellent point! I'm not sure though, what would the best translation for a `char` be in JS? A single-character string?

FreeMasen:
Could it be translated as a u16 and then converted on the JS side via `String.fromCharCode`?

FreeMasen:
Or even the `u32` paired with `String.fromCodePoint`

alexcrichton:
@FreeMasen er well the Rust side of things definitely can use `char`, so it's just what type this maps to in JS. JS doesn't really have a `char` so the closes thing seems to be a string with only one code point I guess?

FreeMasen:
Sorry, i think I misunderstood. 

On the JS side think it would be a `string`. I was alluding to how the `impl FromWasmAbi for char` might work. 

My thought was that `into_abi` could either be convert the `char` to a `string` and use the existing helper functions (`passStringToWasm`, `getStringFromWasm`) or cast the `char` to a `u32` and then implement a new helper function that uses the `fromCodePoint` static method and `codePointAt(0)` on the `String` type.

I still may be misunderstanding, sorry if that doesn't clear anything up.

Pauan:
@alexcrichton Just to be clear, a single code point might correspond to two JS characters:

```js
"\u{1F4A9}".length  // returns 2
```

This is because strings in JS are UTF-16 encoded, and JS does not attempt to hide that fact.

alexcrichton:
Ok I think it seems reasonable to expect that `char` should work and realistically the only equivalent in JS is a `string`. This won't protect against passing strings of more than one code point, but I don't think there's much we can do about that!

Seems good to me to implement this, and a PR would be most welcome!

alexcrichton:
Implemented in https://github.com/rustwasm/wasm-bindgen/pull/206

PR #206, opened 2018-05-21
add char support

FreeMasen:
this should address #198 
I have an example project I used to test this is working. I wasn't sure if it should be included.


alexcrichton:
Thanks for the PR! 

I think the values related to `char` on `JsValue` can be removed though in the sense that `JsValue` is attempting to be as close to a native JS value as it can. In that sense JS doesn't actually have a native character type, so it should be ok to remove.

Other than that mind adding some tests to exercise this functionality? Looks great other than those nits!

FreeMasen:
I will add the example project I used to test this working and remove the `JsValue` entries. I also appears that I need to move the changes into the `if_std!` block?

alexcrichton:
I think the import just needs to be from `core::char` instead of `std::char` and should otherwise be good to go

src/lib.rs
```diff
@@ -305,6 +305,8 @@ externs! {
     fn __wbindgen_is_undefined(idx: u32) -> u32;
     fn __wbindgen_boolean_new(val: u32) -> u32;
     fn __wbindgen_boolean_get(idx: u32) -> u32;
+    fn __wbindgen_char_new(val: u32) -> u32;
+    fn __wbindgen_char_get(idx: u32) -> u32;
```
alexcrichton: I think this can be removed since the `JsValue` methods were removed

crates/cli-support/src/js/mod.rs
```diff
@@ -216,6 +216,25 @@ impl<'a> Context<'a> {
...
+            Ok(format!("
+                function(n, invalid) {{
+                    let obj = getObject(n);
+                    if (typeof(obj) === 'number')
+                        return String.fromCodePoint(obj);
+                    getUint8Memory()[invalid] = 1;
+                    return 0;
+                }}
+            "))
+        })?;
```
alexcrichton: I think this can be removed since the `JsValue` methods were removed

tests/all/char.rs
```diff
@@ -0,0 +1,42 @@
...
+
+            use wasm_bindgen::prelude::*;
+
+            #[wasm_bindgen]
+            pub fn single_char() -> char { 'a' }
+            #[wasm_bindgen]
+            pub fn wide_char() -> char { '💩' }
+
+            #[wasm_bindgen]
+            pub fn parrot(c: char) -> char { c }
```
alexcrichton: Could this be expanded to also assert the value of the character coming in to Rust?

examples/char/.gitignore
```diff
@@ -0,0 +1,4 @@
+package-lock.json
+char.js
+char_bg.js
+char_bg.wasm
```
alexcrichton: Could this ignore `dist` as well? (I think a 1.2MB js file was checked in below)

FreeMasen:
I added a `README.md` for the example and updated the examples `README.md`. I trimmed the char project a little bit as well. I started to try and create a version in `WebAssembly.studio` but that would only be possible once these changes are merged. 

I also removed the large `js` file from the repo.

Once this is all ready did you want me to squash the interim commits into 'add char support'?

alexcrichton:
Looks great! And no need, nowadays with GitHub you can squash at merge time!

kimtuck:
@FreeMasen @alexcrichton Thanks for the fix.  I'm working on some samples using bindgen for a presentation; I could upload the examples someplace (via pull request) if there is a place here that makes sense

alexcrichton:
@kimtuck certainly! Either in this repo under `examples` or gists on https://webassembly.studio/ would be great!

PR #213, opened 2018-05-25
backend: All AST types should implement Debug

fitzgen:


fitzgen:
Pushed a commit to also implement `Eq`.

crates/backend/Cargo.toml
```diff
@@ -17,5 +17,5 @@ spans = ["proc-macro2/nightly"]
 quote = '0.6'
 proc-macro2 = "0.4"
 wasm-bindgen-shared = { path = "../shared", version = "=0.2.11" }
-syn = { version = '0.14', features = ['full', 'visit-mut'] }
+syn = { version = '0.14', features = ['extra-traits', 'full', 'visit-mut'] }
```
alexcrichton: Hm this runs the risk of increasing the compile time of `syn`, so do we want to do this by default?

PR #214, opened 2018-05-29
Add a WebIDL frontend

fitzgen:
This is still a work in progress, but I'd like to merge it so that we can collaborate on improving it.

Parse WebIDL source text and convert it into wasm-bindgen AST, so that we can automatically emit bindings for the types and functions described in the WebIDL.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ wasm-bindgen-cli-support = { path = "crates/cli-support", version = '=0.2.11' }
 [workspace]
 members = [
   "crates/cli",
+  "crates/webidl",
   "examples/hello_world",
   "examples/smorgasboard",
   "examples/console_log",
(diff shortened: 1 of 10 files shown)


alexcrichton:
Looking good! 

I'm wondering now though if we may want to tweak strategy slightly wrt the backend crate? I like the idea of using it but I think that we may want to eventually generate the attribute annotated code rather than expanded code? Right now the output of the attribute is somewhat unstable and changes in point releases but the attribute itself is the current stable interface 

If we do that then I think the tests here can be over the final token stream rather than the ast (which can be cumbersome)

Does that make sense? Do you think something like this may be a better approach? The code to go back to the attributes doesn't exist yet but I think the ast should have most if not all of the necessary information

fitzgen:
Alex and I just had a conversation about this. We decided that it is best to emit the same generated source text that the proc-macro produces (ie call to_tokens on the AST), not the `#[wasm_bindgen]` annotations. While this generated source text is more fragile than the annotations, we intend to emit it in `build.rs`, which side steps the fragilitiy issues since it will always use the current compilation's crate graph's wasm-bindgen, not the wasm-bindgen from when we published the crate.

alexcrichton:
👍 

🚅

PR #217, opened 2018-06-01
Implement `PartialEq` for `JsValue`

alexcrichton:
Dispatch to JS's `===` operator internally

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -300,6 +300,15 @@ impl<'a> Context<'a> {
             "))
         })?;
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
Hm perhaps yeah, I'm hoping we can stick with `===` as it's a pretty simple definiton and it corresponds to exactly what JS does when you use syntactical equality (mostly)

fitzgen:
To be clear I wasn't suggesting we support structural equality, just that we have a test asserting that we match `===`'s behavior.

alexcrichton:
Oh oops sorry, sure can add a test for that!

PR #219, opened 2018-06-01
Fix tests by pinning versions of webpack

alexcrichton:
Looks like a newer version of webpack has broken tests, so let's use
package-lock.json to pin to an older version while we wait for a fix.


xtuc:
Yes I saw that and I assumed that it came from rust nightly because only one build is failing?

But I'm sure that we introduced a regression in the latest versions of Webpack anyway.

PR #220, opened 2018-06-01
Optimize JsValue::{from_bool, undefined, null} constructors

alexcrichton:
This commit optimizes constructing an instance of `JsValue` which is one of
`null`, `undefined`, `true`, or `false`. These are commonly created on the Rust
side of things and since there's only a limited set of values we can easily
prepopulate the global slab with a few entries and use hardcoded indices to
refer to these constants. This should avoid the need to travel into JS to insert
a `null` or and `undefined` into the global slab.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -690,17 +690,23 @@ impl<'a> Context<'a> {
         if !self.exposed_globals.insert("slab") {
             return;
         }
-        self.global(&format!("let slab = [];"));
+        let initial_values = [
+            "{ obj: null }",
+            "{ obj: undefined }",
+            "{ obj: true }",
+            "{ obj: false }",
+        ];
+        self.global(&format!("let slab = [{}];", initial_values.join(", ")));
         if self.config.debug {
-            self.export("assertSlabEmpty", "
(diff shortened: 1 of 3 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -690,11 +690,16 @@ impl<'a> Context<'a> {
...
+            { obj: null },
+            { obj: undefined },
+            { obj: true },
+            { obj: false },
+        ];");
         if self.config.debug {
             self.export("assertSlabEmpty", "
                 function() {
-                    for (let i = 0; i < slab.length; i++) {
+                    for (let i = 4; i < slab.length; i++) {
```
fitzgen: Pull `4` out into a constant so its less magical? Or maybe:

```rust
const INITIAL_SLAB_ENTRIES: &[&str] = &[
    "{ obj: null }",
    // ...
];

self.global(format!("let slab = [{}]", INITIAL_SLAB_ENTRIES.join(",\\n")));

self.export("assertSlabEmpty", format!("
    ...
    for (let i = {}; i < slab.length; i++) {{
        ...
    }}
", INITIAL_SLAB_ENTRIES.len()));
```

PR #223, opened 2018-06-01
webidl: emit methods for regular operations on interfaces

fitzgen:
@alexcrichton you were right that ASTs would get too gnarly to meaningfully test...

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -1,4 +1,4 @@
-use proc_macro2::{TokenTree, TokenStream, Ident, Span};
+use proc_macro2::{Ident, Span, TokenStream, TokenTree};
 use quote::ToTokens;
 use shared;
 use syn;
@@ -121,7 +121,12 @@ pub enum TypeLocation {
 }
 
 impl Program {
-    pub fn push_item(&mut self, item: syn::Item, opts: Option<BindgenAttrs>, tokens: &mut TokenStream) {
+    pub fn push_item(
+        &mut self,
+        item: syn::Item,
+        opts: Option<BindgenAttrs>,
+        tokens: &mut TokenStream,
+    ) {
         match item {
             syn::Item::Fn(mut f) => {
                 let opts = opts.unwrap_or_else(|| BindgenAttrs::find(&mut f.attrs));
@@ -309,7 +314,10 @@ impl Program {
                 BindgenAttrs::find(attrs)
             };
             let module = item_opts.module().or(opts.module()).map(|s| s.to_string());
-            let version = item_opts.version().or(opts.version()).map(|s| s.to_string());
+            let version = item_opts
+                .version()
+                .or(opts.version())
+                .map(|s| s.to_string());
             let js_namespace = item_opts.js_namespace().or(opts.js_namespace()).cloned();
             let mut kind = match item {
                 syn::ForeignItem::Fn(f) => self.push_foreign_fn(f, item_opts),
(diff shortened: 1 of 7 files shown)


crates/backend/src/ast.rs
```diff
@@ -802,116 +807,94 @@ impl BindgenAttrs {
...
         })
     }
 
     fn constructor(&self) -> bool {
-        self.attrs.iter().any(|a| {
-            match a {
-                BindgenAttr::Constructor => true,
-                _ => false,
-            }
+        self.attrs.iter().any(|a| match a {
```
alexcrichton: Hm is there a way we can disable this in rustfmt? This seems like a bug?

crates/webidl/src/lib.rs
```diff
@@ -65,23 +66,31 @@ fn compile_ast(ast: &backend::ast::Program) -> String {
...
+            def.webidl_parse(program, ())?;
         }
         Ok(())
     }
 }
 
-impl WebidlParse for webidl::ast::Definition {
-    fn webidl_parse(&self, program: &mut backend::ast::Program) -> Result<()> {
+impl<'a> WebidlParse<'a> for webidl::ast::Definition {
+    type Extra = ();
```
alexcrichton: Hm once you got down the route of associated types here and `()` all over the place it may just be best to have a `Context` struct with shared state and a bunch of methods for each recursive step. Either way's fine by me, but I tend to shy away from traits too aggressively historically

crates/backend/src/ast.rs
fitzgen: I can undo it certainly. Not sure there is a way to tell rustfmt not to do it in the first place, and I'm not really a fan of deviating from rustfmt's behavior anyways. /me shrugs

crates/webidl/src/lib.rs
```diff
@@ -118,7 +133,234 @@ impl WebidlParse for webidl::ast::NonPartialInterface {
...
+        inputs.push(simple_fn_arg(
+            proc_macro2::Ident::new("self_", proc_macro2::Span::call_site()),
+            self_ref_ty.clone(),
+        ));
+        arguments.push(self_ref_ty);
+
+        for arg in &self.arguments {
+            if arg.optional || arg.variadic {
+                // We don't support optional or variadic functions yet; skip
+                // bindings for this this whole function.
```
alexcrichton: I wonder if for debugging purposes at some point if it'd be good to log whenever a function is skipped with a brief reason as to why?

fitzgen: Yeah, I was wondering about this too. Do you have a preferred method of logging? `env_logger`?

crates/webidl/tests/expected/Event.rs
```diff
@@ -53,6 +53,124 @@ impl From<Event> for ::wasm_bindgen::JsValue {
...
+    use wasm_bindgen::describe::*;
+    inform(FUNCTION);
+    inform(1u32);
+    <&Event as WasmDescribe>::describe();
+    inform(0);
+}
+impl Event {
+    #[allow(bad_style)]
+    #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
+    pub extern "C" fn stopPropagation(&self) {
```
alexcrichton: Hm this looks a little fishy (the `extern "C"`), although it may be a preexisting bug, I'm not entirely sure

crates/backend/src/ast.rs
alexcrichton: Nah I know what you mean. I specifically really dislike this style of formatting because when you read the body you have no idea it's in ` match` unless you go back up and read it (and the `match` is hidden with the closure), so I'm sort of hoping this is a bug in rustfmt that'll get fixed at some point...

crates/webidl/src/lib.rs
alexcrichton: Sure yeah, that should work!

crates/webidl/src/lib.rs
fitzgen: I like `Context` for shared state that is uniform across all `impl`s and `Self::Extra` for things that change between different `impl`s.

I find that traits help me clarify my thinking about what is going on, even if they end up varying a bit because of the `Extra` type.

crates/webidl/src/lib.rs
fitzgen: Will add logging in a follow up PR.

PR #225, opened 2018-06-01
webidl: Add logging and use env_logger in the tests

fitzgen:


crates/webidl/Cargo.toml
```diff
@@ -12,14 +12,21 @@ name = "webidl-expected"
 path = "tests/expected/lib.rs"
 
 [dev-dependencies]
+diff = "0.1.11"
+env_logger = "0.5.10"
 wasm-bindgen = { version = "=0.2.11", path = "../..", default-features = false }
 wasm-bindgen-backend = { version = "=0.2.11", path = "../backend", features = ["extra-traits"] }
-diff = "0.1.11"
+# Force the `log` dependency for testing.
+wasm-bindgen-webidl = { version = "=0.2.11", path = ".", features = ["log"] }
```
alexcrichton: Oh dear now that's something that was never intended to be allowed in Cargo... Ah well!

crates/webidl/Cargo.toml
```diff
@@ -12,14 +12,21 @@ name = "webidl-expected"
...
 
 [dependencies]
 failure = "0.1"
+log = { optional = true, version = "0.4.1" }
```
alexcrichton: The log crate is pretty lightweight, but was there a specific reason you were thinking this should be optional?

fitzgen: Just that it seems unnecessary if you don't need it. Happy to simplify the toml config if you don't think this is worth the headache.

PR #228, opened 2018-06-04
Add prototype of wasm-bindgen-typescript

spastorino:
This is a first prototype of wasm-bindgen-typescript. The idea is to parse TypeScript definition files and be able to output Rust code from it.
This is kind of WIP, though it work in some cases but far for a finished and polished thing. There are still some things to fix and do.

This thing first use `api-extractor` (it's an npm package and needs to be installed) to parse the typescript definition file you want to process and saves a json file with the definitions. After that it parses this json file and ends filling wasm-bindgen IR. There's a main (which probably should not be in the project :smile:) to run the whole thing and print the Rust source on screen.

### Todo
- [ ] It needs to cover all the different data structures that can come on TypeScript files, in particular https://github.com/rustwasm/wasm-bindgen/commit/fa8961e56a1dae0ab8696c5a949dea5706cf2480#diff-7837f9f1900f30ff96bd2bbfe07a3f4eR30
- [ ] It needs to properly convert from JS types to Rust types https://github.com/rustwasm/wasm-bindgen/commit/fa8961e56a1dae0ab8696c5a949dea5706cf2480#diff-3377e08c4ce34b2cb783497ffd1b2e1eR127
- [ ] It needs to run api-extractor passing the -c flag and the right location of api-extractor.json file
- [ ] It needs a lot of test cases to be sure that we end covering all the possible TypeScript constructions
- [ ] Check out tsconfig.json and see if there's something we can do better
- [ ] Check api-extractor.json and see if there's something we can do better
- [ ] It would be nice to evaluate https://github.com/SimplrJS/ts-extractor and consider replacing api-extractor with it. ts-extractor cover more syntaxis and it's more flexible from what I've seen.
- [ ] Probably we would like to refactor this a bit, maybe use a visitor, or implement serialize in a smarter way and dump directly to IR doing smarter stuff if possible or something like that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ wasm-bindgen-cli-support = { path = "crates/cli-support", version = '=0.2.11' }
 [workspace]
 members = [
   "crates/cli",
+  "crates/typescript",
   "crates/webidl",
   "examples/hello_world",
   "examples/smorgasboard",
--- a/crates/typescript/Cargo.toml
+++ b/crates/typescript/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "wasm-bindgen-typescript"
+version = "0.1.0"
+authors = ["Santiago Pastorino <spastorino@gmail.com>"]
+
+[dependencies]
+serde = "1.0"
+serde_derive = "1.0"
+serde_json = "1.0"
+
+proc-macro2 = "0.4"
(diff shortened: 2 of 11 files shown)


fitzgen:
Note: we should also enable tests/building this crate in travis CI -- see the `.travis.yml` file for how its done for the `webidl` crate, for example.

PR #230, opened 2018-06-06
Bump webpack

xtuc:


examples/add/package.json
```diff
@@ -3,7 +3,7 @@
     "serve": "webpack-dev-server"
   },
   "devDependencies": {
-    "webpack": "^4.8.1",
+    "webpack": "^4.11.1",
```
sendilkumarn: `^` will take care of that right. Why bother changing

xtuc: that's true but I just ran a sed on the repo for consistency.

PR #264, opened 2018-06-11
webidl: add initial support for integration tests

ohanar:


tests/all/main.rs
```diff
@@ -209,8 +196,150 @@ impl Project {
...
+        );
+        self.files.push(("run.js".to_string(), runjs));
+    }
+
+    fn ensure_test_entry(&mut self) {
+        if self
+            .files
+            .iter()
+            .find(|(path, _)| path == "test.ts" || path == "test.js")
+            .is_none()
```
fitzgen: This can be cleaned up a little as:

```rust
!self.files.iter().any(|(path, _)| path == "test.ts" || path == "test.js")
```

tests/all/main.rs
```diff
@@ -209,8 +196,150 @@ impl Project {
         self
     }
 
+    fn generate_bindings(&mut self) -> Vec<PathBuf> {
```
fitzgen: Nitpick: to be a little more clear, I think we should name this `generate_webidl_bindings`, since this whole project involves bindings of some sort

ohanar:
@fitzgen Nitpicks addressed :-)

fitzgen:
Thanks again!

Issue #57, opened 2018-03-06
Copy doc comments from Rust to JS


fitzgen:
Would be good for JS programmers reading the generated JS to have documentation on

* Classes
* Methods
* Functions
* Anything else that gets emitted and has a doc comment in the Rust version

FreeMasen:
Would that require a modification to #23 to include the doc comment strings?

alexcrichton:
Indeed! That's fine though, the format isn't specified yet :)

FreeMasen:
After #265 we are now copying doc comments from Rust to JS for the following using the [JSDoc](http://usejsdoc.org/) format. 

- Classes
- Methods
- Public Attributes
- Functions

The only other doc comments that I can think of would be module level comments. 

One of the nice things about the JSDoc comments is that they define the tooltip content when working in VSCode. I would ideally like to see the following annotated lines added to our current translation.

- `@constructor`: When using `no-modules` this would be added to the prototype function
- `@param {type} name`: This would be added for each parameter in a function signature
- `@returns {type}`: This would be added to functions and methods that return non-unit types
  - We could also have unit returns annotated with `void`

Both `@param` and `@returns` optionally can have a description added to them, I do not believe that Rust Doc comments have this feature?

PR #265, opened 2018-06-11
Copy doc comments from Rust to JS

FreeMasen:
Addresses #57

Still to do:
- add tests
- add new example to travis test script

optional additions
- Add comments to .ts file
- Add type annotations for `jsdoc` comments
  - `@param {*} [name]`
  - `@returns {*}`


fitzgen:
@FreeMasen let me know when this is ready for review!

FreeMasen:
@fitzgen This is ready for review.

Let me know if you would like the additional 2 items I listed above

crates/backend/src/ast.rs
```diff
@@ -150,13 +155,29 @@ impl Program {
...
+                                            a.tts.clone().into_iter().filter_map(|t| match t {
+                                                TokenTree::Literal(lit) => Some(format!("{}", lit)),
+                                                _ => None,
+                                            })
+                                        )
+                                    } else {
+                                        None
+                                    }
+                                })
+                                .fold(vec![], |mut acc, a| {acc.extend(a); acc});
```
fitzgen: This should be a call to `extract_doc_comments`, right?

crates/backend/src/ast.rs
```diff
@@ -1072,3 +1105,24 @@ fn replace_self(name: &Ident, item: &mut syn::ImplItem) {
 
     syn::visit_mut::VisitMut::visit_impl_item_mut(&mut Walk(name), item);
 }
+/// Extract the documentation comments from a Vec of attributes
+fn extract_doc_comments(attrs: &Vec<syn::Attribute>) -> Vec<String> {
+    attrs
+    .iter()
+    .filter_map(|a| {
+        //We only care about outer comments for now
+        if a.style == AttrStyle::Outer {
```
fitzgen: Shouldn't this also check that the attributes are `doc`? It seems to me that this will collect non-`doc` attributes as well, which we don't want.

crates/backend/src/ast.rs
```diff
@@ -1072,3 +1105,24 @@ fn replace_self(name: &Ident, item: &mut syn::ImplItem) {
 
     syn::visit_mut::VisitMut::visit_impl_item_mut(&mut Walk(name), item);
 }
+/// Extract the documentation comments from a Vec of attributes
+fn extract_doc_comments(attrs: &Vec<syn::Attribute>) -> Vec<String> {
```
fitzgen: Is there a reason this needs to take `&Vec<syn::Attribute>` rather than `&[syn::Attribute]`? If not, then we should switch to the latter.

examples/comments/README.md
```diff
@@ -0,0 +1 @@
+# TODO
```
fitzgen: Please fill this out :)

tests/all/comments.rs
```diff
@@ -0,0 +1,56 @@
...
+                    self.a.clone()
+                }
+            }
+        "#);
+
+        let (root, target) = p.build();
+        p.gen_bindings(&root, &target);
+        let js = p.read_js();
+        let comments = extract_doc_comments(&js);
+        ///
```
fitzgen: Was there supposed to be something here?

crates/backend/src/ast.rs
FreeMasen: Good catch, I am pretty new at using `syn`/`proc-macro2`, I updated this to check the path segments for a "doc" `Ident` instead of just capturing outer attributes. Let me know if you know of a better way to differentiate doc comment attributes.

crates/backend/src/ast.rs
FreeMasen: yep, this has been updated

crates/backend/src/ast.rs
FreeMasen: There is not reason, just updated it

examples/comments/README.md
FreeMasen: sorry, I think this was from an older commit it does have content.

tests/all/comments.rs
FreeMasen: nope, not sure how this got there, it has been removed. I also updated the test to handle white-space a little better

FreeMasen:
I believe I have addressed all of you comments. Sorry about the commits, I will work on trying to break them apart a little better moving forward or just squash them all down.

Issue #89, opened 2018-03-29
Pretty-ify generated JS code


alexcrichton:
Right now it has crazy indentation, we should make it pretty!

@ashleygwilliams says we should use `prettier`!

corbinu:
@alexcrichton does this mean the CLI will depend on Node being installed? 

If so how would you feel if we modify the output to be TypeScript and simply run it through the TSC into JavaScript if people don't specify --typescript? 

That way we get the declaration file automatically.

alexcrichton:
@corbinu hm perhaps! I'm not actually too sure how this would work, I'm pretty unfamiliar with `pretty` myself. I think for now we'd probably want to shell out to something rather than embedding a JS interpreter in wasm-bindgen itself, but we can also probably gracefully handle the case where it's not available and just opportunistically do it.

ashleygwilliams:
i was thinking that we might just first implement it for when tests are run as that's really the time where a prettified output would be appreciated. post that i think there might be some fun compile stuff we might be able to do to make it work, or there's always just porting it entirely to rust ;)

corbinu:
Ok no worries. I can put something together that shells out

corbinu:
Ok so this is what I have so far. 

I am building a Node cli app that will use TypeScript and Prettier to do a bunch of the work. 

However I also have been testing and I created a new bindgen for Duktape. Looks like I can get the JS to run inside of Duktape as well. (looks like I can keep that app under 5mb but don't think it should be directly included as most won't need it). 

That way the cli can try these things in order:

1. See if the Node companion CLI is installed and use that
2. See if npx with an acceptable version of node is installed and if so run the node companion CLI with npx
3. See if the Duktape wrapped version is installed and use that
4. Print out that the user should either download node or install the wrapped cli with something like "cargo install wasm-bindgen-cli-js-fallback"

I should have a MVP by the start of next week.

alexcrichton:
Awesome! Let me know if you need any help @corbinu

corbinu:
Will do! Still experimenting but going to need a lot of feedback.

alexcrichton:
@kwonoj it's true yeah we probably wouldn't want to ship node with `wasm-bindgen`, but that being said I think it'd be possible to opportunistically try to prettyify code and silently avoid that if it's otherwise not available.

Another option here is to simply generate prettier code! I've added a few commits to do this historically, and there's just a few places left that produce ugly code by default I believe

[...]

PR #268, opened 2018-06-14
Format

FreeMasen:
To address #89.

note: this has a shared base with #165, I can rebase it if needed

I have updated the overall approach to fixing this problem. Instead of attempting to get our indentation correct on smaller pieces, I have moved this to a single step over a whole js module. 

I then ran [eslint](https://eslint.org/) against each of the out.js files generated by `cargo test`. Any errors that were reported I attempted to fix them. Since this approach seemed valuable I added an eslint step to the travis test that will review all `out.js` and `out_bg.js` files for formatting. Adding an eslint test should help reduce any incorrectly formatted code moving forward.


fitzgen:
@FreeMasen can you rebase and squash this PR? Thanks :)

FreeMasen:
@fitzgen rebased and squashed

crates/cli-support/src/js/js2rust.rs
```diff
@@ -1,6 +1,6 @@
 use failure::Error;
 
-use super::{indent, Context};
+use super::{Context};
```
fitzgen: Nitpick: when there is a single import, it is idiomatic style to drop the `{}`:

```rust
use super::Context;
```

crates/cli-support/src/js/mod.rs
```diff
@@ -1748,31 +1744,31 @@ impl<'a, 'b> SubContext<'a, 'b> {
...
+//         .unwrap_or(0);
+//     let mut ret = String::new();
+//     for line in s.lines() {
+//         if !line.trim().is_empty() {
+//             ret.push_str(&line[amt_to_strip..]);
+//         }
+//         ret.push_str("\n");
+//     }
+//     ret
+// }
```
fitzgen: Let's remove this code if it is not used anymore.

crates/cli-support/src/lib.rs
```diff
@@ -380,3 +380,28 @@ impl fmt::Display for MyError {
...
+fn indent_recurse(mut lines: ::std::str::Lines, current_indent: usize) -> String {
+    if let Some(line) = lines.next() {
+        let mut trimmed = line.trim().to_owned();
+        let mut next_indent = current_indent;
+        let mut current_indent = current_indent;
+        if trimmed.ends_with('{') {
+            next_indent += 1;
+        }
+        if trimmed.starts_with('}') || trimmed.ends_with('}') {
+            if current_indent > 0 { current_indent -= 1; }
```
fitzgen: Nitpick: we shouldn't put `if` blocks on the same line as their condition:

```rust
if current_index > 0 {
    current_index -= 1;
}
```

.eslintrc.js
```diff
@@ -0,0 +1 @@
+module.exports = { env: { es6: true, browser: true, commonjs: true, node: true }, extends: 'eslint:recommended', parserOptions: { sourceType: 'module' }, rules: { indent: [ 'error', 4 ], 'linebreak-style': [ 'error', 'unix' ], quotes: [ 'error', 'single' ], semi: [ 'error', 'always' ], 'no-console': 0, 'no-undef': 'warn', 'no-unused-vars': 'warn' } };
```
fitzgen: Can we pretty print this configuration object so it is easier to read?

crates/cli-support/src/js/mod.rs
```diff
@@ -120,8 +119,7 @@ impl<'a> Context<'a> {
             Ok(format!("
                 function(idx) {{
                     // If this object is on the stack promote it to the heap.
-                    if ((idx & 1) === 1)
-                        return addHeapObject(getObject(idx));
+                    if ((idx & 1) === 1) return addHeapObject(getObject(idx));
```
fitzgen: I'd prefer we add curly braces to these one-statement blocks but we can bikeshed style and which eslint rules we want to subscribe to in follow up issues.

FreeMasen: I tried to one-line anything that felt short enough, otherwise I added curly braces. ESLint doesn't actually care about that but the very simple re_indent function I wrote was choking on the second line.

.eslintrc.js
FreeMasen: yep, done!

crates/cli-support/src/lib.rs
FreeMasen: fixed!

crates/cli-support/src/js/mod.rs
FreeMasen: oops, sorry about that, this has been removed

crates/cli-support/src/js/js2rust.rs
FreeMasen: fixed!

PR #272, opened 2018-06-15
webidl: add support for static methods and attributes

### LINKED ISSUE #251: Support emitting getters for static WebIDL attributes
https://heycam.github.io/webidl/#idl-attributes

Given

```webidl
interface Foo {
  static attribute long bar;
};
```

we should emit a `fn get_bar() -> i32` method to get `Foo`'s static `bar` attribute.

This is similar to #250 except that there is no `&self` parameter.

ohanar:
Fixes #251.

### DIFF
--- a/tests/all/webidl/simple.rs
+++ b/tests/all/webidl/simple.rs
@@ -176,3 +176,59 @@ fn named_constructor() {
         )
         .test();
 }
+
+#[test]
+fn static_method() {
+    project()
+        .file(
+            "foo.webidl",
+            r#"
+            interface Foo {
+                [Pure]
+                static double swap(double value);
+            };
+        "#,
+        )
+        .file(
+            "foo.ts",
+            r#"
+            export class Foo {
+                private static value: number = 0;
+                static swap(value: number): number {
+                    const res = Foo.value;
+                    Foo.value = value;
+                    return res;
+                }
+            }
+            "#,
+        )
+        .file(
+            "src/lib.rs",
+            r#"
+            #![feature(proc_macro, wasm_custom_section, wasm_import_module)]
+
+            extern crate wasm_bindgen;
+
+            use wasm_bindgen::prelude::*;
+
+            pub mod foo;
+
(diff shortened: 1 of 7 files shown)


tests/all/webidl/simple.rs
```diff
@@ -176,3 +176,59 @@ fn named_constructor() {
...
+
+            use wasm_bindgen::prelude::*;
+
+            pub mod foo;
+
+            use foo::Foo;
+
+            #[wasm_bindgen]
+            pub fn test() {
+                let tmp = Foo::swap(3.14159) == 0.;
```
fitzgen: Nitpick: we should use `assert_eq!(..)` here

tests/all/webidl/simple.rs
```diff
@@ -176,3 +176,59 @@ fn named_constructor() {
...
+                assert!(tmp);
+                let tmp = Foo::swap(2.71828) == 3.14159;
+                assert!(tmp);
+                let tmp = Foo::swap(2.71828) != 3.14159;
```
fitzgen: And `assert_ne!(..)` here

fitzgen:
@ohanar, would you be interested in joining as a collaborator for wasm-bindgen? Specifically with regards to the webidl frontend, since you've now written most of the code that makes it up as well as its testing infrastructure :)

Responsibilities involved are basically what's listed [here](https://github.com/rustwasm/team/blob/master/GOVERNANCE.md#repositories), and specifically reviewing pull requests related to webidl from time to time.

ohanar:
Nitpick addressed. There is a weird issue (at least on my machine) where, in the testing framework, asserts don't ever fail without binding to a temporary variable. I had assumed this extended to `assert_eq` and `assert_ne`, but it doesn't seem to.

I would be happy to be a collaborator. As you said it sounds like my main responsibility would be in reviewing pull requests, which I am happy to do.

fitzgen:
Awesome :) Invite sent!

PR #274, opened 2018-06-18
Expose objects and functions from the JavaScript global scope

fitzgen:
These are bindings to JavaScript's standard, built-in objects and their methods and properties.

This does *not* include any Web, Node, or any other JS environment APIs. Only the things that are guaranteed to exist in the global scope by the ECMAScript standard. They are available in the `wasm_bindgen::js` module when the `js_globals` feature is enabled.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects

This is intentionally an incomplete set of imports, and I intend to create a big issue with a big checklist of everything else that we need to add.

@ohanar @spastorino if one of you could review this, that would be very appreciated :)

### DIFF
--- a/tests/all/js_globals.rs
+++ b/tests/all/js_globals.rs
@@ -1,0 +1,150 @@
+// Keep these tests in alphabetical order, just like the imports in `src/js.rs`.
+
+use super::project;
+
+#[test]
+#[cfg(feature = "std")]
+fn decode_uri() {
+    project()
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn test() {
+                let x = js::decode_uri("https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B")
+                    .ok()
+                    .expect("should decode URI OK");
+                assert_eq!(x, "https://mozilla.org/?x=шеллы");
+
+                assert!(js::decode_uri("%E0%A4%A").is_err());
+            }
+        "#)
+        .file("test.ts", r#"
+            import * as wasm from "./out";
+
+            export function test() {
+                wasm.test();
+            }
+        "#)
+        .test();
+}
+
+#[test]
+#[cfg(feature = "std")]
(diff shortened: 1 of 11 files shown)


fitzgen:
> This is intentionally an incomplete set of imports, and I intend to create a big issue with a big checklist of everything else that we need to add.

And here is the big issue with a big checklist: https://github.com/rustwasm/wasm-bindgen/issues/275

tests/all/js_globals.rs
```diff
@@ -0,0 +1,150 @@
...
+            pub fn test() {
+                let x = js::decode_uri("https://mozilla.org/?x=%D1%88%D0%B5%D0%BB%D0%BB%D1%8B")
+                    .ok()
+                    .expect("should decode URI OK");
+                assert_eq!(x, "https://mozilla.org/?x=шеллы");
+
+                assert!(js::decode_uri("%E0%A4%A").is_err());
+            }
+        "#)
+        .file("test.ts", r#"
```
ohanar: This `test.ts` isn't needed anymore. The changes I made for WebIDL will automatically generate a `test.ts` equivalent to this one if no `test.ts` is found.

tests/all/js_globals.rs
```diff
@@ -0,0 +1,150 @@
...
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn test() {
+                let x = js::encode_uri("ABC abc 123");
+                assert_eq!(x, "ABC%20abc%20123");
+            }
+        "#)
+        .file("test.ts", r#"
```
ohanar: Same goes for this one.

tests/all/js_globals.rs
```diff
@@ -0,0 +1,150 @@
...
+
+            export function test() {
+                wasm.test();
+            }
+        "#)
+        .test();
+}
+
+#[allow(non_snake_case)]
+mod Object {
```
ohanar: Rather than adding an inline module, maybe a new directory should be added? (Especially in preparation for more to be added to `js_globals`.)

tests/all/js_globals.rs
fitzgen: Good catch! I forgot about that :)

spastorino:
I've arrived a bit late to the party but this looks perfect!!!

PR #276, opened 2018-06-19
Add a guide, change log, enumerate team members, release template, etc

fitzgen:
A bunch of organizational / administrative tidy ups :)


fitzgen:
Sounds like a good plan to me! Thanks :)

guide/src/what-else-can-we-do.md
```diff
@@ -0,0 +1,148 @@
...
+  if (a !== b)
+    throw new Error(`${a} != ${b}`);
+  console.log(`found ${a} === ${b}`);
+}
+
+function main() {
+  assertEq(concat('a', 'b'), 'ab');
+
+  // Note the `new Foo()` syntax cannot be used, static function
+  // constructors must be used instead. Additionally objects allocated
```
konstin: You can use `new Foo()` since #127 by annotating the new method with `#[wasm_bindgen(constructor)]`.

fitzgen: Woops, I literally just cut up the README into pieces and turned it into an mdbook -- would you like to send a PR fixing this incorrectness?

PR #278, opened 2018-06-20
Js binding for to_string

jonathan-s:
A JS binding for to_string. 
Rust is still fairly new to me. So there is one thing I was wondering. 

I tried adding the code below, but got a compiler error, saying that it doesn't find the function from_serde in JsValue. But from what I can see, from_serde is part of `JsValue`?
```
#[wasm_bindgen]
pub fn test() {
    let json = JsValue::from_serde("{test: 2}").unwrap();
    let object = js::Object{ obj: json };
    assert_eq!(object.to_string(), "[object Object!]");
}
```


src/js.rs
```diff
@@ -85,4 +85,11 @@ extern {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
     #[wasm_bindgen(method, js_name = hasOwnProperty)]
     pub fn has_own_property(this: &Object, property: &str) -> bool;
+
+    /// The toString() method returns a string representing the object.
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString
```
fitzgen: Nitpick: add a newline before the URL, like the other comments.

jonathan-s:
I managed to create a new JS object in Rust. From my understanding, you wanted to create a non-empty JS object in rust, you would have to expand the struct to be able to contain values? 

Or how would you assign values to the object?

fitzgen:
> I managed to create a new JS object in Rust. From my understanding, you wanted to create a non-empty JS object in rust, you would have to expand the struct to be able to contain values?

To support getting and setting arbitrary properties, we would need more bindings to things like `Reflect.get`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/get

jonathan-s:
Thanks for the review and the insight! A change coming right up.

tests/all/js_globals/Object.rs
```diff
@@ -54,3 +54,35 @@ fn has_own_property() {
...
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn to_string(obj: &js::Object) -> String {
+                obj.to_string()
+            }
+
+            #[wasm_bindgen]
+            pub fn test() {
```
fitzgen: Oops, should have caught this earlier. The `test` function in a `.rs` is only called by default if you don't supply a `test.ts` file. So in this case, this rust `test` is never called. This should be added to the `test.ts` file:

```js
wasm.test();
```

jonathan-s: Good call! The error message that you get from wasm, is a lot less user friendly. The sasert statement in rust should now run and pass.

fitzgen: Yes, this is what https://github.com/rustwasm/console_error_panic_hook is for, but we haven't hooked it up into the tests. I'll file an issue!

PR #280, opened 2018-06-20
attempts to implement Promise constructor

belfz:
The closure-taking functions/constructors seem to be quite difficult to deal with; especially in `Promise` constructor, where a closure takes a form of:
```js
new Promise( /* executor */ function(resolve, reject) { ... } );
```
where `resolve` and `reject` are callable.

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -16,6 +16,7 @@
 //! example, `decodeURI` in JavaScript is exposed as `decode_uri` in these
 //! bindings.
 
+use closure::*;
 use wasm_bindgen_macro::*;
 use JsValue;
 if_std! {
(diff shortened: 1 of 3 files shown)


belfz:
This one fails too, but I tried using `Box` for `resolve` and `reject` args.

The error is:
```bash
92 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ the trait `convert::FromWasmAbi` is not implemented for `std::boxed::Box<core::ops::FnMut(JsValue)>`
```

ohanar:
I don't believe we have any support for passing JS callbacks to rust at the moment, so I don't think you can deal with the `resolve` and `reject` arguments for now.

belfz:
Thanks @ohanar, that's what I was afraid of. I'll wait for confirmation from @fitzgen, and if it's indeed impossible right now, I'll just give up on the `Promise` constructor for now.

PR #282, opened 2018-06-20
Non closure array bindings

wismer:
Here's a few of `Array` for today. I tried to include `toSource` but it's not implemented in any browser except Firefox and I assume there aren't any typescript declarations for it either.

I'll continue working on some more non-closure related items in the checklist - as much as I would love to figure the callback issue out, I'll wait until the rest is done first. Please let me know if there's anything I can do to enhance the tests or got the signatures wrong (or add a missing test - pretty sure I got them all). 

And please note that I decided to go with including all optional arguments for those that had them. Never knew that `indexOf` allows you to look in reverse.


fitzgen:
> I tried to include toSource but it's not implemented in any browser except Firefox and I assume there aren't any typescript declarations for it either.

Woops, I meant to remove these from the list. They're gone now.

PR #285, opened 2018-06-21
implements Object.prototype.propertyIsEnumerable() binding

belfz:


belfz:
I used `&JsValue` as an argument here, because at least both JavaScript's numbers and Strings can be used as parameter names in objects. There are tests for both cases.

Moreover, I think that existing binding for `has_own_property` should use the `&JsValue` as well, instead of the currently used `&str` (I can fix & PR that one, too).

fitzgen:
> Moreover, I think that existing binding for has_own_property should use the &JsValue as well, instead of the currently used &str (I can fix & PR that one, too).

Agreed, thanks for pointing this out and volunteering to fix it :)

belfz:
I forgot to save the changes in the file 🤦‍♂️ . Sorry for the confusion.
Updated the code just now (plus I added the `Symbol` case you suggested).

Thanks!

PR #288, opened 2018-06-21
allows using &JsValue as an arg to Object's has_own_property

belfz:
The binding for `Object`'s `has_own_property` should expect a property argument of type `&JsValue`. In JavaScript, a property can be of type `String`, `Number` and even `Symbol`. The previous implementation allowed `&str` only.

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -84,7 +84,7 @@ extern {
     ///
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty
     #[wasm_bindgen(method, js_name = hasOwnProperty)]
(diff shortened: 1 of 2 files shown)


jonathan-s:
`JsValue` encompasses every value in JavaScript. Which means that a JsValue can also be an array. Would it be a good idea to try and prevent this from happening making a new type that is a subset of the String, Number and Symbol?

belfz:
@jonathan-s the interesting fact is that in case of JS, it is perfectly valid for an array to be a property "name", too:
```js
const a = [1, 2, 3];
const o = { [a]: 42 };
o[a]; // 42
```

I mentioned only `String`, `Number` and `Symbol` originally because IMHO these are the most reasonable types to be used as property names. Still, other types are valid too.

PR #296, opened 2018-06-22
Attribute for static methods

fitzgen:
This hooks up the existing support for static methods that @ohanar added for WebIDL to a `#[wasm_bindgen(static_method_of = Class)]` attribute, and implements bindings to `Object.keys` to exercise the code path during tests.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -813,7 +813,7 @@ impl Struct {
                     ty: field.ty.clone(),
                     getter: Ident::new(&getter, Span::call_site()),
                     setter: Ident::new(&setter, Span::call_site()),
-                    comments
+                    comments,
                 });
             }
         }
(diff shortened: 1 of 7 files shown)


crates/backend/src/ast.rs
```diff
@@ -410,6 +411,24 @@ impl Program {
...
+                arguments: syn::PathArguments::None,
+            }));
+
+            let ty = syn::Type::Path(syn::TypePath {
+                qself: None,
+                path: syn::Path {
+                    leading_colon: None,
+                    segments,
+                },
+            });
```
ohanar: This logic largely redundant with `wasm_bindgen_webidl::util::ident_ty`, it might be good to move some of the utility functions out of that crate and into the backend.

ohanar:
Looks good to me. I'll let you decide if you want to address the small amount of refactoring here or leave it for the future.

alexcrichton:
FWIW this was sort of the intention of `js_namespace`, but having a few options isn't the end of the world! @fitzgen mind also updating the [design doc](https://github.com/rustwasm/wasm-bindgen/blob/master/guide/src/design.md) with the new attribute?

fitzgen:
Yes, I can do that when I get a chance. The reason `js_namespace` didn't work for this is because it creates a free function rather than a static method.

PR #297, opened 2018-06-23
use yarn for ci

ohanar:
Currently we use `npm install` for installing node packages, however, this makes no guarantee about the exact version of the package used. Instead, we should use `npm ci` to ensure that `package-lock.json` is respected. Unfortunately, this currently causes webpack to barf on one of our tests (`imports::versions`), so rather than try to debug this terrible mess, I elected to just use yarn instead (which was designed for reproducible builds in the first place).

This makes no changes to the documentation, and the project should still work perfectly fine without yarn installed, this simply is a change to the ci infrastructure.


.appveyor.yml
```diff
@@ -19,10 +19,13 @@ build: false
 
 test_script:
   - rustup target add wasm32-unknown-unknown
-  - npm install
+  - yarn install --freeze-lockfile
```
sendilkumarn: While I understand this change, I would rather stick with `npm`.

ohanar: I'm find with staying with `npm`, just not `npm install`. I couldn't be bothered with figuring out why webpack fails to function for one of our tests with `npm ci`, so I just elected to use yarn (since it just works).

If you would like to make a competing pull request that makes travis & appveyor use (and work) with `npm ci`, I would be happy to review it in favor of this.

fitzgen: > While I understand this change, I would rather stick with npm.

Can you explain why? @ohanar makes the case for why `yarn` is a good fit here, but I don't know the case for why we should stick with `npm`.

corbinu: @fitzgen Yarn has some pretty serious issues and its advantages were nullified by npm@6. Given that I am using this to build node modules rather than just front-end code which yarn is designed for I would really prefer we stay with NPM. 

A great article on the issues with yarn: https://mixmax.com/blog/to-yarn-and-back-again-npm

ohanar: @corbinu Yarn isn't the only one to have issues, it wouldn't have come into existence if npm didn't have it's own baggage. I'm not really interested in taking sides of yarn vs npm (the internet already has too much of that :smiley:), what I am interested is in reducing the surface area for unrelated build failures. I would be perfectly happy with a `npm ci` solution (as the article you linked suggested) or a yarn solution, but I do not want to stick with status quo.

I would like to note it is very easy for us to switch between the two, so I would be inclined to move ahead with this branch in the absence of a working npm alternative at the moment.

corbinu: I am sorry but I think the fact that yarn publish still doesn't work correctly means it can't be a full replacement in the CI and using two different things one to install and one to publish. Keep in mind that npm install also respects the package-lock file so I really don't see whats the difference with the status quo. npm ci just installs faster.

ohanar: We don't have any need to publish (and I don't see us having the need to publish anytime soon), we only need npm or yarn for dev dependencies for integration tests.

Running `npm install` doesn't fully respect `package-lock.json`, it still updates the file. Right now, if you freshly clone the repository and `npm install`, you get a different `package-lock.json` (npm inserts some extra fields). Run `npm install` a second time and you are back to the currently checked in `package-lock.json`. Fundamentally this means that the CI is not really running any pull requests or branches cleanly, but rather on one slightly modified by npm. `npm ci` doesn't have this issue.

corbinu: @ohanar Please see the npm docs on how to resolve conflicts: https://docs.npmjs.com/files/package-locks#resolving-lockfile-conflicts

sendilkumarn: @ohanar There will be a lot of discrepancy when you mix `yarn` and `npm`. They both run on different algorithms and get dependencies entirely differently and there will be a definite mismatch. 

Usage wise, yes there are a lot of discussions on this, which is the best one. IMO we should stick with `npm` because that is more consistent. 

> Right now, if you freshly clone the repository and npm install, you get a different package-lock.json (npm inserts some extra fields). 

We can remove the `^` in the package.json and that will solve most of the issues that we are talking here. 

But still if you prefer to use `yarn` I would suggest remove `package-lock.json` completely from the repo add it in `.gitignore` and use only `yarn.lock`.

Pauan: > There will be a lot of discrepancy when you mix yarn and npm. They both run on different algorithms and get dependencies entirely differently and there will be a definite mismatch.

Absolutely, so why aren't we testing it with *both* npm and yarn? Since downstream users might be using either one.

> We can remove the ^ in the package.json and that will solve most of the issues that we are talking here.

Unfortunately, that's not true because of transitive dependencies.

ohanar:
I'm closing this, and will implement @Pauan suggestion when I have a moment.

PR #299, opened 2018-06-23
Adds valueOf and toString to Number

jonathan-s:
While doing this I saw that `Result<String, JsValue>` can't cross the javascript boundary. As it isn't described in wasmdescribe. Would it be desirable to ever have that described in wasmdescribe, or should you explicitly handle any errors in Rust code.

### DIFF
--- a/tests/all/js_globals/Number.rs
+++ b/tests/all/js_globals/Number.rs
@@ -4,6 +4,40 @@ use super::project;
 
 
 #[test]
+fn to_string() {
+    project()
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn to_string(this: &js::Number, radix: u8) -> String {
+                let result = this.to_string(radix);
+                let result = match result {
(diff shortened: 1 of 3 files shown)


tests/all/js_globals/Number.rs
```diff
@@ -0,0 +1,127 @@
...
+        "#)
+        .file("test.ts", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                let number = 1234.45;
+                assert.equal(wasm.to_locale_string(number, "de-DE"), "1,234.45");
+                assert.equal(wasm.to_locale_string(number, "en-US"), "1,234.45");
+                assert.equal(wasm.to_locale_string(number, "zh-Hans-CN-u-nu-hanidec"), "1,234.45");
```
jonathan-s: If you use this locale string for javascript in the browser you'll get chinese characters. Any idea why that doesn't work here?

fitzgen: I assume it must be an issue with node.js then.

ohanar: This test fails for me, I get the chinese characters. Also the "de-DE" fails for me, I get "1.234,45".

fitzgen: Hm, I wonder if it depends on some other system-specific things then. Perhaps we should just disable the test.

PR #302, opened 2018-06-24
js globals: return JsString rather than String

ohanar:
This reduces the need to copy strings in and out of wasm memory. We should also potentially consider doing the same for arguments (rather than taking `&str`).

### DIFF
--- a/tests/all/js_globals/Number.rs
+++ b/tests/all/js_globals/Number.rs
@@ -14,7 +14,7 @@ fn to_locale_string() {
             use wasm_bindgen::js;
 
             #[wasm_bindgen]
-            pub fn to_locale_string(this: &js::Number, locale: String) -> String {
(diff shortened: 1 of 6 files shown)


tests/all/js_globals/Number.rs
```diff
@@ -3,7 +3,9 @@
 use super::project;
 
 
-#[test]
+// TODO: test seems to be system dependent, disable for now
+// #[test]
+#[allow(dead_code)]
```
jonathan-s: Delete these two lines and it'll be fine. 
https://github.com/rustwasm/wasm-bindgen/pull/302/files#diff-c29ca216e01d6480124fdf536ba30e97R29
https://github.com/rustwasm/wasm-bindgen/pull/302/files#diff-c29ca216e01d6480124fdf536ba30e97R31

ohanar: Good point.

Issue #244, opened 2018-06-08
Ignore anything that has `ChromeOnly` extended attribute


fitzgen:
Context: `ChromeOnly` is for things that are only exposed to priveleged code in Firefox (where we have been getting our WebIDL sources from). These items aren't exposed to the Web, so we don't want to generate bindings to them.

I'm imagining a `is_chrome_only` function that takes extended attributes and returns `bool`. Then we would check the extended attributes on interfaces, attributes, dictionaries, and operations with this function. If it returns true, and is ChromeOnly, then we would skip the item and not emit imports to it.

teovoinea:
I’d like to pick this issue up! 😊

fitzgen:
Awesome! Let me know if you have any questions. The WebIDL frontend lives in `crates/webidl`, and you can run its tests via `cargo test` within that directory, see `crates/webidl/tests/*`.

teovoinea:
Just to clarify, when you said "takes extended attributes", did you mean one function for each of [these](https://docs.rs/webidl/0.6.0/webidl/ast/index.html) `*ExtendedAttribute` structs?

fitzgen:
Yeah, for example: https://github.com/rustwasm/wasm-bindgen/blob/master/crates/webidl/tests/fixtures/Event.webidl#L45

ohanar:
Fixed by #303.

PR #303, opened 2018-06-24
Added is_chrome_only (#244)

teovoinea:
Implements #244

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -128,6 +128,10 @@ impl WebidlParse<()> for webidl::ast::Interface {
 
 impl WebidlParse<()> for webidl::ast::Typedef {
     fn webidl_parse(&self, program: &mut backend::ast::Program, _: ()) -> Result<()> {
+        if util::is_chrome_only(&self.extended_attributes) {
+            return Ok(())
+        }
+
         let dest = rust_ident(&self.name);
         let src = match webidl_ty_to_syn_ty(&self.type_, TypePosition::Return) {
             Some(src) => src,
@@ -154,6 +158,10 @@ impl WebidlParse<()> for webidl::ast::Typedef {
 
 impl WebidlParse<()> for webidl::ast::NonPartialInterface {
     fn webidl_parse(&self, program: &mut backend::ast::Program, _: ()) -> Result<()> {
+        if util::is_chrome_only(&self.extended_attributes) {
+            return Ok(())
+        }
+
         program.imports.push(backend::ast::Import {
             module: None,
             version: None,
@@ -293,6 +301,10 @@ impl<'a> WebidlParse<&'a str> for webidl::ast::Operation {
 
 impl<'a> WebidlParse<&'a str> for webidl::ast::RegularAttribute {
     fn webidl_parse(&self, program: &mut backend::ast::Program, self_name: &'a str) -> Result<()> {
+        if util::is_chrome_only(&self.extended_attributes) {
+            return Ok(())
+        }
+
         create_getter(
             &self.name,
             &self.type_,
@@ -317,6 +329,10 @@ impl<'a> WebidlParse<&'a str> for webidl::ast::RegularAttribute {
 
 impl<'a> WebidlParse<&'a str> for webidl::ast::StaticAttribute {
     fn webidl_parse(&self, program: &mut backend::ast::Program, self_name: &'a str) -> Result<()> {
+        if util::is_chrome_only(&self.extended_attributes) {
+            return Ok(())
+        }
+
         create_getter(
             &self.name,
             &self.type_,
@@ -341,6 +357,10 @@ impl<'a> WebidlParse<&'a str> for webidl::ast::StaticAttribute {
 
 impl<'a> WebidlParse<&'a str> for webidl::ast::RegularOperation {
     fn webidl_parse(&self, program: &mut backend::ast::Program, self_name: &'a str) -> Result<()> {
+        if util::is_chrome_only(&self.extended_attributes) {
+            return Ok(())
+        }
+
         create_basic_method(
             &self.arguments,
             self.name.as_ref(),
@@ -356,6 +376,10 @@ impl<'a> WebidlParse<&'a str> for webidl::ast::RegularOperation {
 
 impl<'a> WebidlParse<&'a str> for webidl::ast::StaticOperation {
     fn webidl_parse(&self, program: &mut backend::ast::Program, self_name: &'a str) -> Result<()> {
+        if util::is_chrome_only(&self.extended_attributes) {
+            return Ok(())
+        }
+        
         create_basic_method(
             &self.arguments,
             self.name.as_ref(),
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -6,6 +6,7 @@ use heck::SnakeCase;
 use proc_macro2::Ident;
 use syn;
 use webidl;
+use webidl::ast::ExtendedAttribute;
 
 fn shared_ref(ty: syn::Type) -> syn::Type {
     syn::TypeReference {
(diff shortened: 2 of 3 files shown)


crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
     Ok(compile_ast(&ast))
 }
 
+/// ChromeOnly is for things that are only exposed to priveleged code in Firefox.
```
jonathan-s: Minor nitpick, it's spelt privileged.

crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
     Ok(compile_ast(&ast))
 }
 
+/// ChromeOnly is for things that are only exposed to privileged code in Firefox.
+pub fn is_chrome_only(ext_attrs: &Vec<Box<ExtendedAttribute>>) -> bool {
```
ohanar: Please take a slice instead of a vec reference.

crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
...
+    ext_attrs.iter().map(|external_attribute| {
+        let mut ret = false;
+        match external_attribute {
```
ohanar: I believe if you explicitly deref (or double deref) here, we can avoid using box patterns. I'm not aware of the current stabilization status of box patterns, so I would rather avoid them if possible.

crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
     Ok(compile_ast(&ast))
 }
 
+/// ChromeOnly is for things that are only exposed to privileged code in Firefox.
+pub fn is_chrome_only(ext_attrs: &Vec<Box<ExtendedAttribute>>) -> bool {
+    ext_attrs.iter().map(|external_attribute| {
```
ohanar: Use the `any` method rather than a `map` and a `fold`.

crates/webidl/src/lib.rs
```diff
@@ -154,6 +195,10 @@ impl WebidlParse<()> for webidl::ast::Typedef {
 
 impl WebidlParse<()> for webidl::ast::NonPartialInterface {
     fn webidl_parse(&self, program: &mut backend::ast::Program, _: ()) -> Result<()> {
+        if is_chrome_only(&self.extended_attributes) {
```
ohanar: Would you add this to the other implemented `webidl::ast::*` that have extended attributes. Doing so will almost certainly require updating the `Event.rs` file under the test directory (which is good, since then your changes will actually be tested).

crates/webidl/src/lib.rs
ohanar: Also, please move this into the `util` module.

crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
     Ok(compile_ast(&ast))
 }
 
+/// ChromeOnly is for things that are only exposed to privileged code in Firefox.
+pub fn is_chrome_only(ext_attrs: &Vec<Box<ExtendedAttribute>>) -> bool {
+    ext_attrs.iter().map(|external_attribute| {
+        let mut ret = false;
```
ohanar: You don't need an extra mutable variable here, the `match` block can evaluate to your return value

crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
...
+        match external_attribute {
+            box ExtendedAttribute::ArgumentList(al) => {
+                if al.name == "ChromeOnly" {
+                    ret = true;
+                    println!("ArgumentList");
+                }
```
ohanar: You can replace this whole `if` block with `al.name == "ChromeOnly"` (no semicolon) and similarly for the later branches. Doing so will cause the `match` block to evaluate to a `bool`.

crates/webidl/src/lib.rs
```diff
@@ -70,6 +72,45 @@ pub fn compile(webidl_source: &str) -> Result<String> {
...
+                    println!("IdentifierList");
+                }
+            },
+            box ExtendedAttribute::NamedArgumentList(nal) => {
+                if nal.lhs_name == "ChromeOnly" {
+                    ret = true;
+                    println!("NamedArgumentList");
+                }
+            },
+            box ExtendedAttribute::NoArguments(na) => {
```
ohanar: Add a starting underscore to a variable if it is unused. E.g. `na` becomes `_na` (you could also just use `_`).

ohanar:
Thank you so much for doing this! Most of the changes that I requested are pretty minor things, but would just help the code flow a bit better.

crates/webidl/src/lib.rs
teovoinea: I had quite a few weird ownership issues with a double deref, not sure what I was doing wrong and ended up on the box syntax in the end. Maybe you could show me an example of how to get it working?

crates/webidl/src/lib.rs
teovoinea: So I added it to the other implemented `webidl::ast::*` and didn't have to make any changes to `Event.rs`, did I miss something?

teovoinea:
Thanks for the feedback! I made most of the changes, and left a couple of comments where I had issues! :smile:

crates/webidl/src/util.rs
```diff
@@ -285,3 +286,40 @@ pub fn create_setter(
...
 }
+
+pub fn wrap_import_function(function: backend::ast::ImportFunction) -> backend::ast::Import {
+    backend::ast::Import {
+        module: None,
+        version: None,
+        js_namespace: None,
+        kind: backend::ast::ImportKind::Function(function),
+    }
+}
```
ohanar: `wrap_import_function` got moved in an earlier commit, I think you accidentally added it back here again, would you remove it?

crates/webidl/src/util.rs
```diff
@@ -285,3 +286,40 @@ pub fn create_setter(
...
+
+/// ChromeOnly is for things that are only exposed to priveleged code in Firefox.
+pub fn is_chrome_only(ext_attrs: &[Box<ExtendedAttribute>]) -> bool {
+    ext_attrs.iter().any(|external_attribute| {
+        return match external_attribute {
```
ohanar: To remove the `box` pattern, replace `external_attribute` with `&**external_attribute`

crates/webidl/src/util.rs
```diff
@@ -285,3 +286,40 @@ pub fn create_setter(
...
+            },
+            box ExtendedAttribute::IdentifierList(il) => {
+                println!("IdentifierList");
+                il.lhs == "ChromeOnly"
+            },
+            box ExtendedAttribute::NamedArgumentList(nal) => {
+                println!("NamedArgumentList");
+                nal.lhs_name == "ChromeOnly"
+            },
+            box ExtendedAttribute::NoArguments(_na) => {
```
ohanar: We are missing a pattern here, which is why you didn't get any change needed for `Event.rs`, namely:
```rust
ExtendedAttribute::NoArguments(webidl::ast::Other::Identifier(name)) => name == "ChromeOnly",
```

ohanar:
@teovoinea I think I responded to your questions there, let me know if I missed any.

teovoinea:
@ohanar Thanks! Everything made sense :)

ohanar:
Thanks!

DJMcNab:
Does this fix #244/is it open for a reason?

ohanar:
@DJMcNab No, Github didn't autodetect it. I just closed it.

PR #305, opened 2018-06-24
Adds Number to_fixed and to_exponential function

sendilkumarn:


src/js.rs
```diff
@@ -246,6 +246,20 @@ extern {
     #[wasm_bindgen(catch, method, js_name = toPrecision)]
     pub fn to_precision(this: &Number, precision: u8) -> Result<String, JsValue>;
 
+    /// The toFixed() method returns a string representing the Number
+    /// object using fixed-point notation.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed
+    #[wasm_bindgen(catch, method, js_name = toFixed)]
+    pub fn to_fixed(this: &Number, digits: u8) -> Result<String, JsValue>;
```
ohanar: Please return `JsString`s rather than `String`s. Returning `String`s forces us to copy into wasm memory every time a method is called. I just merged a pull request that makes it easy to convert to and from `JsString` and `String`, you might want to merged that to make it easier to write tests.

sendilkumarn: Ah the new change. Done 👍

ohanar:
Thanks @sendilkumarn!

PR #307, opened 2018-06-24
`String.prototype.substr()` support

elpiel:
I wanted to open this PR as well as ask some questions. Not sure how the whole binding works, but I do know some stuff currently are not supported as the optional parameters( as well as `NaN` support). But I also wanted to raise the question: `What about all the special cases for specific handling?`, for example I can give `substr`:
```
If length is undefined, substr() extracts characters to the end of the string.
If length is a negative number, it is treated as 0.

For both start and length, NaN is treated as 0.
```

How can these cases be handled as a whole and what should we do about them until we can actually handle them?

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -359,7 +359,6 @@ extern {
     /// the start index and a number of characters after it.
     ///
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr
-    /// TODO: Add `NaN` and `undefined` support
     #[wasm_bindgen(method, js_class = "String")]
     pub fn substr(this: &JsString, start: i32, length: i32) -> JsString;
 }
--- a/src/js.rs
+++ b/src/js.rs
@@ -340,6 +340,14 @@ extern {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice
     #[wasm_bindgen(method, js_class = "String")]
     pub fn slice(this: &JsString, start: u32, end: u32) -> JsString;
+
+    /// The substr() method returns the part of a string between
+    /// the start index and a number of characters after it.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr
+    /// TODO: Add `NaN` and `undefined` support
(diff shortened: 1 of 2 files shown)


src/js.rs
```diff
@@ -340,6 +340,14 @@ extern {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice
     #[wasm_bindgen(method, js_class = "String")]
     pub fn slice(this: &JsString, start: u32, end: u32) -> JsString;
+
+    /// The substr() method returns the part of a string between
+    /// the start index and a number of characters after it.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr
+    /// TODO: Add `NaN` and `undefined` support
```
elpiel: Added todo as we don't yet have that support(or at least not that I know of).

tests/all/js_globals/JsString.rs
```diff
@@ -30,3 +30,39 @@ fn slice() {
...
+
+            export function test() {
+                let aString = "Mozilla";
+
+                assert.equal(wasm.create_substr(aString, 0, 1), "M");
+                assert.equal(wasm.create_substr(aString, 1, 0), "");
+                assert.equal(wasm.create_substr(aString, -1, 1), "a");
+                assert.equal(wasm.create_substr(aString, 1, -1), "");
+                // TODO: Uncomment and test these assertions, once we have support for optional parameters
+                // assert.equal(wasm.create_substr(aString, -3), "lla");
```
elpiel: I did add those tests though(for the optional parameter) and do want to know what should be added as a todo if needed at all.

src/js.rs
fitzgen: The `NaN` behavior is on the JS engine implementation side, not our side, so we don't need to worry about it.

elpiel:
Should I remove the todos then? And how should we handle the Optional parameters - again with todos or you want another way of doing it?

fitzgen:
For optional parameters, it is often best to just require their presence. It is a judgement call on a case-to-case basis though.

elpiel:
I removed the TODO, don't know if I should leave the `TODO`.

PR #311, opened 2018-06-25
ci: Cache mdbook installs to reduce CI build time

fitzgen:
We spend a lot of time installing and compiling `mdbook` from source. This caches `mdbook` across builds.


ohanar:
Looks good to me.

fitzgen:
Thanks!

PR #314, opened 2018-06-25
Array.prototype.filter

fitzgen:
The key to making the higher order functions like `Array.prototype.filter` work, is to *not* take values by reference in the closure. That is a little too tricky for the macros in `src/{convert,describe}.rs` and is also unnecessary since these things are all indices into a table under the hood anyways -- not something that is too big to efficiently pass directly.


ohanar:
Looks good to me.

PR #315, opened 2018-06-26
use both npm and yarn in ci

ohanar:


.travis.yml
```diff
@@ -46,7 +70,11 @@ matrix:
 
     # All examples work
     - rust: nightly
-      before_install: rustup target add wasm32-unknown-unknown
+      before_install:
+        - rustup target add wasm32-unknown-unknown
+        - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash
```
sendilkumarn: `0.33.11 ?`

fitzgen: Is there a way we can just always install the latest version?

Anyways, I want to land this ASAP and start getting the test coverage in place, so we can fix this in a follow up.

PR #317, opened 2018-06-26
fix little typo in backend crate

xeqlol:


xeqlol:
Ah, commited to wrong branch. Will create another pull request with clean changes.

PR #318, opened 2018-06-26
WeakMap bindings

xeqlol:
Added bindings for **WeakMap's** _new_, _get_, _set_, _delete_ and _has_ methods. :rocket:


src/js.rs
```diff
@@ -434,6 +434,45 @@ extern {
     pub fn value_of(this: &Object) -> Object;
 }
 
+// WeakMap
+#[wasm_bindgen]
+extern {
+    pub type WeakMap;
+
+    /// Returns the function that created an WeakMap wrapper.
```
fitzgen: Nitpick: this is currently documenting the `.constructor` property, but instead we should document the `WeakMap` constructor function itself:

> The WeakMap object is a collection of key/value pairs in which the keys are weakly referenced.  The keys must be objects and the values can be arbitrary values.

src/js.rs
```diff
@@ -434,6 +434,45 @@ extern {
...
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap/has
+    #[wasm_bindgen(method)]
+    pub fn has(this: &WeakMap, key: JsValue) -> bool;
+
+    /// The delete() method removes the specified element from a WeakMap object.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap/delete
+    #[wasm_bindgen(method)]
+    pub fn delete(this: &WeakMap, key: JsValue) -> bool;
```
fitzgen: Here and throughout, we should have `key` be `Object` rather than `JsValue`, since a `WeakMap`'s keys must be objects.

xeqlol: Got it 👍 Going to fix this ASAP!

PR #325, opened 2018-06-26
Support String.prototype.concat

rail44:


src/js.rs
```diff
@@ -447,6 +447,13 @@ extern {
     #[wasm_bindgen(method, js_class = "String", js_name = charAt)]
     pub fn char_at(this: &JsString, index: u32) -> JsString;
 
+
+    /// The concat() method concatenates the string arguments to the calling string and returns a new string.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat
+    #[wasm_bindgen(method, js_class = "String")]
+    pub fn concat(this: &JsString, string_2: &JsString) -> JsString;
```
sendilkumarn: How to pass on multiple strings here?

fitzgen: You mean via variadic arguments? It isn't supported yet, unfortunately.

rail44: https://github.com/rustwasm/wasm-bindgen/pull/325/files#diff-697eab89bc36e4c0667b1491284ca386R54 😭

Issue #233, opened 2018-06-06
Example of Default Output that is compatible for Node and Browsers


mbalex99:
In the main README.md, it states:

> By default the generated JS uses ES modules and is compatible with both Node and browsers (but will likely require a bundler for both use cases).

But it seems like the ES module can't be consumed by node directly. Like I can't do this:

```
const myLib = require('myLib')
const sum = myLib.add(1, 2)
console.log(sum)
```

However the `--nodejs` flag for the compiler allows this to work seamlessly. Do we need to use webpack for both browser setup and node setup with the default es6 modules?

Ideally, it would be fine if the default usage in the browser would need webpack, but what is the ideal build setup for using it in nodejs without webpack?

fitzgen:
Would love to receive a PR that clarified the README!

> Do we need to use webpack for both browser setup and node setup with the default es6 modules?

I don't know how node plans on supporting es6 modules, but maybe @ashleygwilliams knows more.

> Ideally, it would be fine if the default usage in the browser would need webpack, but what is the ideal build setup for using it in nodejs without webpack?

If you aren't supporting the Web, and are only supporing node, then the `--nodejs` flag should work fine for you. There is also the `--no-modules` flag to do a UMD-style thing.

mbalex99:
Well node supports es6 modules if you do this:
```
node --experimental-modules myFile.mjs 
```

However the problem is in the top of the generated bindgen file you have something like:

```
import * as wasm from './mylib_bg';
```

Here are the files that it generated (default settings):
* mylib.js
* mylib.d.ts
* mylib_bg.wasm

and running: 

```
$ mv mylib.js mylib.mjs 
$ node --experimental-modules myFile.mjs 
```

Results in:

```
(node:98839) ExperimentalWarning: The ESM module loader is experimental.
{ Error: Cannot find module ./mylib_bg
    at search (internal/modules/esm/default_resolve.js:28:12)
    at Loader.resolve [as _resolve] (internal/modules/esm/default_resolve.js:64:11)
    at Loader.resolve (internal/modules/esm/loader.js:56:18)
    at Loader.getModuleJob (internal/modules/esm/loader.js:88:40)
    at ModuleWrap.promises.module.link (internal/modules/esm/module_job.js:35:40)
    at link (internal/modules/esm/module_job.js:34:36) code: 'MODULE_NOT_FOUND' 
```

FreeMasen:
[...]

PR #326, opened 2018-06-26
Both

FreeMasen:
This is to address #233

This adds a new flag to the CLI `--both`, using this flag will allow users to generate libraries that will be usable from both Node.js and compiled with webpack.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ matrix:
         - |
           (cd examples/comments && sed -i 's/: "webpack-dev-server"/: "webpack"/' package.json && ./build.sh)
         - |
-          (cd examples/node-and-browser && sed -i 's/: "webpack-dev-server"/: "webpack"/' package.json && ./build.sh && npm start)
+          (cd examples/node_and_browser && sed -i 's/: "webpack-dev-server"/: "webpack"/' package.json && ./build.sh && npm start)
 
     # Build the guide.
     - rust: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ matrix:
           (cd examples/closures && sed -i 's/: "webpack-dev-server"/: "webpack"/' package.json && ./build.sh)
         - |
           (cd examples/comments && sed -i 's/: "webpack-dev-server"/: "webpack"/' package.json && ./build.sh)
+        - |
+          (cd examples/node-and-browser && sed -i 's/: "webpack-dev-server"/: "webpack"/' package.json && ./build.sh && npm start)
 
     # Build the guide.
     - rust: stable
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,8 @@ members = [
   "examples/asm.js",
   "examples/char",
   "examples/import_js",
-  "examples/comments"
+  "examples/comments",
+  "examples/node_and_browser"
 ]
 
 [profile.release]
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -59,7 +59,7 @@ impl<'a> Context<'a> {
         if let Some(ref c) = comments {
             self.globals.push_str(c);
         }
-        let global = if self.config.nodejs {
+        let global = if self.config.nodejs || self.config.both {
             if contents.starts_with("class") {
                 format!("{1}\nmodule.exports.{0} = {0};\n", name, contents)
             } else {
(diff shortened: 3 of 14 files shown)


FreeMasen:
@fitzgen The one thing that is missing currently from the PR is a new test in the tests module. I have added an example project and updated the `.travis.yml` to build that project for both browser and node.

I am at a loss as to how to actually test this using the current testing system. The system seems to want to bundle with webpack and then execute in the node environment. When I tried duplicating [node](../blob/master/tests/all/node.rs) with the new flag it was failing to resolve the `./out.js` functions. 

The goal of this PR is to actually remove the requirement of using webpack when working in a Node.js environment while allowing webpack to correctly bundle the same files for the browser. That being said I am not sure how I should be testing this?

FreeMasen:
One of the things that brought me to this issue was attempting to create [this](https://github.com/FreeMasen/wasm_cmark_parse) proof of concept when using `wasm-pack`.  

Initially I ended up having to create 2 different packages, one for node.js and one for the browser, otherwise the library wouldn't be usable as a dependency. By implementing these changes in that project I was able to get the same set of files to be usable from either environment.  The above link has an examples folder with a project using the parent project as an NPM installed dependency from both environments. 

Also, I'll try and come up with a better flag.

FreeMasen:
@xtuc As things stand today, to get a Web Assembly module working in Node (without Experimental ESM turned on) we need to execute the following.

```js
const imports = require('imports.js');
const fs = require('fs');
const bytes = fs.readFileSync('./thing.wasm');
const wasm = WebAssembly.Module(bytes);
const wasmInstance = WebAssembly.Instance(wasmModule, imports);
``` 
or something similar.  While Webpack does support CJS, it cannot resolve a `require('fs')` and will fail to compile. 

This causes a problem for someone who wants to publish a WebAssembly project to NPM and allow it to be used by other developers. So it is less a issue of ESM vs CJS but the current state of WASM in Node (supprted but not `require`able).

FreeMasen:
@fitzgen 
Some possible options for the flag could be
- `--dynamic`
- `--browser --nodejs`
  - both flags
- `--dual-env`
- `--hybrid`
- `--eval`
- `eval-fs`
I am also open to suggestions.

xtuc:
I think we should create a documentation about usage with CJS and ESM on different platforms.

## Node

Node itself **does not** support wasm, and it's not related to the module system being used (if CJS or ESM). 

With the experimental ESM support I've created a loader that supports wasm, see https://github.com/wasm-tool/node-loader. Otherwise for CJS, as you described, you'll need to load, provide the imports and instantiate the wasm module on your own.

## Webpack

Webpack 4 added the support for wasm module and has support for ESM since even longer out of the box. I would recommend to publish an ESM version because it's still compatible with CJS (apart from the default export I mentioned before).

FreeMasen:
> Node itself does not support wasm, 

I don't understand how this could be true, if I use the `WebAssembly` global to build my own wasm module, I can use it today. It is exactly what I am doing in the new example I have added in this PR as an example of this working today. Am I misunderstanding something?

xtuc:
Sorry, I was a bit unclear. Node has support for wasm, of course. What I meant is that there's not integration with the module system or basically "managed". As you demonstrated the loading, imports and instatiation is manual.

FreeMasen:
I understand the resistance for moving forward with this solution. I do think it would be good to provide some solution that works today. I wonder if this might be better applied as an optional post-processing step, similar to `wasm2es6js`?

xmclark:
So maybe I missed something, but I do not believe there is any issue with having an npm wasm package support both the browser and node.js and we do not need to inline the wasm like with `wasm2es6js`. package.json has a convention for supporting different targets e.g. browser and node.js. That convention is supported in both webpack and rollup.js.

In a package.json, there are two top level fields called 'main' and 'browser'. The main field is typically required, but the browser is optional. The bundlers will look for both, and fall back on the 'browser' field when targeting the browser.

A good example is superagent which is a request library for node.js (and the browser). [Check out their package.json](https://github.com/visionmedia/superagent/blob/master/package.json#L56). They have a different file loaded when webpack or rollup bundles superagent for the browser.

The wasm load code would be partially duplicated, but the wasm, the js shims and any additional would be common. You get support in both browsers.

Pauan:
@xmclark I agree with you, and I believe everything you said is correct.

However, I want to clarify something for the benefit of others:

> The wasm load code would be partially duplicated

This duplication only occurs when downloading the npm package: when running the code in Node.js or bundling for the browser there is no duplication.

So there is no downside to the duplication (other than an *extremely slightly* longer npm download).

This pattern (using the `main` and `browser` fields in `package.json`) is extremely common, it is the idiomatic way to have code which works in both Node.js and the browser.

alexcrichton:
I'm gonna close this as it's been quiet for quite some time now, but we can of course resubmit/reopen!

Issue #275, opened 2018-06-18
js-sys: Expose bindings to ALL the global JS things


fitzgen:
This is about exposing **ALL** of the globally available JS APIs through the `js-sys` crate. Things that are guaranteed by the ECMAScript standard, *not* Web/Node/etc APIs.

[A good overview/list/documentation of these APIs is available here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects) and I've also made a checklist below. As we implement bindings for these APIs, I will check them off.

## How to Implement New Bindings

* Comment here saying which thing you are going to make bindings for (so that we don't accidentally duplicate effort). I'll add your username next to the checkbox item.

* Open the [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects) for the relevant JS API.

* Open `crates/js-sys/src/lib.rs` in your editor; this is the file where we are implementing the bindings.

* Follow the instructions in `crates/js-sys/src/lib.rs` about how to add new bindings: https://github.com/rustwasm/wasm-bindgen/blob/aa348f963ff4ae7ddd769365ad1b197cd134e24e/crates/js-sys/src/lib.rs#L28-L44

* Add a test for the new binding to `crates/js-sys/tests/wasm/MyType.rs`

* Run the JS global API bindings tests with `cargo test -p js-sys --target wasm32-unknown-unknown`

* Send a pull request! :smile_cat: 

-------------------------------------------------------------------------------

Depends on this PR for the initial skeleton and infrastructure:

* [X] https://github.com/rustwasm/wasm-bindgen/pull/274

-------------------------------------------------------------------------------

All `String` bindings depend on:

* [x] https://github.com/rustwasm/wasm-bindgen/issues/287

-------------------------------------------------------------------------------

* [x] `Array`

    * [x] `Array.length` (@robertDurst)

    * [x] `Array.from()`

    * [x] `Array.isArray()`

    * [x] `Array.of()`

    * [x] `Array.prototype.concat()`

    * [x] `Array.prototype.copyWithin()`

    * [x] `Array.prototype.entries()`

    * [x] `Array.prototype.every()`

    * [x] `Array.prototype.fill()`

    * [x] `Array.prototype.filter()`

    * [x] `Array.prototype.find()`

    * [x] `Array.prototype.findIndex()`

    * [x] `Array.prototype.forEach()`

    * [x] `Array.prototype.includes()`

[...]

PR #334, opened 2018-06-27
feat(js) Implement `Number.new`, `Date.now` and `Date.UTC` bindings

Hywan:
* `Date.now` is a static method,
* `Date.UTC` is a static method,
* To test `Date.UTC`, I needed `Number.new` so I created it,
* `Date.UTC` has 5 optional parameters (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC) I didn't implement (cf. https://github.com/rustwasm/wasm-bindgen/issues/275#issuecomment-398549004). Is it OK?


src/js.rs
```diff
@@ -431,6 +431,13 @@ extern {
     #[wasm_bindgen(constructor)]
     pub fn new() -> Date;
 
+    /// The `Date.now()` method returns the number of milliseconds
+    /// elapsed since January 1, 1970 00:00:00 UTC.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now
+    #[wasm_bindgen(static_method_of = Date)]
+    pub fn now() -> Number;
```
fitzgen: I think this should return a `u64` -- it isn't returning a `Number` *object* (as opposed to a number) so this isn't correct as-written.

src/js.rs
```diff
@@ -516,6 +516,15 @@ extern {
...
     pub fn to_utc_string(this: &Date) -> JsString;
 
+    /// The `Date.UTC()` method accepts the same parameters as the
+    /// longest form of the constructor, and returns the number of
+    /// milliseconds in a `Date` object since January 1, 1970,
+    /// 00:00:00, universal time.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC
+    #[wasm_bindgen(static_method_of = Date, js_name = UTC)]
+    pub fn utc(year: Number, month: Number) -> Number;
```
fitzgen: Ditto to every `Number` here.

Hywan:
Anyway, the `Number::new` constructor is still in the patch even if it's not really related now ;-).

fitzgen:
And `u32` isn't good enough in this case because we really do need numbers larger than 2^32 - 1 for date things, right? Ah well, I guess it has to be `f64`.

PR #336, opened 2018-06-27
implements Array.prototype.every()

belfz:


src/js.rs
```diff
@@ -103,6 +103,13 @@ extern {
     #[wasm_bindgen(method)]
     pub fn concat(this: &Array, array: &Array) -> Array;
 
+    /// The every() method tests whether all elements in the array pass the test
+    /// implemented by the provided function.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every
+    #[wasm_bindgen(method)]
+    pub fn every(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool) -> bool;
```
jonathan-s: do we need the function to be mutable?

fitzgen: > do we need the function to be mutable?

Yes, because `FnMut::call_mut` requires an `&mut self` receiver: https://doc.rust-lang.org/nightly/std/ops/trait.FnMut.html#tymethod.call_mut

PR #339, opened 2018-06-28
feat(js): JSON bindings

xeqlol:
Static methods _parse_ and _stringify_ for **JSON** are implemented. (Issue #275)

Make sure these boxes are checked! 📦✅

- [ ] You have the latest version of `rustfmt` installed and have your 
      cloned directory set to nightly
```bash
$ rustup override set nightly
$ rustup component add rustfmt-preview --toolchain nightly
```
- [ ] You ran `rustfmt` on the code base before submitting
- [x] You reference which issue is being closed in the PR text

✨✨ 😄 Thanks so much for contributing to wasm-bindgen! 😄 ✨✨

P.s. Not sure that PR template is applied currectly. 😄
P.s.s. It seems `cargo fmt` formats a lot of code (not only code in this PR), so i guess that formatting should be done in separate massive PR.


src/js.rs
```diff
@@ -682,6 +682,15 @@ extern {
     pub fn delete(this: &WeakSet, value: Object) -> bool;
 }
 
+// JSON
+#[wasm_bindgen]
+extern {
+    pub type JSON;
```
fitzgen: Rust style for acronyms within identifiers would be to use `Json` here:

```rust
#[js_name = JSON]
pub type Json;
```

The other option, since `JSON` is sort of a module/namespace in JS, would be to add bindings like this:

```rust
mod json {
    #[wasm_bindgen]
    extern {
        #[wasm_bindgen(catch, js_namespace = JSON)]
        pub fn stringify(value: JsValue) -> Result<JsString, JsValue>;
    }
}
```

xeqlol:
Oops, did something wrong with merge. I'll close this and open new PR with clear changes.

PR #340, opened 2018-06-28
Use rustfmt for standarized formatting

ohanar:
We have lots of multi-line strings as part of function calls, these seem to be a bit funky under rustfmt. I fixed up a number of them, but there are likely still a good number of them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ matrix:
       install: true
       script: cargo build --manifest-path crates/cli/Cargo.toml
 
+    # check formatting
+    - rust: nightly
+      before_install: rustup component add rustfmt-preview --toolchain nightly
+      script: cargo fmt -- --check
+
     # Tests pass on nightly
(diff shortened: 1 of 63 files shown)


.travis.yml
```diff
@@ -8,6 +8,11 @@ matrix:
       install: true
       script: cargo build --manifest-path crates/cli/Cargo.toml
 
+    # check formatting
+    - rust: nightly
+      before_install: rustup component add rustfmt-preview --toolchain nightly
+      script: cargo fmt -- --check
```
alexcrichton: FWIW I've [disabled](https://github.com/rustwasm/wasm-bindgen/commit/942673e15f86dfd8971a35f990da4a543c49b0a3) doing this on Travis, in practice I've found this to just end up being a barrier to contribution unfortunately (also rustfmt's output is pretty unstable). We can periodically take PRs to update formatting though!

PR #342, opened 2018-06-28
feat(js): add bindings for Map

jannikkeye:
This branch adds bindings for the following methods and properties of Map:

- [x] Map.new
- [x] Map.clear
- [x] Map.set
- [x] Map.get
- [x] Map.has
- [x] Map.delete
- [x] Map.size
- [x] Map.entries
- [x] Map.values
- [x] Map.keys

I had trouble adding the missing ones for example `Map[@@iterator]` and `Map.forEach`. I wasn't able to find similar things implemented already.

Am I correct in guessing that to implement `Map[@@iterator]` we first have to support Symbol.iterator? I was also wondering how the function identifier should look like, as `@@iterator` does not work. Maybe just `pub fn iterator`?
I wasn't able to correctly add and test `Map.forEach`. Hopefully someone can point me in the right direction.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,8 +8,8 @@
       "integrity": "sha512-bPHp6Ji8b41szTOcaP63VlnbbO5Ny6dwAATtY6JTjh5N2OLrb5Qk/Th5cRkRQhkWCt+EJsYrNB0MiL+Gpn6e3g==",
       "dev": true,
       "requires": {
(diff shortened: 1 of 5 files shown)


jannikkeye:
Travis seems to fail but not on any tests. Is this caused by something I did? The logs on Travis don't make it any clearer to me.

PR #344, opened 2018-06-28
Binding for Math.cos,cosh,exp,expml,fround,imul,log,log10,log1p,log2

autodidaddict:
Bindings and tests for several math functions.

### DIFF
--- a/tests/all/js_globals/Math.rs
+++ b/tests/all/js_globals/Math.rs
@@ -78,7 +78,6 @@ fn acosh() {
             export function test() {
                 assert.equal(wasm.acosh(1), 0);
                 assert.equal(wasm.acosh(2), Math.acosh(2));
-                assert.equal(wasm.acosh(2), Math.acosh(2));
             }
         "#)
         .test()
@@ -315,7 +314,7 @@ fn cos() {
 
             export function test() {
                 assert.equal(wasm.cos(0), 1);
-                // other assertions failing due to rounding errors
+                assert.equal(wasm.cos(1.5), Math.cos(1.5));
             }
         "#)
         .test()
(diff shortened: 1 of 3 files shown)


autodidaddict:
I don't know why the build seems to have failed? All the new tests work locally.

jonathan-s:
`No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

So it's an issue with the Travis in this case.

autodidaddict:
Can we re-run that build?

tests/all/js_globals/Math.rs
```diff
@@ -293,6 +294,117 @@ fn clz32() {
...
+                js::Math::cos(x)
+            }
+        "#)
+        .file("test.ts", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                assert.equal(wasm.cos(0), 1);
+                // other assertions failing due to rounding errors
```
jonathan-s: If want to make more assertions what you could do is the following:
```
assert.equal(wasm.cos(1.34), Math.cos(1.34)))
```
The same goes for the other assertions where you have a very precise float. Doing the above could make it a bit more easy to read as well :)

tests/all/js_globals/Math.rs
```diff
@@ -78,6 +78,7 @@ fn acosh() {
             export function test() {
                 assert.equal(wasm.acosh(1), 0);
                 assert.equal(wasm.acosh(2), Math.acosh(2));
+                assert.equal(wasm.acosh(2), Math.acosh(2));
```
jonathan-s: This seems to be a line added by mistake?

autodidaddict: I didn't even work on `acosh` .. weird.

autodidaddict:
Updated in response to your comments.

autodidaddict:
Next time I will definitely split the commits per function. Not sure if I can do that retroactively

PR #347, opened 2018-06-28
Add bindings for Set.xx

jannikkeye:
Here is the new PR for the methods described in #343 but each method is implemented in a separate commit. I also adjusted the line lengths and removed the whitespace where mentioned.


alexcrichton:
All green!

PR #350, opened 2018-06-28
Support returning custom types in imports

### LINKED ISSUE #320: Support returning Rust struct from JS
Should it be possible to return (a pointer to) a Rust struct from Javascript? I noticed that trying to compile this, fails:

```rust
#![feature(proc_macro, wasm_custom_section, wasm_import_module)]

extern crate wasm_bindgen;

use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern {
    pub fn f(x: MyStruct) -> MyStruct;
}

#[wasm_bindgen]
pub struct MyStruct;
```

In general, I see three ways to cross the FFI boundary between Rust and JS:
- Serialise data structures when passing the boundary
- Construct, manipulate, and query all data structures in JS by calling JS functions from Rust/wasm
- Construct, manipulate, and query all data structures in Rust/wasm by calling Rust/wasm functions from JS

It looks like there is support for the first option (with `from_serde` and `into_serde`) and there is a lot of stuff happing for the second option (like #275). Seeing that the example above isn't possible (yet?), makes me think that the third approach isn't a good idea perhaps? Or at least suboptimal to the other approaches?

Thank you for your patience if this is a trivial question or something, and thank you for the awesome work you have put into this project!

alexcrichton:
Closes #320

### DIFF
--- a/crates/cli-support/src/js/rust2js.rs
+++ b/crates/cli-support/src/js/rust2js.rs
@@ -298,6 +298,26 @@ impl<'a, 'b> Rust2Js<'a, 'b> {
             );
             return Ok(());
         }
+
+        if let Some(class) = ty.rust_struct() {
+            if ty.is_by_ref() {
+                bail!("cannot invoke JS functions returning custom ref types yet")
+            }
+            // Insert an assertion to the type of the returned value as
+            // otherwise this will cause memory unsafety on the Rust side of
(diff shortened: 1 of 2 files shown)


Procrat:
Awesome, Alex! Thanks a lot for fixing that so quickly! :heart:

PR #351, opened 2018-06-28
Support wildcard arguments in foreign functions

### LINKED ISSUE #346: Extern function with an underscore as argument name causes panic saying "unsupported pattern in foreign function"
Minimal example:

```rust
#[wasm_bindgen]
pub extern {
    pub fn f(_: u8);
}
```

This is probably not the intended behaviour as the extern block compiles fine without the bindgen attribute.

alexcrichton:
No real reason to not support them!

Closes #346

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -573,6 +573,7 @@ impl ToTokens for ast::ImportFunction {
         let mut abi_argument_names = Vec::new();
         let mut abi_arguments = Vec::new();
         let mut arg_conversions = Vec::new();
+        let mut arguments = Vec::new();
         let ret_ident = Ident::new("_ret", Span::call_site());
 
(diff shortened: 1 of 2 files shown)


ohanar:
If we are going to add support for wildcard patterns, couldn't we just add support for all patterns? It just makes sense to me to take an all or nothing approach to this.

alexcrichton:
True! A lot of AST constructs are gonna be on a sort of case-by-case basis, as other sorts of patterns are probably broken for a few other reasons as well. In general though it's pretty rare to see something like `fn foo(ref x: u32)` in an `extern` block though

crates/backend/src/codegen.rs
```diff
@@ -696,6 +696,8 @@ impl ToTokens for ast::ImportFunction {
 
         };
 
+        // println!("{}", invocation);
+
```
ohanar: Other than these leftovers, everything looks good to me.

PR #355, opened 2018-06-29
Don't generate JS bindings for unused imports

alexcrichton:
If a JS import's shim isn't actually imported that means that somewhere along
the way it was optimized out or it was never used in the first place! In that
case we can skip generation of the JS bindings for it as it's not needed.

### DIFF
--- a/tests/all/imports.rs
+++ b/tests/all/imports.rs
@@ -1,3 +1,6 @@
+use std::fs::File;
+use std::io::Read;
+
 use super::project;
 
 #[test]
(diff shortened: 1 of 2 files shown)


tests/all/imports.rs
```diff
@@ -668,3 +671,40 @@ fn custom_type() {
...
+            export function test() {
+                run();
+            }
+        "#)
+        .test();
+
+    let out = ::root().join("out.js");
+    let mut contents = String::new();
+    File::open(&out).unwrap()
+        .read_to_string(&mut contents).unwrap();
```
fitzgen: We have a helper function for reading `out.js` now :)

```rust
let project = project()
    ...

let contents = project.read_js();
```

alexcrichton: Oh nice!

alexcrichton:
Er I botched the branches but this is now on master

PR #356, opened 2018-06-30
add more detailed error messages for invalid #[wasm_bindgen] invocations

kzvi:
Allows this code:

```rust
#[wasm_bindgen(readonly)]
pub fn a() {}

#[wasm_bindgen(js_name)]
pub fn b() {}

#[wasm_bindgen(a = b)]
pub fn c() {}

#[wasm_bindgen(js_name = 1)]
pub fn d() {}

#[wasm_bindgen(version = "1", version = "2")]
extern {}
```

To produce this:

```
error: attribute is not available in this context
  --> src/lib.rs:15:16
   |
15 | #[wasm_bindgen(readonly)]
   |                ^^^^^^^^
   |
   = help: attribute "readonly" is not allowed before exported function

error: attribute "js_name" requires a value
  --> src/lib.rs:18:16
   |
18 | #[wasm_bindgen(js_name)]
   |                ^^^^^^^

error: unrecognized attribute "a"
  --> src/lib.rs:21:16
   |
21 | #[wasm_bindgen(a = b)]
   |                ^

error: failed to parse value for attribute "js_name"
  --> src/lib.rs:24:26
   |
24 | #[wasm_bindgen(js_name = 1)]
   |                          ^
   |
   = help: expected ident

error: duplicate attribute "version"
  --> src/lib.rs:27:31
   |
27 | #[wasm_bindgen(version = "1", version = "2")]
   |                               ^^^^^^^^^^^^^

error: aborting due to 5 previous errors
```

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -2,6 +2,8 @@ use proc_macro2::{Ident, Span, TokenStream, TokenTree};
 use quote::ToTokens;
 use shared;
 use syn;
+use syn::spanned::Spanned;
+use syn::synom::Parser;
 use util;
 
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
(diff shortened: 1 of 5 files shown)


kzvi:
I added `#[cfg(feature = "spans")]` before each error message, but I'm not sure if that is enough to make it work with the stable compiler. Would it be better to create a type that imitates the unstable `Diagnostic` type but that can be used without the span API?

alexcrichton:
True! Ideally though I think we'd have a method that doesn't involve adding `#[cfg]` on all APIs but instead is what you're thinking, imitating the diagnostic apis but can be used without spans.

That also allows us to decouple the `spans` feature from this feature, which I think are pretty separable.

kzvi:
I'm going to close this because I don't want to work on it and it might be a better idea to wait until this API is finalized anyway.

PR #363, opened 2018-07-02
Expose bindings/object is* methods

belfz:
Implements bindings for `Object.isExtensible()`, `Object.isFrozen()` and `Object.isSealed()` static methods.


alexcrichton:
Awesome, thanks!

PR #365, opened 2018-07-02
binding for string.endsWith

gaurikholkar:
Part of #275

### DIFF
--- a/tests/all/js_globals/JsString.rs
+++ b/tests/all/js_globals/JsString.rs
@@ -627,3 +627,33 @@ fn value_of() {
         )
         .test()
 }
+
+#[test]
+fn ends_with() {
+    project()
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn string_ends_with(this: &js::JsString, search_string: &js::JsString, length: u32) -> bool {
+                this.ends_with(search_string, length)
+            }
+        "#)
+        .file("test.ts", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                let str = "To be, or not to be, that is the question.";
+
+                assert.ok(wasm.string_ends_with(str, 'question', 0));
+                assert.ok(!wasm.string_ends_with(str, 'to be', 0));
+                assert.ok(wasm.string_ends_with(str, 'to be', 19));
+            }
+        "#)
+        .test()
+}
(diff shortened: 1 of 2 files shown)


tests/all/js_globals/JsString.rs
```diff
@@ -627,3 +627,33 @@ fn value_of() {
...
+            }
+        "#)
+        .file("test.ts", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                let str = "To be, or not to be, that is the question.";
+
+                assert.ok(wasm.string_ends_with(str, 'question', 0));
```
belfz: You have a bug here. This should rather read `assert.ok(wasm.string_ends_with(str, 'question.', str.length));` in order to pass (notice the dot at the end of search string).

tests/all/js_globals/JsString.rs
```diff
@@ -627,3 +627,33 @@ fn value_of() {
...
+            pub fn string_ends_with(this: &js::JsString, search_string: &js::JsString, length: u32) -> bool {
+                this.ends_with(search_string, length)
+            }
+        "#)
+        .file("test.ts", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                let str = "To be, or not to be, that is the question.";
```
belfz: This is minor, but please favor use of `const`, unless `let` is really needed.

tests/all/js_globals/JsString.rs
belfz: After you correct that, please make sure to run the tests with `cargo test js_globals::JsString` :)

alexcrichton:
A good question! I'm not sure an `Option` will work in the sense that it'd have to pass `undefined` as the argument value, but another option here is something where we expose both versions of the function under different names. For example we could have:

```rust
#[wasm_bindgen(method, js_name = endsWith)]
fn ends_with(this: &JsString, tail: &JsString) -> bool;
#[wasm_bindgen(method, js_name = endsWith)]
fn ends_with_len(this: &JsString, tail: &JsString, length: u32) -> bool;
```

belfz:
@alexcrichton I was thinking that passing `None` could be mapped to `undefined` when crossing the wasm/js boundaries. Although now that I am thinking about it again, maybe we need a `js::Undefined` type, which would act as a "subtype" of all other types?

 While what you proposed could indeed work, I'm afraid it would become too verbose, especially in case of functions/methods that accept more than one optional parameter. Plus, we have to remember, that in such cases we could still want such functions to be called with `undefined` value for arguments in the middle.

alexcrichton:
Indeed it may not be generally applicable! I think for something like `ends_with` though it may be the most ergonomic though

sendilkumarn:
Oops sorry @gaurikholkar Closing this PR as I think it is already merged into master.

PR #370, opened 2018-07-02
Add `design.md` to the online mdbook

alexcrichton:
At the same time split it up into some more digestable sections


fitzgen:
Is this the same content as what is in "Contributing > Design" ?

alexcrichton:
Aha yes indeed! I dunno why I kept overlooking that... I've moved it around to expand where it is today (broken up into multiple pages now)

jonathan-s:
I've had a read through and it looks good to me. Just a thought when it comes to the term `ABI`, I assume that it stands for Application Binary Interface? Perhaps that could be clarified somewhere?

I also had a thought that there is quite a few code snippets in that documentation. These code snippets could possibly break at some point. Perhaps this crate could help out with making sure that the rust code in the documentation is still viable -> https://github.com/budziq/rust-skeptic

fitzgen:
> Just a thought when it comes to the term ABI, I assume that it stands for Application Binary Interface? Perhaps that could be clarified somewhere?

Yep and good idea :)

> I also had a thought that there is quite a few code snippets in that documentation. These code snippets could possibly break at some point. Perhaps this crate could help out with making sure that the rust code in the documentation is still viable -> https://github.com/budziq/rust-skeptic

Another good idea :) We should at least check that they compile, actually testing some them will require even more headless browser integration, I think.

jonathan-s:
I'll create an issue that makes sure that the code in the markdown compiles. :)

PR #372, opened 2018-07-03
Fix the constructor explanation in the guide

konstin:
This fixes the explanation of constructors in the guide as discussed in https://github.com/rustwasm/wasm-bindgen/pull/276#discussion_r199004235.


belfz:
Looks good! Similar could be added to another example, that currently makes use of `.new` - it's in `examples/char` :)

alexcrichton:
Thanks!

PR #373, opened 2018-07-03
Add support getter and setter for static props 

rail44:


tests/all/js_globals/Symbol.rs
```diff
@@ -0,0 +1,37 @@
...
+        )
+        .file(
+            "test.ts",
+            r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                assert.ok(wasm.symbol_has_instance());
+                assert.ok([] instanceof Array);
```
belfz: Correct me if I'm wrong, but I think that line alone doesn't bring much value in this test. Could you instead try to include a code snippet from [Symbol.hasInstance page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance), asserting that `[] instanceof Array1` (notice that `Array1` is a new class here, which implements `static [Symbol.hasInstance](instance)` method).

tests/all/js_globals/Symbol.rs
```diff
@@ -0,0 +1,37 @@
...
+        "#,
+        )
+        .file(
+            "test.ts",
+            r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                assert.ok(wasm.symbol_has_instance());
```
belfz: I think it could be much more meaningful to assert that a return value of that method is of `symbol` type - for example using `assert.equal(typeof wasm.symbol_has_instance(), "symbol")` 🙂

rail44: I see!

tests/all/js_globals/Symbol.rs
rail44: Oh, I missed to revert chaging to create test that absolutely passing :bowing_man:

belfz: 👍 although now that I'm thinking about it again, it might be even pointless to include that code snippet - that would be testing the JS itself (not the WASM part).

rail44:
Although fixing up, I'm getting following error by `cargo test has_instance` :disappointed: 

```
running "node" "/home/satoshi/src/github.com/rail44/wasm-bindgen/target/generated-tests/test0/bundle.js"
exit: exit code: 1
dur: 0.205ms
stderr ---
TypeError: Cannot assign to read only property 'Symbol(Symbol.hasInstance)' of function 'function Array1() {
    }'
    at /home/satoshi/src/github.com/rail44/wasm-bindgen/target/generated-tests/test0/1.bundle.js:19:40
    at Object../test.ts (/home/satoshi/src/github.com/rail44/wasm-bindgen/target/generated-tests/test0/1.bundle.js:23:2)
    at __webpack_require__ (/home/satoshi/src/github.com/rail44/wasm-bindgen/target/generated-tests/test0/bundle.js:46:30)
    at __webpack_require__.t (/home/satoshi/src/github.com/rail44/wasm-bindgen/target/generated-tests/test0/bundle.js:151:
3)
    at <anonymous>
```

And there is `1.bundle.js`

```javascript
exports.ids = [1];
exports.modules = {

/***/ "./test.ts":
/*!*****************!*\
  !*** ./test.ts ***!
  \*****************/
/*! no static exports found */
/***/ (function(module, exports, __webpack_require__) {

"use strict";

Object.defineProperty(exports, "__esModule", { value: true });
var assert = __webpack_require__(/*! assert */ "assert");
var wasm = __webpack_require__(/*! ./out */ "./out.js");
var Array1 = /** @class */ (function () {
    function Array1() {
    }
    Array1[wasm.symbol_has_instance()] = function (instance) {
        return Array.isArray(instance);
    };
    return Array1;
}());
function test() {
    assert.ok(wasm.symbol_has_instance());
    assert.ok([] instanceof Array1);
}
exports.test = test;


/***/ })

};;
//# sourceMappingURL=1.bundle.js.map
```

It seems to be checked by Typescript in spite of its already compiled js from ts.

rail44:
I'm going to get this again tommorow...

belfz:
@rail44 I think I found a solution - that should work:
```typescript
import * as assert from "assert";
import * as wasm from "./out";

class Array1 {}
Object.defineProperty(Array1, wasm.symbol_has_instance(), {
    value: (instance: any) => Array.isArray(instance)
});

export function test() {
    assert.equal(typeof wasm.symbol_has_instance(), "symbol");
    assert.ok([] instanceof Array1);
}
```

It looks like a TypeScript's bug to me. For some reason, it transpiled yours to a code that attempts to define that property in incorrect way, thus raising a TypeError. [Babel had a similar issue before](https://github.com/babel/babel/issues/2286).

rail44:
@belfz thanks for your workaround!
I've just push a commit 😃

tests/all/js_globals/Symbol.rs
```diff
@@ -0,0 +1,41 @@
...
+
+            #[wasm_bindgen]
+            pub fn symbol_has_instance() -> js::Symbol {
+                js::Symbol::has_instance()
+            }
+
+        "#,
+        )
+        .file(
+            "test.ts",
```
belfz: TypeScript has been temporarily disabled, so this test fails with `thread 'js_globals::Symbol::has_instance' panicked at 'webpack needs to be enabled to use typescript'`. Could you please change that line to `test.js`?

tests/all/js_globals/Symbol.rs
```diff
@@ -0,0 +1,41 @@
...
+            r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+            class Array1 {}
+            Object.defineProperty(Array1, wasm.symbol_has_instance(), {
+                value: (instance: any) => Array.isArray(instance)
```
belfz: Because we can't use TypeScript at the moment, the `any` type annotation here has to be removed as well.

tests/all/js_globals/Symbol.rs
rail44: Sorry,  I didn't realize it 😅

Issue #368, opened 2018-07-02
js_globals::Date::to_locale_date_string test failing locally


fitzgen:
Even with the latest node 10

```
$ nvm use 10
Now using node v10.5.0 (npm v6.1.0)
```

it fails locally for me:

```
failures:

---- js_globals::Date::to_locale_date_string stdout ----
···················································
running "cargo" "build" "--target" "wasm32-unknown-unknown"
exit: exit code: 0
dur: 0.601ms
stderr ---
   Compiling test28 v0.0.1 (file:///Users/fitzgen/src/wasm-bindgen/target/generated-tests/test28)
    Finished dev [unoptimized + debuginfo] target(s) in 0.55s

···················································
running "npm" "run" "run-webpack"
exit: exit code: 0
dur: 5.041ms
stdout ---

> @ run-webpack /Users/fitzgen/src/wasm-bindgen/target/generated-tests/test28
> webpack

Hash: 0025335020a4964bfc1e
Version: webpack 4.11.1
Time: 3668ms
Built at: 07/02/2018 1:11:20 PM
                           Asset          Size  Chunks             Chunk Names
                       bundle.js      9.64 KiB    main  [emitted]  main
                     0.bundle.js        68 KiB       0  [emitted]  
a7af4adcd48892d78296.module.wasm  unknown size       0  [emitted]  
                     1.bundle.js     747 bytes       1  [emitted]  
                   bundle.js.map      8.98 KiB    main  [emitted]  main
                 0.bundle.js.map      43.8 KiB       0  [emitted]  
                 1.bundle.js.map     656 bytes       1  [emitted]  
[./out.js] 37.7 KiB {0} [built]
[./out_bg.wasm] 235 KiB {0} [built]
[./run.js] 610 bytes {main} [built]
[./test.ts] 425 bytes {1} [built]
[assert] external "assert" 42 bytes {main} [built]
[process] external "process" 42 bytes {main} [built]
[util] external "util" 42 bytes {main} [built]

···················································
running "node" "/Users/fitzgen/src/wasm-bindgen/target/generated-tests/test28/bundle.js"
exit: exit code: 1
dur: 0.296ms
stderr ---
{ AssertionError [ERR_ASSERTION]: 'Wednesday, December 19, 2012' == 'Thursday, December 20, 2012'
    at Object.test (/Users/fitzgen/src/wasm-bindgen/target/generated-tests/test28/1.bundle.js:19:12)
[...]

PR #374, opened 2018-07-03
test(js) Fix expectations regarding given locales

Hywan:
I've no idea why these tests can be green on Travis.

Hope my patch will not make them red… while they _should_ be green :-).


alexcrichton:
Oh dear, looks like it failed on both Travis and AppVeyor :(

sendilkumarn:
I think @fitzgen idea to check only whether it is String is better for now. And we have to admit, It is a pain to handle the locales

jonathan-s:
Related to #368

Hywan:
So if I follow the recent discussions, we should update the patch to only check that it returns a non-empty string, without caring about the exact result?

I'm good with this approach.

alexcrichton:
Sounds good to me!

PR #378, opened 2018-07-03
Bindings for Math.pow, round, sign, sin, sinh, sqrt, tan, tanh and trunc

jhenninger:


jonathan-s:
Only comment I have is that it would be good if all test cases included taking decimal numbers as well so that it's covered in the tests. Otherwise 👍

jhenninger:
Done, at least for the tests added in this PR.

PR #380, opened 2018-07-03
bindings for Generator.next(), Generator.return() and Generator.throw()

akryvomaz:


src/js.rs
```diff
@@ -369,6 +369,19 @@ extern "C" {
...
+#[wasm_bindgen]
+extern {
+    pub type Generator;
+
+    /// The next() method returns an object with two properties done and value.
+    /// You can also provide a parameter to the next method to send a value to the generator.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/next
+    #[wasm_bindgen(method, structural)]
+    pub fn next(this: &Generator, value: &JsValue) -> JsValue;
```
fitzgen: This could totally throw an exception, so we need to add teh `wasm_bindgen(catch)` attribute and have the return type be `Result<JsValue, JsValue>`.

tests/all/js_globals/Generator.rs
```diff
@@ -0,0 +1,51 @@
...
+            import * as wasm from "./out";
+
+            export function test() {
+                function* generator() {
+                    const reply = yield '2 * 2';
+
+                    return reply === 4;
+                }
+
+                const gen = generator();
```
fitzgen: And then we should also have a test with:

```js
function* G() {
    throw new Error("uh oh");
}

assert(wasm.next_throws_error(G()));
```

where we have this rust:

```rust
#[wasm_bindgen]
pub fn next_throws_error(g: &Generator) -> bool {
    g.next().is_err()
}
```

src/js.rs
```diff
@@ -374,6 +374,12 @@ extern "C" {
 extern {
     pub type Generator;
 
+    /// The return() method returns the given value and finishes the generator.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/return
+    #[wasm_bindgen(method, structural, js_name = return)]
+    pub fn gen_return(this: &Generator, value: &JsValue) -> JsValue;
```
fitzgen: In the WebIDL frontend, we mangle Rust keywords into identifiers by appending an underscore, so I think we should be consistent and do the same here:

```rust
#[wasm_bindgen(method, structural, js_name = return)] 
pub fn return_(this: &Generator, value: &JsValue) -> JsValue;
```

src/js.rs
```diff
@@ -386,6 +386,13 @@ extern {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/next
     #[wasm_bindgen(method, structural)]
     pub fn next(this: &Generator, value: &JsValue) -> JsValue;
+
+    /// The throw() method resumes the execution of a generator by throwing an error into it
+    /// and returns an object with two properties done and value.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/throw
+    #[wasm_bindgen(method, structural)]
+    pub fn throw(this: &Generator, error: &Error) -> JsValue;
```
fitzgen: This should also have `catch`, since the generator can (and likely will) throw an exception when one calls this method.

tests/all/js_globals/Generator.rs
```diff
@@ -89,3 +89,49 @@ fn next() {
...
+                    yield 2;
+                }
+
+                const gen = generator();
+                gen.next();
+
+                try {
+                    wasm.gen_throw(gen, new Error('Something went wrong'));
+                } catch(err) {
+                    assert.equal(err.message, 'Something went wrong');
```
fitzgen: Rather than asserting that we unwound over the wasm, we should have the wasm verify that it can catch the exception as an `Err` result.

tests/all/js_globals/Generator.rs
akryvomaz: Rare use case but its real, you are right. Well thought!

PR #381, opened 2018-07-03
Speed up Travis by running Webpack in fewer tests

alexcrichton:
This commit is a large refactoring of the test suite which primarily restructures it to avoid needing to run webpack most of the time. Running webpack can take quite some time (2s per test locally) so this is hopefully going to be a good speed boost for tests on Travis. 

The test harness still retains the ability to support Webpack to test things like Typescript and upcoming headless testing, but most tests don't need this.


.travis.yml
```diff
@@ -1,20 +1,38 @@
...
-
     # Tests pass on nightly
     - rust: nightly
-      before_install:
-        - rustup target add wasm32-unknown-unknown
-        - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
-        - source ~/.nvm/nvm.sh
-        - nvm install v10.5
+      env: JOB=test-bindgen
+      before_install: *INSTALL_NODE_VIA_NVM
```
fitzgen: wat. I had no idea that this was possible!

xtuc:
well we should make it faster in Webpack too

alexcrichton:
@xtuc certainly! This is perhaps an egregious case though which webpack may or may not want to optimize for. We're basically running webpack hundreds of times over little projects all over the place, and each webpack invocation takes about 2s locally. Much of that 2s is probably related to loading dependencies or processing things like TypeScript. We definitely want to keep Webpack for TypeScript and headless tests! I was just thinking for Travis's sake it might be best to avoid running it for all the tests when it may not be necessary!

PR #382, opened 2018-07-04
Add bindings for Proxy

srenatus:
This is for #275.

I'm not entire sure about the bindings added for `Proxy.revocable` -- I would want to be more explicit about the return type [here](https://github.com/rustwasm/wasm-bindgen/compare/master...srenatus:sr/js_globals/Proxy#diff-2e17c66ff42936abe2d7eae32713e8caR998), but I haven't found a way to express _"it's an `Object` with `proxy` of type `Proxy` and `revoke` a function"_.

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -1030,6 +1030,20 @@ extern "C" {
     pub fn values(object: &Object) -> Array;
 }
 
+// Proxy
+#[wasm_bindgen]
+extern {
+    pub type Proxy;
+
+    /// The Proxy object is used to define custom behavior for fundamental
+    /// operations (e.g. property lookup, assignment, enumeration, function
+    /// invocation, etc).
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
+    #[wasm_bindgen(constructor)]
+    pub fn new(target: &JsValue, handler: &Object) -> Proxy;
+}
+
 // Set
 #[wasm_bindgen]
 extern {
--- a/src/js.rs
+++ b/src/js.rs
@@ -1042,6 +1042,12 @@ extern {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
     #[wasm_bindgen(constructor)]
     pub fn new(target: &JsValue, handler: &Object) -> Proxy;
+
+    /// The Proxy.revocable() method is used to create a revocable Proxy object.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/revocable
+    #[wasm_bindgen(static_method_of = Proxy)]
+    pub fn revocable(target: &JsValue, handler: &Object) -> Object;
 }
 
 // Set
--- a/tests/all/js_globals/Proxy.rs
+++ b/tests/all/js_globals/Proxy.rs
@@ -1,0 +1,37 @@
+#![allow(non_snake_case)]
+
+use project;
+
+#[test]
+fn new() {
+    project()
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn new_proxy(target: JsValue, handler: js::Object) -> js::Proxy {
+                js::Proxy::new(&target, &handler)
+            }
+        "#)
+        .file("test.ts", r#"
+            import * as assert from "assert";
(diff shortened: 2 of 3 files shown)


belfz:
I think that maybe it would be a good idea to create a new type for that special case (eg `RevocableProxy`) and add the `revoke` method signature there :)

srenatus:
@belfz I'm not sure I can figure this out... I've tried adding
```rust
#[wasm_bindgen]
pub struct RevocableProxy {
    proxy: Proxy,
    revoke: FnMut(),
}
```
outside of `extern {}`, [here](https://github.com/rustwasm/wasm-bindgen/pull/382/files#diff-2e17c66ff42936abe2d7eae32713e8caR980), but I keep getting complaints about `__wbindgen_if_not_std` not being found:
```
   Compiling wasm-bindgen v0.2.11 (file:///Users/stephan/Misc/Rust/_wasm/wasm-bindgen)
error[E0432]: unresolved import `wasm_bindgen`
   --> src/js.rs:988:1
    |
988 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^ no `__wbindgen_if_not_std` in `wasm_bindgen`. Did you mean to use `__wbindgen_cb_forget`?

error: cannot determine resolution for the macro `__wbindgen_if_not_std`
   --> src/js.rs:988:1
    |
988 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^
    |
    = note: import resolution is stuck, try simplifying macro imports

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `wasm-bindgen`.

To learn more, run the command again with --verbose.
```

And indeed, the macro doesn't seem to be defined? (The struct mentioned above might not be the right one, but I'm stuck early here... 😕)

srenatus:
Trial-end-erroring around, I seem to see a different situation when omitting the `#[wasm_bindgen]` from the struct:
```
error[E0277]: the trait bound `js::RevocableProxy: describe::WasmDescribe` is not satisfied
   --> src/js.rs:994:1
    |
994 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^ the trait `describe::WasmDescribe` is not implemented for `js::RevocableProxy`
    |
note: required by `describe::WasmDescribe::describe`
   --> src/describe.rs:48:5
    |
48  |     fn describe();
    |     ^^^^^^^^^^^^^^

error[E0277]: the size for value values of type `(dyn core::ops::FnMut() + 'static)` cannot be known at compilation time
   --> src/js.rs:994:1
    |
994 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: within `js::RevocableProxy`, the trait `core::marker::Sized` is not implemented for `(dyn core::ops::FnMut() + 'static)`
    = note: to learn more, visit <https://doc.rust-lang.org/book/second-edition/ch19-04-advanced-types.html#dynamically-sized-types--sized>
    = note: required because it appears within the type `js::RevocableProxy`
    = note: the return type of a function must have a statically known size

error: Could not compile `wasm-bindgen`.

To learn more, run the command again with --verbose.
```

srenatus:
@belfz I'd be happy to follow some example, or any other pointers :) (Not trying to make this more effort than having you do it yourself 😉 )

belfz:
@srenatus have you tried defining it inside it's own `extern` block? I mean, this should be similar to what we already have for `ArrayIterator` type :)

srenatus:
@belfz thanks for your help, I did indeed make this more complicated than necessary 😄 Is this closer to what you thought? This still breaks trying to access `RevocableProxy.prototype` (there's no such thing in JS)

belfz:
Right, I wasn't aware of that. Turns out we'll have to probably fallback to `Object` type in that case? cc @alexcrichton

srenatus:
I'll just drop the third commit then -- but I'll wait for @alexcrichton's input 😃

srenatus:
It returns an object with a property 'proxy' that is of type Proxy,
together with a function 'revoke' for revoking it.

So I suppose if we can't encode that in a type somehow, it's got to be
Object?

alexcrichton:
Sounds reasonable to me!

srenatus:
✔️ updated, back to `Object`.

srenatus:
This was failing on travis because one of the jobs seems to have timed out. Rebased, let's see if we're lucky this time.

srenatus:
Ok rebasing introduced something else. Looking into it.

srenatus:
✔️ Caught up with what was merged last night:

- With #381, the tests currently need to be `.js` files, not `.ts` files.
- Had to ensure I've got a recent enough `node` version locally. Tests failed using node 7.x. Using [nodenv](https://github.com/nodenv/nodenv) and 10.4.0 resolved that.

belfz:
LGTM 👍 Thanks, @srenatus!

PR #384, opened 2018-07-04
bindings for parseInt/parseFloat

liigo:
`parseInt`, `parseFloat`, `Number.parseInt`, `Number.parseFloat`

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -93,13 +93,13 @@ extern "C" {
     ///
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
     #[wasm_bindgen(js_name = parseInt)]
-    pub fn parse_int(value: &str, radix: u8) -> JsValue;
+    pub fn parse_int(text: &str, radix: u8) -> JsValue;
 
     /// The parseFloat() function parses an argument and returns a floating point number.
     ///
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat
     #[wasm_bindgen(js_name = parseFloat)]
-    pub fn parse_float(value: &str) -> JsValue;
+    pub fn parse_float(text: &str) -> JsValue;
 }
 
 // UInt8Array
@@ -792,6 +792,20 @@ extern "C" {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number
     #[wasm_bindgen(constructor)]
     pub fn new(value: JsValue) -> Number;
+
+    /// The Number.parseInt() method parses a string argument and returns an 
+    /// integer of the specified radix or base.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseInt
+    #[wasm_bindgen(static_method_of = Number, js_name = parseInt)]
+    pub fn parse_int(text: &str, radix: u8) -> Number;
+
+    /// The Number.parseFloat() method parses a string argument and returns a 
+    /// floating point number.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseFloat
+    #[wasm_bindgen(static_method_of = Number, js_name = parseFloat)]
+    pub fn parse_float(text: &str) -> Number;
 
     /// The toLocaleString() method returns a string with a language sensitive
     /// representation of this number.
--- a/src/js.rs
+++ b/src/js.rs
@@ -87,6 +87,19 @@ extern "C" {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite
     #[wasm_bindgen(js_name = isFinite)]
     pub fn is_finite(value: &JsValue) -> bool;
+
+    /// The `parseInt()` function parses a string argument and returns an integer
+    /// of the specified radix (the base in mathematical numeral systems).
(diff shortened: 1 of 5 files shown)


liigo:
cc #275

alexcrichton:
Thanks! Is `is_nan` required though? I think that `NaN` is a `Number`, right?

jonathan-s:
Also, was there a problem in exposing isNaN the 'normal' way?

Pauan:
@jonathan-s `Number.isNaN` is a recent addition (older browsers like IE don't have it), but despite that I'm in favor of using it. Users can use polyfills if they wish to support older browsers.

liigo:
@alexcrichton 
> NaN compares unequal (via ==, !=, ===, and !==) to any other value -- including to another NaN value.
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN

I don't know how to check a Number is `NaN` without `isNaN` and `Number.isNaN`?

tests/all/js_globals/Number.rs
```diff
@@ -29,6 +29,46 @@ fn new() {
...
+            }
+
+            #[wasm_bindgen]
+            pub fn parse_float(text: &str) -> Number {
+                Number::parse_float(text)
+            }
+        "#,
+        )
+        .file(
+            "test.ts", r#"
```
srenatus: ⚠️ This needs to be `test.js` to become green on travis before we've addressed [this](https://github.com/rustwasm/wasm-bindgen/blob/efa4a2b8fa5da3e5aecd1a71f060c1954fc58cf4/tests/all/project_builder.rs#L233).

fitzgen:
> I don't know how to check a Number is NaN easily without isNaN and Number.isNaN?

One way is to do `x !== x`, which is only true if `x` is `NaN`.

alexcrichton:
@liigo the existing `JsValue::is_*` methods are intended to test for differently typed values, not what sort of value it is within that type. As @fitzgen mentions once a `f64` is extracted it can be tested via a number of methods in Rust to see if it's nan

liigo:
@alexcrichton I've removed `JsValue::is_nan`.

src/js.rs
```diff
@@ -80,6 +80,19 @@ extern "C" {
...
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
+    #[wasm_bindgen(js_name = parseInt)]
+    pub fn parse_int(text: &str, radix: u8) -> JsValue;
+
+    /// The parseFloat() function parses an argument and returns a floating point number.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat
+    #[wasm_bindgen(js_name = parseFloat)]
+    pub fn parse_float(text: &str) -> JsValue;
```
alexcrichton: Should this perhaps return `f64` as I think it always returns a float?

src/js.rs
```diff
@@ -80,6 +80,19 @@ extern "C" {
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval
     #[wasm_bindgen(catch)]
     pub fn eval(js_source_text: &str) -> Result<JsValue, JsValue>;
+
+    /// The `parseInt()` function parses a string argument and returns an integer
+    /// of the specified radix (the base in mathematical numeral systems).
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt
+    #[wasm_bindgen(js_name = parseInt)]
+    pub fn parse_int(text: &str, radix: u8) -> JsValue;
```
alexcrichton: Should this return an `f64` as it returns an integer or `NaN`?

tests/all/js_globals/Number.rs
alexcrichton: I think it's safe to change this to `test.js` now

src/js.rs
liigo: I tried, but ...

```
---- js_globals::parse_int_float stdout ----
···················································
running "cargo" "build" "--target" "wasm32-unknown-unknown"
exit: exit code: 0
stderr ---
   Compiling test0 v0.0.1 (file:///D:/github/other/wasm-bindgen/target/generated-tests/test0)
    Finished dev [unoptimized + debuginfo] target(s) in 0.61s

dur: 0.812s
···················································
running wasm-bindgen
dur: 1.739s
···················································
running "node" "--experimental-modules" "run.mjs"
exit: exit code: 1
stderr ---
(node:23524) ExperimentalWarning: The ESM module loader is experimental.
SyntaxError: Unexpected token import
    at ModuleJob.loaders.set [as moduleProvider] (internal/loader/ModuleRequest.js:32:13)
    at <anonymous>

thread 'js_globals::parse_int_float' panicked at 'assertion failed: output.status.success()', crates\test-project-builder\src\lib.rs:767:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
dur: 0.328s


failures:
    js_globals::parse_int_float
```

the `#[wasm_bindgen]`-ed export function is:

```
export function __wbg_f_parseFloat_parse_float_n(arg0, arg1) {
    let varg0 = getStringFromWasm(arg0, arg1);
    return parseFloat(varg0);  // here returns a js `number`, not `f64`
}
```

tests/all/js_globals/Number.rs
liigo: done.

liigo:
all commemts addressed.

alexcrichton:
👍

PR #385, opened 2018-07-04
bindings for date.getDay/getFullYear

toVersus:
#275


belfz:
I think we'll have to decide on the return values - I see `f64` used interchangeably with the `Number`. Otherwise, this PR looks good to me. Thanks, @toVersus!

PR #386, opened 2018-07-04
Reflect.xx bindings

jannikkeye:
This PR adds all the bindings for `Reflect`.
I wasn't sure how to handle exception throwing on the rust side.


src/js.rs
```diff
@@ -978,6 +978,114 @@ extern "C" {
...
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/getPrototypeOf
+    #[wasm_bindgen(static_method_of = Reflect, js_name = getPrototypeOf, catch)]
+    pub fn get_prototype_of(target: &JsValue) -> Result<JsValue, JsValue>;
+
+    /// The static Reflect.has() method works like the in operator as a function.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/has
+    #[wasm_bindgen(static_method_of = Reflect, catch)]
+    pub fn has(target: &JsValue, property_key: &JsValue) -> Result<JsValue, JsValue>;
```
belfz: I think that this signature is incorrect. The `target` should rather be of `&Object` type, thus eliminating a need for `catch` (`TypeError` is only thrown when `target` is not an `Object` - but here we avoid that problem at type-checking level).

Could you please review other methods and apply the change (where needed)?

tests/all/js_globals/Reflect.rs
```diff
@@ -0,0 +1,703 @@
...
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js;
+
+            #[wasm_bindgen]
+            pub fn apply(target: &js::Function, this_argument: &JsValue, arguments_list: &js::Array) -> JsValue {
+                let result = js::Reflect::apply(target, this_argument, arguments_list);
+                let result = match result {
```
belfz: Could you avoid name shadowing? This code could be a bit more concise with:
```rust
match result {
    Ok(val) => val,
    Err(_err) => "TypeError".into(),
}
```

`match` is an expression that returns value - note that I omitted the semicolon at the end :)

tests/all/js_globals/Reflect.rs
```diff
@@ -0,0 +1,703 @@
...
+                    Ok(val) => val,
+                    Err(_err) => "TypeError".into()
```
belfz: Have you looked up the contents of `_err`? Maybe it could be used as a return value here (I'm not sure what's there, but it could be worth checking).

tests/all/js_globals/Reflect.rs
jannikkeye: This I actually took from [here](https://github.com/rustwasm/wasm-bindgen/blob/2694dd4dee28e36981e026e60d5bdf141f51562d/tests/all/js_globals/Number.rs#L75) without thinking to much. You are absolutely right. Maybe those tests could use some refactoring there as well for future reference.

tests/all/js_globals/Reflect.rs
jannikkeye: _err indeed contains the actual error message. Again, we should probably apply those changes [here](https://github.com/rustwasm/wasm-bindgen/blob/2694dd4dee28e36981e026e60d5bdf141f51562d/tests/all/js_globals/Number.rs#L75) as well.

src/js.rs
jannikkeye: I checked all other functions and this applies to all of them. Thanks for clarifying!

tests/all/js_globals/Reflect.rs
belfz: great 👍

src/js.rs
```diff
@@ -978,6 +978,114 @@ extern "C" {
...
+
+    /// The static Reflect.construct() method acts like the new operator, but as a function.
+    /// It is equivalent to calling new target(...args). It gives also the added option to 
+    /// specify a different prototype.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/construct
+    #[wasm_bindgen(static_method_of = Reflect)]
+    pub fn construct(target: &Function, arguments_list: &Array) -> JsValue;
+    #[wasm_bindgen(static_method_of = Reflect, js_name = construct)]
+    pub fn construct_with_new_target(target: &Function, arguments_list: &Array, new_target: &Function) -> JsValue;
```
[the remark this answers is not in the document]
jannikkeye: This will probably not be as good if there are multiple optional properties. If this is how optional properties will be handled it will have to be documented properly.

srenatus:
ℹ️ Just a heads-up: These test files need to be `test.js` to become green on travis before we've addressed [this](https://github.com/rustwasm/wasm-bindgen/blob/efa4a2b8fa5da3e5aecd1a71f060c1954fc58cf4/tests/all/project_builder.rs#L233).

belfz:
Looks good to me! Thank you @jannikkeye and @srenatus !

PR #387, opened 2018-07-04
feat(js) Implement the `WebAssembly.validate` binding.

Hywan:
Cf. https://github.com/rustwasm/wasm-bindgen/issues/275.

`validate` should accept only a typed array or an `ArrayBuffer`, but I feel like it is not possible to get a strict type here (there is no common type in JS to represent to type). So I typed with `JsValue`.

I don't know how to support exception. And I don't even know the code will behave if an exception will be thrown :-). Any clue?

### DIFF
--- a/tests/all/js_globals/WebAssembly.rs
+++ b/tests/all/js_globals/WebAssembly.rs
@@ -31,3 +31,38 @@ fn validate() {
         "#)
         .test()
 }
+
+#[test]
+fn validate_with_invalid_input() {
+    project()
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use JsValue;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js::WebAssembly;
+
+            #[wasm_bindgen]
+            pub fn validate_wasm(wasm: JsValue) -> JsValue {
+                match WebAssembly::validate(wasm) {
+                    Ok(value) => value.into(),
+                    Err(err) => err
+                }
+            }
+        "#)
+        .file("test.js", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                try {
+                    wasm.validate_wasm(42);
+                    assert.ok(false);
+                } catch (e) {
+                    assert.ok(true);
+                }
+            }
+        "#)
+        .test()
+}
--- a/tests/all/js_globals/WebAssembly.rs
+++ b/tests/all/js_globals/WebAssembly.rs
@@ -1,0 +1,30 @@
+#![allow(non_snake_case)]
+
+use super::project;
+
+#[test]
+fn validate() {
+    project()
+        .file("src/lib.rs", r#"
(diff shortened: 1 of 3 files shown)


belfz:
For exceptions, check out the `catch` attribute, eg:
```rust
#[wasm_bindgen(catch, js_name = decodeURI)]
pub fn decode_uri(encoded: &str) -> Result<JsString, JsValue>;
```
here, `catch` means that this method can throw an error - on rust's side it is handled with the [Result type](https://doc.rust-lang.org/std/result/) (1st type parameter represents a "success" type, while the second is an error) :)

Hywan:
Thanks!

tests/all/js_globals/WebAssembly.rs
```diff
@@ -0,0 +1,30 @@
...
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use JsValue;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js::WebAssembly;
+
+            #[wasm_bindgen]
+            pub fn validate_wasm(wasm: JsValue) -> bool {
+                WebAssembly::validate(wasm).unwrap_or(false)
```
belfz: Hmm. This will "swallow" a possible error, so you can't really test for that case. Could you take a look at how methods that possibly throw are tested (eg. `tests/all/js_globals/Number.rs`). I am thinking about something like:
```rust
match WebAssembly::validate(wasm) {
    Ok(value) => value,
    Err(_err) => "TypeError".into(), // additionally, you can check the contents of the '_err' - I am not sure now, but I think it should contain the "TypeError" string
}
```

Hywan: In this case, the output type is `bool` or `String`, and I'm not sure how to type that :-).

belfz: You're right - sorry for that confusion! I wonder whether simply returning a result from that method would work for both success and error cases on js side (so that the error could be caught within `try/catch` or `assert.throws`).

belfz: ...or, we can try a testing strategy similar to that of `decode_uri_component` in `test/all/js_globals/mod.rs` 🙂

jannikkeye: Btw. the contents of _err are the actual javascript error message.
Istead of `Err(_err) => "TypeError".into()` you can just do `Err(_err) => _err`.

tests/all/js_globals/WebAssembly.rs
```diff
@@ -0,0 +1,30 @@
...
+            }
+        "#)
+        .file("test.ts", r#"
```
srenatus: ⚠️ This needs to be `test.js` to become green on travis before we've addressed [this](https://github.com/rustwasm/wasm-bindgen/blob/efa4a2b8fa5da3e5aecd1a71f060c1954fc58cf4/tests/all/project_builder.rs#L233).

Hywan:
I've added a new test case. I don't use `assert.throws` because it somehow bugs. A simple `try`/`catch` does the job well.

belfz:
Thanks @Hywan, looks good to me.

PR #388, opened 2018-07-04
feat(js) Add `ArrayBuffer::new`, `::is_view` & `::slice` bindings.

Hywan:
Cf. https://github.com/rustwasm/wasm-bindgen/issues/275

My wish was to include `get ArrayBuffer[@@species]` but I've no idea how to tackle this binding :-). Any recommendation?

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -254,6 +254,39 @@ extern "C" {
     pub fn unshift(this: &Array, value: JsValue) -> u32;
 }
 
+// ArrayBuffer
+#[wasm_bindgen]
+extern "C" {
+    pub type ArrayBuffer;
+
+    /// The `ArrayBuffer` object is used to represent a generic,
+    /// fixed-length raw binary data buffer. You cannot directly
+    /// manipulate the contents of an `ArrayBuffer`; instead, you
+    /// create one of the typed array objects or a `DataView` object
+    /// which represents the buffer in a specific format, and use that
+    /// to read and write the contents of the buffer.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer
+    #[wasm_bindgen(constructor)]
+    pub fn new(length: f64) -> ArrayBuffer;
+
+    /// The `slice()` method returns a new `ArrayBuffer` whose contents
+    /// are a copy of this `ArrayBuffer`'s bytes from begin, inclusive,
+    /// up to end, exclusive.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice
+    #[wasm_bindgen(static_method_of = ArrayBuffer, js_name = isView)]
+    pub fn is_view(value: JsValue) -> bool;
+
+    /// The `slice()` method returns a new `ArrayBuffer` whose contents
+    /// are a copy of this `ArrayBuffer`'s bytes from begin, inclusive,
+    /// up to end, exclusive.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice
+    #[wasm_bindgen(method)]
+    pub fn slice(this: &ArrayBuffer, begin: f64) -> ArrayBuffer;
+}
+
 // Array Iterator
 #[wasm_bindgen]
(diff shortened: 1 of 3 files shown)


tests/all/js_globals/ArrayBuffer.rs
```diff
@@ -0,0 +1,83 @@
...
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js::ArrayBuffer;
+
+            #[wasm_bindgen]
+            pub fn slice_at_2(arraybuffer: ArrayBuffer) -> ArrayBuffer {
+                ArrayBuffer::slice(&arraybuffer, 2f64)
```
belfz: `slice` is an instance method. Could you please us `arraybuffer.slice(2f64)` instead?

tests/all/js_globals/ArrayBuffer.rs
```diff
@@ -0,0 +1,83 @@
...
+    project()
+        .file("src/lib.rs", r#"
+            #![feature(proc_macro, wasm_custom_section)]
+
+            extern crate wasm_bindgen;
+            use wasm_bindgen::prelude::*;
+            use wasm_bindgen::js::ArrayBuffer;
+
+            #[wasm_bindgen]
+            pub fn slice_at_2(arraybuffer: ArrayBuffer) -> ArrayBuffer {
```
belfz: this argument should be passed as a reference (`&ArrayBuffer`) :)

src/js.rs
```diff
@@ -254,6 +254,39 @@ extern "C" {
...
+    #[wasm_bindgen(static_method_of = ArrayBuffer, js_name = isView)]
+    pub fn is_view(value: JsValue) -> bool;
+
+    /// The `slice()` method returns a new `ArrayBuffer` whose contents
+    /// are a copy of this `ArrayBuffer`'s bytes from begin, inclusive,
+    /// up to end, exclusive.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice
+    #[wasm_bindgen(method)]
+    pub fn slice(this: &ArrayBuffer, begin: f64) -> ArrayBuffer;
```
belfz: we might need another method here, that takes into account an optional `end` argument - eg. `pub fn slice_with_end(this: &ArrayBuffer, begin: f64, end: f64) -> ArrayBuffer`

alexcrichton: I think the arguments here should be integers as well since they're indices as well, right?

src/js.rs
```diff
@@ -254,6 +254,39 @@ extern "C" {
...
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer
+    #[wasm_bindgen(constructor)]
+    pub fn new(length: f64) -> ArrayBuffer;
+
+    /// The `slice()` method returns a new `ArrayBuffer` whose contents
+    /// are a copy of this `ArrayBuffer`'s bytes from begin, inclusive,
+    /// up to end, exclusive.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice
```
alexcrichton: This URL I think should be https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView?

src/js.rs
belfz: True!

tests/all/js_globals/ArrayBuffer.rs
Hywan: Fixed :-).

tests/all/js_globals/ArrayBuffer.rs
Hywan: Fixed :-).

src/js.rs
Hywan: Fixed :-). Good catch!

src/js.rs
Hywan: Fixed, and fixed :-).

Hywan:
Thanks for the review! It is a perfect illustration of: “Why you should not write a PR 5mn before a meeting”.

PR #393, opened 2018-07-05
Test release builds on Travis

alexcrichton:
Profiling locally shows a 3x speed improvement by testing release builds,
presumably due to the amount of wasm processing happening! Let's see what
happens on Travis and if we get a similar speedup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
         # to run `npm install` twice or by using `npm ci` (which is currently broken)
         - npm install
       script:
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks like [17 minutes](https://travis-ci.org/rustwasm/wasm-bindgen/jobs/400462448) compared to [15 minutes](https://travis-ci.org/rustwasm/wasm-bindgen/builds/400570504), so not as big of a gain on Travis!

PR #394, opened 2018-07-05
js: Return scalar types instead of `Number` objects

fitzgen:
99% of the time we don't want to deal with opaque handles to `Number`
objects.


fitzgen:
appveyor is super backed up, going to go ahead and merge this for now.

PR #395, opened 2018-07-05
Add Array.prototype.values binding

brisad:


fitzgen:
@brisad does the version of Node you ahve installed have `Array#values`?

Also, I think we already have bindings for this method -- maybe we forgot to check it off.

fitzgen:
> Also, I think we already have bindings for this method -- maybe we forgot to check it off.

NVM, I misread.

brisad:
@fitzgen Oh, good question. It seems like it doesnt:

```
$ node
> let a = [1,2,3];
undefined
> a.entries()
Object [Array Iterator] {}
> a.values()
TypeError: a.values is not a function
```
I also just upgraded from Node v8 to v10.

Travis CI seems to have the same problem, so it appears it is not just me.

brisad:
And the Browser compatibility section of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/values indeed states that Node.js does not have support for it.

fitzgen:
@brisad you could polyfill `Array.prototype.find` in the test.js file -- want to try that?

alexcrichton:
Alternatively this could perhaps use `.headless(true)` to run it in a browser?

brisad:
@fitzgen I tried assigning a `values` function to `Array.prototype`, if that is what you meant. I did it in the `test.js` file, both outside and inside the test function, but it does not seem to take. I get the same error every time.

@alexcrichton For some reason running any `headless`-tests I run locally seem to never finish. I tried pushing up to CI with an addition to this PR, but that also failed.

brisad:
@alexcrichton Sure! After looking more closely it turned out I didn't have `webpack-dev-server` installed for some reason. So after running `yarn` again I could run the test. Then I could see that I had made a mistake of adding square brackets in the call to `new Array()`.

I updated the PR with the fixes so now everything should be set.

I realized that the dev server wasn't running when `TcpStream::connect("127.0.0.1:8080")` never succeeded. Perhaps we can add some "Connecting to dev server..." debug message, along with a "connected" confirmation, to aid debugging? I can add something in a PR if you want?

alexcrichton:
Ok great! Let's merge this once tests are green.

Oh dear that sounds like a nasty bug though! Let's do a separate PR for that? I think it'd be reasonable to place a cap on the number of times we connect to 8080 before printing an error and failing the test?

PR #400, opened 2018-07-06
Forward attributes on `type` declaration to definition

alexcrichton:
This'll allow things like `#[derive(Clone)]` or `#[derive(Debug)]` to control
traits for these types.


fitzgen:
Er -- nvm, these are *imports* of JS things not *exports* of Rust things.

PR #401, opened 2018-07-06
Add support for static props of Symbol

rail44:


rail44:
How should I do about `Symbol.unscopable` ?
Although working with `with` statement, but it's unavailable in strict mode.

fitzgen:
> How should I do about Symbol.unscopable ? Although working with with statement, it's unavailable in strict mode.

I'm not sure I follow -- can you explain in more detail?

PR #410, opened 2018-07-07
Generate normal Rust functions for imported bindings

alexcrichton:
No need to use the C ABI, it just mixes up the documentation!


ohanar:
It looks like the end of the generated output got cut off...

PR #411, opened 2018-07-07
Backend refactor

ohanar:
I was getting tired trying to figure out what `BindgenAttrs` actually were used in codegen when working on some things for WebIDL, so I decided to try to refactor the backend crate to make it more clear. This moves all logic that is specific to macros to the macro crate, so that the backend crate is now simply for codegen (and a few common util functions).


alexcrichton:
Awesome refactoring, thanks!

PR #415, opened 2018-07-07
uses (JS) `const` for stack and slab declarations

belfz:
While this is rather a minor thing, while [reading through the guide](https://rustwasm.github.io/wasm-bindgen/design/js-objects-in-rust.html) I noticed that for JS `stack` and `slab` the `let` keyword is used, where in fact it should be a `const`.


alexcrichton:
Thanks!

PR #417, opened 2018-07-07
Add Array.prototype.forEach bindings

NAlexPear:
This PR should add bindings to `forEach`.

This is also my first PR to `wasm-bindgen`, so let me know if I've goofed anything up. Possible places for improvement:

1. This implementation is returning a `JsValue`, when I suspect it really ought to return a `JsValue::undefined()`, but my Rust-fu wasn't strong enough to get that granular. Let me know if `JsValue` isn't restrictive enough.
2. The tests are really calling out for a spy _a la_ `sinon`, but I didn't want to add another dependency to the project for one test.

Referencing issue #275.

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -148,6 +148,12 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn find(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool) -> JsValue;
 
+    /// The forEach() method executes a provided function once for each array element.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
+    #[wasm_bindgen(method, js_name = forEach)]
+    pub fn for_each(this: &Array, callback: &Function) -> JsValue;
+
     /// The includes() method determines whether an array includes a certain
     /// element, returning true or false as appropriate.
     ///
--- a/tests/all/js_globals/Array.rs
+++ b/tests/all/js_globals/Array.rs
@@ -42,6 +42,57 @@ fn filter() {
 }
 
 #[test]
+fn for_each() {
+    project()
+        .file(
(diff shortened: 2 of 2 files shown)


src/js.rs
```diff
@@ -148,6 +148,12 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn find(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool) -> JsValue;
 
+    /// The forEach() method executes a provided function once for each array element.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
+    #[wasm_bindgen(method, js_name= forEach)]
```
FrankHassanabad: space between `js_name` and `forEach` like so? `js_name = forEach)]`

NAlexPear: Good catch, just fixed.

src/js.rs
```diff
@@ -148,6 +148,12 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn find(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool) -> JsValue;
 
+    /// The forEach() method executes a provided function once for each array element.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach
+    #[wasm_bindgen(method, js_name = forEach)]
+    pub fn for_each(this: &Array, callback: &Function) -> JsValue;
```
alexcrichton: I think the `&Function` may want to perhaps be a variant of `&mut FnMut(...)` to invoke the closure in Rust?

fitzgen: Yes, we want to use `&mut FnMut(...)` for the callback.

@FrankHassanabad, see the bindings for `find` or `filter` for examples.

PR #419, opened 2018-07-08
adds julia set example

belfz:
[View this example online](https://webassembly.studio/?f=qmg1c6dbjos)

This is an example of more complex computations needed to generate a fractal (Julia Set).
It's also an example of direct interaction with CanvasRenderingContext2D's API via `#[wasm_bindgen]` macro.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ members = [
   "examples/smorgasboard",
   "examples/console_log",
   "examples/dom",
+  "examples/julia_set",
   "examples/math",
   "examples/performance",
   "examples/wasm-in-wasm",
--- a/examples/julia_set/.gitignore
+++ b/examples/julia_set/.gitignore
@@ -1,0 +1,2 @@
+julia_set.js
+julia_set_bg.wasm
--- a/examples/julia_set/Cargo.toml
+++ b/examples/julia_set/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "julia_set"
+version = "0.1.0"
+authors = ["Marcin Baraniecki <marcinbar1@gmail.com>"]
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+wasm-bindgen = { path = "../.." }
--- a/examples/julia_set/README.md
+++ b/examples/julia_set/README.md
@@ -1,0 +1,14 @@
+# Julia Set
+
+[View this example online](https://webassembly.studio/?f=qmg1c6dbjos)
+
+This is an example of more complex computations needed to generate a fractal (Julia Set).
+It's also an example of direct interaction with CanvasRenderingContext2D's API via `#[wasm_bindgen]` macro.
+
+You can build the example with:
+
+```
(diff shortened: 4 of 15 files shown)


alexcrichton:
This is awesome! Quite the fun demo to play with :)

I wonder if we might want to try to show off the ability to do this with zero copies? It should be possible to create a `Uint8ClampedArray` without actually copying any data which goes straight to the canvas I think?

alexcrichton:
Oh sure! So right now you're doing:

```rust
#[wasm_bindgen]
extern "C" {
    pub type Uint8ClampedArray;

    #[wasm_bindgen(constructor)]
    pub fn new(arr: &[u8]) -> Uint8ClampedArray;
}
```

where `Uint8ClampedArray::new(arr)` is actually copying all the data from `arr` to the `Uint8ClampedArray` instance. What happens here in JS is that a binding like this is generated:

```js
function __wbindgen_Uint8ClampedArray_new(ptr, len) {
    return new Uint8ClampedArray(getUint8Memory().subarray(ptr, ptr + len))
}
```

where `getUint8Memory().subarray(...)` is basically taking a slice of wasm's memory, not copying any data but creating a `Uint8Array`. Afterwards though `new Uint8ClampedArray(t: Uint8Array)` is copying all the data from one array to another. This means that the `Uint8ClampedArray` instance is actually copying all the data from wasm's memory into a different array and saving it for later processing.


What we basically want is something like this:

```rust
#[wasm_bindgen]
extern "C" {
    pub type ImageData;

    #[wasm_bindgen(constructor, clamped)]
    pub fn new(arr: &[u8], width: u32, height: u32) -> ImageData;
}
```

Note that this isn't implemented today! But ideally we'd have the `new ImageData(..)` constructor directly take a `Uint8ClampedArray` that's just a slice into wasm's own memory.

Now that I say all this though it may be pretty beyond this PR though to do something like this. There's a few reasons this isn't great:

* First we'd have to design a way to use `Uint8ClampedArray` in JS instead of `Uint8Array`.
* Next, this API is actually a "lie" in the sense that `ImageData` internally should look like `ImageData<'arr>` but the type signature doesn't say that, meaning it's probably easy to get some corruption
* And finally, this PR is just adding a cool example, and this is spinning a bit large now!

What do you think? I'd be tempted to merge this and perhaps open an issue about it?

belfz:
Wow! Honestly, I didn't even think much about how the memory is managed under the hood in that case. But that's a nice lesson, and moreover - I think that this might be a nice use case for future implementations of things you just mentioned.

I agree with you - let's merge that, keep it as a nice example to show off and then improve it when we're done with necessary changes!

Thanks again, Alex! I'll also try to understand the challenge harder because right now, these low-level interactions within the memory are still relatively new stuff for me :)

alexcrichton:
Ok sounds great to me!

alexcrichton:
Ok I've opened https://github.com/rustwasm/wasm-bindgen/issues/421 and https://github.com/rustwasm/wasm-bindgen/issues/423 to cover these issues

PR #429, opened 2018-07-09
[WIP] Test out various configurations to speed up CI

alexcrichton:
This is similar to #393 but I'm also curious what the effect of the number of
test threads has...


alexcrichton:
Looks like it [is indeed a big win](https://travis-ci.org/rustwasm/wasm-bindgen/builds/401819001?utm_source=github_status&utm_medium=notification) now for release builds! I'll submit a separate PR.

PR #430, opened 2018-07-09
Js docs

### LINKED ISSUE #57: Copy doc comments from Rust to JS

FreeMasen:
This addresses two of the 3 remaining items in #57, all user exported functions will now be annotated with the JSDoc `@param` and `@returns` statements.

Upon further investigation into applying the `@constructor` statement I discovered that we are still using the `class` keyword when using `no-modules`, meaning that this would not be required.

This should effectively close #57 unless we are looking to provide an output that uses the function constructor + prototype syntax.

### DIFF
--- a/crates/cli-support/src/js/js2rust.rs
+++ b/crates/cli-support/src/js/js2rust.rs
@@ -368,6 +368,14 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
         Ok(self)
     }
 
+    pub fn js_doc_comments(&self) -> String {
+        let mut ret: String = self.js_arguments.iter().map(|a| {
+            format!("@param {{{}}} {}\n", a.1, a.0)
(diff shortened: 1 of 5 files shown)


FreeMasen:
Sorry about the vague commit messages, I was attempting to open this against my fork only but apparently that isn't a thing GitHub can do. I will squash out all the junk after the travis tests run.

FreeMasen:
Looks like this is currently breaking the JS formatting due to the JSDoc convention of including type information inside of `{}`. Since these new lines end with a `}` they are being incorrectly formatted.

#427 is pending and re-writes the formatting code entirely so I am going to hold out on a fix until that gets merged.

Issue #260, opened 2018-06-08
Support WebIDL enumerations


fitzgen:
https://heycam.github.io/webidl/#idl-enums

```webidl
enum MealType { "rice", "noodles", "other" };
```

These are always string literals.

I don't think we want to translate these into Rust `enum`s because Rust is allowed to assume they are exhaustive, meaning that if the definitions are extended in the future, we would get UB in Rust.

Instead, we can translate it into something like:

```rust
mod MealType {
    const Rice: &'static str = "rice";
    // ...
}

// usage:
foo(MealType::Rice);
```

ohanar:
I would also vote for mapping them to rust enum's, especially given that the non-exhaustive enum's rfc was accepted and has already been implemented in the nightly compiler.

fitzgen:
SGTM!

PR #433, opened 2018-07-10
WebIDL Enum Support

Slowki:
Adds basic support for enums to the WebIDL interface generator.

Per the consensus on #260 the enums are represented as normal Rust enums. Right now I'm just having it panic when an invalid string is converted, I was thinking of adding something along the lines of [this nice hack](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md#how-we-do-this-today) instead, but I wanted some feedback first to see if there's a better approach.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -626,7 +626,7 @@ impl ToTokens for ast::ImportEnum {
             }
 
             impl #name {
-                fn from_js_value(obj: ::wasm_bindgen::JsValue) -> Option<#name> {
+                #vis fn from_js_value(obj: ::wasm_bindgen::JsValue) -> Option<#name> {
                     obj.as_string().and_then(|obj_str| match obj_str.as_str() {
                         #(#variant_strings => Some(#variant_paths_ref),)*
                         _ => None,
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -4,7 +4,7 @@ use std::env;
 use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
 
 use ast;
-use proc_macro2::{Ident, Span, TokenStream};
+use proc_macro2::{Ident, Literal, Span, TokenStream};
 use quote::ToTokens;
 use serde_json;
 use shared;
@@ -500,6 +500,7 @@ impl ToTokens for ast::ImportKind {
             ast::ImportKind::Function(ref f) => f.to_tokens(tokens),
             ast::ImportKind::Static(ref s) => s.to_tokens(tokens),
             ast::ImportKind::Type(ref t) => t.to_tokens(tokens),
+            ast::ImportKind::Enum(ref e) => e.to_tokens(tokens),
         }
     }
 }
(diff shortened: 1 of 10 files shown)


ohanar:
I haven't had the time to look through your code thoroughly (will make sure to get a chance tomorrow), but upon a first pass through everything seems fine.

I think the hack you suggested sounds like a good approach, nothing better comes to mind immediately (when I commented on the issue, I think I was thinking of doing the same thing).

crates/backend/src/codegen.rs
```diff
@@ -586,6 +587,98 @@ impl ToTokens for ast::ImportType {
...
+                }
+            }
+
+            impl ::wasm_bindgen::convert::IntoWasmAbi for #name {
+                type Abi = <::wasm_bindgen::JsValue as
+                    ::wasm_bindgen::convert::IntoWasmAbi>::Abi;
+
+                fn into_abi(self, extra: &mut ::wasm_bindgen::convert::Stack) -> Self::Abi {
+                    match self {
+                        #(#variant_paths_ref => ::wasm_bindgen::JsValue::from_str(#variant_strings).into_abi(extra),)*
```
alexcrichton: Could this use the `From` impl below to convert this to a `JsValue`?

alexcrichton:
Nice! Just one small comment but otherwise this looks great to me

crates/backend/src/codegen.rs
```diff
@@ -586,6 +587,98 @@ impl ToTokens for ast::ImportType {
...
+                ) -> Self {
+                    #name::from(::wasm_bindgen::JsValue::from_abi(js, extra))
+                }
+            }
+
+            impl From<::wasm_bindgen::JsValue> for #name {
+                fn from(obj: ::wasm_bindgen::JsValue) -> #name {
+                    let obj_str = match obj.as_string() {
+                        Some(string_value) => string_value,
+                        None => panic!("Can't convert a non-string into {}", #name_string),
```
fitzgen: I agree that when we receive a non-string value or non-enum string value as an argument to a function that expects this enum, we should panic.

However, I don't think we should implement `From<JsValue>` for these and have it panic. This is too much of a footgun for folks who are just reading docs and see the impl and don't know what it does under the hood.

Instead, I think we should make a normal, non-trait `pub fn from_js_value(JsValue) -> Option<Self>` method, and the `FromWasmAbi` trait impl can `unwrap` the result. This gives us the same behavior when sending them to and from the ABI, but without the footgun.

crates/backend/src/codegen.rs
Slowki: :+1:

crates/backend/src/codegen.rs
Slowki: I think panicking in `FromWasmAbi` is something we should avoid so that we can recover in cases where JS functions return invalid string values.
### Example ###
```webidl
enum Foo { "bar", "baz" }
Foo getFoo();
```

```js
getFoo() {
    return "quux"
}
```

```rust
// Panics
match getFoo() {
    Foo::Bar => ...,
    Foo::Baz => ...,
}
```

### Possible Solutions ###
* We could use this [idiom from libstd](https://github.com/rust-lang/rust/blob/master/src/libstd/io/error.rs#L190) and return `__Invalid` instead of panicking
```rust
enum Foo {
    Bar,
    Baz,
    #[doc(hidden)]
    __Invalid,
}
```
* We could just expose a catch-all variant
```rust
enum Foo {
    Bar,
    Baz,
    InvalidValue(String),
}
```
* Or we could make WebIDL functions which return enums return `Result`s.

ohanar: +1. I definitely wouldn't want my web app to panic if a browser gets updated with an updated version of a standard, and I haven't gotten around to upgrading my copy of `web-sys`.

The main downside to including the invalid value in the actual enum is that the enum no longer can be copy. I would be inclined to either just include an `__Invalid` variant, or have WebIDL functions returning enums to instead return `Result`s, with an error type that contains the invalid string.

Slowki: I like the `Result` solution the best, it seems like it's the most convenient for the end user. The only downside is that functions that throw will end up returning `Result<Result<Enum, String>, ...>`, but there might be a nice way to flatten that out.

crates/backend/src/codegen.rs
```diff
@@ -586,6 +587,91 @@ impl ToTokens for ast::ImportType {
...
+
+        (quote! {
+            #[allow(bad_style)]
+            #[derive(Copy, Clone, Debug)]
+            #vis enum #name {
+                #(#variants = #variant_indexes_ref,)*
+            }
+
+            impl #name {
+                fn from_js_value(obj: ::wasm_bindgen::JsValue) -> Option<#name> {
```
alexcrichton: This may want to use `#vis` out in front to allow this to be used outside the module perhaps?

crates/backend/src/codegen.rs
fitzgen: Another advantage of having non exhaustive `enum`s is that we won't have to do breaking version bumps when updating WebIDLs.

ohanar: I am working on adding support for partial and mixin interfaces right now. As part of those changes, I am doing a two pass process that would make it trivial to implement the `Result` return type solution. (This is in part to fix a bug in the current crate where interface objects are being passed by value, rather than by reference, as per the spec.)

I should have time today to finish up a first draft, and submit a PR. It is based off of this branch, so I would return to the non-exhaustive `enum` problem in a follow-up PR (especially if you want to go down the `Result` path).

Slowki: @ohanar Sounds good. I'll make `from_js_value` visible and resolve the merge conflict, then I'm good with this being merged.

@fitzgen That's a good point, definitely a huge + for non-exhaustive over `Result`s.

crates/backend/src/codegen.rs
Slowki: :+1:

Slowki:
Feel free to merge, I'll make a separate PR for the non-exhaustive enums issue.

fitzgen:
> Feel free to merge, I'll make a separate PR for the non-exhaustive enums issue.

SGTM! Thanks @Slowki :)

PR #434, opened 2018-07-10
fix `GlobalStack::pop()`

liigo:
I'm not very sure if this is a good fix. Seems like no direct usage of `GlobalStack::pop()` through out the tree.


alexcrichton:
Thanks for the PR! I think that this is actually unused nowadays, but the previous implementation was the intended implementation for this feature. Since it's unused, though, want to remove it entirely?

liigo:
`into_abi(&mut GlobalStack::new())`

Here `GlobalStack` is used as trait `Stack`, so we can't remove `impl Stack for GlobalStack` entirely.
Just mark its `push` and `pop` methods as `unreachable!()`? This feels uncomfortable.

alexcrichton:
Ah yes what I mean is removing the `pop` function from `GlobalStack` as well as the implementation, I don't believe its used anywhere any more

PR #436, opened 2018-07-10
js::DataView binding

liigo:
All done!

- [x] DataView 
- [x] DataView.prototype.buffer
- [x] DataView.prototype.byteLength
- [x] DataView.prototype.byteOffset
- [x] DataView.prototype.getFloat32()
- [x] DataView.prototype.getFloat64()
- [x] DataView.prototype.getInt16()
- [x] DataView.prototype.getInt32()
- [x] DataView.prototype.getInt8()
- [x] DataView.prototype.getUint16()
- [x] DataView.prototype.getUint32()
- [x] DataView.prototype.getUint8()
- [x] DataView.prototype.setFloat32()
- [x] DataView.prototype.setFloat64()
- [x] DataView.prototype.setInt16()
- [x] DataView.prototype.setInt32()
- [x] DataView.prototype.setInt8()
- [x] DataView.prototype.setUint16()
- [x] DataView.prototype.setUint32()
- [x] DataView.prototype.setUint8()

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -378,6 +378,152 @@ extern "C" {
     pub fn value_of(this: &Boolean) -> bool;
 }
 
+// DataView
+#[wasm_bindgen]
+extern "C" {
+    pub type DataView;
+
+    /// The `DataView` view provides a low-level interface for reading and 
+    /// writing multiple number types in an `ArrayBuffer` irrespective of the
+    /// platform's endianness.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView
+    #[wasm_bindgen(constructor)]
+    pub fn new(buffer: &ArrayBuffer, byteOffset: usize, byteLength: usize) -> DataView;
+
+    /// The ArrayBuffer referenced by this view. Fixed at construction time and thus read only.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/buffer
+    #[wasm_bindgen(method)]
+    pub fn buffer(this: &DataView) -> ArrayBuffer;
+
+    /// The length (in bytes) of this view from the start of its ArrayBuffer.
+    /// Fixed at construction time and thus read only.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/byteLength
+    #[wasm_bindgen(method, js_name = byteLength)]
+    pub fn byte_length(this: &DataView) -> usize;
+
+    /// The offset (in bytes) of this view from the start of its ArrayBuffer.
+    /// Fixed at construction time and thus read only.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/byteOffset
+    #[wasm_bindgen(method, js_name = byteOffset)]
+    pub fn byte_offset(this: &DataView) -> usize;
+
+    /// The getInt8() method gets a signed 8-bit integer (byte) at the 
+    /// specified byte offset from the start of the DataView.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getInt8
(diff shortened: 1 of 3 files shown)


tests/all/js_globals/DataView.rs
```diff
@@ -0,0 +1,54 @@
...
+                assert_eq!(v.get_float64(0), 123456789.123456);
+
+                v.set_int8(0, 4+2);
+            }
+        "#)
+        .file("test.js", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
```
liigo: It seems like this function was not invoked normally?

src/js.rs
```diff
@@ -369,6 +369,152 @@ extern "C" {
...
+    /// platform's endianness.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView
+    #[wasm_bindgen(constructor)]
+    pub fn new(buffer: &ArrayBuffer, byteOffset: usize, byteLength: usize) -> DataView;
+
+    /// The ArrayBuffer referenced by this view. Fixed at construction time and thus read only.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/buffer
+    #[wasm_bindgen(method)]
```
alexcrichton: I think this should be a `getter` which is causing the tests to fail?

src/js.rs
```diff
@@ -369,6 +369,152 @@ extern "C" {
...
+    pub fn buffer(this: &DataView) -> ArrayBuffer;
+
+    /// The length (in bytes) of this view from the start of its ArrayBuffer.
+    /// Fixed at construction time and thus read only.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/byteLength
+    #[wasm_bindgen(method, js_name = byteLength)]
```
alexcrichton: Like above, I think this should be a `getter`

(perhaps paired with a `setter` as well?

liigo:
@alexcrichton Thank you! `buffer` / `byteOffset` / `byteLenght` are read-only properties, so only getter, no setter.

liigo:
done.

tests/all/js_globals/DataView.rs
```diff
@@ -0,0 +1,55 @@
...
+
+                v.set_int8(0, 42);
+            }
+        "#)
+        .file("test.js", r#"
+            import * as assert from "assert";
+            import * as wasm from "./out";
+
+            export function test() {
+                var bytes = new Int8Array(10);
```
belfz: That's a minor thing, but could you use `const` instead of `var` here? Thanks! :-)

src/js.rs
```diff
@@ -378,6 +378,152 @@ extern "C" {
...
+    pub fn byte_offset(this: &DataView) -> usize;
+
+    /// The getInt8() method gets a signed 8-bit integer (byte) at the 
+    /// specified byte offset from the start of the DataView.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getInt8
+    #[wasm_bindgen(method, js_name = getInt8)]
+    pub fn get_int8(this: &DataView, byte_offset: usize) -> i8;
+
+    /// The getInt8() method gets a signed 8-bit integer (byte) at the specified
```
belfz: Looks like it's a description of `getInt8` for `get_uint8` (note: unsigned int) here

liigo: thank you!

tests/all/js_globals/DataView.rs
liigo: there is `bytes[2] = 2;` in the following line, and then rust code will modify its buffer.

liigo: OK. I learned more about JS `const` (const != immutable in JS). Will change. Thank you!

belfz: Thanks :) I agree, the JS's `const` is somewhat confusing.

belfz:
Thank you @liigo!

PR #437, opened 2018-07-10
Add support for methods of Symbol

rail44:
support Symbol, except for `Symbol.prototype[@@toPrimitive]` and `Symbol.unscopables`


rail44:
and, i've forgotten to add comment for bindings :sweat_smile:

PR #438, opened 2018-07-10
bindings for Date.setXXX

toVersus:
I'm not sure my decision is correct, but I ignored the bindings for optional arguments because it seems unintuitive for the methods to take those arguments at all times.

Referencing issue #275.


alexcrichton:
Thanks! Looks like time zones (or something like that) may be coming into effect though? (failing tests)

toVersus:
Yeah, my lack of consideration...so, I will fix it!

fitzgen:
Thanks!

PR #440, opened 2018-07-10
Add console.log(error); [--no-modules]

limira:


crates/cli-support/src/js/mod.rs
```diff
@@ -426,12 +426,18 @@ impl<'a> Context<'a> {
...
-                                .then(response => response.arrayBuffer())
-                                .then(buffer => WebAssembly.instantiate(buffer, {{ './{module}': __exports }}))
-                                .then(({{instance}}) => {{
-                                    wasm = init.wasm = instance.exports;
-                                    return;
-                                }});
+                                .then(response => response.arrayBuffer(), reason => console.log('Error start fetching', wasm_path, ':', reason))
+                                .then(
+                                    buffer => WebAssembly.instantiate(buffer, {{ './{module}': __exports }}),
+                                    reason => console.log('Error on streaming', wasm_path, ':', reason)
```
fitzgen: I don't think it is really appropriate for libraries to log to stdout freely -- shouldn't the consumer of the API be chaining `.then`s that do error logging and all that?

limira: Oh, I see! I just don't know how to use a Promise before. This is the first time I see it! OK, I will close this now!

crates/cli-support/src/lib.rs
```diff
@@ -317,7 +317,7 @@ or you can update the binary with
     cargo install -f wasm-bindgen-cli
 
 if this warning fails to go away though and you're not sure what to do feel free
-to open an issue at https://github.com/alexcrichton/wasm-bindgen/issues!
+to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!
```
fitzgen: This would still be great to merge if you'd like to make a new PR :)

Issue #449, opened 2018-07-10
Add the fetch API to web-sys


fitzgen:
This is a tracking issue for adding support for the fetch API to the `web-sys` crate.

General info on the fetch API: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

How to add support for new Web APIs to `web-sys`: https://rustwasm.github.io/wasm-bindgen/web-sys/supporting-more-web-apis.html

## Interfaces

Here is a list of interfaces we need to support:

* [ ] `fetch()`
* [x] `Headers`
* [ ] `Request`
* [x] `Response`
* [ ] `Body`

## Example

* [ ] Add an example to `examples/` that shows off working with the fetch API

brisad:
While `Headers` does show up in the docs, it does not have any methods. Should it still be added like that (without tests), or should I check into why `wasm_bindgen_webidl` doesn't export the methods? I see a bunch of warnings of unsupported definitions when building.

PR #452, opened 2018-07-10
Add Response interface part of fetch API

brisad:
#449 

Only part of the properties and methods of `Response` were generated, but I made a test for some of the generated ones.


crates/web-sys/tests/all/response.rs
```diff
@@ -0,0 +1,37 @@
+use super::project;
+
+#[test]
+fn response() {
+    project()
+        .add_local_dependency("web-sys", env!("CARGO_MANIFEST_DIR"))
+        .headless(true)
```
fitzgen: We are going to end up repeating these three lines in every single test in this crate -- would you be interested in writing a follow up PR that defines a helper function to preconfigure these bits in `crates/web-sys/tests/all/main.rs` ?

brisad: Certainly! :)

PR #453, opened 2018-07-10
Binding for Array.prototype.map()

dorayakikun:
This is my first PR.
Could you review it?

Referencing issue #275.

Thank you.


fitzgen:
I think this is jsut intermittent infra issues causing the CI failure -- restarted that job and we will see what happens!

Thanks again @dorayakikun !

PR #454, opened 2018-07-10
Build documentation for master branch on gh-pages

alexcrichton:
Don't link it from the book as the book likely wants to use published crates.io
versions, but it should be available to browse if need be

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,8 +114,8 @@ matrix:
       env: JOB=check-nightly-cli
       script: cargo check --manifest-path crates/cli/Cargo.toml
 
-    # Build the guide.
-    - rust: stable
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -125,6 +125,9 @@ matrix:
         - cargo install-update -a
       script:
         - (cd guide && mdbook build)
+        - cargo doc -p wasm-bindgen --no-deps
+        - cargo doc -p web-sys --no-deps
+        - mv target/doc guide/book/api
```
fitzgen: So will the API docs be used be used?

alexcrichton: Not directly by the book, but I'd find them useful at least to browser a git version of the documentation to checkup on things like the `web-sys` crate and its APIs

PR #457, opened 2018-07-11
webidl: Turn the `[Throws]` extended attributes into `Result<T, JsValue>`

fitzgen:
This sets the `catch` flag on the emitted AST when an operation/attribute has the `[Throws]` extended attribute on it.

Additionally, constructors aren't annotated with `[Throws]` but can still throw exceptions, so we must conservatively assume *every* constructor can throw an error.

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -77,7 +77,7 @@ fn compile_ast(mut ast: backend::ast::Program) -> String {
     let mut defined = BTreeSet::from_iter(
         vec![
             "str", "char", "bool", "JsValue", "u8", "i8", "u16", "i16", "u32", "i32", "u64", "i64",
-            "usize", "isize", "f32", "f64",
+            "usize", "isize", "f32", "f64", "Result",
         ].into_iter()
             .map(|id| proc_macro2::Ident::new(id, proc_macro2::Span::call_site())),
     );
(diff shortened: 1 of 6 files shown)


crates/webidl/src/lib.rs
```diff
@@ -210,19 +210,38 @@ impl<'a> WebidlParse<&'a webidl::ast::NonPartialInterface> for webidl::ast::Exte
...
+            // always throw.
+            //
+            // From https://heycam.github.io/webidl/#Constructor (emphasis
+            // mine):
+            //
+            // > The prose definition of a constructor must either return an IDL
+            // > value of a type corresponding to the interface the
+            // > `[Constructor]` extended attribute appears on, **or throw an
+            // > exception**.
+            let throws = true;
```
alexcrichton: This seems a little unfortunate :(

I wonder though if we can add configuration options to the generation to handle this? I'd expect that the majority of webidl constructors don't actually throw (or we're not interested in what they throw). We could perhaps explicitly whitelist some constructors as throwing but all others would be infallible?

fitzgen: We could certainly add our own `[WbgNoThrow]` extended attribute if we get to a place where we want  to explicitly whitelist some constructors.

alexcrichton: Perhaps yeah, although I was thinking moreso the build script rather than modifying the webidl files themselves, but either way's fine!

ohanar: It would be nice if WebIDL had a `NoThrowsConstructor` or something like that.

I think for now, we should probably just leave this as is, and maybe come back to some whitelist thing -- maybe make an issue about it?

fitzgen: I'll file a follow up.

fitzgen: https://github.com/rustwasm/wasm-bindgen/issues/458

PR #459, opened 2018-07-11
Fix Language Statistics

Slowki:
Right now [linguist](https://github.com/github/linguist) is reporting mostly IDL:
linguist-before
And that bothers me for some reason, so I added a `.gitattributes` to ignore the WebIDL.
linguist-after


fitzgen:
Neat!

Issue #255, opened 2018-06-08
Support WebIDL partial interfaces


fitzgen:
https://heycam.github.io/webidl/#dfn-partial-interface

```webidl
interface SomeInterface {
  /* interface_members... */
};

partial interface SomeInterface {
  /* interface_members... */
};

```

This is essentially the same as

```webidl
mixin SomeInterfaceExtra {...}
SomeInterface includes SomeInterfaceExtra
```

Therefore I think we should block on support for mixins, and desugar this into basically the above.

Blockers:

* [ ] #246

ohanar:
FYI, I'm working on this (I don't believe it actually depends on #246, but I'm looking at that too).

alexcrichton:
I believe this was done in https://github.com/rustwasm/wasm-bindgen/pull/460 (yay!)

PR #460, opened 2018-07-12
webidl: add support for partial interfaces and mixins

### LINKED ISSUE #246: Support WebIDL interface mixins
https://heycam.github.io/webidl/#idl-interface-mixins
https://heycam.github.io/webidl/#using-mixins-and-partials

Example:

```webidl
interface Entry {
  readonly attribute unsigned short entryType;
  // ...
};

interface mixin Observable {
  void addEventListener(DOMString type,
                        EventListener listener,
                        boolean useCapture);
  // ...
};

Entry includes Observable;
```

Given that example, I think we would want to

* extend the `Entry`'s bindings with all the things inside the mixin
* turn the `Observable` mixin into a trait with the appropriate methods
* and turn `Interface includes Mixin` into a trait implementation that proxies `<Entry as Observable>::addEventListener` to `Entry::addEventListener`

This way you don't *have* to always import the `Observable` trait to get an `addEventListener` method, but you can still abstract over anything that is `Observable`.

This will require some plumbing through the AST and its `ToTokens` implementations.

ohanar:
Fixes #246 and #255.

This is a major change to how webidl is processed. This adds
a two phase process, where the first phase records the names of
various types and indexes the mixins (and might do more in the
future). The actual program building happens in the second phase.

As part of this, this also makes it so that interface objects
are passed by reference, rather than by value. The spec isn't
exactly clear on this, but Mozilla's C++ reflection suggestions
seem to indicate that they should be passed by reference (see
https://developer.mozilla.org/en-US/docs/Mozilla/WebIDL_bindings).


crates/webidl/src/lib.rs
```diff
@@ -96,60 +94,156 @@ trait WebidlParse<Ctx> {
...
+                    }
+                }
+            }
+            None => warn!("Tried to include missing mixin {}", self.includee),
+        }
+        Ok(())
+    }
+}
+
+impl<'a, 'b> WebidlParse<&'a FirstPass<'b>> for webidl::ast::Interface {
```
alexcrichton: I'd personally recommend avoiding the `WebidlParse` trait entirely as this looks like it may be getting a little too verbose, but otherwise perhaps the `FirstPass` parameter could be part of the trait signature to avoid the wordiness of lifetimes here?

crates/webidl/tests/all/simple.rs
```diff
@@ -44,14 +44,14 @@ fn method() {
                     let pi = Foo::new(3.14159).unwrap();
                     let e = Foo::new(2.71828).unwrap();
                     // TODO: figure out why the following doesn't fail
```
alexcrichton: FWIW this should be fixed now in upstream rust if you wanna do a drive-by cleanup as well

alexcrichton:
Seems like a great cleanup to me and a good way to structure it as well!

crates/webidl/src/lib.rs
```diff
@@ -96,60 +94,156 @@ trait WebidlParse<Ctx> {
     fn webidl_parse(&self, program: &mut backend::ast::Program, context: Ctx) -> Result<()>;
 }
 
-impl WebidlParse<()> for Vec<webidl::ast::Definition> {
+fn first_pass<'a>(definitions: &'a [webidl::ast::Definition]) -> FirstPass<'a> {
+    use webidl::ast::*;
+
+    let mut first_pass = FirstPass::default();
+    for def in definitions {
+        if let Definition::Interface(Interface::NonPartial(NonPartialInterface { name, .. })) = def
```
fitzgen: Nitpick: stylistically I'd prefer using a match with a `_` arm to a series of `if let`s.

crates/webidl/src/util.rs
```diff
@@ -173,192 +58,344 @@ fn result_ty(t: syn::Type) -> syn::Type {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum TypePosition {
+    Argument,
+    Return,
+}
 
-        raw_ident(&format!("__widl_f_{}_{}", rust_name, ns))
-    };
+#[derive(Default)]
+pub struct FirstPass<'a> {
```
fitzgen: I think a doc comment here explaining what the first pass over the WebIDL does, what information is collected, and for what use and why would be very valuable for future code readers and contributors.

crates/webidl/src/util.rs
```diff
@@ -173,192 +58,344 @@ fn result_ty(t: syn::Type) -> syn::Type {
...
+        name: &str,
+        ty: &webidl::ast::Type,
+        self_name: &str,
+        is_static: bool,
+        is_structural: bool,
+        catch: bool,
+    ) -> Option<backend::ast::ImportFunction> {
+        let ret = match self.webidl_ty_to_syn_ty(ty, TypePosition::Return) {
+            None => {
+                warn!("Attribute's type does not yet support reading: {:?}", ty);
```
fitzgen: Should this warning message s/reading/returning/ ?

fitzgen: I guess this is how it was before, but the message no longer makes sense to me, and I assume I originally wrote it...

crates/webidl/src/lib.rs
fitzgen: +1 for `FirstPass` as part of the trait signature, since pretty much every impl is using it.

ohanar:
I ended up moving out the new first pass into a new module, and restructuring the new code a bit to make it more extensible in the future. Let me know if you have any new feedback, I hope to finish up addressing your previous comments tomorrow.

PR #466, opened 2018-07-13
Remove debug sections by default

alexcrichton:
The changes on master Rust insert debug sections now (yay!) but this means that
wasm binaries by default pick up debug sections from the standard library, so
let's remove them by default in wasm-bindgen unless `--debug` is passed

### DIFF
--- a/crates/cli-support/Cargo.toml
+++ b/crates/cli-support/Cargo.toml
@@ -19,5 +19,5 @@ serde_derive = "1.0"
 serde_json = "1.0"
 tempfile = "3.0"
 wasm-bindgen-shared = { path = "../shared", version = '=0.2.11' }
-wasm-gc-api = "0.1"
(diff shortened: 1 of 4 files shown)


alexcrichton:
@fitzgen they're [dwarf indeed!](https://github.com/llvm-mirror/lld/commit/b5c6a5aca8d591cf1ead4e7f2947b4c21affb713)

fitzgen:
Are they generated even when we don't have `debug = true` in `Cargo.toml` etc? It seems strange to me that we should be removing this stuff.

alexcrichton:
Yes, the standard library comes with debug information and, like Linux, final binaries have debuginfo sections by default in release mode for this reason

fitzgen:
Ok. We probably want to separate the idea of debug info and debug assertions at some point in time, but happy to move forward with this for now.

PR #470, opened 2018-07-13
Support WebIDL constants

### LINKED ISSUE #259: Support WebIDL constants
https://heycam.github.io/webidl/#idl-constants

We can turn these into Rust constants. Just need to extend `backend::ast` to be able to represent them.

jrakow:
Fix #259

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -29,12 +29,12 @@ use std::path::Path;
 use backend::defined::{ImportedTypeDefinitions, RemoveUndefinedImports};
 use backend::util::{ident_ty, rust_ident, wrap_import_function};
 use failure::ResultExt;
-use heck::CamelCase;
+use heck::{CamelCase, ShoutySnakeCase};
 use quote::ToTokens;
 
 use util::{
-    create_basic_method, create_function, create_getter, create_setter, webidl_ty_to_syn_ty,
-    TypePosition,
+    create_basic_method, create_function, create_getter, create_setter, webidl_const_ty_to_syn_ty,
+    webidl_const_v_to_backend_const_v, webidl_ty_to_syn_ty, TypePosition,
 };
 
 /// Either `Ok(t)` or `Err(failure::Error)`.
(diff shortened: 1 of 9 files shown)


crates/webidl/tests/all/consts.rs
```diff
@@ -0,0 +1,190 @@
...
+            r#"
+                #![feature(proc_macro, wasm_custom_section, wasm_import_module)]
+                extern crate wasm_bindgen;
+                use wasm_bindgen::prelude::*;
+
+                pub mod foo;
+
+                #[wasm_bindgen]
+                pub fn test() {
+                    assert_eq!(foo::floats::F, 0.0);
```
ohanar: Maybe make this `0.0_f32` to insure that `F` isn't accidentally a `f64` .

crates/webidl/tests/all/consts.rs
```diff
@@ -0,0 +1,190 @@
...
+                    assert!(foo::floats::NEG_INF.is_infinite());
+                    assert!(foo::floats::NEG_INF.is_sign_negative());
+                    assert!(foo::floats::INF.is_infinite());
+                    assert!(foo::floats::INF.is_sign_positive());
+                    assert!(foo::floats::NAN.is_nan());
+
+                    assert_eq!(foo::doubles::D, 0.0);
```
ohanar: And `0.0_f64` here.

crates/webidl/tests/all/consts.rs
```diff
@@ -0,0 +1,190 @@
...
+                    const unsigned long umin = 0;
+                    const unsigned long umax = 4294967295;
+                };
+                interface LongLong {
+                    const long long imin = -9223372036854775808;
+                    const long long imax = 9223372036854775807;
+                    const unsigned long long umin = 0;
+                    // bug in webidl
+                    // https://github.com/sgodwincs/webidl-rs/issues/15
+                    //const unsigned long long umax = 18446744073709551615;
```
ohanar: This is unfortunate. Upon scanning the docs, it doesn't look like [weedle](https://crates.io/crates/weedle) has this bug (it seems to preserve the literal string). I've wondered if we should use weedle instead since its dependencies are much more aligned with our own.

crates/webidl/src/lib.rs
```diff
@@ -474,3 +474,28 @@ impl<'a> WebidlParse<()> for webidl::ast::Enum {
         Ok(())
     }
 }
+
+impl<'a> WebidlParse<&'a str> for webidl::ast::Const {
+    fn webidl_parse(
+        &self,
+        program: &mut backend::ast::Program,
+        interface_name: &'a str,
```
ohanar: Maybe `self_name` here to be more consistent with the other impls.

crates/webidl/tests/all/consts.rs
fitzgen: It would also let us skip lalrpop, which is an incredible crate, but a long build time to hoist on to downstream users :-/

fitzgen: Looks like it was already fixed, that was fast

PR #471, opened 2018-07-13
Improve documentation around `link_this_library`

alexcrichton:
I've started noticing this in non-LTO builds and initially tried to remove it. I
was unsuccessful but decided to better document my adventures to hopefully
improve future onlookers!

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -447,7 +447,13 @@ pub struct GlobalStack {
 }
 
 const GLOBAL_STACK_CAP: usize = 16;
-static mut GLOBAL_STACK: [u32; GLOBAL_STACK_CAP] = [0; GLOBAL_STACK_CAP];
+
+// Increase the alignment to 8 here because this can be used as a
+// BigUint64Array pointer base which requires alignment 8
+#[repr(align(8))]
+struct GlobalData([u32; GLOBAL_STACK_CAP]);
+
+static mut GLOBAL_STACK: GlobalData = GlobalData([0; GLOBAL_STACK_CAP]);
 
 impl GlobalStack {
     #[inline]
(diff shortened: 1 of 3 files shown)


src/convert.rs
```diff
@@ -447,7 +447,13 @@ pub struct GlobalStack {
 }
 
 const GLOBAL_STACK_CAP: usize = 16;
-static mut GLOBAL_STACK: [u32; GLOBAL_STACK_CAP] = [0; GLOBAL_STACK_CAP];
+
+// Increase the alignment to 8 here because this can be used as a
+// BigUint64Array pointer base which requires alignment 8
+#[repr(align(8))]
```
alexcrichton: Apparently the original change in this PR caused the `u64::works` test to fail due to the pointer being returned by `__wbindgen_global_argument_ptr` being aligned to 4 instead of 8. We use this pointer as the base of a `BigUint64Array`, however, so we needed a bigger alignment on global. AKA a preexisting bug!

Presumably this was uncovered because the function changes below added a static which tweaked alignment of fields in this crate, but just wanted to make sure this wasn't thought as being totally random!

src/lib.rs
```diff
@@ -687,5 +687,53 @@ pub mod __rt {
...
+    /// it) will be placed into the same object file as the two intrinsics
+    /// above. That means if this function is called and referenced we'll pull
+    /// in the object file and link the intrinsics.
+    ///
+    /// Note that this is an #[inline] function to remove the function call
+    /// overhead we inject in functions, but right now it's unclear how to do
+    /// this in a zero-cost fashion. The lowest cost seems to be generating a
+    /// store that can't be optimized away (to a global), which is listed below.
+    ///
+    /// Ideas for how to improve this are most welcome!
```
fitzgen: It seems like all this should eventually be replaced with `#[used]` on the functions in question. Is there a reason why that wouldn't work (once we allow `#[used]` on functions)?

alexcrichton: I believe so!

PR #475, opened 2018-07-14
Bindings for Aate.prototype.findIndex(), toLocaleString()

dorayakikun:
#275 I will add Array.prototype.findIndex(), toLocaleString()
I am not sure whether to define the "this_arg" argument in findIndex() :(

### DIFF
--- a/src/js.rs
+++ b/src/js.rs
@@ -172,6 +172,13 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn find(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool) -> JsValue;
 
+    /// The findIndex() method returns the index of the first element in the array that 
+    /// satisfies the provided testing function. Otherwise -1 is returned.
+    /// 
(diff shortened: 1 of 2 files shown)


src/js.rs
```diff
@@ -172,6 +172,13 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn find(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool) -> JsValue;
 
+    /// The findIndex() method returns the index of the first element in the array that 
+    /// satisfies the provided testing function. Otherwise -1 is returned.
+    /// 
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex
+    #[wasm_bindgen(method, js_name = findIndex)]
+    pub fn find_index(this: &Array, predicate: &mut FnMut(JsValue, u32, Array) -> bool, this_arg: &JsValue) -> u32;
```
fitzgen: I don't think we need `this_arg` because our closure can't access it. Want to do a follow up to remove it?

dorayakikun: Thank you for your Advice.
I'll try to remove it 😃

PR #476, opened 2018-07-14
split const integers into signed and unsigned

jrakow:
Integrate the changes in sgodwincs/webidl-rs#15


jrakow:
Conflicts with #460 which contains changes to the original `const` PR.
Ignore this until #460 is merged. I will then rebase or something.

PR #477, opened 2018-07-14
WebIDL: Handle Invalid Enum Returns

### LINKED ISSUE #260: Support WebIDL enumerations

Slowki:
As discussed in #260 and #433 there are two main ways to handle invalid enum returns from JS in the WebIDL codegen:

### Method 1 ###
Convert enum returns in WebIDL to `Result<ENUM, String>`s

#### Advantages ####
* Fairly ergonomic and idiomatic
* You can add special handling logic for browser specific returns by examining the `String` in `Err`

#### Disadvantages ####
* Additive changes will break any `match`s that don't have a wildcard arm(unlike in Method 2, which forces a wildcard match), so the user might have to change their code when they update their `web-sys` version

### Method 2 ###
Non-exhaustive enums.

#### Advantages ####
* Any additive changes to the WebIDL enums(and as such the `web-sys` crate's Rust enum variants) don't break code because non-exhaustive enums force wildcard matches

#### Disadvantages ####
* Not very ergonomic on the end user since their code could end up with a lot of wildcard matches even in cases where the enum is coming from their own Rust code 
* Wildcard matches wouldn't give you specific information on the value returned(unless we want to make the enums un-copyable), just that it wasn't one of the known variants, so you couldn't add special handling for browser specific return values

I opted for Method 1 because it seems like it gives the cleanest code when using enums, and I like the idea of being able to handle browser specific returns that aren't document in WebIDL. Do you folks agree, or should we go with Method 2/something else?

Closes #260

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -616,7 +616,7 @@ impl ToTokens for ast::ImportEnum {
             }
 
             impl #name {
-                #vis fn from_js_value(obj: ::wasm_bindgen::JsValue) -> Option<#name> {
+                #vis fn from_js_value(obj: &::wasm_bindgen::JsValue) -> Option<#name> {
                     obj.as_string().and_then(|obj_str| match obj_str.as_str() {
                         #(#variant_strings => Some(#variant_paths_ref),)*
                         _ => None,
@@ -647,7 +647,7 @@ impl ToTokens for ast::ImportEnum {
                     js: Self::Abi,
                     extra: &mut ::wasm_bindgen::convert::Stack,
                 ) -> Self {
-                    #name::from_js_value(::wasm_bindgen::JsValue::from_abi(js, extra)).expect(#expect_string)
+                    #name::from_js_value(&::wasm_bindgen::JsValue::from_abi(js, extra)).expect(#expect_string)
                 }
             }
 
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -583,6 +583,7 @@ impl ToTokens for ast::ImportEnum {
         let expect_string = format!("attempted to convert invalid JSValue into {}", name);
         let variants = &self.variants;
         let variant_strings = &self.variant_values;
+        let attrs = &self.rust_attrs;
 
         let mut current_idx: usize = 0;
         let variant_indexes: Vec<Literal> = variants
@@ -609,7 +610,7 @@ impl ToTokens for ast::ImportEnum {
 
         (quote! {
             #[allow(bad_style)]
-            #[derive(Copy, Clone, Debug)]
+            #(#attrs)*
             #vis enum #name {
                 #(#variants = #variant_indexes_ref,)*
             }
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -580,7 +580,7 @@ impl ToTokens for ast::ImportEnum {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         let vis = &self.vis;
         let name = &self.name;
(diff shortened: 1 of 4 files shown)


crates/backend/src/codegen.rs
```diff
@@ -624,17 +624,26 @@ impl ToTokens for ast::ImportEnum {
...
-                        _ => None,
-                    })
+                /// Attempt to convert a `JSValue` to an enum, returning Ok if
+                /// the value maps to the enum's variants, or an `Err` containing
+                /// the `JsValue`'s string value if it doesn't map to any of the
+                /// variants.
+                /// 
+                /// # Panics
+                /// This function will panic if the `JsValue` isn't a string
+                #vis fn from_js_value(obj: ::wasm_bindgen::JsValue) -> Result<#name, String> {
```
alexcrichton: Since this doesn't actually need ownership, should this take `&JsValue`?

crates/backend/src/codegen.rs
```diff
@@ -624,17 +624,26 @@ impl ToTokens for ast::ImportEnum {
 
         (quote! {
             #[allow(bad_style)]
-            #[derive(Copy, Clone, Debug)]
+            #[derive(Copy, Clone, PartialEq, Debug)]
```
alexcrichton: I've just realized (after the fact, sorry!) that in other places in wasm-bindgen we don't explicitly list derive modes but rather pass through the original annotations. Here there's no `#[wasm_bindgen]` equivalent for generating an imported `enum` so it doesn't matter too too much, but long-term we may want this line to be `#(#rust_attrs)*` and store attributes in the `ImportEnum`

src/convert.rs
```diff
@@ -46,6 +46,27 @@ pub trait FromWasmAbi: WasmDescribe {
     unsafe fn from_abi(js: Self::Abi, extra: &mut Stack) -> Self;
 }
 
+/// A trait for anything that can be recovered by-value from the wasm ABI
+/// but whose conversion might error, for example a JS string being converted
+/// to a Rust enum.
+pub trait TryFromWasmAbi: WasmDescribe + Sized {
```
alexcrichton: I'm a little wary of adding a trait like this, I'm afraid it'll be difficult to work with in general. The conversion traits are used in various locations throughout imports/exports/closures/etc. This also necessitates `FromWasmAbi` for `Result` which seems unfortunate in the sense that you can't do:

```rust
#[wasm_bindgen]
pub fn foo(a: CustomEnumType) {
    // ...
}
```

but are instead forced to do:

```rust
#[wasm_bindgen]
pub fn foo(a: Result<CustomEnumType, String>) {
    // ...
}
```

I might be leaning a bit towards dealing with nonexhaustive enums due to this interaction of how to add implicit conversions? I would also be ok, though, having a fallible constructor in Rust but panicking in the conversions at the boundary (as it seems like that's what may happen most of the time anyway?)

Slowki: Non-exhaustive enums are fine by me, the only awkward thing with them is the potential panic if the Rust code gets an invalid enum then tries to return it back to JS, which I can't think of a decent way around.

crates/backend/src/codegen.rs
Slowki: :+1:

alexcrichton:
Awesome, thanks!

PR #478, opened 2018-07-14
Clean up Some Clippy Warnings

Slowki:
This PR cleans up some Clippy warnings

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -216,7 +216,7 @@ impl Function {
 
 impl Export {
     pub(crate) fn rust_symbol(&self) -> Ident {
-        let mut generated_name = format!("__wasm_bindgen_generated");
+        let mut generated_name = String::from("__wasm_bindgen_generated");
         if let Some(class) = &self.class {
             generated_name.push_str("_");
             generated_name.push_str(&class.to_string());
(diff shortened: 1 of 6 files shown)


crates/backend/src/ast.rs
```diff
@@ -39,7 +39,7 @@ pub struct Import {
 
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
 pub enum ImportKind {
-    Function(ImportFunction),
+    Function(Box<ImportFunction>),
```
alexcrichton: We aren't really going for speed here, could this be left as is to keep it consistent ?

crates/backend/src/codegen.rs
```diff
@@ -31,19 +31,19 @@ fn to_ident_name(s: &str) -> Cow<str> {
 impl ToTokens for ast::Program {
     // Generate wrappers for all the items that we've found
     fn to_tokens(&self, tokens: &mut TokenStream) {
-        for export in self.exports.iter() {
```
alexcrichton: I personally prefer the previous way this was written, could it stay as it was?

alexcrichton:
Thanks!

crates/backend/src/ast.rs
Slowki: :+1:

PR #479, opened 2018-07-14
Remove target_args from Array.find_index()

dorayakikun:
Because closure can't access this_arg.
Referencing issue #475 .


alexcrichton:
Thanks!

PR #493, opened 2018-07-17
`String.prototype.padEnd()`, `String.prototype.padStart()`

data-pup:
This adds bindings for the `String.prototype.padEnd()` and `String.prototype.padStart()` methods :)

(Regarding #275)


data-pup:
Took care of the conflict :)

PR #486, opened 2018-07-17
Bindings for TypedArray subclasses

cdisselkoen:
More progress towards #275.  This exposes the bindings for constructors and `fill` methods for `Uint16Array`, `Uint32Array`, `Uint8ClampedArray`, `Int8Array`, `Int16Array`, `Int32Array`, `Float32Array`, and `Float64Array`, following very closely the already-existing support for `Uint8Array` supplied in #316.  Small tweaks to the existing `Uint8Array` code (moved to alphabetical position in `js.rs`, refactored tests) are included in this PR.

Feedback welcome - this is my first contribution to this project, and actually to any project written in the Rust language.

PR #496, opened 2018-07-17
Fix symbol conflicts defined in two crates

alexcrichton:
Not a lot of attention has been paid to dealing with conflicts of symbols
between crates and different `#[wasm_bindgen]` blocks. This commit starts to fix
this issue by unblocking #486 which first ran into this. Currently there's a bug
where if two independent crates bind the same JS API they'll generate the same
symbol which causes conflicts for things like LTO or linking in general.

This commit starts to add a "salt" to all symbols generated by `wasm-bindgen`
(these are all transparent to the user) to ensure that each crate's invocations
are kept apart from one another and using the correct bindings.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -1,6 +1,4 @@
-use std::borrow::Cow;
 use std::collections::HashSet;
-use std::env;
 use std::sync::atomic::{AtomicUsize, Ordering, ATOMIC_USIZE_INIT};
 
 use ast;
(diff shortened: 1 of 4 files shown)


examples/julia_set/build.sh
```diff
@@ -4,7 +4,7 @@
 
 set -ex
 
-cargo +nightly build --target wasm32-unknown-unknown --release
+cargo +nightly build --target wasm32-unknown-unknown
```
fitzgen: Why remove `--release` here?

alexcrichton: Oops! That was a mistake

PR #500, opened 2018-07-18
Add binding for String.prototype.repeat()

toVersus:
#275


fitzgen:
It looks like #499 already added `String.prototype.repeat` -- sorry for the confusion!

toVersus:
Oops, I didn't notice the duplication! I'm trying to bind another one:thumbsup:

PR #506, opened 2018-07-18
[WIP] webidl: use weedle instead of webidl-rs

ohanar:
This is an experiment of using weedle instead of webidl-rs. Some pros and cons:

**Pros:**
- Uses nom rather than lalrpop which is a heavy dependency put on downstream users.
- Separates out types a bit more, which is nice for converting to syn types.

**Cons:**
- Error messages seem basically non-existent.
- Currently has a bug that appears to have truncated output (causing web-sys to fail). I haven't investigated this deeply.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -184,14 +184,28 @@ pub struct Const {
     pub value: ConstValue,
 }
 
-#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq))]
-/// same as webidl::ast::ConstValue
+#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
 pub enum ConstValue {
-    BooleanLiteral(bool),
-    FloatLiteral(f64),
-    SignedIntegerLiteral(i64),
-    UnsignedIntegerLiteral(u64),
+    Boolean(bool),
+    Float(FloatValue),
+    Integer(IntegerValue),
     Null,
+}
+
+#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
+pub enum FloatValue {
+    Literal(String),
+    Value(f64),
+    NegInfinity,
+    Infinity,
+    NaN,
+}
+
+#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
+pub enum IntegerValue {
+    Literal(String),
+    SignedValue(i64),
+    UnsignedValue(u64),
 }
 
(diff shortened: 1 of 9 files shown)


fitzgen:
I think this is the right direction long term, thanks for digging in @ohanar!

Here are some things off the top of my head that I think we should do before merging this PR:

* [ ] `webidl` has been around longer than `weedle` has and been used more, therefore it has likely explored more edge cases and fixed more edge case bugs than `weedle` has. Hopefully those are captured in the `webidle` test suite. We should port all of the `webidl` tests to `weedle`.

* [ ] We need OK error messages for *our* debugging when adding new webidl frontend features (but we don't need *good* error messages in general and our users don't need them at all by default since we are only processing known-good webidls in `web-sys` and basically no one else will use the webidl frontend).

* [ ] it would be great to fuzz `weedle`, but I view this as a nice-to-have since we aren't really working with untrusted input.

ohanar:
I've been pretty busy recently, but I got a chance to look into this more thoroughly in the past week. My general observations:
- Currently weedle never errors. This can be easily fixed as it is a small oversight, but the best context you could give at the moment without more thorough changes is pointing to the definition block that failed.
- Weedle parses the subset of the WebIDL grammar that is used by the WebIDL spec. The main place that I've found that these differ is for extended attributes, where the grammar is far more general that what the spec needs. mozilla-central's webidl files use this more general grammar with `attribute="some string literal"` all over the place. From a cursory glace through webidl-rs' source, it doesn't seem like it supports the full general grammar of extended attributes either (although some commented out code looks like there was the intention at some point), however, it seems to parse the main offender I mentioned above just fine.
- The truncating issue that I experienced with web-sys was just the combination of the two previous bullet points.
- I fuzzed on my personal machine weedle for the better part of a week and had no issues.

Overall, I think we probably would need better error reporting before using it, and doing that would require some fairly serious changes to the library. @csharad offered to transfer ownership to us if we decided to use it since he didn't think he would be working on it in the near future.

Given all the issues at the moment, I think we should hold off on this for the moment, as there are more pressing priorities. @fitzgen Would you rather I leave this PR open as a placeholder, or should I make a new issue and close this.

fitzgen:
> @fitzgen Would you rather I leave this PR open as a placeholder, or should I make a new issue and close this.

If the intention is not to merge this PR but to do more fundamental work and then make a new PR, then we should close this.

> I fuzzed on my personal machine weedle for the better part of a week and had no issues.

Nice!

ohanar:
I'm going to close this then.

PR #507, opened 2018-07-18
Add support for optional slice types

alexcrichton:
This commit adds support for optional slice types to wasm-bindgen, allowing these types to be used for exported and imported APIs. They currently map to `undefined` in JS for `None` in Rust, and `Some` in Rust for other values in JS.


alexcrichton:
After this I hope to change the macro generation for imported types to also generate Into/From wasm abi implementations for `Option<Foo>` to allow optional custom types as well.

alexcrichton:
I've updated to tweak this a bit, it now interprets *both* `null` and `undefined` as mapping to `None`. The motivation for this that I've arbitrarily chosen [`Headers.get`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get) to fill out support for. In [WebIDL this returns an optional `ByteString`](https://github.com/rustwasm/wasm-bindgen/blob/e99c6b7c61f7ed69b25769d856c7e06b506c1921/crates/web-sys/webidls/enabled/Headers.webidl#L26) but the actual behavior is that it returns `null`. 

This means that when you pass `None` to JS it'll always generate `undefined` but when JS passes Rust either `null` or `undefined` it'll map to `None`. Hopefully this behavior will work for more APIs!

alexcrichton:
Urgh actually so after implementing more of this I don't believe this strategy will work. `impl ForeignTrait for Option<LocalType>` is denied by the orphan rules, and I didn't realize that! That means we'll need a different strategy here for foreign types defined in other crates (like web-sys).

alexcrichton:
Ok! I've taken a different approach now for how the underlying infrastructure works and I've also implemented automatic implementations for `#[wasm_bindgen]` imported types, so should be good to go now!

src/convert/traits.rs
```diff
@@ -1,6 +1,3 @@
-//! This is mostly an internal module, no stability guarantees are provided. Use
-//! at your own risk.
```
fitzgen: Are we going to make this stuff part of the stable interface now?

alexcrichton: Nah I hope to still keep this as an internal implementation detail (these docs just moved to `src/convert/mod.rs`). I'm not sure how long we'll be able to pull this off but I figured we may as well try!

PR #512, opened 2018-07-19
Move the `js` module to a `js_sys` crate

alexcrichton:
This should allow us to independently version the `js_sys` crate from `wasm_bindgen` itself, while also making `wasm_bindgen` a bit lighter!


fitzgen:
This *also* sped up our CI times a bunch! :tada:

PR #518, opened 2018-07-19
Explicitly ignore "Exposed" attributes

alexcrichton:
It looks like these are primarily targeted at informing whether functionality is
either on web workers, windows, or both. For now we'll generate the same
bindings regardless, and users will need to be proactive about what they're
using. In that case there shouldn't be any need for us to process these, so
avoid warning about them!

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -353,6 +353,14 @@ impl<'a> WebidlParse<&'a webidl::ast::NonPartialInterface> for webidl::ast::Exte
             {
                 add_constructor(arguments, &interface.name)
             }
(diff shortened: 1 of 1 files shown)


fitzgen:
We had previously talked about using features for each exposed thing -- do you think it is better to ask users to be proactive in doing feature detection?

fitzgen:
(woops, wrong comment button)

alexcrichton:
Hm an interesting idea! It's probably more realistic to divide this crate into web-worker and not than it is for libc to be divided into "pre-this-version-of-linux" and not. In that sense we most likely could proactively use these for those sorts of features!

PR #521, opened 2018-07-20
Add some notes about viewing bindings output.

derekdreery:
I found this helpful when debugging, so I thought others might too.


derekdreery:
ci failure is spurious.

alexcrichton:
👍

PR #523, opened 2018-07-20
add bindings for String.to_locale_XXX_case

toVersus:
I ignored tricky behavior of locale parameter passing as Array implicitly.

Ref. #275


alexcrichton:
Thanks!

Issue #517, opened 2018-07-19
Travis regularly times out on "npm ci"


alexcrichton:
I'm not really sure why but Travis will regularly time out with `npm ci` not emitting output for 10 minutes. This seems to happen sporadically but fairly regularly. Some example logs are:

* https://travis-ci.org/rustwasm/wasm-bindgen/jobs/405978557
* https://travis-ci.org/rustwasm/wasm-bindgen/jobs/405978581
* https://travis-ci.org/rustwasm/wasm-bindgen/jobs/405964606
* https://travis-ci.org/rustwasm/wasm-bindgen/jobs/405869282
* https://travis-ci.org/rustwasm/wasm-bindgen/jobs/405837840
* https://travis-ci.org/rustwasm/wasm-bindgen/jobs/405837842

Does anyone know why this is happening? Is there a way we can help debug this?

Alexendoo:
You can use `npm ci --verbose` to see what operation it is hanging on

alexcrichton:
Hm well this hasn't happened in awhile that I've seen, so maybe the verbose output actually helps us! In any case we should be set next time it times out to investigate further, thanks @Alexendoo!

alexcrichton:
Looks like we hit this first on https://travis-ci.org/rustwasm/wasm-bindgen/jobs/409944853

PR #527, opened 2018-07-20
Verbose npm ci output

Alexendoo:
Temporarily increase the output of npm ci to see what's happening regarding #517


Alexendoo:
Hopefully hits the error on the first run, if it doesn't there is a re-run button on travis rather than having to merge it

alexcrichton:
Bah well no worries, will be good to have anyway. Thanks for this!

PR #529, opened 2018-07-21
Move the macro support code into its own crate

konstin:
This is partially reversing #411. The reason for this change is that is otherwise impossible to use wasm-bindgen-backend in others crates without copying parser.rs. It also makes the `cfg_attr(feature = "extra-traits", ...)` work again as the wasm-bindgen-macro crate doesn't have that feature.


alexcrichton:
Thanks for the PR! I'm ok doing this, but FWIW the backend and other crates of wasm-bindgen are intended to be purely internal and have API breaking changes without being signaled through semver, so I wouldn't recommend using them from other projects. I'm curious, what's your use case for using the crates externally?

ohanar:
I would personally vote against this, the backend crate really should just be for codegen. Would making the parser module public in the macro crate be sufficient for your usage? (and adding an "extra-traits" feature to the macro crate as well).

konstin:
I'm trying to unify helix, pyo3 and wasm-bindgen under a single proc macro, which already works for classes and methods: https://github.com/konstin/capybara

I'm aware that the backend isn't a real public api with semver garuantess; I've already migrated the code with the last two or three minor versions.

> Would making the parser module public in the macro crate be sufficient for your usage?

That would be the ideal solution, but afaik proc macro lib crates still can't export any other items.

alexcrichton:
I wonder, could the macro be split into a `macro` crate that purely calls one function in a `macro-support` crate? That way other libraries could link to `macro-support` in an experimental fashion

konstin:
IMHO the backend crate is already close enough to what `macro-support` would do. The problem is that syn upgrades break downstream crates anyway. For example the only real effort in updatig 0.2.10 to 0.2.11 was porting capybara from syn 0.13 to syn 0.14.

ohanar:
The backend crate also serves as a backend for other frontends (at the moment mainly WebIDL, although I think there is also the intention to support Typescript as well at some point in the future).

The main reason that I moved stuff out of the backend crate was to make sure that codegen wasn't implicitly depending on stuff specific to the rust macro frontend (which it was at the time). I think it would be less of an issue if all the macro frontend stuff was in a separate module (which is what this PR does), but I still don't think it really belongs in the backend crate. Just my $0.02 :smile:.

konstin:
As long as the macro-frontend-api stays public, I'm fine with every way of structuring this. I wouldn't mind changing this PR to adding an extra crate for the macro frontend stuff.

konstin:
I've moved the code into a new `macro-support` crate, that exports `BindgenAttrs` and an `expand` function. Should be ready for another review now.

alexcrichton:
Thanks!

PR #531, opened 2018-07-21
Add documentation for passing arbitrary data

David-OConnor:
Bindgen can indirectly pass data structs more varied and complex than those directly supported via serialization. This pull adds a page describing how to do so.


alexcrichton:
Thanks so much!

PR #538, opened 2018-07-23
Add Math.hypot binding

twilco:
Adds `Math.hypot()` binding for #275.


alexcrichton:
Awesome, thanks!

crates/js-sys/src/lib.rs
```diff
@@ -1154,6 +1154,12 @@ extern "C" {
     #[wasm_bindgen(static_method_of = Math)]
     pub fn fround(x: f64) -> f32;
 
+    /// The Math.hypot() function returns the square root of the sum of squares of its arguments.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/fround
```
jontro: Shouldn't this be https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/hypot ?

alexcrichton: Ah yes indeed! @jontro want to file a follow-up PR?

jontro: Done! https://github.com/rustwasm/wasm-bindgen/pull/567

Issue #514, opened 2018-07-19
Add iterable support for webidl


jonathanKingston:
In web-sys to get a functional DOM implementation `iterable<Node?>` is needed by `NodeList`.

https://heycam.github.io/webidl/#dfn-iterable-declaration which is supported by `webidl::ast::Iterable`.

`pub fn get_elements_by_name(&self, element_name: &str) -> NodeList`

jonathanKingston:
To add `for_each` support we will need to support https://github.com/rustwasm/wasm-bindgen/issues/257

@fitzgen part of me things `NodeList` and others should just implement [Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html). I know this isn't the correct DOM API however the interface describes some of the same methods.
I don't know enough about the abstraction across the buffer but perhaps the nodes have already passed into wasm when code has called `get_elements_by_name` that it could be represented as a vec or similar in wasm?

The alternative is implement all the functions that iterables support in webidl like:
https://heycam.github.io/webidl/#es-forEach and https://heycam.github.io/webidl/#es-iterable-values

Oddly with the latest `Option` support I can't seem to provide a non throwing function return type. Is only `Result<Option<T>, E>` implemented for function responses and not: `Option<T>`?

I managed to get the code spitting out a `values` with a response, this should be implementing `@@iterator` perhaps it could just return a vector of the type?

Does this need a RFC?

alexcrichton:
Interesting! I think I'd personally be on board with the strategy of using `Iterator` in Rust and just defining it for iterable types where it makes sense. For example could `get_elements_by_name` return something like `impl Iterator`?

Returns from functions should support `Option<T>`, yeah, but do you have an example of a function that didn't work for you?

For now I think it's ok to avoid an RFC, we're still in the somewhat early stages of web-sys!

Pauan:
I don't think `get_elements_by_name` should return `impl Iterator`, because that would prevent using useful methods (like `length`, or indexing).

Instead I think the best course of action is to return `NodeList` (and then implement `Iterator` for `NodeList`).

jonathanKingston:
> Returns from functions should support Option<T>, yeah, but do you have an example of a function that didn't work for you?

```
let return_value = webidl::ast::ReturnType::NonVoid(self.value_type.clone());
let args = [];
first_pass
.create_basic_method(
    &args,
    Some(&"item".to_string()),
    &return_value,
    self_name,
    false,
    false
)
[...]

PR #541, opened 2018-07-24
Initial support for Document, EventTarget, NodeList and Iterator 

jonathanKingston:
@alexcrichton https://github.com/rustwasm/wasm-bindgen/compare/master...jonathanKingston:web-sys/document-support?expand=1#diff-9f9955b41f2425125eaf42c53a05c94dR558 is the mentioned issue from https://github.com/rustwasm/wasm-bindgen/issues/514#issuecomment-407090466

The other question I had was surrounding: `element.query_selector_all('x').get(0)` returning a `Node` not an `Element`.  Is this something https://github.com/rustwasm/rfcs/pull/2 will tackle?


alexcrichton:
Awesome, thanks for this!

I've switched the `throws` there to `false` and I think I see the issue! The first problem is that `Option` wasn't defined in [this list](https://github.com/rustwasm/wasm-bindgen/blob/cefe1681ab37091ca07080fe1234390e5b5aff47/crates/webidl/src/lib.rs#L83-L86). After that though it looked like the path traversal code wasn't handling generics so there were lots of compilations errors. After [filling out the missing cases](https://gist.github.com/alexcrichton/7a74a6781ad7bea5e9974f3ec3aed1f3) though it should be all good to go!

crates/webidl/src/lib.rs
```diff
@@ -520,6 +522,48 @@ impl<'a> WebidlParse<&'a str> for webidl::ast::RegularAttribute {
...
+            return Ok(());
+        }
+
+        let throws = util::throws(&self.extended_attributes);
+        let return_value = webidl::ast::ReturnType::NonVoid(self.value_type.clone());
+        let args = [];
+        first_pass
+            .create_basic_method(
+                &args,
+                Some(&"values".to_string()),
```
alexcrichton: I wasn't quite sure what these two methods were doing, is this what it means to be iterable? (a `values` and an `item` method?)

jonathanKingston: It comes from the `iterable` in webidl yeah: https://heycam.github.io/webidl/#es-iterable

I couldn't get `forEach` to work as we don't have callbacks yet.

`entries`, `keys` and `values` and some other behaviour is needed.

I messed up though `item` is a getter on the webidl which isn't supported yet.

jonathanKingston:
I filled in the `Parenthesized` `PathArguments` and it looks like it is working (I have more in Element now like `getAttribute` which wasn't working before).

I commented out the code for now as I think I need to implement a trait for iterator similar to how `OptionFromWasmAbi` and `OptionIntoWasmAbi` work.

I think perhaps I could remove the function comments and it should be merged as is (without iterator).

crates/webidl/src/lib.rs
jonathanKingston: Thanks for this direction, it really helped me understand the codebase a lot more too!

crates/backend/src/defined.rs
```diff
@@ -147,19 +161,74 @@ impl ImportedTypes for syn::TypePath {
...
+                for arg in data.args.iter() {
+                    arg.imported_types(f);
+                }
+            }
+//TOCHECK
+            syn::PathArguments::Parenthesized(data) => {
+                for input in data.inputs.iter() {
+                    input.imported_types(f);
+                }
+                // TODO do we need to handle output here?
```
alexcrichton: I think so yeah if you get a chance, makes sense to go ahead and add it here!

alexcrichton:
Awesome looks great to me!

Do you think we'll need a new trait for the iterator types? I'm actually curious if we can get by with an `Iterator` trait implementation? Or perhaps things like `values` return `impl Iterator`?

jonathanKingston:
Wow cool thanks for merging!

> Do you think we'll need a new trait for the iterator types? I'm actually curious if we can get by with an Iterator trait implementation? Or perhaps things like values return impl Iterator?

More experimenting needed I think, I am hoping we can use that but I'm still learning here.

alexcrichton:
Sure thing! I'm all for incremental changes anyway :)

PR #542, opened 2018-07-24
Implement Math.min() and Math.max() bindings

twilco:
`Math.min()` and `Math.max()` for #275.

And while I have your attention... :)

How do we want to do constants?  Doesn't look like anyone has made a constant for any binding yet.  Tried this, but figured it was too good to be true.

```
// Math
#[wasm_bindgen]
extern "C" {
    pub type Math;

    const E: f64 = 2.718281828459045;

    /// The Math.abs() function returns the absolute value of a number, that is
    /// Math.abs(x) = |x|
    ///
    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs
    #[wasm_bindgen(static_method_of = Math)]
    pub fn abs(x: f64) -> f64;
```

Error message is: Static constants in extern blocks cannot have values

`let` also does not work.


alexcrichton:
Awesome, thanks!

For constants... that's a good question! I think though that something like this should in theory work (although it may need some backend work)

```rust
#[wasm_bindgen]
extern {
    #[wasm_bindgen(static_method_of = Math)]
    static E: f64;
}
```

(or something like that)

As I write this I'm quite certain this'll need some backend work! For now it's ok to skip those :)

PR #547, opened 2018-07-25
Extra docs

derekdreery:
I always try to understand a codebase by writing docs, here they are.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -2,22 +2,39 @@ use proc_macro2::{Ident, Span};
 use shared;
 use syn;
 
+/// An abstract syntax tree representing a rust program. Contains
+/// extra information for joining up this rust code with javascript.
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq))]
 #[derive(Default)]
 pub struct Program {
+    /// rust -> js interfaces
     pub exports: Vec<Export>,
+    /// js -> rust interfaces
     pub imports: Vec<Import>,
+    /// rust enums
     pub enums: Vec<Enum>,
+    /// rust structs
     pub structs: Vec<Struct>,
+    /// rust type aliases
     pub type_aliases: Vec<TypeAlias>,
+    /// rust consts
     pub consts: Vec<Const>,
 }
 
+/// A rust to js interface. Allows interaction with rust objects/functions
+/// from javascript.
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
 pub struct Export {
+    /// The javascript class name.
     pub class: Option<Ident>,
+    /// The type of `self` (either `self`, `&self`, or `&mut self`)
     pub method_self: Option<MethodSelf>,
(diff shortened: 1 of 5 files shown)


crates/backend/src/ast.rs
```diff
@@ -36,7 +36,9 @@ pub struct Export {
     pub constructor: Option<String>,
     /// The rust function
     pub function: Function,
+    ///
```
fitzgen: We should either not have a doc comment at all, or the comment should have some content.

crates/backend/src/ast.rs
```diff
@@ -36,7 +36,9 @@ pub struct Export {
...
     pub comments: Vec<String>,
+    /// The name of the rust object the function belongs to TODO is this correct?
```
fitzgen: This is the name of the rust function/method, not the object. Check out `impl ToTokens for ast::Export` to see how it is used.

crates/macro/src/parser.rs
```diff
@@ -44,6 +48,7 @@ impl BindgenAttrs {
             .next()
     }
 
+    /// Get the first module attribute
```
fitzgen: s/module/version/

crates/backend/src/ast.rs
```diff
@@ -255,6 +275,7 @@ impl Export {
     }
 }
 
+
```
fitzgen: Nitpick: remove this newly introduced empty line.

crates/backend/src/ast.rs
```diff
@@ -230,6 +249,7 @@ impl Export {
         Ident::new(&generated_name, Span::call_site())
     }
 
+    /// ???
```
fitzgen: This is the name of the shim function that gets exported and takes the raw ABI form of its arguments and converts them back into their normal, "high level" form before calling the actual function.

crates/macro/src/parser.rs
derekdreery: oops

derekdreery:
Ready for review again.

crates/webidl/src/first_pass.rs
```diff
@@ -1,3 +1,12 @@
+//! Because some WebIDL constructs are defined in multiple places
+//! (keyword `partial` is used to add to an existing construct),
+//! We need to first walk the webidl to collect all non-partial
+//! constructs so that we have containers in which to put the
+//! partial ones.
+//!
+//! Only `interface`s, `dictionary`s, `enum`s and `mixin`s can
+//! be partial.
+
```
ohanar: A couple of more things that have been on my mind and should go in the first pass in the future (likely not exhaustive):
- `typedef`s should not be exposed on the rust side, but should be processed by the binding's generator (i.e. this crate). To do that, `typedef` declarations have to be recorded before building the AST, so in the first pass.
- should we want to implement up-casting conversion by more than one level, we need to be able to walk the inheritance tree up to the root while building the AST, doing so would require recording that in the first pass

More generally, the fist pass is used to record information that needs to be globally accessible while building the backend AST.

PR #559, opened 2018-07-26
Initial support for in-browser and headless testing

alexcrichton:
This commit provides what I hope is the initial support for headless and in-browser testing. Most support has gone into two separate locations:

* In `crates/test` the test harness was souped up to handle in-browser tests as well as node.js tests. It detects at runtime where it's running and renders output appropriately.
* In `wasm-bindgen-test-runner` the there are now two new modules. One is a server based on [rouille](https://github.com/tomaka/rouille) which serves as an HTTP server for running tests against. Another is a headless testing module which implements the WebDriver protocol to run a browser against this headless HTTP server.

The goal here is to continue to migrate all of wasm-bindgen's test away from the test project builder and towards a "hopefully idiomatic" solution that can be adopted by others in the community.

I'm curious to hear thoughts on the implementation and strategy here though!


alexcrichton:
One thing I'd like to get thoughts on as well, the `crates/js-sys/tests/wasm/main.rs` tests is quite the large test suite now, and I've gotten it working with browsers but it breaks compatibility with node (unless we want to use node's experimental modules). I'm curious if others think we should prioritize running tests in the browser? Or in node?

fitzgen:
Ideally we would run in both node and browsers -- what's the issue? ES6 modules? I think we can use node's experimental modules flag in our tests, that sounds fine to me.

fitzgen:
Tests aren't passing -- should I still review this or wait?

alexcrichton:
Nah I'll rebase over the recent changes, get tests passing, and add comments. I'll ping you after all that though and it'll be in a better state to review!

alexcrichton:
Ok should be good to go! Tests may be failing but I'll mostly be wrangling CI, no major changes should be needed to get tests working.

alexcrichton:
A green bill of health!

crates/test/src/rt/detect.rs
```diff
@@ -0,0 +1,62 @@
...
+    // values for `this` when using the `apply` function. Add it all up, and you
+    // get the below code:
+    //
+    // * Manufacture a function
+    // * Call `apply` where we specify `this` but the function ignores it
+    // * Once we have `this`, use a structural getter to get the value of `self`
+    // * Last but not least, test whether `self` is defined or not.
+    //
+    // Whew!
+    let this = new_function("return this")
```
fitzgen: Do we not have bindings to `new Function` in `js_sys` already? Maybe we should just add that instead of making this custom binding.

crates/test/src/rt/mod.rs
```diff
@@ -20,6 +26,15 @@ pub struct Context {
     current_log: RefCell<String>,
     current_error: RefCell<String>,
     ignore_this_test: Cell<bool>,
+    formatter: Box<Formatter>,
+}
+
+trait Formatter {
```
fitzgen: Why not just use `std::io::Write` and `writeln!(..)`?

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -0,0 +1,444 @@
...
+        .stderr(Stdio::piped())
+        .stdin(Stdio::null());
+    let mut child = cmd.spawn()
+        .context(format!("failed to spawn {:?} binary", driver.path()))?;
+    let mut stdout = child.stdout.take().unwrap();
+    let mut stderr = child.stderr.take().unwrap();
+    let mut stdout = Some(thread::spawn(move || read(&mut stdout)));
+    let mut stderr = Some(thread::spawn(move || read(&mut stderr)));
+    let print_driver_stdio = Cell::new(true);
+    let _f = OnDrop(|| {
```
fitzgen: I feel like all this could be wrapped up in a `BackgroundChild` type like it is/was in the test-project-builder

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -0,0 +1,444 @@
...
+    shell.status(&format!("Visiting {}...", url));
+    client.goto(&id, &url)?;
+    shell.status("Loading page elements...");
+    let output = client.element(&id, "#output")?;
+    let logs = client.element(&id, "#console_log")?;
+    let errors = client.element(&id, "#console_error")?;
+
+    // At this point we need to wait for the test to finish before we can take a
+    // look at what happened. There appears to be no great way to do this with
+    // the webdriver protocol today (in terms of synchronization), so for now we
```
fitzgen: What did the geckodriver npm package we used to use do? Was it polling under the hood?

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -0,0 +1,444 @@
...
+}
+
+enum Method<'a> {
+    Get,
+    Post(&'a str),
+    Delete,
+}
+
+impl Client {
+    fn new_session(&self, driver: &Driver) -> Result<String, Error> {
```
fitzgen: Did you consider using https://crates.io/crates/fantoccini? I guess that would pull in a lot of dependencies via tokio et al...

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -0,0 +1,444 @@
...
+                None => continue,
+            };
+            return Ok((ctor(name.into()), env_args(name)))
+        }
+
+        // TODO: download an appropriate driver? How to know which one to
+        //       download?
+
+        bail!("failed to find a suitable WebDriver binary to drive headless \
+               testing")
```
fitzgen: This message should include info about the env vars that we check and how we look for `foodriver` on `$PATH` or else we will drive users crazy

crates/test/src/lib.rs
```diff
@@ -20,6 +20,31 @@ macro_rules! console_log {
...
+///
+///     wasm_bindgen_test_configure!(foo bar baz);
+///
+/// where all of `foo`, `bar`, and `baz`, would be recognized options to this
+/// macro. The currently known options to this macro are:
+///
+/// * `run_in_browser` - requires that this test is run in a browser rather than
+///   node.js, which is the default for executing tests.
+///
+/// This macro may be invoked at most one time per test suite.
```
fitzgen: A "test suite" is a test crate, right? like `tests/browser.rs` or `tests/all/*`? *Not* per-module, right? Would be good to clarify this point.

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -165,6 +174,15 @@ enum Driver {
 }
 
 impl Driver {
+    /// Attempts to find an appropriate WebDriver server binary to execute tests
+    /// with. Performs a number of heuristics to find one available, including:
```
fitzgen: Starting to feel like this webdriver module could be its own crate...

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
alexcrichton: Yeah I initially started there but it ended up pulling lots of deps we didn't end up really needing, so I opted to go with the hopefully simple route of just handrolling what we need

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
alexcrichton: I'm not actually sure, but I sort of assume so! I couldn't find anything in the webdriver protocol for synchronization myself :(

crates/test/src/lib.rs
alexcrichton: True!

crates/test/src/rt/mod.rs
alexcrichton: Hm perhaps! I think it's too much abstraction though because we can't generate any errors here and it'd require utf8 parsing and whatnot. This isn't really reused anywhere so we shouldn't lose too much I think without the trait implementation.

alexcrichton:
Updated!

crates/test/src/rt/mod.rs
```diff
@@ -0,0 +1,379 @@
...
+// Here's a rough and (semi) high level overview of what happens when this crate
+// runs.
+//
+// * First, the user runs `cargo test --target wasm32-unknown-unknown`
+//
+// * Cargo then compiles all the test suites (aka `tests/*.rs`) as wasm binaries
+//   (the `bin` crate type). These binaries all have entry points that are
+//   `main` functions, but it's actually not used. The binaries are also
+//   compiled with `--test`, which means they're linked to the standard `test`
+//   crate, but this crate doesn't work on wasm and so we bypass it entirely.
```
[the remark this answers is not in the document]
alexcrichton: Ah yeah those two properties though were why I consider it as "not working" in the sense that failing on the first failure isn't great for a test harness and similarly not being able to output anything is a bummer :(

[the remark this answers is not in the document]
kuviman: It is though nicer to have regular tests in case of developing a library that should work on both wasm and natively, even if limited like this.

I remember there is like wasm_syscall feature in std that should be able to enable printing, not sure about unwind.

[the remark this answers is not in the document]
alexcrichton: It would be nice to use `#[test]`, yeah, but customizing that and getting it working on stable is a long way off. The `wasm_syscall` feature is unlikely to ever be stabilized

brisad:
Previously I could run `js-sys` tests with this command:
`WASM_BINDGEN_FIREFOX_BIN_PATH=... cargo +nightly test -p js-sys --target wasm32-unknown-unknown`,
but after this change I only get the message
> Running server on http://127.0.0.1:8000 

in the terminal, and nothing more happens. How can I resolve that?

alexcrichton:
Ah sorry about that @brisad! The testing strategy has been changed now. By default those tests are "browser enabled" but headless mode isn't enabled. The intention is that you'd visit the page in a browser to see if it passes.

It's not a great UI though! In the meantime you can set `CI=1` in the environment and that'll enable headless testing, which'll be back to the normal `cargo test` workflow

brisad:
@alexcrichton Ah, I was super close trying out `CI=1` since I saw that somewhere before commenting here :)

Thanks for the explanation! Works fine!

PR #561, opened 2018-07-26
Enable History Web API

twilco:
Didn't see a tracking issue for this one, but here is the first of (hopefully) many!


twilco:
Also, not sure my workflow is the best...this is my first time working against a forked repo.  Currently I just run `git pull git@github.com:rustwasm/wasm-bindgen.git master` before I get to work on anything, and it looks like these (among my actual commits) are just continuing to stack up.  What should I do differently in my workflow to avoid this?

alexcrichton:
Awesome! Thanks so much!

There's a few ways you can update with the master branch with a git repo. Typically you don't actually need to do it unless there's a merge conflict, but when doing so one thing I'd recommend is to give the "remote" (this repository) a unique name, something like:

```
git remote add rustwasm git@github.com:rustwasm/wasm-bindgen.git`
```

(and using https://github.com/rustwasm/wasm-bindgen as a URL should also work)

After that if you want to pull in changes you can do:

```
git pull rustwasm master
```

to create a merge commit, or you can also do:

```
git fetch rustwasm
git rebase rustwasm/master
```

if you'd like to do a *rebase* which is slightly different in terms of how the history looks, but otherwise similar esque!

Issue #447, opened 2018-07-10
Add HTML elements' interfaces to web-sys


fitzgen:
This is a tracking issue for adding support for all the HTML element interfaces to the `web-sys` crate.

General info on HTML interfaces: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model#HTML_interfaces

How to add support for new Web APIs to `web-sys`: https://rustwasm.github.io/wasm-bindgen/web-sys/supporting-more-web-apis.html

## Interfaces

Here is a list of interfaces we need to support:

* [ ] `HTMLAnchorElement`
* [ ] `HTMLAppletElement`
* [ ] `HTMLAreaElement`
* [ ] `HTMLAudioElement`
* [ ] `HTMLBaseElement`
* [ ] `HTMLBodyElement`
* [ ] `HTMLBRElement`
* [ ] `HTMLButtonElement`
* [ ] `HTMLCanvasElement`
* [ ] `HTMLDataElement`
* [ ] `HTMLDataListElement`
* [ ] `HTMLDialogElement`
* [ ] `HTMLDirectoryElement`
* [ ] `HTMLDivElement`
* [ ] `HTMLDListElement`
* [ ] `HTMLElement`
* [ ] `HTMLEmbedElement`
* [ ] `HTMLFieldSetElement`
* [ ] `HTMLFontElement`
* [ ] `HTMLFormElement`
* [ ] `HTMLFrameElement`
* [ ] `HTMLFrameSetElement`
* [ ] `HTMLHeadElement`
* [ ] `HTMLHeadingElement`
* [ ] `HTMLHtmlElement`
* [ ] `HTMLHRElement`
* [ ] `HTMLIFrameElement`
* [ ] `HTMLImageElement`
* [ ] `HTMLInputElement`
* [ ] `HTMLKeygenElement`
* [ ] `HTMLLabelElement`
* [ ] `HTMLLegendElement`
* [ ] `HTMLLIElement`
* [ ] `HTMLLinkElement`
* [ ] `HTMLMapElement`
* [ ] `HTMLMediaElement`
* [ ] `HTMLMenuElement`
* [ ] `HTMLMetaElement`
* [ ] `HTMLMeterElement`
* [ ] `HTMLModElement`
* [ ] `HTMLObjectElement`
* [ ] `HTMLOListElement`
* [ ] `HTMLOptGroupElement`
* [ ] `HTMLOptionElement`
* [ ] `HTMLOutputElement`
[...]

PR #568, opened 2018-07-27
Adding in initial support for all HTML*Element interfaces.

jonathanKingston:
I'll keep adding the tests for this today when I get chance. However everything looks to be working locally (so happy to merge whenever you are :+1: ).


jonathanKingston:
I think this covers all of #447 minus the tests highlighted.

jonathanKingston:
The only issue I can see is the short elements interface names are lowercase:
`HtmlbrElement`

jonathanKingston:
@alexcrichton I changed the camelcasing to use a little helper function which fixing the naming of `HTMLBRElement` to be `HtmlBrElement` in Rust.

alexcrichton:
Nice! The last commit looks good to me as well.

It looks like there's a test failure, but it's likely not your fault as well (cc @xtuc, maybe a webpack bug?). @jonathanKingston want to tweak or comment out the test for now to get CI passing? With https://github.com/rustwasm/wasm-bindgen/pull/559 we can hopefully rewrite these tests to be a bit speedier as well!

fitzgen:
Wow!

alexcrichton:
👍

xtuc:
The error `wasm validation error: at offset 229: byte size mismatch in function section` isn't emitted by Webpack but by Firefox.

alexcrichton:
@xtuc yes but it could be a bug in the wasm file that Webpack produces, right?

xtuc:
Yes that's possible, could you please indicate me how to get the failing binary?

alexcrichton:
Ah it's passed now in that these tests have been enabled on master via other means, but if it comes up again I'll let you know!

PR #571, opened 2018-07-27
Adds `Array.prototype.splice()` to js-sys

joelgallant:
Looking at `concat`, `push`, `unshift`, etc. that take varargs, I found no pattern that allowed for it. If there is one, I'd be happy to improve this (and those) APIs.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -314,6 +314,13 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn sort(this: &Array) -> Array;
 
+    /// The splice() method changes the contents of an array by removing existing elements and/or
+    /// adding new elements.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice
+    #[wasm_bindgen(method)]
+    pub fn splice(this: &Array, start: u32, delete_count: u32, item: &JsValue) -> Array;
+
     /// The toLocaleString() method returns a string representing the elements of the array.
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -314,6 +314,13 @@ extern "C" {
     #[wasm_bindgen(method)]
     pub fn sort(this: &Array) -> Array;
 
+    /// The splice() method changes the contents of an array by removing existing elements and/or
+    /// adding new elements.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
```
alexcrichton: I think this may be a copy/paste error at the end with `slice` instead of `splice`?

crates/js-sys/src/lib.rs
```diff
@@ -314,6 +314,13 @@ extern "C" {
...
+    #[wasm_bindgen(method)]
+    pub fn splice(this: &Array, start: u32, delete_count: u32, items: &JsValue) -> Array;
```
alexcrichton: I think the argument `items` here should be `item`, right? In that if an array is passed it's not treated as varargs, just as an array in an array?

joelgallant: totally. tried to treat it as an array at first and forgot to change the name.

crates/js-sys/src/lib.rs
joelgallant: :+1:

alexcrichton:
Thanks!

PR #573, opened 2018-07-27
Try to enable all webidls

derekdreery:
I tried to enable all webidl files, then disabled the ones that broke the build. They were caused by

 - enum variant not valid ident ("2x") - (only 1 case of this).
 - overloaded methods (e.g. `new` with a few different argument combinations)
 - optional primitive types like `Option<i16>` not implementing `OptionIntoWasmAbi`


sendilkumarn:
wow 😮 that is a brave move 👍

fitzgen:
Wow indeed!!

I'm a bit hesitant to enable *all* the webidls we can at this moment in time without also writing a bunch more sanity tests. I'm not sure what the best approach here is.

Long term, I would like to have **super** good test coverage of the webidl frontend and all its features, plus just sanity integration tests of the actual DOM APIs and tests for any regressions we uncover related to cross browser support and all that.

Maybe that means we should just enable all these interfaces and then really double down on improving the webidl frontend and getting it feature complete?

Do you have more thoughts @alexcrichton ?

alexcrichton:
I think I agree with @fitzgen basically. I'd love to have some more tests, but this is a lot of functionality and it's not necessarily required. I definitely agree that the main body of work from here will now be improving the webidl frontend, filling out missing features and enabling even more of the currently ignored APIs.

all in all, I'm 👍  to merge

derekdreery:
You might want to consider that build times go up a lot, since there's about 200000 lines of code in bindings.rs

If you don't want to enable all the widls, I can move them back into available but keep the fixes that are in this PR.

derekdreery:
I've modified the PR a bit
 1. Categorized the failing WebIDL files by reason.
 2. Added a list of all WebIDL files to `README.md` with a checkbox that will be ticked as each file has full test coverage and is fully working.
 3. Updated the contributing documentation to reflect the fact that there is work to do even if a WebIDL file is added, if it doesn't yet have a full test suite.

Hopefully this helps to address the concern:

> Long term, I would like to have super good test coverage of the webidl frontend and all its features, plus just sanity integration tests of the actual DOM APIs and tests for any regressions we uncover related to cross browser support and all that.

Let me know what you think!

alexcrichton:
This sounds and looks great to me, thanks again for doing this! Perhaps the files with blank idents could also be unavailable temporarily? (generating `BlankName` seems a bit odd for now)

derekdreery:
I'll make that change.

derekdreery:
I changed the code to error on blank `rust_ident`, and it turns out all the errors came from enum variants, so I've moved those webidl files into `unavailable_enum_ident`.

alexcrichton:
Ok awesome, thanks @derekdreery!

PR #578, opened 2018-07-28
Adding tests for script and style elements.

jonathanKingston:


alexcrichton:
Looks great, thanks! Looks like some stray `.swp` files were introduced though?

jonathanKingston:
Ah dang, I knew that would eventually get me. Thanks

Issue #444, opened 2018-07-10
Add 2D canvas support to web-sys


fitzgen:
This is a tracking issue for adding support for the canvas 2D API to the `web-sys` crate.

General info on canvas 2D: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API

How to add support for new Web APIs to `web-sys`: https://rustwasm.github.io/wasm-bindgen/web-sys/supporting-more-web-apis.html

## Interfaces

Here is a list of interfaces we need to support:

* [ ] HTMLCanvasElement
* [ ] CanvasRenderingContext2D
* [ ] CanvasGradient
* [ ] CanvasPattern
* [ ] ImageBitmap
* [ ] ImageData
* [ ] TextMetrics

## Example

* [ ] Add an example to `examples/` that shows off working with canvas 2D

teovoinea:
I’d like to pick this up! 😊

teovoinea:
Is there a story for handling mutliple constructor functions? Right now doing a `cargo build` on `CanvasRenderingContext2D.webidl` errors with

```bash
the name `__wbindgen_describe___widl_f_new_Path2D` is defined multiple times
```

which makes sense based on the idl

https://github.com/teovoinea/wasm-bindgen/blob/729e67423b8a27f89b40de295fcaf1f93dbf1b50/crates/web-sys/webidls/enabled/CanvasRenderingContext2D.webidl#L392-L399

afdw:
What do you think, which names in Rust should be used for constructors/functions with same names, but different argument types? Should they be generated from types names? Or manually annotated somewhere?

alexcrichton:
The current thinking is that we'd do something programmatic like:

* First constructor is named `new`
* Second constructor with arg `foo` is named `new_with_foo`
* Third constructor with arg `foo` and `bar` is named `new_with_foo_and_bar`
* etc...

And at least see how far that gets us, although we're also ok commenting out the duplicate constructors for now!

afdw:
The problem is that for WebSockets all 4 overloads of method `send` are useful, and they have same argument names:
```webidl
  [Throws]
  void send(DOMString data);

  /*[Throws]
  void send(Blob data);

  [Throws]
  void send(ArrayBuffer data);

  [Throws]
  void send(ArrayBufferView data);
```

alexcrichton:
[...]

PR #579, opened 2018-07-28
Add renaming of conflicting constructors and operations

afdw:
Also enable WebSocket and Blob for testing.
image
image


crates/webidl/src/first_pass.rs
```diff
@@ -103,43 +114,177 @@ impl FirstPass for webidl::ast::Enum {
...
     }
 }
 
-impl FirstPass for webidl::ast::NonPartialInterface {
-    fn first_pass<'a>(&'a self, record: &mut FirstPassRecord<'a>) -> Result<()> {
-        if record.interfaces.insert(self.name.clone()) {
-            warn!("Encountered multiple declarations of {}", self.name);
+impl FirstPass<()> for webidl::ast::NonPartialInterface {
+    fn first_pass<'a>(&'a self, record: &mut FirstPassRecord<'a>, (): ()) -> Result<()> {
+        if record.interfaces.contains_key(&self.name) {
```
alexcrichton: Could this perhaps use `record.interfaces.entry(...).or_insert_with(...)` to deduplicate these branches a bit?

crates/webidl/src/first_pass.rs
```diff
@@ -103,43 +114,177 @@ impl FirstPass for webidl::ast::Enum {
...
+            member.first_pass(record, &self.name)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl FirstPass<()> for webidl::ast::PartialInterface {
+    fn first_pass<'a>(&'a self, record: &mut FirstPassRecord<'a>, (): ()) -> Result<()> {
+        if !record.interfaces.contains_key(&self.name) {
```
alexcrichton: Similar to above this may look a bit cleaner with the `entry` method, but if this shows up in a few places it could also be a helper method perhaps?

crates/webidl/src/util.rs
```diff
@@ -134,6 +134,91 @@ pub enum TypePosition {
     Return,
 }
 
+trait TypeToString {
```
alexcrichton: Could this have a comment indicating what it's used for?

alexcrichton:
Nice!

I'm wondering if we can perhaps trim down the renamings as well? The current implementation is a bit ambitious covering *everything* but some of the names generated can likely be super funky. I wonder if we could whitelist a few at a time to make sure the methods being generated have semi-reasonable names and review as we go along?

For example optional arguments come out as `nullable_*` but it may be more rusty to do `opt_*`, and similarly `unsigned_short` is probably just better as `u16` in Rust.

I wonder as well, is it perhaps more clear to include the name of the argument from WebIDL rather than the type? That may make the constructors a bit more readable by default as well

crates/webidl/src/first_pass.rs
afdw: I wanted to do something like https://play.rust-lang.org/?gist=3c1e682a8747dd4c2a70c8b9f81898fc, but this does not compile.

afdw:
Now it looks like this:
image
image
Constructor with `url` and `protocols` argument names uses argument types naming because there is another constructor, but it is currently unavailable:
```webidl
[Exposed=(Window,Worker),
 Constructor(DOMString url),
 Constructor(DOMString url, DOMString protocols),
 Constructor(DOMString url, sequence<DOMString> protocols)]
interface WebSocket : EventTarget {
```

ohanar:
Very small stylistic thing is that `new_with_url` etc. seems a bit strange to me, maybe for overloaded constructors, we could do `from_url`, `from_foo_and_bar`, ... or `with_url`, `with_foo_and_bar`, ... and then just `new` for the constructor with no arguments if it exists.

afdw:
Yes, I was thinking about this problem; we need better naming here. I have just implemented what @alexcrichton suggested: https://github.com/rustwasm/wasm-bindgen/issues/444#issuecomment-408619825

alexcrichton:
Ah good points @ohanar! The construction schemes for `HashMap` seem to agree, perhaps for these named constructors we should drop `new_` and just have `with_foo_and_bar`?

@afdw if you want to implement that real quick I can merge!

afdw:
@alexcrichton done.

alexcrichton:
Looks awesome, thanks @afdw!

PR #580, opened 2018-07-29
Create bindings for RegExp

brisad:
This adds bindings for most of RegExp
#275

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -2095,16 +2095,6 @@ extern {
     #[wasm_bindgen(static_method_of = RegExp, getter)]
     pub fn input() -> JsString;
 
-    /// The lastIndex is a read/write integer property of regular
-    /// expression instances that specifies the index at which to
-    /// start the next match.
-    ///
-    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex
-    #[wasm_bindgen(method, getter, structural, js_name = lastIndex)]
-    pub fn last_index(this: &RegExp) -> u32;
-    #[wasm_bindgen(method, setter, structural, js_name = set_lastIndex)]
-    pub fn set_last_index(this: &RegExp, index: u32);
-
     /// The non-standard lastMatch property is a static and read-only
     /// property of regular expressions that contains the last matched
     /// characters. RegExp.$& is an alias for this property.
@@ -2143,7 +2133,9 @@ extern {
     ///
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
     #[wasm_bindgen(constructor)]
-    pub fn new(pattern: &JsValue, flags: &str) -> RegExp;
+    pub fn new(pattern: &str, flags: &str) -> RegExp;
+    #[wasm_bindgen(constructor)]
+    pub fn new_regexp(pattern: &RegExp, flags: &str) -> RegExp;
 
     /// The non-standard rightContext property is a static and
     /// read-only property of regular expressions that contains the
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -2050,6 +2050,152 @@ extern "C" {
     pub fn set_prototype_of(target: &Object, prototype: &JsValue) -> bool;
 }
 
+// RegExp
+#[wasm_bindgen]
+extern {
+    #[derive(Clone, Debug)]
+    pub type RegExp;
+
+    /// The exec() method executes a search for a match in a specified
+    /// string. Returns a result array, or null.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec
+    #[wasm_bindgen(method)]
+    pub fn exec(this: &RegExp, text: &str) -> Option<Array>;
+
+    /// The flags property returns a string consisting of the flags of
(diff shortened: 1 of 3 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -2050,6 +2050,152 @@ extern "C" {
...
+extern {
+    #[derive(Clone, Debug)]
+    pub type RegExp;
+
+    /// The exec() method executes a search for a match in a specified
+    /// string. Returns a result array, or null.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec
+    #[wasm_bindgen(method)]
+    pub fn exec(this: &RegExp, text: &str) -> Option<Array>;
```
brisad: MDN states that on a successful match the returned value is an array. But in the example later on one can see that there are also properties `index` and `input` added to this array. I put Array here for now as I don't know how to add those properties. How can that be done?

alexcrichton: I think it's fine to return an `Array` for now, but the alternative is you could return a custom type here with `input` and `index` properties defined. For now though it's likely ok to rely on users to do that locally if they need to access the values.

crates/js-sys/src/lib.rs
```diff
@@ -2050,6 +2050,152 @@ extern "C" {
...
+    pub fn input() -> JsString;
+
+    /// The lastIndex is a read/write integer property of regular
+    /// expression instances that specifies the index at which to
+    /// start the next match.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex
+    #[wasm_bindgen(method, getter, structural, js_name = lastIndex)]
+    pub fn last_index(this: &RegExp) -> u32;
+    #[wasm_bindgen(method, setter, structural, js_name = set_lastIndex)]
```
alexcrichton: I think this `js_name` may not be quite right? Also is `structural` required here to get this working?

brisad: Ah, yes, I should have left a comment on this one.

At first I set `js_name = lastIndex` for the setter. But then I got a custom attribute panic with the message: "setters must start with `set_`." Simply adding it to the `js_name` solved the problem, and also made the test pass, so I left it at that. But it does look odd. What is the correct way to do it?

As for `structural`, I must admit I don't fully grok the documentation for it. But leaving it out gives the error `Error: descriptor for id='lastIndex' not found`. MDN documents this one method as `regexp.lastIndex` rather than `RegExp.prototype.lastIndex`, and so I thought `structural` causes it to not go through the property. But that was mostly a guess.

crates/js-sys/tests/wasm/RegExp.rs
```diff
@@ -0,0 +1,138 @@
...
+#[wasm_bindgen_test]
+fn multiline() {
+    let re = RegExp::new(&JsValue::from("foo"), "m");
+    assert!(re.multiline());
+}
+
+#[wasm_bindgen_test]
+fn new() {
+    let re = RegExp::new(&JsValue::from("foo"), "");
+    let re = RegExp::new(&JsValue::from(re), "g");
```
brisad: This test demonstrates that `RegExp::new` not only can accept a string as its first argument, but also another instance of `RegExp`. But can this be made more ergonomic?

The majority of the use cases of `RegExp` will probably use a string as its first argument, and it would be nice to be able to do `RegExp::new("foo", "g")` instead.

brisad:
Oh, I found that `RegExp` constructor seems to be throwing if the second argument has invalid flags. I couldn't find it documented in MDN though.

Should a `catch` attribute? (I tried but I get an error saying: `first argument of method must be a bare type`. I'm guessing it has something to do with that it is an constructor)

crates/js-sys/src/lib.rs
brisad: And just to be clear, the actual rust name always started with `set_` while I was facing that problem.

crates/js-sys/tests/wasm/RegExp.rs
alexcrichton: One possibility would be to do:

```rust
#[wasm_bindgen]
extern {
    #[wasm_bindgen(constructor)]
    fn new(s: &str, flags: &str) -> RegExp;
    #[wasm_bindgen(constructor)]
    fn new_regexp(r: &RegExp, flags: &str) -> RegExp;
}
```

maybe that could work here?

alexcrichton:
@brisad oh hm I think `catch` may not be supported on constructors yet? That may need a dedicated bug to support that. In the meantime though it's ok to not return a `Result`

crates/js-sys/src/lib.rs
alexcrichton: Ah ok the `structural` definitely makes sense to avoid the prototype (that's the attribute's purpose), but I think the `js_name` here needs to be `lastIndex` because otherwise this will set the `set_lastIndex` property on the object

brisad: Alright. Since `js_name = lastIndex` panics I'll remove `last_index` from this PR for now. I can open a new issue for the panic.

crates/js-sys/tests/wasm/RegExp.rs
brisad: Yes, indeed. I'll update the PR accordingly.

brisad:
@alexcrichton I've updated the PR after your review comments

alexcrichton:
Awesome, thanks!

PR #581, opened 2018-07-29
Add documentation and MDN links for webidl files. Fixes #513

### LINKED ISSUE #513: Automatic documentation generation for web-sys bindings
It'd be awesome if we could automatically hook up something like `Element::class_name` to the [MDN docuementation for the method](https://developer.mozilla.org/en-US/docs/Web/API/Element/className)!

jonathanKingston:
image


alexcrichton:
Nice! This is gonna make browsing the documentation *much* nicer

PR #592, opened 2018-07-31
Fix `web-sys` history unit test

twilco:
In an actual browser, the changing of the browser history using the binding worked a little too well, and caused the test to fail if you refreshed the page or manually used the back and forward buttons.  The stateful stuff has been removed - the remaining two assertions should adequately test that the binding works, which is the point of these tests anyway.


alexcrichton:
Thanks!

PR #596, opened 2018-07-31
Testing web-sys for input, heading and title elements.

jonathanKingston:
Submitting this as it bitrotted heavily and I don't want to lose it.
Input needs some follow up work. I verified everything besides the exising history test is working in Chrome.

I wasn't able to get this to work via cli?
Is the new command:
```
cargo +nightly test --target wasm32-unknown-unknown
```
I ripped out the existing TODO comments and moved to checking off the readme instead.


alexcrichton:
Awesome, thanks!

That CLI command should work yeah (in `web-sys`) and it should spawn a server in the background. Could you gist the logs though if that's failing to execute?

(the test framework is still pretty new and we're trying to iron out the kinks!)

PR #602, opened 2018-08-01
Start migrating `wasm_bindgen` tests to `wasm_bindgen_test`

alexcrichton:
This commit starts migrating the `wasm_bindgen` tests to the `wasm_bindgen_test`
framework, starting to assemble the coffin for
`wasm-bindgen-test-project-builder`. Over time all of the tests in
`tests/all/*.rs` should be migrated to `wasm_bindgen_test`, although they may
not all want to go into a monolithic test suite so we can continue to test for
some more subtle situations with `#[wasm_bindgen]`.

In the meantime those, the `tests/all/api.rs` tests can certainly migrate!


tests/wasm/api.rs
```diff
@@ -0,0 +1,137 @@
+use wasm_bindgen_test::*;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(module = "tests/wasm/api.js", version = "*")]
```
fitzgen: Do we need the version thing here?

alexcrichton: Unfortunately yeah, I've been meaning to relax this in `#[wasm_bindgen]` for awhile, but since the module doesn't start with `"./"` it requires `version` to be present (it's a bit of a hack today)

PR #605, opened 2018-08-01
Upgrade `failure` and fix deprecation warnings

alexcrichton:
Should be pretty standard!


crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -48,7 +48,7 @@ fn main() {
         Err(e) => e,
     };
     eprintln!("error: {}", err);
-    for cause in err.causes().skip(1) {
```
fitzgen: Yay! I've literally had this `skip(1)` in every single project that used `failure`!

Issue #601, opened 2018-08-01
Improve #[wasm_bindgen] errors with `compile_error!`


alexcrichton:
Right now the procedural macro simply panics if anything goes wrong, but this is pretty bad! The error messages coming out look pretty opaque at times and are often quite subpar, especially in that they're pointing to the wrong items. 

I've just realized, however, that the procedural macro can generate `compile_error!` invocations which gives us the ability to generate custom compiler errors as well as custom error spans. An example of this is:

https://gist.github.com/alexcrichton/35f9b54585b39d16a887cffd4eabdb2d

We should refactor the macro crate to use `Result` for error handling and use a structured error type which has a string and start/end spans, and that way we can generate nicely spanned error messages, yay!

mkane0814:
Yes please! These errors have been really frustrating to work around. Does anyone know what causes them?

alexcrichton:
I think this is effectively done now, yay!

PR #608, opened 2018-08-01
Result-ify `src/parser.rs` 

alexcrichton:
This commit converts all of `src/parser.rs` away from panics to using
`Diagnostic` instead. Along the way this adds a test case per changed `panic!`,
ensuring that we don't regress in these areas!

cc #601


alexcrichton:
> How does one run the UI tests? 

We've got some [handy documentation](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/macro/README.md#testing) about that!

> Is it the same as all of our other tests? 

Indeed! A simple `cargo test -p ui-tests` should do the trick.

> Also, are rustc error messages stable enough that we can rely on this stderr content remaining the same?

I believe that since all this output is *our* error messages it should hopefully be pretty stable. That being said I wouldn't be surprised if these tests break over time. I'd like to eventually add UI tests as well for failing conditions like "you specified an argument that didn't implement `IntoWasmAbi`" to make sure it looks reasonable, but that sort of error message is practically guaranteed to change over time.

I think we may eventually move this test suite into an "allow failures" category on Travis and only update it every now and then (such as when the test suite itself is modified), but for now we hopefully won't have too many issues!

PR #627, opened 2018-08-03
Run Chrome headless tests on CI

### LINKED ISSUE #622: web-sys: Add Chrome headless testing to CI
We'll want to edit the travis config to download `chromedriver` and also probably Chrome itself. We should the configure js-sys and web-sys tests to both run with `geckodriver` and `chromedriver` on CI, ensuring that tests pass in both Firefox and Chrome.

alexcrichton:
Closes #622

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,12 @@ INSTALL_GECKODRIVER: &INSTALL_GECKODRIVER
   |
     curl --retry 5 -LO https://github.com/mozilla/geckodriver/releases/download/v0.21.0/geckodriver-v0.21.0-linux64.tar.gz
     tar xf geckodriver-v0.21.0-linux64.tar.gz
-    export GECKODRIVER=`pwd`/geckodriver
+
+INSTALL_CHROMEDRIVER: &INSTALL_CHROMEDRIVER
+  |
+    curl --retry 5 -LO https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip
+    unzip chromedriver_linux64.zip
+    export CHROMEDRIVER_ARGS=--verbose
 
 DEPLOY_TO_GITHUB: &DEPLOY_TO_GITHUB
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -78,10 +82,14 @@ matrix:
       install:
         - *INSTALL_NODE_VIA_NVM
         - *INSTALL_GECKODRIVER
+        - *INSTALL_CHROMEDRIVER
       script:
-        - cargo test -p web-sys --target wasm32-unknown-unknown
+        - export RUST_LOG=wasm_bindgen_test_runner
+        - CHROMEDRIVER=`pwd`/chromedriver cargo test -p web-sys --target wasm32-unknown-unknown
```
fitzgen: I think this should probably be a separate travis job

.travis.yml
```diff
@@ -90,10 +98,14 @@ matrix:
       install:
         - *INSTALL_NODE_VIA_NVM
         - *INSTALL_GECKODRIVER
+        - *INSTALL_CHROMEDRIVER
       script:
-        - cargo test -p js-sys --target wasm32-unknown-unknown
+        - export RUST_LOG=wasm_bindgen_test_runner
+        - CHROMEDRIVER=`pwd`/chromedriver cargo test -p js-sys --target wasm32-unknown-unknown
```
fitzgen: ditto

alexcrichton:
Ok I think I figured out the incantation to get this working.

@fitzgen I agree in principle yeah but CI already takes so long I don't think we can spare two more whole builders when running Chrome after Firefox takes ~seconds

fitzgen:
> @fitzgen I agree in principle yeah but CI already takes so long I don't think we can spare two more whole builders when running Chrome after Firefox takes ~seconds

Ah, OK. I was thinking this would help speed things up, but if not then definitely *don't* do this!

PR #630, opened 2018-08-03
Add support for optional numbers

### LINKED ISSUE #624: WebIDL: enable bindings that take `Option<scalar>`
Right now support for things like `Option<u32>` isn't implemented, so we'll need to implement that first to enable the bindings in https://github.com/rustwasm/wasm-bindgen/tree/master/crates/web-sys/webidls/unavailable_option_primitive

afdw:
Fixes #624.


alexcrichton:
Thanks for this!

It looks like the in-memory representation of optional types is a `u32` indicating whether it's `Some` or `None` and then the payload itself? That seems plausible to me!

@fitzgen was experimenting using the "global stack" as well for the payload and/or discriminant, as this I believe would enable a blanket implementation for all `Option<T>` rather than needing a per-type solution (like we need today)

Another option I was thinking of is that if we want to go the super-specialized route we could support `Option<T>` for all `T` that aren't `f64` or `u64` by using `f64` as the ABI value, choosing an out-of-range sentinel for `None` for each type (e.g. `None::<u32>` becomes NaN, `None::<f32>` becomes `Some(f64::max_value())`, etc)

I'm not sure what the best option is here, but I'm curious what @fitzgen thinks of this too!

afdw:
Other approaches are possible, yes. But implementing manually means that it's possible to do it better, more compact (like I do with optional `i8`, `u8`, `i16` and `u16` types, using a single `u32` argument/return type everywhere). And we don't have too many primitives.

Is passing (and converting) float significantly cheaper than passing 2 integers / integer and 32-bit float?

alexcrichton:
True yeah! I personally like the idea of using more compact representations as well, and I think using `u32` where possible is a great example of that.

In terms of cost I'm not actually sure, but I'd imagine that it's pretty negligible for wasm-bindgen's purposes

crates/cli-support/src/descriptor.rs
```diff
@@ -154,7 +154,27 @@ impl Descriptor {
         }
     }
 
-    pub fn get_64bit(&self) -> Option<bool> {
+    pub fn is_primitive(&self) -> bool {
```
alexcrichton: Could this perhaps be named slightly differently to something like `is_wasm_native_type` or something like that?

crates/cli-support/src/descriptor.rs
```diff
@@ -154,7 +154,27 @@ impl Descriptor {
...
+        match *self {
+            Descriptor::I32
+            | Descriptor::U32
+            | Descriptor::F32
+            | Descriptor::F64 => true,
+            _ => return false,
+        }
+    }
+
+    pub fn is_as_u32(&self) -> bool {
```
alexcrichton: Could this perhaps be renamed to `is_abi_as_u32`?

crates/cli-support/src/js/js2rust.rs
```diff
@@ -192,7 +192,74 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
...
+                self.cx.expose_is_like_none();
+                self.js_arguments.push((name.clone(), "number".to_string()));
+
+                if self.cx.config.debug {
+                    self.cx.expose_assert_num();
+                    self.prelude(&format!(
+                        "\n\
+                            if (!isLikeNone({0})) {{\n\
+                                _assertNum({0});\n\
+                            }}\n\
```
alexcrichton: Nowadays all the `\n\` markers here should be fine to clean up, the auto-formatting will strip them all away anyway!

fitzgen:
I don't have any concrete work in progress, so don't block this on me!

Regarding blanket vs targeted impls: I also doubt that there is observable overhead to putting a discriminant on the global stack. But targeted impls mean that we have to write many impls instead of a single one. I suspect that we will want to move to a blanket impl eventually. But it is also an implementation detail that can be changed at a later date.

afdw:
I think it's now implemented for all of the primitive types.

alexcrichton:
Thanks so much @afdw!

PR #639, opened 2018-08-04
Migrate from the `webidl` crate to `weedle`

### LINKED ISSUE #620: use weedle instead of webidl-rs
(See #506 for prior experimental implementation of this issue.)

There are currently a number of issues with weedle that need to be addressed upstream before we can really use it.
- Currently weedle never errors. This can be easily fixed as it is a small oversight, but the best context you could give at the moment without more thorough changes is pointing to the definition block that failed.
- Weedle parses the subset of the WebIDL grammar that is used by the WebIDL spec. The main place that these differ is for extended attributes, where the grammar is far more general that what the spec needs. mozilla-central's webidl files use this more general grammar with attribute="some string literal" all over the place. From a cursory glace through webidl-rs' source, it doesn't seem like it supports the full general grammar of extended attributes either (although some commented out code looks like there was the intention at some point), however, it seems to parse the main offender previously mentioned just fine.
- web-sys gets truncated due to a combination of the two previous bullet points.

Note that the author of weedle is not planning on working on it for the near future, and has offered to transfer ownership should we decide to pursue this issue.

alexcrichton:
This commit migrates the `wasm-bindgen-webidl` crate from the `webidl` parser to
`weedle`. The main rationale for doing this is that `webidl` depends on
`lalrpop`, which is quite a large dependency and takes a good deal of time to
compile. The `weedle` crate, however, depends on `nom` and is much faster to
compile.

Almost all translations were pretty straightforward. Some abstractions changed
and/or were introduced in this commit when moving to `weedle` like the
`ToSynType` trait, but otherwise the generated bindings should be the same. It's
been verified that the `weedle`-generated bindings are exactly the same as the
`webidl`-generated bindings, with the one exception of `weedle` generates one
more method, `WebGpuCommandEncoder::transition_buffer`. It's not clear currently
why `webidl` didn't generate this method, as its [idl] is pretty straightforward!

This commit is using a [fork] of `weedle` currently which has a number of fixes
for parsing our WebIDL, although all the fixes are quite minor!

Closes #620

[idl]: https://github.com/rustwasm/wasm-bindgen/blob/d66b834afd9b45daefc4dcd35a74f3274fc0ab12/crates/web-sys/webidls/enabled/WebGPU.webidl#L499
[fork]: https://github.com/alexcrichton/weedle/tree/fix-for-web-sys

### DIFF
--- a/crates/webidl/Cargo.toml
+++ b/crates/webidl/Cargo.toml
@@ -20,4 +20,4 @@ proc-macro2 = "0.4.8"
 quote = '0.6'
 syn = { version = '0.14', features = ['full'] }
 wasm-bindgen-backend = { version = "=0.2.15", path = "../backend" }
-webidl = "0.7.0"
+weedle = "0.6"
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -1,12 +1,16 @@
-use std::iter::{self, FromIterator};
+use std::iter::FromIterator;
 
 use backend;
 use backend::util::{ident_ty, leading_colon_path_ty, raw_ident, rust_ident};
 use heck::{CamelCase, SnakeCase};
 use proc_macro2::Ident;
 use syn;
-use webidl;
-use webidl::ast::ExtendedAttribute;
+use weedle;
+use weedle::attribute::{ExtendedAttributeList, ExtendedAttribute};
+use weedle::argument::{Argument, SingleArgument};
+use weedle::common::Identifier;
+use weedle::types::*;
+use weedle::literal::{ConstValue, FloatLit, IntegerLit};
 
 use first_pass::FirstPassRecord;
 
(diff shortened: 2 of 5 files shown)


alexcrichton:
cc @ohanar, curious to get your take on this!

crates/webidl/src/util.rs
```diff
@@ -482,65 +738,66 @@ impl<'a> FirstPassRecord<'a> {
...
+                }
             }
         }
 
         Some(res)
     }
 
     /// Create a wasm-bindgen function, if possible.
-    pub fn create_function<'b, I>(
+    pub fn create_function<'b, 'c: 'b>(
```
ohanar: These lifetimes aren't needed anymore, are they?

crates/webidl/src/util.rs
```diff
@@ -34,35 +38,380 @@ pub fn mdn_doc(class: &str, method: Option<&str>) -> String {
...
+                                panic!("literal too big: {}", orig_text)
+                            }
+                        }
+                        ast::ConstValue::UnsignedIntegerLiteral(n)
+                    }
+                }
+            };
+            match lit {
+                IntegerLit::Hex(h) => mklit(h.0, 16, 2),
+                IntegerLit::Oct(h) => mklit(h.0, 8, 2),
```
ohanar: WebIDL octal literals start with the prefix of `0` rather than `0o`, so the offset here should be 1, not 2.

crates/webidl/src/util.rs
```diff
@@ -34,35 +38,380 @@ pub fn mdn_doc(class: &str, method: Option<&str>) -> String {
...
+                let (negative, text) = if orig_text.starts_with("-") {
+                    (true, &orig_text[1..])
+                } else {
+                    (false, orig_text)
+                };
+                if text == "0" {
+                    return ast::ConstValue::SignedIntegerLiteral(0)
+                }
+                let text = &text[offset..];
+                match i64::from_str_radix(text, base) {
```
ohanar: Why can't we just parse it as a u64 in the first place? I.e. something like this:
```rust
let n = u64::from_str_radix(text, base).unwrap_or_else(|_| panic!("literal too big: {}", orig_text));
if negative {
    let n = if n > (i64::min_value() as u64).wrapping_neg() {
        panic!("literal too big: {}", orig_text)
    } else {
        n.wrapping_neg() as i64
    };
    ast::ConstValue::SignedIntegerLiteral(n)
} else {
    ast::ConstValue::UnsignedIntegerLiteral(n)
}
```

crates/webidl/src/util.rs
```diff
@@ -34,35 +38,380 @@ pub fn mdn_doc(class: &str, method: Option<&str>) -> String {
...
+            ArrayBuffer(_)
+            | DataView(_)
+            | Error(_)
+            | FrozenArrayType(_)
+            | Object(_)
+            | Promise(_)
+            | RecordType(..)
+            | Sequence(_)
+            | Symbol(_) => {
+                None
```
afdw: When we will add an actual support for these types, it will be quiet likely that adding support for stringifying will be forgotten. So, as it should be very trivial, adding support for turning these types into strings now might be a better approach?

crates/webidl/Cargo.toml
```diff
@@ -20,4 +20,4 @@ proc-macro2 = "0.4.8"
 quote = '0.6'
 syn = { version = '0.14', features = ['full'] }
 wasm-bindgen-backend = { version = "=0.2.15", path = "../backend" }
-webidl = "0.7.0"
+weedle = { git = 'https://github.com/alexcrichton/weedle', branch = 'fix-for-web-sys' }
```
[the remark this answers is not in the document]
alexcrichton: Certainly! I wanted to make sure this PR looked good from stakeholders and such and after that I'd be totally fine reaching out to transfer (if they're still willing)

crates/webidl/src/util.rs
alexcrichton: Ah that's much better!

alexcrichton:
@csharad would you still be up for transferring the `weedle` crate to us?

csharad:
@alexcrichton Absolutely, I would be happy to. To whom should I transfer it to?

alexcrichton:
@csharad ok great! You can transfer it to me and I'll transfer it to the `rustwasm` organization.

csharad:
@alexcrichton You already have a `weedle` repo, I could not transfer.

alexcrichton:
oops, deleted now!

csharad:
@alexcrichton What about the crate ownership?

alexcrichton:
Thanks @csharad! I've transferred the repository to https://github.com/rustwasm/weedle, set up @rustwasm/weedle as a team, and have added you, me, fitzgen, and @ohanar on the team!


@csharad also, mind adding me as an owner on crates.io?

csharad:
@alexcrichton Sent you the request.

alexcrichton:
@csharad ah I don't see it on crates.io :(

alexcrichton:
Ok great, got it now! And the team is also added as an owner -- https://crates.io/crates/weedle

afdw:
You probably also want to update links the repository and the homepage.

PR #640, opened 2018-08-04
Implement RFC #2 - casting hierarchy between JS values

alexcrichton:
This is an implementation of https://github.com/rustwasm/rfcs/pull/2 for the `wasm-bindgen` suite of crates. Nothing too too tricky here and the commits inline should explain what they're doing.

One question to grapple with is that today we generate `From<JsValue> for ImportedType`, but that's not intended from RFC 2, so we'll want to remove it at some point. This PR currently doesn't delete it so we can avoid making a major bump to 0.3, but we may want to go ahead and bump to 0.3 depending on other schedules!

> **Note:** at the time of this writing the RFC isn't merged yet, so this shouldn't be merged until that's done.


src/cast.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    /// Performs a zero-cost unchecked cast into the specified type.
+    ///
+    /// This method will convert the `self` value to the type `T`, where both
+    /// `self` and `T` are simple wrappers around `JsValue`. This method **does
+    /// not check whether `self` is an instance of `T`**. If used incorrectly
+    /// then this method may cause runtime exceptions in both Rust and JS, this
+    /// shoudl be used with caution.
```
fitzgen: "shoudl" -> "should"

crates/backend/src/codegen.rs
```diff
@@ -527,84 +529,85 @@ impl ToTokens for ast::ImportType {
                 obj: ::wasm_bindgen::JsValue,
             }
 
-            impl ::wasm_bindgen::describe::WasmDescribe for #name {
-                fn describe() {
-                    ::wasm_bindgen::JsValue::describe();
+            #[allow(bad_style)]
+            const #const_name: () = {
```
fitzgen: lol this is a nice/nasty trick

crates/backend/src/codegen.rs
```diff
@@ -594,18 +596,63 @@ impl ToTokens for ast::ImportType {
                     }
                 }
 
+                // TODO: remove this on the next major version
```
fitzgen: Can one mark trait impls as deprecated?

src/cast.rs
afdw: Still not fixed in 2 other places.

alexcrichton: Oops thanks, fixed now.

guide/src/design/import-customization.md
```diff
@@ -168,3 +168,25 @@ possibilities!
 
   All of these functions will call `console.log` in JS, but each identifier
   will have only one signature in Rust.
+
+* `extends = Bar` - this can be used to say that an imported type extends
```
fitzgen: This is going to need to move to `guide/src/reference/attributes/on-js-imports/extends.md` and a corresponding entry in the SUMMARY.md table of contents

alexcrichton:
Updated!

guide/src/design/import-customization.md
```diff
@@ -0,0 +1,170 @@
+# Customizing import behavior
```
fitzgen: This file got accidentally resurrected.

alexcrichton: Oops, should be dead for real now

PR #646, opened 2018-08-04
Allow js_name attribute to accept a string

### LINKED ISSUE #583: Support quoted strings as js_name values
In addition to supporting `js_name = foo` we should also support a quoted variant like `js_name = "$1"`. This is needed to be able to create bindings for the methods `$1-$9` in `RegExp`. (#275)

brisad:
I think there's room for improvement in these changes. But I'm adding it here to show how far I got, and to see whether it's a viable approach.

Fixes #583

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -149,11 +149,11 @@ impl BindgenAttrs {
     }
 
     /// Get the first js_name attribute
-    fn js_name(&self) -> Option<&Ident> {
+    fn js_name(&self) -> Option<&str> {
         self.attrs
             .iter()
             .filter_map(|a| match a {
-                BindgenAttr::JsName(s) => Some(s),
+                BindgenAttr::JsName(s) => Some(&s[..]),
                 _ => None,
             })
             .next()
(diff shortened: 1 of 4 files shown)


alexcrichton:
This looks great to me, thanks!

This strategy I think will definitely work out. We'll want to keep an `Ident` for the Rust name but that's already done here too!

brisad:
Thanks! I'll finish this one up then and add some tests.

crates/macro-support/src/parser.rs
```diff
@@ -365,9 +369,10 @@ impl<'a> ConvertToAst<(BindgenAttrs, &'a Option<String>)> for syn::ForeignItemFn
     fn convert(self, (opts, module): (BindgenAttrs, &'a Option<String>))
         -> Result<Self::Target, Diagnostic>
     {
-        let js_name = opts.js_name().unwrap_or(&self.ident).clone();
+        let default_name = self.ident.to_string();
```
brisad: I used an explicit `let` here to make it live long enough. Is there a better way to do this? Now we're unconditionally allocating a string even though it might not be needed.

brisad:
@alexcrichton This PR is ready for review. :-)

alexcrichton:
👍 

Thanks so much!

Issue #603, opened 2018-08-01
Migrate `wasm_bindgen` crate's tests to `wasm_bindgen_test`


alexcrichton:
The test framework is shaping up quite nicely now, so we should purge the last and final use of `wasm-bindgen-test-project-builder`! The steps for this will be to:

* ~~First, wait for https://github.com/rustwasm/wasm-bindgen/pull/602 to merge~~
* Next, select a test module (for example `tests/all/foo.rs`) to migrate. Comment here so make sure work isn't duplicated!
* Move the file to `tests/wasm/foo.rs`
* Add `pub mod foo;` to `tests/wasm/main.rs`
* Copy all JS into `tests/wasm/foo.js`
* Migrate `#[test]` to `#[wasm_bindgen_test]`
* Copy all Rust directly into `#[wasm_bindgen_test]` functions, or leave them as separate exported functions
* Run `cargo test --target wasm32-unknown-unknown` until tests pass!

All tests currently live in their own namespace so there's a lot of name conflicts (lots of `class Foo` classes in JS). When copying out JS and Rust exports they may need to be renamed appropriately to avoid conflicting with other names throughout the test suite.

An example of how `tests/all/api.rs` was moved can be found at https://github.com/rustwasm/wasm-bindgen/pull/602

Tests to migrate:

* [x] `api.rs`
* [x] `char.rs`
* [x] `classes.rs`
* [x] `closures.rs`
* [ ] `comments.rs`
* [x] `dependencies.rs`
* [x] `duplicates.rs`
* [x] `enums.rs`
* [x] `import_class.rs`
* [x] `imports.rs`
* [x] `jsobjects.rs`
* [x] `math.rs`
* [x] `node.rs`
* [x] `non_debug.rs`
* [x] `non_wasm.rs`
* [x] `simple.rs`
* [x] `slice.rs`
* [x] `structural.rs`
* [ ] `typescript.rs`
* [x] `u64.rs`
* [x] `validate_prt.rs`

alexcrichton:
I'm working on `import_class.rs`

alexcrichton:
Done now!

PR #647, opened 2018-08-05
Port tests that use only basic features

afdw:
Part of #603.


alexcrichton:
Thanks so much for this!

PR #650, opened 2018-08-05
Enable structural mode for all operations if there is a Global extended attribute on the interface

afdw:


crates/web-sys/webidls/enabled/Window.webidl
```diff
@@ -25,7 +25,9 @@
 // interface XULControllers;
 
 // http://www.whatwg.org/specs/web-apps/current-work/
-[PrimaryGlobal, LegacyUnenumerableNamedProperties, NeedResolve]
```
alexcrichton: Would it be possible to recognize `PrimaryGlobal` instead of adding the modifications here perhaps?

alexcrichton:
Nice! Could a small test for `Window` and one of its properties also be added to ensure it works?

crates/web-sys/webidls/enabled/Window.webidl
afdw: `PrimaryGlobal` was removed from spec ([old](https://www.w3.org/TR/WebIDL-1/#PrimaryGlobal), [new](https://heycam.github.io/webidl/#Global)), I have just applied this commit: https://github.com/whatwg/html/commit/8c69161a98acea9eeef1e43e481ca3cc6287d393

afdw:
Currently test for WebIDL are run in Node, so adding a test for `Window` might be impossible. But I have added a test where the method can only be accessed on an actual instance of the class, not on the prototype.

chinedufn:
Not quite sure if it helps in this instance but you can set globals in Node.js -> https://github.com/chinedufn/percy/blob/69373a62ad1e4f4d512d6c11a3e8f5fc6b84fc2c/tests/jsdom/tests/test-utils.js#L16

afdw:
But it is `jsdom` , not a real browser.

It is actually possible to write a test for real browsers, but these will be `web-sys` tests, located here: https://github.com/rustwasm/wasm-bindgen/tree/b4601295d0c5908a70bf8e4c7249dda35491995d/crates/web-sys/tests/wasm

PR #652, opened 2018-08-05
Add support for getters, setters and deleters

### LINKED ISSUE #248: Support WebIDL getters and setters
https://heycam.github.io/webidl/#idl-special-operations

```webidl
interface Dictionary {
  readonly attribute unsigned long propertyCount;

  getter double (DOMString propertyName);
  setter void (DOMString propertyName, double propertyValue);
};
```

These are the moral equivalent of `Index` and `IndexMut` traits in Rust, except that those must return references and we can't support that (especially with fallible getters and setters). Maybe just translate these into `get` and `set` functions?

afdw:
Fixes #248.


alexcrichton:
Thanks! Would it be possible to add some tests as well? I wonder if the attribute as well could be something like webidl_setter and such? The name "special" doesn't necessarily imply much.

Finally, could some small docs be added to the guide for these attributes?

afdw:
It is not necessarily tied to WebIDL. It just means another syntax in JS (mode dynamic variants of getters and setters and the deletion syntax).

alexcrichton:
Hm true yeah! "dynamic" seems plausible (preferrable to "special" to me at least), we could also do something like "index", "index_set", "delete" or something like that to

afdw:
Maybe `IndexingGetter`, `IndexingSetter` and `IndexingDeleter`?

alexcrichton:
Sure yeah, internally I don't mind too much what we call these, I'm mostly worried for the name of the attribute which is part of the public API. Names seem fine by me though!

alexcrichton:
Thanks so much!

PR #657, opened 2018-08-06
Add `ArrayBuffer.prototype.byteLength` to js-sys

ctjhoa:


crates/js-sys/src/lib.rs
```diff
@@ -361,6 +361,16 @@ extern "C" {
...
     pub fn new(length: u32) -> ArrayBuffer;
 
+    /// The byteLength property of an object which is an instance of type ArrayBuffer
+    /// it's an accessor property whose set accessor function is undefined,
+    /// meaning that you can only read this property.
+    /// The value is established when the array is constructed and cannot be changed.
+    /// This property returns 0 if this ArrayBuffer has been detached.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/byteLength
+    #[wasm_bindgen(method, getter, structural, js_name = byteLength)]
```
alexcrichton: I think the `structural` part here can likely be removed?

alexcrichton:
Thanks!

crates/js-sys/src/lib.rs
ctjhoa: You're right. bad copy paste

PR #663, opened 2018-08-07
guide: add `indexing_{getter,setter,deleter}` attribute pages

fitzgen:
r? @afdw

### DIFF
--- a/guide/src/reference/attributes/on-js-imports/indexing-getter-setter-deleter.md
+++ b/guide/src/reference/attributes/on-js-imports/indexing-getter-setter-deleter.md
@@ -1,0 +1,82 @@
+# `indexing_getter`, `indexing_setter`, and `indexing_deleter`
+
+These three attributes indicate that a method is an dynamically intercepted
+getter, setter, or deleter on the receiver object itself, rather than a direct
+access of the receiver's properties. It is equivalent calling the Proxy handler
+for the `obj[prop]` operation with some dynamic `prop` variable in JavaScript,
+rather than a normal static property access like `obj.prop` on a normal
+JavaScript `Object`.
+
+This is useful for binding to `Proxy`s and some builtin DOM types that
+dynamically intercept property accesses.
+
+* `indexing_getter` corresponds to `obj[prop]` operation in JavaScript. The
+  function annotated must have a `this` receiver parameter, a single parameter
+  that is used for indexing into the receiver (`prop`), and a return type.
+
+* `indexing_setter` corresponds to the `obj[prop] = val` operation in
+  JavaScript. The function annotated must have a `this` receiver parameter, a
+  parameter for indexing into the receiver (`prop`), and a value parameter
+  (`val`).
+
+* `indexing_deleter` corresponds to `delete obj[prop]` operation in
+  JavaScript. The function annotated must have a `this` receiver and a single
+  parameter for indexing into the receiver (`prop`).
+
+These must always be used in conjunction with the `structural` and `method`
+flags.
+
+For example, consider this JavaScript snippet that uses `Proxy`:
+
+```js
+const foo = new Proxy({}, {
+    get(obj, prop) {
+        return prop in obj ? obj[prop] : prop.length;
+    },
+    set(obj, prop, value) {
+        obj[prop] = value;
+    },
+    deleteProperty(obj, prop) {
+        delete obj[prop];
+    },
+});
+
+foo.ten;
+// 3
+
+foo.ten = 10;
+foo.ten;
+// 10
+
+delete foo.ten;
+foo.ten;
+// 3
+```
+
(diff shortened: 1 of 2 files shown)


guide/src/reference/attributes/on-js-imports/indexing-getter-setter-deleter.md
```diff
@@ -0,0 +1,39 @@
+# `indexing_getter`, `indexing_setter`, and `indexing_deleter`
+
+These three attributes indicate that a method is a getter, setter, or deleter
+method on the receiver object itself, rather than one of the receiver's
+properties. This is useful for binding to special objects like `Array` and
+`NodeList` that override indexing operations.
```
afdw: Not sure about this part.
Firstly, normal `getter` and `setter` with `structural` can also access properties pretty much the same way, the main difference is that the name of the property with normal ones is always the same and specified in name of the method or in the value of the property, while in indexing is it dynamic and is specified in the argument. Would it be possible to describe the difference more clearly here?
Secondly, I do not think that are good example. `NodeList` definitely does not generate any of these 3 attributes, not sure about `Array`. I think just not giving an example can be fine, but it is possible to write JS code that corresponds to the Rust example (it would probably use `Proxy`, check out the test: https://github.com/rustwasm/wasm-bindgen/blob/9b65e57585c6ad334ac5fb3867656f3e734747a5/crates/webidl-tests/simple.js#L96-L108). Like this I think:
```js
new Proxy({}, {
    get(obj, prop) {
        return obj.hasOwnProperty(prop) ? obj[prop] : 0xFFFFFFFF;
    },
    set(obj, prop, value) {
        obj[prop] = value;
    },
    deleteProperty(obj, prop) {
        delete obj[prop];
    },
});
```

guide/src/reference/attributes/on-js-imports/indexing-getter-setter-deleter.md
```diff
@@ -0,0 +1,39 @@
...
+
+* `indexing_getter` corresponds to `obj[prop]` in JavaScript. The function
+  annotated must have a `this` receiver parameter and a single parameter that is
+  used for indexing into the receiver.
+
+* `indexing_setter` corresponds to `obj[prop] = val` in JavaScript. The function
+  annotated must have a `this` receiver parameter, a parameter for indexing into
+  the receiver, and a value parameter.
```
afdw: How about adding names have used in JS syntax? Like:
> a parameter for indexing into the receiver (`prop`), and a value parameter (`val`)

guide/src/reference/attributes/on-js-imports/indexing-getter-setter-deleter.md
```diff
@@ -0,0 +1,39 @@
+# `indexing_getter`, `indexing_setter`, and `indexing_deleter`
+
+These three attributes indicate that a method is a getter, setter, or deleter
+method on the receiver object itself, rather than one of the receiver's
+properties. This is useful for binding to special objects like `Array` and
+`NodeList` that override indexing operations.
+
+* `indexing_getter` corresponds to `obj[prop]` in JavaScript. The function
+  annotated must have a `this` receiver parameter and a single parameter that is
+  used for indexing into the receiver.
```
afdw: Maybe specify what the function should return here?

fitzgen:
Thanks for the notes @afdw! I've updated the commit, if you'd like to take another look.

guide/src/reference/attributes/on-js-imports/indexing-getter-setter-deleter.md
```diff
@@ -0,0 +1,57 @@
+# `indexing_getter`, `indexing_setter`, and `indexing_deleter`
+
+These three attributes indicate that a method is a getter, setter, or deleter
+method on the receiver object itself, rather than one of the receiver's
+properties. This is useful for binding to types that use `Proxy` to intercept
+property accesses.
```
afdw: > These three attributes indicate that a method is a getter, setter, or deleter method on the receiver object itself, rather than one of the receiver's properties.

This is still a bit confusing (and looks like it is not changed yet comparing to previous version).
I read this as, for example for setter, either "`object = value`, rather than `object.property = value`" or "`object.property = value`, rather than `object.property.anotherProperty = value`", both senses are wrong. The actual difference is in whether access is static or dynamic, otherwise they are pretty similar.

It is not only used for `Proxy`, that is just an example that can be implemented in JS, but its main usage is for built-in types in browsers. So something like this would probably be better:
> This is useful for binding to types that intercept property accesses (for example, some build-in types in browsers or objects that are wrapped with `Proxy`).

fitzgen:
@afdw, ok I updated some more and expanded the example code as well. I'm going to go ahead and merge this version, and if you have further improvements, please send follow up PRs!

afdw:
Yes, thanks, that already looks better.

PR #665, opened 2018-08-07
Support named "special" operations in WebIDL

alexcrichton:
This commit adds support for two different features of the "special" operations
in WebIDL. First, it implements the desugaring [described by WebIDL][1] where
this:

    interface Dictionary {
      getter double getProperty(DOMString propertyName);
      setter void setProperty(DOMString propertyName, double propertyValue);
    };

becomes ...

    interface Dictionary {
      double getProperty(DOMString propertyName);
      void setProperty(DOMString propertyName, double propertyValue);

      getter double (DOMString propertyName);
      setter void (DOMString propertyName, double propertyValue);
    };

where specifically a named `getter` generates both a getter and a named
function.

Second it implements the distinction between two different types of getters in
WebIDL, described as:

> Getters and setters come in two varieties: ones that take a DOMString as a
> property name, known as named property getters and named property setters, and
> ones that take an unsigned long as a property index, known as indexed property
> getters and indexed property setters.

The name `get` is given to DOMString arguments, and the name `get_idx` is given
to index property getters.

[1]: https://heycam.github.io/webidl/#idl-special-operations

### DIFF
--- a/crates/webidl/src/first_pass.rs
+++ b/crates/webidl/src/first_pass.rs
@@ -11,7 +11,7 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use weedle::argument::Argument;
 use weedle::attribute::ExtendedAttribute;
-use weedle::interface::StringifierOrStatic;
+use weedle::interface::{StringifierOrStatic, Special};
 use weedle::mixin::MixinMember;
 use weedle::namespace::NamespaceMember;
 use weedle;
@@ -61,7 +61,7 @@ pub(crate) struct NamespaceData<'src> {
     pub(crate) operations: BTreeMap<OperationId<'src>, OperationData<'src>>,
 }
 
-#[derive(PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, Copy)]
 pub(crate) enum OperationId<'src> {
     Constructor,
     Operation(Option<&'src str>),
@@ -171,7 +171,7 @@ fn first_pass_operation<'src>(
     record: &mut FirstPassRecord<'src>,
     first_pass_operation_type: FirstPassOperationType,
     self_name: &'src str,
-    id: OperationId<'src>,
+    ids: &[OperationId<'src>],
     arguments: &[Argument<'src>],
 ) -> Result<()> {
     let mut names = Vec::with_capacity(arguments.len());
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for taking a look! It's true yeah the main motivation for doing this at all was mostly for consistency in that we want to support `getter DOMString(DOMString name)` as well as `getter foo DOMString(DOMString name)` so this was applying the desugaring the spec recommended.

For handling name collisions I like your idea better though, I'll plumb these into the overloaded map.

alexcrichton:
Ok updated!

crates/webidl/src/first_pass.rs
```diff
@@ -124,18 +125,29 @@ fn first_pass_operation<'src>(
...
-        .or_insert(false);
+    let mut ids = vec![id];
+    for special in specials {
+        ids.push(match special {
+            Special::Getter(_) => OperationId::IndexingGetter,
+            Special::Setter(_) => OperationId::IndexingSetter,
+            Special::Deleter(_) => OperationId::IndexingDeleter,
+            Special::LegacyCaller(_) => continue,
+        });
+    }
```
afdw: I think moving code that handles specials here will not help reusing, as specials can be only on interfaces (not even on interface mixins). So it should be fine to just pass a vector of operations ids into this function?

alexcrichton: Good idea, updated!

afdw:
OK, so now there are mostly readability changes.

I am not sure why there is even a vector of specials instead of 1 special. [Grammar from the spec](https://heycam.github.io/webidl/#index-prod-SpecialOperation):
> ```
> SpecialOperation ::
>     Special RegularOperation
> 
> Special ::
>     getter
>     setter
>     deleter
> ```

But [here](https://heycam.github.io/webidl/#idl-operations) the spec says that there can be multiple keywords:
> ```webidl
> interface interface_identifier {
>   /* special_keywords... */ return_type identifier(/* arguments... */);
>   /* special_keywords... */ return_type (/* arguments... */);
> };
> ```
But I believe that it is an another mistake that is not fixed after https://github.com/heycam/webidl/pull/562 and https://github.com/heycam/webidl/pull/573.

alexcrichton:
Ah seems reasonable to fix upstream in weedle! You mentioned some readability changes though, perhaps a comment that wasn't submitted?

afdw:
No, I meant that now the diff on this PR is mostly changes that are aimed to increase the readability of the code.

alexcrichton:
Ah true yeah, it also handles though multiple `getter` methods with different types as well as the expansion of a getter+name into two methods

PR #666, opened 2018-08-08
Instanceof renamed import types

fitzgen:


crates/shared/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 #[macro_use]
```
alexcrichton: It looks like this file accidentally became executable?

I'm actually really curious, I've seen this happen from time to time, do you know what might cause it?

crates/shared/src/lib.rs
```diff
@@ -82,7 +82,8 @@ pub struct ImportStatic {
 
 #[derive(Deserialize, Serialize)]
 pub struct ImportType {
-    pub name: String,
+    pub rust_name: String,
```
alexcrichton: We actually don't need to change the name here I think, the generated JS bindings don't need to know about the rust name, right? (we just need to be sure to fill in the JS name when this struct is generated)

alexcrichton:
r=me, but looks like a web-sys test is failing?

crates/shared/src/lib.rs
fitzgen: Yeah, I've seen it too, and I have no idea why. It is possible a mistyped emacs key chord does it? I really don't know... I'm certainly not running `chmod +x`! :-p

crates/shared/src/lib.rs
fitzgen: Ah good idea, and that way we don't need to wonder if we have tracked down all the potential bits where we use the wrong version!

fitzgen:
> r=me, but looks like a web-sys test is failing?

Gah :( I couldn't run those tests because webdriver is hanging for me.

Also, it seems that the `WASM_BINDGEN_FIREFOX_BIN_PATH` var is not used anymore. I'll file some follow up issues.

Issue #670, opened 2018-08-08
Add `extends = ...` attributes to `js-sys` types


fitzgen:
We should add all the inheritance relations for global JS types to `js-sys`.

We can use this snippet of JS to log the prototype chain for something, eg `Uint8Array`:

```js
function logPrototypeChain(cls) {
  let obj = cls.prototype;
  do {
    console.log(obj.constructor.name);
    obj = obj.__proto__;
  } while (obj != null);
}

logPrototypeChain(Uint8Array);
```

Pasting that snippet in a browser's devtools console logs:

```text
Uint8Array
TypedArray
Object
```

So we would add `#[wasm_bindgen(extends = Object, extends = TypedArray)` to `Uint8Array`'s import definition and a corresponding test to the appropriate file in `crates/js-sys/tests/..` like:

```rust
use wasm_bindgen::JsCast;

#[wasm_bindgen_test]
fn test_blah_inheritance() {
    let blah = js_sys::Blah::new(); // its constructor might require different args

    // Assert JS actually agrees with our set of base classes.
    assert!(blah.is_instance_of::<Blah>());
    assert!(blah.is_instance_of::<BlahBase>());
    // etc for each base class...
    assert!(blah.is_instance_of::<js_sys::Object>());

    // Ensure that we generated the appropriate `AsRef` upcasting implementations
    // for each base class.
    let _: &BlahBase = blah.as_ref();
    // etc for each base class...
    let _: &js_sys::Object = blah.as_ref();
}
```

Here is an example PR for the `RegExp` type: https://github.com/rustwasm/wasm-bindgen/pull/671

See also `js-sys` contributing documentation: https://rustwasm.github.io/wasm-bindgen/js-sys.html

-------------------------------------------------------

Here are all the JS global import types for which we need to add this information:

* [x] `Array`
* [x] `ArrayBuffer`
* [x] `Boolean`
* [x] `DataView`
* [x] `Date`
* [x] `Error`
* [x] `EvalError`
* [x] `Float32Array`
* [x] `Float64Array`
* [x] `Function`
* [x] `Generator`
* [x] `Int16Array`
* [x] `Int32Array`
* [x] `Int8Array`
* [x] `Intl.Collator`
* [x] `Intl.DateTimeFormat`
* [x] `Intl.NumberFormat`
* [x] `Intl.PluralRules`
* [x] `JSON`
* [x] `Map`
* [x] `Math`
* [x] `Number`
* [x] `Promise`
[...]

PR #672, opened 2018-08-09
js-sys: Add extends attributes for js_sys::Number

bokuweb:
Part of #670

### DIFF
--- a/crates/js-sys/tests/wasm/Number.rs
+++ b/crates/js-sys/tests/wasm/Number.rs
@@ -1,8 +1,16 @@
 use std::f64::{INFINITY, NAN};
 
+use wasm_bindgen::JsCast;
 use wasm_bindgen::JsValue;
 use wasm_bindgen_test::*;
 use js_sys::*;
+
+#[wasm_bindgen_test]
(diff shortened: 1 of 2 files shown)


crates/js-sys/tests/wasm/Number.rs
```diff
@@ -1,9 +1,17 @@
 use std::f64::{INFINITY, NAN};
 
+use wasm_bindgen::JsCast;
 use wasm_bindgen::JsValue;
 use wasm_bindgen_test::*;
 use js_sys::*;
 
+#[wasm_bindgen_test]
+fn number_inheritance() {
+    let number: i32 = 10;
```
fitzgen: This line should be:

```rust
let number = Number::new(&JsValue::from(10));
```

Note that if you open a JS console:

```
> 10 instanceof Number
false
> new Number(10) instanceof Number
true
```

JS has a distinction between `Number` objects and its number values, which is a bit tricky!

bokuweb:
Thanks for your quick response :) On this occasion I will learn about `wsm-bindgen`.

PR #674, opened 2018-08-09
Add extends attributes for several types

eminence:
Part of #670 

Marking as WIP because I wasn't able to run tests locally.  If/when travis CI finishes successfully on this PR, feel free to merge


Pauan:
@eminence Symbol is a value type (like boolean, string, and number), so you have to detect it like this:

```js
typeof x === "symbol"
```

fitzgen:
Will merge when CI comes back green.

eminence:
I think something went wrong during the merge.  I just pushed a fixup

eminence:
We are green!

fitzgen:
Thanks @eminence :)

PR #677, opened 2018-08-09
[670] add extends for ArrayBuffer

sepiropht:


fitzgen:
Thanks for the pull request @sepiropht; unfortunately `ArrayBuffer` was already added in #674 so we won't merge this PR. There are other types that need the `extends` attribute still, however!

PR #678, opened 2018-08-09
Add support webidl namespaces.

derekdreery:
I'm opening this early to get feedback.

This PR aims to get namespaces supported, so you can do

```rust
console::log("string")
```

So far it can generate code from the AST, which I reconstructed from the following code snippet which I checked compiles correctly.

```rust
extern crate wasm_bindgen;

#[no_mangle]
#[allow(non_snake_case)]
#[doc(hidden)]
pub extern "C" fn __wbindgen_describe___wbg_log_894f9ca062837e26() {
    use wasm_bindgen::describe::*;
    inform(FUNCTION);
    inform(1u32);
    <&str as WasmDescribe>::describe();
    inform(0);
}

mod console {
#[allow(bad_style)]
#[doc = ""]
pub fn log(s: &str) {
    ::wasm_bindgen::__rt::link_mem_intrinsics();
    extern "C" {
        fn __wbg_log_894f9ca062837e26(s: <&str as ::wasm_bindgen::convert::IntoWasmAbi>::Abi)
            -> ();
    }
    unsafe {
        let _ret = {
            let mut __stack = ::wasm_bindgen::convert::GlobalStack::new();
            let s = <&str as ::wasm_bindgen::convert::IntoWasmAbi>::into_abi(s, &mut __stack);
            __wbg_log_894f9ca062837e26(s)
        };
        ()
    }
}
#[allow(bad_style, unused_variables)]
#[cfg(not(all(target_arch = "wasm32", not(target_os = "emscripten"))))]
#[doc = ""]
pub fn log(s: &str) {
    panic!(
        "cannot call wasm-bindgen imported functions on \
         non-wasm targets"
    );
}
}

#[allow(non_upper_case_globals)]
#[link_section = "__wasm_bindgen_unstable"]
#[doc(hidden)]
pub static __WASM_BINDGEN_GENERATED_50eb08d47552231d :
[ u8 ; 283usize ] = *
b"\x17\x01\x00\x00{\"exports\":[],\"enums\":[],\"imports\":[{\"module\":null,\"js_namespace\":\"console\",\"kind\":{\"kind\":\"function\",\"shim\":\"__wbg_log_894f9ca062837e26\",\"catch\":false,\"method\":null,\"structural\":false,\"function\":{\"name\":\"log\"}}}],\"structs\":[],\"version\":\"0.2.15 (5b935526f)\",\"schema_version\":\"8\"}";

pub fn greet(name: &str) {
    console::log(&format!("Hello, {}!", name));
}
#[export_name = "greet"]
#[allow(non_snake_case)]
pub extern "C" fn __wasm_bindgen_generated_greet(
    arg0: <str as ::wasm_bindgen::convert::RefFromWasmAbi>::Abi,
) {
    ::wasm_bindgen::__rt::link_mem_intrinsics();
    let _ret = {
        let mut __stack = unsafe { ::wasm_bindgen::convert::GlobalStack::new() };
        let arg0 = unsafe {
            <str as ::wasm_bindgen::convert::RefFromWasmAbi>::ref_from_abi(arg0, &mut __stack)
        };
        let arg0 = &*arg0;
        greet(arg0)
    };
}
#[no_mangle]
#[allow(non_snake_case)]
#[doc(hidden)]
pub extern "C" fn __wbindgen_describe_greet() {
    use wasm_bindgen::describe::*;
    inform(FUNCTION);
    inform(1u32);
    <&str as WasmDescribe>::describe();
    inform(0);
}
#[allow(non_upper_case_globals)]
#[link_section = "__wasm_bindgen_unstable"]
#[doc(hidden)]
pub static
__WASM_BINDGEN_GENERATED_bbc0315d281cc3c3 : [ u8 ; 214usize ] = *
b"\xd2\x00\x00\x00{\"exports\":[{\"class\":null,\"method\":false,\"consumed\":false,\"constructor\":null,\"function\":{\"name\":\"greet\"},\"comments\":[]}],\"enums\":[],\"imports\":[],\"structs\":[],\"version\":\"0.2.15 (5b935526f)\",\"schema_version\":\"8\"}"
;
```

### DIFF
--- a/crates/webidl/src/first_pass.rs
+++ b/crates/webidl/src/first_pass.rs
@@ -350,7 +350,7 @@ impl<'src> FirstPass<'src, ()> for weedle::NamespaceDefinition<'src> {
             .namespaces
             .entry(self.identifier.0)
             .and_modify(|entry| entry.partial = false)
-            .or_insert_with(Namespace::empty_non_partial);
+            .or_insert_with(NamespaceData::empty_non_partial);
 
         if util::is_chrome_only(&self.attributes) {
             return Ok(())
--- a/crates/webidl/src/first_pass.rs
+++ b/crates/webidl/src/first_pass.rs
@@ -41,17 +41,6 @@ pub(crate) struct InterfaceData<'src> {
     pub(crate) superclass: Option<&'src str>,
 }
 
-impl<'src> Default for InterfaceData<'src> {
-    fn default() -> Self {
-        InterfaceData {
-            partial: true,
-            global: false,
-            operations: Default::default(),
-            superclass: Default::default(),
-        }
-    }
-}
-
 /// We need to collect namespace data during the first pass, to be used later.
 #[derive(Default)]
 pub(crate) struct NamespaceData<'src> {
@@ -60,9 +49,10 @@ pub(crate) struct NamespaceData<'src> {
     pub(crate) operations: BTreeMap<Option<&'src str>, OperationData<'src>>,
 }
 
-impl<'src> Default for NamespaceData<'src> {
-    fn default() -> Self {
-        NamespaceData {
+impl<'src> NamespaceData<'src> {
+    /// Same as `Default::default` but sets `partial` to true.
+    pub(crate) fn default_for_partial() -> Self {
+        Self {
             partial: true,
             operations: Default::default(),
         }
@@ -178,7 +168,7 @@ impl<'src> FirstPass<'src, ()> for weedle::InterfaceDefinition<'src> {
             let interface = record
                 .interfaces
                 .entry(self.identifier.0)
-                .or_insert_with(Default::default);
+                .or_default();
             interface.partial = false;
             interface.superclass = self.inheritance.map(|s| s.identifier.0);
         }
@@ -344,6 +334,93 @@ impl<'src> FirstPass<'src, ()> for weedle::TypedefDefinition<'src> {
             warn!("Encountered multiple declarations of {}", self.identifier.0);
         }
 
+        Ok(())
+    }
+}
+
+impl<'src> FirstPass<'src, ()> for weedle::NamespaceDefinition<'src> {
+    fn first_pass(&'src self, record: &mut FirstPassRecord<'src>, (): ()) -> Result<()> {
+        {
+            let namespace = record
+                .namespaces
+                .entry(self.identifier.0)
+                .or_default();
+            namespace.partial = false;
+        }
+
+        if util::is_chrome_only(&self.attributes) {
+            return Ok(())
+        }
+
+        // We ignore all attributes.
+
+        for member in &self.members.body {
+            member.first_pass(record, self.identifier.0)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl<'src> FirstPass<'src, ()> for weedle::PartialNamespaceDefinition<'src> {
+    fn first_pass(&'src self, record: &mut FirstPassRecord<'src>, (): ()) -> Result<()> {
+        record
+            .namespaces
+            .entry(self.identifier.0)
+            .or_insert_with(NamespaceData::default_for_partial);
+
+        if util::is_chrome_only(&self.attributes) {
+            return Ok(())
+        }
+
+        for member in &self.members.body {
+            member.first_pass(record, self.identifier.0)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl<'src> FirstPass<'src, &'src str> for weedle::namespace::NamespaceMember<'src> {
+    fn first_pass(&'src self,
+                  record: &mut FirstPassRecord<'src>,
+                  namespace_name: &'src str) -> Result<()>
+    {
+        match self {
+            weedle::namespace::NamespaceMember::Operation(op) => {
+                op.first_pass(record, namespace_name)
+            }
+            _ => Ok(()),
+        }
+    }
+}
+
+impl<'src> FirstPass<'src, &'src str> for weedle::namespace::OperationNamespaceMember<'src> {
+    fn first_pass(&'src self,
+                  record: &mut FirstPassRecord<'src>,
+                  namespace_name: &'src str) -> Result<()>
+    {
+        let identifier = self.identifier.map(|s| s.0);
+        let arguments = &self.args.body.list;
+        let mut names = Vec::with_capacity(arguments.len());
+        for argument in arguments {
+            match argument {
+                Argument::Single(arg) => names.push(arg.identifier.0),
+                Argument::Variadic(_) => return Ok(()),
+            }
+        }
+        record
+            .namespaces
+            .get_mut(namespace_name)
+            .unwrap() // call this after creating the namespace
+            .operations
+            .entry(identifier)
+            .and_modify(|operation_data| operation_data.overloaded = true)
+            .or_insert_with(Default::default)
+            .argument_names_same
+            .entry(names)
+            .and_modify(|same_argument_names| *same_argument_names = true)
+            .or_insert(false);
         Ok(())
     }
 }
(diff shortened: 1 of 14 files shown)


derekdreery:
I'm planning to write tests once I've done the webidl-to-backend bit (which I've *just* started), I wouldn't expect you to accept this without tests. However, I think it's hard to just test the codegen without manually creating AST in the test - once the webidl bit is working we can test both at the same time.

fitzgen:
Yep, sounds good to me!

fitzgen:
Let me/alex/ohanar know when you're ready for another round of review :)

derekdreery:
There is another option here:

The webidl suggests that the console methods are pseudo-static methods on the Console type. So instead of generating a module, we could generate a `Console` struct and add the methods there. Note sure which is better.

fitzgen:
I had written out a comment about how I prefer modules, but then I discovered this in the webidl spec:

> A namespace is a specification of a set of namespace members (matching NamespaceMembers), which are the regular operations and read only regular attributes that appear between the braces in the namespace declaration. These operations and attributes describe the behaviors packaged into the namespace.
>
> [...]
>
> ```ebnf
> Namespace ::
>     namespace identifier { NamespaceMembers } ;
> 
> NamespaceMembers ::
>     ExtendedAttributeList NamespaceMember NamespaceMembers
>     ε
> 
> NamespaceMember ::
>     RegularOperation
>     readonly AttributeRest
> ```

and verified that our webidls only have operations and attributes within namespaces. That means we don't need to worry about nested types/interfaces/mixins. Which means that the first and second options are looking a lot more tempting, and require much less invasive changes to wasm-bindgen.

Additionally, if you carefully read [these steps on creating an operation](https://heycam.github.io/webidl/#dfn-create-operation-function) you'll find that namespace operations do not get the namespace object passed as a `this` value. This doesn't matter in the end because `static_method_of = ...` will bind the static method to its class anyways, and if we used a `static` object like `static console; console.log(..);` we would be invoking the method with a `this` receiver as well.

All of that said, I am partial to using static methods, the equivalent of

```rust
#[wasm_bindgen]
extern {
    type console;

    #[wasm_bindgen(static_method_of = console)]
    fn log(...);
}

console::log(...);
```

because it matches what `js-sys` does with `Math`.

afdw:
There is a problem with naming here though: while `Math` is not a bad name for a Rust type, `console` is, as it starts with a lowercase letters. It is a bit unconventional.

ohanar:
The main problem with using static methods rather than modules (at least as I see it), is that you can't import a static method into your local namespace. I.e.
```rust
pub mod console_module {
    pub fn log(...) { ... }
}
// vs
pub struct ConsoleType;
impl ConsoleType {
    pub fn log(...) { ... }
}

use console_module::log; // this works
use ConsoleType::log; // this doesn't
```

derekdreery:
I've almost finished implementing this as per the module version, (finished modulo bugs/errors), and anyone who wants to change over to one of the other variants will find it much easier after this PR has merged (since you can just go in and edit the AST transformation from webidl to backend), so I would consider punting on the decision and seeing what people's experience of the current version is.

The reason I've stopped is I'm getting the following error

```
error[E0308]: mismatched types
   --> crates/webidl/src/lib.rs:895:32
    |
895 |             js_namespace: Some(ns_names.js_name.clone()),
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `proc_macro2::Ident`, found &str
    |
    = note: expected type `proc_macro2::Ident`
               found type `&str`

```

which is easy to solve, but I was wondering if an `Ident` is the right type for the `js_namespace`, since it is only used on the javascript side. Unless the valid ident rules are the same in javascript and rust (I don't know if they are or not), we might as well just use a &str, because `Ident` suggests that it has meaning in rust, which it doesn't.

As always, I love to find out I'm wrong, so tell me if I am :).

Also please don't merge this until I have added tests, and implemented partial namespaces (easy because I already did all the first_pass stuff).

derekdreery:
Ok @fitzgen @alexcrichton ready for review :).

ohanar:
I have no issue running your altered hello_world example. I remember when I first started looking at this project, I was getting a bunch of validation errors, that eventually went away when I restarted my system. I never really figured out what was causing the issue.

crates/webidl/src/first_pass.rs
```diff
@@ -312,6 +338,93 @@ impl<'src> FirstPass<'src, ()> for weedle::TypedefDefinition<'src> {
...
+            }
+        }
+        record
+            .namespaces
+            .get_mut(namespace_name)
+            .unwrap() // call this after creating the namespace
+            .operations
+            .entry(identifier)
+            .and_modify(|operation_data| operation_data.overloaded = true)
+            .or_insert_with(Default::default)
```
ohanar: You could just use `or_default` here.

crates/webidl/src/first_pass.rs
```diff
@@ -40,6 +41,24 @@ pub(crate) struct InterfaceData<'src> {
...
 
+/// We need to collect namespace data during the first pass, to be used later.
+#[derive(Default)]
+pub(crate) struct NamespaceData<'src> {
+    /// Whether only partial namespaces were encountered
+    pub(crate) partial: bool,
+    pub(crate) operations: BTreeMap<Option<&'src str>, OperationData<'src>>,
+}
+
+impl<'src> NamespaceData<'src> {
```
ohanar: Why not just make this the implementation for `Default`?

crates/webidl/src/lib.rs
```diff
@@ -783,3 +792,149 @@ impl<'src> WebidlParse<'src, &'src str> for weedle::interface::ConstMember<'src>
...
+        &'src self,
+        program: &mut backend::ast::Program,
+        first_pass: &FirstPassRecord<'src>,
+        (): (),
+    ) -> Result<()> {
+        if util::is_chrome_only(&self.attributes) {
+            return Ok(());
+        }
+
+        let rust_name = rust_ident(self.identifier.0.to_snake_case().as_str());
```
ohanar: This is repeated a lot. Maybe make a `snake_case_ident` function in utils? (It would match the `camel_case_ident` function`.)

derekdreery:
@ohanar I've tried updating everything (rust, npm, cargo) and building from scratch, and I still get the offest error. Very strange, and unfortunate since I can't debug what I've done.

crates/webidl/src/first_pass.rs
derekdreery: The reason it's the way it currently is is that I felt implementing `Default` is misleading - there isn't an 'obvious' default value for the `partial` field. I thought that having separate methods helped convey meaning.

In fact, if I had followed my own plan I would not implement Default at all, I'd have 2 methods for partial and non-partial. I'll change to this so you can see what I mean, if you don't like it I can change it to just the Default :).

crates/webidl/src/lib.rs
derekdreery: I think the camel_case_ident function exists to handle cases like `HTMLBR` where the generic converter cannot deduce there should be a space, like `HTML_BR`.

derekdreery:
I'm not sure what's causing the build error now - it's failing in chrome but not firefox.

ohanar:
So it seems like the issue you are running up against is some upstream bug (see https://github.com/webpack/webpack/issues/7760 and https://github.com/xtuc/webassemblyjs/issues/407).

As for tests, you could add something like this to `webidl-test`:
```webidl
namespace Math {
    double pow(double base, double exponent);
}
```
And then assert some things with it in a corresponding rust file.

derekdreery:
yep using webpack from `git://github.com/webpack/webpack.git#b310b9b45c7e8a60c40d13dae967f74e712ec224` fixed it and I get the output I expect!

I'll work on some tests.

derekdreery:
I've added some tests - I'm not sure they ran though because when I typed `cargo test -p wasm-bindgen-webidl`, it said "running 0 tests".

afdw:
Try `cargo test -p webidl-tests --target wasm32-unknown-unknown`.

derekdreery:
Should be ready for review again.

ohanar:
Looks good!

PR #680, opened 2018-08-09
Js sys promise arguments by shared ref

fitzgen:
Not a breaking change because we haven't published js-sys with `Promise` bindings yet, thankfully.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1868,24 +1868,28 @@ impl<'a, 'b> SubContext<'a, 'b> {
                                 ),
                             }
                         } else {
-                            let location = if *is_static { "" } else { ".prototype" };
+                            let (location, binding) = if *is_static {
+                                ("", format!(".bind({})", class))
+                            } else {
+                                (".prototype", "".into())
+                            };
 
                             match kind {
                                 shared::OperationKind::Regular => {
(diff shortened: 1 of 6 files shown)


alexcrichton:
💯

crates/cli-support/src/js/mod.rs
```diff
@@ -1868,24 +1868,28 @@ impl<'a, 'b> SubContext<'a, 'b> {
                                 ),
                             }
                         } else {
-                            let location = if *is_static { "" } else { ".prototype" };
+                            let (location, binding) = if *is_static {
+                                ("", format!(".bind({})", class))
```
alexcrichton: This runs a mild risk in the future of not being quite as fast as it's creating a new JS function which may bypass optimizations if you hook up a C++ function directly, but I think we can cross that bridge when we get there

fitzgen: I know SpiderMonkey (and I think v8 does this too) will understand bound functions and treat them as calls to the original thing rather than inserting any kind of trampoline. My hope is that this means we aren't nullifying and potential host bindings good-ness, but agreed that we can cross that bridge when we get to it.

alexcrichton: Oh nice! That means this'll work perfectly as is I believe

PR #681, opened 2018-08-09
[670] add extends for some types

sepiropht:


crates/js-sys/tests/wasm/Symbol.rs
```diff
@@ -107,3 +107,11 @@ fn value_of() {
     let a = gensym(JsValue::undefined());
     assert_eq!(JsValue::from(a.value_of()), JsValue::from(a));
 }
+
+#[wasm_bindgen_test]
+fn symbol_inheritance() {
```
fitzgen: And I think this was accidentally included?

crates/js-sys/tests/wasm/JsString.rs
```diff
@@ -261,3 +262,11 @@ fn value_of() {
     let greeting = JsString::from("Hello world!");
     assert_eq!(greeting.value_of(), "Hello world!");
 }
+
+#[wasm_bindgen_test]
+fn js_string_inheritance() {
```
fitzgen: This needs the corresponding `extends = ...` on the `JsString` type import.

crates/js-sys/tests/wasm/Symbol.rs
sepiropht: No, It wasn't accidentally. Symbol is like other, isn't it ?
It has 2 Constructor which it inherit from : Symbol and Object

crates/js-sys/tests/wasm/JsString.rs
```diff
@@ -261,3 +262,10 @@ fn value_of() {
...
+    let string = JsString::from("Hello world!");
+    assert!(string.is_instance_of::<JsString>());
```
sepiropht: I have added extend for String type but test still don't pass. What i 'm missing ?

fitzgen: Similar to `Number`, JS has both string values and `String` objects:

strings

Despite string values not being an object, `String.prototype == Object.getPrototypeOf("")` is true and all the prototype methods exist on string values.

As far as moving forward with this PR goes, we can have a `JsString.js` file that exports a function that constructs a new `String` object:

```js
exports.new_string_object = () => new String("hi");
```

and import that in the `JsString.rs` test, and use that for the `is_instance_of` test.

sepiropht: Whoa ! Good to know. Thank for the reply.

alexcrichton:
Thanks @sepiropht!

PR #683, opened 2018-08-10
Add initial support and tests for JSON

eminence:
Not yet added is the optional arguments to `JSON.parse()` and `JSON.stringify()`.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -2663,6 +2663,31 @@ extern "C" {
     pub fn validate(bufferSource: &JsValue) -> Result<bool, JsValue>;
 }
 
+// JSON
+#[wasm_bindgen]
+extern "C" {
+
+    #[derive(Clone, Debug)]
+    pub type JSON;
+
+    /// The `JSON.parse()` method parses a JSON string, constructing the
+    /// JavaScript value or object described by the string.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
+    #[wasm_bindgen(catch, static_method_of = JSON)]
+    pub fn parse(text: &str) -> Result<JsValue, JsValue>;
+
+    /// The JSON.stringify() method converts a JavaScript value to a JSON string,
+    /// optionally replacing values if a replacer function is specified or
+    /// optionally including only the specified properties if a replacer array is
+    /// specified.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
+    #[wasm_bindgen(static_method_of = JSON)]
+    pub fn stringify(obj: &JsValue) -> JsString;
+
+}
+
 // JsString
 #[wasm_bindgen]
 extern "C" {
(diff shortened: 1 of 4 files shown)


eminence:
I'm looking at the optional args to `JSON.stringify()`, and I'm not totally sure how to design the API.  There are 2 optional arguments, and the second argument can take two types (function or array).  If you expand all the combos into different functions, you get this big list:

* `stringify(obj: &JsValue) -> JsString`
* `stringify_replacer_func(obj: &JsValue, replacer_function: &mut FnMut(JsValue, JsValue) -> JsValue)) -> JsString`
* `stringify_replacer_array(obj: &JsValue, replacer_array: &Array) -> JsString`
* `stringify_replacer_func_spacer(obj: &JsValue, replacer_function: &mut FnMut(JsValue, JsValue) -> JsValue, spacer: &JsValue)) -> JsString`
* `stringify_replacer_array_spacer(obj: &JsValue, replacer_array: &Array, spacer: &JsValue) -> JsString`

Is there a better way to do this?

fitzgen:
> Is there a better way to do this?

Not really, at the moment, unfortunately.

crates/js-sys/src/lib.rs
```diff
@@ -2663,6 +2663,31 @@ extern "C" {
...
+    pub fn parse(text: &str) -> Result<JsValue, JsValue>;
+
+    /// The JSON.stringify() method converts a JavaScript value to a JSON string,
+    /// optionally replacing values if a replacer function is specified or
+    /// optionally including only the specified properties if a replacer array is
+    /// specified.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
+    #[wasm_bindgen(static_method_of = JSON)]
+    pub fn stringify(obj: &JsValue) -> JsString;
```
fitzgen: Unfortunately, `JSON.stringify` can throw errors, even without the replacer functions:

json-lol

Therefore, we should add the `catch` attribute and return `Result<JsString, JsValue>` here.

eminence:
Ahh, good catch about errors from stringify.  I pushed a fix for that, plus a test.  Note:  I couldn't figure out how to add properties to an `Object`, so I wrote a small javascript helper to do this.  If there's a better way to do this in pure-rust, let me know.

Regarding the other optional args, I'll prepare another PR with support and tests for those.

crates/js-sys/tests/wasm/JSON.rs
```diff
@@ -63,7 +70,22 @@ fn stringify() {
...
     assert_eq!(rust_str, "[1,true,\"hello\"]");
-}
\ No newline at end of file
+}
+
+#[wasm_bindgen_test]
+fn stringify_error() {
+    let func = Function::new_no_args("throw new Error(\"rust really rocks\")");
+    let obj = Object::new();
+    set_in_object(&obj, "toJSON", &JsValue::from(func));
```
fitzgen: This should do it:

```rust
let prop = JsValue::from("toJSON");
Reflect::set(obj.as_ref(), prop.as_ref(), func.as_ref());
```

eminence: Great, thanks!  Pushed 1 more commit that fixes this

PR #685, opened 2018-08-10
A few more js sys things

fitzgen:


crates/js-sys/src/lib.rs
```diff
@@ -250,6 +250,40 @@ extern "C" {
...
+    /// with actual undefined values).
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of
+    ///
+    /// # Notes
+    ///
+    /// There are a few bindings to `of` in `js-sys`: `of1`, `of2`, etc...
+    /// with different arities.
+    #[wasm_bindgen(static_method_of = Array, js_name = of)]
+    pub fn of1(a: &JsValue) -> Array;
```
sendilkumarn: can we have a converter (with annotation) that converts the variable args into an vector and use that here?

PR #687, opened 2018-08-11
Misc fixes and enable tests on appveyor

eminence:
The first commit removes an unused import, which reduces the amount of warning spam on build.

The second commit was needed to get the tests to build on my windows machine (which uses back-slashes in paths).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,6 +22,11 @@ test_script:
   - npm ci
   - cargo test
   - cargo build --release -p wasm-bindgen-cli
+  - where chromedriver
+  - set CHROMEDRIVER=C:\Tools\WebDriver\chromedriver.exe
+  - cargo test -p js-sys --target wasm32-unknown-unknown
+  - cargo test -p web-sys --target wasm32-unknown-unknown
+  - cargo test -p webidl-tests --target wasm32-unknown-unknown
 
 branches:
   only:
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -543,7 +543,7 @@ impl ToTokens for ast::ImportType {
             const #const_name: () = {
                 use wasm_bindgen::convert::{IntoWasmAbi, FromWasmAbi, Stack};
                 use wasm_bindgen::convert::{OptionIntoWasmAbi, OptionFromWasmAbi};
-                use wasm_bindgen::convert::{RefFromWasmAbi, GlobalStack};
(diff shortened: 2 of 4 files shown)


eminence:
Opps.  Turns out that import was actually used.  Trying again with an `allow(unused_imports)` instead.

alexcrichton:
Thanks! Could some tests be added as well?

eminence:
I didn't add any new features, but sure.  What did you have in mind?

ohanar:
The appveyor script is pretty minimal right now, it would be good to have at least the `js-sys` and `webidl-tests` crates tested on there.

alexcrichton:
@eminence oh primarily just tests to ensure that the behavior doesn't regress accidentally in the future through a refactor accidentally gone wrong. 

For example I'm not sure what generates the unused imports, but having a windows-specific tests with a Windows-specific path would be good to have!

eminence:
Ok!  So I don't know how to add a test for unused imports (which wasn't a problem, just a small annoyance), but I took @ohanar's suggestion and enabled from more tests on Appveyor.

Appveyor already includes geckodriver and chromedriver as pre-installed software on its image, so this was pretty easy.  Unfortunately, I wasn't able to get geckodriver to work, so this PR only uses chromedriver.  (Any experts can review the geckodriver failure logs [here](https://ci.appveyor.com/project/eminence/wasm-bindgen/build/1.0.4)).

.appveyor.yml
```diff
@@ -22,6 +22,13 @@ test_script:
   - npm ci
   - cargo test
   - cargo build --release -p wasm-bindgen-cli
+  - where chromedriver
+  - set CHROMEDRIVER=C:\Tools\WebDriver\chromedriver.exe
+  - cargo test -p js-sys --target wasm32-unknown-unknown
+  - cargo test -p web-sys
+  - cargo test -p web-sys --target wasm32-unknown-unknown
+  - cargo test -p webidl-tests
```
ohanar: The web-sys and webidl-tests on the native platform are unnecessary, and just take extra time building. These crates only have tests for `wasm32-unknown-unknown`.

alexcrichton:
@eminence oh I think I just ran across the `GlobalStack` unused import issue, which is a legit but I think we need to fix. In that sense could the unused import be fixed instead of ignored?

.appveyor.yml
eminence: Great.  I suspected that was the case, but thanks for confirming!

alexcrichton:
Looks great, thanks!

PR #691, opened 2018-08-12
Add bindings for Intl.DateTimeFormat

thomaseizinger:
This PR adds bindings for `Intl.DateTimeFormat`.


thomaseizinger:
Ups, copy-pasta error :see_no_evil: 
I did run "the" tests. That I added and they passed. That was not enough apparently :D
Already fixed it. Fingers crossed that they pass this time :)

fitzgen:
> I have the following error locally:

Make sure you have node 10, as bigints are only available in >= 10. Personally, I recommend using `nvm`, which is sort of like `rustup` but for node.

PR #692, opened 2018-08-13
Add bindings for static method String.from_char_code

quelledanielle:
Seems like this is the first static binding with a different `js_class` name?

Not sure if these are most appropriate decisions:
- Added multiple bindings with different arities
- Used `f64` as the argument type since `char_code_at` returns `f64`

(Note: I'm getting failures for `String.prototype.trimStart does not exist` and `.trimEnd`)

@fitzgen

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -2940,6 +2940,34 @@ extern "C" {
     #[wasm_bindgen(method, js_class = "String", js_name = endsWith)]
     pub fn ends_with(this: &JsString, search_string: &str, length: i32) -> bool;
 
+    /// The static String.fromCharCode() method returns a string created from
+    /// the specified sequence of UTF-16 code units.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode
+    ///
+    /// # Notes
+    ///
+    /// There are a few bindings to `from_char_code` in `js-sys`: `from_char_code1`, `from_char_code2`, etc...
+    /// with different arities.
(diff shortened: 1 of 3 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -2940,6 +2940,34 @@ extern "C" {
...
+    /// the specified sequence of UTF-16 code units.
+    ///
+    /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode
+    ///
+    /// # Notes
+    ///
+    /// There are a few bindings to `from_char_code` in `js-sys`: `from_char_code1`, `from_char_code2`, etc...
+    /// with different arities.
+    #[wasm_bindgen(static_method_of = JsString, js_class = "String", js_name = fromCharCode)]
+    pub fn from_char_code1(a: f64) -> JsString;
```
fitzgen: These should be using `u32` parameter types.

`char_code_at` returns `NaN` for out of bounds indices, which is why it returns `f64`, but we don't have that issue here.

crates/js-sys/tests/wasm/JsString.rs
```diff
@@ -58,6 +58,19 @@ fn ends_with() {
     assert_eq!(js.ends_with("to be", 19), true);
 }
 
+#[wasm_bindgen_test]
+fn from_char_code() {
+    let s = "½+¾=";
```
fitzgen: Great test, glad to see some non-ascii chars coverage in here.

quelledanielle: Thanks, MDN 😄

PR #702, opened 2018-08-14
Implement support for WebIDL dictionaries

### LINKED ISSUE #241: Support creating bindings for WebIDL dictionaries
https://heycam.github.io/webidl/#idl-dictionaries

Dictionaries are passed by-value, and are structural rather than nominal (aka "duck typed").

For passing them from Rust to an imported function, we can translate dictionaries from

```c
dictionary SomeDict {
    bool on;
    long which;
}
```

into the AST equivalent of

```rust
#[wasm_bindgen(js_name = __SomeDict)]
pub struct SomeDict {
    pub on: bool,
    pub i32: which,
}
```

alexcrichton:
This commit adds support for generating bindings for dictionaries defined in
WebIDL. Dictionaries are associative arrays which are simply objects in JS with
named keys and some values. In Rust given a dictionary like:

    dictionary Foo {
        long field;
    };

we'll generate a struct like:

    pub struct Foo {
        obj: js_sys::Object,
    }

    impl Foo {
        pub fn new() -> Foo { /* make a blank object */ }

        pub fn field(&mut self, val: i32) -> &mut Self {
            // set the field using `js_sys::Reflect`
        }
    }

    // plus a bunch of AsRef, From, and wasm abi impls

At the same time this adds support for partial dictionaries and dictionary
inheritance. All dictionary fields are optional by default and hence only have
builder-style setters, but dictionaries can also have required fields. Required
fields are exposed as arguments to the `new` constructor.

Closes #241

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -74,6 +74,9 @@ impl TryToTokens for ast::Program {
                 errors.push(e);
             }
         }
+        for d in self.dictionaries.iter() {
+            d.to_tokens(tokens);
+        }
 
         Diagnostic::from_vec(errors)?;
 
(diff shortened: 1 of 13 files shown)


crates/webidl/src/lib.rs
```diff
@@ -924,3 +928,90 @@ impl<'src> WebidlParse<'src, (&'src str, &'src mut backend::ast::Module)> for we
...
+                        // dictionary at all, but if it's not required we can
+                        // avoid generating bindings for the field and keep
+                        // going otherwise.
+                        if member.required.is_some() {
+                            return false
+                        }
+                    }
+                }
+            }
+            dst[start..].sort_by_key(|f| f.name.clone());
```
ohanar: This is sorting by the snake case identifier, rather than the original identifier, which could produce the wrong result in edge cases:
```webidl
dictionary Example {
    boolean fooBaz;
    boolean foo_bar;
};
```
That is correctly ordered per the spec, but this line will reorder it to `[foo_bar, foo_baz]`.

crates/backend/src/codegen.rs
```diff
@@ -1135,6 +1138,135 @@ impl<'a> TryToTokens for ast::Module {
...
+        (quote! {
+            pub fn #name(&mut self, val: #ty) -> &mut Self {
+                use wasm_bindgen::JsValue;
+                ::js_sys::Reflect::set(
+                    self.obj.as_ref(),
+                    &JsValue::from(stringify!(#name)),
+                    &JsValue::from(val),
+                );
+                self
+            }
```
ohanar: Without corresponding getters, if a dictionary is returned, it is clumsy to use (have to upcast it to an Object, which I have concerns about).

crates/backend/src/codegen.rs
```diff
@@ -1135,6 +1138,135 @@ impl<'a> TryToTokens for ast::Module {
...
+                }
+                impl AsRef<JsValue> for #name {
+                    fn as_ref(&self) -> &JsValue { self.obj.as_ref() }
+                }
+                impl AsMut<JsValue> for #name {
+                    fn as_mut(&mut self) -> &mut JsValue { self.obj.as_mut() }
+                }
+
+                // interop w/ Object
+                impl From<#name> for Object {
```
ohanar: I don't think there is any guarantee that we could do this with the host bindings proposal. In particular, I don't see any mention that WebIDL dictionaries have to be ECMAScript Objects. (This is really more of a concern for the AsRef and AsMut impls than the conversion impl, but I think we should implement all or none.)

crates/backend/src/codegen.rs
```diff
@@ -1135,6 +1138,135 @@ impl<'a> TryToTokens for ast::Module {
...
+        }).to_tokens(tokens);
+    }
+}
+
+impl ToTokens for ast::DictionaryField {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        let name = &self.name;
+        let ty = &self.ty;
+        (quote! {
+            pub fn #name(&mut self, val: #ty) -> &mut Self {
```
ohanar: You use `pub` here, even though the Dictionary might not have a public visibility.

crates/webidl/src/first_pass.rs
```diff
@@ -118,14 +127,19 @@ impl<'src> FirstPass<'src, ()> for weedle::Definition<'src> {
...
+            .or_insert_with(Default::default)
+            .definition = Some(self);
+        Ok(())
+    }
+}
 
+impl<'src> FirstPass<'src, ()> for weedle::PartialDictionaryDefinition<'src> {
+    fn first_pass(&'src self, record: &mut FirstPassRecord<'src>, (): ()) -> Result<()> {
+        record.dictionaries.entry(self.identifier.0)
+            .or_insert_with(Default::default)
```
ohanar: Maybe `or_default()` here (and above)?

crates/backend/src/codegen.rs
```diff
@@ -1135,6 +1138,135 @@ impl<'a> TryToTokens for ast::Module {
...
+                    }
+                }
+
+                impl FromWasmAbi for #name {
+                    type Abi = <Object as FromWasmAbi>::Abi;
+                    unsafe fn from_abi(abi: Self::Abi, extra: &mut Stack) -> Self {
+                        #name { obj: Object::from_abi(abi, extra) }
+                    }
+                }
+
```
ohanar: Missing `RefFromWasmAbi`.

eminence:
One places that uses a Dictionary is AudioContext.webidl:

```webidl
dictionary AudioContextOptions {
             float        sampleRate = 0;
};

[Pref="dom.webaudio.enabled",
 Constructor(optional AudioContextOptions contextOptions)]
interface AudioContext : BaseAudioContext {
 //...
}
```

While the `AudioContextOptions` dictionary gets correctly generated, and a no-arg constructor is generated for the `AudioContext` struct, there is no `AudioContext` constructor that accepts an `AudioContextOptions` struct.     Is this expected?

alexcrichton:
Thanks for taking a look @ohanar! It's a good point that I don't think we quite know how this is going to map to the host bindings proposal. We could definitely be more conservative and expose none of the internal details about `Object`.

I'm not sure what the best route here is. I don't think we know how this is gonna end up with host bindings (these functions may just not ever be bound with host bindings?). In any case though I think it's *definitely* the case that we're going to publish multiple major versions of `web-sys`. There's a lot of API design decisions we're only guessing about and we're definitely going to get some of them wrong. In that sense I'd probably be in favor of exposing the impls today with the assumption that we may very well reverse this decision in the future if a better binding strategy comes about.

crates/backend/src/codegen.rs
```diff
@@ -74,6 +74,9 @@ impl TryToTokens for ast::Program {
                 errors.push(e);
             }
         }
+        for d in self.dictionaries.iter() {
+            d.to_tokens(tokens);
```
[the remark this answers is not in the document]
alexcrichton: Nah it's not currently implemented because I think it's ambiguous whether you want bracket delimiters or everything just concatenated together

crates/backend/src/codegen.rs
```diff
@@ -1135,6 +1138,135 @@ impl<'a> TryToTokens for ast::Module {
...
+        let required_names = &self.fields.iter()
+            .filter(|f| f.required)
+            .map(|f| &f.name)
+            .collect::<Vec<_>>();
+        let required_types = &self.fields.iter()
+            .filter(|f| f.required)
+            .map(|f| &f.ty)
+            .collect::<Vec<_>>();
+        let required_names2 = required_names;
+        let required_names3 = required_names;
```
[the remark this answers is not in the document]
alexcrichton: Unfortunately yeah, it has to do with the mystery of the `quote!` macro, although I'm not quite sure why. If the same variable name is used below it generates a bunch of scary compiler errors.

alexcrichton:
@eminence turned out to be a bug, should be fixed now!

crates/webidl/src/idl_type.rs
```diff
@@ -433,8 +433,10 @@ impl<'a> IdlType<'a> {
             IdlType::UnsignedShort => Some(ident_ty(raw_ident("u16"))),
             IdlType::Long => Some(ident_ty(raw_ident("i32"))),
             IdlType::UnsignedLong => Some(ident_ty(raw_ident("u32"))),
-            IdlType::LongLong => Some(ident_ty(raw_ident("i64"))),
-            IdlType::UnsignedLongLong => Some(ident_ty(raw_ident("u64"))),
+            // Note that these map to `f64` because JS doesn't really have a
+            // native type for u64, so WebIDL surely just means a JS number
+            IdlType::LongLong => Some(ident_ty(raw_ident("f64"))),
+            IdlType::UnsignedLongLong => Some(ident_ty(raw_ident("f64"))),
```
alexcrichton: This is a pretty subtle update in this PR, but lots of dictionaries apparently have `long long` specified in them. It's unclear what WebIDL could mean *other* than `f64` for JS because 64-bit integers aren't possible, so these two now map to f64

ohanar: From the [spec](https://heycam.github.io/webidl/#idl-long-long):
>The long long type is a signed integer type that has values in the range [−9223372036854775808, 9223372036854775807].
In other words, WebIDL says a long long is a 64-bit integer. Presumably with host-bindings this won't be an issue, it is however an issue with current day shims. I don't think we should change all the signatures of methods just because of modern day limitations.

ohanar:
@alexcrichton My vote would be to be on the more conservative side. I don't see that much value added by having the ability to convert to an Object, so the cost/value ratio doesn't really add up for me.

ohanar:
Well, really conversion to a JSValue is needed to power dictionaries-in-dictionaries, not conversion to js_sys::Object (which was my main complaint).

alexcrichton:
@ohanar oh interesting, can you expand on why `JsValue` is ok but `Object` isn't? I figured that we'd want to remove both to be conservative but having a conversion to `JsValue` seems like it'd be natural to have a conversion to `Object` as well (in that adding `Object` isn't really that much more of a promise in the API, the conversion to `JsValue` is already the big one)

eminence:
With the latest fix, I confirm that the missing web audio constructors are now available, and that I was able to successfully get a sine-wave demo working!   This obviously isn't a rigorous test of every dictionary, but I'm very happy that this particular one is working :smiley:  :notes:

alexcrichton:
I've gone ahead and removed the asref/asmut/from implementations for `Object`

ohanar:
So I guess my concern is that Objects are much more general maps than dictionaries (which are more akin to structs), and I would be a bit surprised if they were re-used for host bindings, just out of concern of efficiency. On the other hand, I would be less surprised if dictionaries were still opaque data structures, so they would presumably be a subtype of anyref, which will likely be the underlying type for a JsValue in the future.

This is obviously me making guesses, but I generally want to take as conservative of an approach as reasonably possible, just so we can minimize potential breakage in the future.

PR #703, opened 2018-08-14
Guide: explain how to deploy the applicaton to a web server

lnicola:
`wasm-bindgen` was pretty much my first contact with the `npm` ecosystem, and I was stuck for a really long time trying to publish my application without manually copying the `index.html` file.

See https://github.com/webpack/webpack/issues/536, https://stackoverflow.com/questions/32155154/webpack-config-how-to-just-copy-the-index-html-to-the-dist-folder, and others.

---
EDIT: The main point here is that the guide currently tells how to run a `webpack` dev server, but not how to get from there to a production server.

In its current form, this PR:

- teaches the reader how to run `webpack` outside of the dev server
- adds `html-webpack-plugin`, a `webpack` plugin that can copy or generate an `index.html` file
- removes `index.html`, since the generated one is good enough, but still describes how to use a custom one

This should be useful for readers who are more familiar with Rust than `webpack` and want to deploy their application.


alexcrichton:
Thanks for the PR! I think though that this sort of publishing documentation may be best suited for `wasm-pack` perhaps? That tool is more focused on publishing than wasm-bindgen is

alexcrichton:
er sorry, didn't mean to close yet

lnicola:
I don't know. I tried to follow the `wasm-bindgen` tutorial, and got a bit stuck at this step, so it might be useful for people coming from the Rust (not JavaScript) side. Note that by publishing I mean getting something to put on a web server, not creating a new `npm` package.

`wasm-pack` seems to be even more biased towards JavaScript developers, to the point I'm not even sure what it does.

lnicola:
Travis errors can't be caused by my changes, can they?

alexcrichton:
Hm ok, but I don't think that either `file-loader` or `npx` should be needed to run wasm-bindgen-generated code. Do you know why those are both listed here?

eminence:
I share some of @lnicola confusion about wasm-pack.  To quote it's [README](https://github.com/rustwasm/wasm-pack):

> `wasm-pack` helps you build and publish rust-generated WebAssembly to the npm registry to be used alongside any other javascript package in workflows that you already use, such as webpack or greenkeeper.

If I want to publish some wasm and JS code to a static website, I don't want to publish anything to the rpm registry.  So it's not clear how `wasm-pack` helps me in this case.

Maybe we should open an issue on the wasm-pack repo to get their input on the use-cases wasm-pack is trying to support?

lnicola:
> Hm ok, but I don't think that `file-loader` should be needed to run wasm-bindgen-generated code.

It's not needed, especially if you use `--no-modules`. But the guide uses `webpack` (which actually does things that one might want, like minifying your JS code) and `webpack-dev-server` to invoke it and serve your application over HTTP (running WASM off the file system doesn't work too well,).

But when you complete your project and actually want to put it under a production web server, the guide doesn't tell you how. `npx webpack` should do it, but doesn't copy your `index.html` file, so you end up with something you can't run, and need to manually copy it each time.

`file-loader` does that instead, and you can actually see I added it as a "dev" dependency, so it's technically not needed to run your code, only to build it.

> or `npx`

`npx webpack` is just a fancy way of running `./node_modules/.bin/webpack`.

And by the way, it seems that `webpack-dev-server` has been superseded by `webpack-serve`.

@eminence 

> Maybe we should open an issue on the wasm-pack repo to get their input on the use-cases wasm-pack is trying to support?

If I'm reading it right, it's a tool aimed at the `npm` ecosystem that makes it easier to prepare packages using Rust and WASM. So it's for `node` developers who most likely know their way around `webpack`. My PR here tries to make things a little easier for Rust developers who aren't into `node`.

guide/src/whirlwind-tour/basic-usage.md
```diff
@@ -103,6 +103,8 @@ support! Let's take a look at that and see how it works.
 First create an `index.js` file:
 
 ```js
+require('file-loader?name=[name].[ext]!./index.html');
```
sendilkumarn: You can use [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin/blob/master/README.md) instead

guide/src/whirlwind-tour/basic-usage.md
```diff
@@ -112,7 +114,8 @@ js.then(js => {
 
 Note that we're using `import(..)` here because Webpack [doesn't
 support][webpack-issue] synchronously importing modules from the main chunk just
-yet.
+yet. The `file-loader` plugin will copy your `index.html` file when you publish
```
sendilkumarn: Instead of saying, it will publish the application. It would be great to explain it here

guide/src/whirlwind-tour/basic-usage.md
```diff
@@ -170,6 +174,14 @@ $ npm run serve
 If you open [http://localhost:8080](http://localhost:8080) in a browser you should see a `Hello, world!`
 dialog pop up!
 
+To publish the application, simply run:
+
+```shell
+$ npx webpack
```
sendilkumarn: Well we can just use npm run webpack. Simpler and default

lnicola: That doesn't work: `npm ERR! missing script: webpack`. I think `npm run` only works with scripts defined in your `package.json`, like `serve` is defined in the guide.

I'm not sure what you mean by default, `npx` seems to be part of `npm`: https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b.

guide/src/whirlwind-tour/basic-usage.md
lnicola: Is there any reason to use that instead? By default, it generates its own HTML instead of copying the one in your project (i.e. the one the guide told you to write). That's not what you want if you're planning to interact with the user.

[This](https://github.com/jantimon/html-webpack-plugin/blob/master/README.md#writing-your-own-templates) might work, but it involves learning about a templating language, and overall seems much more involved than dropping a single `require` line.

guide/src/whirlwind-tour/basic-usage.md
lnicola: I'm not sure what you mean. Is "publishing" a poor term for "getting something that can run outside of the dev server"? If so, can you suggest a different one?

guide/src/whirlwind-tour/basic-usage.md
sendilkumarn: It can also copy the template and if anybody who knows html and templating, they can also playaround with it. 

So templating language is not a must and it is much preferred way to handle html files.

guide/src/whirlwind-tour/basic-usage.md
sendilkumarn: wasm-pack too publishes your application and definitely you are not talking about the same publication here. Doesn't it make sense to explain that?  (the overloading of application is confusing)

guide/src/whirlwind-tour/basic-usage.md
lnicola: It seems much more powerful, but is templating something useful in a "Getting started with `wasm-bindgen`" guide? I went for what seemed to be the simplest solution, in order not to scare away non-`node` devs.

But sure, I'll try to switch to that instead.

guide/src/whirlwind-tour/basic-usage.md
sendilkumarn: Oops, Well I was under the assumption that we already have the script. 

By default, I mean everybody uses that and knows about it.

lnicola: Well, `npx` sounds easier, since that's what it's intended for (running tools from `.bin`), and there's no need to add something to `package.json`.

guide/src/whirlwind-tour/basic-usage.md
lnicola: Done

guide/src/whirlwind-tour/basic-usage.md
lnicola: Done

guide/src/whirlwind-tour/basic-usage.md
```diff
@@ -170,6 +164,18 @@ $ npm run serve
 If you open [http://localhost:8080](http://localhost:8080) in a browser you should see a `Hello, world!`
 dialog pop up!
 
+Notice that `html-webpack-plugin` has generated an HTML page which includes `index.js`.
+
```
eminence: I think I would be a little sad to see the index.html file removed.  I liked it because (1) it showed that there wasn't anything special in it and (2) I immediately knew where to start expanding my application.

So maybe you could add a line or two of instruction text that says something like "If you want to use your own index.html file instead of the generated one, do steps XYZ"

lnicola: Done

alexcrichton:
Hm I'm still slightly confused in the sense of why is the extra dependency needed and why is `npx` used here still? (is that installed with node by default?) 

There's a lot of information about webpack we're not explaining in the tutorial and we're in theory just showing a bare-bones how to get it up and running and it's up to authors to connect the dots with webpack tutorials afterwards, right?

lnicola:
> why is npx used here still? (is that installed with node by default?)

`npx webpack` is equivalent to `./node_modules/.bin/webpack`. `npx` is part of `node` and is the preferred way of running `nodejs` dev binaries.

> I'm still slightly confused in the sense of why is the extra dependency needed

The guide stops with running the web server. I've read the guide, but then I wasn't able to deploy my app. That sounds like a small step from the web server, right?

> it's up to authors to connect the dots with webpack tutorials afterwards

Sure. I've googled for "webpack deploy site". The first results are:

- https://itnext.io/from-zero-to-deploy-set-up-react-stack-with-webpack-3-20b57d6cb8d7, which starts with `yarn` and talks about `webpack 3`; we're on `4`
- https://webpack.js.org/guides/production/, which seems promising, but talks more about merging multiple `webpack` configuration files and throws in a lot of JS which doesn't make much sense in the context of the `wasm-bindgen` guide; in hindsight, it also uses `html-webpack-plugin`, but doesn't say what it does or why it is needed
- https://www.npmjs.com/package/webpack-deploy -- I'm not sure what it does
- https://survivejs.com/webpack/techniques/deploying/ -- this one cares more about `gh-pages`
- https://github.com/CanopyTax/single-spa/issues/177 -- a random issue on a "javascript metaframework"
- https://github.com/productboard/webpack-deploy -- this again
- https://www.sitepoint.com/bundle-static-site-webpack/ -- talks about `webpack 3`, JS transpilers, and other stuff

After a while I figured that I should just run `webpack`, but then stumbled on the missing `index.html` problem. I asked on IRC and somebody told me to manually copy the HTML every time or make a symlink. I also googled for that and found a four-year old `webpack` [issue](https://github.com/webpack/webpack/issues/536) with 131 likes that's still unresolved and also [this](https://stackoverflow.com/questions/32155154/webpack-config-how-to-just-copy-the-index-html-to-the-dist-folder) SO answer which pointed me to `file-loader` and `html-webpack-plugin`.

---

If the guide starts with `webpack`, I think it should be thorough instead of stopping in the middle. I read the guide, built my application, but I was then unable to (properly) deploy it because I had no experience with `webpack`. It was a very frustrating experience.

Of course, I'm not saying that the guide should explain every `webpack` feature (especially with the `npm` ecosystem in constant flux as it seems to be). But we shouldn't expect the users to trawl through four-year old feature requests looking for how to actually get some output from it.

lnicola:
I updated the PR description.

eminence:
I think I'm with @alexcrichton in that it's not the place for the wasm-bindgen to cover the depths and details of wasmpack.  I think the guide's "quick start" does a really good job getting people able to see for themselves what neat stuff wasm-bindgen can do, and if that means using some mysterious webpack development tools, then so be it.

But I do think there is a place for a full and in-depth guide to deploying a non-trivial application into production.  I can imagine that wasm-bindgen might attract rust developers who are not web developers, but they see the power of rust on the web and want to start becoming web developers :).    Would an out-of-tree guide be the best place for such a guide?  (It could of course be linked to from the wasm-bindgen guide)

lnicola:
> it's not the place for the wasm-bindgen to cover the depths and details of wasmpack

I agree with that, and I'm fine with not going into more detail about `webpack`. But after completing the guide, your app is stuck in the dev server and you can't get it out. Reading the guide, I'm presented with `webpack` (which I didn't want to learn), I copy-paste some commands, I can get my app to work, but there's no way to show it to someone else. The plan should be:

1. Write some Rust code
2. Write some JavaScript to interop with it
3. Run this magical `webpack` thingy
4. ???
5. Copy `dist` to your web server and post on Twitter about your first Rust web app

I'm just trying to fill in step 4 in that list.

alexcrichton:
After takling with @lnicola on IRC I think we do want to effectively merge this PR. That being said I do think we'll want to tweak a few things.

The main rationale I think for merging this is that the current setup for specifying `index.html` appears to be incorrect. If `webpack` as a bare command is executed then `index.html` is missing from the output directory, whereas using the html plugin here will generate it. 

While I do think this is breaching a bit too much into the territory of "webpack is a bit magical" I think it's fine in the sense that it's a lot of copy-paste that can be glossed over for the uninterested or investigated for those for are.

I think though that we'll want to update all of the `examples` folder as well as they're following the same pattern. @lnicola would you be up for doing that too?

eminence:
Wherever the guide lives, I'm glad that someone is working on it!

lnicola:
Sure, I'll update the examples, but probably not until next week.

lnicola:
Updated and rebased, sorry for the delay.

Tarnadas:
I also think that the repository is in desperate need of instructions on how to use `wasm-bindgen` in production, therefore I created PR #815

alexcrichton:
Thanks @lnicola!

PR #708, opened 2018-08-15
Adds Extends to TypedArrays

sendilkumarn:
While the `UInt8Array` (and others) have the `TypedArray` in the prototype chain. 

`TypedArray` by itself is not known to `window` or `global`. I don't think we can test that here. 

Is there something that I am missing

### DIFF
--- a/crates/js-sys/tests/wasm/TypedArray.rs
+++ b/crates/js-sys/tests/wasm/TypedArray.rs
@@ -13,6 +13,7 @@ macro_rules! each {
         $m!(Int16Array);
         $m!(Int32Array);
         $m!(Float32Array);
+        $m!(Float64Array);
     )
 }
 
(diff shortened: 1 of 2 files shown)


crates/js-sys/tests/wasm/TypedArray.rs
```diff
@@ -16,6 +17,32 @@ macro_rules! each {
...
+        assert!(arr.is_instance_of::<$arr>());
+        assert!(arr.is_instance_of::<Object>());        
+    })
+}
+#[wasm_bindgen_test]
+fn inheritence() {
+    each!(test_inheritence);
+}
+
+macro_rules! each {
```
fitzgen: This `each!` macro is the same as the one above it -- we should only use the original `each!` macro and delete this one.

crates/js-sys/tests/wasm/TypedArray.rs
```diff
@@ -16,6 +17,32 @@ macro_rules! each {
     )
 }
 
+macro_rules! test_inheritence {
+    ($arr:ident) => ({
+        let arr = $arr::new(&JsValue::undefined());
+        assert!(arr.is_instance_of::<$arr>());
+        assert!(arr.is_instance_of::<Object>());
```
fitzgen: This should also have

```rust
let _: &Object = arr.as_ref();
```

sendilkumarn: I don't get this ?

PR #713, opened 2018-08-16
Add experimental support for `WeakRef`

### LINKED ISSUE #704: Prototype WeakRef support with a SpiderMonkey shell polyfill
https://gist.github.com/tschneidereit/51dd7d5b51e16e21f8281b111fe638d1 is a polyfill for the upcoming `WeakRef` proposal, and we're definitely going to want to make use of it to avoid the need to manually call `free` on Rust structs exported to JS.

We should add an option to wasm-bindgen (probably an env var for now) to tailor output of wasm-bindgen for the SpiderMonkey JS shell and use these weakref bindings!

alexcrichton:
This commit adds experimental support for `WeakRef` to be used to automatically
free wasm objects instead of having to always call the `free` function manually.
Note that when enabled the `free` function for all exported objects is still
generated, it's just optionally invoked by the application.

Support isn't exposed through a CLI flag right now due to the early stages of
the `WeakRef` proposal, but the env var `WASM_BINDGEN_WEAKREF` can be used to
enable this generation. Upon doing so the output can then be edited slightly as
well to work in the SpiderMonkey shell and it looks like this is working!

Closes #704

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -490,6 +490,39 @@ impl<'a> Context<'a> {
         let mut dst = format!("class {} {{\n", name);
         let mut ts_dst = format!("export {}", dst);
 
+        let (mkweakref, freeref) = if self.config.weak_refs {
+            // When weak refs are enabled we use them to automatically free the
+            // contents of an exported rust class when it's gc'd. Note that a
+            // manual `free` function still exists for deterministic
+            // destruction.
+            //
+            // This is implemented by using a `WeakRefGroup` to run finalizers
+            // for all `WeakRef` objects that it creates. Upon construction of
+            // a new wasm object we use `makeRef` with "holdings" of a thunk to
+            // free the wasm instance.  Once the `this` (the instance we're
+            // creating) is gc'd then the finalizer will run with the
+            // `WeakRef`, and we'll pull out the `holdings`, our pointer.
+            //
+            // Note, though, that if manual finalization happens we want to
+            // cancel the `WeakRef`-generated finalization, so we retain the
+            // `WeakRef` in a global map. This global map is then used to
+            // `drop()` the `WeakRef` (cancel finalization) whenever it is
+            // finalized.
+            self.expose_cleanup_groups();
+            let mk = format!("
+                const cleanup_ptr = this.ptr;
+                const ref = CLEANUPS.makeRef(this, () => free{}(cleanup_ptr));
+                CLEANUPS_MAP.set(this.ptr, ref);
+            ", name);
+            let free = "
+                CLEANUPS_MAP.get(ptr).drop();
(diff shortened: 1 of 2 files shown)


alexcrichton:
f? @tschneidereit

crates/cli-support/src/js/mod.rs
```diff
@@ -490,6 +490,45 @@ impl<'a> Context<'a> {
         let mut dst = format!("class {} {{\n", name);
         let mut ts_dst = format!("export {}", dst);
 
+        let (mkweakref, freeref) = if self.config.weak_refs {
+            // When weak refs are enabled we use them to automatically free the
+            // contents of an exported rust class when it's gc'd. Note that a
+            // manual `free` function still exists for deterministic
+            // destruction.
+            //
+            // This is implemented by using a `WeakRefGroup`-per-exported-class
```
tschneidereit: Why do you create a group per class? To support different protocols for freeing instances? It doesn't look like there are differences in protocol right now, so ideally we'd just have one group per WebAssembly instance.

The idea with `WeakRefGroup` is to use it to guarantee cleanup as long as an interconnected system hasn't been GC'd. Specifically, for a WebAssembly module instance, you'd tie the group's lifetime to the instance's lifetime—most likely by just setting a property on the instance pointing to the group. If the entire instance is GC'd, then there's no need to do individual cleanup, so the group should be GC'd without running finalizers, too. Conversely, as long as the group is around, finalizers are guaranteed to run: the group keeps all the `WeakRef` instances created through it alive.

fitzgen: > Why do you create a group per class? To support different protocols for freeing instances? It doesn't look like there are differences in protocol right now, so ideally we'd just have one group per WebAssembly instance.

They have different destructors, and therefore each class has different drop-and-free functions. See my comment elsewhere in the PR about how we could have a single group, though.

crates/cli-support/src/js/mod.rs
```diff
@@ -490,6 +490,45 @@ impl<'a> Context<'a> {
...
+            // which contains a finalizer for all `WeakRef` objects that it
+            // creates. Upon construction of a new wasm object we use `makeRef`
+            // with "holdings" of the wasm pointer value (just an integer).
+            // Once the `this` (the instance we're creating) is gc'd then the
+            // finalizer will run with the `WeakRef`, and we'll pull out the
+            // `holdings`, our pointer.
+            //
+            // Note, though, that if manual finalization happens we want to
+            // cancel the `WeakRef`-generated finalization, so we retain the
```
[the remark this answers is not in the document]
alexcrichton: Correct!

crates/cli-support/src/js/mod.rs
alexcrichton: Ah yeah I wasn't initially sure on the expense of having a `WeakRefGroup`, but using just one is totally plausible!

Issue #443, opened 2018-07-10
Add WebAudio support to web-sys


fitzgen:
This is a tracking issue for adding support for the web audio API to the `web-sys` crate.

General info on web audio: https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API

How to add support for new Web APIs to `web-sys`: https://rustwasm.github.io/wasm-bindgen/web-sys/supporting-more-web-apis.html

## Interfaces

Here is a list of interfaces we need to support:

* [ ] `AudioContext`
* [ ] `AudioContextOptions`
* [ ] `AudioNode`
* [ ] `AudioParam`
* [ ] `AudioParamMap`
* [ ] `BaseAudioContext`
* [ ] `The ended event`
* [ ] `AudioScheduledSourceNode`
* [ ] `OscillatorNode`
* [ ] `AudioBuffer`
* [ ] `AudioBufferSourceNode`
* [ ] `MediaElementAudioSourceNode`
* [ ] `MediaStreamAudioSourceNode`
* [ ] `BiquadFilterNode`
* [ ] `ConvolverNode`
* [ ] `DelayNode`
* [ ] `DynamicsCompressorNode`
* [ ] `GainNode`
* [ ] `WaveShaperNode`
* [ ] `PeriodicWave`
* [ ] `IIRFilterNode`
* [ ] `AudioDestinationNode`
* [ ] `MediaStreamAudioDestinationNode`
* [ ] `AnalyserNode`
* [ ] `ChannelSplitterNode`
* [ ] `ChannelMergerNode`
* [ ] `AudioListener`
* [ ] `PannerNode`
* [ ] `StereoPannerNode`
* [ ] `OfflineAudioContext`
* [ ] `complete (event)`
* [ ] `OfflineAudioCompletionEvent`
* [ ] `JavaScriptNode`
* [ ] `WaveTableNode`

## Example

* [ ] Add an example to `examples/` that shows off working with web audio

eminence:
I took a look at this over the weekend.  Nearly all of these interfaces are "enabled", but many of them are incomplete due to lack of dictionary support (#241).   But once that is fixed up, I'll be happy to develop some Web Audio examples

fitzgen:
[...]

PR #716, opened 2018-08-17
Initial example of using the WebAudio APIs from web-sys

eminence:
Part of #443

(I tried to also host this on https://webassembly.studio/, but I'm not sure how to do that as it doesn't have web-sys)


alexcrichton:
Awesome, thanks! It's fine to skip webassembly.studio for now, but it looks like the build.sh script needs to be made executable?

eminence:
> it looks like the build.sh script needs to be made executable?

Oops!  Fixed

PR #723, opened 2018-08-18
Add extends attributes for js_sys:Generator

mstallmo:
Part of #670 

I'm pretty new to working with JS Generators and as near as I can tell they just inherit from Object and that's it but I would love a second pair of eyes on that to confirm. 

Thanks!


alexcrichton:
Perfect!

PR #724, opened 2018-08-18
Fix missing WindowOrWorkerGlobalScope partial interface mixins.

eminence:
Without the `mixin` keyword, wasm_bindgen_webidl would report:

    Partial interface WindowOrWorkerGlobalScope missing non-partial interface

Also, including the "mixin" keyword here is consistent with the official
webidl spec (for example see https://fetch.spec.whatwg.org/#fetch-method).

The most important thing this change unlocks is the `fetch` method (CC #449)

### DIFF
--- a/crates/web-sys/webidls/enabled/WindowOrWorkerGlobalScope.webidl
+++ b/crates/web-sys/webidls/enabled/WindowOrWorkerGlobalScope.webidl
@@ -43,25 +43,25 @@ interface mixin WindowOrWorkerGlobalScope {
 };
 
 // https://fetch.spec.whatwg.org/#fetch-method
-partial interface WindowOrWorkerGlobalScope {
+partial interface mixin WindowOrWorkerGlobalScope {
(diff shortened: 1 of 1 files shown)


eminence:
With this change, the only remaining instance of "missing non-partial interface" is:

    Partial interface ConsoleInstance missing non-partial interface

Which I thing comes from the [following ](https://github.com/rustwasm/wasm-bindgen/blob/302f7ba21d0480eb8c66e58c1d112fc25ebf8055/crates/web-sys/webidls/enabled/Console.webidl#L223)(which looks ignoreable):

```webidl
// this interface is just for testing
partial interface ConsoleInstance {
  [ChromeOnly]
  void reportForServiceWorkerScope(DOMString scope, DOMString message,
                                   DOMString filename, unsigned long lineNumber,
                                   unsigned long columnNumber,
                                   ConsoleLevel level);
};
```

alexcrichton:
Looks like a great fix to me, thanks for tracking this down!

Issue #503, opened 2018-07-18
Add webidl variadic (`...`) support to web-sys


derekdreery:
Like `optional` this requires deciding how to represent this in rust. It could be a slice (so `...any` becomes `&[JsValue]`, etc.), or a varargs function. Since varargs isn't implemented yet ([tracking issue]), probably best to go with a slice.

Example of a variadic function in WebIDL (Console API):

```webidl
[UseCounter]
void log(any... data);
```

[tracking issue]: https://github.com/rust-lang/rust/issues/44930

fitzgen:
Fixed in #726

PR #726, opened 2018-08-18
Support variadic javascript function parameters

derekdreery:
The main bit to do is the actual logic to convert the slice into a javascript array, and then do `fn.apply(something, our_slice)`.

I've put some diagnostic logic in to stop incorrect usage, I think there are more opportunities for improvement here.

A question: I want to test failures when the attribute is in the wrong place. Is there a way to do this?

Relevant issue: #503.


derekdreery:
This is ready for review.

I think it needs more tests, but I'm not sure what type. Please advise :)

derekdreery:
I also need to implement this the other way -> so that javascript can call a variadic rust function, and the variadic part of the arguments gets turned into a `Vec<T>` or similar (I think it needs to be owned).

derekdreery:
I think the ci error is spurious.

crates/cli-support/src/js/rust2js.rs
```diff
@@ -528,10 +543,31 @@ impl<'a, 'b> Rust2Js<'a, 'b> {
...
+            if self.js_arguments.is_empty() {
+                unreachable!("the last argument of a variadic must be a slice");
+            }
+            let last_arg = self.js_arguments.len() - 1; // check implies >= 0
+            // The variadic part of the arguments could be passed as as a typed array in
+            // javascript, so we cannot just do `.concat(..)`, as this seems to convert the
+            // contents to strings (javascript is very strange).
+            write!(ret, "\
+                let args = [{}];\n\
+                {}.forEach(extra_arg => args.push(extra_arg));\n\
```
alexcrichton: We're already using pretty modern syntax elsewhere, so instead of this could it instead generate `foo(a, b, ...c)` perhaps?

derekdreery: This seems to work - I was expecting the problem mentioned in the comment.

I'll add plenty of tests to try with lots of different types.

derekdreery: done

crates/macro-support/src/parser.rs
```diff
@@ -1074,6 +1095,40 @@ fn assert_no_lifetimes(decl: &syn::FnDecl) -> Result<(), Diagnostic> {
     Diagnostic::from_vec(walk.diagnostics)
 }
 
+/// This method always fails if the BindgenAttrs contain variadic
+fn assert_not_variadic(attrs: &BindgenAttrs) -> Result<(), Diagnostic> {
+    match attrs.variadic() {
+        true => Err(Diagnostic::error("the `variadic` attribute can only be applied to imported \
```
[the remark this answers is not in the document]
derekdreery: done

tests/wasm/variadic.js
```diff
@@ -0,0 +1,13 @@
+const wasm = require('wasm-bindgen-test.js');
+const assert = require('assert');
+
+function variadic_sum() {
```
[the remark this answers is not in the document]
derekdreery: done

derekdreery:
I'm adding loads of tests for all the types I can think of. I'm planning to comment out the ones we decide not to support for now.

Because the last slice needs to go through the language boundary before being expanded, all the slice types need to implement `IntoWasmAbi`, some like `&[&str]` don't yet.

I also should support `Vec<T>` as well as `&[T]` for passing the `rest` args.

crates/macro-support/src/parser.rs
```diff
@@ -1074,6 +1095,40 @@ fn assert_no_lifetimes(decl: &syn::FnDecl) -> Result<(), Diagnostic> {
...
+    let ty = match arg.value() {
+        syn::FnArg::Captured(ref arg_cap) => &arg_cap.ty,
+        _ => return Err(not_slice_error(&arg))
+    };
+    match ty {
+        syn::Type::Reference(ref ref_ty) => match &*ref_ty.elem {
+            syn::Type::Slice(_) => Ok(()),
+            _ => Err(not_slice_error(ty))
+        },
+        _ => Err(not_slice_error(ty))
```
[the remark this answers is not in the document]
derekdreery: done.

I've refactored this, but will have to change it again if I support `Vec`s as well.

derekdreery:
Stretch goal: the last parameter can be anything that implements `IntoIter<Iter=<impl IntoWasmAbi>>`

derekdreery:
> I think we'll eventually want to support things like &[JsValue] or &[ImportedType], but those might be sort of hard to add support for :(

I think we can't support this yet since IntoWasmAbi isn't implemented for &[JsValue]. Maybe best to do in a separate PR. I can add the test for it and comment it out for now.

derekdreery:
Ready for review again.

derekdreery:
gentle ping :)

alexcrichton:
Ah oops, sorry for the delay!

After reading this again, what do you think of basically entirely removing the slice validation of the last argument? That way the attribute would be sort of "duck typed" but as long as the last item was iterable it should be able to be passed to the splat we then pass to JS, right? That way we can naturally pick up support for more list-like types as they're implemented.

Additionally could you modify the `guide/*` assorted files to ensure that `variadic` is documented there too?

derekdreery:
I've implemented the duck-typing suggestion, and it works fine. The issue is an incorrect usage of the `variadic` tag will be a runtime error, not compiletime, but maybe this is ok.

derekdreery:
Ready for review again.

crates/cli-support/src/js/rust2js.rs
```diff
@@ -528,10 +543,24 @@ impl<'a, 'b> Rust2Js<'a, 'b> {
         ret.push_str(") {\n");
         ret.push_str(&self.prelude);
 
-        let mut invoc = self.ret_expr.replace(
-            "JS",
-            &format!("{}({})", invoc, self.js_arguments.join(", ")),
-        );
+        let mut invoc = if self.variadic {
+            if self.js_arguments.is_empty() {
+                unreachable!("the last argument of a variadic must be a slice");
```
afdw: Should not it be `bail_span!` instead of `unreachable!`? And the text is a bit outdated, probably should say something like "variadic functions should have at least 1 argument".

derekdreery: I think that before this was unreachable, but since I removed the checking code it now might not be. I'll check.

crates/cli-support/src/js/rust2js.rs
```diff
@@ -528,10 +542,24 @@ impl<'a, 'b> Rust2Js<'a, 'b> {
         ret.push_str(") {\n");
         ret.push_str(&self.prelude);
 
-        let mut invoc = self.ret_expr.replace(
-            "JS",
-            &format!("{}({})", invoc, self.js_arguments.join(", ")),
-        );
+        let mut invoc = if self.variadic {
+            if self.js_arguments.is_empty() {
+                return Err(failure::err_msg("a function with no arguments cannot be variadic"));
```
alexcrichton: Stylistically I think this could just be `bail!("...")`

alexcrichton:
Looks great to me, thanks!

PR #727, opened 2018-08-19
Add binding for String.prototype.localeCompare

quelledanielle:
I went with a similar signature to those in `Intl` but let me know if you have a more convenient suggestion.

cc @twilco


alexcrichton:
Thanks!

PR #729, opened 2018-08-19
Initial example for the Fetch API

eminence:
CC #449

### DIFF
--- a/examples/fetch/.gitignore
+++ b/examples/fetch/.gitignore
@@ -1,0 +1,4 @@
+fetch.d.ts
+fetch.js
+fetch_bg.wasm
+package-lock.json
--- a/examples/fetch/src/lib.rs
+++ b/examples/fetch/src/lib.rs
@@ -1,0 +1,85 @@
+#![feature(use_extern_macros)]
+
+extern crate wasm_bindgen;
+extern crate js_sys;
+extern crate web_sys;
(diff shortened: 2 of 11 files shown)


alexcrichton:
This looks awesome, thanks!

I think there's a permissions error on `build.sh`, but otherwise looks good to go!

Also, perhaps this could show an example of using the serde feature to deserialize the JSON as well?

examples/fetch/.gitignore
```diff
@@ -0,0 +1,4 @@
+fetch.d.ts
+fetch.js
+fetch_bg.wasm
+package-lock.json
```
jrakow: There is a newline missing here.

eminence: Should be fixed now

eminence:
Example has been updated to show off serde magic.

I would have liked for `run()` to return a Promise to a `Branch` object.  `future_to_promise` requires that the future resolve to a `JsValue` type, so while I could have returned the `Branch` as a `JsValue`, I wasn't sure how to convert that back into a `Branch` object in javascript land.  Any hints?

examples/fetch/src/lib.rs
```diff
@@ -0,0 +1,85 @@
+#![feature(use_extern_macros)]
```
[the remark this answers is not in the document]
eminence: Removed \o/

PR #730, opened 2018-08-19
Add extends to js sys Intl.Collater, Intl.DateTimeFormat, Intl.NumberFormat, and Intl.PluralRules

mstallmo:
Part of #670 

Added ```extends = Object``` and accompanying tests to all of the Int.* bindings in js-sys. I did all of them as individual commits but would be happy to squash them into one commit if that would be preferred.

Cheers!


alexcrichton:
Thanks so much!

PR #736, opened 2018-08-20
Remove dependency on `wasmi`

alexcrichton:
This is a pretty heavyweight dependency which accounts for a surprising amount
of runtime for larger modules in `wasm-bindgen`. We don't need 90% of the crate
and so this commit bundles a small interpreter for instructions we know are only
going to appear in describe-related functions.

### DIFF
--- a/crates/cli-support/Cargo.toml
+++ b/crates/cli-support/Cargo.toml
@@ -18,5 +18,5 @@ serde = "1.0"
 serde_json = "1.0"
 tempfile = "3.0"
 wasm-bindgen-shared = { path = "../shared", version = '=0.2.17' }
+wasm-bindgen-wasm-interpreter = { path = "../wasm-interpreter", version = '=0.2.17' }
 wasm-gc-api = "0.1.9"
-wasmi = "0.3"
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -94,7 +94,7 @@ pub enum VectorKind {
 impl Descriptor {
     pub fn decode(mut data: &[u32]) -> Descriptor {
         let descriptor = Descriptor::_decode(&mut data);
-        assert!(data.is_empty());
+        assert!(data.is_empty(), "remaining data {:?}", data);
         descriptor
     }
 
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -10,6 +10,7 @@ use wasm_gc;
 
 use super::Bindgen;
 use descriptor::{Descriptor, VectorKind};
+use wasm_interpreter::Interpreter;
 
 mod js2rust;
 use self::js2rust::Js2Rust;
@@ -41,7 +42,7 @@ pub struct Context<'a> {
 
     pub exported_classes: HashMap<String, ExportedClass>,
     pub function_table_needed: bool,
-    pub run_descriptor: &'a Fn(&str) -> Option<Vec<u32>>,
+    pub interpreter: &'a mut Interpreter,
 }
 
 #[derive(Default)]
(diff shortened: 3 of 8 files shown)


crates/cli-support/src/interpret.rs
```diff
@@ -0,0 +1,155 @@
...
+//! As a recap, the wasm-bindgen macro generate "descriptor functions" which
+//! basically as a mapping of rustc's trait resolution in executable code. This
+//! allows us to detect, after the macro is invoke, what trait selection did and
+//! what types of functions look like. By executing descriptor functions they'll
+//! each invoke a known import (with only one argument) some number of times,
+//! which gives us a list of `u32` values to then decode.
+//!
+//! The interpreter here is only geared towards this one exact use case, so it's
+//! quite small and likely not extra-efficient.
+
```
fitzgen: missing `#![deny(missing_docs)]` ;)

crates/cli-support/src/interpret.rs
```diff
@@ -0,0 +1,155 @@
...
+                assert_eq!(i.value_type(), ValueType::I32);
+                i.count()
+            })
+            .unwrap_or(0);
+        let mut locals = vec![0; locals as usize];
+        let before = self.stack.len();
+        for instr in body.code().elements() {
+            match instr {
+                I32Const(x) => self.stack.push(*x),
+                SetLocal(i) => locals[*i as usize] = self.stack.pop().unwrap(),
```
fitzgen: Are we sharing locals across stack frames? Wait, no we aren't because we are also using recursion in the interpreter.

crates/cli-support/src/interpret.rs
```diff
@@ -0,0 +1,155 @@
...
+                GetLocal(i) => self.stack.push(locals[*i as usize]),
+                Call(idx) => {
+                    if Some(*idx) == self.describe_idx {
+                        self.descriptor.push(self.stack.pop().unwrap() as u32);
```
fitzgen: This is jsut inlining the imported describe function, right? A comment would be nice here, it isn't totally clear why this is special cased.

crates/cli-support/src/interpret.rs
```diff
@@ -0,0 +1,155 @@
...
+        self.call(idx, code);
+        assert_eq!(self.stack.len(), 0);
+        assert_eq!(self.sp, self.mem.len() as i32);
+        Some(&self.descriptor)
+    }
+
+    fn call(&mut self, idx: u32, code: &CodeSection) {
+        use parity_wasm::elements::Instruction::*;
+
+        let body = &code.bodies()[(idx as usize) - self.imports];
```
fitzgen: Assert that we aren't invoking an imported type (and it shouldn't be the describe function either because of the inline handling of it below).

```rust
assert!(idx >= self.imports);
```

alexcrichton:
> If this is the route we intend to go down, let's move the interpreter to a new crate, so we fully leverage improvements to compile times. Additionally, let's write some basic instruction sequences / modules for sanity testing.

👍 

> The pipeline ... correct?

Almost! You're right on the "before" pipeline, but afterwards it's actually just adding support to `parity-wasm` is all we need. We use `parity-wasm` to deserialize and reserialize a module (after some modifications), but `wasm-bindgen` itself doesn't actually have any need to support new features in wasm. For example everything I was working on with atomics is purely related to getting the tooling to *ignore* atomics and basically pass-through all the bits and pieces (aka deserialization and serialization support in `parity-wasm`).

In that sense getting `wasm-bindgen` to support SIMD instructions should *purely* be a case of getting `parity-wasm` to support serializing/deserializing SIMD. Otherwise `wasm-bindgen` needs no knowledge of SIMD instructions!

> I think we can get the same pipeline with less code to maintain by pre-processing the wasm and replacing all new instructions with unreachable before we give it to wasmi

Almost! I don't think this is quite true though in practice. The `wasmi` crate also links against `parity-wasm` and has exhaustive matches against the `Instruction` enumeration (as one would reasonbly expect). That means that we can't just replace them with `unreachable` but for `wasmi`'s code to compile we'll need to actually edit the source code to do *something* on all the new instructions (even if it's just panicking).

In light of all that I'm pretty confident that this strategy is going to reduce the maintenance burden on us. We need to pay this one-time cost to actually implement some of an interpreter but over time I expect us to basically never need to implement more instructions or tweak it. This then also has the added benefit of giving us a nice speed boost for larger modules!

PR #737, opened 2018-08-20
Renamed the binding to `document` as `wbg_document`

### LINKED ISSUE #714: #[wasm_bindgen] on `impl {}` produces duplicate function in wasm-bindgen-test
Hi, 
I am trying to implement some tests for my crate using `wasm-bindgen-test`. If I have `#[wasm_bindgen]` on an `impl ... {...}`, the test failed because of duplicate definition of:
```
// in file: `/wbg-tmp/wasm-bindgen-test.js`
export function __wbg_static_accessor_document_document
```
[Here is the code that cause the problem](https://gitlab.com/limira-rs/simi/blob/wasm-bindgen-test/simi/tests/wasm/counter.rs). (Note that it is on branch `wasm-bindgen-test`, not on master.) If I uncomment line 53, the problem will occur.

csharad:
When `wasm-bindgen-test` is used to test user code which declares bindings to `static document`, duplicate js bindings to it are generated. It is caused due to internal bindings used in `wasm-bindgen-test`.

Fixes #714

### DIFF
--- a/crates/test/src/rt/browser.rs
+++ b/crates/test/src/rt/browser.rs
@@ -17,7 +17,8 @@ pub struct Browser {
 #[wasm_bindgen]
 extern {
     type HTMLDocument;
-    static document: HTMLDocument;
+    #[wasm_bindgen(js_name = document)]
+    static wbg_document: HTMLDocument;
     #[wasm_bindgen(method, structural)]
     fn getElementById(this: &HTMLDocument, id: &str) -> Element;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I think it's probably best to fix the internal bug though which [hopefully isn't too hard](https://github.com/rustwasm/wasm-bindgen/pull/741), so I'm gonna close this in favor of that PR.

PR #738, opened 2018-08-20
feat(js-sys) Implement `String.split` with regexp.

Hywan:
Tests are not running locally. Don't know why yet. Meanwhile, let's wait on Travis.


alexcrichton:
Thanks!

PR #739, opened 2018-08-20
feat(js-sys) Implement `String.replace(&str, …)`

Hywan:
In `String.replace(needle, haystack)`, `needle` can be a string or a regular expression. So far, only the regular expression type is supported. This patch implements the string type.

I'm not sure about the namings though.


alexcrichton:
Looks good to me, thanks!

Issue #701, opened 2018-08-14
WebIDL: Support the BufferSource type


alexcrichton:
We've currently got a number of fields/methods that aren't bound because the `BufferSource` type isn't recognized. [Defined by the spec](https://www.w3.org/TR/WebIDL-1/#h-common-buffersource) this type is basically `TypedArray or DataView or ArrayBuffer`. We unfortunately don't have a great way to bind this type tight now!

How should we support this type? Should we just take `&JsValue` and call it a day?

fitzgen:
For inputs, we can overload these similar to other unions.

For outputs, we can probably use `JsValue` and call it a dat.

+cc @afdw

alexcrichton:
Added in https://github.com/rustwasm/wasm-bindgen/pull/745!

PR #745, opened 2018-08-23
Add ArrayBufferView & BufferSource; Enables Webcrypto API

gnunicorn:
_needs [Weedle#14](https://github.com/rustwasm/weedle/pull/14)_

Adds type definitions for `ArrayBufferView` and `BufferSource` to the parser; both are represented as `[u8]` in Rust for now (see comments below).

Among others, this means the Webcrypto and SubtleCrypto-APIs are now generated properly and accessible from rust.

### DIFF
--- a/crates/webidl/src/idl_type.rs
+++ b/crates/webidl/src/idl_type.rs
@@ -441,7 +441,7 @@ impl<'a> IdlType<'a> {
             | IdlType::DomString
             | IdlType::ByteString
             | IdlType::UsvString => match pos {
-                TypePosition::Argument => Some(shared_ref(ident_ty(raw_ident("str")))),
+                TypePosition::Argument => Some(shared_ref(ident_ty(raw_ident("str")), true)),
                 TypePosition::Return => Some(ident_ty(raw_ident("String"))),
             },
             IdlType::Object => {
@@ -456,15 +456,15 @@ impl<'a> IdlType<'a> {
                 Some(leading_colon_path_ty(path))
             },
             IdlType::DataView => None,
-            IdlType::Int8Array => Some(array("i8", pos)),
-            IdlType::Uint8Array => Some(array("u8", pos)),
+            IdlType::Int8Array => Some(array("i8", pos, true)),
+            IdlType::Uint8Array => Some(array("u8", pos, true)),
             IdlType::Uint8ClampedArray => None, // FIXME(#421)
-            IdlType::Int16Array => Some(array("i16", pos)),
-            IdlType::Uint16Array => Some(array("u16", pos)),
-            IdlType::Int32Array => Some(array("i32", pos)),
-            IdlType::Uint32Array => Some(array("u32", pos)),
-            IdlType::Float32Array => Some(array("f32", pos)),
-            IdlType::Float64Array => Some(array("f64", pos)),
+            IdlType::Int16Array => Some(array("i16", pos, true)),
+            IdlType::Uint16Array => Some(array("u16", pos, true)),
+            IdlType::Int32Array => Some(array("i32", pos, true)),
+            IdlType::Uint32Array => Some(array("u32", pos, true)),
+            IdlType::Float32Array => Some(array("f32", pos, true)),
+            IdlType::Float64Array => Some(array("f64", pos, true)),
 
             IdlType::ArrayBufferView | IdlType::BufferSource => {
                 let path = vec![rust_ident("js_sys"), rust_ident("Object")];
@@ -474,7 +474,7 @@ impl<'a> IdlType<'a> {
             | IdlType::Dictionary(name) => {
                 let ty = ident_ty(rust_ident(camel_case_ident(name).as_str()));
                 if pos == TypePosition::Argument {
-                    Some(shared_ref(ty))
+                    Some(shared_ref(ty, true))
                 } else {
                     Some(ty)
                 }
@@ -488,7 +488,7 @@ impl<'a> IdlType<'a> {
                 let path = vec![rust_ident("js_sys"), rust_ident("Promise")];
                 let ty = leading_colon_path_ty(path);
                 if pos == TypePosition::Argument {
-                    Some(shared_ref(ty))
+                    Some(shared_ref(ty, true))
                 } else {
                     Some(ty)
                 }
--- a/crates/webidl/src/idl_type.rs
+++ b/crates/webidl/src/idl_type.rs
@@ -466,9 +466,10 @@ impl<'a> IdlType<'a> {
             IdlType::Float32Array => Some(array("f32", pos)),
             IdlType::Float64Array => Some(array("f64", pos)),
 
-            // we alias ArrayBufferView & BufferSource to &u8 in Rust
-            IdlType::ArrayBufferView
-            | IdlType::BufferSource => Some(array("u8", pos)),
+            IdlType::ArrayBufferView | IdlType::BufferSource => {
+                let path = vec![rust_ident("js_sys"), rust_ident("Object")];
+                Some(leading_colon_path_ty(path))
+            },
             IdlType::Interface(name)
             | IdlType::Dictionary(name) => {
                 let ty = ident_ty(rust_ident(camel_case_ident(name).as_str()));
@@ -573,6 +574,8 @@ impl<'a> IdlType<'a> {
                 .iter()
                 .flat_map(|idl_type| idl_type.flatten())
                 .collect(),
+            IdlType::ArrayBufferView | IdlType::BufferSource =>
+                vec![IdlType::Object, IdlType::Uint8Array],
 
             idl_type @ _ => vec![idl_type.clone()],
         }
--- a/crates/webidl/src/idl_type.rs
+++ b/crates/webidl/src/idl_type.rs
@@ -33,6 +33,7 @@ pub(crate) enum IdlType<'a> {
     DataView,
     Int8Array,
     Uint8Array,
+    Uint8ArrayMut,
     Uint8ClampedArray,
     Int16Array,
     Uint16Array,
@@ -368,6 +369,7 @@ impl<'a> IdlType<'a> {
             IdlType::DataView => dst.push_str("data_view"),
             IdlType::Int8Array => dst.push_str("i8_array"),
             IdlType::Uint8Array => dst.push_str("u8_array"),
+            IdlType::Uint8ArrayMut => dst.push_str("u8_array"),
             IdlType::Uint8ClampedArray => dst.push_str("u8_clamped_array"),
             IdlType::Int16Array => dst.push_str("i16_array"),
             IdlType::Uint16Array => dst.push_str("u16_array"),
@@ -441,7 +443,7 @@ impl<'a> IdlType<'a> {
             | IdlType::DomString
             | IdlType::ByteString
             | IdlType::UsvString => match pos {
-                TypePosition::Argument => Some(shared_ref(ident_ty(raw_ident("str")), true)),
+                TypePosition::Argument => Some(shared_ref(ident_ty(raw_ident("str")), false)),
                 TypePosition::Return => Some(ident_ty(raw_ident("String"))),
             },
             IdlType::Object => {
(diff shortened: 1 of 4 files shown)


gnunicorn:
It looks like the changes of #579 have the side-effect that for typedef-aliases, we are generating a whole lot of slightly different methods, just to be compatible. For this common type and local type 
```IDL
typedef (Int8Array or Int16Array or Int32Array or
         Uint8Array or Uint16Array or Uint32Array or Uint8ClampedArray or
         Float32Array or Float64Array or DataView) ArrayBufferView;

typedef ArrayBufferView BufferSource;
dictionary Algorithm {
  required DOMString name;
};
```

being parameters in this function in `SubtleCrypto`
```IDL

  [Throws]
  Promise<any> encrypt(AlgorithmIdentifier algorithm,
                       CryptoKey key,
                       BufferSource data);
```

We end up generating **40** method definitions with slightly different types:

screenshot-20180823180137-1444x1562


---

I was thinking: What if we changed the way these bigger-type defs are defined, so that for anything with 3 or more potential cases, we'd create an enum holding these cases instead? So in this case, we'd create:

```rust
pub enum ArrayBufferView {
  Int8Array(Int8Array),
  Int16Array(Int16Array),
  Int32Array(Int32Array),
  Uint8Array(Uint8Array),
  Uint16Array(Uint16Array),
  Uint32Array(Uint32Array),
  Uint8ClampedArray(Uint8ClampedArray),
  Float32Array(Float32Array)
  Float64Array(Float64Array),
  DataView(DataView)
}

encrypt_with_object(
    &self, 
    algorithm: Object, 
    key: &CryptoKey, 
    data: &ArrayBufferView
) -> Result<Promise, JsValue>
```

What do you think, @alexcrichton ?

alexcrichton:
Thanks for the PR! I think we'll probably want to avoid the `Common.webidl` file in the sense that those are built into the WebIDL spec and we can probably just inline the definitions directly into the webidl parser itself. Supporting things like `BufferSource` was briefly brought up in https://github.com/rustwasm/wasm-bindgen/issues/701 and does indeed not have a great solution today unfortunately :(

I agree that the forty different methods is probably a bit excessive for this! I wonder though if we can perhaps get away with a custom definition of a `BufferSource` which is either `JsValue` or `&[u8]`? That may allow it to be easily called from Rust but also allow any number of values to be passed in. What do you think?

gnunicorn:
>
> Yeah, for the webcrypto going for `&[u8]` is reasonable I think - you
usually don't want to change the cipher, key and such. So I'll define
`BufferSource` in here then?

Do we still want `Common.webidl` embedded in the parser then? And do we
want to make the same assumption there about `ArrayBufferView`? I don't
know if this is used (and might break) somewhere else...

alexcrichton:
I think it's probably best to fold `Common.webidl` directly into the parser, I think it should be relatively easy to process during `idl_type.rs` processing, although if it's troublesome let me know!

I think we can basically assume that `BufferSource` and `ArrayBufferView` are the same thing for our purposes (binding generation)

alexcrichton:
Ok great! I've merged and published the `weedle` change, want to update that here too?

alexcrichton:
Thanks! Looks though like though a rebase is needed?

I was also thinking that we may want to support either a `JsValue` or a `&[u8]` to allow passing in direct JS values to these methods (all the JS types they actually support). What do you think of that?

gnunicorn:
> I was also thinking that we may want to support either a `JsValue` or a `&[u8]` to allow passing in direct JS values to these methods (all the JS types they actually support). What do you think of that?

Sounds reasonable. Let me see if I can get the parser to do that (tomorrow).

gnunicorn:
@alexcrichton actually, I am facing a different error I can't really figure out how to handle. see the CI for what I mean. It appears that by implementing the `BufferSource`/`ArrayBufferView` we are triggering some casting issues - namely that `[u8]` can't be easily cast into `JsValue`. Considering the  way that Js handles these, the entire `TypedArray`-Concept, I don't see how we can solve that in a generic manner. However, I haven't quite figured out how `JsValue` works though. Any ideas/hints?

```
                                                                                                    ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::JsCast` is not implemented for `[u8]`
     |
     = note: required because of the requirements on the impl of `std::convert::From<&[u8]>` for `wasm_bindgen::JsValue`
     = note: required because of the requirements on the impl of `std::convert::From<std::option::Option<&[u8]>>` for `wasm_bindgen::JsValue`
     = note: required by `std::convert::From::from`

error: Could not compile `web-sys`.

```

alexcrichton:
Ah it looks like that has to do with dictionary support, so for now we probably just need to say that dictionaries aren't supported with `BufferSource` (or they always use `JsValue` instead of `&[u8]` for now). For now though to get this working [this code](https://github.com/rustwasm/wasm-bindgen/blob/447b42b7accb316a062ff30c79556bea6264e225/crates/webidl/src/lib.rs#L1040-L1071) will need to be updated to not only reject slices but also reject optional slices.

gnunicorn:
Okay, I was able to make the tests pass again, but I am not really happy about [the change, as it isn't very specific](https://github.com/rustwasm/wasm-bindgen/pull/745/commits/8937155a5ea91c9aac9a4cb855aabdd28f6a1489#diff-9f9955b41f2425125eaf42c53a05c94dR1055). Unfortunately, at this part of the code base, we don't have a good reference to learn that this is an `Option<&[u8]>`as far as I can see.

regarding the passing of `JsValue` and `&[u8]`, I am still fiddling with things. I was trying to let our parser emulate it as a `Union<[JsValue, &[u8]]>` but it doesn't really want to. I am not sure, I've understood how unions work in there...

gnunicorn:
Okay, I was able to get them to be passed off as a union of `&[u8]` and `JsValue`. However, I am running into a similar issue somewhere down the line again (see CI for details).

```
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::OptionIntoWasmAbi` is not implemented for `wasm_bindgen::JsValue`
     |
     = note: required because of the requirements on the impl of `wasm_bindgen::convert::IntoWasmAbi` for `std::option::Option<wasm_bindgen::JsValue>`
```

This must be happening somewhere else this time, as we don't have any reference for `Option` at the code part you pointed me earlier. Do you have a hint, where I have to fix this, @alexcrichton ? 

And/or maybe we want to reasses this approach, it feels like I am messing with a lot of things and do some dirty chnages to the codebase, and have no idea what impacts those have down the line (`ArrayBufferView` is used extensively throughout the binary-oriented-parts of the webIDL specs).

alexcrichton:
Ah it looks like this is just a standard error in the sense that methods are being bound that aren't actually supported. It looks like they're emitting `Option<JsValue>` as presumably `BufferSource` is pretty deep in a union argument. Perhaps this could be fixed by translating to `Object` instead of `JsValue`?

gnunicorn:
Okay, that seems to work. Now we have every method twice, once with `_with_object` and once `_with_u8_array`, similarly as it is with other versions. Still had to remove the `ArrayBufferView` return-value on `Crypto.get_random_values` though for it to compile (which makes sense), but I think that is fine as it is filling in the buffer you are passing it (shouldn't it therefore be marked as `mut` in the API?).

Ready for review :) .

crates/web-sys/webidls/enabled/Crypto.webidl
```diff
@@ -17,5 +17,5 @@ interface Crypto {
   readonly attribute SubtleCrypto subtle;
 
   [Throws]
-  ArrayBufferView getRandomValues(ArrayBufferView array);
+  void getRandomValues(ArrayBufferView array);
```
alexcrichton: Hm it's unfortunate that we need to change this here, could we perhaps change this though so in return position an `ArrayBufferView` is an `Object` but in argument position it's either an object or a u8 array?

gnunicorn: It was little awkward, but I got the parser to do just that: consider it a union for arguments, assume it to be an JS `Object` when returned.

gnunicorn:
@alexcrichton done. Also rebased and fixed the conflict that arose from that.

gnunicorn:
Attaching the docs:
screenshot-20180831200610-1132x1287
---

screenshot-20180831200558-501x1524

crates/webidl/src/idl_type.rs
```diff
@@ -466,7 +474,16 @@ impl<'a> IdlType<'a> {
             IdlType::Uint32Array => Some(array("u32", pos)),
             IdlType::Float32Array => Some(array("f32", pos)),
             IdlType::Float64Array => Some(array("f64", pos)),
-
+            IdlType::ArrayBufferView | IdlType::BufferSource => {
+                if pos == TypePosition::Argument {
+                    // ArrayBuffer and BufferSource can be represented as `Object` or &[u8]
+                    IdlType::Union(vec![IdlType::Object, IdlType::Uint8Array]).to_syn_type(pos)
```
alexcrichton: I think for a simplification this can probably be `Object` as well actually in the sense that the "flatten" expansion happens below and here it's for things like attributes and whatnot (which largely only work with `Object`)

gnunicorn: done.

gnunicorn:
Follow up issue (maybe we want to address that in a separate PR/issue): Currently the WebCrypto IDL (and thus our API) specifies the returning value of many as `Promise<Any>`, [however in practice this ends up being an `ArrayBuffer` for many cases](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/encrypt#Return_value). Now the issue: the way `JsValue` is currently implemented, it isn't actually possible to convert it into an `ArrayBuffer` or do anything with the content of it :( .

alexcrichton:
Hm interesting! Can't `JsValue` be converted into an `ArrayBuffer` with `dyn_into` and the `JsCast` trait?

Additionally I had one final thought recently, which is we may not actually be able to work with `&[u8]` unfortunately. One of the APIs here, `getRandomValues`, actually mutates that argument. In that sense it's a bit of a violation to pass in `&[u8]` and then have it all get updated :(

We could conservatively require `&mut [u8]` everywhere perhaps?

gnunicorn:
thanks for the tip `dyn_into::<Uint8Array>()` seems to ... compile :D (not actually tested yet).

Re `mut &[8]`: yes, I agree. However, IDL doesn't seems to have a notion of non-mutability, so in doubt we'd have to mark everything mutable? While this seems excessive, I don't think we have much of a choice, as JS is mutable by default and whenever we pass something into a JS-function, this might happen. So to keep rust as safe as can be, yes, it should prevent any changes when we passed them out, thus we should mark them as `mut &[u8]` ...

gnunicorn:
@alexcrichton I made all referenced types `&mut` now (not only `[u8]`), but I left the conditional change in there (second to last commit) so if we want to have `mut` only on some refs, we could still easily make that happen.

However, this now leads to a few API ... oddities... Like:

screenshot-20180903133126-916x701

or

screenshot-20180903133113-1088x831

---

the later feeling a lot more strange. not only because of `&mut str`, but also because we are now passing in the `CryptoKey` mutable, which clearly shouldn't change when en- or decrypting, but technically the underlying code _could_... I am just afraid this causes weird wrapping and way to many unnecessary copies down the line of API-users :(

crates/webidl/src/util.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) fn shared_ref(ty: syn::Type) -> syn::Type {
     syn::TypeReference {
         and_token: Default::default(),
         lifetime: None,
-        mutability: None,
+        mutability: Some(syn::token::Mut::default()),
```
alexcrichton: Oh I think this isn't quite where we'll want to do this because this makes *everything* a mutable reference. In general `&mut SomeJsType` doesn't actually make sense in that the instance of `SomeJsType` isn't being changed (the pointer we're using at least). The instance itself may be modified but that's sort of just the normal story with interior mutability in Rust. 

To that end I think we'll want a dedicated new type like `Uint8ArrayMut` which would be the only type that uses `&mut`

gnunicorn:
Okay, this now uses a newly created `Uint8ArrayMut` internally, which is the only `&mut` now. However, all `BufferArrayView` and `SourceBuffer` will be rendered as such, even if the spec doesn't actually expect them to be modified. However, this is the only way to do that if we don't want to modify the WebIDL specs.

This now creates:
screenshot-20180906161544-932x788
screenshot-20180906161528-933x736

alexcrichton:
Awesome, thanks!

PR #750, opened 2018-08-25
Add support for modules importing memory

alexcrichton:
The default of Rust wasm binaries is to export the memory that they contain, but
LLD also supports an `--import-memory` option where memory is imported into a
module instead. It's looking like importing memory is along the lines of how
shared memory wasm modules will work (they'll all import the same memory).

This commit adds support to wasm-bindgen to support modules which import memory.
Memory accessors are tweaked to no longer always assume that the wasm module
exports its memory. Additionally JS bindings will create a `memory` option
automatically because LLD always imports memory from an `env` module which won't
actually exist.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -43,6 +43,7 @@ pub struct Context<'a> {
     pub exported_classes: HashMap<String, ExportedClass>,
     pub function_table_needed: bool,
     pub interpreter: &'a mut Interpreter,
+    pub memory_init: Option<ResizableLimits>,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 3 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -1690,6 +1717,29 @@ impl<'a> Context<'a> {
     fn use_node_require(&self) -> bool {
         self.config.nodejs && !self.config.nodejs_experimental_modules
     }
+
+    fn memory(&mut self) -> &'static str {
+        if self.module.memory_section().is_some() {
+            return "wasm.memory"
```
fitzgen: Nitpick: missing `;`

PR #761, opened 2018-08-28
web-sys: Add support for `Global`-scope methods

### LINKED ISSUE #659: Expose a getter for the global window
I think we could do something like

```rust
#[wasm_bindgen]
extern {
    #[wasm_bindgen(getter)]
    pub fn window() -> Window;
}
```

inside `crates/web-sys/src/lib.rs` and consider that the entry point to all globally exposed properties and methods.

alexcrichton:
This commit adds further support for the `Global` attribute to not only emit
structural accessors but also emit functions that don't take `&self`. All
methods on a `[Global]` interface will not require `&self` and will call
functions and/or access properties on the global scope.

This should enable things like:

    Window::location() // returns `Location`
    Window::fetch(...) // invokes the `fetch` function

Closes #659

### DIFF
--- a/crates/webidl-tests/global.js
+++ b/crates/webidl-tests/global.js
@@ -1,0 +1,4 @@
+global.global_no_args = () => 3;
+global.global_with_args = (a, b) => a + b;
+global.global_attribute = 'x';
+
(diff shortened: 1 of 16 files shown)


crates/webidl-tests/global.js
```diff
@@ -0,0 +1,4 @@
+global.global_no_args = () => 3;
```
fitzgen: This should probably be

```js
global.global_no_args = function () {
    assert(this == global);
    return 3;
}
```

just to make sure we properly bind the global method to its global object.

PR #765, opened 2018-08-30
Change how naming overloaded methods works

alexcrichton:
This is a pretty large changeset instigated by an initial refactoring. The issue was originally discovered with the method [`put_image_data_with_imagedata_and_dx_and_dy`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.CanvasRenderingContext2D.html#method.put_image_data_with_imagedata_and_dx_and_dy). That, and the method currently below it, aren't ideally named, and this commit sets out to fix that!

We want to generally compress the names as much as possible to make them a bit nicer to call. To that end we ignore arguments that are the same among all methods of the same name. Unfortunately though this logic is only applied to *single* method definitions (with union parameters) rather than across method definitions.

This comit is a large refactoring which achieves this goal of deduplicating argument names across function definitions by:

* The `FirstPassRecord` and the first pass of the AST continues to collect information. This is now the only pass of the AST though, it's never traversed again.
* All operations during this first pass are recorded into hash maps, and these maps have enough data to create methods for the backend afterwards.
* The `WebIdlParse` trait has been removed in favor of just walking over the data members of the first pass object.

Notably this allows us to collect all same-named methods into the same data structure as part of the first pass, and then in the second pass, once we have all this information, we can deduce names for all of the generated bindings.

Along the way a good deal of code was either removed or simplified as a result of this architectural change. The `util.rs` module is attempting to consolidation function manufacturing as much as possible, but getters/setters are still distinct and need to be unified (future work).

Breaking changes in web-sys for this PR include:

* Names of functions will change (as seen above), but for the most part the new names should either be shorter and/or more clear.
* Functions with variadic arguments are not longer bound. It was discovered that we just bound these functions with a single argument, but this likely isn't what we want to do long-term!

Otherwise it's intended at least that no other breaking changes are contained within! In any case, curious to see what others think!


afdw:
Thanks for the refactoring, I was thinking about this direction of changes.

crates/webidl/src/first_pass.rs
```diff
@@ -71,7 +73,7 @@ pub(crate) struct DictionaryData<'src> {
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, Copy)]
 pub(crate) enum OperationId<'src> {
-    Constructor,
+    Constructor(IgnoreTraits<&'src str>),
```
afdw: This looks a bit confusing, especially with `IgnoreTraits`. `OperationId` meant to be an identified of operation, not a storage of data about the operation. Is it possible to move the JS class name to the `OperationData`?

alexcrichton: I was originally thinking that yeah but ended up feeling that it was best here because it only makes sense in the context of a constructor, not for other operations. I'd be fine either way though!

crates/webidl/src/first_pass.rs
```diff
@@ -207,37 +229,60 @@ fn first_pass_operation<'src>(
...
+
     let mut names = Vec::with_capacity(arguments.len());
     for argument in arguments {
         match argument {
             Argument::Single(single) => names.push(single.identifier.0),
             Argument::Variadic(variadic) => names.push(variadic.identifier.0),
         }
     }
-    let operations = match first_pass_operation_type{
+    let (operations, operations2) = match first_pass_operation_type{
```
fitzgen: nitpick: space before `{` or just rustfmt this badboy

crates/webidl/src/first_pass.rs
```diff
@@ -247,7 +292,16 @@ fn first_pass_operation<'src>(
             .entry(names.clone())
             .and_modify(|same_argument_names| *same_argument_names = true)
             .or_insert(false);
+        operations2.entry(*id)
+            .or_default()
+            .signatures
+            .push(Signature {
```
fitzgen: Ah, nice.

crates/webidl/src/util.rs
```diff
@@ -693,6 +638,149 @@ impl<'src> FirstPassRecord<'src> {
...
+            actual_signatures.push(ExpandedSig {
+                orig: signature,
+                args: Vec::with_capacity(signature.args.len()),
+            });
+
+            for (i, arg) in signature.args.iter().enumerate() {
+                let cur = actual_signatures.len();
+                // If any argument in this signature can't be converted we have
+                // to throw out the entire signature, so revert back to the
+                // beginning and then keep going.
```
fitzgen: Implementing the filtering out of unsupported arg types is making the logic harder to follow than necessary -- especially with the backtracking via truncate. Can we do a `filter_map` before we even start the `'outer` loop so that we aren't doing so much all at once?

crates/webidl/src/util.rs
```diff
@@ -693,6 +638,149 @@ impl<'src> FirstPassRecord<'src> {
...
+                    }
+                };
+
+                // The first arugment gets pushed directly in-place, but all
+                // future expanded arguments will cause new signatures to be
+                // created. If we have an optional argument then we consider the
+                // already existing signature as the "none" case and the flatten
+                // below will produce the "some" case, so we've already
+                // processed the first argument effectively.
+                let mut first = !arg.optional;
```
fitzgen: the `first` variable = whether or not this is in the set of arguments that all signatures for this method share, right?

The comment doesn't quite match that, and says "the first argument" blah blah, when it is really "the set of common arguments" blah blah, right? We should definitely clarify this for future readers!

crates/webidl/src/util.rs
```diff
@@ -693,6 +638,149 @@ impl<'src> FirstPassRecord<'src> {
...
+            }
+            OperationId::IndexingGetter => "get",
+            OperationId::IndexingSetter => "set",
+            OperationId::IndexingDeleter => "delete",
+        };
+
+        let mut ret = Vec::new();
+        for signature in actual_signatures.iter() {
+            // Ignore signatures with invalid return types
+            let ret_ty = match signature.orig.ret.to_idl_type(self) {
```
fitzgen: I don't think that returns are overloaded (right?) so if we did the return type filtering before the signature expansion, then we could potentially avoid doing work that will get thrown away.

crates/webidl/src/util.rs
```diff
@@ -693,6 +638,149 @@ impl<'src> FirstPassRecord<'src> {
...
+                    first = false;
+                } else {
+                    rust_name.push_str("_and_");
+                }
+
+                // If this name of the argument for this signature is unique
+                // then that's a bit more human readable so we include it in the
+                // method name. Otherwise the type name should disambiguate
+                // correctly.
+                if !any_same_name {
```
fitzgen: It is generally more readable to do `if foo { a() } else { b() }` rather than `if !foo { b() } else { a() }`

crates/webidl/src/util.rs
fitzgen: Readign some more, I think I am misunderstanding what this loop does, because what I thoguth it does is handled below. Now I am not really sure what is up with this.

crates/webidl/src/lib.rs
```diff
@@ -453,61 +453,61 @@ impl<'src> FirstPassRecord<'src> {
...
+    //         },
+    //         IdlType::Interface(iface_name),
+    //         kind,
+    //         structural,
+    //         throws,
+    //         None,
+    //     ) {
+    //         program.imports.push(wrap_import_function(import_function));
+    //     }
+    // }
```
fitzgen: I'm assuming all this will be removed in follow up commits?

examples/julia_set/Cargo.toml
```diff
@@ -8,4 +8,3 @@ crate-type = ["cdylib"]
 
 [dependencies]
 wasm-bindgen = { path = "../.." }
-web-sys = { path = "../../crates/web-sys" }
```
fitzgen: Wait -- why?

examples/julia_set/src/lib.rs
```diff
@@ -1,13 +1,18 @@
...
 mod fractal;
 use fractal::get_julia_set;
 use fractal::complex::Complex;
 
+#[wasm_bindgen]
+extern "C" {
+    pub type ImageData;
+
+    #[wasm_bindgen(constructor)]
+    pub fn new(arr: &Uint8ClampedArray, width: u32, height: u32) -> ImageData;
```
fitzgen: Why are we removing `web-sys` usage from this example?

examples/console_log/src/lib.rs
```diff
@@ -22,7 +18,4 @@ pub fn run() {
     log("Hello from Rust!");
     log_u32(42);
     log_many("Logging", "many values!");
-
-    console::log(JsValue::from("Another message from rust!"));
```
fitzgen: Again: why?

crates/webidl/src/first_pass.rs
```diff
@@ -261,84 +289,123 @@ impl<'src> FirstPass<'src, ()> for weedle::InterfaceDefinition<'src> {
...
-                if list.lhs_identifier.0 == "NamedConstructor" =>
-            {
-                first_pass_operation(
-                    record,
-                    FirstPassOperationType::Interface,
-                    self_name,
-                    &[OperationId::Constructor],
-                    &list.args.body.list,
-                )
+            InterfaceMember::Const(const_) => {
```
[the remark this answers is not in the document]
alexcrichton: I thought so too! Unfortunately though mixin/interface consts get different treatment but weedle parses them both as `ConstMember`, so I had to inline the different logic

crates/webidl/src/util.rs
alexcrichton: Nah your first comment makes sense! I think I've clarified it now though

crates/webidl/src/util.rs
alexcrichton: Yeah I wasn't sure about that, I'll leave a TODO for future work

crates/webidl/src/util.rs
alexcrichton: Makes sense yeah (english made sense the other way to me) but in any case this ended up changing by the end

crates/webidl/src/lib.rs
alexcrichton: indeed!

examples/julia_set/Cargo.toml
alexcrichton: Ah we talked about this on IRC but I was overzealous at the start of this PR and this ended up not working out, so this is me backing out intermediate changes

examples/console_log/src/lib.rs
alexcrichton: This one is because variadic support has been removed (as it was erroneously there before!)

afdw: Yes, I have accidentally added variadic support while unifying how single and variadic arguments work, but I thee decided to keep it like this because many operations with variadic arguments are useful even if you actually can call them using only a single argument.

alexcrichton: It's true yeah a single argument works, but I think we want to discuss this a bit to figure out what future support for these APIs will look like and avoid having our hands tied by having an existing convention

crates/webidl/src/first_pass.rs
alexcrichton: Perhaps! I'm gonna go ahead this PR and leave this for a future cleanup, but `self_name` in the original `first_pass_operation` method is always the same for constructors so there's no clear entry point for this special case with renamed constructors. We can always continue to iterate on this in tree though!

PR #772, opened 2018-08-30
Add `links` section to `Cargo.toml`

fitzgen:
Because only a single `wasm_bindgen` version can be used in a dependency graph, pretend we link a native library so that `cargo` will provide better error messages than the esoteric linker errors we would otherwise trigger.


alexcrichton:
Looks like CI is forcing us to have a build scripts?

fitzgen:
Added an empty build.rs file.

PR #785, opened 2018-09-02
Add initial support for unions in return types, add more fixes for case of identifiers

afdw:
I have tried to add a better support for unions return types using enums with payload, but it turns out that it would be a bit harder than I have expected (we need a way for passing enums with payload and we need a way of identifying type that JS function returned). `push_camel_case_name` is useful for generating names for such enums.

So I have decided to just return `JsValue`, as there are not many such operations:
```
"FormData" Operation(Some("get")) Nullable(Union([Interface("Blob"), Interface("Directory"), UsvString]))
"FormData" Operation(Some("getAll")) Sequence(Union([Interface("Blob"), Interface("Directory"), UsvString]))
"DataTransfer" Operation(Some("getFilesAndDirectories")) Promise(Sequence(Union([Interface("File"), Interface("Directory")])))
"HTMLAllCollection" Operation(Some("item")) Nullable(Union([Interface("Node"), Interface("HTMLCollection")]))
"HTMLAllCollection" Operation(Some("namedItem")) Nullable(Union([Interface("Node"), Interface("HTMLCollection")]))
"HTMLAllCollection" IndexingGetter Nullable(Union([Interface("Node"), Interface("HTMLCollection")]))
"HTMLFormControlsCollection" Operation(Some("namedItem")) Nullable(Union([Interface("RadioNodeList"), Interface("Element")]))
"HTMLFormControlsCollection" IndexingGetter Nullable(Union([Interface("RadioNodeList"), Interface("Element")]))
"Directory" Operation(Some("getFilesAndDirectories")) Promise(Sequence(Union([Interface("File"), Interface("Directory")])))
```


afdw:
It turns that sometimes there are nullable types with unions inside them. But `Option<JsValue>` is not `IntoWasmAbi` or `FromWasmAbi`. Is it possible to implement it like this:
```rust
impl OptionIntoWasmAbi for JsValue {
    #[inline]
    fn none() -> <JsValue as IntoWasmAbi>::Abi {
        ::JSIDX_UNDEFINED
    }
}

impl OptionFromWasmAbi for JsValue {
    #[inline]
    fn is_none(abi: &<JsValue as FromWasmAbi>::Abi) -> bool {
        *abi == ::JSIDX_UNDEFINED || *abi == ::JSIDX_NULL
    }
}
```
It compiles with this change. But doc of the `none` function says:
```rust
    /// Returns an ABI instance indicating "none", which JS will interpret as
    /// the `None` branch of this option.
    ///
    /// It should be guaranteed that the `IntoWasmAbi` can never produce the ABI
    /// value returned here.
```
But `IntoWasmAbi` can produce this ABI value on `JsValue::UNDEFINED`. It means that `None` and `Some(JsValue::UNDEFINED)` would be serialized into the same ABI value, which is not good.

The other approach is to always return `JsValue` regardless of how it is nested.

afdw:
I have implemented the second approach: nullable unions become just `JsValue`.

alexcrichton:
Interesting, thanks for the PR!

So a sort of crazy idea that could also work. What if, instead of `JsValue`, we take the "greatest common ancestor" of all the types in a union? `JsValue` is sort of hard to work with in that it includes primitives and, as you've seen, doesn't implement conversions for `Option` of itself. Almost all of these unions, however, can probably be `Object` instead, and sometimes even more specific if there's a common ancestor.

I think if we don't implement that sort of logic which may want to instead stick with `Object` unless we find a primitive (requiring `JsValue`) in the list.

afdw:
Yes, but this PR only adds an initial support so that these few functions can be called at all. So would it be useful to implement better temporary support, or we can wait until it is possible to have a better one? If yes, I will try to implement it here.

alexcrichton:
In terms of temporary support I think I'd slightly prefer `Object` so we can translate a nullable union to `Option<...>` (are there any unions with primitives though or nullable unions with primitives?)

Overall though we're close to publishing web-sys and a change like this would be a breaking change, so it may be good to have a semi-final implementation! (`Object` I think is "semi-final enough", whereas finding the common ancestor is probably much more complicated and may not actually be too beneficial)

afdw:
OK, done I think.

crates/webidl/src/idl_type.rs
```diff
@@ -336,8 +335,88 @@ terms_to_idl_type! {
 }
 
 impl<'a> IdlType<'a> {
+    /// Generates a camel case type name.
+    #[allow(dead_code)]
```
alexcrichton: Since this is dead, should we leave it omitted for now?

crates/webidl/src/idl_type.rs
```diff
@@ -478,7 +549,24 @@ impl<'a> IdlType<'a> {
             },
             IdlType::Enum(name) => Some(ident_ty(rust_ident(camel_case_ident(name).as_str()))),
 
-            IdlType::Nullable(idl_type) => Some(option_ty(idl_type.to_syn_type(pos)?)),
+            IdlType::Nullable(idl_type) => {
+                match **idl_type {
+                    IdlType::Union(ref idl_types) =>
+                        if idl_types
+                            .iter()
+                            .all(|idl_type|
```
alexcrichton: To confirm, is this ever actually `false`? If so it seems fine to keep the branch in, but if not I think it'd be fine to change this to an `assert!` and always emit an `Object`

afdw: Yes, unions can contain not only interfaces. See, for example, `BasePropertyIndexedKeyframe` or `IDBObjectStoreParameters`.

alexcrichton: Ah ok, good to know!

alexcrichton:
👍

PR #787, opened 2018-09-03
Update wasm2es6js.rs

bspeice:
Binaryen renamed the tool to wasm2js instead of wasm2asm - https://github.com/WebAssembly/binaryen/pull/1642

### DIFF
--- a/crates/cli-support/src/wasm2es6js.rs
+++ b/crates/cli-support/src/wasm2es6js.rs
@@ -11,14 +11,14 @@ use parity_wasm::elements::*;
 
 pub struct Config {
     base64: bool,
-    wasm2asm: bool,
+    wasm2js: bool,
     fetch_path: Option<String>,
 }
 
 pub struct Output {
     module: Module,
     base64: bool,
-    wasm2asm: bool,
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I think though that `--wasm2asm` is basically a deprecated mode now that `wasm2js` exists, as the output of `wasm2js` shoudl be equivalent to `wasm2es6js --wasm2asm`. If you use the `wasm2js` tool though is it not working for you?

bspeice:
Not totally sure what you mean - are you referring to keeping the `--wasm2asm` flag live in `wasm-bindgen` and just changing the internal call to `binaryen` to use `wasm2js`?

bspeice:
Ah, that makes more sense. I'm actually mostly in favor of deleting the flag; if it's just a shim around `binaryen` I'd rather reduce functionality and make a note in the docs that functionality provided by `wasm2asm` is now a part of `wasm2js` in `binaryen`. I'll get started on that unless I hear otherwise.

alexcrichton:
Sounds good to me!

PR #790, opened 2018-09-05
Gate `web-sys` APIs on activated features

alexcrichton:
Currently the compile times of `web-sys` are unfortunately prohibitive,
increasing the barrier to using it. This commit updates the crate to instead
have all APIs gated by a set of Cargo features which affect what bindings are
generated at compile time (and which are then compiled by rustc). It's
significantly faster to activate only a handful of features vs all thousand of
them!

A magical env var is added to print the list of all features that should be
generated, and then necessary logic is added to ferry features from the build
script to the webidl crate which then uses that as a filter to remove items
after parsing. Currently parsing is pretty speedy so we'll unconditionally parse
all WebIDL files, but this may change in the future!

For now this will make the `web-sys` crate a bit less ergonomic to use as lots
of features will need to be specified, but it should make it much more
approachable in terms of first-user experience with compile times.


fitzgen:
FYI: lots of CI failures

crates/web-sys/Cargo.toml
```diff
@@ -22,3 +22,1156 @@ js-sys = { path = '../js-sys', version = '0.2.4' }
...
+# compiling this crate which dumps the total list of features to a file called
+# `foo`.
+#
+# Each one of these features activates the corresponding type, allowing bindings
+# to be generated for it. Note that we may eventually add "groupings" of
+# features to enable a convenient set of features all at once. For now, though,
+# the features must all be manually activated.
+[features]
+AbortController = []
+AbortSignal = []
```
fitzgen: Yeeeeeaaaaaaah!

crates/webidl/src/lib.rs
```diff
@@ -104,43 +109,70 @@ fn parse(webidl_source: &str) -> Result<backend::ast::Program> {
...
-    );
+    )
+}
+
+/// Run codegen on the AST to generate rust code.
+fn compile_ast(mut ast: backend::ast::Program) -> String {
+    // Iteratively prune all entries from the AST which reference undefined
+    // fields. Each pass may remove definitions of types and so we need to
+    // reexecute this pass to see if we need to keep removing types until we
+    // reach a steady state.
```
fitzgen: One might describe the pruning as a monotonic function, and iterating to a fixed point with this monotonic function as a "monotone framework"... ;)

alexcrichton:
Ok I think CI is green, AppVeyor seems to have terminated but doesn't explain why...

crates/webidl/src/lib.rs
```diff
@@ -555,8 +596,39 @@ impl<'src> FirstPassRecord<'src> {
...
+        &self,
+        item: impl ImportedTypeReferences,
+        doc: &mut Option<String>,
+    ) {
+        let doc = match doc {
+            Some(doc) => doc,
+            None => return,
+        };
+        let mut required = BTreeSet::new();
+        item.imported_type_references(&mut |f| {
```
fitzgen: Super happy that this ended up being generally useful!

crates/web-sys/Cargo.toml
fitzgen: Oh also: this comment about features should probably go into the README and/or src/lib.rs too

crates/webidl/src/lib.rs
alexcrichton: 🎓

PR #793, opened 2018-09-06
Bump to 0.2.20

alexcrichton:


fitzgen:
Don't want to wait for the closures PR? Any reason in particular for doing a release now rather than some other time?

alexcrichton:
Ah the motivation here was remembering that the 0.2.19 binaries for Windows weren't working so with an upcoming wasm-pack release to download precompiled binaries I wanted to make sure those were fixed. I'm fine waiting a bit though for the closures PR!

PR #797, opened 2018-09-07
Implement the `Iterator` trait for JS iterators

### LINKED ISSUE #777: Implement Rust's `Iterator` trait for `js_sys::Iterator`
with `type Item = Result<JsValue, JsValue>;` meaning that it unpacks the value from the returned `js_sys::IteratorNext`

alexcrichton:
This commit implements the standard library's `Iterator` trait for the
`js_sys::Iterator` type, using the iterator protocol described on [MDN]

Closes #777

[MDN]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1174,6 +1174,79 @@ extern {
     pub fn next(this: &Iterator) -> Result<IteratorNext, JsValue>;
 }
 
+pub struct Iter<'a> {
+    js: &'a Iterator,
+    state: IterState,
+}
+
+pub struct IntoIter {
+    js: Iterator,
+    state: IterState,
+}
+
+struct IterState {
+    done: bool,
+}
+
+impl<'a> IntoIterator for &'a Iterator {
+    type Item = Result<JsValue, JsValue>;
+    type IntoIter = Iter<'a>;
+
+    fn into_iter(self) -> Iter<'a> {
+        Iter { js: self, state: IterState::new() }
+    }
+}
+
+impl<'a> std::iter::Iterator for Iter<'a> {
+    type Item = Result<JsValue, JsValue>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.state.next(self.js)
+    }
+}
+
+impl IntoIterator for Iterator {
+    type Item = Result<JsValue, JsValue>;
+    type IntoIter = IntoIter;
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -1174,6 +1174,20 @@ extern {
...
 }
 
+impl std::iter::Iterator for Iterator {
+    type Item = JsValue;
+
+    fn next(&mut self) -> Option<JsValue> {
+        let val = Iterator::next(&*self)
+            .expect("JS object didn't obey iterator protocol");
+        if val.done() {
+            None
```
fitzgen: I *thought* this was wrong, and that if both done and value are present (happens when returning from a generator), that we should add the value to the iterator, but it seems that it doesn't get into the iterable in JS:

done-and-value

Carry on!

crates/js-sys/src/lib.rs
```diff
@@ -1174,6 +1174,20 @@ extern {
     pub fn next(this: &Iterator) -> Result<IteratorNext, JsValue>;
 }
 
+impl std::iter::Iterator for Iterator {
+    type Item = JsValue;
+
+    fn next(&mut self) -> Option<JsValue> {
+        let val = Iterator::next(&*self)
+            .expect("JS object didn't obey iterator protocol");
```
fitzgen: Iterables (such as generators) can totally throw, and it isn't really violating the protocol :-/

generator-throws

I think we should have `Iterator<Item = Result<JsValue, JsValue>>` and then fuse the iterator such that calling next after the first `Err` always returns `None` and doesn't call back into JS again.

alexcrichton: Interesting! 

I don't think we could auto-fuse though unless we start setting a JS property on the iterator (which seems icky) or have a wrapper in Rust. I'd be ok returning a `Result<JsValue, JsValue>` but that would make fusing difficult for sure

fitzgen: I think the fusing is important for combinators and `.collect()` to work correctly.

Perhaps we should implement `IntoIterator` for `&'a js_sys::Iterator` witha  custom type that has a `bool` for fusing?

alexcrichton:
`IntoIterator` sounds like a great idea to me, updated to that!

PR #798, opened 2018-09-07
Event listeners are also callbacks

fitzgen:
Not sure when it makes sense to land this -- once we support callback interfaces, so we can actually test this?

### DIFF
--- a/crates/web-sys/webidls/enabled/EventListener.webidl
+++ b/crates/web-sys/webidls/enabled/EventListener.webidl
@@ -10,6 +10,8 @@
  * liability, trademark and document use rules apply.
  */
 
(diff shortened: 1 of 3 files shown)


fitzgen:
Also, these method name manglings will suck :-/

alexcrichton:
Hm so before we commit to this, how does this work in browsers? How does Gecko accept a function instead of just the callback interface?

fitzgen:
> How does Gecko accept a function instead of just the callback interface?

Asking some firefox DOM folks...

fitzgen:
Hm, actualyl I think we might not want to merge this: https://heycam.github.io/webidl/#es-user-objects

PR #803, opened 2018-09-10
Change how filtering is done in WebIDL

### LINKED ISSUE #802: web-sys: required Cargo features aren't exhaustive
For example `Window::fetch_with_str`'s [requirement list](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Window.html#method.fetch_with_str) only includes the `Window` feature but to actually get access to the  function, you need at least `Window`, `Request`, and `Response`.

alexcrichton:
Instead of actually modifying the `FirstPassRecord` let's instead just skip
relevant entries when we come across them. This should help us retain knowledge
that `optional SomeImportedType arg` can be bound even though `SomeImportedType`
may not exist.

One small tweak was needed to modify the AST afterwards to remove `extends`
annotations which aren't actually defined, but other than that this should...

Closes #802

### DIFF
--- a/crates/backend/src/defined.rs
+++ b/crates/backend/src/defined.rs
@@ -289,6 +289,9 @@ impl ImportedTypes for ast::ImportType {
         F: FnMut(&Ident, ImportedTypeKind),
     {
         f(&self.rust_name, ImportedTypeKind::Definition);
+        for class in self.extends.iter() {
+            f(class, ImportedTypeKind::Reference);
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


fitzgen:
https://travis-ci.org/rustwasm/wasm-bindgen/jobs/426774974#L990-L1006

```
error[E0432]: unresolved import `web_sys::RequestInit`
-->examples/fetch/src/lib.rs:12:24
|
12|use web_sys::{Request, RequestInit, RequestMode, Response, Window};
|^^^^^^^^^^^no `RequestInit` in the root
error[E0599]: no function or associated item named `new_with_str_and_init` found for type `web_sys::Request` in the current scope
-->examples/fetch/src/lib.rs:49:15
|
49|    let req = Request::new_with_str_and_init("https://api.github.com/repos/rustwasm/wasm-bindgen/branches/master", &request_options).unwrap();
|^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^function or associated item not found in `web_sys::Request`
Some errors occurred: E0432, E0599.
For more information about an error, try `rustc --explain E0432`.
error: Could not compile `fetch`.
```

PR #804, opened 2018-09-10
Expand LongLong to `(i32 or f64)` instead of `i64`

### LINKED ISSUE #800: BigInt doesn't play nicely with WebGL bindings
I've been experimenting with the `WebGLRenderingContext` bindings, and I've discovered that there's a weird interaction between the WebIDL types, `wasm-bindgen`'s type mapping, and what browsers actually support in their WebGL implementations. Basically, there are several methods which take a `GLintptr`, which is a `long long`, which `wasm-bindgen` maps to `i64` and `BigInt`. But the actual API throws an exception if you pass it a `BigNum`. Using an `i32` or `u32` seems to work fine.

The simple solution would seem to be changing the definition of `GLintptr` in the WebIDL, but is there another alternative that isn't so hacky?

alexcrichton:
This commit tweaks WebIDL expansion of the "long long" and "unsigned long long"
types to expand to a union of an 32-bit integer and a double. This reflects how
almost none of the APIs on the web today actually work with a `BigInt` (what the
previous Rust type of `i64` translates to) and how JS itself fundamentally
operates with these APIs.

Eventually this may not be necessary if we can natively connect to C++ engines
with the `i64` type, but until that day comes this should provide more useful
interfaces as they shoudl work in all browsers.

Closes #800

### DIFF
--- a/crates/webidl-tests/consts.rs
+++ b/crates/webidl-tests/consts.rs
@@ -26,11 +26,6 @@ fn ints() {
     assert_eq!(ConstLong::IMAX, i32::max_value());
     assert_eq!(ConstLong::UMIN, u32::min_value());
     assert_eq!(ConstLong::UMAX, u32::max_value());
-
-    assert_eq!(ConstLongLong::IMIN, i64::min_value());
-    assert_eq!(ConstLongLong::IMAX, i64::max_value());
-    assert_eq!(ConstLongLong::UMIN, u64::min_value());
-    assert_eq!(ConstLongLong::UMAX, u64::max_value());
 }
 
 #[wasm_bindgen_test]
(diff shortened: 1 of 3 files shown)


fitzgen:
https://travis-ci.org/rustwasm/wasm-bindgen/jobs/426788978#L647

```
error[E0599]: no associated item named `IMIN` found for type `consts::ConstLongLong` in the current scope
-->crates/webidl-tests/consts.rs:30:16
|
30|    assert_eq!(ConstLongLong::IMIN, i64::min_value());
|^^^^^^^^^^^^^^^^^^^associated item not found in `consts::ConstLongLong`
```

PR #805, opened 2018-09-10
Implement WebIDL callback interfaces

### LINKED ISSUE #258: Support WebIDL callback interfaces
https://heycam.github.io/webidl/#dfn-callback-interface

These are like dictionaries (structural, not nominal; user-defined, not host defined) for sets of callback functions.

Example:

```webidl
callback interface OpenCloseListener {
  void onOpen();
  void onClose();
};
```

These should probably be translated into traits, and then operations that take or return these should take `Box<OpenCloseListener>`.

alexcrichton:
This commit implements callback interfaces for WebIDL, the final WebIDL
construct that we were unconditionally ignoring! Callback interfaces are
implemented as dictionaries of callbacks. Single-operation callback interfaces
are also expanded when flattening to accept a `Function` as well, in accordance
with the WebIDL spec.

New features have been added to `web-sys` for all the new callback interface
types. Additionally the `EventTarget.webidl` was tweaked to not have
`EventListener?` as this is required for all functional usage and there's no
need to keep that sort of web browser compat here.

Closes #258


crates/webidl/src/first_pass.rs
```diff
@@ -690,6 +691,25 @@ impl<'src> FirstPass<'src, ()> for weedle::CallbackDefinition<'src> {
     }
 }
 
+impl<'src> FirstPass<'src, ()> for weedle::CallbackInterfaceDefinition<'src> {
+    fn first_pass(&'src self, record: &mut FirstPassRecord<'src>, _: ()) -> Result<()> {
+        if util::is_chrome_only(&self.attributes) {
+            return Ok(())
+        }
+        if self.inheritance.is_some() {
+            warn!("skipping callback interface with inheritance: {}",
```
fitzgen: Do we ever hit this in our webidls?

alexcrichton: Thankfully no! I made sure to check that before writing this :)

Issue #616, opened 2018-08-02
User-facing documentation overhaul


fitzgen:
As discussed in today's WG meeting, we are aiming for a docs overhaul / focus on being a good reference for the Rust 2018 Release Candidate milestone. That's 6 weeks from now: 2018-09-13.

Use case: generally, when I am looking up wasm-bindgen docs, I am not trying to read a whole thing, I am trying to figure out whether some specific X is supported, or how Y rust thing gets translated into JS or vice versa. Our docs should have *great* reference material for quickly answering those kinds of queries.

* [x] Split existing user-facing docs into two sections: https://github.com/rustwasm/wasm-bindgen/pull/635
  1. Prose-y whirlwind tutorial
  2. Bullet-point-y reference material

* [x] Audit CLI flags reference and make sure it is detailed and complete: https://github.com/rustwasm/wasm-bindgen/pull/637
  * Should note that usually one should use `wasm-pack` instead of running wasm-bindgen CLI by hand

* [x] Audit attributes reference and make sure it is detailed and complete
  * When I place this attribute on a struct, what does that mean? What about on a function? Which attributes are valid on which things?
  * #661 

* [x] Supported types reference
  * [x] Audit for coverage completeness: https://github.com/rustwasm/wasm-bindgen/pull/637
  * [x] Should include strategy/implementation for translation where appropriate (eg, note that `String` involves making copy)
  * [x] should be able to answer "how does X rust thing translate into JavaScript, and how do I do that?"

* [x] Audit/create specific features reference sections:
  * [x] Promises and Futures: https://github.com/rustwasm/wasm-bindgen/pull/695
  * [x] Arbitrary data with Serde: #664 
  * [x] Rust Closures: #661 
  * [x] JavaScript Closures: #661 

* [x] Include domain-specific examples of using `web-sys` in the guide (e.g. webgl, webaudio, dom manipulation, etc)

* [x] Add a "Testing with `wasm-bindgen-test`" section
  * [x] Should include isntructions for getting CI set up

* [x] How to read/write arbitrary properties on a JS object

* [x] Working with Duck-Typed Interfaces

Anything I am missing?

+cc @ashleygwilliams @alexcrichton

alexcrichton:
I think that just about covers it all!

fitzgen:
More of this just landed in #661

PR #812, opened 2018-09-11
guide: Add section on accessing properties of untyped values

fitzgen:
Part of #616


alexcrichton:
I wonder if it would also be beneficial to show usage of the `structural` attribute here? I've used that historically to access data I know has the same "shape" and it should allow some different types as well (like slices).

Something like that may also belong elsewhere in the guide though too

fitzgen:
Yeah, I think there is a place for another "Working with Duck-Typed Interfaces" section in addition to what we have now.

alexcrichton:
Sounds good to me!

PR #815, opened 2018-09-12
Production example

Tarnadas:


alexcrichton:
Thanks for the PR! I think I'm slightly confused, though, of how much we want to show off here. I definitely agree that we should be showing off a good default set of production settings for common bundlers like webpack (maybe rename this to `production-webpack`?), but I'm not sure how much we want to dive into other likely opinionated topics like npm/yarn, how to structure a project, how to serve files, etc.

In that sense I wonder if it makes sense to perhaps trim this down to just what you'd expect to be run with webpack? Maybe just the production config needed there to generate an appropriately sized wasm file and such to ship? Or is that detracting too much from the example?

examples/production/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "target": "esnext",
```
eminence: I'm curious if it's possible to get this example working with an "ES6" target, instead of "ESNext"

Tarnadas: The reason behind ESNext target is the concatenation with Babel. There is no reason for TypeScript to target anything lower than ESNext since Babel does the next transpilation step anyway

Tarnadas:
You're right about Yarn, this should be removed and replaced with npm.

I am also unsure about the name and what should be shown, but I think that *something* should be used for serving files and replacing Express with another backend shouldn't be too hard.

There's also a big difference whether you want to use TypeScript or not, maybe a name including `webpack` and `typescript` would be better. On the other side `wasm-bindgen` generates TypeScript typings by default, which could be called opinionated as well. Due to the lack of proper watch mode for JavaScript and Rust files, I included `nodemon` for watching and building Rust files. Should that be kept or moved somewhere else?

examples/production/tsconfig.json
eminence: Just for my own education, can you explain what babel is doing here?   (In my own personal and small typescript projects, I've been able to directly use the output from the typescript compiler, and haven't needed an additional transpilation step)

Tarnadas:
```js
        {
          test: /\.tsx?$/,
          exclude: /node_modules/,
          use: [
            {
              loader: 'babel-loader',
              ...
            },
            {
              loader: 'awesome-typescript-loader'
            }
          ]
        }
```
Webpack does the chaining from bottom to top, so it passes the output from `awesome-typescript-loader` to `babel-loader`. Babel is just more powerful in the transpilation step and can do a lot more than TSC.
You can e.g. use [browserlist](http://browserl.ist/) via `browsers: [  '>1%'  ]` to just transpile specific features, which aren't supported by these browsers, resulting in more specialized and therefore better transpiled files.
This is also true for the polyfill step done by Babel. `useBuiltIns: 'entry'` will look into the browserlist and just give you the polyfills you really need and use. Note that I haven't added `babel-polyfill` as a dependency here, which must either be done at the index.ts file or within the entry chunk definition.

alexcrichton:
This came up during our WG discussion today, and @ashleygwilliams brought up a good point that an example like this may actually cause *confusion* because when you first look at it then it appears like wasm-bindgen need some sort of special configuration for wasm-bindgen to work with webpack. It's in fact the opposite, though, in that webpack shouldn't need any configuration to get working with wasm-bindgen!

In that sense is there perhaps a section of the documentation that we can write down referencing official webpack docs? It'd be great if we could say something like "take a look at these official docs, and maybe do a change like $this to plug in wasm". That way it would be more clear that very little configuration is necessary for wasm-bindgen code!

chinedufn:
On that note ^ -> maybe move this to `Tarnadas/rust-wasm-starter` ?

And maybe have a book section that links to community projects.. or something.. This way more opinionated things can still get a little sunlight even if they don't belong into `wasm-bindgen` proper.

alexcrichton:
Ok I'm gonna close this as it's been quiet for quite some time now, but we can always of course consider resubmission!

PR #816, opened 2018-09-12
Remove some crufty AppVeyor configuration

alexcrichton:
The `crt-static` feature should be long since stable as well as incremental
compilation.


sendilkumarn:
Thanks @alexcrichton

PR #818, opened 2018-09-12
guide: Add documentation for testing with `wasm-bindgen-test`

### LINKED ISSUE #616: User-facing documentation overhaul

fitzgen:
Fixes #616


guide/src/wasm-bindgen-test/browsers.md
```diff
@@ -0,0 +1,69 @@
+# Testing in Headless Browsers
+
+## Configure Your Test Crate
+
+Add this to the root of your test crate, e.g. `$MY_CRATE/tests/wasm.rs`:
+
+```rust
+wasm_bindgen_test_configure!(run_in_browser);
```
alexcrichton: This may want to use `use wasm_bindgen_test::wasm_bindgen_test_configure;` just above it (to bring the macro into scope)

guide/src/wasm-bindgen-test/continuous-integration.md
```diff
@@ -0,0 +1,73 @@
...
+INSTALL_GECKODRIVER: &INSTALL_GECKODRIVER
+  |
+    curl --retry 5 -LO https://github.com/mozilla/geckodriver/releases/download/v0.21.0/geckodriver-v0.21.0-linux64.tar.gz
+    tar xf geckodriver-v0.21.0-linux64.tar.gz
+
+INSTALL_CHROMEDRIVER: &INSTALL_CHROMEDRIVER
+  |
+    curl --retry 5 -LO https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip
+    unzip chromedriver_linux64.zip
+    export CHROMEDRIVER_ARGS=--verbose
```
alexcrichton: Technically this shouldn't be required

guide/src/wasm-bindgen-test/continuous-integration.md
```diff
@@ -0,0 +1,73 @@
...
+request!](https://github.com/rustwasm/wasm-bindgen)
+
+## Travis CI
+
+```yaml
+INSTALL_NODE_VIA_NVM: &INSTALL_NODE_VIA_NVM
+  |
+    curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash
+    source ~/.nvm/nvm.sh
+    nvm install v10.5
```
alexcrichton: I think this isn't necessarily required actually, the version of `node` installed automatically on Travis should already have wasm support

guide/src/wasm-bindgen-test/continuous-integration.md
```diff
@@ -0,0 +1,73 @@
...
+
+INSTALL_GECKODRIVER: &INSTALL_GECKODRIVER
+  |
+    curl --retry 5 -LO https://github.com/mozilla/geckodriver/releases/download/v0.21.0/geckodriver-v0.21.0-linux64.tar.gz
+    tar xf geckodriver-v0.21.0-linux64.tar.gz
+
+INSTALL_CHROMEDRIVER: &INSTALL_CHROMEDRIVER
```
alexcrichton: Instead of using this yaml syntax it may be best to inline it directly into the `install` section below to avoid confusing those who don't know about the syntax (which is probably almost everyone)

guide/src/wasm-bindgen-test/usage.md
```diff
@@ -0,0 +1,107 @@
...
+}
+
+#[wasm_bindgen_test]
+fn fail() {
+    assert_eq!(1, 2);
+}
+```
+
+Writing tests is the same as normal Rust `#[test]`s, except we are using the
+`#[wasm_bindgen_test]` attribute.
```
alexcrichton: This might be a good spot to mention `pub mod ...`

guide/src/wasm-bindgen-test/continuous-integration.md
fitzgen: Node comes with travis, even for non-node/js projects?

alexcrichton: I believe so! I could be wrong though, I'd have to check...

guide/src/wasm-bindgen-test/continuous-integration.md
```diff
@@ -0,0 +1,63 @@
...
+language: rust
+rust: nightly
+
+addons:
+  firefox: latest
+  chrome: stable
+
+install:
+  - rustup target add wasm32-unknown-unknown
+  - cargo install wasm-bindgen-cli
```
chinedufn: I'd imagine that a bunch of people will just copy this so ideally this would be a fast-enough starting point. 

This can take minutes, so could it make sense to instead either:

option 1. Download a release binary and move it to either `~/.cargo/bin` or `/usr/local/bin`

option 2. Cache cargo deps (I see that the `wasm-bindgen` repo's travis seems to do this?)

Happy to do either if you think it makes sense!

fitzgen: @chinedufn totally -- good call! I think option 1 is more straight forward, and we could also have a comment mentioning something like

> Download and unpack the `wasm-bindgen` CLI tools. Alternatively, use `wasm-pack` to manage `wasm-bindgen` binaries for you.

Thanks!

chinedufn: Thanks for the guidance!

For posterity https://github.com/rustwasm/wasm-bindgen/pull/824

PR #835, opened 2018-09-16
Create basic WebGL example

blm768:
This example doesn't do much, but it gets a triangle on the screen.

### DIFF
--- a/examples/webgl/src/lib.rs
+++ b/examples/webgl/src/lib.rs
@@ -1,0 +1,121 @@
+extern crate js_sys;
+extern crate wasm_bindgen;
+extern crate web_sys;
+
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::JsCast;
+use web_sys::{WebGlProgram, WebGlRenderingContext, WebGlShader};
+
+#[wasm_bindgen]
+pub fn draw() {
+    let document = web_sys::Window::document().unwrap();
+    let canvas = document.get_element_by_id("canvas").unwrap();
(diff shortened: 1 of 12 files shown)


examples/webgl/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+use wasm_bindgen::JsCast;
+use web_sys::{WebGlProgram, WebGlRenderingContext, WebGlShader};
+
+pub fn compile_shader(
+    context: &WebGlRenderingContext,
+    shader_type: u32,
+    source: &str,
+) -> Result<WebGlShader, String> {
+    let shader = context
+        .create_shader(shader_type as u32)
```
chinedufn: incredibly, incredibly minor - but it seems like `shader_type` is already a `u32` so need for casting?

examples/webgl/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        Ok(program)
+    } else {
+        Err(context
+            .get_program_info_log(&program)
+            .unwrap_or_else(|| "Unknown error creating program object".into()))
+    }
+}
+
+#[wasm_bindgen]
```
chinedufn: Minor - but what about moving this function to the top of the file since it's the meat of the demo?

blm768:
@chinedufn Good catch on the cast. That's what I get for copying from my other project. ;)

I think I've got the guide section in, but let me know if something doesn't look right.

PR #836, opened 2018-09-16
Link to API docs from README

chinedufn:
Just started moving a project from my own `externs` to `web-sys` and need to look at the docs to see how to call different methods.

Spent a few minutes trying to remember where the links were so wanted to add them to the README.

This PR adds those links to the `README`. A bit of extra README real estate, but feels worth it IMO.


alexcrichton:
Oops, thanks!

PR #837, opened 2018-09-16
Update weedle: use special instead of specials

afdw:
https://github.com/rustwasm/weedle/pull/15


afdw:
OK, after updating weedle it seems to build fine. Docs of web-sys do not have any differences:
image

alexcrichton:
Thanks!

PR #848, opened 2018-09-17
Tweak more `Closure` docs

### LINKED ISSUE #843: Hard to pass functions to web-sys
Tehe wasm-bindgen reference has [a chapter for passing Rust closures to JS](https://rustwasm.github.io/wasm-bindgen/reference/passing-rust-closures-to-js.html), which involves using `Closure` from wasm-bindgen.

However, functions exposed from web-sys doesn't seem to be using that. For example, `Window::set_timeout_with_callback` accepts a `&Function` which doesn't seem to have any conversion method with `Closure`, and it doesn't seem to be possible to construct a `Function` from Rust code (without writing JavaScript code inline).

It would be great if it is possible to pass Rust function or closure to web-sys functions which accept callbacks.

alexcrichton:
Show off usage of the stable `Closure::wrap` instead of `Closure::new` and
additionally add an explicit example of using it with `web_sys`.

Closes #843

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -3,6 +3,11 @@
 //! This module defines the `Closure` type which is used to pass "owned
 //! closures" from Rust to JS. Some more details can be found on the `Closure`
 //! type itself.
+<<<<<<< a37fa45100baddc36695cbf01110bcd9ab638065
+=======
+//!
+//! The `nightly` feature is required for the `Closure` type to be used.
+>>>>>>> Tweak more `Closure` docs
 
 use std::cell::UnsafeCell;
 #[cfg(feature = "nightly")]
@@ -37,6 +42,8 @@ use throw;
 ///
 /// # Example
 ///
+/// Sample usage of `Closure` to invoke the `setTimeout` API.
+///
 /// ```rust,no_run
 /// #[wasm_bindgen]
 /// extern {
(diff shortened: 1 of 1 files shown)


src/closure.rs
```diff
@@ -55,9 +57,9 @@ use throw;
 /// pub fn run() -> ClosureHandle {
 ///     // First up we use `Closure::new` to wrap up a Rust closure and create
```
chinedufn: Looks like the comment needs a tweak

src/closure.rs
```diff
@@ -69,6 +71,37 @@ use throw;
...
+///
+/// #[wasm_bindgen]
+/// pub fn run() -> ClosureHandle {
+///     let cb = Closure::wrap(Box::new(move || {
+///         web_sys::console::log_1(&"timeout elapsed!".into());
+///     }) as Box<FnMut()>);
+///
+///     let window = web_sys::window().unwrap();
+///     window.set_timeout_with_callback_and_timeout_and_arguments_0(
+///         // Note this method call, which uses `as_ref()` to get a `JsValue`
```
chinedufn: 🔥

src/closure.rs
alexcrichton: Oops, thanks!

PR #850, opened 2018-09-18
Allow returning `Result` from functions 

### LINKED ISSUE #841: Allow returning Result<T, JsValue> for throwing
Many functions in web-sys return a `Result<_, JsValue>` presumably because they can throw in certain cases. But Rust code isn't really the best place to handle those, so it would be better if we can just pass (rethrow) it back to JavaScript.

I would suggest have wasm-bindgen support function returning `Result<T, JsValue>` where `T` is the current supported return types, and handle the `Err(_)` branch as throw.

This way, we can also easily redirect exceptions from web-sys functions to the caller rather than need to `unwrap()` everywhere.

alexcrichton:
This commit adds support for exporting a function defined in Rust that returns a
`Result`, translating the `Ok` variant to the actual return value and the `Err`
variant to an exception that's thrown in JS.

The support for return types and descriptors was rejiggered a bit to be a bit
more abstract and more well suited for this purpose. We no longer distinguish
between functions with a return value and those without a return value.
Additionally a new trait, `ReturnWasmAbi`, is used for converting return values.
This trait is an internal implementation detail, however, and shouldn't surface
itself to users much (if at all).

Closes #841

### DIFF
--- a/guide/src/reference/types/result.md
+++ b/guide/src/reference/types/result.md
@@ -1,0 +1,20 @@
+# `Result<T, JsValue>`
+
+| `T` parameter | `&T` parameter | `&mut T` parameter | `T` return value | `Option<T>` parameter | `Option<T>` return value | JavaScript representation |
+|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
+| No | No | No | No | No | Yes | Same as `T`, or an exception |
+
+The `Result` type can be returned from functions exported to JS as well as
+closures in Rust. Only `Result<T, JsValue>` is supported where `T` can be
+converted to JS. Whenever `Ok(val)` is encountered it's converted to JS and
+handed off, and whenever `Err(error)` is encountered an exception is thrown in
+JS with `error`.
+
+You can use `Result` to enable handling of JS exceptions with `?` in Rust,
+naturally propagating it upwards to the wasm boundary. Furthermore you can also
+return custom types in Rust so long as they're all convertible to `JsValue`.
+
+Note that if you import a JS function with `Result` you need
+`#[wasm_bindgen(catch)]` to be annotated on the import (unlike exported
+functions, which require no extra annotation). This may not be necessary in the
+future though and it may work "as is"!.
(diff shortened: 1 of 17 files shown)


afdw:
One alternative would be always using rethrow behavior if the function return `Result`, not requiring the `rethrow` attribute, if we are sure that returning `Result` is useful only for this purpose, not for, say, returning it directly to JS.

eminence:
Can the `rethrow` attribute be used in conjunction with the `constructor` attribute?

alexcrichton:
Ok I've updated after brainstorming with @fitzgen who had the same gut reaction as @afdw and we figured out a way to implement it!

@eminence we can certainly add support for it!

guide/src/reference/types/result.md
```diff
@@ -0,0 +1,20 @@
+# `Result<T, JsValue>`
```
fitzgen: Thanks for updating the guide!

src/lib.rs
```diff
@@ -565,6 +572,26 @@ pub fn throw(s: &str) -> ! {
...
+/// function will not return and the wasm stack will be popped until the point
+/// of entry of wasm itself.
+///
+/// Note that it is very easy to leak memory with this function because this
+/// function, unlike `panic!` on other platforms, **will not run destructors**.
+/// It's recommended to return a `Result` where possible to avoid the worry of
+/// leaks.
+#[cold]
+#[inline(never)]
+pub fn rethrow(s: JsValue) -> ! {
```
fitzgen: nitpick: there is nothing about this function that requires that `s` was ever thrown before (I could be returning a result as the first throw) so this name is a bit off.

How about:

* `throw` -> `throw_str`
* `rethrow` -> `throw_val`

?

src/lib.rs
```diff
@@ -827,4 +854,14 @@ pub mod __rt {
         static FOO: AtomicUsize = ATOMIC_USIZE_INIT;
         FOO.store(0, Ordering::SeqCst);
     }
+
+    pub trait IsResult {
```
fitzgen: Where is this being used? What is this for?

alexcrichton: Oops, left over from previous implementation

src/lib.rs
alexcrichton: Sounds good to me!

PR #853, opened 2018-09-19
Add `#[derive(Clone, Debug)]` to all web-sys types

alexcrichton:
They're all cloneable and debuggable!

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -468,11 +468,18 @@ impl<'src> FirstPassRecord<'src> {
             name,
             mdn_doc(name, None),
         ));
+        let derive = syn::Attribute {
+            pound_token: Default::default(),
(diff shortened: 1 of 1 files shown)


chinedufn:
Something that comes to mind here:

```
let mut elem` = document.get_element_by_id("id");
let mut elem2 = elem.clone();
```

In most Rust code mutating `elem1` wouldn't mutate `elem2`, but it seems like in our `web-sys` case mutating one would mutate the other and that feels unexpected.. I think?

fitzgen:
@chinedufn it's the same as something like `Rc<RefCell<...>>`: shared ownership with interior mutability. If you think about it, JS things are always shared with JS land, and JS is not bound by the mutability xor aliasing that Rust usually follows, so everything in JS is implicitly mutable. Given that, I think it makes sense.

chinedufn:
Makes sense!

PR #854, opened 2018-09-19
Add the last four bindings for Object

brisad:
#275


fitzgen:
CI failure is unrelated, comes from two merges that don't literally conflict, but still logically conflict.

PR #860, opened 2018-09-20
Reorganize and rewrite examples

alexcrichton:
This commit is a large-ish scale reorganization of our examples. The
main goal here is to have a dedicated section of the guide for example,
and all examples will be listed there. Each example's `README` is now
just boilerplate pointing at the guide along with a blurb about how to
run it.

Some examples like `math` and `smorgasboard` have been deleted as they
didn't really serve much purpose, and others like `closures` have been
rewritten with `web-sys` instead of hand-bound bindings.

Overall it's hoped that this puts us in a good and consistent state for
our examples, with all of them being described in the guide, excerpts
are in the guide, and they're all relatively idiomatically using
`web-sys`.


examples/wasm-in-wasm/build.sh
```diff
@@ -8,5 +8,5 @@ cargo +nightly build --target wasm32-unknown-unknown
 cargo +nightly run --manifest-path ../../crates/cli/Cargo.toml \
   --bin wasm-bindgen -- \
   ../../target/wasm32-unknown-unknown/debug/wasm_in_wasm.wasm --out-dir .
-npm install
+#npm install
```
afdw: Looks like this was accidentally left commented.

afdw:
Another thing that could be done is unifying package names: some examples are named like `wasm-bindgen-…-demo`, some are not.

examples/wasm-in-wasm/build.sh
alexcrichton: Oops, thanks!

alexcrichton:
Oh sure thing, makes sense!

examples/canvas/index.html
```diff
@@ -4,6 +4,5 @@
   </head>
   <body>
     <canvas id="canvas" height="150" width="150" />
-    <script src='./index.js'></script>
```
fitzgen: I assume that this is still needed to load the JS no?

examples/char/index.html
```diff
@@ -49,8 +49,7 @@
     <body>
         <button id="add-counter" type="button">add counter</button>
         <div id="container">
-            
+
         </div>
-        <script type="text/javascript" src="index.js"></script>
```
fitzgen: Are we using the fancy webpack thing where it automatically includes `index.js`?

PR #862, opened 2018-09-21
Enable union types without interfaces in WebIDL 

### LINKED ISSUE #817: web-sys: Unable to add body to a Request
I'm trying to create a `Request` with a body but I can't figure out how (or it isn't possible yet). Neither `Request` nor `RequestInit` offer a way to add a body. Ultimately I'd like to do something like [this](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Uploading_JSON_data).

alexcrichton:
Bind them all as `JsValue` as that's the "least common ancestor" we can
work with. Fixes up one location in WebIDL where `Option<JsValue>`
arose as we haven't implemented that.

Closes #817

### DIFF
--- a/crates/webidl/src/idl_type.rs
+++ b/crates/webidl/src/idl_type.rs
@@ -521,19 +521,54 @@ impl<'a> IdlType<'a> {
             },
             IdlType::Enum(name) => Some(ident_ty(rust_ident(camel_case_ident(name).as_str()))),
 
-            IdlType::Nullable(idl_type) => Some(option_ty(idl_type.to_syn_type(pos)?)),
+            IdlType::Nullable(idl_type) => {
+                let inner = idl_type.to_syn_type(pos)?;
+
+                // TODO: this is a bit of a hack, but `Option<JsValue>` isn't
+                // supported right now. As a result if we see `JsValue` for our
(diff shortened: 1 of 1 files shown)


afdw:
I have previously implemented this with a simple check for unions in code that handles nullable types while converting them to syn types: if it is a nullable type with a union type inside it, then return `IdlType::Any.to_syn_type(pos)`. This can save us from modifying WebIDL files (with a bit more complicated code logic though, but hopefully it will be only temporary).

alexcrichton:
I'd be fine either way, but I feel like that sort of check is quite complicated and would lean towards this solution since it only comes up in one location. Do you have the code though for what that looks like?

afdw:
See here: https://github.com/rustwasm/wasm-bindgen/pull/785/commits/095f86fa51ef42b6c167b6f6b925b257dcee7922
I am OK with both ways too, I only meant that todo in code has less chance to be forgotten then one in a WebIDL file.

alexcrichton:
Ah ok right! I've updated with a slight variant of that to just check for any `JsValue` coming out instead of specifically unions, and that should hopefully avoid some duplication

PR #866, opened 2018-09-21
More conservatively patch closure descriptors

### LINKED ISSUE #864: wasm-bindgen fails to process wasm generated with LTO
When I enable LTO on my wasm project, wasm-bindgen aborts with the following:
```
wasm-bindgen rust/out/calories_calc.wasm \
		--no-modules --out-dir out
  left: `Some(18)`,
 right: `Some(0)`', /Users/upsuper/.cargo/git/checkouts/wasm-bindgen-340397a49d85992c/f871cf6/crates/wasm-interpreter/src/lib.rs:349:25
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: std::panicking::begin_panic_fmt
   7: wasm_bindgen_wasm_interpreter::Interpreter::call
   8: wasm_bindgen_wasm_interpreter::Interpreter::interpret_closure_descriptor
   9: wasm_bindgen_cli_support::js::closures::rewrite
  10: wasm_bindgen_cli_support::js::Context::finalize
  11: wasm_bindgen_cli_support::Bindgen::_generate
  12: wasm_bindgen::main
  13: std::rt::lang_start::{{closure}}
  14: std::panicking::try::do_call
  15: __rust_maybe_catch_panic
  16: std::rt::lang_start_internal
  17: main
```

I can probably try reduce the code if that would help.

alexcrichton:
Previously `wasm-bindgen` would take its `breaks_if_inlined` shims and
attempt to remove them entirely, replacing calls to `breaks_if_inlined`
to the imported closure factories. This worked great in that it would
remove the `breaks_if_inlined` funtion entirely, removing the "cost" of
the `#[inline(never)]`.

Unfortunately as #864 discovered this is "too clever by half". LLVM's
aggressive optimizations won't inline `breaks_if_inlined`, but it may
still change the ABI! We can't replace calls to `breaks_if_inlined` if
the signature changes, because the function its calling has a fixed signature.

This commit cops out a bit and instead of replacing calls to
`breaks_if_inlined` to the imported closure factories, we instead
rewrite calls to `__wbindgen_describe_closure` to the closure factories.
This means that the `breaks_if_inlined` shims do not get removed. It
also means that the closure factory shims have a third and final
argument (what would be the function pointer of the descriptor function)
which is dead and unused.

This should be a functional solution for now and let us iterate on a
true fix later on (if needed). For now the cost of this
`#[inline(never)]` and the extra unused argument should be quite small.

Closes #864

### DIFF
--- a/crates/cli-support/src/js/closures.rs
+++ b/crates/cli-support/src/js/closures.rs
@@ -47,6 +47,7 @@ pub fn rewrite(input: &mut Context) -> Result<(), Error> {
 
     info.delete_function_table_entries(input);
     info.inject_imports(input)?;
+    info.rewrite_calls(input);
     Ok(())
 }
 
@@ -63,7 +64,13 @@ struct ClosureDescriptors {
     ///
     /// This map is later used to replace all calls to the keys of this map with
     /// calls to the value of the map.
-    code_idx_to_descriptor: BTreeMap<u32, (u32, Descriptor)>,
+    code_idx_to_descriptor: BTreeMap<u32, DescribeInstruction>,
+}
+
+struct DescribeInstruction {
+    new_idx: u32,
+    instr_idx: usize,
+    descriptor: Descriptor,
 }
 
 impl ClosureDescriptors {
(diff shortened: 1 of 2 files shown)


crates/cli-support/src/js/closures.rs
```diff
@@ -261,10 +275,37 @@ impl ClosureDescriptors {
         }
         Ok(())
     }
+
+    /// The final step, rewriting calls to `__wbindgen_describe_closure` to the
+    /// imported functions
+    ///
+    /// Here we take the final step, rewriting calls to
+    /// `__wbindgen_describe_closure` to the imported functions, ensuring
+    /// there's no lingering references to `__wbindgen_describe_closure`.
```
fitzgen: This comment's paragraphs seem oddly redundant.

crates/cli-support/src/js/closures.rs
```diff
@@ -261,10 +275,37 @@ impl ClosureDescriptors {
...
+    fn rewrite_calls(&self, input: &mut Context) {
+        // FIXME: Ok so this is a bit sketchy in that it introduces overhead.
+        // What we're doing is taking a our #[inline(never)] shim and *not*
+        // removing it, only switching the one function that it calls internally.
+        //
+        // This isn't great because now we have this non-inlined function which
+        // would certainly benefit from getting inlined. It's a tiny function
+        // though and surrounded by allocation so it's probably not a huge
+        // problem in the long run.
```
fitzgen: If we are going to make excuses for ourselves here, we might as well mention that `wasm-opt` will likely inline the function for us anyways.

PR #871, opened 2018-09-22
Support `#[wasm_bindgen(setter, js_name = ...)]` 

### LINKED ISSUE #584: Cannot properly create setters when js_name is provided
While attempting to create the binding for `regexp.lastIndex` with the following code,
```rust
/// The lastIndex is a read/write integer property of regular
/// expression instances that specifies the index at which to
/// start the next match.
///
/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex
#[wasm_bindgen(method, getter, structural, js_name = lastIndex)]
pub fn last_index(this: &RegExp) -> u32;
#[wasm_bindgen(method, setter, structural, js_name = lastIndex)]
pub fn set_last_index(this: &RegExp, index: u32);
```
I got the following error when compiling:
> 
> error: custom attribute panicked
>     --> crates/js-sys/src/lib.rs:2054:1
>      |
> 2054 | #[wasm_bindgen]
>      | ^^^^^^^^^^^^^^^
>      |
>      = help: message: setters must start with `set_`
> 

Since the method is called `set_last_index` this error appears to be invalid. And according to https://github.com/rustwasm/wasm-bindgen/pull/580#discussion_r205978745 `js_name = lastIndex` is correct.

alexcrichton:
Previously we'd require the explicit `js_name` to *also* start with
`set_`, but when explicitly specified it shouldn't be mangled at all!

Closes #584 

Note that along the way this also improve the error message we get from failures to infer names.

### DIFF
--- a/crates/macro/ui-tests/invalid-setter.stderr
+++ b/crates/macro/ui-tests/invalid-setter.stderr
@@ -4,11 +4,5 @@ error: setters must start with `set_`, found: a
 10 |     fn a(this: &A, b: i32);
    |        ^
 
-error: setters must start with `set_`, found: x
-  --> $DIR/invalid-setter.rs:15:46
-   |
-15 |     #[wasm_bindgen(setter, method, js_name = x)]
-   |                                              ^
-
-error: aborting due to 2 previous errors
+error: aborting due to previous error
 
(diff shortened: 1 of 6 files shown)


crates/macro/ui-tests/invalid-setter.stderr
```diff
@@ -0,0 +1,8 @@
+error: setters must start with `set_`, found: a
+  --> $DIR/invalid-setter.rs:10:8
+   |
+10 |     fn a(this: &A, b: i32);
+   |        ^
```
fitzgen: nice

PR #877, opened 2018-09-24
Add some blurbs about optimizing for size

### LINKED ISSUE #826: Discuss size optimization
https://rustwasm.github.io/wasm-bindgen/whirlwind-tour/basic-usage.html says:

> After this you'll have a wasm file at target/wasm32-unknown-unknown/debug/js_hello_world.wasm. Don't be alarmed at the size, this is an unoptimized program!

Even compiled in release mode, however, this still seems true. Please consider explaining this in a bit more detail to show what the real sizes look like after processing.

alexcrichton:
The main thing we mention is to explicitly not measure the output of the
compiler, but only the ouput of `wasm-bindgen` itself.

Closes #826


joshtriplett:
This looks great, thanks! That answers all the questions I'd had when I first encountered the original verbiage.

PR #878, opened 2018-09-24
Implement support for `Uint8ClampedArray`

### LINKED ISSUE #421: Support creating `Uint8ClampedArray` views into wasm's memory
This [first came up](https://github.com/rustwasm/wasm-bindgen/pull/419#issuecomment-403296031) with #419 but some APIs on the web specifically take `Uint8ClampedArray`, so we should support taking a view into wasm's memory with this type of array! Today there's no easy way to create a `Uint8ClampedArray` which doesn't copy data from wasm's memory into a temporary buffer.

A strawman is something like:

```rust
#[wasm_bindgen]
extern {
    #[wasm_bindgen(clamped)]
    fn foo(a: &[u8]);
}
```

where that sort of vaguely means that all `&[u8]` arguments are actually clamped arrays instead of `Uint8Array`, but it's not great as it's not too configurable!

alexcrichton:
This commit implements support for binding APIs that take
`Uint8ClampedArray` in JS. This is pretty rare but comes up in a
`web-sys` binding or two, and we're now able to bind these APIs instead
of having to omit the bindings.

The `Uint8ClampedArray` type is bound by using the `Clamped` marker
struct in Rust. For example this is declaring a JS API that takes
`Uint8ClampedArray`:

    use wasm_bindgen::Clamped;

    #[wasm_bindgen]
    extern {
        fn takes_clamped(a: Clamped<&[u8]>);
    }

The `Clamped` type currently only works when wrapping the `&[u8]`, `&mut
[u8]`, and `Vec<u8>` types. Everything else will produce an error at
`wasm-bindgen` time.

Closes #421

### DIFF
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -35,6 +35,7 @@ tys! {
     CHAR
     OPTIONAL
     UNIT
+    CLAMPED
 }
 
 #[derive(Debug)]
@@ -63,6 +64,7 @@ pub enum Descriptor {
     Char,
     Option(Box<Descriptor>),
     Unit,
+    Clamped(Box<Descriptor>),
 }
 
 #[derive(Debug)]
@@ -81,6 +83,7 @@ pub struct Closure {
 pub enum VectorKind {
     I8,
     U8,
+    ClampedU8,
     I16,
     U16,
     I32,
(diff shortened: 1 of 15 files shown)


alexcrichton:
Ah and this commit also performs a drive-by fix to bind all slice types as `&mut [T]` instead of `&mut [u8]` and `&[T]` for everything else. In all cases for web-sys we have to pessimistically assume that the input is being modified as we don't actually know if it will be modified or not.

Issue #873, opened 2018-09-23
Don't require allocation for returning string from web-sys


upsuper:
Currently when functions return `String` for string value from web-sys, this makes it impossible to remove allocation support from wasm when using web-sys.

I think it is possible to avoid needing to allocate. One possible approach is to define a trait like
```rust
trait StringAlloc {
    unsafe fn alloc(&mut self, len: usize) -> Option<*const u8>;
}
```
and have web-sys functions which want to return string accept a `&mut dyn StringAlloc` instead.

We can then impl that trait for `String` and have some wrapper function so people don't mind doing allocation can continue using the easy approach, while people who don't want allocation can impl that trait for their own string types like `ArrayString`.

alexcrichton:
This seems like an interesting idea! Can you elaborate one some of the use cases you're thinking of as well? I'm not really sure how this would fit into the existing APIs as-is, but knowing more about the end goal may help!

upsuper:
As I mentioned in the issue description, the end goal I'm thinking is to be able to have allocation-free wasm with web-sys.

For small applications, allocator is a significant part of codesize, even it's wee_alloc. (Actually in my tiny application, wee_alloc takes more codesize then the default.) So I guess it might be useful if it's possible to use web-sys without requiring allocator.

Also heap allocation is slower than pre-allocated slice in stack I suppose?

Majority of methods returning `String` are attributes in IDL, so we can probably do something like:
```rust
// binding to JS code, return whether allocation succeeded.
fn get_id(&self, &mut dyn StringAlloc) -> bool;
fn id(&self) -> String {
    let mut result = String::new();
    self.get_id(&mut result).unwrap();
    result
}
```

Oh, and it reminds me that, with doing so we can probably keep all heap allocation in the Rust side, and may no longer need to expose `__wbindgen_malloc` to JavaScript.

alexcrichton:
This seems plausible to do but it does seem slightly esoteric in motivation. I've found that in practice `dlmalloc` is a pretty small portion of the binary and `wee_alloc` is *absolutely tiny* so if `wee_alloc` dominates the binary size it's not clear if that's the best use case for wasm. Not to say this shouldn't ever get done, it just doesn't seem too high priority to me personally at least.

upsuper:
[...]

PR #879, opened 2018-09-24
Improve codegen for `Closure<T>`

### LINKED ISSUE #874: Support passing non-closure functions to js without allocation?
Currently we can use `Closure::wrap` to create a `Closure` and cast it to `Function` for passing to Web API. However it isn't always necessary to have a closure, and in those cases an additional allocation may be unfortunate.

Can we / do we have some approach to pass function pointer like `fn(Event)` to Web API without using `Box`?

alexcrichton:
This commit improves the codegen for `Closure<T>`, primarily for ZST
where the closure doesn't actually capture anything. Previously
`wasm-bindgen` would unconditionally allocate an `Rc` for a fat pointer,
meaning that it would always hit the allocator even when the `Box<T>`
didn't actually contain an allocation. Now the reference count for the
closure is stored on the JS object rather than in Rust.

Some more advanced tests were added along the way to ensure that
functionality didn't regress, and otherwise the calling convention for
`Closure` changed a good deal but should still be the same user-facing.
The primary change was that the reference count reaching zero may cause
JS to need to run the destructor. It simply returns this information in
`Drop for Closure` and otherwise when calling it now also retains a
function pointer that runs the destructor.

Closes #874

### DIFF
--- a/crates/cli-support/src/js/closures.rs
+++ b/crates/cli-support/src/js/closures.rs
@@ -226,26 +226,32 @@ impl ClosureDescriptors {
 
             let (js, _ts, _js_doc) = {
                 let mut builder = Js2Rust::new("", input);
+                builder.prelude("this.cnt++;");
                 if closure.mutable {
                     builder
                         .prelude("let a = this.a;\n")
                         .prelude("this.a = 0;\n")
                         .rust_argument("a")
+                        .rust_argument("b")
                         .finally("this.a = a;\n");
                 } else {
-                    builder.rust_argument("this.a");
+                    builder.rust_argument("this.a")
+                        .rust_argument("b");
                 }
+                builder.finally("if (this.cnt-- == 1) d(this.a, b);");
                 builder
                     .process(&closure.function)?
-                    .finish("function", "this.f")
+                    .finish("function", "f")
             };
             input.expose_add_heap_object();
             input.function_table_needed = true;
             let body = format!(
-                "function(ptr, f, _ignored) {{
-                    let cb = {};
-                    cb.f = wasm.__wbg_function_table.get(f);
-                    cb.a = ptr;
+                "function(a, b, fi, di, _ignored) {{
+                    const f = wasm.__wbg_function_table.get(fi);
+                    const d = wasm.__wbg_function_table.get(di);
+                    const cb = {};
+                    cb.a = a;
+                    cb.cnt = 1;
                     let real = cb.bind(cb);
                     real.original = cb;
                     return addHeapObject(real);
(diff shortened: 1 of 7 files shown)


alexcrichton:
cc @upsuper, mind testing this out to see if it works for your use case?

upsuper:
Sorry for the delay. As I mentioned in [#873 (comment)](https://github.com/rustwasm/wasm-bindgen/issues/873#issuecomment-423855019) my main hope is to be able to remove allocation completely, which is probably not a goal worth pursuing with lots of effort for wasm...

And I actually tested this patch, but it seems to somehow break #864 again. I see the following when I try to run `wasm-bindgen`:
```
  left: `3`,
 right: `1`', crates/wasm-interpreter/src/lib.rs:391:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: std::panicking::begin_panic_fmt
   7: wasm_bindgen_wasm_interpreter::Interpreter::call
   8: wasm_bindgen_wasm_interpreter::Interpreter::interpret_closure_descriptor
   9: wasm_bindgen_cli_support::js::closures::rewrite
  10: wasm_bindgen_cli_support::js::Context::finalize
  11: wasm_bindgen_cli_support::Bindgen::_generate
  12: wasm_bindgen::main
  13: std::rt::lang_start::{{closure}}
  14: std::panicking::try::do_call
  15: __rust_maybe_catch_panic
  16: std::rt::lang_start_internal
  17: main
```

alexcrichton:
Oops the panic was a mistake and should be fixed now. This PR does remove the allocation if your closure doesn't actually capture anything (like you're calling a function pointer or only capture ZST variables)

upsuper:
Yeah, I can confirm that the allocation goes away if I pass in a closure with no capture or only ZST variable capture. Thanks!

alexcrichton:
Awesome, thanks!

r? @fitzgen

crates/cli-support/src/js/closures.rs
```diff
@@ -226,26 +226,32 @@ impl ClosureDescriptors {
...
+                    .finish("function", "f")
             };
             input.expose_add_heap_object();
             input.function_table_needed = true;
             let body = format!(
-                "function(ptr, f, _ignored) {{
-                    let cb = {};
-                    cb.f = wasm.__wbg_function_table.get(f);
-                    cb.a = ptr;
+                "function(a, b, fi, di, _ignored) {{
```
fitzgen: What is `b` here?

alexcrichton: The a/b pointers here are the two parts of a trait object, with fi being the wrapper that we invoke which does wasm-bindgen translation of arguments.

PR #887, opened 2018-09-25
Iterating over js values

fitzgen:


alexcrichton:
Looks good to me! Do you have an idea for how we will handle the js-sys breaking change? Just publish a new version?

PR #902, opened 2018-09-27
Add documentation about supported targets

### LINKED ISSUE #892: Clarity on which targets are supported
I'm having a hard time figuring out whether or not this is intended to work with wasm32-unknown-emscripten. There seem to be lots of checks of the form:

```
```

But I'm not sure if that is meant to preclude generation of code known not to work or not needed on Emscripten, or if that is meant to make this crate a no-op when compiled for wasm32-unknown-emscripten. And I can't find a clear answer in any of the docs I've checked. Apologies if it is documented somewhere and I haven't found it.

FWIW, I'm aware of Rust's preference for wasm32-unknown-unknown, and if I didn't need to link against C libraries that will never be ported to Rust, I'd prefer that too. I'm just currently jumping through lots of hoops to make this code work in wasm, and if wasm-bindgen doesn't support Emscripten at all, it'd be helpful to see that in some prominant place, instead of learning that after I've jumped through bunches of hoops only to get a .wasm file that won't work. :)

And if it *does* work under Emscripten, some general pointers on how to get a working .wasm file would also be helpful. I'm happy to try distilling those points into user-friendly docs if I have a bit of guidance. The issue I'm hitting now is that `crate-type=["cdylib"]` isn't supported under Emscripten and doesn't generate a .wasm file. So now I'm trying to create a bin crate with an empty `main` and a more JS-friendly API which I export via `#[wasm_bindgen]`, but that's going to take a bit of work, and I'd rather not do that if what I'll ultimately get is an incompatible .wasm file. :)

Thanks!

alexcrichton:
* Main target is wasm32-unknown-unknown
* All other targets work ok, but imports panic
* Emscripten explicitly not supported at this time

Closes #892

### DIFF
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -37,6 +37,7 @@
   - [Working with Duck-Typed Interfaces](./reference/working-with-duck-typed-interfaces.md)
   - [Command Line Interface](./reference/cli.md)
   - [Optimizing for Size](./reference/optimize-size.md)
+  - [Supported Rust Targets](./reference/rust-targets.md)
   - [Supported Types](./reference/types.md)
     - [Imported JavaScript Types](./reference/types/imported-js-types.md)
     - [Exported Rust Types](./reference/types/exported-rust-types.md)
--- a/guide/src/reference/rust-targets.md
+++ b/guide/src/reference/rust-targets.md
@@ -1,0 +1,72 @@
+# Supported Rust Targets
+
+> **Note**: This section is about Rust target triples, not targets like node/web
+> workers/browsers. More information on that coming soon!
+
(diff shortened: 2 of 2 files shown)


guide/src/reference/targets.md
```diff
@@ -0,0 +1,69 @@
+# Supported Targets
```
fitzgen: I think JS developers are going to be thinking "node or Web or workers or ..." here. It might be worth clarifying somehow. Maybe "Supported Target Triples"? That's not great... do you have any other idea for disambiguating?

alexcrichton: Ah excellent point, how about "Rust Targets" along with a note at the top saying we're not talking about web workers and such? (although that'll also make good documentation eventually)

fitzgen: Sounds great!

Issue #897, opened 2018-09-27
Webaudio example gives error in Safari on Mac (works in Firefox)


kud1ing:
The web development console says:
```
[Error] Unhandled Promise Rejection: ReferenceError: Can't find variable: AudioContext
	(anonymous function)
	promiseReactionJob
```

This is Safari 12 on macOS Mojave.

chinedufn:
Web audio is [supported with a prefix in safari](https://caniuse.com/#feat=audio-api), so the example [would need to fall back to `webkitAudioContext` if there was no `AudioContext`](https://github.com/rustwasm/wasm-bindgen/blob/826619870f4cb043663530c8c6a60e486aa5b2b7/examples/webaudio/src/lib.rs#L6)

---

From a quick search it doesn't seem like `web-sys` supports any prefixed APIs (I could very well be missing something though!). 

If that's the case I'm also not sure if that's an intentional decision or not.

alexcrichton:
Thanks for the report and thanks for the info @chinedufn! We [discussed this during today's meeting](https://paper.dropbox.com/doc/RustWebAssembly-WG--AN0Kj_TSmeR5csz9JFryWJQsAg-dcXi5F7Bqlvg10zd7dKve#:uid=466221811040185070621428&h2=2018/09/27) and the conclusion was that we'd like to add support to `web-sys` to whitelist APIs as "these are alternative class names". That way we can get this working by default even in Safari where it's renamed to `webkitAudioContext`.

alexcrichton:
https://github.com/rustwasm/wasm-bindgen/pull/906 fixes using `webkitAudioContext` on Safari, but unfortunately the example still doesn't work in Safari. The `AudioScheduledSourceNode` type doesn't exist in Safari (and edge presumably) but rather the methods are just listed in the parent classes. I've written up a more detailed issue about that at https://github.com/rustwasm/wasm-bindgen/issues/908

alexcrichton:
The combination of https://github.com/rustwasm/wasm-bindgen/pull/909 and https://github.com/rustwasm/wasm-bindgen/pull/906 should fix this issue

kud1ing:
Is this deployed? The demo now gives:

```
ReferenceError: Can't find variable: AudioScheduledSourceNode
Eval Code
eval
./webaudio.js — 0.index.js:11
__webpack_require__ — index.js:144
__webpack_require__
promiseReactionJob
```

alexcrichton:
[...]

PR #906, opened 2018-09-28
Implement a `polyfill` attribute for imports

alexcrichton:
Allow using imported APIs under alternative names, such as prefixed
names, for web APIs when the exact API differs across browsers.

Note that this is intended to assist with https://github.com/rustwasm/wasm-bindgen/issues/897 but it does not fully fix it for reasons I'll mention over there shortly.


alexcrichton:
@fitzgen makes sense to me! I've submitted that as https://github.com/rustwasm/wasm-bindgen/pull/914

Issue #908, opened 2018-09-28
Gecko's WebIDL can be wrong for other browsers


alexcrichton:
We originally lifted our WebIDL definitions from Gecko, which means we're highly likely to work in Firefox which uses that exact WebIDL (yay!). Over time though we're discovering that Gecko's WebIDL isn't what's used by other browsers.

Examples discovered so far:

* [`AudioScheduledStartNode`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.AudioScheduledSourceNode.html) exists in our WebIDL, but in webkit the methods [are defined inherently](https://github.com/WebKit/webkit/blob/d779d13d239e5f893fd23864d5c9b48d5b6e925c/Source/WebCore/Modules/webaudio/OscillatorNode.idl#L51-L52) in classes which inherit from `AudioScheduledStartNode` (like `OscillatorNode`)
* [`BaseAudioContext`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.BaseAudioContext.html) doesn't exist in Safari, looking like [it's also inlined on `AudioContext`](https://github.com/WebKit/webkit/blob/d779d13d239e5f893fd23864d5c9b48d5b6e925c/Source/WebCore/Modules/webaudio/AudioContext.idl#L79-L90).

These definition issues aren't generally a problem for the web platform but they're a problem for us in that we're plucking our raw function pointers for maximal performance, and those raw function pointers are located in different places.

Ok with that in mind, how do we want to solve this? I don't think we want to include WebIDL from all browsers and take the union, but rather what I think we can do is to basically just massage our WebIDL to become more browser-agnostic over time. Put another way, we should do this:

* For `AudioScheduledStartNode`, switch it to a mixin.
* All classes which inherit from `AudioScheduledStartNode` now have `Foo includes AudioScheduledStartNode` and inherit from `AudioScheduledStartNode`'s original parent

Transformations like that mean that we're modeling the web platform slightly differently, but it should effectively have the same result. The `wasm-bindgen` generated bindings [already look up the inheritance chain](https://github.com/rustwasm/wasm-bindgen/blob/11bcaf42d5300adc99fd3d10ac726e568c304e88/crates/cli-support/src/js/mod.rs#L1609-L1616) for the right function pointer, which means that we'll still use all the native methods on `BaseAudioContext` in Firefox and we'll otherwise use the methods on `AudioContext` in Safari. 

[...]

PR #909, opened 2018-09-28
Fix our WebIDL for Safari

alexcrichton:
This commit employs the strategy described in #908 to apply a
non-breaking change to fix WebIDL to be compatible with all browsers,
including Safari.

The problem here is that `BaseAudioContext` and `AudioScheduledSourceNode`
are not types in Safari, but they are types in Firefox/Chrome. The fix
here was to move the contents of these two interfaces into mixins, and
then include the mixins in all classes which inherit from these two
classes. That should have the same effect as defining the methods
inherently on the original interface.

Additionally a special `[RustDeprecated]` attribute to WebIDL was added
to signify interfaces this has happened to. Currently it's directly
tailored towards this case of "this intermediate class doesn't exist in
all browsers", but we may want to refine and extend the deprecation
message over time.

Although it's possible we could do this as a breaking change to
`web-sys` I'm hoping that we can do this as a non-breaking change for
now and then eventually on the next breaking release batch all these
changes together, deleting the intermediate classes. This is also
hopefully a good trial run for how stable web-sys can be when it's
actually stable!

cc #897
cc #908

### DIFF
--- a/crates/web-sys/webidls/enabled/AudioBufferSourceNode.webidl
+++ b/crates/web-sys/webidls/enabled/AudioBufferSourceNode.webidl
@@ -32,7 +32,12 @@ interface AudioBufferSourceNode : AudioScheduledSourceNode {
     attribute double loopStart;
     attribute double loopEnd;
 
+    attribute EventHandler onended;
+
     [Throws]
     void start(optional double when = 0, optional double grainOffset = 0,
                optional double grainDuration);
+
+    [Throws]
+    void stop (optional double when = 0);
 };
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -488,18 +488,15 @@ impl<'src> FirstPassRecord<'src> {
         data: &InterfaceData<'src>,
     ) {
         let mut doc_comment = Some(format!("The `{}` object\n\n{}", name, mdn_doc(name, None),));
-        let derive = syn::Attribute {
-            pound_token: Default::default(),
-            style: syn::AttrStyle::Outer,
-            bracket_token: Default::default(),
-            path: Ident::new("derive", Span::call_site()).into(),
-            tts: quote!((Debug, Clone)),
-        };
+
+        let mut attrs = Vec::new();
+        attrs.push(parse_quote!( #[derive(Debug, Clone)] ));
+        self.add_deprecated(data, &mut attrs);
         let mut import_type = backend::ast::ImportType {
             vis: public(),
             rust_name: rust_ident(camel_case_ident(name).as_str()),
             js_name: name.to_string(),
-            attrs: vec![derive],
+            attrs,
             doc_comment: None,
             instanceof_shim: format!("__widl_instanceof_{}", name),
             extends: Vec::new(),
@@ -539,6 +536,7 @@ impl<'src> FirstPassRecord<'src> {
             self.member_attribute(
                 program,
                 name,
+                data,
                 member.modifier,
                 member.readonly.is_some(),
                 &member.type_,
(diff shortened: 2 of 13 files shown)


crates/web-sys/webidls/enabled/AudioBufferSourceNode.webidl
```diff
@@ -32,7 +32,12 @@ interface AudioBufferSourceNode : AudioScheduledSourceNode {
...
     attribute double loopEnd;
 
+                    attribute EventHandler onended;
+
     [Throws]
     void start(optional double when = 0, optional double grainOffset = 0,
                optional double grainDuration);
+
+    [Throws]
+    void stop (optional double when = 0);
```
alexcrichton: Note that this class didn't switch to `AudioBufferSourceNode includes rustAudioScheduledSourceNode;` because the mixin's `start` method conflicts with the `start` method in this class, so it was just easier to inline the few properties here.

crates/web-sys/webidls/enabled/AudioBufferSourceNode.webidl
```diff
@@ -32,7 +32,12 @@ interface AudioBufferSourceNode : AudioScheduledSourceNode {
     attribute double loopStart;
     attribute double loopEnd;
 
+                    attribute EventHandler onended;
```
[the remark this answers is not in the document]
alexcrichton: Heh [it is indeed wrong](https://github.com/rustwasm/wasm-bindgen/blob/11bcaf42d5300adc99fd3d10ac726e568c304e88/crates/web-sys/webidls/enabled/AudioScheduledSourceNode.webidl#L14)!
 
(a copy/paste error)

crates/webidl/src/lib.rs
```diff
@@ -668,10 +669,25 @@ impl<'src> FirstPassRecord<'src> {
...
+        if !data.deprecated {
+            return
+        }
+        let note = format!("This API is deprecated as it was found to not \
+                            work in all browsers, typically because the bound \
+                            class doesn't exist in all browsers. You can \
+                            probably stop using this class and instead \
+                            use the class that inherits from this, which \
+                            should include all the same methods and work \
+                            in all browsers in theory!");
```
fitzgen: It would be nice if we could tailor the deprecation message for the particular interfaces. Something like:

```webidl
// BaseAudioContext.webidl

[Deprecated="reason why this is deprecated detailed here..."]
interface BaseAudioContext {
  // ...
};
```

examples/webaudio/Cargo.toml
```diff
@@ -16,8 +16,6 @@ features = [
   'AudioDestinationNode',
   'AudioNode',
   'AudioParam',
-  'AudioScheduledSourceNode',
-  'BaseAudioContext',
```
fitzgen: Isn't removing features a breaking change?

examples/webaudio/src/lib.rs
```diff
@@ -45,22 +43,14 @@ impl Drop for FmOsc {
 #[wasm_bindgen]
 impl FmOsc {
     #[wasm_bindgen(constructor)]
-    pub fn new() -> FmOsc {
-        let ctx = web_sys::AudioContext::new().unwrap();
-        let primary;
-        let fm_osc;
-        let gain;
-        let fm_gain;
+    pub fn new() -> Result<FmOsc, JsValue> {
```
fitzgen: Nice.

examples/webaudio/Cargo.toml
alexcrichton: Oh these aren't removed from `web-sys` itself, just here because the features are no longer needed

crates/webidl/src/lib.rs
alexcrichton: Sure yeah, I can upate this

examples/webaudio/Cargo.toml
fitzgen: D'oh I totally misread what file this was

PR #910, opened 2018-09-29
Fix polyfill of `TextEncoder` and `TextDecoder`

### LINKED ISSUE #895: Full MS Edge Support
After building a project with wasm-bindgen and getting it working as expected in Chrome/Safari/FF I decided to give Edge a shot since I expected things to work there as well since it supports WebAssembly in the latest version. However, when I loaded up my code nothing seemed to happen. I don't get any errors in the console, but nothing seems to work. After debugging and paring things down a bit the issue _might_ be related to using serde, but I'm not sure that's entirely what is responsible. 

I'm seeming to have the same issue running the `fetch` from the examples directory too - https://rustwasm.github.io/wasm-bindgen/exbuild/fetch. Loading that URL seems to work fine except in Edge where, similar to what I'm seeing, nothing happens, no errors, no console output. Plenty of other projects in the examples directory seem to work fine, so there appears to be something partial that doesn't work in Edge.

I did notice that when using serde the JS shim that wasm-bindgen creates has the following:

```
const TextDecoder = typeof self === 'object' && self.TextDecoder
    ? self.TextDecoder
    : require('util').TextDecoder;
``` 

And I know that `TextDecoder` [isn't available within Edge yet](https://caniuse.com/#feat=textencoder). However I tried just jamming in a quick polyfill for that but it didn't seem to change anything. Is there any docs/known issues around supporting Edge?

alexcrichton:
This commit does a few things, including:

* Fixing the generated JS of `wasm-bindgen` to allow polyfills to work.
  (a minor tweak of the generated JS)

* All examples are updated to include a Webpack-specific polyfill for
  these two types to get examples working in Edge.

* A new page has been added to the guide about supported browsers. This
  mentions known caveats like IE 11 requiring `wasm2js` as well as
  documenting some `TextEncoder` and `TextDecoder` workarounds for Edge.

Closes #895


guide/src/reference/browser-support.md
```diff
@@ -0,0 +1,61 @@
...
+## Caveats
+
+* **IE 11** - `wasm-bindgen` by default requires support for
+  `WebAssembly`, but no version of IE currently supports `WebAssembly`. You can
+  support IE by [compiling wasm files to JS using `wasm2js`][w2js] (you can [see
+  an example of doing this too](../examples/wasm2js.html)). Note
+  that at this time no bundler will do this by default, but we'd love to
+  document plugins which do this if you are aware of one!
+
+* **Edge** - the `TextEncoder` and `TextDecoder` types are not currently
```
xtuc: `types` -> `API`?

Also note that these aren't ECMAScript API but web plateform API, so more or less plateform specifics.

xtuc:
This will fix the issue, but it seems that now `wasm-bindgen` strongly requires TextEncoder/Decoder and a way to provide it.

What about inlining the `text-encoding` in the generated code? So the issue will stay on our side and not leak on the user's side.

xtuc:
I won't have the time to test with `babel-prevet-env` before a couple of days, so don't count on this (sorry), but it _could_ provide a better UX with dealing more generally with Polyfilling wasm-bindgen generated stuff.

alexcrichton:
Oh sure! I searched "webpack textencoder polyfill" and it pointed at this, but I by no means claim this is a good solution. This is also in no rush, so we can hold off if you want time to test babel as well

crates/cli-support/src/js/mod.rs
```diff
@@ -1065,58 +1065,32 @@ impl<'a> Context<'a> {
...
             return;
         }
+        self.expose_text_processor("TextDecoder");
+    }
+
+    fn expose_text_processor(&mut self, s: &str) {
         if self.config.nodejs_experimental_modules {
-            self.imports
-                .push_str("import { TextDecoder } from 'util';\n");
+            self.imports.push_str(&format!("impot {{ {} }} from 'util';\n", s));
```
ernieturner: I assume this should be `import ...`?

alexcrichton: Oops yes indeed! Thanks for testing regardless!

alexcrichton:
@xtuc also sorry but to confirm, do you mean inlining the implementations of `TextEncoder` and `TextDecoder` into the generated JS?

alexcrichton:
It makes sense yeah but IIRC we specifically in the last mtg decided to not include the polyfill ourselves but rather rely on bundlers?

xtuc:
Yes, that correct. I assumed that we wanted to ship this quickly, and the less likely way to break is to include it. 

Once that we have a stable polyfilling story, we could switch to bundlers. One issue is for other platforms like the fetch API on Nodejs?

alexcrichton:
Oh I don't think we're in a rush per se for this so I think it's fine to take ab it longer. For `fetch` is that going to be implemented in Node? I was thinking this'd just be used for "clearly should be implemented here but just hasn't been done so already"

alexcrichton:
I'm gonna go ahead and merge this for the time being, but we can of course iterate on strategies for these types in Edge!

clouds56:
Why can't we provide `TextEncoder`/`TextDecoder` ourselves? I thought it would be as easy as convert between `std::str::String` and `Vec<u16>`
```rust
  #[wasm_bindgen]
  pub fn text_encode(s: JsString) -> Vec<u8> {
    let length = s.length();
    let utf16 = (0..length).map(|i| s.char_code_at(i) as u16).collect::<Vec<_>>();
    let string = String::from_utf16_lossy(&utf16);
    string.into()
  }
  #[wasm_bindgen]
  pub fn text_decode(v: &[u8]) -> JsString {
    let string = String::from_utf8_lossy(v);
    let utf16 = string.encode_utf16().collect::<Vec<_>>();
    let result = JsString::from_char_code(&utf16);
    drop(result.clone());
    result
  }
```

Pauan:
@clouds56 That would probably work, yeah, but since Edge 79+ now supports TextEncoder/TextDecoder, a polyfill isn't really needed anymore.

clouds56:
Sure, most browser support TextEncoder/TextDecoder now.
But I think it is the only dependency in web API, (while other things like Uint8Array are dependencies of Javascript API), and seems every wasm module would requires this (for `__wbindgen_throw`)
I think it would be great if we could **optionally** depends on TextDecoder.

I tried to write the 2 api via wasm_bindgen like above, but I cannot load it since this module also requires TextDecoder. It's a catch 22.

Pauan:
@clouds56 What is your use case? Every platform supported by wasm-bindgen supports `TextEncoder` and `TextDecoder`.

PR #915, opened 2018-10-01
Add a number of `#[inline]` annotation through crates

alexcrichton:
Adding `#[inline]` will typically improve codegen for optimized builds
without LTO (so far the majority in practice) by allowing functions that
otherwise couldn't be inlined across codegen units to get inlined
across codegen units.

Right now `wasm-bindgen` has a lot of functions that are very small and
delegate to other functions, but aren't otherwise candidates for
inlining because they're concrete.

I was poking around in release-mode wasm recently and noticed an
alarming number of functions for tiny pieces of functionality, which
motivates this patch!


fitzgen:
I was just thinking the other day that making all generated import functions `#[inline]` is probably worth it.

PR #918, opened 2018-10-02
Implement inline conversion to/from JS values 

### LINKED ISSUE #811: Support converting js_sys typed arrays to Vec
Currently there's no way to convert `TypedArray`s to `Vec<T>`s in rust, e.g., `js_sys::Uint8Array` to `Vec<u8>`, even though one can pass (via an implicit copy) a `Uint8Array` across the wasm boundary to rust functions expecting `Vec<u8>` or `&[u8]`. See [this comment](https://github.com/rustwasm/wasm-bindgen/issues/5#issuecomment-420401682) for motivation.

alexcrichton:
These commits were originally spawned from https://github.com/rustwasm/wasm-bindgen/issues/811. When working with `web-sys` not all values are well typed, but rather some values are simply `JsValue` and they're worked with elsewhere. Those values, however, sometimes need to be manufactured or interpreted as specific Rust types without handy conversions available.

This commit adds two new methods, `JsValue::{from_wasm_abi,into_wasm_abi}`, which can convert to/from a `JsValue` using the `{From,Into}WasmAbi` traits.

The implementation here is a bit heinous as it relies on the same closure rewriting logic we already have. I'm also starting to get a little worried abou the large slew of conversions we're accruing without necessarily a lot of direction about which conversion should be used when. Some thoughts on this would definitely be appreciated!

Closes #811


alexcrichton:
Ah sorry I've let this languish a bit! I agree that this is probably adding too much surface area for conversions to be worth it, and it makes it a bit confusing as to what's being converted where and when. I haven't had a chance to sit down and try to think about the fundamental value this is adding (if there's anything other than the convenient slice <-> jsvalue conversions), but in the meantime I think it's best that we defer this to a later date

dakom:
Just to clarify- by sending I mean passing from Rust to JS via a JsValue, likely by way of a js-sys Function callback

alexcrichton:
Good questions! For now the easiest way is to use `wasm_bindgen::memory` and create an offset directly into wasm's own memory and then hand that off, that way there's no copying/allocations!

dakom:
Thanks!

Do you mind explaining a bit more the relationship between `JsValue::from_wasm_abi` and `wasm_bindgen::memory` when it comes to passing a `[f64]` to a js-sys Function callback (and have it arrive in JS as a `Float64Array`)

I'm happy to tinker around until it works - but I'd appreciate the leg-up!

chinedufn:
@dakom here's a quick example in case it helps!

```rust
let memory_buffer = wasm_bindgen::memory()
  .dyn_into::<WebAssembly::Memory>()
  .unwrap()
  .buffer();

let vertices: [f32; 12] = [
   0., 0., 0., // Bottom Left
    1., 0., 0., // Bottom Right
    1., 0., -1., // Top Right
    0., 0., -1., // Top Left
];

let vertices_location = vertices.as_ptr() as u32 / 4;

let vert_array = js_sys::Float32Array::new(&memory_buffer)
  .subarray(vertices_location, vertices_location + vertices.len() as u32);

let indices: [u16; 6] = [0, 1, 2, 0, 2, 3];

let indices_location = indices.as_ptr() as u32 / 2;

let indices_array = js_sys::Uint16Array::new(&memory_buffer)
  .subarray(indices_location, indices_location + indices.len() as u32);
```

For `f64` you'd just need to divide by 8 (I haven't used f64 myself but I'd imagine this would work fine)

dakom:
@chinedufn - I've finally gotten around to actually playing with this hands-on, and having a bit of trouble understanding... is all that pointer math also needed when staying directly in Rust/Wasm (and not crossing the boundary like I was asking above)?

At a glance - it seems the goal of this PR was to allow something like `js_sys::from(&vertices);` ?

chinedufn:
Sorry just saw this! - not entirely sure what you're asking

dakom:
Nevermind - I have a slightly better understanding of what's going on now, and love the idea of being able to pass a slice of WASM memory as an ArrayBufferView :D

PR #938, opened 2018-10-06
Fix unimportant typo

killercup:
Wanted to play with this and noticed a missing 'm'


afdw:
There are 14 other places with the same typo: https://github.com/rustwasm/wasm-bindgen/search?q=coments It would be good to fix them as well.

killercup:
Oh wow, of course! Will do!

PR #945, opened 2018-10-08
Don't use JSON for custom section format

alexcrichton:
This commit migrates away from using Serde for the custom section in
wasm executables. This is a refactoring of a purely-internal data
structure to `wasm-bindgen` and should have no visible functional change
on users.

The motivation for this commit is two fold:

* First, the compile times using `serde_json` and `serde_derive` for the
  syntax extension isn't the most fun.
* Second, eventually we're going to want to stablize the layout of the
  custom section, and it's highly unlikely to be json!

Primarily, though, the intention of this commit is to improve the
cold-cache compile time of `wasm-bindgen` by ensuring that for new users
this project builds as quickly as possible. By removing some heavyweight
dependencies from the procedural macro, `serde`, `serde_derive`, and
`serde_json`, we're able to get a pretty nice build time improvement for
the `wasm-bindgen` crate itself:

|             | single-core build | parallel build |
|-------------|-------------------|----------------|
| master      |             36.5s |          17.3s |
| this commit |             20.5s |          11.8s |

These are't really end-all-be-all wins but they're much better
especially on the spectrum of weaker CPUs (in theory modeled by the
single-core case showing we have 42% less CPU work in theory).

### DIFF
--- a/crates/backend/Cargo.toml
+++ b/crates/backend/Cargo.toml
@@ -19,6 +19,5 @@ lazy_static = "1.0.0"
 log = "0.4"
 proc-macro2 = "0.4.8"
 quote = '0.6'
-serde_json = "1.0"
-syn = { version = '0.15', features = ['full', 'visit'] }
+syn = { version = '0.15', features = ['full'] }
 wasm-bindgen-shared = { path = "../shared", version = "=0.2.25" }
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -1,8 +1,7 @@
+use Diagnostic;
 use proc_macro2::{Ident, Span};
 use shared;
 use syn;
-
-use Diagnostic;
 
 /// An abstract syntax tree representing a rust program. Contains
 /// extra information for joining up this rust code with javascript.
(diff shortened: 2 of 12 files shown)


alexcrichton:
It's worth pointing out that these are relatively modest gains and in any large app are guaranteed to be swallowed by up the compile time of all the surrounding crates. In that sense this is hyper-focusing on the one use case of "new users who just started working on a small project". I want to be careful to not over-rotate on this use case (but also not under-value it!). 

To that end I'd be particularly curious to see other's thoughts on the code organization here. While the `wasm-bindgen-shared` crate no longer uses `#[derive]` to keep things in sync, it still exists and keeps things in sync via a `macro_rules!` soup of macros. I definitely don't think this is as clear as serialize/deserialize, but I feel that the complexity here is striking the right balance between compile times and edit-ability. That is, we can edit everything pretty easily (if you ignore the definitions of the macros) and everything should be kept in sync naturally.

alexcrichton:
Oh FWIW that's a "solved problem" effectively in terms of both share [these API definitions](https://github.com/rustwasm/wasm-bindgen/blob/d1a71fcce676ea8e7b403615b607d2bb3b1f5052/crates/shared/src/lib.rs#L7-L115) which is parsed by macros for encoding/decoding. The implementation of base primitives (like how to encode a `u32`) is "duplicated" but that's sort of to be expected (and a closed, finite, set)

PR #947, opened 2018-10-10
Use the global allocator, not the system allocator

fitzgen:
This was previously causing us to accidentally always pull in the system allocator, even if users were trying to just use a custom global allocator.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -809,7 +809,7 @@ pub mod __rt {
     }
 
     if_std! {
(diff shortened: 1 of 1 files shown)


frehberg:
In case of using different global allocator (for example wee_alloc), without this patch the functions __wbindgen_malloc etc. would still link against dlmalloc, increasing wasm-binary-footprint by around 20KB, and causing issues using two different dynamic memory allocators, internally and the one used by exported functions.

PR #954, opened 2018-10-10
Defer future execution to the next event loop tick

alexcrichton:
Previously whenever a future readiness notification came in we would
immediately start polling a future. This ends up having two downsides,
however:

* First, the stack depth may run a risk of getting blown. There's no
  recursion limit to defer execution to later, which means that if
  futures are always ready we'll keep making the stack deeper.

* Second, and more worrisome in the near term, apparently future
  adapaters in the `futures` crate (namely the unsync oneshot channel)
  doesn't actually work if you immediately poll on readiness. This may
  or may not be a bug in the `futures` crate but it's good to fix it
  here anyway.

As a result whenever a future is ready to get polled again we defer its
polling to the next turn of the event loop. This should ensure that the
current call stack is always drained and we're effectively enqueueing
the future to be polled in the near future.

### DIFF
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -108,6 +108,7 @@ extern crate js_sys;
 extern crate wasm_bindgen;
 
 use std::cell::{Cell, RefCell};
+use std::rc::Rc;
 use std::sync::Arc;
 
 use futures::executor::{self, Notify, Spawn};
@@ -336,15 +337,14 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
 
     impl Notify for Package {
         fn notify(&self, _id: usize) {
-            match self.notified.replace(State::Notified) {
-                // we need to schedule polling to resume, so we do so
-                // immediately for now
-                State::Waiting(me) => Package::poll(&me),
+            let me = match self.notified.replace(State::Notified) {
+                // we need to schedule polling to resume, so keep going
+                State::Waiting(me) => me,
 
                 // we were already notified, and were just notified again;
                 // having now coalesced the notifications we return as it's
                 // still someone else's job to process this
-                State::Notified => {}
+                State::Notified => return,
 
                 // the future was previously being polled, and we've just
                 // switched it to the "you're notified" state. We don't have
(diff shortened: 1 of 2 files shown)


crates/futures/src/lib.rs
```diff
@@ -353,8 +353,32 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
                 // continue polling. For us, though, there's nothing else to do,
                 // so we bail out.
                 // later see
-                State::Polling => {}
-            }
+                State::Polling => return,
+            };
+
+            // Use `setTimeout` to place our execution onto the next turn of the
+            // event loop, enqueueing our poll operation. We don't currently
```
fitzgen: I don't think we want to use `setTimeout` -- that will incure larger-than-necessary latency.

Instead, we should do the equivalent of `Promise.resolve().then(() => executeOurStuff())` which will execute on the micro-task queue, and not bounce all the way out to the browser's main event loop.

Pauan: In stdweb [we use `MutationObserver` (falling back to `Promise.prototype.then` on Nodejs)](https://github.com/koute/stdweb/blob/bc07dac9d275e6d0b517d304527d3a593df208f6/src/webcore/executor.rs#L268-L310).

The reason we do this is because `Promise.prototype.then` allocates (to create the new Promise), whereas with `MutationObserver` we can schedule the next microtask tick without any allocation (well, except the allocation for the array of `MutationRecord`, I guess).

We also handle everything in Rust as much as possible:

1. It [maintains a `VecDeque` of pending tasks](https://github.com/koute/stdweb/blob/bc07dac9d275e6d0b517d304527d3a593df208f6/src/webcore/executor.rs#L111).

2. When there are multiple pending tasks, [it ensures that it only schedules a single microtask tick](https://github.com/koute/stdweb/blob/bc07dac9d275e6d0b517d304527d3a593df208f6/src/webcore/executor.rs#L220-L228).

3. When the microtask tick finishes, [it then loops over the pending tasks](https://github.com/koute/stdweb/blob/bc07dac9d275e6d0b517d304527d3a593df208f6/src/webcore/executor.rs#L165).

This ensures that no matter how many tasks are pending, it only needs to schedule one microtask tick, which made a *huge* difference for performance (the old implementation was similar to wasm-bindgen and spun up a new microtask tick for each task).

There's a few reasons for the big performance boost:

* By keeping everything in Rust as much as possible, it can make use of Rust's great data structures and optimizations.

* Calling between wasm and JS is slow (at least in stdweb), so this minimizes the amount of wasm<->JS marshalling (because it only schedules one microtask tick per cycle, rather than one microtask tick per task).

* It cuts back massively on the amount of garbage that JS produced (which was causing a lot of garbage collector thrashing).

   We're talking huge differences here: with [my animation demo](https://pauan.github.io/rust-dominator/examples/animation/) the old executor got ~10 FPS (and generated a lot of garbage), but [when I changed it to use a `VecDeque`](https://github.com/koute/stdweb/pull/165) it jumped up to ~60 FPS (and generated almost no garbage).

In any case, I agree that it should be using the microtask queue (`MutationRecord` or Promises).

alexcrichton: Oh good idea, thanks for the info @fitzgen! @Pauan I think we can eventually go towards a more complicated world like that, but I'd prefer to have a benchmark and/or example which motivates it to avoid too much premature complexity.

PR #957, opened 2018-10-10
js_sys: Add `#[derive(Clone, Debug)]` to `Promise`

alexcrichton:
I think we just forgot this from earlier!


alexcrichton:
FWIW I also did a quick audit and the only ones still missing this are `JSON` and `Math`, neither of which have instances to work with anyway.

PR #961, opened 2018-10-13
Todo mvc

jonathanKingston:
A [TODO MVC](https://todomvc.com/) implementation using web-sys.

### DIFF
--- a/examples/todomvc/Cargo.toml
+++ b/examples/todomvc/Cargo.toml
@@ -1,0 +1,41 @@
+[package]
+name = "todomvc"
+version = "0.1.0"
+authors = ["The wasm-bindgen Developers"]
+
+[lib]
+crate-type = ["cdylib"]
+
+[profile.release]
+lto = true
+opt-level = 's'
+
+[build-dependencies]
+askama = "0.7.2"
+
+[dependencies]
+js-sys = { path = "../../crates/js-sys" }
+wasm-bindgen = { path = "../../" }
+askama = "0.7.2"
+
+[dependencies.web-sys]
+path = "../../crates/web-sys"
+features = [
+  'console',
+  'CssStyleDeclaration',
+  'Document',
+  'DomStringMap',
+  'DomTokenList',
+  'Element',
+  'Event',
+  'EventTarget',
+  'HtmlBodyElement',
+  'HtmlElement',
+  'HtmlInputElement',
+  'KeyboardEvent',
+  'Location',
+  'Node',
+  'NodeList',
+  'Storage',
+  'Window',
+]
--- a/examples/todomvc/README.md
+++ b/examples/todomvc/README.md
@@ -1,0 +1,17 @@
+# TODO MVC
+
+[View documentation for this example online][dox] or [View compiled example
+online][compiled]
+
+[compiled]: https://rustwasm.github.io/wasm-bindgen/exbuild/todomvc/
+[dox]: https://rustwasm.github.io/wasm-bindgen/examples/todomvc.html
+
+You can build the example locally with:
+
+```
+$ ./build.sh
+```
+
+(or running the commands on Windows manually)
+
+and then visiting http://localhost:8080 in a browser should run the example!
--- a/examples/todomvc/build.sh
+++ b/examples/todomvc/build.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# For more comments about what's going on here, see the `hello_world` example
+
+set -ex
+cd "$(dirname $0)"
+
+cargo +nightly build --target wasm32-unknown-unknown
+
+cargo +nightly run --manifest-path ../../crates/cli/Cargo.toml \
+  --bin wasm-bindgen -- \
+  ../../target/wasm32-unknown-unknown/debug/todomvc.wasm --out-dir .
+
+npm install
+npm run serve
--- a/examples/todomvc/package.json
+++ b/examples/todomvc/package.json
@@ -1,0 +1,13 @@
+{
+  "scripts": {
+    "build": "webpack",
+    "serve": "webpack-dev-server"
+  },
+  "devDependencies": {
+    "text-encoding": "^0.7.0",
+    "html-webpack-plugin": "^3.2.0",
+    "webpack": "^4.11.1",
+    "webpack-cli": "^3.1.1",
+    "webpack-dev-server": "^3.1.0"
+  }
+}
--- a/examples/todomvc/src/element.rs
+++ b/examples/todomvc/src/element.rs
@@ -1,0 +1,208 @@
+extern crate wasm_bindgen;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::JsCast;
+
+/// Wrapper for `web_sys::Element` to simplify calling different interfaces
+pub struct Element {
+    el: Option<web_sys::Element>,
+}
+
+impl From<Element> for Option<web_sys::Node> {
+    fn from(obj: Element) -> Option<web_sys::Node> {
+        if let Some(el) = obj.el {
+            Some(el.into())
+        } else {
+            None
+        }
+    }
+}
+
+impl From<Element> for Option<web_sys::EventTarget> {
+    fn from(obj: Element) -> Option<web_sys::EventTarget> {
+        if let Some(el) = obj.el {
+            Some(el.into())
+        } else {
+            None
+        }
+    }
+}
+
+impl Element {
+    pub fn qs(selector: &str) -> Option<Element> {
+        let body: web_sys::Element = web_sys::window()?.document()?.body()?.into();
+        let el = body.query_selector(selector).ok()?;
+        Some(Element { el })
+    }
+
+    /// Add event listener to this node
+    pub fn add_event_listener<T>(&mut self, event_name: &str, handler: T)
+    where
+        T: 'static + FnMut(web_sys::Event),
+    {
+        let cb = Closure::wrap(Box::new(handler) as Box<FnMut(_)>);
+        if let Some(el) = self.el.take() {
+            let el_et: web_sys::EventTarget = el.into();
+            el_et.add_event_listener_with_callback(event_name, cb.as_ref().unchecked_ref());
+            cb.forget();
+            if let Ok(el) = el_et.dyn_into::<web_sys::Element>() {
+                self.el = Some(el);
+            }
+        }
+    }
+
+    /// Delegate an event to a selector
+    pub fn delegate<T>(
+        &mut self,
+        selector: &'static str,
+        event: &str,
+        mut handler: T,
+        use_capture: bool,
+    ) where
+        T: 'static + FnMut(web_sys::Event) -> (),
+    {
+        let el = match self.el.take() {
+            Some(e) => e,
+            None => return,
+        };
+        if let Some(dyn_el) = &el.dyn_ref::<web_sys::EventTarget>()
+        {
+            if let Some(window) = web_sys::window() {
+                if let Some(document) = window.document() {
+                    // TODO document selector to the target element
+                    let tg_el = document;
+
+                    let cb = Closure::wrap(Box::new(move |event: web_sys::Event| {
+                        if let Some(target_element) = event.target() {
+                            let dyn_target_el: Option<
+                                &web_sys::Node,
+                            > = wasm_bindgen::JsCast::dyn_ref(&target_element);
+                            if let Some(target_element) = dyn_target_el {
+                                if let Ok(potential_elements) =
+                                    tg_el.query_selector_all(selector)
+                                {
+                                    let mut has_match = false;
+                                    for i in 0..potential_elements.length() {
+                                        if let Some(el) = potential_elements.get(i) {
+                                            if target_element.is_equal_node(Some(&el)) {
+                                                has_match = true;
+                                            }
+                                            break;
+                                        }
+                                    }
+
+                                    if has_match {
+                                        handler(event);
+                                    }
+                                }
+                            }
+                        }
+                    }) as Box<FnMut(_)>);
+
+                    dyn_el.add_event_listener_with_callback_and_bool(
+                        event,
+                        cb.as_ref().unchecked_ref(),
+                        use_capture,
+                    );
+                    cb.forget(); // TODO cycle collect
+                }
+            }
+        }
+        self.el = Some(el);
+    }
+
+    /// Find child `Element`s from this node
+    pub fn qs_from(&mut self, selector: &str) -> Option<Element> {
+        let mut found_el = None;
+        if let Some(el) = self.el.as_ref() {
+            found_el = Some(Element {
+                el: el.query_selector(selector).ok()?,
+            });
+        }
+        found_el
+    }
+
+    /// Sets the inner HTML of the `self.el` element
+    pub fn set_inner_html(&mut self, value: String) {
+        if let Some(el) = self.el.take() {
+            el.set_inner_html(&value);
+            self.el = Some(el);
+        }
+    }
+
+    /// Sets the text content of the `self.el` element
+    pub fn set_text_content(&mut self, value: &str) {
+        if let Some(el) = self.el.as_ref() {
+            if let Some(node) = &el.dyn_ref::<web_sys::Node>() {
+                node.set_text_content(Some(&value));
+            }
+        }
+    }
+
+    /// Removes a class list item from the element
+    ///
+    /// ```
+    /// e.class_list_remove(String::from("clickable"));
+    /// // removes the class 'clickable' from e.el
+    /// ```
+    pub fn class_list_remove(&mut self, value: &str) {
+        if let Some(el) = self.el.take() {
+            el.class_list().remove_1(&value);
+            self.el = Some(el);
+        }
+    }
+
+    /// Given another `Element` it will remove that child from the DOM from this element
+    /// Consumes `child` so it can't be used after it's removal.
+    pub fn remove_child(&mut self, mut child: Element) {
+        if let Some(child_el) = child.el.take() {
+            if let Some(el) = self.el.take() {
+                if let Some(el_node) = el.dyn_ref::<web_sys::Node>() {
+                    let child_node: web_sys::Node = child_el.into();
+                    el_node.remove_child(&child_node);
+                }
+                self.el = Some(el);
+            }
+        }
+    }
+
+    /// Sets the whole class value for `self.el`
+    pub fn set_class_name(&mut self, class_name: &str) {
+        if let Some(el) = self.el.take() {
+            el.set_class_name(&class_name);
+            self.el = Some(el);
+        }
+    }
+
+    /// Sets the visibility for the element in `self.el`
+    pub fn set_visibility(&mut self, visible: bool) {
+        if let Some(el) = self.el.take() {
+            {
+                let dyn_el: Option<&web_sys::HtmlElement> = wasm_bindgen::JsCast::dyn_ref(&el);
+                if let Some(el) = dyn_el {
+                    el.set_hidden(!visible);
+                }
+            }
+            self.el = Some(el);
+        }
+    }
+
+    /// Sets the visibility for the element in `self.el` (The element must be an input)
+    pub fn set_value(&mut self, value: &str) {
+        if let Some(el) = self.el.take() {
+            if let Some(el) = wasm_bindgen::JsCast::dyn_ref::<web_sys::HtmlInputElement>(&el) {
+                el.set_value(&value);
+            }
+            self.el = Some(el);
+        }
+    }
+
+    /// Sets the checked state for the element in `self.el` (The element must be an input)
+    pub fn set_checked(&mut self, checked: bool) {
+        if let Some(el) = self.el.take() {
+            if let Some(el) = wasm_bindgen::JsCast::dyn_ref::<web_sys::HtmlInputElement>(&el) {
+                el.set_checked(checked);
+            }
+            self.el = Some(el);
+        }
+    }
+}
--- a/examples/todomvc/src/lib.rs
+++ b/examples/todomvc/src/lib.rs
@@ -1,0 +1,70 @@
+//! # TODO MVC
+//!
+//! A [TODO MVC](https://todomvc.com/) implementation written using [web-sys](https://rustwasm.github.io/wasm-bindgen/web-sys/overview.html)
+#![warn(missing_docs)]
+
+extern crate wasm_bindgen;
+use wasm_bindgen::prelude::*;
+
+extern crate js_sys;
+extern crate web_sys;
+use std::rc::Rc;
+
+#[macro_use]
+extern crate askama;
+
+/// Controller of the program
+pub mod controller;
+/// Element wrapper to the DOM
+pub mod element;
+/// Schedule messages to the Controller and View
+pub mod scheduler;
+/// Stores items into localstorage
+pub mod store;
+/// Handles constructing template strings from data
+pub mod template;
+/// Presentation layer
+pub mod view;
+use crate::controller::{Controller, ControllerMessage};
+use crate::scheduler::Scheduler;
+use crate::store::Store;
+use crate::view::{View, ViewMessage};
+
+/// Message wrapper enum used to pass through the scheduler to the Controller or View
+pub enum Message {
+    /// Message wrapper to send to the controller
+    Controller(ControllerMessage),
+    /// Message wrapper to send to the view
+    View(ViewMessage),
+}
+
+/// Used for debugging to the console
+pub fn exit(message: &str) {
+    let v = wasm_bindgen::JsValue::from_str(&message.to_string());
+    web_sys::console::exception_1(&v);
+    std::process::abort();
+}
+
+fn app(name: &str) {
+    let sched = Rc::new(Scheduler::new());
+    let store = match Store::new(name) {
+        Some(s) => s,
+        None => return,
+    };
+    let controller = Controller::new(store, Rc::downgrade(&sched));
+    if let Some(mut view) = View::new(sched.clone()) {
+        let sch: &Rc<Scheduler> = &sched;
+        view.init();
+        sch.set_view(view);
+        sch.set_controller(controller);
+        sched.add_message(Message::Controller(ControllerMessage::SetPage(
+            "".to_string(),
+        )));
+    }
+}
+
+/// Entry point into the program from JavaScript
+#[wasm_bindgen]
+pub fn run() {
+    app("todos-wasmbindgen");
+}
--- a/examples/todomvc/src/scheduler.rs
+++ b/examples/todomvc/src/scheduler.rs
@@ -1,0 +1,138 @@
+use crate::controller::Controller;
+use crate::exit;
+use crate::view::View;
+use crate::Message;
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// Creates an event loop that starts each time a message is added
+pub struct Scheduler {
+    controller: Rc<RefCell<Option<Controller>>>,
+    view: Rc<RefCell<Option<View>>>,
+    events: RefCell<Vec<Message>>,
+    running: RefCell<bool>,
+}
+
+impl Scheduler {
+    /// Construct a new `Scheduler`
+    pub fn new() -> Scheduler {
+        Scheduler {
+            controller: Rc::new(RefCell::new(None)),
+            view: Rc::new(RefCell::new(None)),
+            events: RefCell::new(Vec::new()),
+            running: RefCell::new(false),
+        }
+    }
+
+    pub fn set_controller(&self, controller: Controller) {
+        if let Ok(mut controller_data) = self.controller.try_borrow_mut() {
+            controller_data.replace(controller);
+        } else {
+            exit("This might be a deadlock");
+        }
+    }
+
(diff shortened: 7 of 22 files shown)


examples/todomvc/README.md
```diff
@@ -0,0 +1,18 @@
+# TODO MVC using wasm-bingen and web-sys
+
+[wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) and [web-sys](https://rustwasm.github.io/wasm-bindgen/api/web_sys/) coded [TODO MVC](https://todomvc.com/)
+
+The code was rewritten from the [ECMA 6 version](http://todomvc.com/examples/vanilla-es6/).
```
sendilkumarn: ES6 or ECMAScript 6 :)

examples/todomvc/README.md
```diff
@@ -0,0 +1,18 @@
...
+The core differences are:
+- Having an [Element wrapper](/src/element.rs) that takes care of dyn refs in web-sys
+- A [Scheduler](/src/scheduler.rs) that allows Controller and View to speak to each other by emulating something similar to the JS event loop.
+
+
+## Size
+
+The size of the project hasn't undergone much work to make it optimised yet.
+
+~96kb release build
```
sendilkumarn: Love this ❤️

examples/todomvc/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "todomvc"
+version = "0.1.0"
+authors = ["The wasm-bindgen Developers"]
+
+[lib]
+crate-type = ["cdylib"]
+
+[profile.release]
+debug = false
```
alexcrichton: I believe this is already implied by `profile.release`, so it should be fine to remove

examples/todomvc/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+  'HtmlBodyElement',
+  'Storage',
+  'CssStyleDeclaration',
+  'DomTokenList',
+  'DomStringMap',
+  'Event',
+  'KeyboardEvent',
+  'EventTarget',
+  'NodeList',
+  'Location'
```
alexcrichton: Could this list be sorted?

examples/todomvc/README.md
```diff
@@ -0,0 +1,18 @@
+# TODO MVC using wasm-bingen and web-sys
+
+[wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) and [web-sys](https://rustwasm.github.io/wasm-bindgen/api/web_sys/) coded [TODO MVC](https://todomvc.com/)
+
+The code was rewritten from the [ECMA 6 version](http://todomvc.com/examples/vanilla-es6/).
+
+The core differences are:
```
alexcrichton: Thanks for writing this up! Could this be added as a page in the examples guide at `guide/src/examples`? The other examples round here have a bland README template as well for what to put in the repo here.

examples/todomvc/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+cargo +nightly build --target wasm32-unknown-unknown --verbose --release &&
+wasm-bindgen ../../target/wasm32-unknown-unknown/release/todomvc.wasm --out-dir . &&
+npm install &&
+npm run serve
```
alexcrichton: Could this copy the `build.sh` scripts from the other examples to have a few extra comments as well? (also use `wams-bindgen` as a path dependency)

examples/todomvc/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "scripts": {
+    "serve": "webpack-dev-server"
+  },
+  "devDependencies": {
+    "webpack": "^4.11.1",
+    "webpack-cli": "^3.1.0",
+    "webpack-dev-server": "^3.1.0"
+  }
+}
```
alexcrichton: Could this copy the `package.json` of other examples to ensure consistency as well>?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn qs(selector: &str) -> Option<Element> {
+        let body: web_sys::Element = web_sys::window()?.document()?.body()?.into();
+        let el = body.query_selector(selector).ok()?;
+        Some(Element { el })
+    }
+
+    /// Add event listener to this node
+    pub fn add_event_listener<T>(&mut self, event_name: &str, handler: T)
+    where
+        // TODO rewrite without static
```
alexcrichton: FWIW writing this without `'static` would actually be quite hard (and I'm not sure it could be sound)?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+        T: 'static + FnMut(web_sys::Event) -> (),
+    {
+        let cb = Closure::wrap(Box::new(handler) as Box<FnMut(_)>);
+        if let Some(el) = self.el.take() {
+            let el_et: web_sys::EventTarget = el.into();
+            el_et.add_event_listener_with_callback(event_name, cb.as_ref().unchecked_ref());
+            cb.forget(); // TODO cycle collect this
+            if let Ok(el) = wasm_bindgen::JsCast::dyn_into(el_et): Result<web_sys::Element, _> {
```
alexcrichton: could `el_et.dyn_into::<web_sys::Element>()` be used here to avoid the type ascription?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+        let body: web_sys::Element = web_sys::window()?.document()?.body()?.into();
+        let el = body.query_selector(selector).ok()?;
+        Some(Element { el })
+    }
+
+    /// Add event listener to this node
+    pub fn add_event_listener<T>(&mut self, event_name: &str, handler: T)
+    where
+        // TODO rewrite without static
+        T: 'static + FnMut(web_sys::Event) -> (),
```
alexcrichton: I think the `-> ()` can be removed here as well

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn delegate<T>(
+        &mut self,
+        selector: &'static str,
+        event: &str,
+        mut handler: T,
+        use_capture: bool,
+    ) where
+        T: 'static + FnMut(web_sys::Event) -> (),
+    {
+        if let Some(el) = self.el.take() {
```
alexcrichton: Could this use an early return to avoid the indentation?

```rust
let el = match self.el.take() {
    Some(e) => e,
    None => return,
};
// ...
```

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+            //  if let Some(element_et) = self.into(): Option<web_sys::EventTarget> {
+            //let element_et: web_sys::EventTarget = el.into();
+            if let Some(dyn_el) = wasm_bindgen::JsCast::dyn_ref(&el): Option<&web_sys::EventTarget>
```
alexcrichton: Same as above w/ `el.dyn_into()`

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+        selector: &'static str,
+        event: &str,
+        mut handler: T,
+        use_capture: bool,
+    ) where
+        T: 'static + FnMut(web_sys::Event) -> (),
+    {
+        if let Some(el) = self.el.take() {
+            //  if let Some(element_et) = self.into(): Option<web_sys::EventTarget> {
+            //let element_et: web_sys::EventTarget = el.into();
```
alexcrichton: Accidental leftovers?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+                        cb.forget(); // TODO cycle collect
+                    }
+                }
+            }
+            self.el = Some(el);
+        }
+    }
+
+    /// Find child `Element`s from this node
+    pub fn qs_from(&mut self, selector: &str) -> Option<Element> {
```
alexcrichton: Does this require `&mut`? It seems like `as_ref()` could be used perhaps?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+            found_el = Some(Element {
+                el: el.query_selector(selector).ok()?,
+            });
+            self.el = Some(el);
+        }
+        found_el
+    }
+
+    /// Sets the inner HTML of the `self.el` element
+    pub fn set_inner_html(&mut self, value: String) {
```
alexcrichton: This technically only requires `&self` as it's all interior mutability down to JS, but I could understand this as wanting to start to impose Rust stylistic idioms here too

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+        if let Some(el) = self.el.take() {
+            el.set_inner_html(&value);
+            self.el = Some(el);
+        }
+    }
+
+    /// Sets the text content of the `self.el` element
+    pub fn set_text_content(&mut self, value: String) {
+        if let Some(el) = self.el.take() {
```
alexcrichton: This pattern seems a little odd to me, is there a reason `.as_ref()` without moving ownership wasn't used?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+            if let Some(node) = wasm_bindgen::JsCast::dyn_ref(&el): Option<&web_sys::Node> {
```
alexcrichton: Another instance of `el.dyn_ref::<T>()` to use here too

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    /// Removes a class list item from the element
+    ///
+    /// ```
+    /// e.class_list_remove(String::from("clickable"));
+    /// // removes the class 'clickable' from e.el
+    /// ```
+    pub fn class_list_remove(&mut self, value: String) {
```
alexcrichton: Perhaps `value: &str`?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
+use crate::dbg;
+extern crate wasm_bindgen;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::JsCast;
+
+/// Wrapper for `web_sys::Element` to simplify calling different interfaces
+pub struct Element {
```
alexcrichton: Should this `#[derive(Clone)]` for more ergonomic usage in Rust? (it's sort of like an `Rc` basically)

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+                        // TODO document selector to the target element
+                        let tg_el = document;
+
+                        let cb = Closure::wrap(Box::new(move |event: web_sys::Event| {
+                            dbg("got fn call delegated");
+                            if let Some(target_element) = event.target() {
+                                let dyn_target_el: Option<
+                                    &web_sys::Node,
+                                > = wasm_bindgen::JsCast::dyn_ref(&target_element);
+                                if let Some(target_element) = dyn_target_el {
```
alexcrichton: Some early returns here could definitely help out with the indentation!

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+                                    if let Ok(potential_elements) =
+                                        tg_el.query_selector_all(selector)
+                                    {
+                                        let mut has_match = false;
+                                        /*
+                                        dbg(format!(
```
alexcrichton: stray commented code?

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+                    let child_node: web_sys::Node = child_el.into();
+                    el_node.remove_child(&child_node);
+                }
+                self.el = Some(el);
+            }
+        }
+    }
+
+    /// Sets the whole class value for `self.el`
+    pub fn set_class_name(&mut self, class_name: String) {
```
alexcrichton: `class_name: &str` here I think

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// ```
+    pub fn class_list_remove(&mut self, value: String) {
+        if let Some(el) = self.el.take() {
+            el.class_list().remove_1(&value);
+            self.el = Some(el);
+        }
+    }
+
+    /// Given another `Element` it will remove that child from the DOM from this element
+    pub fn remove_child(&mut self, mut child: Element) {
```
alexcrichton: This is another case where it could be `child: &Element`, but it depends on the goals of this I think (taking ownership is a Rust-ism which make more sense here)

examples/todomvc/src/element.rs
```diff
@@ -0,0 +1,235 @@
...
+                if let Some(el) = dyn_el {
+                    el.set_hidden(!visible);
+                }
+            }
+            self.el = Some(el);
+        }
+    }
+
+    /// Sets the visibility for the element in `self.el` (The element must be an input)
+    pub fn set_value(&mut self, value: String) {
```
alexcrichton: `value: &str` here

examples/todomvc/src/lib.rs
```diff
@@ -0,0 +1,66 @@
+//! # TODO MVC
+//!
+//! A [TODO MVC](https://todomvc.com/) implementation written using [web-sys](https://rustwasm.github.io/wasm-bindgen/web-sys/overview.html)
+#![feature(type_ascription)]
+#![feature(option_replace)]
```
alexcrichton: Could this avoid nightly features to compile on stable?

examples/todomvc/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    View(ViewMessage),
+}
+
+/// Used for debugging to the console
+pub fn dbg(message: &str) {
+    let v = wasm_bindgen::JsValue::from_str(&message.to_string());
+    web_sys::console::log_1(&v);
+}
+
+fn app(name: &str) -> Option<()> {
```
alexcrichton: How come the `Option<()>` return value here?

examples/todomvc/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    use std::borrow::Borrow;
+    let sched = Rc::new(Scheduler::new());
+    let store = Store::new(name)?;
+    let controller = Controller::new(store, Rc::downgrade(&sched));
+    let mut view = View::new(sched.clone())?;
+    let sch: &Rc<Scheduler> = sched.borrow();
```
alexcrichton: I think `.borrow()` here might just be the `&` operator?

examples/todomvc/src/scheduler.rs
```diff
@@ -0,0 +1,138 @@
...
+            events: RefCell::new(Vec::new()),
+            running: RefCell::new(false),
+        }
+    }
+
+    pub fn set_controller(&self, controller: Controller) {
+        if let Ok(mut controller_data) = self.controller.try_borrow_mut() {
+            controller_data.replace(controller);
+        } else {
+            dbg("This might be a deadlock");
```
alexcrichton: Should this be a panic instead of a debug log? (same with many of the others below)

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+pub struct Store {
+    local_storage: web_sys::Storage,
+    data: Option<ItemList>,
+    name: String,
+}
+
+impl Store {
+    /// Creates a new store with `name` as the localstorage value name
+    pub fn new(name: &str) -> Option<Store> {
+        if let Some(window) = web_sys::window() {
```
alexcrichton: Could this use `?` to tighten this up a bit?

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+    /// Caches the store into `self.data` to reduce calls to JS
+    ///
+    /// Uses mut here as the return is something we might want to manipulate
+    ///
+    fn get_local_storage(&mut self) -> &Option<ItemList> {
+        let mut item_list = ItemList::new();
+        if let Some(ref _data) = self.data {
+            return &self.data;
+        }
+        if let Ok(Some(value)) = self.local_storage.get_item(&self.name) {
```
alexcrichton: Heh This is definitely a case where we probably want to use `?` :)

The outer function here can return `Option<&ItemList>` and that should enable `?` to be used everywhere here.

alexcrichton: (either that or I think some `and_then` and/or early returns/continues should be used to avoid the indentation)

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+            let s = item.title.clone();
+            child.push(&JsValue::from(&s));
+            child.push(&JsValue::from(item.completed));
+            child.push(&JsValue::from(item.id.to_string()));
+
+            array.push(&JsValue::from(child));
+        }
+        if let Ok(storage_string) = JSON::stringify(&JsValue::from(array)) {
+            let storage_string: String = storage_string.to_string().into();
+            dbg(&storage_string);
```
alexcrichton: Is it intended to leave these in? Or shoudl they be removed at the end?

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+        if let Some(todos) = self.data.take() {
+            let todos = todos.into_iter();
+
+            let todos = todos
+                .map(|mut todo| {
+                    if id == todo.id {
+                        todo.update(&update);
+                    }
+                    todo
+                }).collect();
```
alexcrichton: Instead of `into_iter()` and `collect()` at the end, could this use an iterator which doesn't consume?

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+
+    /// Update an item in the Store.
+    ///
+    /// `ItemUpdate` update Record with an id and a property to update
+    pub fn update(&mut self, update: ItemUpdate) {
+        let id = update.id();
+        self.get_local_storage();
+        if let Some(todos) = self.data.take() {
```
alexcrichton: Is the take/set here necessary? Or could we reference this as it just lives in the storage?

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+            self.set_local_storage(todos);
+        }
+    }
+
+    /// Insert an item into the Store.
+    ///
+    /// `Item` item Item to insert
+    pub fn insert(&mut self, item: Item) {
+        self.get_local_storage();
+        if let Some(mut todos) = self.data.take() {
```
alexcrichton: I wonder if instead of take/set there could perhaps be a method like `sync_to_local_storage` or something like that?

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+                }).collect();
+            self.set_local_storage(todos);
+        }
+    }
+
+    /// Insert an item into the Store.
+    ///
+    /// `Item` item Item to insert
+    pub fn insert(&mut self, item: Item) {
+        self.get_local_storage();
```
alexcrichton: It might be good to have a precondition to all these methods that they're already fetched from local storage, as then it shouldn't be necessary to load this at the start of all functions I think?

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+
+    /// Remove items from the Store based on a query.
+    /// query is an `ItemQuery` query Query matching the items to remove
+    pub fn remove(&mut self, query: ItemQuery) {
+        self.get_local_storage();
+        if let Some(todos) = self.data.take() {
+            let todos = todos
+                .into_iter()
+                .filter(|todo| !query.matches(todo))
+                .collect();
```
alexcrichton: This could use `Vec::retain` to avoid the into_iter/collect

examples/todomvc/src/store.rs
```diff
@@ -0,0 +1,321 @@
...
+pub struct ItemListSlice<'a> {
+    list: Vec<&'a Item>,
+}
+impl<'a> ItemListSlice<'a> {
+    /*
+TODO implement?
+    fn new(boxed_slice: Box<[&'a Item]>) -> ItemListSlice<'a> {
+        ItemListSlice { list: boxed_slice }
+    }
+*/
```
alexcrichton: stray dead code?

examples/todomvc/src/template.rs
```diff
@@ -0,0 +1,54 @@
+use store::{ItemList, ItemListTrait};
+
+fn escape_html(val: String) -> String {
```
alexcrichton: One thing that might be neat to show off is pulling in a crate for this, and presumably one of these already exists on crates.io?

examples/todomvc/src/template.rs
```diff
@@ -0,0 +1,54 @@
...
+    val.replace('&', "&alt;").replace('<', "&lt;").replace('>', "&gt;").replace('"', "&quot").replace('\'', "&#39;")
+}
+
+pub struct Template {}
```
alexcrichton: Along the lines of showing off crates.io it might be cool to pull in something like handlebars here

examples/todomvc/src/view.rs
```diff
@@ -0,0 +1,478 @@
...
+            main,
+            toggle_all,
+            new_todo,
+            callbacks: Vec::new(),
+        })
+    }
+
+    pub fn init(&mut self) {
+        if let Some(window) = web_sys::window() {
+            if let Some(document) = window.document() {
```
alexcrichton: Another few cases for early-return to help reduce indentation

examples/todomvc/src/view.rs
```diff
@@ -0,0 +1,478 @@
...
+                        }
+                    }
+                }) as Box<FnMut()>);
+
+                let window_et: web_sys::EventTarget = window.into();
+                window_et.add_event_listener_with_callback(
+                    "hashchange",
+                    set_page.as_ref().unchecked_ref(),
+                );
+                set_page.forget(); // Cycle collect this
```
alexcrichton: It looks like this is pretty common throughout the app (forgetting event listener callbacks). Is this happening often like with every click? If so, would it be good to figure out a way to deallocate these?

examples/todomvc/src/view.rs
```diff
@@ -0,0 +1,478 @@
...
+        }
+    }
+
+    /// Populate the todo list with a list of items.
+    fn show_items(&mut self, items: ItemList) {
+        self.todo_list.set_inner_html(Template::item_list(&items));
+    }
+
+    /// Gets the selector to find a todo item in the DOM
+    fn get_selector_string(id: String) -> String {
```
alexcrichton: In general string arguments to functions are often `&str` instead of `String`

alexcrichton:
Oh I think that an entry in the top-level workspace may also want to be added to get CI passing?

examples/todomvc/src/element.rs
jonathanKingston: Cool this was mostly my lack of knowledge here. Will remove the comment :).

examples/todomvc/Cargo.toml
jonathanKingston: This was from: https://rustwasm.github.io/book/game-of-life/code-size.html#disable-debug-symbols

examples/todomvc/src/scheduler.rs
jonathanKingston: This was also from https://rustwasm.github.io/book/game-of-life/code-size.html#disable-debug-symbols maybe use `process::abort()` within the debug fn?

examples/todomvc/src/element.rs
jonathanKingston: This was on purpose so it would consume `child` such that it isn't usable. I will add a comment that this is on purpose.

examples/todomvc/src/element.rs
jonathanKingston: Mostly because I forgot `as_ref` was possible :laughing: I will simplify where possible :) thanks! (and for all the other reviews).

jonathanKingston: Oh I think mostly because `.into` needs ownership so I got into the habit of needing ownership.

examples/todomvc/src/element.rs
Pauan: Yes `'static` is necessary to avoid undefined behavior (in particular, without `'static` the closure could access stack-local variables, but it would access them after the stack frame is gone, causing use-after-free).

The only way to avoid `'static` is to use a bounded API (similar to the [thread scoping API in crossbeam](https://docs.rs/crossbeam-utils/0.5.0/crossbeam_utils/thread/fn.scope.html)):

```rust
scoped_event_listener(|scope| {
    scope.add_event_listener(node, name, callback);
});
```

As soon as the `|scope| { ... }` closure returns, it will immediately remove all of the event listeners which were created inside of the closure. This guarantees that the callbacks won't be called after the stack frame is gone.

However, such an API is completely and utterly useless: event listeners in the DOM are always asynchronous, meaning that they *always* run after the stack frames are gone.

So the end result of using `scoped_event_listener` is that it would attach a bunch of event listeners, and then immediately remove them (the callbacks will never be called).

So unfortunately we're stuck with `'static` for callbacks, and I don't think that will ever change.

----

`'static` causes a lot of restrictions, but there's two workarounds:

1. If you only need to access the variables inside of one closure, you can just use `move`, which will move the variables from the stack into the closure. This is extremely fast.

   This works just fine with `'static` closures, because the variables aren't on the stack anymore. It even works for mutable variables:

   ```rust
   let mut foo = 0;

   node.add_event_listener("click", move |_| {
       foo += 1;
   });
   ```

2. However, sometimes you need to access the variables in multiple callbacks, or you need to access them both inside and outside the callback.

   In that case you can use `Rc` or `Arc` and then clone them and move the clone into the closure using something like this:

   ```rust
   let foo = Rc::new(...);

   {
       let foo = foo.clone();

       node.add_event_listener("click", move |_| {
           foo.some_method();
       });
   }

   {
       let foo = foo.clone();

       node.add_event_listener("mousemove", move |_| {
          foo.some_other_method();
       });
   }
   ```

   This has a very small performance cost, but it's often necessary.

Of course you can mix and match both techniques:

```rust
let foo = Rc::new(...);

{
    let foo = foo.clone();
    let mut bar = 0;

    node.add_event_listener("click", move |_| {
        bar += 1;
        foo.some_method(bar);
    });
}
```

I see you're already using stuff like that with the `Scheduler` type. I commonly use a pattern like this:

```rust
struct State {
    foo: Cell<bool>,
    bar: RefCell<Vec<String>>,
}

impl State {
    fn new() -> Rc<Self> {
        Rc::new(Self {
            foo: Cell::new(false),
            bar: RefCell::new(vec![]),
        })
    }
}

...

let state = State::new();
```

This ensures that the entire app's state is in a single unified struct (which can use sub-structs if it wishes to), and it only requires a single `Rc`, so it's easy and cheap to pass it around to callbacks.

Internally it uses `Cell` and `RefCell` to mutate individual fields. It could instead use `Rc<RefCell<State>>` to mutate the entire struct, but I'm not a big fan of that.

Both styles are valid, but doing it this way lets you call methods directly on `state` (without needing to `borrow` or `borrow_mut` it first), and it can be a bit faster since it can use `Cell`.

examples/todomvc/Cargo.toml
alexcrichton: Aha! I sent a PR - https://github.com/rustwasm/book/pull/116

examples/todomvc/src/scheduler.rs
alexcrichton: Ah I mean that `dbg` returns after logging, but these conditions seem like they mean the application is in an erroneous state so it should somehow abort, although I'm fine either aborting via `process::abort()` or through `panic!`

jonathanKingston: I have moved to change this to be called `exit` which triggers an exception in the JS and `process::abort` annoyingly it doesn't actually stop the JS or wasm from working though.

examples/todomvc/src/element.rs
jonathanKingston: I added this, not sure if it can be utilised in the current code though.

examples/todomvc/src/template.rs
jonathanKingston: the 'handlebars' crate seems to add serde which would likely bloat the demo a lot and it looks like maud isn't compiling at the moment. Could this be a follow up bug perhaps? Unless you have other suggestions?

examples/todomvc/src/element.rs
jonathanKingston: Yeah I was hoping to make Element require borrowing and ownership, especially given we don't have any JS changing the objects. However adding `#[derive(Clone)]` goes against that tbh.

jonathanKingston:
@Pauan wow yeah sorry, it said it wasn't sending!

I don't see my latest PR either :/

jonathanKingston:
@alexcrichton is there any further changes you need here?

alexcrichton:
Oh sorry I was busy last week and didn't get a chance to get to this! In any case though I think this is definitely in a good place now, and we can continue iterating in-tree if necessary. Thanks so much again @jonathanKingston for writing up all this!

alexcrichton:
It looks like there's a lot of build warnings when building this example? Could the warnings perhaps be fixed?

jonathanKingston:
@alexcrichton on it :)

PR #965, opened 2018-10-14
Removed buffer copy from webgl example

djozis:
Change webgl example to use a buffer wrapping instead of copying.

Please scrutinize as I'm both a Rust and wasm-bindgen newbie, and I couldn't get the example to compile locally on Windows (running "wasm-pack build" I got error: failed to read `\\?\C:\...\github\wasm-bindgen\crates/cli\Cargo.toml`) so I tested for errors by compiling it in a side-project instead against latest published versions.

### DIFF
--- a/examples/webgl/src/lib.rs
+++ b/examples/webgl/src/lib.rs
@@ -47,9 +47,10 @@ pub fn draw() {
 
     let vertices: [f32; 9] = [-0.7, -0.7, 0.0, 0.7, -0.7, 0.0, 0.0, 0.7, 0.0];
     let memory_buffer = wasm_bindgen::memory().dyn_into::<WebAssembly::Memory>().unwrap().buffer();
+    let verticles_location = vertices.as_ptr() as u32 / 4;
     let vert_array = js_sys::Float32Array::new(&memory_buffer).subarray(
-        &vertices as *const f32 as u32 / 4,
-        &vertices as *const f32 as u32 / 4 + vertices.len() as u32,
+        verticles_location,
+        verticles_location + vertices.len() as u32,
     );
 
     let buffer = context.create_buffer().unwrap();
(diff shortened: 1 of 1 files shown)


examples/webgl/src/lib.rs
```diff
@@ -44,17 +45,18 @@ pub fn draw() {
...
 
-    let vertices = [-0.7, -0.7, 0.0, 0.7, -0.7, 0.0, 0.0, 0.7, 0.0];
-    let vert_array = js_sys::Float32Array::new(&wasm_bindgen::JsValue::from(vertices.len() as u32));
-    for (i, f) in vertices.iter().enumerate() {
-        vert_array.fill(*f, i as u32, (i + 1) as u32);
-    }
+    let vertices: [f32; 9] = [-0.7, -0.7, 0.0, 0.7, -0.7, 0.0, 0.0, 0.7, 0.0];
+    let memory_buffer = wasm_bindgen::memory().dyn_into::<WebAssembly::Memory>().unwrap().buffer();
+    let vert_array = js_sys::Float32Array::new(&memory_buffer).subarray(
+        &vertices as *const f32 as u32 / 4,
```
alexcrichton: Could this use `vertices.as_ptr()` instead of raw pointer casts?

jsheard: With the `new Uint8Array(buffer, byteOffset, length)` constructor it could be simplified to

```
let vert_array = js_sys::Uint8Array::new(&memory_buffer, vertices.as_ptr() as u32, vertices.len() * mem::size_of::<f32>());
```

That constructor doesn't seem to be exposed in js_sys currently though.

djozis: Updated to use as_ptr() and tested on my side project.
Turns out the generated wasm is identical, even extracting verticles_location made no difference.

As a novice user, I was confused that I couldn't use the constructor @jsheard mentioned (or the float equivalent), but then I found the subarray work-around.

alexcrichton:
Thanks!

PR #966, opened 2018-10-15
Add overloads with endianness parameter to DataView gets and sets

katis:


alexcrichton:
Looks great to me, thanks so much!

PR #970, opened 2018-10-16
Bind more Typed Array constructors in `js-sys`

alexcrichton:
Bind `new Int8Array(length)`, `new Int8Array(buffer, offset)` and `new
Int8Array(buffer, offset, length)`.


alexcrichton:
We can certainly consider that yeah! I think the documentation may be harder but we could probably figure out a way

PR #972, opened 2018-10-16
Implement support for WebAssembly threads

alexcrichton:
... and add a parallel raytracing demo!

This commit adds enough support to `wasm-bindgen` to produce a workable
wasm binary *today* with the experimental WebAssembly threads support
implemented in Firefox Nightly. I've tried to comment what's going on in
the commits and such, but at a high level the changes made here are:

* A new transformation, living in a new `wasm-bindgen-threads-xform`
  crate, prepares a wasm module for parallel execution. This performs a
  number of mundane tasks which I hope to detail in a blog post later on.
* The `--no-modules` output is enhanced with more support for when
  shared memory is enabled, allowing passing in the module/memory to
  initialize the wasm instance on multiple threads (sharing both module
  and memory).
* The `wasm-bindgen` crate now offers the ability, in `--no-modules`
  mode, to get a handle on the `WebAssembly.Module` instance.
* The example itself requires Xargo to recompile the standard library
  with atomics and an experimental feature enabled. Afterwards it
  experimentally also enables threading support in wasm-bindgen.

I've also added hopefully enough CI support to compile this example in a
builder so we can upload it and poke around live online. I hope to
detail more about the technical details here in a blog post soon as
well!

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -6,7 +6,7 @@ use decode;
 use failure::{Error, ResultExt};
 use parity_wasm::elements::*;
 use shared;
-use wasm_bindgen_gc;
+use gc;
 
 use super::Bindgen;
 use descriptor::{Descriptor, VectorKind};
@@ -388,12 +388,26 @@ impl<'a> Context<'a> {
             ))
         })?;
 
+        self.bind("__wbindgen_module", &|me| {
+            if !me.config.no_modules {
+                bail!("`wasm_bindgen::module` is currently only supported with \
+                       --no-modules");
+            }
+            me.expose_add_heap_object();
+            Ok(format!(
+                "
+                function() {{
+                    return addHeapObject(init.__wbindgen_wasm_module);
+                }}
+                ",
+            ))
+        })?;
+
         self.bind("__wbindgen_rethrow", &|me| {
             me.expose_take_object();
             Ok(String::from("function(idx) { throw takeObject(idx); }"))
         })?;
 
-        self.create_memory_export();
         self.unexport_unused_internal_exports();
         closures::rewrite(self)?;
         self.gc();
@@ -415,7 +429,76 @@ impl<'a> Context<'a> {
 
         self.rewrite_imports(module_name);
 
-        let mut js = if self.config.no_modules {
+        let mut js = if self.config.threads.is_some() {
+            // TODO: It's not clear right now how to best use threads with
+            // bundlers like webpack. We need a way to get the existing
+            // module/memory into web workers for now and we don't quite know
+            // idiomatically how to do that! In the meantime, always require
+            // `--no-modules`
+            if !self.config.no_modules {
+                bail!("most use `--no-modules` with threads for now")
+            }
+            self.memory(); // set `memory_limit` if it's not already set
+            let limits = match &self.memory_init {
+                Some(l) if l.shared() => l.clone(),
+                _ => bail!("must impot a shared memory with threads"),
+            };
+
+            let mut memory = String::from("new WebAssembly.Memory({");
+            memory.push_str(&format!("initial:{}", limits.initial()));
+            if let Some(max) = limits.maximum() {
+                memory.push_str(&format!(",maximum:{}", max));
+            }
+            if limits.shared() {
+                memory.push_str(",shared:true");
+            }
+            memory.push_str("})");
+
+            format!(
+                    "\
+(function() {{
+    var wasm;
+    var memory;
+    const __exports = {{}};
+    {globals}
+    function init(module_or_path, maybe_memory) {{
+        let result;
+        const imports = {{ './{module}': __exports }};
+        if (module_or_path instanceof WebAssembly.Module) {{
+            memory = __exports.memory = maybe_memory;
+            result = WebAssembly.instantiate(module_or_path, imports)
+                .then(instance => {{
+                    return {{ instance, module: module_or_path }}
+                }});
+        }} else {{
+            memory = __exports.memory = {init_memory};
+            const response = fetch(module_or_path);
+            if (typeof WebAssembly.instantiateStreaming === 'function') {{
+                result = WebAssembly.instantiateStreaming(response, imports);
+            }} else {{
+                result = response
+                    .then(r => r.arrayBuffer())
+                    .then(bytes => WebAssembly.instantiate(bytes, imports));
+            }}
+        }}
+        return result.then(({{instance, module}}) => {{
+            wasm = init.wasm = instance.exports;
+            init.__wbindgen_wasm_instance = instance;
+            init.__wbindgen_wasm_module = module;
+            init.__wbindgen_wasm_memory = __exports.memory;
+        }});
+    }};
+    self.{global_name} = Object.assign(init, __exports);
+}})();",
+                    globals = self.globals,
+                    module = module_name,
+                    global_name = self.config.no_modules_global
+                        .as_ref()
+                        .map(|s| &**s)
+                        .unwrap_or("wasm_bindgen"),
+                    init_memory = memory,
+            )
+        } else if self.config.no_modules {
             format!(
                     "\
 (function() {{
@@ -635,20 +718,6 @@ impl<'a> Context<'a> {
             exports.entries_mut().push(entry);
             break;
         }
-    }
-
-    fn create_memory_export(&mut self) {
-        let limits = match self.memory_init.clone() {
-            Some(limits) => limits,
-            None => return,
-        };
-        let mut initializer = String::from("new WebAssembly.Memory({");
-        initializer.push_str(&format!("initial:{}", limits.initial()));
-        if let Some(max) = limits.maximum() {
-            initializer.push_str(&format!(",maximum:{}", max));
-        }
-        initializer.push_str("})");
-        self.export("memory", &initializer, None);
     }
 
     fn rewrite_imports(&mut self, module_name: &str) {
@@ -1621,7 +1690,7 @@ impl<'a> Context<'a> {
 
     fn gc(&mut self) {
         self.parse_wasm_names();
-        wasm_bindgen_gc::Config::new()
+        gc::Config::new()
             .demangle(self.config.demangle)
             .keep_debug(self.config.keep_debug || self.config.debug)
             .run(&mut self.module);
@@ -1671,7 +1740,6 @@ impl<'a> Context<'a> {
                 _ => None,
             }).next()
             .expect("must import memory");
-        assert_eq!(entry.module(), "env");
         assert_eq!(entry.field(), "memory");
         self.memory_init = Some(mem.limits().clone());
         "memory"
--- a/crates/threads-xform/src/lib.rs
+++ b/crates/threads-xform/src/lib.rs
@@ -1,0 +1,672 @@
+#[macro_use]
+extern crate failure;
+extern crate parity_wasm;
+
+use std::collections::HashMap;
+
+use failure::{Error, ResultExt};
+use parity_wasm::elements::*;
+
+const PAGE_SIZE: u32 = 1 << 16;
+
+/// Configuration for the transformation pass in this module.
+///
+/// Created primarily through `new` and then executed through `run`.
+pub struct Config {
+    maximum_memory: u32,
+    thread_stack_size: u32,
+}
+
+impl Config {
+    /// Create a new configuration with default settings.
+    pub fn new() -> Config {
+        Config {
+            maximum_memory: 1 << 30, // 1GB
+            thread_stack_size: 1 << 20, // 1MB
+        }
+    }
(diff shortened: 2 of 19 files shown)


examples/raytrace-parallel/index.html
```diff
@@ -0,0 +1,245 @@
...
+      function run() {
+        if (typeof WebAssembly === 'undefined') {
+          alert('WebAssembly is not available!');
+          return
+        }
+        if (typeof SharedArrayBuffer === 'undefined') {
+          alert('SharedArrayBuffer is not available, is the preference not enabled?');
+          return
+        }
+        delete WebAssembly.instantiateStreaming;
```
xtuc: Why do you need this?

examples/raytrace-parallel/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+// First up, load our wasm
+wasm_bindgen('./raytrace_parallel_bg.wasm')
+  .then(run)
+  .catch(e => {
+    if (e instanceof WebAssembly.CompileError) {
+      let msg = 'failed to compile wasm module';
+      msg += `\nerror: ${e.message}`;
+      msg += "\ncurrently this requires nightly Firefox";
+      alert(msg);
```
xtuc: This already came up in the WebAssembly group, we should have a way to feature test. You could have an unrelated compilation error here.

examples/raytrace-parallel/index.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+  });
+
+const { Scene, WorkerPool } = wasm_bindgen;
+
+function run() {
+  // The maximal concurrency of our web worker pool is `hardwareConcurrency`,
+  // so set that up here and this ideally is the only location we create web
+  // workers.
+  pool = new WorkerPool(navigator.hardwareConcurrency);
```
xtuc: Maybe it's just me but I took me some time to get that this is exported from wasm, I think the destructuring shouldn't be outside the function and not at all.

examples/raytrace-parallel/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+        }) as Box<FnMut(Event)>);
+        let mut workers = Vec::new();
+        for _ in 0..max {
+            // TODO: what do do about `./worker.js`:
+            //
+            // * the path is only known by the bundler. How can we, as a
+            //   library, know what's going on?
+            // * How do we not fetch a script N times? It internally then
+            //   causes another script to get fetched N times...
+            let worker = Worker::new("./worker.js")?;
```
xtuc: This is an interesting issue. If you were in a JS file, the bundler would detect the import and rewrite to its own architecture.

Here it's just a random string, no module system will be able to detect it and importing it doesn't make any sense since we want the URL.

In the web it's easy to move/load the file at the root `/` but that won't work in Node.

it would make sense to me to have the bundler injects some runtime functions for plateform specific stuff, like generating a path/url. Unfortunately, type representation are language specific, how do we pass a string? 

To sum up: :man_shrugging:

guide/src/SUMMARY.md
```diff
@@ -26,6 +26,7 @@
   - [web-sys: WebAudio](./examples/web-audio.md)
   - [web-sys: WebGL](./examples/webgl.md)
   - [web-sys: A Simple Paint Program](./examples/paint.md)
+  - [Parallel Raytracing](./examples/raytrace.md)
```
xtuc: Well, synchronisation is involved, it's more concurrent than parallel, but I don't really mind.

xtuc:
Great work @alexcrichton!

crates/cli-support/src/js/mod.rs
```diff
@@ -415,7 +429,75 @@ impl<'a> Context<'a> {
 
         self.rewrite_imports(module_name);
 
-        let mut js = if self.config.no_modules {
+        let mut js = if self.config.threads.is_some() {
+            if !self.config.no_modules {
+                // TODO: unfortunately there's bugs in bundlers like webpack, so
+                // this is only known to work with `--no-modules` for now. We
+                // should fix this with other integrations in the future! (node
+                // too!)
```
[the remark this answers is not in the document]
alexcrichton: Oh this is probably badly worded, it's definitely not active bugs moreso "this is a new world where it's totally unclear how this will all look". In that sense you're right, yeah, accessing `WebAssembly.Module` instances is one issue but another is just shipping the memory/module to a different thread and letting webpack start on the thread

examples/raytrace-parallel/index.html
alexcrichton: Oh this is [copied from here](https://github.com/rustwasm/wasm-bindgen/blob/ab8be88e7b5db590e17051317d742deb877631fb/examples/no_modules/index.html#L7-L15), tl;dr; the python built-in webservers don't know about the wasm content tyep

examples/raytrace-parallel/index.js
alexcrichton: It's true! I believe the feature testing method is intended to be done with tiny wasm modules and something like `new WebAssembly.Instance(buffer)` where `buffer` is like 20 bytes long and does precisely just one operation. I was largely just lazy here :)

crates/cli-support/src/js/mod.rs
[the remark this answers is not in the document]
xtuc: Yeah we need to think about that more.

examples/raytrace-parallel/src/lib.rs
xtuc: Maybe Host-bindings would help

crates/threads-xform/src/lib.rs
```diff
@@ -0,0 +1,675 @@
...
+                    Type::Function(t) => t,
+                }
+            })
+            .position(|t| t.params().is_empty() && t.return_type().is_none());
+        match pos {
+            Some(i) => i as u32,
+            None => {
+                let f = FunctionType::new(Vec::new(), None);
+                section.types_mut().push(Type::Function(f));
+                (section.types().len() - 1) as u32
```
fitzgen:
```suggestion
                (section.types().len() - 1) as u32 // TODO: checked cast
```

crates/threads-xform/src/lib.rs
```diff
@@ -0,0 +1,675 @@
...
+        }
+    };
+    module.function_section_mut()
+        .expect("module has no function section")
+        .entries_mut()
+        .push(Func::new(type_idx));
+
+    // ... and finally flag it as the new start function
+    let idx = code_idx +
+        (module.import_section().map(|m| m.functions()).unwrap_or(0) as u32);
```
fitzgen:
```suggestion
        module.import_count(elements::ImportCountType::Function)
```

crates/threads-xform/src/lib.rs
```diff
@@ -0,0 +1,675 @@
...
+        match &mut module.sections_mut()[i] {
+            Section::Type(_) |
+            Section::Import(_) |
+            Section::Function(_) |
+            Section::Table(_) |
+            Section::Memory(_) |
+            Section::Global(_) |
+            Section::Export(_) => continue,
+            Section::Start(i) => {
+                *i = start;
```
fitzgen: Suggest using a different index variable here (`start_idx`?) because `i` is used above, so this is easy to mis-read.

crates/threads-xform/src/lib.rs
```diff
@@ -0,0 +1,675 @@
...
+                External::Function(i) => *i,
+                _ => unreachable!(),
+            };
+            let types = module.type_section().unwrap();
+            let fty = match &types.types()[type_idx as usize] {
+                Type::Function(f) => f,
+            };
+            // Validate the type for this intrinsic
+            match entry.field() {
+                "__wbindgen_thread_id" => {
```
fitzgen: Ah cool so these are all implemented with "intrinsics", same as our others. Nice!

examples/raytrace-parallel/src/lib.rs
fitzgen: In this particular case, we could do this:

```rust
let workerjs = include_str!("worker.js");
late workerjs_b64 = web_sys::window().unwrap().btoa(workerjs)?;
let worker = Worker::new(&format!("data:text/javascript;base64,{}", workerjs_b64))?;
```

guide/src/SUMMARY.md
fitzgen: This is nittiest of nitpicking but:

As long as there are multiple workers that are running on different cores, then this *is* parallelism.

Just because there are synchronization points doesn't make something not parallel.

concurrent = operations in-progress (but maybe not doing work) at the same time

parallel = doing work at the same time

For example, threading is always concurrent, but it might not be parallel if there is only a single core that is only actually working on one thread at any given moment and is time slicing between all of the threads.

src/lib.rs
```diff
@@ -633,6 +634,18 @@ pub fn throw_val(s: JsValue) -> ! {
     }
 }
 
+/// Returns a handle to this wasm instance's `WebAssembly.Module`
+///
+/// Note that this is only available when the final wasm app is built with
+/// `--no-modules`, it's not recommended to rely on this API yet! This is
+/// largely just an experimental addition to enable threading demos. Using this
+/// may prevent your wasm module from building down the road.
+pub fn module() -> JsValue {
```
fitzgen: Want to `doc(hidden)` this to emphasize the experimental, unstable nature?

examples/raytrace-parallel/index.js
fitzgen: This is nittiest of nitpicking but:

As long as there are multiple workers that are running on different cores, then this is parallelism.

Just because there are synchronization points doesn't make something not parallel.

concurrent = operations in-progress (but maybe not doing work) at the same time

parallel = doing work at the same time

For example, threading is always concurrent, but it might not be parallel if there is only a single core that is only actually working on one thread at any given moment and is time slicing between all of the threads.

examples/raytrace-parallel/src/lib.rs
xtuc: Sorry @fitzgen I didn't ping you, the discussion is here https://github.com/webpack/webpack/issues/8235#issue-371444670.

I think that exposing the bundler's API is much better than inlining a big JS there.

crates/threads-xform/src/lib.rs
```diff
@@ -0,0 +1,675 @@
...
+    }
+
+    Ok(ret)
+}
+
+fn import_memory_zero(module: &mut Module)
+    -> Result<(), Error>
+{
+    // If memory is exported, let's switch it to imported. If memory isn't
+    // exported then there's nothing to do.
```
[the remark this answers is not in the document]
alexcrichton: Correct! This is cleaned up a bit later by the later phases, although this is a bit confusingly worded...

crates/threads-xform/src/lib.rs
alexcrichton: Eh I've opted to just leave out the TODO for now, the chance of this actually getting hit is basically zero

alexcrichton:
Ok I've addressed the points here and we should hopefully be good to go now! Gonna merge this and we can iterate online as well.

DanielRamosAcosta:
Hey guys, [the example of the docs](https://rustwasm.github.io/wasm-bindgen/exbuild/raytrace-parallel/) isn't working for me, I'm in Firefox Beta 64.0b14 (64-bit), and I have enabled the `javascript.options.shared_memory` option. Here is the error I'm receiving:

```
CompileError: "wasm validation error: at offset 65375: unrecognized opcode: fe 19"
	result https://rustwasm.github.io/wasm-bindgen/exbuild/raytrace-parallel/raytrace_parallel.js:619:24
```

alexcrichton:
Ah the encoding has been tweaked a few times (it's all unstable after all!), but if you try nightly the demo should work!

DanielRamosAcosta:
It still gives me the same error... 😕

alexcrichton:
@DanielRamosAcosta hm odd! I'm on "65.0a1 (2018-11-30) (64-bit)" and the [online example](https://rustwasm.github.io/wasm-bindgen/exbuild/raytrace-parallel/) works for me. What nightly are you using?

DanielRamosAcosta:
Oh, I still in 64.0b14, let me try updating.

DanielRamosAcosta:
image
Strange, Firefox says me that I'm updated, where did you get the 65?

alexcrichton:
Hm I'm not sure! If you use the ["Nightly" channel from here](https://www.mozilla.org/en-US/firefox/channel/desktop/) does it still give you 64?

whaqzhzd:
hi, I opened the flag in the new version of Chrome. But I still can't run. It's unknown for what reason。https://developers.google.com/web/updates/2018/10/wasm-threads

alexcrichton:
@whaqzhzd unfortunately I think the Chrome/Firefox implementations differ, but as they become more stable and match we'll be sure to track it!

PR #974, opened 2018-10-18
Shrink binary size of distributed `wasm-bindgen`

alexcrichton:
This shaves a little over 2MB off the download locally for Linux,
removing debuginfo (which no one's probably gonna use anyway) as well as
switching from jemalloc to the system allocator.


sendilkumarn:
We can even consider to optimize it with `opt-level = z` right to shave off some extra bits 👍

alexcrichton:
Perhaps yeah! I was just thinking we could get some easy wins on Linux which was disproportionately larger than other platforms, but we've got more options available to us to shrink further on other platforms

PR #987, opened 2018-10-28
Add gc support for locals in functions

alexcrichton:
This commit adds support for running a gc pass over locals in a
function. This will remove dead local declarations for a function
(completely unused) as well as compact existing entries to ensure that
we don't have two local declarations of the same type.

While this was initially intended for some future support of emitting
shims in `wasm-bindgen`, it turns out this pass is firing quite a lot
over existing functions generated by LLVM. Looks like we may see benefit
from this today with slightly smaller wasm binaries!

### DIFF
--- a/crates/gc/src/lib.rs
+++ b/crates/gc/src/lib.rs
@@ -8,8 +8,9 @@ extern crate parity_wasm;
 extern crate log;
 extern crate rustc_demangle;
 
-use std::mem;
 use std::collections::HashSet;
+use std::iter;
+use std::mem;
 
 use parity_wasm::elements::*;
 
@@ -132,6 +133,8 @@ fn run(config: &mut Config, module: &mut Module) {
             module.sections_mut().remove(i);
         }
     }
+
+    gc_locals(module);
 }
 
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


crates/gc/src/lib.rs
```diff
@@ -794,3 +796,128 @@ impl<'a> RemapContext<'a> {
...
+    idx: u32,
+    body: &mut FuncBody,
+    ty: &FunctionType,
+    names: &mut Option<&mut LocalNameSection>,
+) {
+    let mut local_tys = ty.params().to_vec();
+    for local in body.locals_mut().drain(..) {
+        for _ in 0..local.count() {
+            local_tys.push(local.value_type());
+        }
```
fitzgen:
```rust
local_tys.extend(iter::repeat(local.value_type()).take(local.count()));
```

crates/gc/src/lib.rs
```diff
@@ -794,3 +796,128 @@ impl<'a> RemapContext<'a> {
...
+    for instr in body.code().elements() {
+        match instr {
+            Instruction::GetLocal(i) => used[*i as usize] = true,
+            Instruction::SetLocal(i) => used[*i as usize] = true,
+            Instruction::TeeLocal(i) => used[*i as usize] = true,
+            _ => {}
+        }
+    }
+
+    let mut map = vec![u32::max_value(); local_tys.len()];
```
fitzgen: Why not just use `Option<u32>` and `unwrap` down below instead of asserting against `u32::max_value()`?

alexcrichton: 👍

Issue #994, opened 2018-10-29
Regession: 0.2.26 fail to build hello world example


la10736:
Follow hello world wasm example at https://rustwasm.github.io/book/game-of-life/hello-world.html fail at

```
> wasm-pack build

...
  [7/9] Copying over your README...
  [8/9] wasm-bindgen already installed...
| [9/9] Running WASM-bindgen...
wasm-bindgen failed to execute properly. stderr:

thread 'main' panicked at 'assertion failed: *i != u32::max_value()', crates/gc/src/lib.rs:805:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Change in `Cargo.toml` the `wasm-bindgen` dependency in
```
 wasm-bindgen = "=0.2.25"
```
fix the issue.

alexcrichton:
Egad sorry about this and thanks for the quick report! This should be fixed in https://github.com/rustwasm/wasm-bindgen/pull/995 and I'll publish a new version of wasm-bindgen once it's merged.

PR #996, opened 2018-10-29
Add tests for the wasm-gc crate 

alexcrichton:
This commit adds a test harness and the beginnings of a test suite for
the crate that performs GC over a wasm module. This crate historically
has had zero tests because it was thought that it would no longer be
used once LLD landed with `--gc-sections`, but `wasm-bindgen` has come
to rely more and more on `wasm-gc` for various purposes.

The last release of `wasm-bindgen` was also released with a bug in the
recently refactored support in the `wasm-gc` crate, providing a perfect
time and motivation to start writing some tests!

All tests added here are `*.wat` files which contain the expected output
after the gc pass is executed. Tests are automatically updated with
`BLESS_TESTS=1` in the environment, which is the expected way to
generate the output for each test.

### DIFF
--- a/crates/gc/tests/all.rs
+++ b/crates/gc/tests/all.rs
@@ -1,0 +1,146 @@
+extern crate parity_wasm;
+extern crate rayon;
+extern crate tempfile;
+extern crate wasm_bindgen_gc;
+
+use std::env;
+use std::error::Error;
+use std::fs;
+use std::io;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
+
+use rayon::prelude::*;
+use parity_wasm::elements::Module;
+use tempfile::NamedTempFile;
+
+struct Test {
+    input: PathBuf,
+}
+
+fn main() {
+    let mut tests = Vec::new();
+    find_tests(&mut tests, "tests/wat".as_ref());
+
+    run_tests(&tests);
+}
+
+fn find_tests(tests: &mut Vec<Test>, path: &Path) {
+    for entry in path.read_dir().unwrap() {
+        let entry = entry.unwrap();
+        let path = entry.path();
+        if entry.file_type().unwrap().is_dir() {
+            find_tests(tests, &path);
+            continue
+        }
+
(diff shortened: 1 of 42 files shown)


alexcrichton:
This also did indeed catch https://github.com/rustwasm/wasm-bindgen/issues/994 while writing tests, and the bug there was that we'd gc a table away but try to preserve the element section which didn't work because the table was gc'd away. This now properly filters out the element section with dead elements and then only remaps the indices of the remaining live elements.

crates/gc/tests/all.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+fn main() {
+    let mut tests = Vec::new();
+    find_tests(&mut tests, "tests/wat".as_ref());
+
+    run_tests(&tests);
+}
+
+fn find_tests(tests: &mut Vec<Test>, path: &Path) {
```
fitzgen: Rather than implementing our own test harness, I think it is usually easier/better to have a build.rs generate normal `#[test]`s and let cargo be the harness. Then we wouldn't have to fiddle with rayon to get proper parallel test running, and we wouldn't need to create our own `Test` type and keep track of results ourselves, etc etc

The downside is that we have to create another test-only crate to avoid making a build.rs that isn't needed for normal builds

alexcrichton: Yeah I opted to not grow another crate because we already have so many, and I figured this logic wasn't really all that complicated and will probably mirror 90% of what will be in the build script anyway

alexcrichton: @fitzgen would you be ok leaving this as-is? I can remove some of the fanciness like rayon to simplify this to just executing tests which may make it much more amenable

fitzgen: If we aren't going to generate `#[test]`s in `build.rs`, then we should use rayon.

Also, I should have mentioned this in the original review, but these tests are awesome, and it is great to see this PR! ^_^

alexcrichton: Ok! FWIW rayon lived up to its promise and was a 2 line change to parallelize, so it's easy enough to back out too!

PR #997, opened 2018-10-30
Remove compiler warnings in todo example

jonathanKingston:
- Simplify usage of elements into element.rs
- Fix a bug with focus not working on edit.

I'm currently ignoring Results from the DOM as for most cases I don't really think there is much recovery, perhaps they should behave like a panic?

### DIFF
--- a/examples/todomvc/src/element.rs
+++ b/examples/todomvc/src/element.rs
@@ -1,10 +1,24 @@
 extern crate wasm_bindgen;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
+use web_sys::EventTarget;
 
 /// Wrapper for `web_sys::Element` to simplify calling different interfaces
 pub struct Element {
     el: Option<web_sys::Element>,
+}
+
+impl From<web_sys::Element> for Element {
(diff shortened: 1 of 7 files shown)


examples/todomvc/src/element.rs
```diff
@@ -41,8 +68,8 @@ impl Element {
     {
         let cb = Closure::wrap(Box::new(handler) as Box<FnMut(_)>);
         if let Some(el) = self.el.take() {
-            let el_et: web_sys::EventTarget = el.into();
-            el_et.add_event_listener_with_callback(event_name, cb.as_ref().unchecked_ref());
+            let el_et: EventTarget = el.into();
+            let _r = el_et.add_event_listener_with_callback(event_name, cb.as_ref().unchecked_ref());
```
alexcrichton: Shouldn't results like these be propagated out as an exception?

jonathanKingston:
@alexcrichton sorry for the delay here. I used console_error_panic_hook for the errors however I think it would be nicer to be exceptions though (perhaps that should be a feature flag for this crate?).

alexcrichton:
Ok no worries! We can always continue to iterate in-tree

PR #1002, opened 2018-10-31
Add experimental support for the `anyref` type

alexcrichton:
This commit adds experimental support to `wasm-bindgen` to emit and
leverage the `anyref` native wasm type. This native type is still in a
proposal status (the reference-types proposal). The intention of
`anyref` is to be able to directly hold JS values in wasm and pass the
to imported functions, namely to empower eventual host bindings (now
renamed WebIDL bindings) integration where we can skip JS shims
altogether for many imports.

This commit doesn't actually affect wasm-bindgen's behavior at all
as-is, but rather this support requires an opt-in env var to be
configured. Once the support is stable in browsers it's intended that
this will add a CLI switch for turning on this support, eventually
defaulting it to `true` in the far future.

The basic strategy here is to take the `stack` and `slab` globals in the
generated JS glue and move them into wasm using a table. This new table
in wasm is managed at the fringes via injected shims. At
`wasm-bindgen`-time the CLI will rewrite exports and imports with shims
that actually use `anyref` if needed, performing loads/stores inside the
wasm module instead of externally in the wasm module.

This should provide a boost over what we have today, but it's not a
fantastic strategy long term. We have a more grand vision for `anyref`
being a first-class type in the language, but that's on a much longer
horizon and this is currently thought to be the best we can do in terms
of integration in the near future.

The stack/heap JS tables are combined into one wasm table. The stack
starts at the end of the table and grows down with a stack pointer (also
injected). The heap starts at the end and grows up (state managed in
linear memory). The anyref transformation here will hook up various
intrinsics in wasm-bindgen to the runtime functionality if the anyref
supoprt is enabled.

The main tricky treatment here was applied to closures, where we need JS
to use a different function pointer than the one Rust gives it to use a
JS function pointer empowered with anyref. This works by switching up a
bit how descriptors work, embedding the shims to call inside descriptors
rather than communicated at runtime. This means that we're accessing
constant values in the generated JS and we can just update the constant
value accessed.

### DIFF
--- a/crates/anyref-xform/src/lib.rs
+++ b/crates/anyref-xform/src/lib.rs
@@ -1,0 +1,730 @@
+//! Transformation for wasm-bindgen to enable usage of `anyref` in a wasm
+//! module.
+//!
+//! This crate is in charge of enabling code using `wasm-bindgen` to use the
+//! `anyref` type inside of the wasm module. This transformation pass primarily
+//! wraps exports and imports in shims which use `anyref`, but quickly turn them
+//! into `i32` value types. This is all largely a stopgap until Rust has
+//! first-class support for the `anyref` type, but that's thought to be in the
+//! far future and will take quite some time to implement. In the meantime, we
+//! have this!
+//!
+//! The pass here works by collecting information during binding generation
+//! about imports and exports. Afterwards this pass runs in one go against a
+//! wasm module, updating exports, imports, calls to these functions, etc. The
+//! goal at least is to have valid wasm modules coming in that don't use
+//! `anyref` and valid wasm modules going out which use `anyref` at the fringes.
+
+use failure::{bail, format_err, Error};
+use std::cmp;
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::mem;
+use walrus::ir::*;
+use walrus::{FunctionId, GlobalId, InitExpr, Module, TableId, ValType};
+
+// must be kept in sync with src/lib.rs and ANYREF_HEAP_START
+const DEFAULT_MIN: u32 = 32;
+
+/// State of the anyref pass, used to collect information while bindings are
+/// generated and used eventually to actually execute the entire pass.
+#[derive(Default)]
+pub struct Context {
+    // Functions within the module that we're gonna be wrapping, organized by
+    // type. The `Function` contains information about what arguments/return
+    // values in the function signature should turn into anyref.
+    imports: HashMap<String, HashMap<String, Function>>,
+    exports: HashMap<String, Function>,
+    elements: BTreeMap<u32, (u32, Function)>,
+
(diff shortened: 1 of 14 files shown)


alexcrichton:
This is intended to get some early review and not actually merge at this time. This is all still entirely untested other than hand-inspecting wat output, we're still waiting on support in an engine to play with this!

alexcrichton:
Also if there are thoughts about landing parts of this first, I'd totally be down for that. Some separable pieces that could be landed first are:

* Refactoring `js/mod.rs` to use `self.get_object(..)` and friends everywhere
* Moving invoke functions from runtime to descriptors for closures (making them constants in JS bindings)

alexcrichton:
Ok this is getting to a good spot! I've split out a bunch of PRs from this which aren't directly related to this change, so only the last commit should be the most relevant for this PR itself. 

I'm able to get a number of toy "hello world" style programs working in Firefox with this PR. Small DOM interactions and such work and I can indeed start to see the glacier that is our glue starting to melt! (aka glue is being removed). 

Further work I believe will be blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1505768. I'd still prefer to not land this until there's a fully tested proof-of-concept (aka a larger test suite passing against an implementation). Firefox is the only implementor of `anyref` as of this red-hot-second (AFAIK), and that bug means that we can't actually store all JS values in anyref tables (`undefined` flat out doesn't work, `true` gets converted to a `Boolean`, etc). Once that bug is fixed we'll be able to keep testing on this end and make sure all this support is ship shape!

In the meantime having the current implementation in Firefox (even if not complete) was mega helpful. I've found tons of little issues here and there already!

alexcrichton:
Ok I've rebased this now! This now is built all on top of `walrus` and I've tested to have everything work locally as well (even caught a bug in the old implementation).

@fitzgen mind doing a once-over of this? I think that reference types are getting final enough that we should be good to go with an implementation on our end.

crates/anyref-xform/src/lib.rs
```diff
@@ -0,0 +1,730 @@
...
+        let tee = builder.local_tee(local, alloc);
+        let get_arg = builder.local_get(arg);
+        let get_table = builder.table_get(table, get_arg);
+        let set_table = builder.table_set(table, tee, get_table);
+        let get_local = builder.local_get(local);
+
+        let ty = module.types.add(&[ValType::I32], &[ValType::I32]);
+        let clone_ref = builder.finish(ty, vec![arg], vec![set_table, get_local], module);
+        let name = "__wbindgen_object_clone_ref".to_string();
+        module.funcs.get_mut(clone_ref).name = Some(name);
```
fitzgen: We should add a method for names to the builder. I also vaguely wanted this before.

Filed https://github.com/rustwasm/walrus/issues/61

Issue #982, opened 2018-10-27
Support for Option in .d.ts


henriiik:
Currently, if you return an `Option` from a function in rust, the generated .d.ts file does not reflect that the function can return `undefined`. I am willing to help out if this is a desired change. 🙂

For example if you have the following rust,

```rust
extern crate wasm_bindgen;
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn some_number() -> usize {
    1
}

#[wasm_bindgen]
pub fn some_optional_number() -> Option<usize> {
    None
}
```

you get the following d.ts, 

```typescript
/* tslint:disable */
export function some_number(): number;

export function some_optional_number(): number;
```

But the actual return value has the type `number | undefined`. if you run typescript in strict mode, it will force you to check for undefined, just like rust. 

Or you could go even further and define a new type in typescript, ex:

```typescript
type Option<T> = T | undefined;

export function some_optional_number(): Option<number>;
```

Would you be willing to merge a change like this if I send a PR? 🙂

alexcrichton:
Thanks for the report! This is related to https://github.com/rustwasm/wasm-bindgen/issues/922 where we unfortunately don't test typescript much on CI, but I think this is still imminently fixable and it can surely be fixed through a PR!

I think the `Option<T>` idiom may not be necessary though? It looks like typescript supports directly returning a type that looks like `number | undefined`, so we can probably inline that definition at each function call.

henriiik:
Sounds good! I'll send a PR, probably wont be able to get to it until the weekend though. :)

henriiik:
this is fixed. :)

PR #1008, opened 2018-11-04
update typescript generation to reflect that Option<T> can be undefined

henriiik:
This patch solves issue #982


henriiik:
Thanks! 😀

PR #1019, opened 2018-11-08
Implement rustwasm/rfcs#5, implement `Deref` for imports and `structural` by default

### LINKED ISSUE #1018: Implement RFC 5: structural by default and Deref inheritance
* https://github.com/rustwasm/rfcs/pull/5
* https://rustwasm.github.io/rfcs/005-structural-and-deref.html

alexcrichton:
This commit implements rustwasm/rfcs#5, namely implementing two large changes:

* The `Deref` trait is now implemented unconditionally for all imported JS types
* The `structural` attribute is now the default

Both of those changes have lots of discussion in the RFC, and to support those changes this PR implements a few other changes as well:

* The generation of JS glue for `structural` bindings has been optimized to remove almost all of the shims that were inserted. The JS glue should be much more optimal and perform better (as measured in rustwasm/rfcs#5)
* Fixup a few issues in the test suite discovered in doing the switch

Closes #1018


alexcrichton:
A lot more information for this PR can also be found in the commit messages!

crates/web-sys/tests/wasm/main.rs
```diff
@@ -56,3 +57,11 @@ pub mod table_element;
 pub mod title_element;
 pub mod xpath_result;
 pub mod indexeddb;
+
+#[wasm_bindgen_test]
+fn deref_works() {
+    let x = JsValue::from(3);
```
fitzgen: Maybe do 

```rust
let x = js_sys::Object::new();
let y: &JsValue = &x;
```

so that we aren't doing these unchecked casts that would fail if they were checked?

examples/paint/src/lib.rs
```diff
@@ -8,29 +8,19 @@ use wasm_bindgen::prelude::*;
...
+        .create_element("canvas")?
+        .dyn_into::<web_sys::HtmlCanvasElement>()?;
+    document.body().unwrap().append_child(&canvas)?;
     canvas.set_width(640);
     canvas.set_height(480);
-    (canvas.as_ref() as &web_sys::HtmlElement)
-        .style()
-        .set_property("border", "solid")
-        .unwrap();
+    canvas.style().set_property("border", "solid")?;
```
fitzgen: Oh man this is sooo much nicer

guide/src/web-sys/inheritance.md
```diff
@@ -0,0 +1,55 @@
+# Inheritance in `web-sys`
+
+Inheritance between JS classes is the bread and butter of how the DOM works on
```
fitzgen: Thanks for writing this up!

tests/wasm/host_binding.rs
```diff
@@ -0,0 +1,40 @@
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_test::*;
```
fitzgen: I guess we don't really have any compile fail tests, but it would be nice to test what happens when both `host_binding` and `structural` are used at the same time. I'd expect that to fail to compile.

I guess we *do* have the UI tests, and this kind of test might fit in there.

alexcrichton:
Tests added and updated!

Pauan:
I don't like the name `host_binding`, because it doesn't really have anything to do with host bindings (`structural` and `final` will both be using host bindings).

But since it's such a niche and rarely used feature, I don't have a particularly strong opinion, so I'm not going to block on it.

guide/src/reference/attributes/on-js-imports/host_binding.md
```diff
@@ -0,0 +1,139 @@
+# `host_binding`
+
+The `host_binding` attribute is the converse of the [`structural`
+attribute](structural.html). It configures how `wasm-bindgen` will generate JS
+glue to call the imported function. The naming here is intended convey that this
+attribute is intended to implement the semantics of the future [host bindings
```
Pauan: I think this is misleading, since both `structural` and `host_binding` can (and will) use the host bindings proposal.

guide/src/reference/attributes/on-js-imports/host_binding.md
```diff
@@ -0,0 +1,139 @@
...
+proposal][host-bindings] for WebAssembly.
+
+[host-bindings]: https://github.com/WebAssembly/host-bindings
+[reference-types]: https://github.com/WebAssembly/reference-types
+
+The `host_binding` attribute is intended to be purely related to performance. It
+ideally has no user-visible effect, and well-typed `structural` imports (the
```
Pauan: I think it would be good to specify what "well-typed" means. It actually has little to do with the types, and has a lot more to do with overriding methods.

guide/src/reference/attributes/on-js-imports/host_binding.md
```diff
@@ -0,0 +1,139 @@
...
+default) should be able to transparently switch to `host_binding` eventually.
+
+The eventual performance aspect is that with the [host bindings
```
Pauan: I think this is also misleading, since `structural` is generally faster, and the amount of JS shims is the same (see my comment below comparing `structural` and `host_binding`).

The only potential performance difference is whether it needs to lookup the prototype chain once (with `host_binding`) or multiple times (with `structural`).

guide/src/reference/attributes/on-js-imports/host_binding.md
```diff
@@ -0,0 +1,139 @@
...
+first argument here is actually the `this`.
+
+With that in mind we can further transform this to:
+
+```js
+export const __wbg_bar_a81456386e6b526f = Foo.prototype.bar;
+```
+
+and voila! We, with [reference types][reference-types] and [host
+bindings][host-bindings], now have no JS shim at all necessary to call the
```
Pauan: I would still call that a JS shim. It is JS glue code which is needed to convert a JS value (`Foo.prototype.bar`) into something which is capable of being imported by wasm, since wasm cannot import it directly.

Also, I think it should mention what the `structural` version would look like:

```js
export const __wbg_bar_a81456386e6b526f = function (varg1) { return this.bar(varg1); };
```

Quite short and simple, and also just as performant.

alexcrichton:
It's a good point that `structural` imports will also use host bindings for argument conversions. At this point I'm just going to rename it back to `final`. The only real difference between the two is whether it's a "final import" that can never change or something that's looked up at runtime on every call via the prototype chain (which can also change at runtime).

guide/src/reference/attributes/on-js-imports/host_binding.md
alexcrichton: I think this is appropriately worded as-is because it explicitly says the "eventual performance", not commenting on today's performance. This is quickly becoming an incredibly niche feature that basically no one will use, so I don't think it's necessary to have absolutely exhaustive documentation for such a feature which goes into 100% of the nuances of why one might or might not want to use the attribute

guide/src/reference/attributes/on-js-imports/host_binding.md
alexcrichton: I've switched to phrasing this as "JS function shim", I think being pedantic here about what's a JS shim and what isn't is detracting from the point of the documentatino. 

And again to be clear, this isn't trying to describe the performance today, it's describing the eventual performance where if you have *any JS shim at all* then wasm engines will be unable to route calls to C++ dom functions directly to C++. The whole point of remove the JS shim is to ensure that DOM calls can go straight to C++ with as little fuss as possible.

Pauan: Oh! I see! I finally understand our confusion.

You've been thinking about this as exclusively about JS built-ins (like `Node.prototype.appendChild`), whereas I've been thinking about it as a general purpose mechanism (which can be used on anything, including JS methods).

From the perspective of built-ins, you're completely right that adding in any sort of indirection (i.e. anything other than `export const bar = Foo.prototype.bar;`) is bad for performance, since it can't directly call the built-in C++ function (at least not without a bit of VM magic).

But on the other hand, from the perspective of regular JS classes, you won't gain any performance by doing `export const bar = Foo.prototype.bar;`, since it still has to call into a JS method.

So from a general standpoint, it is indeed a JS shim, but from the built-ins standpoint it isn't a JS shim. Hence our disagreement and confusion.

I think we could use a better term for it ("direct" vs "indirect" JS call, maybe? Or "built-in" vs "non-built-in" method?), but at least now the confusion has been resolved.

As for the name, none of them are great, and we can easily spend months bikeshedding it, so let's just go with `final`.

alexcrichton: Ah yes indeed, this would cause confusion! 

I don't know what this will mean for regular JS classes. I think it's too early to say whether it will or won't have a difference.

PR #1024, opened 2018-11-09
Add a flag to remove the wasm name section

### LINKED ISSUE #1021: Add option to strip the custom name section
When we are making builds for production, we want them stripped and don't want to have the names section. We should have a CLI flag that wasm-pack can pass to strip the name section.

alexcrichton:
This commit adds a `--remove-name-section` flag to the `wasm-bindgen`
command which will remove the `name` section of the wasm file, used to
indicate the names of functions typically used in debugging. This flag
is off-by-default and will primarily be controlled by wasm-pack,
typically being passed by default with `wasm-pack build --release`.

Closes #1021

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -116,6 +116,7 @@ impl<'a> Context<'a> {
     }
 
     pub fn finalize(&mut self, module_name: &str) -> Result<(String, String), Error> {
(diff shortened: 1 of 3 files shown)


alexcrichton:
Unfortunately we don't have many tests for wasm-bindgen's CLI today, but this'll get tested soon through wasm-pack so I'm gonna go ahead and merge

PR #1030, opened 2018-11-12
Add an optimization to directly wire up imported functions

alexcrichton:
This commit adds an optimization to `wasm-bindgen` to directly import
and invoke other modules' functions from the wasm module, rather than
going through a shim in the imported bindings. This will be an important
optimization in the future for the host bindings proposal, but for now
it's largely just a proof-of-concept to show that we can do it and is
unlikely to bring about many performance benefits.

The implementation in this commit is largely refactoring to reorganize a
bit how functions are imported, but the implementation happens in
`generate_import_function`.

With this commit, 71/287 imports in the `tests/wasm/main.rs` suite get
hooked up directly to the ES modules, no shims needed!

### DIFF
--- a/crates/cli-support/src/js/rust2js.rs
+++ b/crates/cli-support/src/js/rust2js.rs
@@ -6,7 +6,7 @@ use descriptor::{Descriptor, Function};
 /// Helper struct for manufacturing a shim in JS used to translate Rust types to
 /// JS, then invoking an imported JS function.
 pub struct Rust2Js<'a, 'b: 'a> {
-    cx: &'a mut Context<'b>,
+    pub cx: &'a mut Context<'b>,
 
     /// Arguments of the JS shim that we're generating, aka the variables passed
     /// from Rust which are only numbers.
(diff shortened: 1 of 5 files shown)


crates/cli-support/src/js/rust2js.rs
```diff
@@ -499,6 +499,37 @@ impl<'a, 'b> Rust2Js<'a, 'b> {
...
+            global_idx: _,
+            shim_arguments: _,
+        } = self;
+
+        !catch &&
+            !variadic &&
+            prelude.is_empty() &&
+            finally.is_empty() &&
+            (ret_expr == "JS;" || ret_expr == "return JS;") &&
+            js_arguments.iter().all(|arg| !arg.contains("(") && !arg.contains(" "))
```
fitzgen: These last two lines on this predicate are a bit funky... What is going on here?

alexcrichton: Heh, good question! I'll add some comments as well, but the gist is that this is trying to detect if nontrivial transformations are needed on the JS return value or any of the arguments, where only bare return expressions are allowed and arguments which are directly forwarding the JS versions (containing `(` typically means a method call and spaces mean something like a ternary expression)

PR #1035, opened 2018-11-17
Fixed link to the correct crate

WAFFO:
Was originally linking to the wrong crate.


alexcrichton:
Oops, thanks!

limira:
I think there is a confusion here!? They are actually in `js_sys`!
* [js_sys::Array](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Array.html)
* [js_sys::Date](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Date.html)

WAFFO:
Oops, I see what limira is saying. In that specific paragraph it is in fact referring to js_sys.

The reason for the confusion is that the [wasm-bindgen tutorial leads to that page from here](https://rustwasm.github.io/book/reference/debugging.html). So users like me are probably going to that page looking for the crate.

Issue #1034, opened 2018-11-16
Documentation: Needs full example of web-sys DOM


David-OConnor:
The [Websys section of the official guide](https://rustwasm.github.io/wasm-bindgen/examples/dom.html) includes how to build WASM modules, but not how you'd use in a webpage. For example, if I create an HTML file, serve it, and add ` <script type="module" src="my_module.js"></script>`, I receive the error in browser console: `Loading failed for the module with source “http://localhost:8000/my_module_bg”.`

I've found I can get it working using Webpack 4, with a minimal package.json, webpack.config.js, and  but is there a way to do this without using the Webpack and npm ecosystem, eg pure Rust?

index.html includes: `<script type="module" src="main.js"></script>`

package.json:
```
{
  "name": "my_module",
  "license": "MIT",
  "version": "0.1.0",
  "dependencies": {},
  "scripts": {
    "serve": "webpack-dev-server"
  },
  "devDependencies": {
    "webpack": "^4.15.1",
    "webpack-cli": "^3.0.8",
    "webpack-dev-server": "^3.1.4"
  }
}
```

webpack.config.js:
```
const path = require('path');

module.exports = {
    entry: path.join(__dirname, 'main.ts'),
    output: {
        filename: 'main.js',
        path: __dirname
    },
    module: {
        rules: [
            {
                exclude: /node_modules/,
            },
        ],

    },
    resolve: {
        extensions: [".tsx", ".ts", ".js", ".wasm"]
    },
    mode: "development"
};
```

main.ts:
```
const rust = import("./my_module");

rust.then(r => r.run())
```

Stated another way: How can I replace the above code with this line in index.html? `<script type="module" src="my_module.js"></script>` (or .wasm, or _bg.wasm )

limira:
Glad to see there are also someone that want to able to use wasm-bindgen without the need of `nodejs`/`webpack`. As far as I know, unfornately, it is not easy to do what you want. The browsers today not fully support `.wasm` files yet. We need to load it and instantiate it. Currently, I known 2 ways doing it: 
1. `wasm-bindgen --no-modules` and use it directly, or
2. `wasm-bindgen --browser` then use `wasm2es6js` to generate shim like [`wasm-bindgen-test-runner` do](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli/src/bin/wasm-bindgen-test-runner/server.rs#L58-L60)

[...]

PR #1044, opened 2018-11-21
Added web-sys dom example withot npm and webpack

ibaryshnikov:
Solves #1034. The same example as in **dom** folder, but using vanillajs.

### DIFF
--- a/examples/no_modules/Cargo.toml
+++ b/examples/no_modules/Cargo.toml
@@ -8,3 +8,13 @@ crate-type = ["cdylib"]
 
 [dependencies]
 wasm-bindgen = "0.2.28"
+
+[dependencies.web-sys]
+version = "0.3.4"
+features = [
+  'Document',
+  'Element',
+  'HtmlElement',
+  'Node',
+  'Window',
+]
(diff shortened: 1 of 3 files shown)


afdw:
Does not it show the same as [the `no_modules` example](https://github.com/rustwasm/wasm-bindgen/tree/091eaa66f36c3b1cf8c4c2c556e2067383c16670/examples/no_modules)?

ibaryshnikov:
@afdw mostly yes
May be it's better to put more examples in a separate repo, i don't know

alexcrichton:
Sorry for the delay, and thanks for the PR! I wonder if this could be included in the existing `no_modules` example perhaps? I think the Rust code here is the exact same, right, it's just the surrounding setup?

ibaryshnikov:
@alexcrichton yes, the Rust part is the same, and the JS part is slightly different. A made a separate commit, so you can check the [diff](https://github.com/ibaryshnikov/wasm-bindgen/commit/ead95baa4d67205d7be57ed060f33bfe758f778c)

alexcrichton:
Ok cool! Perhaps this could be folded in with the existing `no_modules` example then? (as that's just showing off how DOM things still work in `--no-modules` mode)

ibaryshnikov:
@alexcrichton great idea, updated the pr

alexcrichton:
👍

Issue #922, opened 2018-10-03
Test TypeScript output on CI


alexcrichton:
There have been a number of TypeScript-related regressions recently like https://github.com/rustwasm/wasm-bindgen/issues/917, and while we used to test TypeScript on CI we no longer do! We should figure out how to set up a dedicated builder focused on testing the TypeScript output of wasm-bindgen!

emk:
And as mentioned in that issue, if you'd like, I can try to take a look at setting up an example TypeScript+Rust+Webpack project which can be built using `npm run something` sometime this coming weekend.

alexcrichton:
A nice idea! I think it's good to have a test like that but for this repository we can probably just stick to the vanilla typescript compiler and a sample project 

I'd be wary of coupling to too many things outside of wasm bindgens control

emk:
My recent experience is that if I just check run `wasm-bindgen` and `tsc`, there's still a really impressive number of ways that everything can silently produce broken output. (And it feels like I've hit most of them getting this to work.) In particular, the async module loading code requires cooperation from all 3 of `wasm-bindgen`, `tsc`'s `esnext` support, and `webpack`, and it's easy for one tool to be silently out of alignment with the other 2.

In a perfect world, _something_ would run `webpack` and make sure that everything all hooks up together. Basically, my goal here is to be able to upgrade `wasm-bindgen` and expect everything to still work. But there's an excellent argument to be made that this is an "integration" or "acceptance" test, and that as such, maybe it belongs in a new `rust-webpack-typescript-template` repository somewhere.

Still, I'd love for `wasm-bindgen` to "just work" with TypeScript. And I suspect that it may be hard to guarantee that without a good end-to-end test somewhere.

What would be the most useful way for me to approach this goal?

alexcrichton:
Oh well I think the priority here is getting *something* tested, so I'm fine getting that done however and then we can tweak it as necessary!

PR #1048, opened 2018-11-24
Adds support for #[wasm_bindgen(typescript_custom_section)].

tcr:
## Implementation

This diff adds an attribute "typescript_custom_section", inspired by wasm_custom_section. It allows you to annotate a static const string as an addition to `wasm-bindgen`'s generated TypeScript definition files.

```rust
const TS_INTERFACE_EXPORT: &'static str = r"
  interface Square { color: string; width: number; }
";

// Errors
#[wasm_bindgen]
const TS_INTERFACE_EXPORT = r"interface Square { color: string; width: number; } ";
const TS_INTERFACE_EXPORT: &'static [u8] = [100, 101, 98, 117, 103, 103, 101, 114, 59]
struct Square { color: String, width: usize }
```

When wasm-bindgen-cli is invoked with the --typescript parameter, the contents of all typescript_custom_section strings encountered during compilation are appended to the output `.d.ts` file. This allows Rust to be the source of truth for type definitions in both WebAssembly and Typescript, or route around missing functionality in `wasm-bindgen`. Most importantly, it enables external tooling to directly generate their own Typescript types.

## Motivation

`wasm-bindgen` exports JavaScript classes and enums that mirror the structure of Rust structs and (C-style) enums. It also exports TypeScript definitions when the `--typescript` flag is used. For types defined in Rust that you want to share a JSON-type definition for, you may only need to export a TypeScript definition that models it.

For example, sum types can't be exported to TypeScript just using wasm_bindgen:

```rust
#[wasm_bindgen]
enum Versions {
  V1,
  V2(String),
  V3{  arg1: bool, arg2: String, },
}
```

There is no concrete type in JavaScript that corresponds to a tagged union. But if we treat this as plain old JSON-encodable data (as when serializing using serde), we can actually create a Typescript type definition that lives alongside it to express all possible values:

```typescript
const TS_VERSIONS_EXPORT: &'static str = r#"

export type Versions =
  | { "tag": "V1" }
  | { "tag": "V2", "fields": string }
  | { "tag": "V3", "fields": { "arg1": bool, "arg2": string } }
  ;

"#;
```

The primary benefit is ensuring that Rust is the source of truth for shared type definitions (where the alternative is writing your own Typescript definition that must stay in sync with your Rust definition.) But we can leverage even stronger type guarantees, where Typescript gives you semantics that are similar to Rust's tagged unions:

```typescript
// All matches on a tagged union must be exhausted in TypeScript
switch (version.tag) {
  case 'V1': return 1;
  case 'V2': return 2;
  // would be a typescript error, because you didn't define a 'V3' case
}
```

Knowing that subcrates might want to define additional Typescript information is the motivating factor for `typescript_custom_section`. Rather than integrate advanced Typescript generation entirely into `wasm-bindgen`, we can pull this logic wholly into external libraries.

An example of this is [wasm-typescript-definition, a `#[derive(TypescriptDefinition)]` macro](https://github.com/tcr/wasm-typescript-definition) I wrote to try out typescript_custom_section. This enables serde-`Serialize`able values to export an equivalent TypeScript definition:

```rust
#[derive(Serialize, Deserialize, TypescriptDefinition)]
#[serde(tag = "tag", content = "fields")]
enum Versions {
  #[serde(rename = "Version1")]
  V1,
  #[serde(rename = "Version2")]
  V2(String),
  #[serde(rename = "Version3")]
  V3{  arg1: bool, arg2: String, },
}
```

In this case, the `#[derive(TypescriptDefinition)]` attribute would generate its own additional `#[wasm_bindgen(typescript_custom_section)]` section, which would be rolled into your resulting definitions file. It's then able to be used to typecheck user code.

Because this crate is intentionally serde-aware, it can export a Typescript definition that matches the serde encoding exactly. You could deserialize Rust values in Typescript with full typechecking, then return them to Rust with the same guarantee, just by using a type annotation. Using `typescript_custom_section`, no serde-aware logic needs to be added into wasm-bindgen to make this work.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -22,6 +22,8 @@ pub struct Program {
     /// objects" in the sense that they represent a JS object with a particular
     /// shape in JIT parlance.
     pub dictionaries: Vec<Dictionary>,
+    /// custom typescript sections to be included in the definition file
+    pub typescript_custom_sections: Vec<String>,
 }
 
 /// A rust to js interface. Allows interaction with rust objects/functions
--- a/crates/backend/src/encode.rs
+++ b/crates/backend/src/encode.rs
@@ -46,6 +46,7 @@ fn shared_program<'a>(prog: &'a ast::Program, intern: &'a Interner)
         imports: prog.imports.iter()
             .map(|a| shared_import(a, intern))
             .collect::<Result<Vec<_>, _>>()?,
+        typescript_custom_sections: prog.typescript_custom_sections.iter().map(|x| -> &'a str { &x }).collect(),
         // version: shared::version(),
         // schema_version: shared::SCHEMA_VERSION.to_string(),
     })
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2143,6 +2143,10 @@ impl<'a, 'b> SubContext<'a, 'b> {
         for s in self.program.structs.iter() {
             self.generate_struct(s)?;
         }
+        for s in self.program.typescript_custom_sections.iter() {
+            self.cx.typescript.push_str(s);
+            self.cx.typescript.push_str("\n\n");
+        }
 
         Ok(())
     }
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -207,6 +207,13 @@ impl BindgenAttrs {
             _ => false,
         })
     }
+
+    fn typescript_custom_section(&self) -> bool {
+        self.attrs.iter().any(|a| match *a {
+            BindgenAttr::TypescriptCustomSection => true,
+            _ => false,
+        })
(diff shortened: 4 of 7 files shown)


alexcrichton:
This is a really interesting idea! It seems like a pretty plausible building block to me at least.

@fitzgen you may have thoughts on this as well! I'm sort of tempted to merge as-is, but would like to get a second opinion!

fitzgen:
Yeah, this is neat for sure. LGTM!

alexcrichton:
👍 Let's merge then!

alexcrichton:
@tcr if you're interested as well, I'd love to beef up our TypeScript testing story on CI for wasm-bindgen, and it'd be great to get some help with that!

tcr:
@alexcrichton I'd love to be able to add Typescript tests :) I'll join the conversation in #922.

PR #1050, opened 2018-11-25
examples: add param to console_log build

piiih:
This PR adds the `+nightly` param to the console_log `build.sh` file.

I don't know if this is a problem only in my PC, but when I tried to run without the `+nightly` param it didn't work.

Sorry if I'm making a PR in wrong format, but I really didn't find instructions about PRs.

### DIFF
--- a/examples/console_log/build.sh
+++ b/examples/console_log/build.sh
@@ -4,7 +4,7 @@
 
 set -ex
 
-cargo build --target wasm32-unknown-unknown
+cargo +nightly build --target wasm32-unknown-unknown
 cargo run --manifest-path ../../crates/cli/Cargo.toml \
   --bin wasm-bindgen -- \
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This I don't think should be required any more because the example compiles on stable Rust, but can you gist an error message you were getting without this change?

piiih:
Hi @alexcrichton, when I ran without `+nightly` I got this https://gist.github.com/piiih/9a9a4608077e0a7d526ee89c80375a10

alexcrichton:
Ah ok, if you run `rustup target add wasm32-unknown-unknown` does it solve that issue?

piiih:
@alexcrichton Oh yes, my bad, I forgot that I added `wasm` just for nightly, closing...

PR #1055, opened 2018-11-28
Assert all attributes are used by default 

alexcrichton:
This commit implements a system that will assert that all
`#[wasm_bindgen]` attributes are actually used during compilation. This
should help ensure that we don't sneak in stray attributes that don't
actually end up having any meaning, and hopefully make it a bit easier
to learn `#[wasm_bindgen]`!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,10 @@ std = []
 serde-serialize = ["serde", "serde_json", "std"]
 nightly = []
 
+# Whether or not the `#[wasm_bindgen]` macro is strict and generates an error on
+# all unused attributes
+strict-macro = ["wasm-bindgen-macro/strict-macro"]
+
 # This is only for debugging wasm-bindgen! No stability guarantees, so enable
 # this at your own peril!
 xxx_debug_only_print_generated_code = ["wasm-bindgen-macro/xxx_debug_only_print_generated_code"]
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -24,12 +24,16 @@ test = false
 doctest = false
 
 [features]
-default = ["spans", "std"]
+default = ["spans", "std", 'strict-macro']
```
alexcrichton: Note that this is on-by-default, but I don't expect to land it like this. I suspect we'll have to turn this to off-by-default (and perhaps on-by-deafult for the next major release). In the meantime this is being used to test CI.

fitzgen: Since the other strings in this file use double quotes, this new default feature string should also use double quotes.

Cargo.toml
```diff
@@ -24,12 +24,16 @@ test = false
...
 spans = ["wasm-bindgen-macro/spans"]
 std = []
 serde-serialize = ["serde", "serde_json", "std"]
 nightly = []
 
+# Whether or not the `#[wasm_bindgen]` macro is strict and generates an error on
+# all unused attributes
+strict-macro = ['wasm-bindgen-macro/strict-macro']
```
fitzgen: Ditto about double quotes.

PR #1057, opened 2018-11-28
Add a `#[wasm_bindgen(start)]` attribute

### LINKED ISSUE #74: Producing wasm-binary if there's main.rs
Currently, the only way to execute rust code is via exposing a function and then calling it on js side. It would be nice to be able to do something like this: 
`main.rs`
```rust
#![feature(proc_macro)]

extern crate wasm_bindgen;

use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern {
    fn log(s: &str);
}

fn main() {
    log("Hello from Rust");
}
```
Then using the cli to generate the required js and including the js directly in the html to execute it, instead of going through an intermediate state of needing to manually write another js file.

alexcrichton:
This commit adds a new attribute to `#[wasm_bindgen]`: `start`. The
`start` attribute can be used to indicate that a function should be
executed when the module is loaded, configuring the `start` function of
the wasm executable. While this doesn't necessarily literally configure
the `start` section, it does its best!

Only one crate in a crate graph may indicate `#[wasm_bindgen(start)]`,
so it's not recommended to be used in libraries but only end-user
applications. Currently this still must be used with the `crate-type =
["cdylib"]` annotation in `Cargo.toml`.

The implementation here is somewhat tricky because of the circular
dependency between our generated JS and the wasm file that we emit. This
circular dependency makes running initialization routines (like the
`start` shim) particularly fraught with complications because one may
need to run before the other but bundlers may not necessarily respect
it. Workarounds have been implemented for various emission strategies,
for example calling the start function directly after exports are wired
up with `--no-modules` and otherwise working around what appears to be
a Webpack bug with initializers running in a different order than we'd
like. In any case, this in theory doesn't show up to the end user!

Closes #74

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -11,6 +11,7 @@ use shared;
 use super::Bindgen;
 use descriptor::{Descriptor, VectorKind};
 use wasm_interpreter::Interpreter;
+use wasm_utils::Remap;
 
 mod js2rust;
 use self::js2rust::Js2Rust;
@@ -30,6 +31,7 @@ pub struct Context<'a> {
     pub imported_statics: HashSet<&'a str>,
     pub config: &'a Bindgen,
     pub module: &'a mut Module,
+    pub start: Option<String>,
 
     /// A map which maintains a list of what identifiers we've imported and what
     /// they're named locally.
@@ -163,7 +165,6 @@ impl<'a> Context<'a> {
     }
 
     pub fn finalize(&mut self, module_name: &str) -> Result<(String, String), Error> {
-        self.parse_wasm_names();
         self.write_classes()?;
 
         self.bind("__wbindgen_object_clone_ref", &|me| {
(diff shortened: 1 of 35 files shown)


alexcrichton:
For posterity, the major downside of `#[wasm_bindgen(start)]` is that you still have to `crate-type = ["cdylib"]` your crates. It's also not as ergonomic/idiomatic as using a `bin` crate type with `fn main() { ... }`.

The downsides of using `fn main() {}`, however, I think are:

* The standard library isn't really ready for these semantics. It thinks that the "main function" is the entire program, tearing down standard library state when it exists. Contrasted with the wasm start function, the start function is just the beginning of execution and often schedules more work to happen later.
* The standard library's `fn main() {}` also currently brings in some weighty initialization, meaning that binaries are a bit larger than they would otherwise need to be. This is likely surmountable though!

The thinking is that we can start off conservatively with a specific attribute like this, and then in the future if we see a way forward to `fn main() { ... }` working we can do that instead!

crates/cli-support/src/js/mod.rs
```diff
@@ -460,6 +461,37 @@ impl<'a> Context<'a> {
 
         self.unexport_unused_internal_exports();
         closures::rewrite(self)?;
+
+        // Handle the `start` function, if one was specified. If we're in a
+        // --test mode (such as wasm-bindgen-test-runner) then we skip this
+        // entirely. Otherwise we want to first add a start function to the
+        // `start` section if one is specified.
+        //
+        // Afterwards, we need to perfom what's a bit of a hack. Right after we
```
fitzgen: Nitpick: "perfom" -> "perform"

```suggestion
        // Afterwards, we need to perform what's a bit of a hack. Right after we
```

crates/cli-support/src/js/mod.rs
```diff
@@ -460,6 +461,37 @@ impl<'a> Context<'a> {
...
+        //
+        // For ESM outputs bundlers like webpack also don't work because
+        // currently they run the wasm initialization before the JS glue
+        // initialization, meaning that if the wasm start function calls
+        // imported functions the JS glue isn't ready to go just yet.
+        //
+        // To handle `--no-modules` we just unstart the start function and call
+        // it manually. To handle the ESM use case we switch the start function
+        // to calling an imported function which defers the start function via
+        // `Promise.resolve().then(...)` to execute on the next microtask queue.
```
fitzgen: Nitpick: the microtask queue is the queue of (micro) tasks, *not* a tick of the (micro) queue or a turn of the event loop, so this reads a bit funny to me.

```suggestion
        // `Promise.resolve().then(...)` to execute on the next microtask tick.
```

crates/macro/ui-tests/start-function.rs
```diff
@@ -0,0 +1,12 @@
...
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(start)]
+pub fn foo() {}
+
+#[wasm_bindgen(start)]
+pub fn foo2(x: u32) {}
+
+#[wasm_bindgen(start)]
+pub fn foo3<T>() {}
```
fitzgen: Would be good to have a test for two otherwise valid start functions, to ensure that we properly error and don't accidentally just take the first one or something.

examples/closures/src/lib.rs
```diff
@@ -7,7 +7,7 @@ use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 use web_sys::{Document, Element, HtmlElement, Window};
 
-#[wasm_bindgen]
+#[wasm_bindgen(start)]
```
fitzgen: Nice

crates/macro/ui-tests/start-function.rs
alexcrichton: Ah unfortunately we don't have a great place to put such tests today, but I've filed https://github.com/rustwasm/wasm-bindgen/issues/1058 to track this!

PR #1059, opened 2018-11-29
[WIP] Add explicit wasm extension

pd4d10:
For the case that `wasm` is not specified at webpack `resolve.extensions`

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -643,7 +643,7 @@ impl<'a> Context<'a> {
                     .push_str(&format!("wasm = require('./{}_bg');", module_name));
                 format!("var wasm;")
             } else {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Unfortunately though it looks like CI has discovered that we rely on the convention that if you omit the extension it'll find the file via the first extension that matches. The theory is that the actual extension behind that import doesn't matter, and while it could be a wasm file tools like `wasm2es6js` will replace it with a JS file. 

We may be able to fix this directly though with the tests? It seems like this may be worthwhile fixing in the wild!

alexcrichton:
Indeed and I agree! This doesn't pass CI though and that needs fixing before we can land it.

alexcrichton:
Ok I'm gonna close this to help clear out our PR list for now, but we can always land this with working tests!

PR #1069, opened 2018-11-30
Switch from heap/stack to just a heap

alexcrichton:
This commit switches strategies for storing `JsValue` from a heap/stack
to just one heap. This mirrors the new strategy for `JsValue` storage
in #1002 and should make multiplexing those strategies at
`wasm-bindgen`-time much easier.

Instead of having one array which acts as a stack for borrowed values
and one array for a heap of borrowed values, only one JS array is used
for storage of JS values now. This makes `getObject` far simpler by
simply being an array access, but it means that cloning an object now
reserves a new slot instead of reference counting it. If the old
reference counting behavior is needed it's thought that `Rc<JsValue>`
can be used in Rust.

The new "heap" has an initial stack pointer which grows downwards, and a
heap which grows upwards. The heap is a singly-linked-list which is
allocated/deallocated from. The stack grows downwards to zero and
presumably starts generating errors once it underflows. An initial stack
size of 32 is chosen as that should encompass all use cases today, but
we can eventually probably add configuration for this!

Note that the heap is initialized to all `null` for the stack and then
the initial JS values (`undefined`, `null`, `true`, `false`) are pushed
onto the heap in reserved locations.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -114,7 +114,9 @@ enum Import<'a> {
     },
 }
 
-const INITIAL_SLAB_VALUES: &[&str] = &["undefined", "null", "true", "false"];
+const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
+// Must be kept in sync with `src/lib.rs` of the `wasm-bindgen` crate
+const INITIAL_HEAP_OFFSET: usize = 32;
 
 impl<'a> Context<'a> {
     fn export(&mut self, name: &str, contents: &str, comments: Option<String>) {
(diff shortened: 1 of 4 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -916,149 +881,49 @@ impl<'a> Context<'a> {
...
-                    }}
-                    ",
-                    initial_values.len()
-                ),
-                None,
-            );
+        self.global("const heap = [];");
+        let init = format!(
+            "for (let i = 0; i < {}; i++) heap.push(undefined);",
+            INITIAL_HEAP_OFFSET,
```
fitzgen: Slightly better:

```js
const heap = new Array(INITIAL_HEAP_OFFSET);
heap.fill(undefined);
```

alexcrichton: Oh nice! I'll omit the `fill` as I don't think we actually need it, and otherwise this works great

alexcrichton:
> We should have a comment somewhere describing the memory layout of this array and how the heap grows towards the wilderness and the stack grows down.

Oops right! I've updated the internal contributing documentation to reflect this change

PR #1073, opened 2018-12-01
Fix typo in "wasm_bindgen::JsCast" trait docs

daxpedda:
```This method will return `Err(self)` ``` **is** ``` `self.is_instance_of::<T>()` returns `false` ```
should be
```This method will return `Err(self)` ``` **if** ``` `self.is_instance_of::<T>()` returns `false` ```


alexcrichton:
Oops, thanks!

PR #1077, opened 2018-12-02
Rename iterator types and add some docs

derekdreery:
Continued from #1043.


derekdreery:
I've thrown away my alternative implementation, and added some new names/docs to help anyone else like me to understand the design decisions that have been made.

alexcrichton:
I think this has gone somewhat stale over time, so I'm gonna go ahead and close this. It can of course be resubmitted with feedback addressed though!

PR #1082, opened 2018-12-04
 Added out-name flag to the cli, offering an alternative over the crate name

David-OConnor:
Untested! Builds, but unable to test due to this error: 
```
a different version of wasm-bindgen than this binary:

  rust wasm file: 0.2.29
     this binary: 0.2.12 (d37d4ae81)
```

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -25,6 +25,7 @@ pub mod wasm2es6js;
 
 pub struct Bindgen {
     path: Option<PathBuf>,
+    out_name: Option<String>,
     nodejs: bool,
     nodejs_experimental_modules: bool,
     browser: bool,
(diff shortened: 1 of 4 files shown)


David-OConnor:
Looks like there

David-OConnor:
Can't figure out how to make the PR without the old PR files in there

chinedufn:
@David-OConnor WRT to your error have you already tried updating your local `wasm-bindgen-cli` binary?

`cargo install wasm-bindgen-cli -f` would be the easiest way.

Alternatively you could download and unpack a release binary](https://github.com/rustwasm/wasm-bindgen/releases) and move it to your `~/.cargo/bin/` if you don't want to build from source.

David-OConnor:
@chinedufn I think the root cause of that error, and of the bogus files from months ago being committed to the PR, is that I didn't follow the procedures [On this page](https://stackoverflow.com/questions/7244321/how-do-i-update-a-github-forked-repository)

I'll reattack this when I get home from work

PR #1091, opened 2018-12-08
Add support for importing reserved JS words

### LINKED ISSUE #1006: Using #[wasm_bindgen(constructor)] for a module that is a type
I am trying to use an NPM module that only has 1 default export that is a type. Here is the JS code I'm trying to produce:

```js
import ScatterEos from "scatterjs-plugin-eos";
const eos = new ScatterEos();
```

Here are bindings I thought might work:

```rust
#[wasm_bindgen(module = "scatterjs-plugin-eosjs")]
extern "C" {
    pub type EosPlugin;

    #[wasm_bindgen(constructor)]
    pub fn new() -> EosPlugin;
}
```

But this is generating an import statement like this:

```js
import { EosPlugin } from 'scatterjs-plugin-eosjs';
```

Is there anyway to get this working without custom JS glue code?

mvlabat:
This should fix #1006

Implemented the idea explained in https://github.com/rustwasm/wasm-bindgen/issues/1006#issuecomment-442621024.

Now
```rust
#[wasm_bindgen(module = "./SomeModule")]
extern "C" {
    type SomeType;
}
```

should be glued with
```js
import { default as default1 } from './SomeModule';
```

I tested this behaviour on my local example project and it seems to be OK, though I couldn't get it working for the tests and `import_js` example. For some reason `js_name` attribute for types is just ignored there.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -119,54 +119,59 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
 const INITIAL_HEAP_OFFSET: usize = 32;
 
 lazy_static! {
-    static ref JS_KEYWORDS: HashSet<&'static str> = {
-        let mut keywords = HashSet::new();
-        keywords.insert("await");
-        keywords.insert("break");
-        keywords.insert("case");
-        keywords.insert("catch");
-        keywords.insert("class");
-        keywords.insert("const");
-        keywords.insert("continue");
-        keywords.insert("debugger");
-        keywords.insert("default");
-        keywords.insert("delete");
-        keywords.insert("do");
-        keywords.insert("else");
-        keywords.insert("enum");
-        keywords.insert("export");
-        keywords.insert("extends");
-        keywords.insert("false");
-        keywords.insert("finally");
-        keywords.insert("for");
-        keywords.insert("function");
-        keywords.insert("if");
-        keywords.insert("implements");
-        keywords.insert("import");
-        keywords.insert("in");
-        keywords.insert("instanceof");
-        keywords.insert("interface");
-        keywords.insert("new");
-        keywords.insert("null");
-        keywords.insert("package");
-        keywords.insert("private");
-        keywords.insert("protected");
-        keywords.insert("public");
-        keywords.insert("return");
-        keywords.insert("super");
-        keywords.insert("switch");
-        keywords.insert("this");
-        keywords.insert("throw");
-        keywords.insert("true");
-        keywords.insert("try");
-        keywords.insert("typeof");
-        keywords.insert("undefined");
-        keywords.insert("var");
-        keywords.insert("void");
-        keywords.insert("while");
-        keywords.insert("with");
-        keywords.insert("yield");
-        keywords
+    static ref JS_RESERVED_WORDS: HashSet<&'static str> = {
+        let mut words = HashSet::with_capacity(49);
+        words.insert("Infinity");
+        words.insert("NaN");
+        words.insert("arguments");
+        words.insert("await");
+        words.insert("break");
+        words.insert("case");
+        words.insert("catch");
+        words.insert("class");
+        words.insert("const");
+        words.insert("continue");
+        words.insert("debugger");
+        words.insert("default");
+        words.insert("delete");
+        words.insert("do");
+        words.insert("else");
+        words.insert("enum");
+        words.insert("export");
+        words.insert("extends");
+        words.insert("false");
+        words.insert("finally");
+        words.insert("for");
+        words.insert("function");
+        words.insert("if");
+        words.insert("implements");
+        words.insert("import");
+        words.insert("in");
+        words.insert("instanceof");
+        words.insert("interface");
+        words.insert("let");
+        words.insert("new");
+        words.insert("null");
+        words.insert("package");
+        words.insert("private");
+        words.insert("protected");
+        words.insert("public");
+        words.insert("require");
+        words.insert("return");
+        words.insert("super");
+        words.insert("switch");
+        words.insert("this");
+        words.insert("throw");
+        words.insert("true");
+        words.insert("try");
+        words.insert("typeof");
+        words.insert("undefined");
+        words.insert("var");
+        words.insert("void");
+        words.insert("while");
+        words.insert("with");
+        words.insert("yield");
+        words
     };
 }
 
@@ -2701,7 +2706,7 @@ impl<'a> Import<'a> {
 fn generate_identifier(name: &str, used_names: &mut HashMap<String, usize>) -> String {
     let cnt = used_names.entry(name.to_string()).or_insert(0);
     *cnt += 1;
-    if *cnt == 1 && !JS_KEYWORDS.contains(name) {
+    if *cnt == 1 && !JS_RESERVED_WORDS.contains(name) {
         name.to_string()
     } else {
         format!("{}{}", name, cnt)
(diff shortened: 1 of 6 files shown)


mvlabat:
Oh, I see... I've forgotten to put `js_class` attribute for `RenamedTypes`. Though setting `js_class = default` doesn't seem to do anything either. Still digging...

crates/cli-support/src/js/mod.rs
```diff
@@ -118,6 +118,58 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
...
+        keywords.insert("public");
+        keywords.insert("return");
+        keywords.insert("super");
+        keywords.insert("switch");
+        keywords.insert("this");
+        keywords.insert("throw");
+        keywords.insert("true");
+        keywords.insert("try");
+        keywords.insert("typeof");
+        keywords.insert("undefined");
```
Pauan: As weird as it may sound, `undefined` isn't a reserved word.

crates/cli-support/src/js/mod.rs
```diff
@@ -118,6 +118,58 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
...
+        keywords.insert("instanceof");
+        keywords.insert("interface");
+        keywords.insert("new");
+        keywords.insert("null");
+        keywords.insert("package");
+        keywords.insert("private");
+        keywords.insert("protected");
+        keywords.insert("public");
+        keywords.insert("return");
+        keywords.insert("super");
```
Pauan: Missing `static`

crates/cli-support/src/js/mod.rs
```diff
@@ -118,6 +118,58 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
...
+        keywords.insert("finally");
+        keywords.insert("for");
+        keywords.insert("function");
+        keywords.insert("if");
+        keywords.insert("implements");
+        keywords.insert("import");
+        keywords.insert("in");
+        keywords.insert("instanceof");
+        keywords.insert("interface");
+        keywords.insert("new");
```
Pauan: Missing `let`

crates/cli-support/src/js/mod.rs
```diff
@@ -118,6 +118,58 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
 // Must be kept in sync with `src/lib.rs` of the `wasm-bindgen` crate
 const INITIAL_HEAP_OFFSET: usize = 32;
 
+lazy_static! {
+    static ref JS_KEYWORDS: HashSet<&'static str> = {
+        let mut keywords = HashSet::new();
+        keywords.insert("await");
```
Pauan: Missing `arguments`

crates/cli-support/src/js/mod.rs
```diff
@@ -118,6 +118,58 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
...
+        keywords.insert("continue");
+        keywords.insert("debugger");
+        keywords.insert("default");
+        keywords.insert("delete");
+        keywords.insert("do");
+        keywords.insert("else");
+        keywords.insert("enum");
+        keywords.insert("export");
+        keywords.insert("extends");
+        keywords.insert("false");
```
Pauan: Missing `eval`

mvlabat: Is it a reserved word? I think it's just a built-in function and I'm not sure we want to have them in this list.

crates/cli-support/src/js/mod.rs
mvlabat: :) Yeah, I know it's a global property, but I thought it's something we would like to leave here. It's totally legal to have such property name, so in the very rare case it can prevent breaking code.

mvlabat:
@Pauan, so I think we want to add everything from this [list of global objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects), don't we? We should be consistent with adding such things as `undefined`, `arguments`.. Pretty much everything can break the code if a user decides to use such names in imports.

Pauan:
@mvlabat `arguments` and `eval` are reserved words in strict mode:

https://tc39.github.io/ecma262/#sec-identifiers-static-semantics-early-errors

You can verify this by attempting to use the following JS code:

```js
import { arguments, eval } from "foo";
```

You will get a SyntaxError.

As for globals (including `undefined`), we should *not* mangle them.

mvlabat:
@Pauan alright, got it. And fixed

mvlabat:
@Pauan  Ahh.. And it seems like we need some kind of special treatment for `eval`. Like mangling it only for actual import statements. I'm not sure what's the better way to workaround this. Do you have any suggestions?

Pauan:
@mvlabat I think the logic should be something like "if this identifier was imported, then mangle it, otherwise don't". It would need to run on every identifier, not just the imported identifiers.

Pauan:
@mvlabat Consider this JS code:

```js
import { eval as eval1 } from "foo";

console.log(eval1);
```

As you can see, it needs to mangle it at both the import site and also the use site. Otherwise the use site will still be using `eval`, which is wrong.

Though I guess your `RenamedTypes` test probably accounts for that, so maybe it's already fine as-is?

mvlabat:
Oh, you meant this case. Yeah, it must work correctly. If I understand it right, `generate_identifier` isn't even supposed to be called twice in that case. If it did, we would actually have something like that:
```js
import { eval as eval1 } from "foo";

console.log(eval2);
```

And there's nothing my code changed in this respect.

mvlabat:
@alexcrichton, if everything looks ok, can we merge the PR and make a new release? ) If not, I'll surely try to fix it. Still can't use wasm-bindgen built locally with Parcel, and further development of my project is currently blocked without this change. Thank you!

alexcrichton:
Thanks for the PR @mvlabat! 

Instead of using `lazy_static` could this initialize a map per-`Context` and use that instead? Additionally how come only the import-from-a-module use case is handled here? Can these identifiers actually be names of functions in the global environment?

It may also be somewhat overkill to try to add an exhaustive list of keywords here, I suspect basically nothing comes up in practice except "default"?

Pauan:
> Can these identifiers actually be names of functions in the global environment?

Not under normal circumstances, since they're reserved words.

The only way to create/access them on the global environment is to use `window.default` (and similarly for the other names).

mvlabat:
@alexcrichton 
> Instead of using lazy_static could this initialize a map per-Context and use that instead? 
I think that `lazy_static` is a more preferable choice here, because we don't need to reinitialize the set of the keywords: it's not the map that's being mutated, it's really just a static set. Creating it for each context seems to be redundant.

> Additionally how come only the import-from-a-module use case is handled here? Can these identifiers actually be names of functions in the global environment?
If we want to support this, it's possible, I think. Though it would be a really, really rare case. I'm just not sure if we can return right from `generate_identifier` something like this: `window.identifier` (for browsers) or `global.identifier` (for node.js).

> It may also be somewhat overkill to try to add an exhaustive list of keywords here, I suspect basically nothing comes up in practice except "default"?
If we want to support only default exports, then I can publish an alternative PR with much fewer lines of code, that handles only that specific case. And I think all the questions above will just become irrelevant )

alexcrichton:
Ok thanks for the info! I'd be curious to see if we could just handle the `default` case here yeah and see if that'll get us basically 99% of the way there

alexcrichton:
Closing in favor of https://github.com/rustwasm/wasm-bindgen/pull/1106, thanks again @mvlabat!

PR #1093, opened 2018-12-08
Add Uint16 to `buffer_data`

### LINKED ISSUE #1089: Add `buffer_data_with_u16_array` to WebGlRenderingContext
Right now there's a [buffer_data_with_u8_array](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.WebGlRenderingContext.html#method.buffer_data_with_u8_array) which is super convenient if you want o use `WebGLRenderingContext::UNSIGNED_BYTE`.

However, `UNSIGNED_SHORT` is probably more common, so `buffer_data_with_u16_array` would be super useful IMO.

Any reason that this shouldn't exist?

If not - I'd be happy to add it if you can point me in the right direction!

chinedufn:
Closes #1089

---

There was no `IdlType::Uint16ArrayMut` so I went with `Uint16Array`. Is this an issue?

### DIFF
--- a/crates/webidl/src/idl_type.rs
+++ b/crates/webidl/src/idl_type.rs
@@ -637,7 +637,7 @@ impl<'a> IdlType<'a> {
                 .iter()
                 .flat_map(|idl_type| idl_type.flatten())
                 .collect(),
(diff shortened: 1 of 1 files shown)


chinedufn:
I don't need this anymore as I ended up going with this approach https://github.com/rustwasm/wasm-bindgen/pull/918#issuecomment-445495395

PR #1098, opened 2018-12-10
examples: Add a requestAnimationFrame loop example

### LINKED ISSUE #976: Add example of using request animation frame
After searching the web sys docs and wasm-bindgen issues I can't seem to find anything about request animation frame.

Months ago, before web-sys, I tried to use RAF via an extern call and ran into some closure issues (don't remember what exactly but likely just me misunderstanding things). 

Ended up giving up and just calling some JS function from my Rust code that handled my RAF loop for me.

Now that I'm circling back to this.. does the RAF API not exist in web-sys for some reason? Or does it actually exist and I just didn't find it?

fitzgen:
+cc @chinedufn @gregtatum

Fixes #976

### DIFF
--- a/examples/request-animation-frame/src/lib.rs
+++ b/examples/request-animation-frame/src/lib.rs
@@ -1,0 +1,57 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::JsCast;
+
+fn window() -> web_sys::Window {
+    web_sys::window().expect("no global `window` exists")
+}
+
+fn request_animation_frame(f: &Closure<FnMut()>) {
+    window()
+        .request_animation_frame(f.as_ref().unchecked_ref())
+        .expect("should register `requestAnimationFrame` OK");
+}
+
+fn document() -> web_sys::Document {
+    window()
+        .document()
+        .expect("should have a document on window")
+}
+
+fn body() -> web_sys::HtmlElement {
+    document().body().expect("document should have a body")
+}
+
+// This function is automatically invoked after the wasm module is instantiated.
+#[wasm_bindgen(start)]
+pub fn run() -> Result<(), JsValue> {
+    let f = Rc::new(RefCell::new(None));
+    let g = f.clone();
+
+    {
+        let mut i = 0;
+        *g.borrow_mut() = Some(Closure::wrap(Box::new(move || {
+            if i > 300 {
+                body().set_text_content(Some("All done!"));
+
+                // Drop our handle to this closure so that it will get cleaned
+                // up once we return.
+                let _ = f.borrow_mut().take();
+                return;
+            }
+
+            // Set the body's text content to how many times this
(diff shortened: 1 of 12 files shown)


examples/request-animation-frame/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+fn body() -> web_sys::HtmlElement {
+    document().body().expect("document should have a body")
+}
+
+// This function is automatically invoked after the wasm module is instantiated.
+#[wasm_bindgen(start)]
+pub fn run() -> Result<(), JsValue> {
+    let f = Rc::new(RefCell::new(None));
```
chinedufn: Wow! I see! This is so, so cool... Thanks!!

examples/request-animation-frame/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+            i += 1;
+            let text = format!("requestAnimationFrame has been called {} times.", i);
+            body().set_text_content(Some(&text));
+
+            // Schedule ourself for another requestAnimationFrame callback.
+            request_animation_frame(f.borrow().as_ref().unwrap());
+        }) as Box<FnMut()>));
+    }
+
+    request_animation_frame(g.borrow().as_ref().unwrap());
```
chinedufn: How come we don't need to `.forget()` the closure at all in this example?

fitzgen: `forget` is for when you want to leak the closure so that it is permanently alive. This is using `Rc` to keep it alive only until when we are done with the loop.

https://docs.rs/wasm-bindgen/0.2.29/wasm_bindgen/closure/struct.Closure.html#method.forget

alexcrichton:
👍

examples/request-animation-frame/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+// This function is automatically invoked after the wasm module is instantiated.
+#[wasm_bindgen(start)]
+pub fn run() -> Result<(), JsValue> {
+    let f = Rc::new(RefCell::new(None));
+    let g = f.clone();
+
+    {
+        let mut i = 0;
+        *g.borrow_mut() = Some(Closure::wrap(Box::new(move || {
```
dakom: Would be really helpful to have an explanatory comment or two on this line. No pressure, and no doubt the only non-core-Rust thing there is the `Closure::wrap()` ... but still, it would help to get what's going on (like why are we re-assigning it just after we assigned it to a clone?)

alexcrichton: Certainly! How's [this look](https://github.com/rustwasm/wasm-bindgen/commit/fcee4656923a1128d0a2faf781242ee2d7835db4)?

dakom: Excellent, thanks!

gregtatum:
Thank you for the work on this! This was a very clear example, and I got it working pretty quickly on my end, I now have a green moving box 😬

image

Issue #1099, opened 2018-12-10
Port our examples to Rust 2018 edition!


fitzgen:
Let's port our examples to the 2018 edition!

## How to port an example crate

1. Leave a comment here claiming an example, so that we don't duplicate work.
2. Run `cargo fix --edition`
3. Add `edition = "2018"` to the `[package]` section of the example's `Cargo.toml`
4. Run `cargo check --target wasm32-unknown-unknown`
    - If it compiles OK and without any warnings:
        - commit your changes and send a PR!
    - Otherwise:
        - trouble shoot, fix any errors and warnings, go back to the start of step 4.
5. (optional) Run `cargo fix --edition-idioms` and remove all `extern crate` directives

The [2018 edition guide](https://rust-lang-nursery.github.io/edition-guide/introduction.html) might be helpful reading.

## TODO

- [x] `examples/add/`
- [x] `examples/canvas/`
- [x] `examples/char/`
- [x] `examples/closures/`
- [x] `examples/console_log/`
- [x] `examples/dom/`
- [x] `examples/duck-typed-interfaces/`
- [x] `examples/fetch/`
- [x] `examples/guide-supported-types-examples/`
- [x] `examples/hello_world/`
- [x] `examples/import_js/`
- [ ] `examples/julia_set/`
- [ ] `examples/no_modules/`
- [ ] `examples/paint/`
- [ ] `examples/performance/`
- [x] `examples/raytrace-parallel/`
- [x] `examples/README.md`
- [x] `examples/todomvc/`
- [x] `examples/wasm2js/`
- [x] `examples/wasm-in-wasm/`
- [x] `examples/webaudio/`
- [x] `examples/webgl/`

andrehjr:
Can I work on add, canvas, char, and hello_world?

Already tried running `cargo fix --edition` on a few examples, and they mostly work without any warnings or issues. So, they only change necessary for those cases is to add `edition = "2018"` on Cargo.toml?

elpiel:
I will pickup the `console_log` as a start.

andrehjr:
Picking up webgl, webaudio, wasm-in-wasm and todomvc

T5uku5hi:
I'll pickup `dom`, `fetch`, `guide-supported-types-examples`, `import_js`.

elpiel:
I will try to implement `julia_set`, `no_modules`, `paint` and `performance`

andrehjr:
Will try raytrace-parallel now!

[...]

PR #1100, opened 2018-12-11
Port examples - console-log, closures and duck-typed-interfaces to Rust 2018

elpiel:
I've left the `_s` from the fix for duck-typed-interfaces/src/lib.rs, it's not required, but removes the warning for unused variable

Checks `console-log`, `closures` and `duck-typed-interfaces` tasks #1099

### DIFF
--- a/examples/closures/src/lib.rs
+++ b/examples/closures/src/lib.rs
@@ -1,5 +1,3 @@
-use web_sys;
-
 use js_sys::{Array, Date};
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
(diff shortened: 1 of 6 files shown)


examples/closures/src/lib.rs
```diff
@@ -1,6 +1,4 @@
-extern crate js_sys;
-extern crate wasm_bindgen;
-extern crate web_sys;
+use web_sys;
```
alexcrichton: I believe in 2018 you can omit this too!

alexcrichton:
Looks great to me, thanks! Just one last line to remove and I think we're good to go

elpiel:
Yes, I've missed this one.

examples/closures/src/lib.rs
elpiel: Removed

PR #1101, opened 2018-12-11
Port examples: add, canvas, char and hello_world to rust 2018 edition examples

andrehjr:
As part of issue #1099

This PR ports the following examples:
- add
- canvas
- char
- hello_world

also ran `cargo fix --edition-idioms`

### DIFF
--- a/examples/canvas/src/lib.rs
+++ b/examples/canvas/src/lib.rs
@@ -1,7 +1,6 @@
 use std::f64;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
-use web_sys;
 
 #[wasm_bindgen(start)]
 pub fn start() {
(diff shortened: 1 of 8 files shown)


examples/canvas/src/lib.rs
```diff
@@ -1,10 +1,7 @@
-extern crate js_sys;
-extern crate wasm_bindgen;
-extern crate web_sys;
-
 use std::f64;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
+use web_sys;
```
alexcrichton: I think you can actually remove this `web_sys` line in 2018 too!

alexcrichton:
Looks great, thanks so much!

examples/canvas/src/lib.rs
andrehjr: Fixed! 🚀

PR #1106, opened 2018-12-11
Add support for importing default exports

### LINKED ISSUE #1006: Using #[wasm_bindgen(constructor)] for a module that is a type

mvlabat:
This should fix #1006

This PR doesn't mangle reserved JS words, only `default`. This is an alternative to #1091

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,6 +120,11 @@ matrix:
         - if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then aws s3 sync ~/$TRAVIS_BUILD_NUMBER s3://wasm-bindgen-ci/$TRAVIS_BUILD_NUMBER; fi
       if: branch = master
 
+    # The `cli-support` crate's tests pass
+    - name: "test cli-support crate"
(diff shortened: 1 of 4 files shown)


mvlabat:
O_o. For some reason `failure` crate fails to build. Can't reproduce it locally... Is it some kind of CI bug?

alexcrichton:
No worries that was my own mistake!

PR #1110, opened 2018-12-13
Fixes outdated links

jrvidal:
Broken link hunt! :bow_and_arrow: 

It seems that TodoMVC is not served through https...


alexcrichton:
Oops, thanks!

PR #1118, opened 2018-12-19
Use u32 for indexes and sizes rather than usize

pepyakin:


alexcrichton:
Thanks for this!

I think that we probably will hold off landing this just yet due to it being a breaking change, but it's certainly not impossible for us to break `js-sys` I think we just want to do so relatively rarely.

With that being said I think it's also worthwhile to evaluate `u32` vs `usize` for the future. They're actually the same underlying type on the wasm32 target, so this mostly matters when we future-proof for a possible wasm64 target. With wasm64 though I think that JS will still be bound as it always is, so I don't think it's wortwhile using `usize` here as they can't actually take 64-bit numbers in JS. Now what they  *actually* take is 53-bit integers (what a f64 can represent integral-ly), but we can always add `f64` APIs later if need be.

All that to say, I personally agree with the `u32` instead of `usize` here, and I think we should definitely favor it in the future!

derekdreery:
An alternative strategy is to panic when the user tries to use an index >= 1<<53. This would be future-proof, in case wasm64 happens, but would avoid the UB behavior.

Pedantic me: the behavior isn't undefined, it's just defined to something other than what the user expects.

Pauan:
@derekdreery JS only supports 32-bit indexes:

```js
new Array(Math.pow(2, 32))  // Error
new ArrayBuffer(Math.pow(2, 31))  // Error
new DataView(new ArrayBuffer(Math.pow(2, 31) - 1), Math.pow(2, 31))  // Error
```

That will still be true even with a hypothetical wasm64. So the bindings should be `u32`.

Actually, looking at the behavior of Firefox and Chrome, it seems that `Array` should be `u32`, but `ArrayBuffer` and `DataView` (and probably others) should be `i32` (even though they don't allow for negatives).

I'm not sure if that's correct according to the spec (at a quick glance the ES6 spec says that `u53` should be supported).

pepyakin:
Closing this to clean my PR queue. Feel free to reopen if needed!

PR #1142, opened 2019-01-01
Remove separators that cause numbering jumps

andrewdavidmackenzie:
I have removed the "--------------------" separators as they cause a section number to be skipped and hence incorrect section numbering in the left panel TOC.

Take a look at https://rustwasm.github.io/wasm-bindgen/web-sys/index.html left panel, towards the bottom, and you will see that it's jumps from section 3, to 5, to 7 to 9 - which I am assuming is NOT what you wanted.


alexcrichton:
👍

PR #1144, opened 2019-01-02
Allow using include_str! in typescript_custom_section consts

coolreader18:
As it says in the title. This would be useful for separating type definitions from the code, and you would be able to get syntax highlighting for the typescript in your editor as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ default = ["spans", "std"]
 spans = ["wasm-bindgen-macro/spans"]
 std = []
 serde-serialize = ["serde", "serde_json", "std"]
-nightly = []
+nightly = ["wasm-bindgen-macro/nightly"]
 
 # Whether or not the `#[wasm_bindgen]` macro is strict and generates an error on
 # all unused attributes
--- a/crates/macro-support/Cargo.toml
+++ b/crates/macro-support/Cargo.toml
@@ -18,6 +18,6 @@ strict-macro = []
 [dependencies]
 syn = { version = '0.15.0', features = ['visit'] }
 quote = '0.6'
-proc-macro2 = "0.4.9"
+proc-macro2 = { version = "0.4.9", features = ["nightly"] }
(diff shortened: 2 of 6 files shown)


alexcrichton:
Thanks for this! I fear this is a bit heavyweight though reimplementing the macro here, is there a way we can get rustc to expand the macro for us?

coolreader18:
Possibly? I think it would require typescript custom sections to be specially compiled with the program and actually referenced from whatever code the CLI then calls. Idk, I'm not too familiar with the architecture of this project but I don't think there are any APIs in `proc_macro` that we can call to evaluate the macro.

alexcrichton:
I think this just needs to go through to a custom section though, right? Could the code for the custom section allow further compiler expansion in it?

coolreader18:
Maybe something like a struct with a generic type `T` that holds a `syn::Expr` that can be encoded into `T` by using the Expr somehow? I'm not quite sure how the backend works between the macro and CLI.

coolreader18:
Basically, the attribute macro only has access to the AST of the item, no macros are expanded at all.

coolreader18:
Oh wow, I did not know that you could resolve merge conflicts from the GitHub website.

alexcrichton:
Er sorry so what I mean is that this isn't going to merge as-is. We don't want to get in the business of reimplementing macros. Macros can expand to other macro invocations, so if possible we may just need to tweak the expansion to allow that. Otherwise this may just be out of luck for now.

coolreader18:
I understand that. I think what I might try to implement is a custom type that holds a `syn` node, and in where it implements ToTokens it just injects its node.

coolreader18:
If that doesn't work out, I'm fine with closing this PR.

alexcrichton:
That sounds reasonable to me I think! Feel free to post here with results!

alexcrichton:
Ok I'm gonna close this for now to clear out our PR list a bit, but feel free to resubmit with results!

PR #1147, opened 2019-01-04
Add conversions between typed arrays and Rust

### LINKED ISSUE #811: Support converting js_sys typed arrays to Vec

alexcrichton:
For all typed arrays, this commit adds:

* `TypedArray::view(src: &[Type])`
* `TypedArray::copy_to(&self, dst: &mut [Type])`

The `view` function is unsafe because it doesn't provide any guarantees
about lifetimes or mutability. The `copy_to` function is, however, safe.

Closes #811

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -854,6 +854,11 @@ extern "C" {
     /// typed array from the start of its `ArrayBuffer`.
     #[wasm_bindgen(method, getter, js_name = byteOffset)]
     pub fn byte_offset(this: &Float32Array) -> u32;
+
+    /// The `set()` method stores multiple values in the typed array, reading
+    /// input values from a specified array.
+    #[wasm_bindgen(method)]
+    pub fn set(this: &Float32Array, src: &JsValue, offset: u32);
 }
 
 // Float64Array
@@ -938,6 +943,11 @@ extern "C" {
     /// typed array from the start of its `ArrayBuffer`.
     #[wasm_bindgen(method, getter, js_name = byteOffset)]
     pub fn byte_offset(this: &Float64Array) -> u32;
+
+    /// The `set()` method stores multiple values in the typed array, reading
+    /// input values from a specified array.
+    #[wasm_bindgen(method)]
+    pub fn set(this: &Float64Array, src: &JsValue, offset: u32);
 }
 
 // Function
@@ -1157,6 +1167,11 @@ extern "C" {
     /// typed array from the start of its `ArrayBuffer`.
     #[wasm_bindgen(method, getter, js_name = byteOffset)]
     pub fn byte_offset(this: &Int8Array) -> u32;
+
+    /// The `set()` method stores multiple values in the typed array, reading
+    /// input values from a specified array.
+    #[wasm_bindgen(method)]
+    pub fn set(this: &Int8Array, src: &JsValue, offset: u32);
 }
 
 // Int16Array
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -4742,3 +4787,70 @@ pub fn global() -> Object {
...
+                    &mem.buffer(),
+                    rust.as_ptr() as u32,
+                    rust.len() as u32,
+                )
+            }
+
+            /// Copy the contents of this JS typed array into the destination
+            /// Rust slice.
+            ///
+            /// This function will efficiently copy the memory from a typed
```
chinedufn: Curious, what does "efficiently" mean here?

Thanks!

crates/js-sys/src/lib.rs
```diff
@@ -4742,3 +4787,70 @@ pub fn global() -> Object {
 
     GLOBAL.with(|g| g.clone())
 }
+
+macro_rules! arrays {
+    ($($name:ident: $ty:ident,)*) => ($(
+        impl $name {
+            /// Creates a JS typed array which is a few into wasm's linear
```
chinedufn: Just a minor thing that I noticed -> `few` -> `view`

crates/js-sys/src/lib.rs
alexcrichton: Ah this just means that it uses the `set` API which eventually translates to something like a raw `memcpy`. That's the fastest way (even in Rust) to copy chunks of memory. 

The slow way would be to copy bytes one-by-one in wasm which misses on on things like vectorization and eliding bounds checks

crates/js-sys/src/lib.rs
alexcrichton: Oops, I'll be sure to handle before landing!

crates/js-sys/src/lib.rs
chinedufn: _Gains knowledge_

Cool - thanks!!

grovesNL:
This looks great, thanks! Is there a way to return a single element from the typed array as well? There doesn’t seem to be an indexing getter for typed arrays in js-sys, but this is supported in JavaScript.

Pauan:
@grovesNL Currently you would use [`Reflect.get`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/get), like this:

```rust
let value = js_sys::Reflect::get(typed_array.as_ref(), &JsValue::from_f64(0.0)).unwrap();
```

`Reflect.get` works on pretty much everything: objects, arrays, TypedArray, etc.

It's equivalent to the `foo[bar]` syntax in JS.

----

@alexcrichton @fitzgen Any thoughts on adding a convenience method which basically does the above?

I think it'd be especially nice if the convenience method could avoid the `JsValue::from_f64` boxing (since `f64` can be passed natively to/from JS).

If you're okay with it, I can try sending a PR.

grovesNL:
Oh right, using `Reflect.get` makes sense, thanks! I thought there might be a more direct way to index into these arrays. It could be useful to make a note of it from the typed array documentation somewhere if this is the recommended approach going forward (I'd be happy to send a PR).

It feels slightly confusing now because of the availability of `set` there with no corresponding `get`. I even looked through all the traits expecting to see something like `IndexingGetter<u8>` :)

alexcrichton:
Adding a convenience or a method to just get one item sounds like a good idea to me! @Pauan I'd be down for such a PR

Pauan:
PR made at #1225

PR #1148, opened 2019-01-05
Spawn local

derekdreery:
This is a very simple implementation of `spawn_local`, a function that spawns a rust future using the javascript promise infrastructure as the executor.

### DIFF
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -378,7 +378,14 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
     }
 }
 
-/// Spawns a future.
+/// Converts a Rust `Future` on a local task queue.
+///
+/// The `future` provided must adhere to `'static` because it'll be scheduled
+/// to run in the background and cannot contain any stack references.
+///
+/// # Panics
+///
+/// This function has the same panic behavior as `future_to_promise`.
 pub fn spawn_local<F>(future: F)
 where
     F: Future<Item = (), Error = ()> + 'static,
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -377,3 +377,15 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
         }
     }
 }
+
+/// Spawns a future.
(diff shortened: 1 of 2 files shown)


crates/futures/src/lib.rs
```diff
@@ -377,3 +377,22 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
...
+/// The `future` provided must adhere to `'static` because it'll be scheduled
+/// to run in the background and cannot contain any stack references.
+///
+/// # Panics
+///
+/// This function has the same panic behavior as `future_to_promise`.
+pub fn spawn_local<F>(future: F)
+where
+    F: Future<Item = (), Error = ()> + 'static,
+{
```
lcnr: Is there a reason why only a Future<Item = (), Error = ()> is allowed?
I would personally prefer 

```rust
pub fn spawn_local(future: impl Future + 'static) {
    future_to_promise(
        future
            .map(|_| JsValue::undefined())
            .map_err(|_| JsValue::undefined()),
    );
}
```

derekdreery: I changed this to what you suggest :)

derekdreery:
Are the docs good enough? It might be that the module documentation needs to be changed.

crates/futures/src/lib.rs
Pauan: @lcnr It is standard practice. See [`tokio::run`](https://docs.rs/tokio/0.1.13/tokio/runtime/fn.run.html) and [`stdweb::spawn_local`](https://github.com/koute/stdweb/blob/318a1a026609aac61cd8997fb3b5b53aa2cf1481/src/webcore/promise_future.rs#L112).

The reason this is done is to prevent mistakes, and also to force the user to handle errors (rather than silently ignoring them, which is very bad!)

derekdreery: Happy to do it either way. Maybe we should match other libraries though @lcnr. You can always make your own wrapper

```rust
pub fn spawn_local(future: impl Future + 'static) {
    wasm_bindgen_futures::spawn_local(
        future.map(|_| ()).map_err(|_| ())
    )
}
```

lcnr: I thought about it and agree with @Pauan .

PR #1151, opened 2019-01-06
Support webidl sequence

derekdreery:
Use `js_sys::Array` for parameters and return types of webidl type `sequence<_>`.


derekdreery:
woops

Issue #1038, opened 2018-11-18
Methods that take a WebIDL `sequence` argument are missing


RSSchermer:
Some methods seem to be missing from the `WebGl2RenderingContext`:

- `invalidateFramebuffer`
- `invalidateSubFramebuffer`
- `drawBuffers`
- `transformFeedbackVaryings`
- `getUniformIndices`
- `getActiveUniforms`

What they all seem to have in common is that they all take a WebIDL `sequence` as an argument. I've not checked other APIs, so I'm not sure whether or not this affects other APIs besides the WebGl2 API.

LordPython:
I just noticed this when looking at the SubtleCrypto api, which has the same problem:

- `generateKey`
- `deriveKey`
- `importKey`
- `unwrapKey`

are all missing, and all have `sequence` parameters.

alexcrichton:
Er sorry for taking awhile to get back to this! I believe this is [handled here](https://github.com/rustwasm/wasm-bindgen/blob/7768328fc49962c9996d80a72bc20b16856104ce/crates/webidl/src/idl_type.rs#L545) where we don't handle it right now. I suspect that we probably want to return either an `Array` or some sort of Rust-defined iterator here perhaps. It'd definitely be nice to bind these!

This'll need some work to figure out what the best interface is, an implementation, and then a PR!

derekdreery:
Relevant webidl spec: https://heycam.github.io/webidl/#es-sequence

derekdreery:
I've added a patch to use `Array`, because I think all sequences are converted to arrays in ECMAScript. PR: https://github.com/rustwasm/wasm-bindgen/pull/1152. I'm going to see if I get any errors with it like this.

derekdreery:
Update: I've re-read [the spec](https://heycam.github.io/webidl/#es-sequence) and it seems ambiguous. I think `Array` is a safe option.

Pauan:
Rather than guessing, we should test and see what the browsers actually do.

derekdreery:
I've experimented with DocumentOrShadowRoot elementsFromPoint in the console window (Firefox):

```text
$ temp0 = document.elementsFromPoint(1.0, 1.0);
Array [ html ]
$ temp0 instanceof Array
true
```

alexcrichton:
@derekdreery it may be worthwhile testing a few sequences in argument position too, I think it makes sense that coming out they're an array, but going in it may be anything iterable

[...]

PR #1152, opened 2019-01-06
Use JsValue for webidl sequence.

derekdreery:


alexcrichton:
Thanks for this! I'm not sure though if `js_sys::Array` is the right binding here, it looks like according to WebIDL it may suffice to take/return iterators as well? (I'm not super familiar with many of these APIs that take a sequence)

Do you know if other non-array types are accepted/returned?

derekdreery:
I wasn't sure exactly what the semantics are. It sounds like sequences should be valid, but all the examples use `Array`s. Maybe some experimentation is required.

alexcrichton:
Ah yeah it's probably worth playing around with this a bit and some of the newly exposed API. One option is to just bind this as `JsValue` and let users interpret it though

derekdreery:
I've done some experimentation on browsers in https://github.com/rustwasm/wasm-bindgen/issues/1038

derekdreery:
I've just set it to use JsValue for now.

alexcrichton:
Ok looks good to me! Want to add a small test as well exercising this?

Also mind taking a look @fitzgen as well and confirming you're good w/ this strategy?

fitzgen:
Is there a reason not to use `js_sys::Iterator` instead of `JsValue`?

alexcrichton:
@fitzgen I think an `Array` couldn't be passed in because it's not a superclass, right?

fitzgen:
> @fitzgen I think an `Array` couldn't be passed in because it's not a superclass, right?

I'm not suggesting `Array`, only `js_sys::Iterator` which is a duck-typed thing for any JS object that implements the iterator protocol.

Maybe we are talking past each other here?

fitzgen:
Is this issue specifically for parameters?

I would expect `js_sys::Iterator` for iterable return values, and I guess `&JsValue` for parameters.

Pauan:
@alexcrichton Can't you just use `foo(&x.values())`? And we can provide a similar method on other iterable types. Maybe even backed by a trait.

alexcrichton:
@fitzgen ah yeah sorry I'm thinking about parameters, but I think we still need `Array` on return types as well because there's not a `next` method on the `Array` types in JS, right? Ideally we'd return like a `IntoIterator` of some form I think?

@Pauan it's true yeah `.values()` would work! I think we ideally, though, want to expose the capabilities of web engines which is to accept both iterators and arrays so conversions that aren't necessary don't pull in overhead by accident.

fitzgen:
> @fitzgen ah yeah sorry I'm thinking about parameters, but I think we still need `Array` on return types as well because there's not a `next` method on the `Array` types in JS, right? Ideally we'd return like a `IntoIterator` of some form I think?

D'oh. yeah. I don't think we can rely on iterables being arrays though. I guess the most correct thing would be something like `-> impl IntoIterator<Item = JsValue>` or something like that...

fitzgen:
That is pretty definitive. Ignore my comment about "I don't think we can rely on..."

alexcrichton:
Ok I think we're probably good to go in this case with argument position as `JsValue` and return position as `js_sys::Array`. @derekdreery want to update the PR to return `Array` and it should be good to go then!

derekdreery:
Ok sounds good!

derekdreery:
Done. Also rebased and squashed.

PR #1159, opened 2019-01-08
Convert all examples to using wasm-pack

### LINKED ISSUE #1120: Alternative to build.sh?
I noticed many of the examples rely on `build.sh`. I assume most users have some shell installed (even on Windows), but I think it would be better to avoid this dependency if possible.

Is there a portable alternative that could be used instead? For example, using some cargo runner (I think a few exist?) or simply providing an equivalent `build.bat` for Windows might be helpful to new users following the guide.

This is fairly minor, but thought I would mention it here as people will probably be looking to this guide for best practices.

alexcrichton:
This commit migrates all our examples to using `wasm-pack build` to
compile their code and run `wasm-bindgen`. This should make it a bit
easier to understand the examples as there's less to follow during the
build step.

Webpack projects are all using `@wasm-tool/wasm-pack-plugin` as well so
the build step is simple `npm run serve`. Other examples which retain
`build.sh` are just using `wasm-pack build` now

Closes #1120


alexcrichton:
Note that this souldn't be merged yet because I'd still like to wait on the next version of wasm-pack being published. I wanted to try to iron out the kinks in CI though!

.travis.yml
```diff
@@ -93,14 +93,16 @@ matrix:
         - *INSTALL_NODE_VIA_NVM
         - *INSTALL_AWS
         - npm install
+        #- curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+        - cargo install --git https://github.com/rustwasm/wasm-pack
```
fitzgen: This should be

```
curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f
```

examples/add/package.json
```diff
@@ -1,9 +1,10 @@
 {
   "scripts": {
-    "build": "webpack",
-    "serve": "webpack-dev-server"
+    "build": "webpack -p",
+    "serve": "webpack-dev-server -p"
   },
   "devDependencies": {
+    "@wasm-tool/wasm-pack-plugin": "0.2.0",
```
xtuc: I published `@wasm-tool/wasm-pack-plugin@0.2.1` in the meantime, could you please update it? or maybe the bot will update it after?

examples/add/package.json
```diff
@@ -1,9 +1,10 @@
 {
   "scripts": {
-    "build": "webpack",
-    "serve": "webpack-dev-server"
+    "build": "webpack -p",
+    "serve": "webpack-dev-server -p"
```
xtuc: I believe this is Webpack's default, we can omit it but I also like that's explicit.

PR #1161, opened 2019-01-09
Better output from `impl Debug for JsValue`.

derekdreery:
We can use `JSON.stringify` to get a better debug representation of a value. It's slow, but I think there's an acceptance that Debug (unlike maybe Display) is allowed to be slow to write better strings, given that it's used mostly before deployment.

There's also a new method for testing whether something is an array in javascript. I can remove this if you don't think it's useful.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -377,7 +377,7 @@ impl<'a> Context<'a> {
                         }
                         if (type == 'function') {
                             const name = val.name;
-                            if (typeof name == 'string') {
+                            if (typeof name == 'string' && name.length > 0) {
                                 return `Function(${name})`;
                             } else {
                                 return 'Function';
@@ -391,16 +391,16 @@ impl<'a> Context<'a> {
                                 debug += debug_str(val[0]);
                             }
                             for(let i = 1; i < length; i++) {
-                                debug += debug_str(val[i]) + ', ';
+                                debug += ', ' + debug_str(val[i]);
                             }
                             debug += ']';
                             return debug;
                         }
                         // Test for built-in
-                        const builtInMatches = /\\[object ([^])+\\]/.exec(toString.call(val));
+                        const builtInMatches = /\\[object ([^\\]]+)\\]/.exec(toString.call(val));
                         let className;
-                        if (builtInMatches.len > 0) {
-                            className = builtInMatches[0];
+                        if (builtInMatches.length > 1) {
+                            className = builtInMatches[1];
                         } else {
                             // Failed to match the standard '[object ClassName]'
                             return toString.call(val);
@@ -414,6 +414,7 @@ impl<'a> Context<'a> {
                             } catch (_) {
                                 return 'Object';
                             }
+                        // TODO we could test for more things here, like `Set`s and `Map`s.
                         } else {
                             return className;
                         }
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -369,7 +369,7 @@ impl<'a> Context<'a> {
                         if (type == 'symbol') {
                             const description = val.description;
                             if (description == null) {
-                                return 'Symbol()';
+                                return 'Symbol';
                             } else {
                                 return 'Symbol(' + description + ')';
                             }
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -350,6 +350,23 @@ impl<'a> Context<'a> {
             ))
         })?;
 
+        self.bind("__wbindgen_to_string", &|me| {
+            me.expose_pass_string_to_wasm()?;
+            me.expose_get_object();
+            me.expose_uint32_memory();
+            Ok(String::from(
+                "
+                function(i, len_ptr) {
+                    let toString = getObject(i).toString();
+                    if (typeof(toString) !== 'string') return 0;
+                    const ptr = passStringToWasm(toString);
+                    getUint32Memory()[len_ptr / 4] = WASM_VECTOR_LEN;
+                    return ptr;
+                }
+                ",
+            ))
+        })?;
+
         self.bind("__wbindgen_cb_drop", &|me| {
             me.expose_drop_ref();
             Ok(String::from(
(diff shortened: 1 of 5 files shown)


derekdreery:
This should have some tests I think (it doesn't yet).

alexcrichton:
Thanks for this! 

Reading this over, I wonder if we perhaps want to actually just move the entire `Debug` implementation into JS? The code size footprint in Rust is actually pretty large we've seen in the past with float parsing and whatnot, and if we're hopping back and forth to JS to test types and whatnot it may actually just be best to have stringification happen directly in JS.

derekdreery:
Good idea! I'll experiment.

derekdreery:
I've done an implementation in JS. See what you think.

There's still a TODO around no-std. This patch actually regresses to doing nothing. Can we do something to get the string in the case that we don't have an allocator?

crates/cli-support/src/js/mod.rs
```diff
@@ -336,6 +348,74 @@ impl<'a> Context<'a> {
...
         })?;
 
+        self.bind("__wbindgen_debug_string", &|me| {
+            me.expose_pass_string_to_wasm()?;
+            me.expose_get_object();
+            me.expose_uint32_memory();
+            Ok(String::from(
+                "
+                function(i, len_ptr) {
+                    const debug_str = val => {
```
fitzgen: I tried to come up with something that supported as rich of debug string output as this but was a little more terse. Not sure I really ended up on an improvement over what you have here, but for posterity:

```js
function debug(val) {
  switch(typeof val) {
    case "undefined":
    case "number":
    case "boolean":
    case "symbol":
      return String(val);
    case "function":
      return `Function(${val.name || ""})`;
    case "object":
      if (val === null) {
        return "null";
      }
      const s = Object.prototype.toString.call(val);
      if (s === "[object Object]" || s === "[object Array]") {
        try {
          return JSON.stringify(val);
        } catch (_) {}
      }
      return s;
    default:
      return Object.prototype.toString.call(val);
  }
}
```

src/lib.rs
```diff
@@ -298,11 +298,32 @@ impl JsValue {
         unsafe { __wbindgen_is_object(self.idx) == 1 }
     }
 
+    /// Tests whether `Array.isArray(self)`.
+    #[inline]
+    pub fn is_array(&self) -> bool {
```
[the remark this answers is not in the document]
derekdreery: Yeah if it were to be included it should be a separate PR. I'll remove.

derekdreery:
> I don't think there is an easy way, unfortunately.

I think I probably need to restore the old impl then for the no_std case.

crates/cli-support/src/js/mod.rs
derekdreery: This is cool. I did think about doing a big switch. If JSON.stringify can throw, I should catch it and just fall through like you do. If we can capture the function name that would be cool also. I'll try to merge the 2 impls.

Also since I'm assuming there's no stability guarantees on the output here we can always refine later.

 - [ ] get name of the function if possible
 - [ ] if JSON.stringify can throw, catch and ignore any exceptions.

alexcrichton:
I think it's fine to in general not worry much about the `no_std` use case, I don't think that it's really operational at all and likely won't make its way to the next major release of wasm-bindgen. (I personally think I was mistaken to add it originally!)

derekdreery:
I think this is ready for review now. I've tried to cover all the cases in the tests. I found quite a few problems, which I guess is why tests!!!!

derekdreery:
Not sure where the failure's from.

fitzgen:
https://travis-ci.com/rustwasm/wasm-bindgen/jobs/169969977#L894-L906

```
---- wasm::api::debug_output output ----
    JS exception that was thrown:
        TypeError: getObject(...).values is not a function
            at module.exports.__wbg_values_c7388d2dffc70125 (/home/travis/build/rustwasm/wasm-bindgen/target/wasm32-unknown-unknown/wbg-tmp/wasm-bindgen-test.js:4348:42)
            at js_sys::Array::values::h413e65c06a2161cc (wasm-function[8094]:101)
            at wasm::api::debug_output::h53891ce08362a31e (wasm-function[2663]:96)
            at core::ops::function::FnOnce::call_once::h155aa4ffc67eec53 (wasm-function[1035]:22)
            at wasm_bindgen_test::__rt::Context::execute_sync::_$u7b$$u7b$closure$u7d$$u7d$::h3886c0814e7f4942 (wasm-function[3976]:22)
            at _$LT$futures..future..lazy..Lazy$LT$F$C$$u20$R$GT$$GT$::get::hc1c1b592d7026e96 (wasm-function[654]:395)
            at _$LT$futures..future..lazy..Lazy$LT$F$C$$u20$R$GT$$u20$as$u20$futures..future..Future$GT$::poll::hc9d74cbdf88e301e (wasm-function[859]:38)
            at _$LT$wasm_bindgen_test..__rt..TestFuture$LT$F$GT$$u20$as$u20$futures..future..Future$GT$::poll::_$u7b$$u7b$closure$u7d$$u7d$::_$u7b$$u7b$closure$u7d$$u7d$::h89214d033d0eec01 (wasm-function[4309]:71)
            at wasm_bindgen::convert::closures::invoke0_mut::h34f83a9aa7fb5334 (wasm-function[7341]:166)
            at Function.cbarg0 (/home/travis/build/rustwasm/wasm-bindgen/target/wasm32-unknown-unknown/wbg-tmp/wasm-bindgen-test.js:4275:25)
```

I believe that node doesn't support Array.prototype.values yet, so maybe that is what is going on here.

crates/cli-support/src/js/mod.rs
```diff
@@ -299,6 +299,18 @@ impl<'a> Context<'a> {
             ))
         })?;
 
+        self.bind("__wbindgen_is_array", &|me| {
```
fitzgen: Nitpick: remove this

src/lib.rs
```diff
@@ -474,9 +489,11 @@ externs! {
     fn __wbindgen_symbol_new(ptr: *const u8, len: usize) -> u32;
     fn __wbindgen_is_symbol(idx: u32) -> u32;
     fn __wbindgen_is_object(idx: u32) -> u32;
+    fn __wbindgen_is_array(idx: u32) -> u32;
```
fitzgen: Nitpick: remove this

derekdreery: yep sorry missed this last time.

PR #1162, opened 2019-01-10
Simply remove the web-sys bindings 'tested' list

rylev:
As discussed in the wasm wg meeting, this list is pretty confusing as it _seems_ to imply that the APIs that are not checked off are not implemented (at least fully). This is not the case and only refers to which have been tested. It was suggested to remove this list.


rylev:
FYI: CI is broken due to this issue rust-lang/rust#57488

PR #1165, opened 2019-01-11
Exclude the `typescript` crate from the workspace

alexcrichton:
This'll help get `cargo build --all` and similar commands working. While
not critical at all, I execute it from time to time and it's a bummer
when it doesn't work!


alexcrichton:
Perhaps yeah, let's start with this and see how far we get!

PR #1169, opened 2019-01-12
fix spawn_local

lcnr:
`spawn_local` caused `uncaught exception: undefined` on firefox when supplied with an error variant. This should now be fixed.

# Examples

```rust
spawn_local(future::err::<(), ()>(()));
```

I don't know if and how I should write a test for this.

### DIFF
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -113,6 +113,7 @@ use std::sync::Arc;
 
 use futures::executor::{self, Notify, Spawn};
 use futures::prelude::*;
+use futures::future;
 use futures::sync::oneshot;
 use js_sys::{Function, Promise};
 use wasm_bindgen::prelude::*;
@@ -393,6 +394,6 @@ where
     future_to_promise(
         future
             .map(|()| JsValue::undefined())
-            .map_err(|()| JsValue::undefined()),
+            .or_else(|()| future::ok::<JsValue, JsValue>(JsValue::undefined())),
     );
 }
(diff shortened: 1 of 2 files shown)


derekdreery:
Oops my bad.

If you were to add a test (good to make sure this doesn't happen again), the place to do it is after [this test](https://github.com/rustwasm/wasm-bindgen/pull/1148/files#diff-8e76276ff1bef2387f2fb0f55285c837R72), or possibly inside it.

lcnr:
I might require some help with testing this. While I do get the following Warning when testing
> (node:14013) UnhandledPromiseRejectionWarning: undefined
(node:14013) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:14013) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js processwith a non-zero exit code.

the tests are successful and I don't know how to add an eventlistener for `unhandledrejection` in Node.js, as the window object does not exist.

alexcrichton:
@lcnr hm I'm not entirely sure what those warnings are, but want to commit the test here and we can help through CI logs?

lcnr:
@alexcrichton I did commit the test without reverting the fix, so it runs successfully without this warning.

alexcrichton:
Thanks!

I think it's ok for us to manually inspect output for now to make sure this doesn't regress

PR #1175, opened 2019-01-14
Fix an issue where closure rewriting required class internals 

### LINKED ISSUE #1174: Returning a struct from a closure results in "__wrap is not a function" on that struct
I'm on wasm-bindgen@0.2.31 and getting "__wrap is not a function" when calling a closure which returns a struct that is not used anywhere else. 

Here's the smallest example of the bug, I could make: 

```rust
type ClosureType = FnMut() -> BadStruct;

#[wasm_bindgen]
pub struct ClosureHandle(Closure<ClosureType>);

#[wasm_bindgen]
pub struct BadStruct {}

#[wasm_bindgen]
pub fn closure_returner() -> Result<Object, JsValue> {
    let o = Object::new();

    let some_fn = Closure::wrap(Box::new(move || BadStruct {}) as Box<ClosureType>);
    Reflect::set(
        &o,
        &JsValue::from("someKey"),
        &some_fn.as_ref().unchecked_ref(),
    )
    .map_err(|_| "Unable to set property")?;
    Reflect::set(
        &o,
        &JsValue::from("handle"),
        &JsValue::from(ClosureHandle(some_fn)),
    )
    .map_err(|_| "Unable to set closure handle")?;

    Ok(o)
}
```

Typescript:
```typescript
import("../../lib-web/pkg").then(module => {
    const c = module.closure_returner();
    c.someKey(); // Error occurs here
});
```

Currently this works as a workaround:
```
#[wasm_bindgen]
pub fn _dont_call() -> BadStruct {
    unimplemented!();
}
```

Edit: Related to #949, but not exactly the same.

alexcrichton:
Surfaced through previous sanity-checking commits, this reorders some
internal operations to...

Closes #1174

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -21,6 +21,7 @@ mod closures;
 
 pub struct Context<'a> {
     pub globals: String,
+    pub globals_written: bool,
     pub imports: String,
     pub imports_post: String,
     pub footer: String,
@@ -119,6 +120,11 @@ const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
 const INITIAL_HEAP_OFFSET: usize = 32;
 
 impl<'a> Context<'a> {
+    fn should_write_global(&mut self, name: &'static str) -> bool {
+        assert!(!self.globals_written);
+        self.exposed_globals.insert(name)
+    }
+
     fn export(&mut self, name: &str, contents: &str, comments: Option<String>) {
         let contents = contents.trim();
         if let Some(ref c) = comments {
@@ -477,6 +483,10 @@ impl<'a> Context<'a> {
 
         self.rewrite_imports(module_name);
         self.update_producers_section();
+
+        // Cause any future calls to `should_write_global` to panic, making sure
+        // we don't ask for items which we can no longer emit.
+        self.globals_written = true;
 
         let mut js = if self.config.threads.is_some() {
             // TODO: It's not clear right now how to best use threads with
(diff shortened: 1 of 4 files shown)


alexcrichton:
@fitzgen I'm curious if you have thoughts on this, we've had a good number of bugs about the ordering of operations in `finalize` in wasm-bindgen, I feel like we're constantly playing some sort of weird code golf to figure out the right ordering.

Do you have thoughts on how we might better structure this and/or fix this for good in terms of JS generation?

crates/cli-support/src/js/mod.rs
```diff
@@ -478,6 +484,10 @@ impl<'a> Context<'a> {
         self.rewrite_imports(module_name);
         self.update_producers_section();
 
+        // Cause any future calls to `should_write_global` to panic, making sure
+        // we don't ask for items which we can no longer emit.
+        self.globals_written = true;
```
fitzgen: Alternatively, `exposed_globals` could be an `Option<...>` and then we can have `should_write_global` do `self.exposed_globals.as_mut().unwrap().insert(name)` and then here do `self.exposed_globals.take().unwrap()`.

This is equivalent to having a bool flag (which becomes the option discriminant in this scheme) but it is *impossible* to forget to check the flag in `should_write_global` or anything else that wants to touch `exposed_globals` thanks to the type system.

crates/cli-support/src/js/mod.rs
```diff
@@ -56,7 +56,7 @@ pub struct Context<'a> {
     /// wasm-bindgen emits.
     pub direct_imports: HashMap<&'a str, (&'a str, &'a str)>,
 
-    pub exported_classes: HashMap<String, ExportedClass>,
+    pub exported_classes: Option<HashMap<String, ExportedClass>>,
```
fitzgen: Ah ok, this second commit is exactly the pattern that I was suggesting for the first one :-p

crates/cli-support/src/js/mod.rs
alexcrichton: 👍

Issue #1005, opened 2018-11-01
webidl: Add a method of whitelisting an API as not mutating its byte slice argument


alexcrichton:
Spawned off from https://github.com/rustwasm/wasm-bindgen/issues/978, we currently [pessmize all slices to be mutable](https://github.com/rustwasm/wasm-bindgen/blob/0d5f5147093746ec209277fc0c879329c2c13188/crates/webidl/src/util.rs#L75) when passed to JS. This conservative decision is required for correctness, but we typically know on a one-off basis that the mutable borrow isn't actually necessary.

We should add a specific method to whitelist particular APIs and switch the to using an immutable slice instead of a mutable one!

RReverser:
I've just found that `TextDecoder::decode_with_u8_array{_and_options}` requires a mutable slice. Sounds relevant to this issue? I'll be happy to send a PR, but what do I need to do to whitelist these methods?

alexcrichton:
@RReverser just to confirm, but do you mean it *doesn't* need a mutable slice? (it looks like it already takes a mutable slice)

RReverser:
Yeah, I'm saying that I found that it currently requires mutable slice in the Rust API, but it shouldn't.

alexcrichton:
Ah yes, then indeed this is the right issue! Currently a whitelist isn't implemented, though, that would need to be done first. (that's what this issue is about)

RReverser:
Ah sorry, I misread, I thought the issue was merely about finding and marking methods that need to accept immutable slices.

chinedufn:
@alexcrichton where might one start looking in order to implement the whitelist? Any tips / guidance for diving in here?

---

Background:

You end up running into minor inconveniences very frequently when working on WebGL applications since you're using a bunch of these APIs that require mutable slices very frequently.

Nothing too, too painful - just inconvenient.

---

So if this is do-able for a beginner contributor I'd love to take a look!

Thanks!

alexcrichton:
Sure yeah! Here's at least one way I think this could be implemented today:

* First up, probably some cleanup needs to be done to `enum IdlType` because everything becomes mutable slices. For example `Uint8ArrayMut` probably isn't needed any more as opposed to `Uint8Array`
* Next, We'll want to add a variant that represents immutable slices. Perhaps booleans on the existing arrays, or perhaps new variants. After an enum variant is added appropriate places are fixed in compilation to ensure shared slices instead of mutable slices are used in Rust.
[...]

PR #1177, opened 2019-01-14
[WIP] whitelist for slice args that do not need to be mutable

chinedufn:
**Work in progress - Not completed**

@alexcrichton thought I'd open an early PR and ask questions as I go along. Should take me some days to get finished as I'll only be making light touches here and there (busy week ahead).

First question: How should I go about writing tests as I go? I started poking around the codebase to better understand what was going on and where things happen but figured it'd be faster to just come and ask for more guidance.

Cheers!

---

_If opening an unfinished PR isn't appropriate in `wasm-bindgen` just let me know!_

---

https://github.com/rustwasm/wasm-bindgen/issues/1005


alexcrichton:
Sounds good to me! Looking good so far :)

chinedufn:
Sweet thanks!

> First question: How should I go about writing tests as I go? 

So I poked around the code base and read up on the docs and it looked like `web-sys` was to place to add tests. Onwards!

crates/webidl-tests/lib.rs
```diff
@@ -1 +1,3 @@
 // intentionally left blank
+
+// QUESTION FOR REVIEWER: WHY? (I'll include this context in the comment)
```
alexcrichton: Ah this is just because Cargo needs to have *something* like a library or a binary, so this is just a library to pacify cargo, but the library isn't actually used by any of the tests currently

chinedufn: Oh thanks! Even though there's a `main.rs` file?

chinedufn:
Err tried changing branches and changing back since I've run into GitHub diff issues before where that fixed it...

Mot sure how this got closed but I can't re-open it so.. opening a new PR sorry..

PR #1193, opened 2019-01-17
ci: build docs on stable

fitzgen:
This avoids https://github.com/rust-lang/rust/issues/57628 and also generally makes sense since wasm-bindgen buils on stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -223,7 +223,7 @@ matrix:
       if: branch = master
 
     # Build API documentation
-    - rust: nightly
+    - rust: stable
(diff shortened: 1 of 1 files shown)


alexcrichton:
There is one entry which requies nightly (documenting the nightly features), but perhaps we could just leave that out for now?

fitzgen:
From CI logs:

> The command "cargo doc --no-deps --features 'nightly serde-serialize'" exited with 0.

/me shrugs

Not sure it is worth merging this or not...

alexcrichton:
wfm

PR #1201, opened 2019-01-23
Add '| undefined' to optional types in generated .d.ts

rhysd:
Hi,

```rust
#[wasm_bindgen]
struct X {
    pub foo: Option<i32>,
}

#[wasm_bindgen]
impl X {
    pub fn f(x: Option<i32>) -> Option<u32> {
        x.map(|i| i as u32)
    }
}
```

These option values generate TS type definition in `.d.ts` file as follows

```typescript
export class X {
free(): void;
foo: number;

static  f(arg0: number): number | undefined;

}
```

However, by this type definition, `foo` and `arg0`  cannot accept `undefined` though the original values are `Option` when using `strictNullChecks` compiler flag of `tsc`. Passing `undefined` to the `arg0` causes TypeScript compilation error. And the compiler thinks accessing to `foo` always returns non-`undefined` value so compiler cannot check the case where `foo` is `undefined`.

Expected type definition is:

```typescript
export class X {
free(): void;
foo: number | undefined;

static  f(arg0: number | undefined): number | undefined;

}
```

`number | undefined` means a value which is `number` or `undefined`. It should make `tsc` happy.

This patch modifies TS type definition file generation to generate the types `T | undefind` for `Option<T>` values.

### DIFF
--- a/crates/cli-support/src/js/js2rust.rs
+++ b/crates/cli-support/src/js/js2rust.rs
@@ -208,7 +208,8 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
         if optional {
             if arg.is_wasm_native() {
                 self.cx.expose_is_like_none();
-                self.js_arguments.push((name.clone(), "number".to_string()));
+                self.js_arguments
+                    .push((name.clone(), "number | undefined".to_string()));
 
                 if self.cx.config.debug {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

Issue #1200, opened 2019-01-22
Add #[allow(clippy::*)] to all our macro-generated Rust code


fitzgen:
### What

Using `wasm-bindgen` with `clippy` results in warnings like those reported in https://github.com/rustwasm/wasm-bindgen/issues/1112 and like:

```
error: calls to `std::mem::drop` with a reference instead of an owned value. Dropping a reference does nothing.
 --> src/change_list.rs:9:5
  |
9 |     #[wasm_bindgen]
  |     ^^^^^^^^^^^^^^^
  |
  = note: #[deny(clippy::drop_ref)] on by default
  = note: argument has type &wasm_bindgen::JsValue
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_ref
```

You can test it out by installing clippy:

```
rustup self update
rustup component add clippy
```

and then running clippy on `wasm-bindgen/examples/request-animation-frame`:

```
cd wasm-bindgen/examples/request-animation-frame/
cargo clippy
```

This should print out warnings like

```
warning: this function has too many arguments (8/7)
    --> crates/js-sys/src/lib.rs:3798:15
     |
3798 | #[wasm_bindgen]
     |               ^
     |
     = note: #[warn(clippy::too_many_arguments)] on by default
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments
```

We would ideally like there to be zero clippy warnings in code generated by the `#[wasm-bindgen]` macro. (Note that tehre are unrelated clippy warnings in wasm-bindgen's general code base, which we are less interested in. The goal is to support folks using clippy on projects that use wasm-bindgen.)

### How to fix

In all the top-level items we create with the `quote! { ... }` macro in `wasm-bindgen/crates/backend/src/codegen.rs`, we need to add `#[allow(clippy::*)]` to the top-level item (function or static, etc...).

T5uku5hi:
I’d like to pick this issue up!

fitzgen:
Let me know if you have any questions!

T5uku5hi:
Thank you so much!

I tried to fix but I have a problem.
Please read my PR.

Mitchell0069:
[...]

PR #1207, opened 2019-01-25
Add #[allow(clippy::*)] to the top-level items in codegen.rs

T5uku5hi:
This PR it to fix issue #1200.

I added `#[allow(clippy::*)]` to the top-level items in `codegen.rs`, but I have a problem 😣
Running clippy after fixing, warnings remain print out.
I think I misunderstand how to fix.
Please tell me the correct fixing.


fitzgen:
Hi @T5uku5hi, it looks like you've introduced some syntax errors in the macro-expanded code somewhere: https://travis-ci.com/rustwasm/wasm-bindgen/jobs/172934004#L585

Unfortunately, there isn't a super great way to debug syntax errors within macro-expanded code right now. The two best ways to debug this that I know of is to comment out half of your changes, test to see if the syntax error is there or not. Keep doing this and use a binary search to try and find the code that introduced the syntax errors.

>  Running clippy after fixing, warnings remain print out.

You're still seeing *clippy* warnings with this patch, and not the syntax errors that CI has? If so, can you share the output?

T5uku5hi:
@fitzgen 
Thank you for your comment.
I fixed the code too much at once.  I'm really sorry 😣

> You're still seeing clippy warnings with this patch, and not the syntax errors that CI has? If so, can you share the output?

I got the same syntax error as the CI output result.
Here is the output.
[errors_commit_e19306c.txt](https://github.com/rustwasm/wasm-bindgen/files/2798526/errors_commit_e19306c.txt)

----

I would like to consult you about next steps.
Here is the output result before my fixing.
[errors_before_add_[#[allow(clippy)]].txt](https://github.com/rustwasm/wasm-bindgen/files/2798527/errors_before_add_.allow.clippy.txt)

At first, I'll focus on fixing to remove this warning according to your advice.
```
warning: this function has too many arguments (8/7)
    --> crates/js-sys/src/lib.rs:3832:15
     |
3832 | #[wasm_bindgen]
     |               ^
     |
     = note: #[warn(clippy::too_many_arguments)] on by default
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments
```
Is this next step correct?

fitzgen:
If you want to try and pinpoint a single lint warning, https://github.com/dtolnay/cargo-expand might help. But I'm not 100% sure, since debugging proc-macro output is a bit suboptimal currently.

I suggest adding one `#[allow(clippy::*)]` at a time, and verifying that the example project builds after each one. Eventually all the `#[allow(...)]`s will be in place and you can push the code to this PR :)

T5uku5hi:
@fitzgen 
I tried to add `#[allow(clippy::*)]` to lines as much as possible.
And then, I found error messages changed when I added `#[allow(clippy::*)]` to 902 line.
　
**Before**
[errors_before_add_allow_clippy.txt](https://github.com/rustwasm/wasm-bindgen/files/2825564/errors_before_add_allow_clippy.txt)
**After**
[errors_after_add_allow_clippy.txt](https://github.com/rustwasm/wasm-bindgen/files/2825566/errors_after_add_allow_clippy.txt)
　
I thought it seems a big step in fixing this issue, but I don't know what to do next 😣
Please advice me.

fitzgen:
Are these errors from when you run `cargo clippy` in the example project?

limira:
> Are these errors from when you run `cargo clippy` in the example project?

I think the answer is '*yes*'. Here is an excerpt from `errors_after_add_allow_clippy.txt`
```
request-animation-frame $ cargo clippy
   Compiling wasm-bindgen-backend v0.2.33 (/Users/misakimakino/Works/clones/fork-wasm-bindgen/wasm-bindgen/crates/backend)
   Compiling wasm-bindgen-macro-support v0.2.33 (/Users/misakimakino/Works/clones/fork-wasm-bindgen/wasm-bindgen/crates/macro-support)
   Compiling wasm-bindgen-webidl v0.2.27 (/Users/misakimakino/Works/clones/fork-wasm-bindgen/wasm-bindgen/crates/webidl)
   Compiling wasm-bindgen-macro v0.2.33 (/Users/misakimakino/Works/clones/fork-wasm-bindgen/wasm-bindgen/crates/macro)
   Compiling web-sys v0.3.10 (/Users/misakimakino/Works/clones/fork-wasm-bindgen/wasm-bindgen/crates/web-sys)
    Checking wasm-bindgen v0.2.33 (/Users/misakimakino/Works/clones/fork-wasm-bindgen/wasm-bindgen)
```

T5uku5hi:
@fitzgen 
Yes. 
When I run `cargo clippy` in `request-animation-frame dir`, these error occurred

crates/backend/src/codegen.rs
```diff
@@ -899,6 +899,7 @@ impl TryToTokens for ast::ImportFunction {
             abi_arguments.push(quote! { #exn_data_ptr: *mut u32 });
             convert_ret = quote! { Ok(#convert_ret) };
             exceptional_ret = quote! {
+                #[allow(clippy::*)]
```
fitzgen: Ah I see -- we should only add `#[allow(clippy::*)]` to things that are top-level items in the expanded code, not anything that is at the top of a `quote!` macro invocation, since it might take multiple quotes to build up a top-level item that we are emitting.

A good rule of thumb is that we should add it on each `fn` we define.

Does that make sense?

T5uku5hi: Yes!!  I understand you said.
I tried to add `#[allow(clippy::*)] ` to things that are top-level items _at first_,
but this code occurred another error like this.

```
error: expected one of `(`, `)`, `,`, or `=`, found `::`
  --> crates/backend/src/codegen.rs:15:15
   |
15 | #[allow(clippy::*)]
   |               ^^ expected one of `(`, `)`, `,`, or `=` here

error: unexpected token: `)`
  --> crates/backend/src/codegen.rs:15:18
   |
15 | #[allow(clippy::*)]
   |                  ^ unexpected token after this

```
　
I found  that  I should use `#[allow(clippy::drop_ref)]` instead of `#[allow(clippy::*)]` when I read codes written by others 🧐💡
Finally, I fixed it.

limira: @T5uku5hi, did you try `#![allow(clippy::all)]`?

T5uku5hi: @limira 
Oh! I didn't try `all` instead of `*`.
I tried it just now and I well did it!

I'll add `#[allow(clippy::all)]` in all the top-level items.
Thank you for your kindness 😊

T5uku5hi:
@fitzgen 
I added `#[allow(clippy::all)]` to the top-level items in `codegen.rs`.

crates/backend/src/codegen.rs
fitzgen: This is my bad, sorry!

For some reason I thought the syntax was `allow(clippy::*)` and it is actually `allow(clippy::all)` as @limira points out. Sorry about that!

PR #1214, opened 2019-01-28
Support `Option` with custom enums in JS

### LINKED ISSUE #1198: Option<enum> returns undefined
Hello! I have a reproducer for this bug here: https://github.com/attente/wasm-bindgen-option

Basically:

```
use wasm_bindgen::prelude::*;

// doesn't work :(
#[wasm_bindgen]
pub fn foo() -> Option<Foo> {
    Some(Foo::Foo)
}

// works!
#[wasm_bindgen]
pub fn bar() -> Foo {
    Foo::Bar
}

// works!
#[wasm_bindgen]
pub fn baz() -> Option<i32> {
    Some(3)
}

pub enum Foo {
    Foo = 1,
    Bar = 2,
}

impl wasm_bindgen::describe::WasmDescribe for Foo {
    fn describe() {
        wasm_bindgen::describe::inform(wasm_bindgen::describe::I32)
    }
}

impl wasm_bindgen::convert::IntoWasmAbi for Foo {
    type Abi = i32;

    fn into_abi(self, _extra: &mut dyn wasm_bindgen::convert::Stack) -> Self::Abi {
        self as Self::Abi
    }
}

impl wasm_bindgen::convert::OptionIntoWasmAbi for Foo {
    fn none() -> Self::Abi {
        0
    }
}
```

Returning an `enum` that implements`wasm_bindgen::convert::OptionIntoWasmAbi` works. (`bar()`)
Returning an `Option<i32>` works. (`baz()`)
But returning `Option<enum: wasm_bindgen::convert::OptionIntoWasmAbi>` doesn't work. (`foo()`)

The output I got from the reproducer is:

```
Some(Foo::Foo) = undefined
Foo::Bar = 2
3 = 3
```

alexcrichton:
Find a hole automatically to use a sentinel value for `None`, and then
just wire everything up!

Closes #1198

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -953,6 +953,12 @@ impl MacroParse<()> for syn::ItemEnum {
             _ => bail_span!(self, "only public enums are allowed with #[wasm_bindgen]"),
         }
 
+        if self.variants.len() == 0 {
+            bail_span!(self, "cannot export empty enums to JS");
+        }
+
+        let has_discriminant = self.variants[0].discriminant.is_some();
+
         let variants = self
             .variants
             .iter()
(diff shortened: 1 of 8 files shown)


crates/macro-support/src/parser.rs
```diff
@@ -970,6 +984,8 @@ impl MacroParse<()> for syn::ItemEnum {
                             lit: syn::Lit::Int(ref int_lit),
                         }),
                     )) => {
+                        if !has_discriminant {
+                            }
```
fitzgen: Some strange formatting going on here...

PR #1219, opened 2019-01-31
Add the `UnwrapThrowExt<T>` trait

fitzgen:
An extension trait for `Option<T>` and `Result<T, E>` for unwraping the `T` value, or throwing a JS error if it is not available.

These methods should have a smaller code size footprint than the normal `Option::unwrap` and `Option::expect` methods, but they are specific to working with wasm and JS.

On non-wasm32 targets, defaults to the normal unwrap/expect calls.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,10 +36,11 @@ macro_rules! if_std {
 /// use wasm_bindgen::prelude::*;
 /// ```
 pub mod prelude {
-    pub use wasm_bindgen_macro::wasm_bindgen;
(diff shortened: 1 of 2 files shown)


fitzgen:
@alexcrichton any idea what's up with the gc tests? https://travis-ci.com/rustwasm/wasm-bindgen/jobs/174516547#L748-L784

alexcrichton:
Oh that's https://github.com/paritytech/parity-wasm/pull/264, nothing to worry about there!

PR #1223, opened 2019-02-01
Print error messages to the console in the webgl example

LesleyLai:
Make it a bit easier for debugging shaders.

### DIFF
--- a/examples/webgl/Cargo.toml
+++ b/examples/webgl/Cargo.toml
@@ -12,8 +12,9 @@ js-sys = "0.3.10"
 wasm-bindgen = "0.2.33"
 
 [dependencies.web-sys]
-version = "0.3.4"
+version = "0.3.10"
 features = [
+  'console',
   'Document',
   'Element',
   'HtmlCanvasElement',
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Out of curiosity, was the thrown exception not showing up in the console? That was the intention at least, and it seems bad if that's not the case!

LesleyLai:
> 
> 
> Thanks! Out of curiosity, was the thrown exception not showing up in the console? That was the intention at least, and it seems bad if that's not the case!

No, at least in my case, nothing shows up in the console if the shaders cannot compile.

alexcrichton:
Hm ok that sounds worrisome! Mind digging in a bit to see why the exceptions aren't showing up in the logs? This is the same pattern we have for a number of our other examples and it'd be a bummer if we needed to change those as well. I wonder if this is a bug in something else by accident or something like that?

chinedufn:
On mobile so can’t easily read the PR diff right now but just in case this is related...:

When WebGL shaders can’t compile they don’t throw an error with the reason. To see what went wrong you need to explicitly print out the shader info log. This is a common pattern is any WebGlProgram.

@LesleyLai I’m noticing a bunch of indentation from some nested functions so would be nice to not have that if possible.

PR #1230, opened 2019-02-04
futures: Add a test for creating multiple futures from the same promise

fitzgen:
Didn't find the bug that I thought maybe existed (maybe it is just a little more complicated than I suspect?) but figured it is worth keeping this test.


fitzgen:
(The bug I was tracking down was completely unrelated, FWIW.)

Issue #898, opened 2018-09-27
wasm-bindgen-test printing failed string assertions incorrectly


chinedufn:
I'm asserting whether or not to str's are equal. The test works fine when things are passing, but the failed assertion output is off.

Here's the program:

```rust
extern crate wasm_bindgen_test;
use wasm_bindgen_test::*;

#[macro_use]
extern crate virtual_dom_rs;

wasm_bindgen_test_configure!(run_in_browser);

#[wasm_bindgen_test]
pub fn string_assertion(){
    assert_eq!("<div>", "<span>");
}
```

Here's the output:

```
    error output:
        panicked at 'assertion failed: `(left == right)`
          left: `"
"`,
         right: `""`', virtual-dom-rs/tests/diff-patch.rs:14:5
```

wasm-bindgen-cli v0.2.23
wasm-bindgen-test v0.2.23

---

Would love to fix this if you don't mind pointing me in the right direction!

chinedufn:
Alright I made it to here

https://github.com/rustwasm/wasm-bindgen/blob/master/crates/test/src/rt/mod.rs#L517-L525

And I'll stop there since it'll probably be a bit more productive to wait a moment for your tips on where to look / what to try (if possible!) vs. diving in and figuring it out as I go.

PR #1233, opened 2019-02-08
Use textContent instead of innerHTML for extracting logs from headless browser tests

### LINKED ISSUE #1232: HTML in console log is interpreted in tests
With this test:

```
#[wasm_bindgen]
extern "C" {
    fn log(s: &str);
}

#[wasm_bindgen_test]
fn direct_log() {
    log("hello<br><b>world</b>");
    assert!(false);
}
```

When run like this:

    CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test --target wasm32-unknown-unknown

...there are two places where the console output is printed.

First, after the assertion: This looks fine.

img

But later in the console.log excerpt, the HTML seems to get interpreted:

img

This makes logging in tests quite useless if the logs contain < and >, since things in between are interpreted as elements.

Example project: [testproj.tar.gz](https://github.com/rustwasm/wasm-bindgen/files/2840446/testproj.tar.gz)

fitzgen:
Easier to maintain fidelity and less likely to be sabotaged by XSS and `<script>` tags.

Fixes #1232

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/index-headless.html
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/index-headless.html
@@ -13,7 +13,7 @@
      const orig = id => (...args) => {
          const logs = document.getElementById(id);
          for (let msg of args) {
-             logs.innerHTML += `${msg}\n`;
+             logs.textContent += `${msg}\n`;
          }
      };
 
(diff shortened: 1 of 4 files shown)


fitzgen:
I should have caught this in #898...

fitzgen:
Not sure how to test our log output of the runner, but:

screen shot 2019-02-08 at 10 34 58 am

alexcrichton:
👍

dbrgn:
Thanks for the quick fix!

Issue #1213, opened 2019-01-28
TodoMVC example issues


liamcurry:
The TodoMVC example appears to have a couple issues.

- Demo page: https://rustwasm.github.io/wasm-bindgen/exbuild/todomvc/
- Source code: https://github.com/rustwasm/wasm-bindgen/tree/master/examples/todomvc

1. `index.css` is missing from the `gh-pages` branch: https://github.com/rustwasm/wasm-bindgen/tree/gh-pages/exbuild/todomvc This could be fixed by tweaking `webpack.config.js` to handle CSS files and adding `require('./index.css')` to `index.js`.
2. JS errors are thrown when interacting with any inputs:

![image](https://user-images.githubusercontent.com/77758/51869172-c1906400-231d-11e9-88c4-89807b690815.png)

alexcrichton:
@jonathanKingston would you be able to help look into this? If not no worries! I think though with these issues we may want to temporarily remove it until they're resolved before re-posting the example?

alexcrichton:
Ok no worries! I'll wait till next week to temporarily take it down, no need to rush here!

jonathanKingston:
I looked into this and was unable to build initially due to the following change needed:
```
diff --git a/examples/todomvc/index.js b/examples/todomvc/index.js
index 29a450d1..bc0a79ba 100644
--- a/examples/todomvc/index.js
+++ b/examples/todomvc/index.js
@@ -1,4 +1,7 @@
 // For more comments about what's going on here, check out the `hello_world`
 // example
-import('./pkg/todomvc')
+const rust = import('./pkg/todomvc');
+
+rust
+  .then(m => m.run())
   .catch(console.error);
```

There was also a new compiler warning:
```
diff --git a/examples/todomvc/src/controller.rs b/examples/todomvc/src/controller.rs
index ebb0ccf4..bfad84c0 100644
--- a/examples/todomvc/src/controller.rs
+++ b/examples/todomvc/src/controller.rs
@@ -68,7 +68,7 @@ impl Controller {
     }
 
     pub fn set_page(&mut self, raw: String) {
-        let route = raw.trim_left_matches("#/");
+        let route = raw.trim_start_matches("#/");
         self.active_route = route.to_string();
         self._filter(false);
         self.add_message(ViewMessage::UpdateFilterButtons(route.to_string()));
```

[...]

PR #1238, opened 2019-02-11
Fixup compiler warnings for Todo MVC example

jonathanKingston:
See compiler warnings mentioned in https://github.com/rustwasm/wasm-bindgen/issues/1213 but this doesn't fix all the issues there of the changed WASM api.


alexcrichton:
Thanks!

PR #1239, opened 2019-02-11
Expected a string or undefined for debug builds. Fixes #1213

### LINKED ISSUE #1213: TodoMVC example issues

jonathanKingston:


alexcrichton:
Thanks! The intention of this method though is to only accept strings in string-taking APIs and something like `Option<String>` would be used instead for something which also takes `undefined`. Can you elaborate on the use case you're thinking though to take `undefined` here?

alexcrichton:
I'm gonna go ahead and close this for now because I think this is working as intended, but we can of course continue to follow up if necessary!

PR #1240, opened 2019-02-12
Run mdbook-linkcheck on guide

stevebob:
This updates the guide manifest to run linkcheck as part of building. A side effect is the html output is now in a subdirectory named "html".  I also updated the travis ci config to install mdbook-linkcheck with cargo. While I was at it, I started installing mdbook with cargo as well , and since this will compile mdbook, I added `cargo: cache` to prevent the need to compile it each time the test runs.

Previously mdbook was installed with curl rather than cargo. Was that to prevent compiling mdbook each time the test ran?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 sudo: false
+cache: cargo
 
 INSTALL_NODE_VIA_NVM: &INSTALL_NODE_VIA_NVM
   |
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: rust
 sudo: false
+cache: cargo
```
alexcrichton: The default cargo cache also caches the `target` directory which can unfortunately get quite large due to Cargo not cleaning things out. I wonder if the cache could only be set for the one documentation builder though?

.travis.yml
```diff
@@ -212,14 +213,13 @@ matrix:
     # Build mdbook documentation
     - name: "doc: Guide documentation"
       install:
-        - mkdir -p $HOME/mdbook
-        - curl -L https://github.com/rust-lang-nursery/mdBook/releases/download/v0.2.1/mdbook-v0.2.1-x86_64-unknown-linux-musl.tar.gz | tar xzf - -C $HOME/mdbook
```
alexcrichton: In the interest of CI time and a smaller cache, perhaps this could still be downloaded but mdbook-linkcheck is compiled?

.travis.yml
stevebob: I will try this!

.travis.yml
stevebob: Do we still save much time/cache if mdbook is a dependency of mdbook-linkcheck? Building mdbook-linkcheck will still build mdbook, but as a library instead of a binary.

stevebob:
I briefly toyed with running mdbook-linkcheck as a standalone binary rather than as part of mdbook, but it looks like it incorrectly flags local links as broken. It doesn't behave this way when running as part of mdbook.

.travis.yml
alexcrichton: Oh yeah for sure just caching the binaries is fine, it's caching wasm-bindgen's target directory that's a bit large over time

alexcrichton:
Ah ok let's stick to what works then!

alexcrichton:
@stevebob have you had a chance to poke again some more at this?

stevebob:
Sorry I got distracted! I'll play around with this tomorrow.

alexcrichton:
No worries! As a heads up we've just [recently switched to Azure Pipelines](https://github.com/rustwasm/wasm-bindgen/pull/1300) so this may need some more invasive changes on those fronts!

stevebob:
Finally have some time to look at this! I'm going to close this PR and open a new one for azure pipelines for clarity. What's the recommended way to test my changes before submitting? Currently I've removed the "deploy" job from the azure pipelines config as I don't have permission to run it (and I don't want to deploy anyway).

PR #1248, opened 2019-02-13
[HTMLMediaElement] Reintroduce srcObject

selaux:
This was removed when mozilla specific extensions were removed. It is not mozilla specific though and currently the only way to show webcam data in a video element that I am aware of.


alexcrichton:
Thanks! Looks like the original comment was slightly off :)

PR #1260, opened 2019-02-15
Temporarily unconditionally remove producers section

alexcrichton:
We've had a lot of bug reports with upstream webpack currently and while
webpack has a fix it may take a moment to deploy. Let's try and fix
wasm-bindgen in the meantime!

Once webpack is updated we can go back to emitting a producers section
by default and publish a new version of wasm-bindgen.


xtuc:
:tada: thanks

PR #1275, opened 2019-02-19
Support `Option<RustStruct>` in arguments/returns

### LINKED ISSUE #1252: Nullable return types
This example builds:

```rust
#[wasm_bindgen]
pub struct Hello {
    pub a: u32,
}

#[wasm_bindgen]
pub fn say_hello() -> Hello {
    Hello { a: 42 }
}
```

But when changin the return type of `say_hello`  to an Option, it fails:

```rust
#[wasm_bindgen]
pub struct Hello {
    pub a: u32,
}

#[wasm_bindgen]
pub fn say_hello() -> Option<Hello> {
    Some(Hello { a: 42 })
}
```

Error:

```
  error[E0277]: the trait bound `Hello: wasm_bindgen::convert::traits::OptionIntoWasmAbi` is not satisfied
    --> examples/hello_world/src/lib.rs:18:1
     |
  18 | #[wasm_bindgen]
     | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::traits::OptionIntoWasmAbi` is not implemented for `Hello`
     |
     = note: required because of the requirements on the impl of `wasm_bindgen::convert::traits::IntoWasmAbi` for `std::option::Option<Hello>`

```

Is this not supported? I would have expected `Hello | null` as return type.

alexcrichton:
Add all the necessary support in a few locations and we should be good
to go!

Closes #1252

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -248,6 +248,17 @@ impl ToTokens for ast::Struct {
                     (*js).borrow_mut()
                 }
             }
+
+            impl ::wasm_bindgen::convert::OptionIntoWasmAbi for #name {
+                #[inline]
+                fn none() -> Self::Abi { 0 }
+            }
+
+            impl ::wasm_bindgen::convert::OptionFromWasmAbi for #name {
+                #[inline]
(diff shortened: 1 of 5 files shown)


dbrgn:
Fantastic, thanks!

PR #1279, opened 2019-02-20
Add support for `TextEncoder#encodeInto`

### LINKED ISSUE #1172: Add experimental support for TextEncoder.encodeInto
[Recently landed](https://bugzilla.mozilla.org/show_bug.cgi?id=1514664) in Firefox Nightly this API has the promise of being much more efficient for encoding strings when a string is passed from JS to Rust. 

Let's add support in generated bindings to detect this method, [namely in `passStringToWasm`](https://github.com/rustwasm/wasm-bindgen/blob/12893183b65b0d9449f767b5c816339225785579/crates/cli-support/src/js/mod.rs#L981-L993), and call it if it's available!

alexcrichton:
This commit adds support for the recently implemented standard of
[`TextEncoder#encodeInto`][standard]. This new function is a "bring your
own buffer" style function where we can avoid an intermediate allocation
and copy by encoding strings directly into wasm's memory.

Currently we feature-detect whether `encodeInto` exists as it is only
implemented in recent browsers and not in all browsers. Additionally
this commit emits the binding using `encodeInto` by default, but this
requires `realloc` functionality to be exposed by the wasm module.
Measured locally an empty binary which takes `&str` previously took
7.6k, but after this commit takes 8.7k due to the extra code needed for
`realloc`.

[standard]: https://encoding.spec.whatwg.org/#dom-textencoder-encodeinto

Closes #1172

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1,6 +1,6 @@
 use crate::decode;
 use crate::descriptor::{Descriptor, VectorKind};
-use crate::Bindgen;
+use crate::{Bindgen, EncodeInto};
 use failure::{bail, Error, ResultExt};
 use std::collections::{HashMap, HashSet};
 use walrus::{MemoryId, Module};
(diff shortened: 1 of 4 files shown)


alexcrichton:
@fitzgen I'm curious on your thoughts about the file size increase here. One possible option I can think of is that we could detect when `wasm-bindgen`, the crate, is compiled with `-Os` or `-Oz` and perhaps exclude this faster functionality.

fitzgen:
> @fitzgen I'm curious on your thoughts about the file size increase here. One possible option I can think of is that we could detect when `wasm-bindgen`, the crate, is compiled with `-Os` or `-Oz` and perhaps exclude this faster functionality.

I think the ideal place we would like to end up at is something like

* add `-O1` and `-Os` to optimize for speed and size of bindings respectively
  * decide on a default choice between these if none is provided
  * default usage of `TextEncode#encodeInto` based on whether this is present or not
* have a flag to control `TextEncoder#encodeInto` behavior specifically so its default can be overridden

how does that sound?

In terms for this PR, as opposed to an ideal future end state, I think just choosing a default and providing a flag to explicitly set it on or off is good enough to land this.

crates/cli-support/src/js/mod.rs
```diff
@@ -1168,10 +1169,37 @@ impl<'a> Context<'a> {
...
+                return ptr;
+            }
+        ");
+
         self.global(&format!(
             "
             function passStringToWasm(arg) {{
                 {}
+                if (typeof cachedTextEncoder.encodeInto === 'function')
+                    return passStringToWasmFast(arg);
```
fitzgen: Instead of checking every time we pass a string, we should do this:

```js
(function () {
    return typeof TextEncoder.prototype.encodeInto === "function"
        ? passStringToWasmFast
        : passStringToWasm;

    function passStringToWasmFast(arg) { ... }
    function passStringToWasm(arg) { ... }
}())

**Comment by alexcrichton** [comment]
Yeah I think ideally we'd have all the various knobs, but I'm also hoping that we can in the meantime pick reasonable defaults that avoid the need for knobs for as long as possible. Do you think it's worth doing any detection right now? (I'm sort of on the fence about this but would lean towards "no")

If we need to do detection, though, do you think we need to probe much beyond the compile settings for the crate?

**Comment by fitzgen** [comment]
> Do you think it's worth doing any detection right now? (I'm sort of on the fence about this but would lean towards "no")

Detection of whether `-Os` or `-Oz` were used instead of `-O2` or soemthing? No I don't think that is worth attempting.

**Comment by alexcrichton** [comment]
Ok just to make sure I understand, you're ok merging this (with the change to only do the feature detect once) as-is? In the long-term we'd keep the option open though to have configuration for tweaking this output

**Comment by fitzgen** [comment]
> Ok just to make sure I understand, you're ok merging this (with the change to only do the feature detect once) as-is?

I think we should at least have a CLI flag like `--use-encode-into=yes|no` before landing and leave the larger configuration stuff for the future.

**Comment by alexcrichton** [comment]
Updated!

File: crates/cli-support/src/js/mod.rs (hunk 1 above)
Comment (fitzgen) [reply crates/cli-support/src/js/mod.rs]: The `use_encode` body is still checking for `encodeInto`, which I think is an oversight

Comment (alexcrichton) [reply crates/cli-support/src/js/mod.rs]: Oops yes

File: crates/cli-support/src/js/mod.rs [hunk 2]
```diff
@@ -1168,19 +1168,77 @@ impl<'a> Context<'a> {
...
+                while (true) {{
+                    const view = getUint8Memory().subarray(ptr + writeOffset, ptr + size);
+                    const {{ read, written }} = cachedTextEncoder.encodeInto(arg, view);
+                    arg = arg.substring(read);
+                    writeOffset += written;
+                    if (arg.length === 0) {{
+                        break;
+                    }}
+                    ptr = wasm.__wbindgen_realloc(ptr, size, size * 2);
+                    size *= 2;
```
hsivonen: If the allocator isn't a power-of-two allocator, this can overallocate. It is known that the max number of bytes that `encodeInto` can require is JavaScript string length times 3. (It follows that if the Wasm program knows the string to be short-lived on the Wasm side, it makes sense to just allocate for the ma length case up front and not reallocate.)

alexcrichton: Hm good point! Do you think it'd be worthwhile do do something like `malloc(string.length)`, and if that fails `realloc(string.length * 3)` followed by `realloc(the_actual_length)` and that'd be more appropriate than this loop?

hsivonen: If the allocations can be assumed the be short-lived, I suggest going with `malloc(string.length * 3)` without a loop.

If they are assumed to be long-lived, I suggest coming up with a magic number `magic` that characterizes how much smaller the initial allocation needs to than the worst case for it to be worthwhile not to allocate for the worst case right away. 

Then:

```
let min = roundUpToBucketSize(string.length);
let max = string.length * 3;
if (min + magic > max) {
  // min is close enough to max anyway
  let m = malloc(max);
  let (read, written) = encodeInto(...);
  assert(read == string.length);
} else {
  let m = malloc(min);
  let (read, written) = encodeInto(...);
  if (read < string.length) {
     let inputLeft = string.length - read;
     let outputNeeded = inputLeft * 3;
     let m = realloc(m, written + outputNeeded);
     // ... encodeInto with input starting from read and output starting from written
  }
}
```

alexcrichton: Ok thanks for the info! I've opened up https://github.com/rustwasm/wasm-bindgen/issues/1313 to further track this.

PR #1281, opened 2019-02-21
Add support for `FnOnce` to `Closure<T>`

fitzgen:
Sort of. By turning them into `FnMut` and doing a dynamic check that they are only ever called once.

I initially tried to get `Closure<FnOnce(...) -> R>` working, but that proved too difficult.

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -9,10 +9,11 @@ use std::marker::Unsize;
 use std::mem::{self, ManuallyDrop};
 use std::prelude::v1::*;
 
-use JsValue;
 use convert::*;
 use describe::*;
 use throw_str;
+use JsValue;
+use UnwrapThrowExt;
 
 /// A handle to both a closure in Rust as well as JS closure which will invoke
 /// the Rust closure.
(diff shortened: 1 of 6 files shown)


src/closure.rs
```diff
@@ -402,6 +623,56 @@ macro_rules! doit {
...
+
+        #[allow(non_snake_case)]
+        impl<T, $($var,)* R> WasmClosureFnOnce<($($var),*), R> for T
+            where T: 'static + FnOnce($($var),*) -> R,
+                  $($var: FromWasmAbi + 'static,)*
+                  R: ReturnWasmAbi + 'static
+        {
+            type FnMut = FnMut($($var),*) -> R;
+
+            fn into_fn_mut(self) -> Box<Self::FnMut> {
```
alexcrichton: Technically this isn't quite as optimized as we might otherwise want because a `FnOnce` that doesn't actually close over anything will allocate because of the discriminant on `Option<T>`. I think fully supporting this in the limit though will require specialized support in the JS bindings (the JS bindings are the ones to clear the stored pointer).

Mind throwing a comment somewhere about that? I think this is totally fine as a first pass

alexcrichton: (technically the same comment for the one below too)

fitzgen:
https://travis-ci.com/rustwasm/wasm-bindgen/jobs/179355779#L951-L961

```
RuntimeError: memory access out of bounds
    at dlmalloc::dlmalloc::Dlmalloc::insert_large_chunk::h752576fb10674c30 (wasm-function[1556]:244)
    at dlmalloc::dlmalloc::Dlmalloc::free::h4e7d0bccb8a2e263 (wasm-function[659]:806)
    at __rdl_dealloc (wasm-function[10159]:8)
    at __rust_dealloc (wasm-function[10043]:7)
    at alloc::alloc::dealloc::h1b984820f03b24d3 (wasm-function[4389]:100)
    at <alloc::alloc::Global as core::alloc::Alloc>::dealloc::h5015cfcc10ef6ccf (wasm-function[5536]:84)
    at <alloc::rc::Rc<T> as core::ops::drop::Drop>::drop::h6a78b25b82ddd7e0 (wasm-function[2222]:216)
    at core::ptr::real_drop_in_place::h3aa3cadebfeeb41a (wasm-function[8375]:38)
    at core::ptr::real_drop_in_place::hc4e654cf5dfc7dc7 (wasm-function[5383]:84)
    at wasm_bindgen_test::__rt::State::log_test_result::h3d6ffabb89d2894e (wasm-function[672]:841)
```

wat

PR #1298, opened 2019-02-26
Fix headless test runner on macOS High Sierra

LegNeato:


alexcrichton:
Thanks for this! Can you detail a bit about what the bug was? I'm not entirely sure what `#[serde(flatten)]` is doing here but it seems like it's fixing the issue?

LegNeato:
Sure! Apparently post High Sierra the format of `safaridriver`'s response changed.  On my machine, I get:

```json
{
  "status": 0,
  "sessionId": "7E00E58A-859E-4127-9E0A-F67FBE4CB647",
  "value": {
    "version": "13605.3.8",
    "rotatable": false,
    "applicationCacheEnabled": true,
    "databaseEnabled": true,
    "handlesAlerts": true,
    "cleanSession": true,
    "browserName": "safari",
    "javascriptEnabled": true,
    "locationContextEnabled": false,
    "platform": "macOS",
    "webStorageEnabled": true,
    "cssSelectorsEnabled": true,
    "nativeEvents": true,
    "platformName": "macOS"
  }
}
```

Because `Option::or` is eager, for the case where a `value` entry exists but doesn't contain `sessionId`, deserializing would fail.

`serde(flatten)` is supposed to hoist the keys of the substruct into the parent struct, but after manually testing I see it doesn't behave the way I want for this fix. So I pushed a new change that should work for all cases.

In this [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e282af14ef49bea587032fea9f1da2da), you can see the three cases:
* `sessionId` is at the top level with a `value` entry with no `sessionId` ("--legacy" or High Sierra and below)
* `sessionId` is at the top level without a `value` entry
* no `sessionId` at the top level but one in `value`

alexcrichton:
Ah I see, thanks for the explanation! The PR as-is definitely makes sense to me as well!

PR #1300, opened 2019-02-27
Switch CI to Azure Pipelines

alexcrichton:
This commit switches CI for the wasm-bindgen repository from a mixture
of Travis and AppVeyor to Azure Pipelines. One of the main reasons for
this was for me to personally get familiar with Azure Pipelines, but
after finishing it I think that this'd be a good transition for this
repository anyway.

With azure pipelines we're able to get more parallelism (10 vs the 3 on
Travis plus 1 on AppVeyor) as well as house all configuration in the
same location for one service (Azure Pipelines covers all 3 major
platforms).

I've tested this a good deal on my own personal repository as well, and
I believe that this is a lossless representation of our current CI,
releases and all!


alexcrichton:
One feature as well is that a lot of [intermediate artifacts](https://dev.azure.com/rustwasm/wasm-bindgen/_build/results?buildId=2&view=results) are deployed and published for each build. It'd be good to get another set of eyes on the `gh_pages` and `gh_release` artifacts there. One is what is pushed directly to `gh-pages` brranch, and another is what's published on each tag as files for upload (the release tarballs)

fitzgen:
>  [intermediate artifacts](https://dev.azure.com/rustwasm/wasm-bindgen/_build/results?buildId=2&view=results)

I just get a 500 error when I follow that link...

alexcrichton:
Hm ok. @fitzgen do you want to try setting up an azure pipelines account and I'll see if I can add you to the rustwasm organization? Other than that I can also attach the raw artifacts here

[gh_pages.zip](https://github.com/rustwasm/wasm-bindgen/files/2912269/gh_pages.zip)

[gh_release.zip](https://github.com/rustwasm/wasm-bindgen/files/2912270/gh_release.zip)

alexcrichton:
@fitzgen do you have further thoughts on the transition here?

fitzgen:
I poked around the azure setup and I feel like I got as much out of that experience as I can without actualyl getting more experience *using* azure pipelines. So I think we are good to go.

My only question, and I think/hope the answer is "yes", but are dist binaries going to have teh same URLs in github releases? If not, then we need to fix up wasm-pack some before migrating I think.

alexcrichton:
Ok! Let's see how this goes!

PR #1302, opened 2019-02-27
Add doc comments export for typescript

### LINKED ISSUE #1276: Add doc comments from rust to typescript definitions
## 💡 Feature description
It would be nice to have documentation comments from rust in corresponding typescript definitions.

#### 💻 Basic example

If I have the following rust code:
```rust
#[wasm_bindgen]
/// Test comment
pub fn greet() {
    alert("Hello, test!");
}
```
I want to see `Test comment` in .d.ts file similarly to .js file:
```ts
/**
 * Test comment
 * @returns {void}
 */
export function greet(): void;
```

tyleranton:
This PR adds exporting Rust doc comments to TypeScript defs.

This is just pushing the comment string to the `typescript` variable on `Context` right after pushing it to `globals` for JS.

I'm not sure if there are any significant differences between JS Doc and TypeScript docs, so I didn't add in any additional formatting. I'd be happy to try and add that if there are any.

This is my first code contribution, so please let me know if there are additional things needed for this PR.

Thanks!

Fixes #1276

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -122,6 +122,7 @@ impl<'a> Context<'a> {
         let contents = contents.trim();
         if let Some(ref c) = comments {
             self.globals.push_str(c);
+            self.typescript.push_str(c);
         }
         let global = if self.use_node_require() {
             if contents.starts_with("class") {


alexcrichton:
Heh that looks like it wasn't so hard after all! 

Unfortunately we don't have a ton of tests for the TypeScript output right, now, but I'm curious if you've had a chance to test this out locally to make sure comments go to the right location?

tyleranton:
Yep I did! They go in the `.js` and `.d.ts` files respectively (above their functions).

alexcrichton:
Nice! Sounds good to me then

PR #1305, opened 2019-02-28
Implement transitive support for NPM dependencies 

alexcrichton:
This commit implements [RFC 8], which enables transitive and transparent
dependencies on NPM. The `module` attribute, when seen and not part of a
local JS snippet, triggers detection of a `package.json` next to
`Cargo.toml`. If found it will cause the `wasm-bindgen` CLI tool to load
and parse the `package.json` within each crate and then create a merged
`package.json` at the end.

[RFC 8]: https://github.com/rustwasm/rfcs/pull/8

> **Note**: this is currently based on https://github.com/rustwasm/wasm-bindgen/pull/1295 so the first commits can be ignored, and additionally [RFC 8] is not FCP'd or merged yet, so merging this should hold off until that happens.


crates/backend/src/encode.rs
```diff
@@ -83,6 +85,7 @@ impl Interner {
             let msg = "relative module paths aren't supported yet";
             return Err(Diagnostic::span_error(span, msg))
         } else {
+            self.check_for_package_json();
```
fitzgen: I think we want to unconditionally check for `package.json` all the time. The npm package might put stuff on the global scope (generally not recommended, but might be required for polyfills or something). This would mean that the use of the dependency is effectively invisible to us, and we can't assume that if there is no module import, then we can avoid the declared npm dependency.

alexcrichton: Sounds reasonable to me!

alexcrichton:
Ok I've opened a follow-up for this in `wasm-pack` at https://github.com/rustwasm/wasm-pack/issues/606

Issue #1040, opened 2018-11-19
.d.ts for _bg module


likr:
I'm trying to access wasm memory from TypeScript like this example.
https://rustwasm.github.io/wasm-bindgen/reference/types/pointers.html

But, the TypeScript compiler shows an error because the type definition for _bg module does not exist.

Currently, I manually added following .d.ts file to avoid the error.
(@types/webassembly-js-api is needed.)

```
import 'webassembly-js-api';

export var memory : WebAssembly.Memory;
```

I think it is better to automatically generate .d.ts file for _bg module.
Or, is there any other way to solve this?

alexcrichton:
Oops yeah I forgot about importing memory! Definitely something we should emit `.d.ts` files for!

We could also go ahead and emit a `.d.ts` file for the entire wasm module, and it would be pretty easy to do so as we've got the types of all the exports already!

PR #1312, opened 2019-03-05
Add webassembly-js-api for *_bg.d.ts.

konkers:
Per #1040, `import 'webassembly-js-api'` is needed for `WebAssembly.Memory`.

### DIFF
--- a/crates/cli-support/src/wasm2es6js.rs
+++ b/crates/cli-support/src/wasm2es6js.rs
@@ -52,7 +52,7 @@ pub fn typescript(module: &Module) -> Result<String, Error> {
             walrus::ExportItem::Function(i) => i,
             walrus::ExportItem::Memory(_) => {
                 exports.push_str(&format!(
-                    "export const {}: WebAssembly.Memory;\n",
+                    "import 'webassembly-js-api';\nexport const {}: WebAssembly.Memory;\n",
                     entry.name,
                 ));
                 continue;


alexcrichton:
Thanks for the PR!

This inspired me to [write some tests we can run on CI](https://github.com/rustwasm/wasm-bindgen/pull/1314) to ensure that this doesn't regress. In making that though I'm wondering if this is necessary? There it looks like just depending on `@types/webassembly-js-api` is enough to pacity the TypeScript compiler, but I'm curious if perhaps there's a common way to use TypeScript or configure it which does require this? (I'm not too familiar with TypeScript myself)

konkers:
Ooh Typescript tests.  Nice.

So, I'm not super familiar with with TS either.  I looked around a bit.  Essentially you're asking the TS compiler to introduce something into the global scope automatically.  As far as I can tell, that's not a common thing to do.

alexcrichton:
I agree yeah it's nonstandard and in general a bad idea, but for this case specifically `WebAssembly` is like `String` and `Array` where it's ambiently available in JS (part of the JS spec) and doesn't typically need an import. For that I think the import shouldn't be necessary because the JS types are already in scope and don't need imports

Pauan:
Yeah, this needs to be added to the TypeScript compiler. They [already do that for ES6 + DOM APIs](https://github.com/Microsoft/TypeScript/blob/2f9218f346bbc606cccc4aebfca1c2c4f3167fee/src/lib/dom.generated.d.ts). It's easy to add support for WebAssembly.

They have a tracking issue here: https://github.com/Microsoft/TypeScript/issues/29747

alexcrichton:
Thanks for the reference and comment @Pauan! Sounds like we should close this for now but recommend depending on `@types/webassembly-js-api` in the meantim?

Pauan:
Yes, installing `@types/webassembly-js-api` should be enough for now.

If that issue takes too long to fix, we can gently ping the TypeScript team.

alexcrichton:
Sounds good to me!

@konkers would you be interested in perhaps sending a PR to update our documentation for using TypeScript to indicate this?

konkers:
Odd!  I though I tried with just`@types/webassembly-js-api` installed.  Guess not.  I reverted my patch and it's all working now.  Thanks for the pointers @Pauan!

konkers:
Yeah, I'll send a PR for a doc update.

konkers:
I'm thinking that adding an example using typescript with callouts about `@types/webassembly-js-api` and the necessary webpack configs would be best.  Thoughts?

alexcrichton:
@konkers an example sounds like a great idea to me!

abau:
> Yes, installing @types/webassembly-js-api should be enough for now

Just for the record, apparently one has to install `@types/webassembly-js-api@0.0.3` since installing the latest version fails due to:

```
$ npm install --save @types/webassembly-js-api
npm WARN deprecated @types/webassembly-js-api@2.0.0: This is a stub types definition. webassembly-js-api provides its own type definitions, so you do not need this installed.
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/webassembly-js-api - Not found
npm ERR! 404 
npm ERR! 404  'webassembly-js-api@*' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 It was specified as a dependency of '@types/webassembly-js-api'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

PR #1321, opened 2019-03-06
Fixed typo in .slice for typed arrays

### LINKED ISSUE #1317: Should .slice called on TypedArray return the same type as the original?
### Summary

In JS for example `.slice` called on Int32Array will return Int32Array.
In wasm-bindgen there is
```
pub fn slice(this: &Int32Array, begin: u32, end: u32) -> Uint8ClampedArray;
```
Should it return Int32Array instead?

ibaryshnikov:
Closes #1317


alexcrichton:
Thanks!

PR #1328, opened 2019-03-07
Switch the `--browser` argument to `--web`

### LINKED ISSUE #1318: Online example is broken
The documentation page https://rustwasm.github.io/wasm-bindgen/examples/hello-world.html has a link ["view the compiled example online"](https://rustwasm.github.io/wasm-bindgen/exbuild/hello_world/).

This example does not work.

### To reproduce
1. Open https://rustwasm.github.io/wasm-bindgen/exbuild/hello_world/ in your browser.

### Expected behavior
Hello world alert or whatever.

### Actual behavior
Blank page, nothing happens. In the JS console there is the following message:
```
TypeError: Cannot read property '__wbindgen_malloc' of undefined
    at passStringToWasm (hello_world.js:66)
    at Module.greet (hello_world.js:77)
    at rust.then.m (index.js:7)
```

alexcrichton:
This commit reverts part of the implementation of [RFC 6]. That RFC
specified that the `--browser` flag was going to be repurposed for the
new "natively loadable as ES module output", but unfortunately the
breakage is far broader than initially expected. It turns out that
`wasm-pack` passes `--browser` by default which means that a change to
break `--browser` would break all historical versions of `wasm-pack`
which is a bit much for now.

To solve this the `--browser` flag is going back to what it represents
on the current released version of `wasm-bindgen` (optimize away some
node.js checks in a few places for bundler-style output) and a new
`--web` flag is being introduced as the new deployment strategy.

[RFC 6]: https://github.com/rustwasm/rfcs/pull/6

Closes #1318

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -179,7 +179,7 @@ impl<'a> Context<'a> {
                 }
             }
             OutputMode::Web => {
-                // In browser mode there's no need to export the internals of
+                 // In web mode there's no need to export the internals of
                 // wasm-bindgen as we're not using the module itself as the
                 // import object but rather the `__exports` map we'll be
                 // initializing below.
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -166,7 +166,7 @@ impl<'a> Context<'a> {
                     format!("__exports.{} = {};\n", name, contents)
                 }
             }
-            OutputMode::Bundler
+            OutputMode::Bundler { .. }
             | OutputMode::Node {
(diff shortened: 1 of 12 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -178,7 +178,7 @@ impl<'a> Context<'a> {
                     format!("export const {} = {};\n", name, contents)
                 }
             }
-            OutputMode::Browser => {
+            OutputMode::Web => {
                 // In browser mode there's no need to export the internals of
```
fitzgen:
```suggestion
                 // In web mode there's no need to export the internals of
```

Issue #1320, opened 2019-03-06
Span doesn't implement PartialEq and Eq


c410-f3r:
`Span` doesn't implement `PartialEq` and `Eq`, making the `extra-traits` feature broken.
Should `PartialEq` and `Eq` be removed or is it better to wait for an implementation?

Related: [#154](https://github.com/alexcrichton/proc-macro2/issues/154)

https://github.com/rustwasm/wasm-bindgen/blob/9f6c2a612503f99c76c3c2c0446710866f102e9f/crates/backend/src/ast.rs#L200-L210

alexcrichton:
Oh dear this does seem bad! We probably want to remove this feature for now in that case? (removing `PartialEq` and `Eq` also seems fine to me)

dtolnay:
I notice that ast::Function stores `name: String` and `name_span: Span` as separate fields. I would recommend storing these as one field of type proc_macro2::Ident which does have a PartialEq impl, or if the name is not guaranteed to be a legal Rust identifier/keyword then writing your own type similar to Ident to hold the name and span and provide PartialEq and any other traits you need.

c410-f3r:
@dtolnay Thanks!

PR #1331, opened 2019-03-07
Remove PartialEq and Eq from Function

c410-f3r:
For now, remove `PartialEq` and `Eq` for `Function` and everything else that uses it.
As soon as `Span` has `PartialEq`/`Eq` (seems trivial), this commit should be reverted and #1320 closed.


alexcrichton:
Thanks!

PR #1336, opened 2019-03-10
Simplified webgl example

### LINKED ISSUE #1322: Simplify WebGL example
### Motivation

There is a function signature
```rust
pub fn link_program<'a, T: IntoIterator<Item = &'a WebGlShader>>(
    context: &WebGlRenderingContext,
    shaders: T,
) -> Result<WebGlProgram, String> {
```
It's hard to teach, because it requires additional explanations about iterators and lifetimes.
There is also `String::from("")` and `"".into()`, which produces the question "why are they different?".

### Proposed Solution

Use distinct arguments instead of the iterator
```rust
pub fn link_program(
    context: &WebGlRenderingContext,
    vert_shader: &WebGlShader,
    frag_shader: &WebGlShader,
) -> Result<WebGlProgram, String> {
```

### Alternatives

Keep it as is

### Additional Context

I'd like to submit an update for this example if it's a good idea

ibaryshnikov:
Removed `<'a, T: IntoIterator<Item = &'a WebGlShader>>` from `link_program` sigrature, used `String::from` instead of `into` for consistency

Closes #1322

### DIFF
--- a/examples/webgl/src/lib.rs
+++ b/examples/webgl/src/lib.rs
@@ -33,7 +33,7 @@ pub fn start() -> Result<(), JsValue> {
         }
     "#,
     )?;
-    let program = link_program(&context, [vert_shader, frag_shader].iter())?;
+    let program = link_program(&context, &vert_shader, &frag_shader)?;
     context.use_program(Some(&program));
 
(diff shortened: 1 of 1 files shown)


tiye:
looks broken now https://rustwasm.github.io/wasm-bindgen/exbuild/webgl/ am I alone?

PR #1338, opened 2019-03-12
Workaround an upstream `nom` bug

alexcrichton:
This commit works around Geal/nom#843 where the API of the `nom` crate
changes based on feature selection, meaning we need to be compatible
even if another crate in the crate graph enables a feature.

Ideally this'd be fixed in upstream `nom`, and it looks like it will in
the next major version! For now a local catch-all directive should help
out.


thavlik:
Thanks for fixing this so quickly!

PR #1339, opened 2019-03-12
Use basic-http-server from crates.io.

anderejd:
Avoids `python3` as dependency in some examples by using [basic-http-server](https://crates.io/crates/basic-http-server).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -168,7 +168,7 @@ jobs:
           echo "##vso[task.setvariable variable=PATH;]$PATH:$PWD"
         displayName: "install xargo"
       - script: |
-          sed -i 's/python/#python/' examples/raytrace-parallel/build.sh
+          sed -i 's/basic-http-server/#basic-http-server/' examples/raytrace-parallel/build.sh
           (cd examples/raytrace-parallel && ./build.sh)
           cp examples/raytrace-parallel/*.{js,html,wasm} $BUILD_ARTIFACTSTAGINGDIRECTORY
         displayName: "build example"
--- a/examples/raytrace-parallel/build.sh
+++ b/examples/raytrace-parallel/build.sh
@@ -22,4 +22,5 @@ WASM_BINDGEN_THREADS=1 \
     ../../target/wasm32-unknown-unknown/release/raytrace_parallel.wasm --out-dir . \
     --no-modules
 
-python3 -m http.server
+# `cargo install basic-http-server`
+basic-http-server
--- a/examples/without-a-bundler/build.sh
+++ b/examples/without-a-bundler/build.sh
@@ -12,4 +12,5 @@ cargo run --manifest-path ../../crates/cli/Cargo.toml \
     ../../target/wasm32-unknown-unknown/release/without_a_bundler.wasm --out-dir pkg \
     --web
 
-python3 -m http.server
+# `cargo install basic-http-server`
+basic-http-server


Pauan:
Any particular reason to use `basic-http-server` instead of [`https`](https://crates.io/crates/https)?

anderejd:
Nope, I simply did not know of the https crate. One reason I stopped looking for other minimal http servers on crates.io once I found basic-http-server was that alexcrichton is one of the owners.

alexcrichton:
Oh dear I didn't even know that I was on that crate! I've personally been happily using `https` myself. I think the long-term answer here is https://github.com/rustwasm/rfcs/pull/10, because otherwise I don't think there's actually that great of an answer of which server we should use here.

anderejd:
> I don't think there's actually that great of an answer of which server we should use here

My reason for replacing python3 is that I didn't have it installed by default on the machine I happened to be trying out wasm-bindgen on, macOS. That was a speed bump I wanted to smooth out for future users trying out the same wasm-bindgen examples.

> I think the long-term answer here is rustwasm/rfcs#10,

That link refers to wasm-pack, the examples I was trying out was https://github.com/rustwasm/wasm-bindgen/tree/master/examples/without-a-bundler-no-modules and https://github.com/rustwasm/wasm-bindgen/tree/master/examples/without-a-bundler, I could have added that in the original post.  

I regularly have to work with a couple of projects using node and npm due to, reasons, and have had some very bad experiences with those projects and would very much like to avoid any tooling and frameworks that depends on, or in any way interact with nodejs for new projects. Please keep npm and nodejs decoupled from wasm-bindgen, please? :)

anderejd:
>  I wonder if in the meantime this could switch to python -m SimpleHTTPServer

On macOs sure, but Windows doesn't have python by default. I guess my preference is to promote Rust based tools over Python and that preference could be the main driving factor for me posting this PR in the first place. If python is not considered an issue on any platform this PR can obviously be closed.

> For rustwasm/rfcs#10, it's intended to be used for those two example, and wasm-pack doesn't require npm/node for those modes!

Thank you for the clarification, that sounds great! I got the impression that wasm-pack was primarily an npm integration tool from reading the current readme on crates.io.

alexcrichton:
Ah that's true yeah, but the shell scripts really aren't that appropriate for Windows anyway. In general the idea was that the shell scripts are just examples and aren't really meant to be copied/executed literally. Ideally we will have no shell scripts at all (which rustwasm/rfcs#10 empowers), but it'll take us a bit to get there. Other than that we're just left with the question of what to do in the meantime.

Perhaps there could simply be a comment there saying "the files in this directory can now be served up" and point to some docs explaining some various ways how to do that?

Pauan:
@anderejd wasm-pack is intended for all workflows, including without npm.

However, you are right that the docs are misleading, and there is work being done to improve that: https://github.com/rustwasm/book/issues/150

anderejd:
I feel like I'm wasting your time here now :) The PR was intended to smooth over a minor speed bump in one or two of the examples, nothing more. If a proper fix is coming in wasm-pack, that does not depend on npm, that sounds great to me!

alexcrichton:
@anderejd oh no worries! I personally feel that is a somewhat difficult situation in that there's not really any great recommendation here. Using Python builtins are easy but not always availble. Using Rust crates means we have to pick one and users are highly likely to need to install something (which can take awhile). The best of all worlds is definitely `wasm-pack serve` which is intended to have all the bells and whistles needed and already likely to be installed when working with wasm-bindgen.

> EDIT: https://rustwasm.github.io/docs/wasm-pack/#prerequisites says NPM is a prerequisite, I'm confused again. Is wasm-pack usable without NPM? Sorry for going off-topic.

Good point! I've [sent a PR](https://github.com/rustwasm/wasm-pack/pull/569) to clarify these docs.

> EDIT 2: Related to documentation regarding NPM and wasm-pack: rustwasm/wasm-pack#466 I would like the documentation to say something about what wasm-pack brings to the build process that wasm-bindgen-cli is lacking.

A good point! We'll hopefully have a chance to once-over wasm-pack docs soon, but https://github.com/rustwasm/book/issues/150 is also likely going to help out with this.

alexcrichton:
I'm gonna go ahead and close this while we wait for the RFC to settle, but thanks regardless for the PR @anderejd!

Issue #1129, opened 2018-12-24
Support for named and optional parameters in .d.ts


dtysky:
This issue is related to #982 which is fixed, but that only makes the return type will be reflected,  if we use optional parameters, it still does not work. Another issue is that definitions of method will not keep their original name and use `arg0`... to instead, it may lose the semantic of parameters.

```rust
#[wasm_bindgen]
pub struct Color(f64, f64, f64, f64);

#[wasm_bindgen]
impl Color {
  #[wasm_bindgen(constructor)]
  pub fn new(r: Option<f64>, g: Option<f64>, b: Option<f64>, a: Option<f64>) -> Color {
    let mut color = Color(0.0, 0.0, 0.0, 0.0);

    match r {
      Some(value) => color.0 = value,
      None => ()
    }

    match g {
      Some(value) => color.1 = value,
      None => ()
    }

    match b {
      Some(value) => color.2 = value,
      None => ()
    }

    match a {
      Some(value) => color.3 = value,
      None => ()
    }

    color
  }
}
```

will generate dts:

```ts
export class Color {
  free(): void;
  constructor(arg0: number, arg1: number, arg2: number, arg3: number);
}
```

which I expect:

```ts
export class Color {
  free(): void;
  // use optional and keep semantic
  constructor(r?: number, g?: number, b?: number, a?: number);
}
```

Could we fix those issues for better programming experience?
Thanks.

alexcrichton:
Sounds like a good idea to me! I know the typescript generation with `Option` has been buggy in the past, and doing our best to preserve the argument names sounds like a great idea to me as well

c410-f3r:
If you guys don't mind, I will work on this. Just need a little help to figure out the best approach as everything related to optional parameters seems to be scattered across the repository.

alexcrichton:
Sure yeah!

Optional arguments are all handled [roughly around here](https://github.com/rustwasm/wasm-bindgen/blob/0f6e62a09b59465f7b7bbc7fca9a42de05b5723b/crates/cli-support/src/js/js2rust.rs#L230-L356) for TypeScript generation. Currently mostly generate `| undefined` rather than `foo?`, but doing either seems reasonable to me (not sure if there's a difference on the TypeScript side though).

[...]

PR #1344, opened 2019-03-14
Preserve argument names

c410-f3r:
Part of #1129.

* Closures aren't supported.
* Still not sure about https://github.com/rustwasm/wasm-bindgen/blob/e075d0492d4aaffaf38252fe8188b433edb0cb83/crates/backend/src/codegen.rs#L426 and https://github.com/rustwasm/wasm-bindgen/blob/e075d0492d4aaffaf38252fe8188b433edb0cb83/crates/cli-support/src/js/rust2js.rs#L95.


crates/cli-support/src/js/js2rust.rs
```diff
@@ -138,15 +151,22 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
         self
     }
 
-    fn abi_arg(&mut self) -> String {
-        let s = format!("arg{}", self.arg_idx);
+    fn abi_arg(&mut self, opt_arg_name: Option<String>) -> String {
```
alexcrichton: Could this perhaps be `Option<&str>` to avoid some extra clones as necessary in a few places?

crates/shared/src/lib.rs
```diff
@@ -106,6 +106,7 @@ macro_rules! shared_api {
         }
 
         struct Function<'a> {
+            arg_names: Vec<String>,
```
alexcrichton: If possible it'd be great to use `&'a str` here to allow the encoder/decoder to be pretty lean and avoid the extra `String` allocations

c410-f3r:
> Can you also manually confirm that the names of the arguments for TypeScript are coming through as well?

Capture d’écran de 2019-03-14 13-21-32

crates/cli-support/src/js/js2rust.rs
c410-f3r: I wish! syn's Ident doesn't expose `&str` and only implements `ToString`.

crates/shared/src/lib.rs
c410-f3r: I wish! syn's Ident doesn't expose &str and only implements ToString.

alexcrichton:
Ok thanks for checking and clarifying about `String`, thanks again for the PR!

c410-f3r:
@alexcrichton I am the one who thanks you. Thanks for your help and knowledge

PR #1350, opened 2019-03-14
Enable OptionIntoWasmAbi and OptionFromWasmAbi for ast::ImportEnum, RTCRtpTransceiver.webidl

alexlapa:
Some notes:
1. Actually `OptionIntoWasmAbi` for `ast::ImportEnum` can be safely deleted since its not being used in `bindings.rs`.
2. I believe that other disabled webidl's can be enabled.
3. Test should succeed on current Safari Technology Preview (12.2) but i don't know how to run `cargo test` on preview version. Latest Chrome and FF are ok.

Looking forward to your feedback (any :smile: )


alexcrichton:
This looks great to me, thanks!

Mind adding some tests as well in `tests/wasm/*` specifically for `Option<ImportedEnum>` as well?

alexlapa:
@alexcrichton ,

Sure! But i will be away for few days, so i guess i will do it by the end of next week.

alexcrichton:
Ok no worries!

alexcrichton:
Thanks!

PR #1356, opened 2019-03-16
Improve TS tests

c410-f3r:
I don't think it is possible to verify an argument name for #1344 by simply using `tsc`, which is unfortunate.


alexcrichton:
This looks awesome to me, thanks!

PR #1358, opened 2019-03-16
Installing and running the linkcheck mdbook backend

stevebob:
Installs and runs the "mdbook-linkcheck" backend on the guide, to test for broken links. This installs mdbook with cargo instead of downloading a binary because I can't work out a way to install mdbook-linkcheck in a compatible way.

Reincarnation of https://github.com/rustwasm/wasm-bindgen/pull/1240.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -232,15 +232,14 @@ jobs:
     displayName: "Doc - build the book"
     steps:
       - script: |
-          mkdir $HOME/mdbook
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! There's some unrelated failures here (being fixed elsewhere) but the "build the book" builder also failed, I think because the `cargo install` needs to happen after Rust is installed?

stevebob:
Added the "install-rust" template which seems to have fixed the problem.

alexcrichton:
Thanks!

PR #1369, opened 2019-03-19
Replace target flags with `--target`

alexcrichton:
This commit deprecates the `--web`, `--no-modules`, and `--nodejs` flags
in favor of one `--target` flag. The motivation for this commit is to be
consistent between `wasm-bindgen` and `wasm-pack` so documentation for
one is applicable for the other (so we don't have to document everywhere
what the translation is between flags). Additionally this should make it
a bit easier to add new targets (if necessary) in the future as it won't
add to the proliferation of flags.

For now the old flags (like `--web`) continue to be accepted, but
they'll be removed during the next set of breaking changes for
`wasm-bindgen`.


alexcrichton:
I had this idea when reviewing docs recently and was curious what others though, this is a pretty big change though, so if y'all think an RFC is requested I don't mind doing that.

ashleygwilliams:
@alexcrichton this is great. we should chat about the deprecation plan as wasm-pack will have to do a bit of dance around how it supports older wasm-bindgen versions (or alternatively we'll need to do a check that the wasm-bindgen version isn't too old.) not to hard but commenting here to flag so i remember to keep an eye on this!

alexcrichton:
I've filed https://github.com/rustwasm/wasm-pack/issues/597 to track this as well

PR #1371, opened 2019-03-20
Merge TypedArray definitions into a macro

RReverser:
This merges all of the `TypedArray` descendants into the existing `arrays!` macro.

This makes it easier to ensure they are all supported in a uniform manner (have the same set of methods) and to add future methods or `TypedArray` instances.


alexcrichton:
Thanks!

PR #1372, opened 2019-03-20
Fix docs for DataView::get*int32*

RReverser:
Looks like a copy-paste error. Probably in the future it would be nice to deduplicate these like in #1371 to reduce repetitions and likelihood of similar mistakes.


alexcrichton:
Oops, thanks!

PR #1380, opened 2019-03-21
Rewrite the parallel raytracing example with `rayon`

### LINKED ISSUE #1349: Use rayon for the raytrace-parallel example
Currently the multithreading example we have with raytracing uses a home-grown solution for parallelization, but we should be showing off `rayon`!

I've got a [work in progress](https://github.com/alexcrichton/wasm-bindgen/tree/rayon-example) branch for gaming this out, but there's a few notable points:

* Supporting this depends on [changes to Rayon](https://github.com/rayon-rs/rayon/pull/636), which we're still proving out on our end.
* Currently the demo is crashing a lot due to lock contention for memory allocation on the main thread, and this is hoped to be patched [in the upstream standard library](https://github.com/rust-lang/rust/pull/58855)
* The demo was seemingly crashing for other reasons, but I figured it'd be easier to get to the bottom of it once the allocator issue was solved.

In any case wanted to make sure we had a tracking issue for this!

alexcrichton:
One of the best parts about concurrency in Rust is using `rayon` and how
easy it makes parallelization of tasks, so it's the ideal example for
parallel Rust on the web! Previously we've been unable to use `rayon`
because there wasn't a way to customize how rayon threads themselves are
spawned, but [that's now being developed for us][rayon]!

This commit uses that PR to rewrite the `raytrace-parallel` example in
this repository. While not a perfect idiomatic representation of using
`rayon` I think this is far more idiomatic than the previous iteration
of `raytrace-parallel`! I'm hoping that we can continue to iterate on
this, but otherwise show it off as a good example of parallel Rust on
the web.

Closes #1349

[rayon]: https://github.com/rayon-rs/rayon/pull/636

### DIFF
--- a/examples/raytrace-parallel/Cargo.toml
+++ b/examples/raytrace-parallel/Cargo.toml
@@ -11,6 +11,8 @@ crate-type = ["cdylib"]
 console_error_panic_hook = "0.1"
 futures = "0.1"
 js-sys = "0.3.22"
+rayon = "1.1.0"
+rayon-core = "1.5.0"
 raytracer = { git = 'https://github.com/alexcrichton/raytracer', branch = 'update-deps' }
 wasm-bindgen = { version = "0.2.45", features = ['serde-serialize'] }
 wasm-bindgen-futures = "0.3.22"
--- a/examples/raytrace-parallel/src/lib.rs
+++ b/examples/raytrace-parallel/src/lib.rs
@@ -1,25 +1,21 @@
-use std::cell::RefCell;
-use std::cmp;
-use std::rc::Rc;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering::SeqCst};
-use std::sync::{Arc, Mutex, MutexGuard};
-
-use futures::sync::oneshot;
 use futures::Future;
-use js_sys::{Array, Error, Promise, Uint8ClampedArray, WebAssembly};
+use js_sys::{Promise, Uint8ClampedArray, WebAssembly};
+use rayon::prelude::*;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
-use web_sys::{CanvasRenderingContext2d, ErrorEvent, Event, Worker};
-use web_sys::{DedicatedWorkerGlobalScope, MessageEvent};
 
 macro_rules! console_log {
-    ($($t:tt)*) => (log(&format_args!($($t)*).to_string()))
+    ($($t:tt)*) => (crate::log(&format_args!($($t)*).to_string()))
 }
+
+mod pool;
 
 #[wasm_bindgen]
 extern "C" {
     #[wasm_bindgen(js_namespace = console)]
     fn log(s: &str);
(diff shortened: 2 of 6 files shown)


alexcrichton:
I think we'll want to hold off on merging this until the upstream rayon PR is merged, but I wanted to post this for others interested!

examples/raytrace-parallel/src/lib.rs
```diff
@@ -41,304 +37,124 @@ impl Scene {
...
+        // future is resolved we can pull out the final version of the image.
+        let done = pool
+            .run_notify(move || {
+                thread_pool.install(|| {
+                    rgb_data
+                        .par_chunks_mut(pixel_granularity_to_parallelize * 4)
+                        .enumerate()
+                        .for_each(|(i, chunk)| {
+                            for (j, chunk) in chunk.chunks_mut(4).enumerate() {
+                                let i = (i * pixel_granularity_to_parallelize + j) as u32;
```
cuviper: I think you could accomplish the same granularity with less munging like this:

```rust
rgb_data
    .par_chunks_mut(4)
    .with_min_len(pixel_granularity_to_parallelize)
    .enumerate()
    .for_each(|(i, chunk)| { ... });
```

So `with_min_len` sets a lower bound on how far rayon may split the work for parallelism. The inner serial work will be a batch of *at least* that length of items. The `for_each` will still see the individual chunk-4 pixels you actually care about.

Setting a minimum length isn't always needed, but rayon tries to be adaptive and split further when stuff is getting stolen a lot. Sometimes this gets overaggressive, especially at the tail end of a computation when they steal/fight to complete the last bits.

alexcrichton: Oh nice, thanks!

TBH It only seems marginally better to do something like `with_min_len(100)` here and the runtime in a browser seems to oscillate between 1800ms and 1100ms locally but it seems to oscillate a little less using `with_min_len(100)`. I think in the end given how easy it is to tack this on to rayon I'll probably just leave this out

alexcrichton:
And now updated with the merged API on https://github.com/rayon-rs/rayon/pull/636!

r? @fitzgen

alexcrichton:
Or actually you've already approved so I'll merge when green

examples/raytrace-parallel/Cargo.toml
```diff
@@ -11,6 +11,8 @@ crate-type = ["cdylib"]
 console_error_panic_hook = "0.1"
 futures = "0.1"
 js-sys = "0.3.22"
+rayon = "1.0.3"
+rayon-core = "1.0.3"
```
cuviper: I know you're patching to git anyway, but `spawn_handler` will need rayon-1.1 / rayon-core-1.5.
(I'm working on those release notes now...)

alexcrichton: Oh nice! We'll just wait for those in that case :)

cuviper: These are now published!

alexcrichton:
Thanks so much again for the help in `rayon` @cuviper!

PR #1391, opened 2019-03-25
Add specialised passStringToWasm for Node.js

RReverser:
Node.js doesn't currently implement `TextEncoder::encodeInto`. I've raised an upstream issue to add it - https://github.com/nodejs/node/issues/26904 - but it's likely to take some time and will be available only in new releases.

In the meanwhile, it's worth noting that Node.js already has `Buffer::write` which has pretty similar semantics, but doesn't require creating an intermediate view using `.subarray` and instead accepts pointer and length directly.

Also, Node.js has `Buffer::byteLength` helper which allows to efficiently retrieve an encoded byte length of a string upfront, and so allows us to avoid a loop with reallocations.

This change takes leverage of these methods by generating an additional Buffer-based view into the WASM memory and using it for string operations.

I'm seeing up to 35% increase in performance in string-heavy library benchmarks.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1236,8 +1236,6 @@ impl<'a> Context<'a> {
             return Ok(());
         }
         self.require_internal_export("__wbindgen_malloc")?;
-        self.expose_text_encoder();
-        self.expose_uint8_memory();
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

Issue #1386, opened 2019-03-22
Investigate caching of static strings


RReverser:
### Motivation

Long ago I noticed that currently even constant strings are routed through entire encoding/decoding process on each pass between Rust and JS, becoming unnecessarily expensive part of the resulting code.

Now I've noticed that this affects even internal functions which need intermediate strings, thus affecting even more potentially hot paths.

E.g. this is a profile from my WASM library that heavily relies on JS iteration - `js_sys::try_iter` - which, in turn, recreated same constant string for the iteration protocol over and over.

![image](https://user-images.githubusercontent.com/557590/54824791-dd701080-4ca3-11e9-8e6a-f0cb21341a23.png)

We can fix this particular function but, ideally, we should have a more generic solution for constant strings.

### Proposed Solution

Either:

1. Add a step to the postprocessor to walk over the generated WASM, detect calls to `JsValue::from_str` with pointers to a data section and statically extract these into JS.
2. Do the same in runtime by using a cache backed by `lazy_static!` (but then we need to somehow detect whether a string is in the data section, which is a bit harder in runtime but not impossible).
3. Add some sort of a proc-macro `js_const_str!("...")` that would extract these statically like (1), although could make it easier to detect by being more expicit.

All of the proposed solutions require choosing the right balance between complexity of the solution itself vs complexity of usage from API point of view (we don't want to add a function that most users would forget to use).

RReverser:
I specialised `try_iter` in my code to try `Array::is_array` + `unchecked_ref::<Array>().values().into_iter()` first, and the whole library benchmark got 5x faster 😱

I wonder how many more cases are affected by these hidden costs.

alexcrichton:
Thanks for the report! Could you expand a bit on how `try_iter` is connected to static strings in this case? Those seem to be naively unrelated, but hey 5x improvements are always nice

RReverser:
I was surprised about that connection showing up in profile as well, but from looking at the source it might be related to `JsValue::from("next")` being created each time you iterate over a new value?

alexcrichton:
Oh I think that's fixable with a change like:

```diff
diff --git a/crates/js-sys/src/lib.rs b/crates/js-sys/src/lib.rs
index 0a51ec7d..05a5a656 100644
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
[...]

PR #1393, opened 2019-03-26
Remove string allocation from try_iter

RReverser:
This allows to significantly speed up iteration over small collections, where string encoding is the primary overhead.

Related to #1386, but works around only this partial case.


RReverser:
Note that it's still faster to check for `is_array` and specialise for it, but I think that's explainable by number of roundtrips `try_iter` has to do to JS land to be generic.

alexcrichton:
👍

PR #1397, opened 2019-03-26
Simplify ABI for Option<char>

RReverser:
Flatten None to a special u32 value instead of using an intermediate pointer.


RReverser:
The failed check looks like a spurious deployment script failure?

alexcrichton:
Thanks!

programmerjake:
it might be better to pick a different value for the `None` case, such as `~0`, since it could reduce compiled code size when JIT compiled for RISC architectures, since they can usually easily represent -1 as an immediate value, but may have to use multiple instructions or a load instruction to represent larger values.

PR #1400, opened 2019-03-26
Add versions to crates.io patches

LegNeato:
Without this running `cargo build` in the top-level directory on nightly
produced an error for me:

```
error: failed to resolve patches for `https://github.com/rust-lang/crates.io-index`

Caused by:
  patch for `wasm-bindgen-futures` in `https://github.com/rust-lang/crates.io-index` did not resolve to any crates. If this is unexpected, you may wish to consult: https://github.com/rust-lang/cargo/issues/4678
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ members = [
 exclude = ['crates/typescript']
 
 [patch.crates-io]
-wasm-bindgen = { path = '.' }
-wasm-bindgen-futures = { path = 'crates/futures' }
-js-sys = { path = 'crates/js-sys' }
-web-sys = { path = 'crates/web-sys' }
+wasm-bindgen = { path = '.', version = '0.2.40' }
+wasm-bindgen-futures = { path = 'crates/futures', version = '0.3.17' }
+js-sys = { path = 'crates/js-sys', version = '0.3.17' }
+web-sys = { path = 'crates/web-sys', version = '0.3.17' }


alexcrichton:
Thanks for this! Unfortunately though this is a bug in Cargo (alluded by the error message). I'd prefer to avoid the versions here because they'll need to also be updated on publication which we don't always remember, and in general version on `[patch]` is somewhat unidiomatic.

I've been meaning to look closer at the bug in Cargo though, so maybe we can try to prioritize that sooner!

LegNeato:
Ah, cool. Wasn't sure where the issue was. I'll poke at cargo to see if I can take a crack at fixing, thanks for the quick response.

alexcrichton:
Oh if you're interested in poking around Cargo, that'd be awesome!

One of the reasons this bug has eluded us for so long I think is that we haven't been able to get a good test case which reproduces the issue. I'd only seen it sporadically locally every now and then! I think though with wasm-bindgen I've learned that steps like this will reproduce the bug:

* Have package A depend on B from crates.io.
* Add a `[patch]` for B pointing at a local `path` dependency
* Create a lock file.
* Bump versions of A and B (and the version requirement from A to B)
* Attempt to create a lock file -- this should succeed, but I believe it fails

That should help narrow down where the issue is cropping up in Cargo. I think that this is probably an instance of an error that just needs to be ignored in Cargo, we just need to understand where we'd ignore it!

If you need help poking around Cargo, feel free to reach out as well!

PR #1401, opened 2019-03-27
Fix handling of `u32` between Rust and JS

### LINKED ISSUE #1388: u32 interpreted as i32 when passed to JS
### Describe the Bug

A valid u32 larger than _2^31_ turns negative when passed to JavaScript.

### Steps to Reproduce

1. Take the `import_js` example.
2. Change the `x.set_number(10)` line to `x.set_number(2147483647)` or any other u32 larger than _2^31_.
3. See the negative number in the console.

### Expected Behavior

I expect to see a positive number.

### Actual Behavior

We see a negative number.

### Additional Context

Tested on wasm-bindgen 0.2.40.

alexcrichton:
All numbers in WebAssembly are signed and then each operation on them
may optionally have an unsigned version. This means that when we pass
large signed numbers to JS they actually show up as large negative
numbers even though JS numbers can faithfully represent the type.

This is fixed by adding `>>>0` in a few locations in the generated
bindings to coerce the JS value into an unsigned value.

Closes #1388

### DIFF
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -99,6 +99,10 @@ pub enum VectorKind {
     Anyref,
 }
 
+pub struct Number {
+    u32: bool,
+}
+
 impl Descriptor {
     pub fn decode(mut data: &[u32]) -> Descriptor {
         let descriptor = Descriptor::_decode(&mut data);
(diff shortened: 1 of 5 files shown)


crates/cli-support/src/descriptor.rs
```diff
@@ -149,18 +149,21 @@ impl Descriptor {
...
             | Descriptor::I16
             | Descriptor::U16
             | Descriptor::I32
-            | Descriptor::U32
             | Descriptor::F32
-            | Descriptor::F64
-            | Descriptor::Enum { .. } => true,
-            _ => return false,
+            | Descriptor::F64 => Some(false),
+            Descriptor::U32 => Some(true),
```
Pauan: Why is this missing a `|`?

crates/cli-support/src/descriptor.rs
```diff
@@ -149,18 +149,21 @@ impl Descriptor {
         }
     }
 
-    pub fn is_number(&self) -> bool {
+    /// Returns `Some` if this type is a number, and the boolean returned
```
Pauan: I think it'd be nice if it returned a custom `enum` rather than `Option<bool>`. I don't feel too strongly about this though.

tests/wasm/math.rs
```diff
@@ -4,6 +4,20 @@ use wasm_bindgen_test::*;
 #[wasm_bindgen(module = "tests/wasm/math.js")]
 extern "C" {
     fn js_auto_bind_math();
+
+    #[wasm_bindgen(js_name = roundtrip)]
```
Pauan: I think it's a little subtle why these use `f64`, so a comment explaining it would be nice.

crates/cli-support/src/descriptor.rs
alexcrichton: Oh true, I've reorganized a bit!

crates/cli-support/src/descriptor.rs
alexcrichton: 👍

PR #1403, opened 2019-03-27
Add `constructor` property to `Object`

### LINKED ISSUE #1361: `Object` bindings are missing `constructor` property
`Object.prototype.constructor` is an important property when you need to check whether an object was created by a particular constructor and not just inherits it somewhere in the prototype chain.

RReverser:
Fixes #1361, although shortly after adding this I realised that to be useful wasm-bindgen also needs to expose a way to get constructor of any type as a `Function`, but not sure what is the best way to expose that yet...

Something like autogenerated `fn as_constructor() { ... }` on any `#[wasm_bindgen]` imported type maybe?

Anyway, that can probably be done in a separate PR.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1878,6 +1878,13 @@ extern "C" {
     pub fn assign3(target: &Object, source1: &Object, source2: &Object, source3: &Object)
         -> Object;
 
+    /// The constructor property returns a reference to the Object constructor
+    /// function that created the instance object.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor)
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks!

I think that with `Deref` you can already do `Object::constructor(&any_wasm_bindgen_type)`, right?

RReverser:
> I think that with `Deref` you can already do `Object::constructor(&any_wasm_bindgen_type)`, right?

Well you can do even `any_value.constructor()`, yes, but the problem is that you can't do comparisons like `any_value.constructor() == Array::constructor_as_a_function_or_something_like_that()` because there is no method to get a constructor of a statically defined type.

RReverser:
Created a separate issue for tracking https://github.com/rustwasm/wasm-bindgen/issues/1406

Issue #442, opened 2018-07-10
Add WebSocket and web-sys example to `examples/` and the guide


fitzgen:
General info on web sockets: https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API

* [ ] Add an example to `examples/` that shows off working with web sockets
* [ ] Add a subsection to "web-sys >> examples" in the guide that includes this example

PR #1404, opened 2019-03-28
Added WebSocket Example

xarus01:
#442


alexcrichton:
Thanks for this! I wonder if it would be possible to transfer more of the logic to Rust though? Right now it looks like the workhorse of the example's logic is all in JS, and to show off Rust usage it might be cool to have the majority of the logic live in Rust

alexcrichton:
I'm gonna go ahead and close this, but feel free to resubmit with my coment addressed!

Issue #1366, opened 2019-03-19
Support prototype-less objects


RReverser:
This would be a breaking change, but, I think, `Object.keys`, `Object.entries`, `Object.setPrototypeOf` etc. need to be changed to somehow support prototype-less objects.

In JavaScript it's common to have object values without prototype used as dictionaries created by either `Object.create(null)` or a corresponding literal form `{ __proto__: null, ... }`.

These don't have `Object` in their prototype chain, so you can't call methods from `Object.prototype` on them (e.g. `.to_string()` won't work), and `dyn_ref` fails as expected.

However, you should still be able to call `Object.keys`, `Object.values`, `Object.seal` and other static methods, because they support such objects as well as any others.

So we need to somehow separate these two types to allow arbitrary objects in static methods, and only prototyped objects in methods that accept `&self` / `&this`.

RReverser:
@alexcrichton I've also just noticed that current implementation of `Object::try_from` is also wrong due to this issue (it assumes that any object inherits from `Object` prototype, but it's not true).

alexcrichton:
I'll admit that I don't think I know enough about JS to know what to do about this issue, but a PR would be most welcome to fix the state of affairs!

RReverser:
I guess the hardest part of this would be to figure out naming in a way that wouldn't be too confusing (in JS it's `object` as a value type vs `Object` as a class).

Pauan:
I believe the correct check for whether something is an object or not is `Object(foo) === foo` (note the *lack* of `new`, which is very important!)

RReverser:
That performs unnecessary boxing of primitives and I'm not sure what you're gaining over current `JsValue::is_object`.

My suggestion was only about splitting interfaces, but keeping the checks as they already are.

Pauan:
@RReverser There's some things which are objects but aren't `typeof "object"`, such as regexps (in some browsers), and functions.

So at the very least, it would need to be changed to a blacklist (anything that isn't a primitive is assumed to be an object).

But that has its own share of issues, such as if ECMAScript adds in a new primitive (which they did, with `Symbol`).

So `Object(foo) === foo` is the most reliable way of detecting objects, since it actually uses the [`ToObject` primitive](https://tc39.github.io/ecma262/#sec-toobject).

If your concern is about the boxing, it can do a `typeof` check first, and only do the `Object(foo)` check if it isn't a primitive.

[...]

Issue #893, opened 2018-09-26
Figure out how to support interfaces with `NoInterfaceObject` attribute


fitzgen:
Right now, we completely skip generating code for any interface that is marked `[NoInterfaceObject]`. I think we might want to do something like:

* mark all their methods `structural`, since we don't ahve the prototype until we get an instance
* make `JsCast::is_instance_of` always return false for them, because we don't have access to the constructor without getting an instance

cc @alexcrichton @afdw 

### Here are the 71 interfaces with `[NoInterfaceObject]` in our webidls

- `ANGLE_instanced_arrays`
- `Coordinates`
- `DeviceAcceleration`
- `DeviceRotationRate`
- `Exception`
- `EXT_blend_minmax`
- `EXT_color_buffer_float`
- `EXT_color_buffer_half_float`
- `EXT_disjoint_timer_query`
- `External`
- `EXT_frag_depth`
- `EXT_shader_texture_lod`
- `EXT_sRGB`
- `EXT_texture_filter_anisotropic`
- `FontFaceSetIterator`
- `Geolocation`
- `GetUserMediaRequest`
- `HTMLHyperlinkElementUtils`
- `ImageCaptureError`
- `IntlUtils`
- `ListBoxObject`
- `MediaStreamError`
- `MOZ_debug`
- `NavigatorAutomationInformation`
- `OES_element_index_uint`
- `OES_standard_derivatives`
- `OES_texture_float`
- `OES_texture_float_linear`
- `OES_texture_half_float`
- `OES_texture_half_float_linear`
- `OES_vertex_array_object`
- `Position`
- `PositionError`
- `PromiseNativeHandler`
- `RTCIdentityProviderRegistrar`
- `ScrollBoxObject`
- `SVGPathSeg`
- `SVGPathSegArcAbs`
- `SVGPathSegArcRel`
- `SVGPathSegClosePath`
- `SVGPathSegCurvetoCubicAbs`
- `SVGPathSegCurvetoCubicRel`
- `SVGPathSegCurvetoCubicSmoothAbs`
- `SVGPathSegCurvetoCubicSmoothRel`
- `SVGPathSegCurvetoQuadraticAbs`
- `SVGPathSegCurvetoQuadraticRel`
- `SVGPathSegCurvetoQuadraticSmoothAbs`
- `SVGPathSegCurvetoQuadraticSmoothRel`
- `SVGPathSegLinetoAbs`
[...]

PR #1405, opened 2019-03-28
Add support for customising `instanceof` behaviour

### LINKED ISSUE #1367: Specialise `dyn_ref` to improve value detection
I think `dyn_ref` should be specialised where possible to support objects from other realms (for which regular `instanceof` check would fail).

For example, it's more reliable to use `Array::is_array` instead of current `instanceof Array` for `.dyn_ref::<js_sys::Array>()`, `JsValue::is_object` for `.dyn_ref::<Object>()` and probably also existing methods for primitives for conversions to `js_sys::Number`, `js_sys::JsString` and so on.

### LINKED ISSUE #1370: `.dyn_into::<Symbol>` / `.dyn_ref::<Symbol> don't work as expected
This is sort of a child issue to #1367, but, because it's a bug, it seems worth it to track separately since it might be fixed independently from the more generic one.

---

While looking into #1087, I realised that #1367 is needed not only for prototype-less `Object`s and cross-realm interaction, but also for new primitives which don't have automatic boxing.

Currently these are:

 - `Symbol`
 - `BigInt`

and for both `instanceof` check will return false when used on a corresponding primitive, because they're not objects, and so `dyn_ref` and `dyn_into` fail to recognise them too.

Example:

```rust
#[wasm_bindgen]
pub fn symbol_to_string_1(s: Symbol) -> JsString {
	s.to_string()
}

#[wasm_bindgen]
pub fn symbol_to_string_2(s: JsValue) -> JsString {
	s.dyn_into::<Symbol>().unwrap().to_string()
}

#[wasm_bindgen]
pub fn symbol_to_string_3(s: JsValue) -> JsString {
	if s.is_symbol() {
		s.unchecked_into::<Symbol>().to_string()
	} else {
		panic!("Not a symbol")
	}
}
```

Results:

```js
> require('./pkg').symbol_to_string_1(Symbol('x'))
'Symbol(x)'
> require('./pkg').symbol_to_string_2(Symbol('x'))
panicked at 'called `Result::unwrap()` on an `Err` value: JsValue(Symbol(x))', src/libcore/result.rs:997:5
> require('./pkg').symbol_to_string_3(Symbol('x'))
'Symbol(x)'
```

RReverser:
This allows types to define custom Rust code as an override for `JsCast::instanceof`.

 - Fixes #1367 for types that have built-in cross-realm checks, allowing `dyn_ref` and `dyn_into` to accept values from other realms. In particular:
   - `.instanceof::<Array>()` will now automatically use `Array::is_array` instead of actual `... instanceof Array`.
   - `.instanceof::<Function>()` will use `JsValue::is_function`. This also simplifies `Function::try_from` to being an alias for `dyn_ref::<Function>()`.
   - `.dyn_into::<JsString>()`, `.dyn_into::<Number>()`, `.dyn_into::<Boolean>()` will also now use corresponding checks from `JsValue`.
     In theory, these 3 can break checks if someone has been using boxed primitives (`new String(...)`, `new Number(...)`, `new Boolean(...)`), but these are extremely rare in real-world JS code and are already not support by `JsValue` methods. If we do want to support them, I believe it should be done at the `JsValue` level in `as_f64`, `as_bool`, `as_string` too.
 - Fixes #1370 - `.instanceof::<Symbol>()` will now use `JsValue::is_symbol` - previously it would return `false` for any value, even actual `Symbol`s, making it impossible to use `.dyn_ref::<Symbol>()` and `.dyn_into::<Symbol>()`.

cc @alexcrichton @fitzgen

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -183,6 +183,7 @@ pub struct ImportType {
     pub attrs: Vec<syn::Attribute>,
     pub doc_comment: Option<String>,
     pub instanceof_shim: String,
+    pub is_type_of: Option<syn::Expr>,
     pub extends: Vec<syn::Path>,
     pub vendor_prefixes: Vec<Ident>,
 }
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -588,6 +588,13 @@ impl ToTokens for ast::ImportType {
             }
         };
 
+        let is_type_of = self.is_type_of.as_ref().map(|is_type_of| quote! {
+            fn is_type_of(val: &JsValue) -> bool {
+                let is_type_of: fn(&JsValue) -> bool = #is_type_of;
+                is_type_of(val)
+            }
+        });
+
         (quote! {
             #[allow(bad_style)]
             #(#attrs)*
@@ -719,6 +726,8 @@ impl ToTokens for ast::ImportType {
                         drop(val);
                         panic!("cannot check instanceof on non-wasm targets");
                     }
+
+                    #is_type_of
 
                     #[inline]
                     fn unchecked_from_js(val: JsValue) -> Self {
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -126,7 +126,7 @@ extern "C" {
 // Array
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(extends = Object)]
+    #[wasm_bindgen(extends = Object, is_type_of = Array::is_array)]
     #[derive(Clone, Debug, PartialEq, Eq)]
     pub type Array;
 
@@ -466,7 +466,7 @@ extern "C" {
 // Boolean
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(extends = Object)]
+    #[wasm_bindgen(extends = Object, is_type_of = |v| v.as_bool().is_some())]
     #[derive(Clone, PartialEq, Eq)]
     pub type Boolean;
 
@@ -801,7 +801,7 @@ extern "C" {
 // Function
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(extends = Object)]
+    #[wasm_bindgen(extends = Object, is_type_of = JsValue::is_function)]
     #[derive(Clone, Debug, PartialEq, Eq)]
     pub type Function;
 
@@ -897,11 +897,7 @@ impl Function {
     /// If this JS value is not an instance of a function then this returns
     /// `None`.
     pub fn try_from(val: &JsValue) -> Option<&Function> {
-        if val.is_function() {
-            Some(val.unchecked_ref())
-        } else {
-            None
-        }
+        val.dyn_ref()
     }
 }
 
@@ -1141,7 +1137,10 @@ pub fn try_iter(val: &JsValue) -> Result<Option<IntoIter>, JsValue> {
         return Ok(None);
     }
 
-    let iter_fn: Function = iter_fn.unchecked_into();
+    let iter_fn: Function = match iter_fn.dyn_into() {
+        Ok(iter_fn) => iter_fn,
+        Err(_) => return Ok(None)
+    };
     let it = iter_fn.call0(val)?;
     if !it.is_object() {
         return Ok(None);
@@ -1434,7 +1433,7 @@ extern "C" {
 // Number.
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(extends = Object)]
+    #[wasm_bindgen(extends = Object, is_type_of = |v| v.as_f64().is_some())]
     #[derive(Clone)]
     pub type Number;
 
@@ -3127,7 +3126,7 @@ extern "C" {
 // JsString
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(js_name = String, extends = Object)]
+    #[wasm_bindgen(js_name = String, extends = Object, is_type_of = JsValue::is_string)]
     #[derive(Clone, PartialEq, Eq)]
     pub type JsString;
 
@@ -3587,11 +3586,7 @@ impl JsString {
     /// If this JS value is not an instance of a string then this returns
     /// `None`.
     pub fn try_from(val: &JsValue) -> Option<&JsString> {
-        if val.is_string() {
-            Some(val.unchecked_ref())
-        } else {
-            None
-        }
+        val.dyn_ref()
     }
 
     /// Returns whether this string is a valid UTF-16 string.
@@ -3683,6 +3678,7 @@ impl fmt::Debug for JsString {
 // Symbol
 #[wasm_bindgen]
 extern "C" {
+    #[wasm_bindgen(is_type_of = JsValue::is_symbol)]
     #[derive(Clone, Debug)]
     pub type Symbol;
 
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -45,6 +45,7 @@ macro_rules! attrgen {
             (readonly, Readonly(Span)),
             (js_name, JsName(Span, String, Span)),
             (js_class, JsClass(Span, String, Span)),
+            (is_type_of, IsTypeOf(Span, syn::Expr)),
             (extends, Extends(Span, syn::Path)),
             (vendor_prefix, VendorPrefix(Span, Ident)),
             (variadic, Variadic(Span)),
@@ -236,6 +237,11 @@ impl Parse for BindgenAttr {
             });
 
             (@parser $variant:ident(Span, syn::Path)) => ({
+                input.parse::<Token![=]>()?;
+                return Ok(BindgenAttr::$variant(attr_span, input.parse()?));
+            });
+
+            (@parser $variant:ident(Span, syn::Expr)) => ({
                 input.parse::<Token![=]>()?;
                 return Ok(BindgenAttr::$variant(attr_span, input.parse()?));
             });
@@ -515,6 +521,7 @@ impl ConvertToAst<BindgenAttrs> for syn::ForeignItemType {
             .js_name()
             .map(|s| s.0)
             .map_or_else(|| self.ident.to_string(), |s| s.to_string());
+        let is_type_of = attrs.is_type_of().cloned();
         let shim = format!("__wbg_instanceof_{}_{}", self.ident, ShortHash(&self.ident));
         let mut extends = Vec::new();
         let mut vendor_prefixes = Vec::new();
@@ -537,6 +544,7 @@ impl ConvertToAst<BindgenAttrs> for syn::ForeignItemType {
             attrs: self.attrs,
             doc_comment: None,
             instanceof_shim: shim,
+            is_type_of,
             rust_name: self.ident,
             js_name,
             extends,
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -514,6 +514,7 @@ impl<'src> FirstPassRecord<'src> {
             attrs,
             doc_comment: None,
             instanceof_shim: format!("__widl_instanceof_{}", name),
+            is_type_of: None,
             extends: Vec::new(),
             vendor_prefixes: Vec::new(),
         };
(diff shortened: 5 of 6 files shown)


RReverser:
> In theory, these 3 can break checks if someone has been using boxed primitives

Heh, looks like some tests actually check this.

RReverser:
It also seems strangely inconsistent that `Boolean` and `Number` have `fn new` bindings to create a boxed primitive, but `JsString` doesn't... @alexcrichton any chance we could soft-deprecate these on existing primitives?

fitzgen:
Ideally we would have some raw layer that just does the `instanceof` check and matches JS semantics *exactly*, and then on top of that we would add some sort of "is-a?" predicate that Does What I Mean.

I'm not sure how to do this in a backwards compat way.

RReverser:
> I'm not sure how to do this in a backwards compat way.

We can surely separate them by adding

```rust
trait JsCast {
  // ...existing definitions

  fn is_a(val: &JsValue) -> bool {
    Self::instanceof()
  }
}
```

and then overriding this extra method via attribute, but I see little benefit in that because `dyn_ref`, `dyn_into` etc. shouldn't blindly use `instanceof` IMO...

alexcrichton:
Can we perhaps draw inspiration from JS somehow? How does JS typically handle these differences idiomatically? Can we basically make the idiomatic thing in JS the same idiomatic thing in Rust?

RReverser:
> Can we basically make the idiomatic thing in JS the same idiomatic thing in Rust?

That's what this PR attempts to do.

Unfortunately, despite few proposals in the past, there is no central API for cross-realm checks, but the common consensus is that you should use them where available (like in case with `typeof` or  `Array.isArray` or Node's `Buffer.isBuffer` etc.).

For other objects, which don't have corresponding method, the common alternative is to use string tag checks like `Object.prototype.toString.call(...) === '[object ...]'` but for now I decided not to go there as it adds potentially heavy runtime cost, whereas built-ins are free.

Pauan:
@RReverser is correct: there are many ways to check whether an object is a particular type, so there isn't a single "correct" way to do it in JS.

Instead you need to use `typeof` or `instanceof` or `{}.prototype.toString.call` depending on the situation. So I agree with him that being able to customize the behavior of `instanceof` is needed.

However, since `instanceof` is an actual operator in JS, I think it would be *far* less confusing if it was renamed to something else, like `is_a`. That avoids any confusion with the `instanceof` operator. And then we should deprecate the `instanceof` method.

alexcrichton:
I think that sounds like the best of all worlds yeah where we:

* Add a new method, `is_a` to `JsCast`
  * For now, make it a default method calling `instanceof` for backwards compatibility reasons
* Deprecate `instanceof`, recommending `is_a`
* Change all `dyn_*` methods to use `is_a`
* Allow customizing `is_a` (as this PR does)
* Take `is_a` customizations in this PR for Array/String etc

How's that sound?

fitzgen:
That plan sounds great to me.

In an ideal world, I'd change the semantics of `JsCast::instanceof` be exactly JS `instanceof`, but I think that is backwards incompatible.

RReverser:
@alexcrichton Sounds perfectly reasonable to me! I'll update the PR.

RReverser:
> In an ideal world, I'd change the semantics of JsCast::instanceof be exactly JS instanceof, but I think that is backwards incompatible.

@fitzgen I think that's what @alexcrichton suggests too? (keeping `instanceof` as an actual JS `instanceof` for now)

fitzgen:
We implement `JsCast::instanceof` for some things like `JsValue` that have no JS equivalent.

RReverser:
> JsCast::instanceof for some things like JsValue that have no JS equivalent.

Ohh. Yeah, I think these would have to be deprecated.

RReverser:
Btw, do we want to actually call it `is_a` or just `is` would be enough? (just to avoid `a`/`an` confusion)

Pauan:
> @fitzgen In an ideal world, I'd change the semantics of JsCast::instanceof be exactly JS instanceof, but I think that is backwards incompatible.

Personally, I don't see any use for `JsCast::instanceof`, it's almost never what you want.

On the other hand, a `JsValue::instanceof` *would* be useful. It would just use the JS `instanceof` operator, and nothing else:

```rust
impl JsValue {
    pub fn instanceof(&self, class: &JsValue) -> bool {
        // intrinsic which uses `instanceof` in JS
    }
}
```

This would be useful as a low-level building block, but isn't intended to be used directly (instead `JsCast::is_a` is preferred).

RReverser:
> On the other hand, a JsValue::instanceof would be useful. It would just use the JS instanceof operator, and nothing else:

I was thinking about it too, but for that we need to solve #1406 first (I raised it with this particular issue in mind).

Pauan:
@RReverser Right, but `JsValue::instanceof` can be added in a separate PR, so it isn't really related to this PR, so we can decide that later.

My point was basically just that `JsCast::instanceof` isn't needed, so we can just deprecate it.

RReverser:
> My point was basically just that JsCast::instanceof isn't needed, so we can just deprecate it.

Maybe better to deprecate in the same PR as `JsValue::instanceof` is added (in case someone does want a stable way to do that, for whatever weird reason)?

RReverser:
Also, as far as I can tell, for now we don't want to use `O.p.toString` on types without real cross-realm checks so we still need `instanceof` for now as a fallback.

Pauan:
@RReverser Except right now `JsCast::instanceof` *doesn't* always use `instanceof`, it's a weird mishmash of some things. So I'm fine with deprecating it in the same release as `JsCast::is`

> Also, as far as I can tell, for now we don't want to use O.p.toString on types without real cross-realm checks so we still need instanceof for now as a fallback.

Why don't we want to use `toString`?

RReverser:
> Why don't we want to use toString?

See above:
1) @alexcrichton: > For now, make it a default method calling instanceof for backwards compatibility reasons
2) me: > for now I decided not to go there as it adds potentially heavy runtime cost, whereas built-ins are free.

Aside from perf impact, toString is also pretty unreliable, although for different reasons.

RReverser:
> Except right now JsCast::instanceof doesn't always use instanceof, it's a weird mishmash of some things.

Aside from `JsValue`, when else it doesn't use `instanceof`?

Pauan:
> Aside from JsValue, when else it doesn't use instanceof?

I think just `JsValue`.

> for now I decided not to go there as it adds potentially heavy runtime cost, whereas built-ins are free.

I created some benchmarks: https://jsperf.com/array-type-checking3

The builtin gets ~32,891,703 ops/sec. Using `toString` gets ~22,678,401 ops/sec.

A cost? Sure, but that's still *22 million* ops/sec. So in the situations where a builtin doesn't exist, correctness is more important in my opinion.

> toString is also not 100% reliable, although for different reasons.

This is the first time I've heard of that, could you explain more?

alexcrichton:
@RReverser trying to follow the discussion, were you on board with [the previous plan](https://github.com/rustwasm/wasm-bindgen/pull/1405#issuecomment-478675721)? I was looking at https://github.com/rustwasm/wasm-bindgen/issues/893 again today and realized that this'd be a great way to fix that, so was curious if you're still up for pursuing this change!

RReverser:
> trying to follow the discussion, were you on board with the previous plan?

I was :) See https://github.com/rustwasm/wasm-bindgen/pull/1405#issuecomment-478679338

There have been just few follow-up questions which can be summarised as:
1) Can we use a more generic `is` instead of `is_a` / `is_an`?
2) Do we want to switch to `Object.prototype.toString` approach for everything else or do we keep `instanceof` as a default behaviour for now?

alexcrichton:
Since this is in theory not called too much (rather the `dyn_*` methods would be preferred) the name doesn't matter too too much so I'd be wary of taking such a small name as `is`, but I also don't have much against it. If we're reframing this as "this is a loose check to see if the types are compatible and will work together" then I think it's fine for us to change the internal behavior at any time, so for now I'd probably stick with `instanceof` and if necessary we could switch to `Object.prototype.toString`

RReverser:
> so for now I'd probably stick with instanceof and if necessary we could switch to Object.prototype.toString

Oh, actually, this brings me back to another question from the beginning of this PR - do we really want to continue to support boxed variants of `Number` and `Boolean`? If so, current change to `typeof` will break them (because for boxed variants it returns `object` as expected).

Given that we don't support boxed variant of `JsString` and in general boxing primitives into objects is usually a bad idea and rarely happens in real-world code, I'd suggest to soft-deprecate `fn new` for these two types, but the final decision is not up to me.

alexcrichton:
So long as "general idiomatic patterns continue to work" I don't have strong feelings one way or another. I don't really know what it means to not support them or to to support them.

Pauan:
@alexcrichton In JS there is a distinction between primitive and boxed values:

* `"foo"` is a primitive string.
* `new String("foo")` is a boxed string.

Most of the time they behave the same, but there are some differences:

```js
typeof "foo" === "string"
typeof new String("foo") === "object"
```

```js
"foo" instanceof String === false
new String("foo") instanceof String === true
```

On the other hand, these are both true:

```js
{}.toString.call("foo") === "[object String]"
{}.toString.call(new String("foo")) === "[object String]"
```

So you can use that to reliably detect both.

Right now, since `JsString` uses `instanceof`, that means that it's not actually possible to use string primitives like `"foo"` as a `JsString`.

If it used `typeof` then it would *only* work with primitive strings.

If it used `{}.toString.call` then it would work with both.

Boxed strings are *very* rare in JS. So using either `typeof` or `{}.toString.call` is fine, but using `instanceof` is not okay.

Since `{}.toString.call` supports both, it may seem like the obvious choice, but [it does have a bit of a performance cost](https://github.com/rustwasm/wasm-bindgen/pull/1405#issuecomment-478787912), so using `typeof` may be better.

On the other hand, some types in JS don't work with `typeof`, so in those cases I recommend using `{}.toString.call`, since it's the most reliable method I know of for type detection (and it works cross-realm).

RReverser:
Sorry for the delay here - I think we agreed with the plan for this PR for now, I just need few days to finish as working on few projects in parallel :) (as far as I understand, there is no rush since no one raised these issues before anyway, so hopefully few days is fine)

RReverser:
Ok I fixed issues and decided to rename the method to `is_type_of` (with a signature `T::is_type_of(&JsValue) -> bool`) - @alexcrichton you said, it's unlikely to be called directly often and the longer name is preferable, and this seems to reflect what it does the best.

I've also added a default fallback to existing `instanceof` instead of replacing it altogether, as discussed above.

RReverser:
One more note: I haven't added custom check to `type Object` yet because it seems to be a breaking change and is tracked separately in #1366.

RReverser:
Noticed that, just like `Function::try_from`, `JsString::try_from` can also be updated to being an alias for `dyn_ref` instead of an unsafe code, so updated it too.

alexcrichton:
Nice! Could this perhaps switch the direction of the arguments though? Instead of `T::is_type_of(&obj)` could that be written as `obj.is_type_of::<T>()`? (sort of like `instanceof` today). That'd then require a seaparate method which isn't typically called by users but would be used to do the `T::some_other_name(&obj)`.

Additionally, could the documentation be updated to indicate when you might want to use `is_instance_of` vs `is_type_of`?

RReverser:
> That'd then require a seaparate method

Yeah this was the primary reason I didn't do it - from the discussion above I've got an impression that we don't want to encourage this method to be often called directly?

Also note that semantically it's in `obj.is_type_of::<T>()` that the name would have to be different, not in `T::is_type_of` -> `T::some_other_name`, because the direction of the name is rather opposite from `instanceof` ("T is type of value" vs "value is instance of T").

But, anyway, I guess we could add `obj.has_type::<T>`, the question is only whether we actually need it...

alexcrichton:
I don't really mind too much about the naming, but I suspect that we'll want to recommend that this is called almost all the time instead of `is_instance_of`. Most users will use `dyn_*` which won't matter much, but in some cases I suspect that `is_instance_of` is in used, and we'll want to make sure the recommended form is still of the form `self.method_name::<T>()` to ensure it reads left-to-right

RReverser:
Ok I can add `obj.has_type::<T>()` then.

RReverser:
(this was just a rebase to fix conflicts for now)

RReverser:
@alexcrichton Ok done and looks like CI passed.

alexcrichton:
Thanks @RReverser!

PR #1408, opened 2019-03-29
Implement Debug on Closures

### LINKED ISSUE #1387: Closure should impl Debug
### Motivation

When implementing https://github.com/rustwasm/gloo/issues/30, I created this type:

```rust
pub struct EventListener<'a> {
    target: EventTarget,
    kind: &'a str,
    callback: Option<Closure<FnMut(Event)>>,
}
```

I would really like to just slap a `#[derive(Debug)]` and call it a day, but `Closure` doesn't impl `Debug`, so I have to do this instead:

```rust
impl<'a> std::fmt::Debug for EventListener<'a> {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        f.debug_struct("EventListener")
            .field("target", &self.target)
            .field("kind", &self.kind)
            .field("callback", &"Closure { ... }")
            .finish()
    }
}
```

### Proposed Solution

I propose that `Closure` should provide a `Debug` impl, which essentially just outputs `Closure { ... }` (or similar).

### Alternatives

Maybe the `Debug` impl should expose more information? For example, it could call `.toString()` on the JS function.

samcday:
As per the discussion in #1387 I kept this one simple, the Debug output for a Closure is simply `Closure { ... }`. I don't think there's currently any more useful information we could provide anyway.

resolves #1387

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -4,6 +4,7 @@
 //! closures" from Rust to JS. Some more details can be found on the `Closure`
 //! type itself.
 
+use std::fmt;
 #[cfg(feature = "nightly")]
 use std::marker::Unsize;
 use std::mem::{self, ManuallyDrop};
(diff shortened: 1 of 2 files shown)


alexcrichton:
Awesome, thanks!

PR #1316, opened 2019-03-05
Add 'ignore' attribute to wasm_bindgen

tyleranton:
This PR adds `ignore` as an attribute for `#[wasm_bindgen]`. This makes it not generate getters/setters for a pub field marked with `#[wasm_bindgen(ignore)]`.

I'm not sure I did this 100% correct, but it is able to convert!

I'll make any changes/fixes necessary.

Thanks!

Fixes #1284

PR #1410, opened 2019-03-30
Add wasm_bindgen(skip) attribute

### LINKED ISSUE #1284: Add `#[wasm_bindgen(ignore)]` to ignore a `pub` field and not generate getters/setters for it
Hello!

I am attempting to use a HashMap inside a wrapper struct, but running into a conversion error.
Code to reproduce:
```rust
#[wasm_bindgen]
#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
pub struct Parser {
  pub count: usize,
  pub fields: HashMap<String, Field>,
}

#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
pub struct Field {
  pub name: String,
  pub count: usize,
}

#[wasm_bindgen]
impl Parser {
  #[wasm_bindgen(constructor)]
  #[wasm_bindgen(js_name="new")]
  pub fn wasm_new() -> Self {
    Parser {
      count: 0,
      fields: HashMap::new(),
    }
  }

  #[wasm_bindgen(js_name="toJson")]
  pub fn wasm_to_json(&mut self) -> JsValue {
    let field = Field {
      name: "field1".to_string(),
      count: 1
    };
    self.fields.insert("field1".to_string(), field);

    JsValue::from_serde(&self).unwrap()
  }
}
```
Running `wasm-pack build` causes this error:

```shell
  error[E0277]: the trait bound `std::collections::HashMap<std::string::String, std::string::String>: wasm_bindgen::convert::traits::IntoWasmAbi` is not satisfied
    --> src/lib.rs:11:1
     |
  11 | #[wasm_bindgen]
     | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::traits::IntoWasmAbi` is not implemented for `std::collections::HashMap<std::string::String, std::string::String>`

  error[E0277]: the trait bound `std::collections::HashMap<std::string::String, std::string::String>: wasm_bindgen::convert::traits::FromWasmAbi` is not satisfied
    --> src/lib.rs:11:1
     |
  11 | #[wasm_bindgen]
     | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::traits::FromWasmAbi` is not implemented for `std::collections::HashMap<std::string::String, std::string::String>`

```

I know  [the guide on Arbitrary Data Structures](https://rustwasm.github.io/wasm-bindgen/reference/arbitrary-data-with-serde.html) mentions to remove the `#[wasm_bindgen]` macro around the struct, but doing that also errors out. 
Adjusted `Parser` Struct (and everything else from example stays the same):
```rust
#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
pub struct Parser {
  pub count: usize,
  pub fields: HashMap<String, Field>,
}
```
Error:

```shell
  error[E0277]: the trait bound `Parser: wasm_bindgen::convert::traits::IntoWasmAbi` is not satisfied
    --> src/lib.rs:23:1
     |
  23 | #[wasm_bindgen]
     | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::traits::IntoWasmAbi` is not implemented for `Parser`
     |
     = note: required because of the requirements on the impl of `wasm_bindgen::convert::traits::ReturnWasmAbi` for `Parser`
```

Thank you for your time!

alexlapa:
Resolves #1284.
Previous PR  #1316 (closed).

 Adds `wasm_bindgen(skip)` attribute to disable public fields exposure to JS.

Renamed attribute from 'ignore' to 'skip' according to @LegNeato's [consideration](https://github.com/rustwasm/wasm-bindgen/issues/1284#issuecomment-470746290) to align naming with [serde's `#[serde(skip)]`](https://serde.rs/field-attrs.html#serdeskip).


alexcrichton:
Awesome, thanks for this!

Could you also add some documentation to the reference in `guide/src/reference`?

alexlapa:
@alexcrichton ,

> Could you also add some documentation to the reference in guide/src/reference?

Done.

alexcrichton:
Thanks!

Sayan751:
It seems that this attribute only works for struct fields. I think it also makes sense to enable application of this attribute for methods. I would expect when this attribute is applied to a method, it is ignored by wasm-bindgen. For example, it will let me write something like this:
```rust
impl MyStruct {

  #[wasm_bindgen(skip)]
  pub fn my_method() -> &str {
    //...
  }

}
```

The advantage being is that then I can keep all the methods, both exposed to WASM as well as to other Rust libs, inside a single `impl` block. This in turn generates nicer docs, by putting all methods inside a single `impl` block, w/o any superfluous section.

What you guys think?

alexcrichton:
Seems reasonable to me!

emccorson:
> It seems that this attribute only works for struct fields. I think it also makes sense to enable application of this attribute for methods
@Sayan751 Do you know if this was ever implemented? I was surprised to find skip doesn't work on methods and I came across this old issue.

PR #1412, opened 2019-03-31
Add TS type for init fn

### LINKED ISSUE #1390: .d.ts file for --web target does not define the init() function
### Describe the Bug

Compiling a module with --web will create a `.d.ts`, which contains the TypeScript types for all the functions exposed by the module.

The typings for the `init()` function, the default export, are missing.

c410-f3r:
Fixes #1390

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -311,6 +311,7 @@ impl<'a> Context<'a> {
     /// `--target no-modules`, `--target web`, or for bundlers. This is the very
     /// last step performed in `finalize`.
     fn finalize_js(&mut self, module_name: &str, needs_manual_start: bool) -> (String, String) {
+        let mut ts = self.typescript.clone();
         let mut js = String::new();
         if self.config.mode.no_modules() {
             js.push_str("(function() {\n");
@@ -318,7 +319,7 @@ impl<'a> Context<'a> {
 
         // Depending on the output mode, generate necessary glue to actually
         // import the wasm file in one way or another.
-        let mut init = String::new();
+        let mut init = (String::new(), String::new());
         match &self.config.mode {
             // In `--target no-modules` mode we need to both expose a name on
             // the global object as well as generate our own custom start
@@ -371,6 +372,10 @@ impl<'a> Context<'a> {
             }
         }
 
+        let (init_js, init_ts) = init;
+
+        ts.push_str(&init_ts);
+
         // Emit all the JS for importing all our functionality
         js.push_str(&self.imports);
         js.push_str("\n");
@@ -382,7 +387,7 @@ impl<'a> Context<'a> {
         js.push_str("\n");
 
         // Generate the initialization glue, if there was any
-        js.push_str(&init);
+        js.push_str(&init_js);
         js.push_str("\n");
         js.push_str(&self.footer);
         js.push_str("\n");
@@ -394,7 +399,7 @@ impl<'a> Context<'a> {
             js = js.replace("\n\n\n", "\n\n");
         }
 
-        (js, self.typescript.clone())
+        (js, ts)
     }
 
     fn wire_up_initial_intrinsics(&mut self) -> Result<(), Error> {
@@ -842,7 +847,34 @@ impl<'a> Context<'a> {
         Ok(())
     }
 
-    fn gen_init(&mut self, module_name: &str, needs_manual_start: bool) -> String {
+    fn ts_for_init_fn(has_memory: bool) -> String {
+        let (memory_doc, memory_param) = if has_memory {
+            (
+                "* @param {WebAssembly.Memory} maybe_memory\n",
+                ", maybe_memory: WebAssembly.Memory",
+            )
+        } else {
+            ("", "")
+        };
+        format!(
+            "\n\
+            /**\n\
+            * If `module_or_path` is {{RequestInfo}}, makes a request and\n\
+            * for everything else, calls `WebAssembly.instantiate` directly.\n\
+            *\n\
+            * @param {{RequestInfo | BufferSource | WebAssembly.Module}} module_or_path\n\
+            {}\
+            *\n\
+            * @returns {{Promise<any>}}\n\
+            */\n\
+            export function init \
+                (module_or_path: RequestInfo | BufferSource | WebAssembly.Module{}): Promise<any>;
+        ",
+            memory_doc, memory_param
+        )
+    }
+
+    fn gen_init(&mut self, module_name: &str, needs_manual_start: bool) -> (String, String) {
         let mem = self.module.memories.get(self.memory);
         let (init_memory1, init_memory2) = if mem.import.is_some() {
             let mut memory = String::from("new WebAssembly.Memory({");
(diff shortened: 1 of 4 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -865,7 +878,7 @@ impl<'a> Context<'a> {
 
         format!(
             "\
-                function init(module_or_path, maybe_memory) {{
```
alexcrichton: Although only used in some situations, this is used above if `mem.import.is_some()` holds true. I imagine though you got a warning for an unused argument? Perhaps we could only include the argument if `mem.import.is_some()`?

crates/typescript-tests/src/web/init.ts
```diff
@@ -0,0 +1,3 @@
+import * as wbg from '../../pkg/web/typescript_tests';
```
alexcrichton: Nice idea!

crates/cli-support/src/js/mod.rs
```diff
@@ -842,6 +845,16 @@ impl<'a> Context<'a> {
         Ok(())
     }
 
+    fn ts_for_init_fn() -> &'static str {
+        "\n\
+        /**\n\
+        * @param {string} module_or_path\n\
+        * @returns {Promise<any>}\n\
+        */\n\
+        export function init(module_or_path: string): Promise<any>;
```
alexcrichton: This might actually be best documented as `any` instead of `string` perhaps? It may be worth documenting that `string` means "go fetch this URL" (along with types like `URL` and `Request`), but everything else is "just pass this to `WebAssembly.instantiate`

crates/cli-support/src/js/mod.rs
c410-f3r: Oooppss :S This was a totally overlooked decision

crates/cli-support/src/js/mod.rs
RReverser: Better to list actual types that are being expected than use `any`, like:

`RequestInfo | URL | WebAssembly.Module` (or exclude `URL` for consistency with TypeScript's definition of `fetch`)

c410-f3r: I don't think TS has built-in support for WASM types

RReverser: Ugh yeah, looks like for now they live here: https://www.npmjs.com/package/@types/webassembly-js-api

I think it would be ok to add these types as a dependency?

RReverser: (this would also help with the return type of the function)

c410-f3r: Oh NVM,  `wasm-bindgen` already exports WASM types like in https://github.com/rustwasm/wasm-bindgen/blob/c5f18b609922bffce459b572daf70b2ddb190f7b/crates/cli-support/src/wasm2es6js.rs#L55

crates/cli-support/src/js/mod.rs
```diff
@@ -842,7 +847,43 @@ impl<'a> Context<'a> {
...
+        if has_memory {
+            concat!(initial_ts!(), "\
+            * @param {WebAssembly.Memory} maybe_memory\n\
+            *\n\
+            * @returns {Promise<WebAssembly.ResultObject | WebAssembly.Instance>}\n\
+            */\n\
+            export function init (\
+                module_or_path: RequestInfo | BufferSource | WebAssembly.Module,\
+                maybe_memory: WebAssembly.Memory\
+            ): Promise<WebAssembly.ResultObject | WebAssembly.Instance>;
```
RReverser: On which conditions does it return `WebAssembly.Instance` as opposed to `ResultObject`?

crates/cli-support/src/js/mod.rs
```diff
@@ -842,7 +847,43 @@ impl<'a> Context<'a> {
...
+        }
+        if has_memory {
+            concat!(initial_ts!(), "\
+            * @param {WebAssembly.Memory} maybe_memory\n\
+            *\n\
+            * @returns {Promise<WebAssembly.ResultObject | WebAssembly.Instance>}\n\
+            */\n\
+            export function init (\
+                module_or_path: RequestInfo | BufferSource | WebAssembly.Module,\
+                maybe_memory: WebAssembly.Memory\
```
RReverser: This should be marked as optional param (both here and in JSDoc).

crates/cli-support/src/js/mod.rs
c410-f3r: `WebAssembly.instantiate` has two overloading methods ([1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiate), [2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8c9b0319e58804090687ba85e15e6ac909c8452e/types/webassembly-js-api/index.d.ts#L113-L114)). Maybe define those overloads in the definition itself?

crates/cli-support/src/js/mod.rs
c410-f3r: Although `maybe_memory?` will make linters happier, it is going to be odd to see a parameter that is not being used for anything in some situations.

crates/cli-support/src/js/mod.rs
alexcrichton: Oh this doesn't actually resolve to either, but it sort of resolves to this module itself. It's probably fine to just say it resolves to `any` for now and we can refine it necessary if need be.

(it definitely doesn't result in `Instance` or `ResultObject` though)

crates/cli-support/src/js/mod.rs
alexcrichton: Yes this shouldn't be optional because it's a static property whether it's required or not.

crates/cli-support/src/js/mod.rs
```diff
@@ -842,7 +847,43 @@ impl<'a> Context<'a> {
...
+            ");
+        }
+        if has_memory {
+            concat!(initial_ts!(), "\
+            * @param {WebAssembly.Memory} maybe_memory\n\
+            *\n\
+            * @returns {Promise<WebAssembly.ResultObject | WebAssembly.Instance>}\n\
+            */\n\
+            export function init (\
+                module_or_path: RequestInfo | BufferSource | WebAssembly.Module,\
```
alexcrichton: Could this be deduplicated in the case `!has_memory` to avoid the duplication here?

crates/cli-support/src/js/mod.rs
RReverser: Ah this is generated only when memory *has* to be passed from the outside? As in, can I still omit it when memory is defined inside of WASM?

alexcrichton: Correct yeah

alexcrichton:
Thanks @c410-f3r!

(also feel free to ping on branch updates, unfortunately github doesn't always send updates!)

crates/cli-support/src/js/mod.rs
RReverser: I'm going to submit another issue / PR, but just ran into this and appears that it has to be optional after all.

PR #1413, opened 2019-04-01
Attempt to make CI green again

c410-f3r:
This is an attempt to make CI green again. Not sure why the behavior has changed.


alexcrichton:
Thanks for the PR! I was talking with some folks and it looks like this was a regression in azure pipelines tracked at https://github.com/Microsoft/azure-pipelines-agent/issues/2191 (or so I'm told). I was thinking we'd wait until a fix was deployed on their end, although I can poke again soon to see the status of that!

c410-f3r:
@alexcrichton Thanks for the explanation!

Issue #1313, opened 2019-03-05
Investigate possible performance wins with `TextEncoder#encodeInto`


alexcrichton:
There's some discussion [starting here](https://github.com/rustwasm/wasm-bindgen/pull/1279#discussion_r260616542) about how we can probably improve the current logic of using `encodeInto` through some more clever usage and possibly some magic numbers. We should take a look into this! Ideally we'd also take a look at actual performance numbers when doing so.

hsivonen:
I put a slightly polished restatement of my [previous comment](https://github.com/rustwasm/wasm-bindgen/pull/1279#discussion_r260616542) [onto MDN](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encodeInto#Buffer_Sizing).

It would be nice to copy and paste the code resulting from fixing this issue into the "Examples" section of the MDN article.

alexcrichton:
Oh thanks for the link @hsivonen!

RReverser:
Note that another alternative that might be worth considering and measuring is calculating UTF-8 byte length upfront.

I've already added usage of `Buffer.byteLength(...)` in #1391 for Node.js, but for other targets we can do something like:

```js
let size = arg.length;
for (let i = 0; i < arg.length; i++) {
    let code = arg.charCodeAt(i);
    if (code > 0x7f) size++;
    if (code > 0x7ff) size++;
    if (code >= 0xD800 && code <= 0xDBFF) i++; // high surrogate
}
```

After that, we can allocate the right size right away.

hsivonen:
> Note that another alternative that might be worth considering and measuring is calculating UTF-8 byte length upfront.

Gecko used to do this for its internal UTF-16 to UTF-8 conversions, but doing imprecise allocations was better for performance. (With jemalloc, allocations are imprecise anyway. With a precise allocator, it might be more interesting to attempt to do precise allocations.)

RReverser:
> With jemalloc, allocations are imprecise anyway.

Does WASM in Rust use jemalloc? AFAIK Rust switched from it to the system allocator by default, but now that I think about it, I'm not sure which allocator is considered "system" and included in the WASM target.

Although maybe most people use `wee_alloc` instead of the default one? Worth investigating.

alexcrichton:
The wasm target does not use jemalloc, it uses a port of dlmalloc.

RReverser:
[...]

PR #1414, opened 2019-04-01
Optimise encodeInto reallocations

RReverser:
Instead of doubling the size on each iteration, use precise upper limit (3 * JS length) if the string turned out not to be ASCII-only. This results in maximum of 1 reallocation instead of O(log N).

Some dummy examples of what this would change:
 - 1000 of ASCII chars: no change, allocates 1000 bytes and bails out.
 - 1000 ASCII chars + 1 '😃': before allocated 1000 bytes and reallocated to 2000; now allocates 1000 bytes and reallocates to 1006.
 - 1000 of '😃' chars: before allocated 1000 bytes, reallocated to 2000, finally reallocated again to 4000; now allocates 1000 bytes and reallocates to 4000 right away.

Related issue: #1313 

r? @alexcrichton @hsivonen

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1306,13 +1306,12 @@ impl<'a> Context<'a> {
                 while (true) {{
                     const view = getUint8Memory().subarray(ptr + writeOffset, ptr + size);
                     const {{ read, written }} = cachedTextEncoder.encodeInto(arg, view);
-                    arg = arg.substring(read);
-                    writeOffset += written;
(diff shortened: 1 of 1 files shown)


RReverser:
This seems to pass all tests (the failures, as usual, are related to deployment scripts and not code changes).

@alexcrichton can we merge this as-is for now? This might be not the final version of optimisations, but it already significantly reduces costs of `encodeInto` in mixed ASCII & Unicode scenarios.

alexcrichton:
Looks good to me!

PR #1416, opened 2019-04-01
Add warnings about UTF-16 vs UTF-8 strings

### LINKED ISSUE #1348: Incorrect handling of unpaired surrogates in JS strings
### Describe the Bug

It was brought to my attention in https://github.com/Pauan/rust-dominator/issues/10 that JavaScript strings (and DOMString) allow for unpaired surrogates.

When using `TextEncoder`, it will convert those unpaired surrogates into U+FFFD (the replacement character). According to the Unicode spec, this is correct behavior.

The issue is that because the unpaired surrogates are replaced, this is *lossy*, and that lossiness can cause serious issues.

You can read the above dominator bug report for the nitty gritty details, but the summary is that with `<input>` fields (and probably other things), it will send *two* `input` events, one for each surrogate.

When the first event arrives, the surrogate is unpaired, so because the string is immediately sent to Rust, the unpaired surrogate is converted into the replacement character.

Then the second event arrives, and the surrogate is *still* unpaired (because the first half was replaced), so the second half also gets replaced with the replacement character.

This has a lot of very deep implications, including for international languages (e.g. Chinese).

I did quite a bit of reading, and unfortunately I think the only real solution here is to always use `JsString`, and not convert into Rust `String`, because that is inherently lossy. Or if a conversion is done, it needs to do some checks to make sure that there aren't any unpaired surrogates.

alexcrichton:
This commit aims to address #1348 via a number of strategies:

* Documentation is updated to warn about UTF-16 vs UTF-8 problems
  between JS and Rust. Notably documenting that `as_string` and handling
  of arguments is lossy when there are lone surrogates.

* A `JsString::is_valid_utf16` method was added to test whether
  `as_string` is lossless or not.

The intention is that most default behavior of `wasm-bindgen` will
remain, but where necessary bindings will use `JsString` instead of
`str`/`String` and will manually check for `is_valid_utf16` as
necessary. It's also hypothesized that this is relatively rare and not
too performance critical, so an optimized intrinsic for `is_valid_utf16`
is not yet provided.

Closes #1348

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3522,6 +3522,37 @@ impl JsString {
             None
         }
     }
+
+    /// Returns whether this string is a valid UTF-16 string.
+    ///
+    /// This is useful for learning whether `String::from(..)` will return a
+    /// lossless representation of the JS string. If this string contains
+    /// unpaired surrogates then `String::from` will succeed but it will be a
+    /// lossy representation of the JS string because unpaired surrogates will
+    /// become replacement characters.
+    ///
+    /// If this function returns `false` then to get a lossless representation
+    /// of the string you'll need to manually use the `iter` method (or the
+    /// `char_code_at` accessor) to view the raw character codes.
+    ///
+    /// For more information, see the documentation on [JS strings vs Rust
+    /// strings][docs]
+    ///
+    /// [docs]: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/str.html
+    pub fn is_valid_utf16(&self) -> bool {
+        std::char::decode_utf16(self.iter()).all(|i| i.is_ok())
+    }
+
+    /// Returns an iterator over the `u16` character codes that make up this JS
+    /// string.
+    ///
+    /// This method will call `char_code_at` for each code in this JS string,
+    /// returning an iterator of the codes in sequence.
+    pub fn iter<'a>(
+        &'a self,
+    ) -> impl ExactSizeIterator<Item = u16> + DoubleEndedIterator<Item = u16> + 'a {
+        (0..self.length()).map(move |i| self.char_code_at(i) as u16)
+    }
 }
 
 impl PartialEq<str> for JsString {
--- a/guide/src/reference/types/str.md
+++ b/guide/src/reference/types/str.md
@@ -20,3 +20,30 @@ with handles to JavaScript string values, use the `js_sys::JsString` type.
 ```js
 {{#include ../../../../examples/guide-supported-types-examples/str.js}}
 ```
+
+## UTF-16 vs UTF-8
+
+Strings in JavaScript are encoded as UTF-16, but with one major exception: they
+can contain unpaired surrogates. For some Unicode characters UTF-16 uses two
+16-byte values.  These are called "surrogate pairs" because they always come in
+pairs. In JavaScript, it is possible for these surrogate pairs to be missing the
+other half, creating an "unpaired surrogate".
+
+When passing a string from JavaScript to Rust, it uses the `TextEncoder` API to
+convert from UTF-16 to UTF-8. This is normally perfectly fine... unless there
+are unpaired surrogates. In that case it will replace the unpaired surrogates
+with U+FFFD (�, the replacement character). That means the string in Rust is
+now different from the string in JavaScript!
+
+If you want to guarantee that the Rust string is the same as the JavaScript
+string, you should instead use `js_sys::JsString` (which keeps the string in
+JavaScript and doesn't copy it into Rust).
+
+If you want to access the raw value of a JS string, you can use `JsString::iter`,
+which returns an `Iterator<Item = u16>`. This perfectly preserves everything
+(including unpaired surrogates), but it does not do any encoding (so you
+have to do that yourself!).
+
+If you simply want to ignore strings which contain unpaired surrogates, you can
+use `JsString::is_valid_utf16` to test whether the string contains unpaired
+surrogates or not.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -260,6 +260,16 @@ impl JsValue {
     ///
     /// If this JS value is not an instance of a string or if it's not valid
     /// utf-8 then this returns `None`.
+    ///
+    /// # UTF-16 vs UTF-8
+    ///
+    /// JavaScript strings in general are encoded as UTF-16, but Rust strings
+    /// are encoded as UTF-8. This can cause the Rust string to look a bit
+    /// different than the JS string sometimes. For more details see the
+    /// [documentation about the `str` type][caveats] which contains a few
+    /// caveats about the encodings.
+    ///
+    /// [caveats]: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/str.html
     #[cfg(feature = "std")]
     pub fn as_string(&self) -> Option<String> {
         unsafe {
--- a/crates/js-sys/tests/wasm/JsString.rs
+++ b/crates/js-sys/tests/wasm/JsString.rs
@@ -541,3 +541,15 @@ fn raw() {
     );
     assert!(JsString::raw_0(&JsValue::null().unchecked_into()).is_err());
 }
+
+#[wasm_bindgen_test]
+fn is_valid_utf16() {
+    assert!(JsString::from("a").is_valid_utf16());
+    assert!(JsString::from("").is_valid_utf16());
+    assert!(JsString::from("🥑").is_valid_utf16());
+    assert!(JsString::from("Why hello there this, 🥑, is 🥑 and is 🥑").is_valid_utf16());
+
+    assert!(JsString::from_char_code1(0x00).is_valid_utf16());
+    assert!(!JsString::from_char_code1(0xd800).is_valid_utf16());
+    assert!(!JsString::from_char_code1(0xdc00).is_valid_utf16());
+}
--- a/examples/without-a-bundler/index.html
+++ b/examples/without-a-bundler/index.html
@@ -25,7 +25,12 @@
         // Also note that the promise, when resolved, yields the wasm module's
         // exports which is the same as importing the `*_bg` module in other
         // modes
-        await init('./pkg/without_a_bundler_bg.wasm');
+        // await init('./pkg/without_a_bundler_bg.wasm');
+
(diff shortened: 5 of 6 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,22 @@ impl JsString {
             None
         }
     }
+
+    /// Returns whether this string is a valid UTF-16 string.
```
Pauan: These docs are pretty weird: they mention `as_string`, but `as_string` doesn't exist on `JsString`. Instead, the relevant function is `Into::into`.

I also think it should use "unpaired surrogate" rather than "lone surrogate", since that seems to be the normal term for it (and it also gives better Google results).

crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,22 @@ impl JsString {
...
+    /// This is useful for learning whether `as_string()` will return a lossless
+    /// representation of the JS string. If this string contains lone surrogates
+    /// then `as_string()` will succeed but it will be a lossy representation of
+    /// the JS string because lone surrogates will become replacement
+    /// characters.
+    ///
+    /// If this function returns `false` then to get a lossless representation
+    /// of the string you'll need to manually use `char_code_at` accessor to
+    /// access the raw code points.
+    pub fn is_valid_utf16(&self) -> bool {
```
Pauan: Since we already have the code for it, maybe we should extract this into a method? Something like this...

```rust
pub fn iter(&self) -> impl Iterator<Item = u16>
```

Now it's quite easy to convert to `Vec<u16>` if the user wishes to do so.

crates/js-sys/src/lib.rs
alexcrichton: I think that `JsString` derefs to `Object` which derefs to `JsValue` so the `as_string` method should be inherently available? As for changing terms 👍

Pauan: Sure, that may be true, but [it doesn't show up in the docs](https://docs.rs/js-sys/0.3.17/js_sys/struct.JsString.html), and in addition it's not idiomatic: users should use `foo.into()` to convert from `JsString` to `String`, not `foo.as_string().unwrap_throw()`

crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,35 @@ impl JsString {
...
+        std::char::decode_utf16(self.iter()).all(|i| i.is_ok())
+    }
+
+    /// Returns an iterator over the u16 character codes that make up this JS
+    /// string.
+    ///
+    /// This method will call `char_code_at` for each code in this JS string,
+    /// returning an iterator of the codes in sequence.
+    pub fn iter<'a>(&'a self) -> impl ExactSizeIterator<Item = u16> + 'a {
```
Pauan: Can this be `impl ExactSizeIterator<Item = u16> + DoubleEndedIterator<Item = u16> + 'a`?

guide/src/reference/types/str.md
```diff
@@ -20,3 +20,25 @@ with handles to JavaScript string values, use the `js_sys::JsString` type.
 ```js
 {{#include ../../../../examples/guide-supported-types-examples/str.js}}
 ```
+
+## UTF-16 vs UTF-8
+
+Strings in JavaScript are encoded as UTF-16, but with one major exception: they
+can contain unpaired surrogates. For some unicode characters UTF-16 usese two
```
Pauan: Nit: unicode -> Unicode
Nit: usese -> uses

src/lib.rs
```diff
@@ -260,6 +260,28 @@ impl JsValue {
     ///
     /// If this JS value is not an instance of a string or if it's not valid
     /// utf-8 then this returns `None`.
+    ///
+    /// # UTF-16 vs UTF-8
+    ///
+    /// Strings in JavaScript are encoded as UTF-16, but with one major
+    /// exception: they can contain unpaired surrogates. For some unicode
```
Pauan: Same nits as above.

src/lib.rs
```diff
@@ -260,6 +260,28 @@ impl JsValue {
...
+    /// "unpaired surrogate".
+    ///
+    /// This method internally will use `TextEncoder` in JS to convert from
+    /// UTF-16 to UTF-8. This is normally perfectly fine...  unless there are
+    /// unpaired surrogates. In that case it will replace the unpaired
+    /// surrogates with U+FFFD (�, the replacement character). That means the
+    /// string returned from this function is actually different from the string
+    /// in JS!
+    ///
+    /// If you want to guarantee that the returned string is the same as the
```
Pauan: This reads really weird. `is_valid_utf16` isn't used for ensuring the string is the same in Rust, it's used for ignoring the string entirely.

Can we just link this somehow to `reference/types/string.md`? I think that's better than duplicating all that information (and possibly having it get out of sync, etc.)

guide/src/reference/types/str.md
```diff
@@ -20,3 +20,25 @@ with handles to JavaScript string values, use the `js_sys::JsString` type.
...
+convert from UTF-16 to UTF-8. This is normally perfectly fine... unless there
+are unpaired surrogates. In that case it will replace the unpaired surrogates
+with U+FFFD (�, the replacement character). That means the string in Rust is
+now different from the string in JavaScript!
+
+If you want to guarantee that the Rust string is the same as the JavaScript
+string, you should instead use `js_sys::JsString` (which keeps the string in
+JavaScript and doesn't copy it into Rust).
+
+If you simply want to ignore strings which contain unpaired surrogates, you can
```
Pauan: We should add a paragraph here about `JsString::iter`:

> If you want to use the raw bytes of a JS string, you can use `JsString::iter`, which returns an `Iterator<Item = u16>`. This perfectly preserves everything (including unpaired surrogates), but it does not do any encoding (so you have to do that yourself!)

crates/js-sys/src/lib.rs
alexcrichton: Oh I didn't even realize that we could do that, neat!

src/lib.rs
alexcrichton: Agreed yeah, the hyperlink I think is the best way to go

crates/js-sys/src/lib.rs
alexcrichton: Ah good point, I'll update!

crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,37 @@ impl JsString {
             None
         }
     }
+
+    /// Returns whether this string is a valid UTF-16 string.
+    ///
+    /// This is useful for learning whether `String::from(..)` will return a
+    /// lossless representation of the JS string. If this string contains
+    /// unparied surrogates then `String::from` will succeed but it will be a
```
Pauan: Nit: unparied -> unpaired

crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,37 @@ impl JsString {
...
+    /// lossy representation of the JS string because unpaired surrogates will
+    /// become replacement characters.
+    ///
+    /// If this function returns `false` then to get a lossless representation
+    /// of the string you'll need to manually use `char_code_at` accessor to
```
Pauan: Can we mention the `iter` method here? That seems like it would be more idiomatic than manual calls to `char_code_at`.

crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,37 @@ impl JsString {
...
+    ///
+    /// For more information, see the documentation on [JS strings vs Rust
+    /// strings][docs]
+    ///
+    /// [docs]: https://rustwasm.github.io/docs/wasm-bindgen/reference/types/str.html
+    pub fn is_valid_utf16(&self) -> bool {
+        std::char::decode_utf16(self.iter()).all(|i| i.is_ok())
+    }
+
+    /// Returns an iterator over the u16 character codes that make up this JS
```
Pauan: Nit: u16 -> `u16`

guide/src/reference/types/str.md
```diff
@@ -20,3 +20,30 @@ with handles to JavaScript string values, use the `js_sys::JsString` type.
...
+convert from UTF-16 to UTF-8. This is normally perfectly fine... unless there
+are unpaired surrogates. In that case it will replace the unpaired surrogates
+with U+FFFD (�, the replacement character). That means the string in Rust is
+now different from the string in JavaScript!
+
+If you want to guarantee that the Rust string is the same as the JavaScript
+string, you should instead use `js_sys::JsString` (which keeps the string in
+JavaScript and doesn't copy it into Rust).
+
+If you want access the raw value of a JS string, you can use `JsString::iter`,
```
Pauan: Nit: If you want access -> If you want to access

src/lib.rs
```diff
@@ -260,6 +260,16 @@ impl JsValue {
...
     /// If this JS value is not an instance of a string or if it's not valid
     /// utf-8 then this returns `None`.
+    ///
+    /// # UTF-16 vs UTF-8
+    ///
+    /// JavaScript strings in general are encoded as UTF-16, but Rust strings
+    /// are encoded as UTF-8. This can cause the Rust string to look a bit
+    /// different than the JS string sometimes. For more details see the
+    /// [documentation about the `str` type][caveats] which contains a few
+    /// caveats about the encodings here.
```
Pauan: Nit: encodings here -> encodings.

alexcrichton:
👍 Thanks for the thorough review!

crates/js-sys/src/lib.rs
```diff
@@ -3522,6 +3522,37 @@ impl JsString {
...
+    /// Returns whether this string is a valid UTF-16 string.
+    ///
+    /// This is useful for learning whether `String::from(..)` will return a
+    /// lossless representation of the JS string. If this string contains
+    /// unpaired surrogates then `String::from` will succeed but it will be a
+    /// lossy representation of the JS string because unpaired surrogates will
+    /// become replacement characters.
+    ///
+    /// If this function returns `false` then to get a lossless representation
+    /// of the string you'll need to manually use `iter` method (or
```
Pauan: Nit: manually use `iter` -> manually use the `iter`
Nit: or `char_code_at` -> or the `char_code_at`
Nit: the raw code points -> the raw character codes

(["code point" refers to a complete character](https://en.wikipedia.org/wiki/Character_encoding#Terminology), not a specific encoding, so it cannot refer to a byte sequence)

alexcrichton:
Thanks @Pauan!

Issue #1399, opened 2019-03-26
Closures should accept & for arguments


Pauan:
### Motivation

If a Closure accepts `&` by argument it can use the faster stack-allocation rather than the slower heap-allocation.

### Proposed Solution

This should be possible:

```rust
let f = move |e: &Foo| { ... };
let x = Closure::wrap(Box::new(f) as Box<FnMut(&Foo)>);
```

### Additional Context

https://github.com/rustwasm/gloo/issues/30#issuecomment-475866803

alexcrichton:
Out of curiosity, since I think that a lot of code duplication would be required here is there perhaps one or two signatures which would hit 90% of use cases? For example would `FnMut(&T)` be enough to cover basically everything?

Pauan:
@alexcrichton It would certainly cover our particular use case (of event listeners). Though we would need `FnOnce(&T)` as well. So that's 2 signatures.

alexcrichton:
Ok I've posted an initial idea of this in https://github.com/rustwasm/wasm-bindgen/pull/1417

PR #1417, opened 2019-04-01
Support 1-reference argument closures

alexcrichton:
This is work towards #1399, although it's just for one-argument closures
where the first argument is a reference. No other closures with
references in argument position are supported yet


fitzgen:
This is good enough for `gloo_events`, so yeah let's see how far we can go with this...

PR #1424, opened 2019-04-04
Don't put anything before ES module imports

fitzgen:
Because of some incorrect use of `js.push_str(..)`, we could sometimes emit code
before the ES modules imports, which is syntactically invalid:

    const __exports = {};
    import { Thing } from '...'; // Syntax error!

This has been fixed by making sure that the correct `imports` or `imports_post`
string is built up. We now also assert that the `js` string is empty at the
location where we add imports if we're using ES modules.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -354,7 +354,8 @@ impl<'a> Context<'a> {
             | OutputMode::Node {
                 experimental_modules: true,
             } => {
-                js.push_str(&format!("import * as wasm from './{}_bg';\n", module_name));
+                self.imports
+                    .push_str(&format!("import * as wasm from './{}_bg';\n", module_name));
                 if needs_manual_start {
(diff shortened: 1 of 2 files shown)


Pauan:
> we could sometimes emit code before the ES modules imports, which is syntactically invalid

Why do you say that? It's perfectly valid to have code before an `import`.

fitzgen:
> Why do you say that? It's perfectly valid to have code before an `import`.

Hm -- I was getting syntax errors claiming otherwise? Will try and repro again.

fitzgen:
Oh yeah I totally misread the error message and it was acutally because I didn't have `type="module"` on the script tag!

I'm ambivalent on keeping this PR or not then. It seems tidy to have imports before everything else, but it isn't really necessary.

Pauan:
> It seems tidy to have imports before everything else, but it isn't really necessary.

Yeah, I'm totally fine with having them at the top for stylistic/organizational reasons, I just didn't want the spread of incorrect information.

PR #1431, opened 2019-04-08
Add an accessor for the function table

### LINKED ISSUE #1427: Add `wasm_bindgen::function_table()`
So I finally started implementing `wasm-bindgen` compatibility in `stdweb`, [and it works](https://github.com/koute/stdweb/tree/master/examples/wasm-bindgen-minimal)! It is however not yet entirely functional - for that I need access to the function table of the WASM's module. In [the RFC](https://github.com/rustwasm/rfcs/blob/master/text/006-local-js-dependencies.md) it was said that an extra `wasm_bindgen::function_table()` intrinsic will be added to access the function table; from what I can see that hasn't been done yet?

Any way I can help?

alexcrichton:
This commit adds an intrinsics to the `wasm_bindgen` crate which
accesses the `WebAssembly.Table` which is the function table of the
module. Eventually the thinking is that a module would import its own
function table via native wasm functionality (via `anyref` and such),
but until that's implemented let's add a binding for it ourselves!

Closes #1427

### DIFF
--- a/tests/wasm/api.js
+++ b/tests/wasm/api.js
@@ -55,3 +55,9 @@ exports.debug_values = () => ([
     () => (null),
     new Set(),
 ]);
+
+exports.assert_function_table = (x, i) => {
+  const rawWasm = require('wasm-bindgen-test_bg.js');
+  assert.ok(x instanceof WebAssembly.Table);
+  assert.strictEqual(x.get(i), rawWasm.function_table_lookup);
+};
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -783,6 +783,14 @@ impl<'a> Context<'a> {
             ))
         })?;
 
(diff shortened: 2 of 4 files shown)


tests/wasm/api.js
```diff
@@ -55,3 +55,7 @@ exports.debug_values = () => ([
     () => (null),
     new Set(),
 ]);
+
+exports.assert_function_table = x => {
```
fitzgen: Maybe it makes sense to assert that this is the right function table too?

Imagining something like this:

```rust
assert_function_table(wasm_bindgen::function_table(), my_exported_function as *const _ as usize);
```

```js
exports.assert_function_table = (t, i) => {
    assert.ok(t instanceof WebAssembly.Table);
    assert.equal(t.get(i), wasm.my_exported_function);
};
```

alexcrichton: Sounds like a great idea to me!

PR #1434, opened 2019-04-10
Bump to 0.2.41

alexcrichton:


alexcrichton:
Hm ok the web-sys test failure is worrisome but shouldn't block the release, so merging!

alexcrichton:
er other way around

PR #1440, opened 2019-04-11
Getters/Setters for fields

### LINKED ISSUE #222: Support defining getters in Rust
Currently when you export a struct in Rust to JS it only exports public methods, but we should enable the ability to define getters and setters for properties too, for example:

```rust
#[wasm_bindgen]
struct Foo { /* ... */ }

#[wasm_bindgen]
impl Foo {
    pub fn new() -> Foo { ... }

    #[wasm_bindgen(getter)]
    pub fn field1(&self) -> u32 { ... }
    #[wasm_bindgen(setter)]
    pub fn set_field1(&mut self, f: u32) { ... }

    #[wasm_bindgen(getter = anotherName)]
    pub fn another_name(&self) { ... }
    #[wasm_bindgen(setter = anotherName)]
    pub fn set_another_name(&mut self, param: u32) { ... }
}
```

c410-f3r:
Fixes #222.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1,18 +1,24 @@
-use crate::decode;
-use crate::descriptor::{Descriptor, VectorKind};
-use crate::{Bindgen, EncodeInto, OutputMode};
-use failure::{bail, Error, ResultExt};
-use std::collections::{BTreeMap, HashMap, HashSet};
-use std::env;
-use std::fs;
-use walrus::{MemoryId, Module};
-use wasm_bindgen_wasm_interpreter::Interpreter;
-
-mod js2rust;
-use self::js2rust::{ExportedShim, Js2Rust};
-mod rust2js;
-use self::rust2js::Rust2Js;
 mod closures;
+mod js2rust;
+mod rust2js;
+
+use self::{
+    js2rust::{ExportedShim, Js2Rust},
+    rust2js::Rust2Js,
+};
+use crate::{
+    decode,
+    descriptor::{Descriptor, VectorKind},
+    Bindgen, EncodeInto, OutputMode,
+};
+use failure::{bail, Error, ResultExt};
+use std::{
+    collections::{BTreeMap, HashMap, HashSet},
+    env, fs,
+};
+use walrus::{MemoryId, Module};
+use wasm_bindgen_shared::struct_function_export_name;
+use wasm_bindgen_wasm_interpreter::Interpreter;
 
 pub struct Context<'a> {
     pub globals: String,
@@ -2189,13 +2195,8 @@ impl<'a> Context<'a> {
         }
     }
 
-    fn require_class_wrap(&mut self, class: &str) {
-        self.exported_classes
-            .as_mut()
-            .expect("classes already written")
-            .entry(class.to_string())
-            .or_insert_with(ExportedClass::default)
-            .wrap_needed = true;
+    fn require_class_wrap(&mut self, name: &str) {
+        require_class(&mut self.exported_classes, name).wrap_needed = true;
     }
 
     fn import_identifier(&mut self, import: Import<'a>) -> String {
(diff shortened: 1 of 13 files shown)


RReverser:
Is it possible to eliminate the extra method on the JS side and keep only the getter?

c410-f3r:
@RReverser Sure. I was a bit apprehensive about this aspect but I come up with the following rules.

<table>
  <thead>
    <tr>
      <th colspan="4">Input</th>
      <th colspan="2">Output</th>
    </tr>
    <tr>
      <th>pub field</th>
      <th>js_name attr</th>
      <th>getter/setter with name</th>
      <th>getter/setter w/o name</th>
      <th>get/set method name</th>
      <th>ordinary method name</th>
  </thead>
  <tbody>
    <tr>
      <td>-</td>
      <td>-</td>
      <td>-</td>
      <td>-</td>
      <td>-</td>
      <td>method name</td>
    </tr>
    <tr>
      <td>-</td>
      <td>-</td>
      <td>-</td>
      <td>X</td>
      <td>method name</td>
      <td>-</td>
    </tr>
    <tr>
      <td>-</td>
      <td>-</td>
      <td>X</td>
      <td>-</td>
      <td>getter/setter name</td>
      <td>-</td>
    </tr>
    <tr>
      <td>-</td>
      <td>X</td>
      <td>-</td>
      <td>-</td>
      <td>-</td>
      <td>js_name</td>
    </tr>
    <tr>
      <td>-</td>
      <td>X</td>
      <td>-</td>
      <td>X</td>
      <td>method name</td>
      <td>js_name</td>
    </tr>
    <tr>
      <td>-</td>
      <td>X</td>
      <td>X</td>
      <td>-</td>
      <td>getter/setter name</td>
      <td>js_name</td>
    </tr>
    <tr>
      <td>X</td>
      <td>-</td>
      <td>-</td>
      <td>-</td>
      <td>field name</td>
      <td>method name</td>
    </tr>
    <tr>
      <td>X</td>
      <td>-</td>
      <td>-</td>
      <td>X</td>
      <td>field name and method name</td>
      <td>-</td>
    </tr>
    <tr>
      <td>X</td>
      <td>-</td>
      <td>X</td>
      <td>-</td>
      <td>field name and getter/setter name</td>
      <td>-</td>
    </tr>
    <tr>
      <td>X</td>
      <td>X</td>
      <td>-</td>
      <td>-</td>
      <td>field name</td>
      <td>js_name</td>
    </tr>
    <tr>
      <td>X</td>
      <td>X</td>
      <td>-</td>
      <td>X</td>
      <td>field name and method name</td>
      <td>js_name</td>
    </tr>
    <tr>
      <td>X</td>
      <td>X</td>
      <td>X</td>
      <td>-</td>
      <td>field name and getter/setter name</td>
      <td>js_name</td>
    </tr>
  </tbody>
</table>

I will try to inform the user about duplicated getters/setters names in the macro definition based on the given table.

RReverser:
Ah, so this extra method is generated only due to `js_name` in the example above and would be omitted otherwise? That makes sense.

c410-f3r:
> Ah, so this extra method is generated only due to `js_name` in the example above and would be omitted otherwise? That makes sense.

Yeah

crates/cli-support/src/js/mod.rs
```diff
@@ -2634,63 +2635,115 @@ impl<'a, 'b> SubContext<'a, 'b> {
...
             None => return Ok(()),
             Some(d) => d,
         };
-
-        let function_name = if export.is_constructor {
-            "constructor"
-        } else {
-            &export.function.name
+        let docs = |raw_docs| format_doc_comments(&export.comments, Some(raw_docs));
+        let getter_from_method_call = |c: &mut ExportedClass, docs, fn_name, g_name| {
```
alexcrichton: Since this and `setter_from_method_call` are only called once, could their contents be inlined below?

crates/cli-support/src/js/mod.rs
```diff
@@ -2634,63 +2635,115 @@ impl<'a, 'b> SubContext<'a, 'b> {
...
+                    static_prfx = "static ";
+                } else {
+                    j2r.method(export.consumed);
+                }
+                let (js, ts, raw_docs) = finish_j2r(j2r)?;
+                let class = require_class(&mut self.cx.exported_classes, class_name);
+                let docs = docs(raw_docs);
+                match operation.kind {
+                    decode::OperationKind::Getter(maybe_getter_name) => {
+                        if export.function.renamed_via_js_name {
```
alexcrichton: Could a comment be added here explaining what's going on?

alexcrichton: I'm also not sure if this is the best way to approach `js_name`? I would think that we basically just ignore `js_name` on setters/getters because customizing the field name would be `#[wasm_bindgen(getter = otherName)]` or something like that. If a new method wants to be created I think we'd want to do that through a dedicated separate method in the source

crates/cli-support/src/js/mod.rs
```diff
@@ -2634,63 +2635,115 @@ impl<'a, 'b> SubContext<'a, 'b> {
...
+                            };
+                            method(class, &docs, getter_name, "get ", &js, &ts);
+                        }
+                    }
+                    decode::OperationKind::Setter(maybe_setter_name) => {
+                        if export.function.renamed_via_js_name {
+                            method(class, &docs, fn_name, "", &js, &ts);
+                            setter_from_method_call(class, &docs, fn_name, &js, maybe_setter_name);
+                        } else {
+                            let getter_name = if maybe_setter_name.is_empty() {
```
alexcrichton: Should `maybe_setter_name` be an `Option` instead of a `&str` if we test if it's empty here?

alexcrichton: Or actually better yet, never have it as an empty string but the backend codegen does the `unwrap_or` for us ahead of time

tests/wasm/getters_and_setters.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn set_wo_js_wo_swn_with_swon(&mut self, field: i32) {
+        self.field = field;
+    }
+
+    #[wasm_bindgen(getter = new_wo_js_with_gwn_wo_gwon_for_getter)]
+    pub fn wo_js_with_gwn_wo_gwon(&self) -> i32 {
+        self.field
+    }
+    #[wasm_bindgen(setter = new_wo_js_with_swn_wo_swon_for_setter)]
+    pub fn set_wo_js_with_swn_wo_swon(&mut self, field: i32) {
```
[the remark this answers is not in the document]
c410-f3r: Confusing, indeed. These are abbreviations of the above table:

* wo = w/o = without
* js = js_name
* (g|s)wn = (getter|setter) with name
* (g|s)won = (getter|setter) without name

`wo_js_with_swn_wo_swon` would be: without (js_name), with (setter with name) and without (setter without name).

crates/cli-support/src/js/mod.rs
c410-f3r: > If a new method wants to be created I think we'd want to do that through a dedicated separate method in the source

This is going to be wonderful and life is also going to be very much easier but I think that we shouldn't ignore the `js_name` attribute in getters/setters, that would lead to unexpected behavior, confusing users. Instead, we could forbid both declarations in the same function or simply remove the ability to define names in getters/setters (the name would de defined in `js_name`).

tests/wasm/getters_and_setters.rs
```diff
@@ -0,0 +1,115 @@
...
+fn _6_rust() {
+    let rules = _6_js(Rules { field: 11 });
+    assert_eq!(*&rules.field, 12);
+}
+
+#[wasm_bindgen_test]
+fn _7_rust() {
+    let rules = _7_js(Rules { field: 13 });
+    assert_eq!(*&rules.field, 14);
+}
```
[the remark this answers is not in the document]
c410-f3r: Sure

guide/src/reference/attributes/on-js-imports/getter-and-setter.md
```diff
@@ -59,6 +59,67 @@ extern "C" {
...
+    field: Option<i32>,
+}
+
+#[wasm_bindgen]
+impl Baz {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    #[wasm_bindgen(getter = field, js_name = getField, method)]
```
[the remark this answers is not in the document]
c410-f3r: > The method I think is safely implied with &self

Indeed

> Additionally this is in the "on js imports" section, so could a new page be added to the "on rust exports" section?

Of course

crates/cli-support/src/js/mod.rs
alexcrichton: I'd also be fine considering `getter = foo` equivalent to `getter, js_name = foo` and maybe producing an error for something like `getter = foo, js_name = bar`, what do you think?

FWIW in the long run I'd like to turn "unused attribute warnings" into hard errors by default, but while we have a crate feature for that it'd be a breaking change so I don't think we can do it just yet.

tests/wasm/getters_and_setters.rs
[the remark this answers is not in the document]
alexcrichton: Ah ok! Mind throwing this documentation inline?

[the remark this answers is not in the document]
RReverser: Maybe better to just use full long names? It's just tests anyway, doesn't hurt.

crates/cli-support/src/js/mod.rs
c410-f3r: > it'd be a breaking change so I don't think we can do it just yet.

So unfortunate :(
Ok, the best approach right now would be to ignore the `js_name` attribute.
**Table updated**.

crates/shared/src/lib.rs
```diff
@@ -110,6 +109,7 @@ macro_rules! shared_api {
         struct Function<'a> {
             arg_names: Vec<String>,
             name: &'a str,
+            renamed_via_js_name: bool,
```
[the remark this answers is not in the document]
c410-f3r: I recently rewrote code and this field was removed, but its original use is due to this single line:

https://github.com/rustwasm/wasm-bindgen/blob/402e72ed39f6502830c7fe9466f1a02baa854986/crates/backend/src/ast.rs#L359

crates/cli-support/src/js/mod.rs
c410-f3r: Hummm... There must be a way to known if the user provided a name for the getter/setter attribute and `Option` would be more intuitive

tests/wasm/getters_and_setters.rs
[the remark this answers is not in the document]
c410-f3r: Updated with full long names

crates/cli-support/src/js/mod.rs
```diff
@@ -2634,63 +2635,115 @@ impl<'a, 'b> SubContext<'a, 'b> {
...
+                                fn_name
+                            } else {
+                                maybe_setter_name
+                            };
+                            method(class, &docs, getter_name, "set ", &js, &ts);
+                        }
+                    }
+                    _ => method(class, &docs, fn_name, static_prfx, &js, &ts),
```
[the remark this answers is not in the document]
c410-f3r: Should an error be thrown? This line code roughly translates to: Don't do fancy things for anything that isn't a getter or a setter.

crates/cli-support/src/js/mod.rs
c410-f3r: This method no longer exists

crates/cli-support/src/js/mod.rs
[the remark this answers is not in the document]
alexcrichton: Ah yeah I think we should generate an error here because something like `#[wasm_bindgen(indexing_getter)]` (or something like that) isn't fully supported here

alexcrichton:
Thanks again for this @c410-f3r!

PR #1442, opened 2019-04-11
Fix a bug using `encodeInto` truncating strings

### LINKED ISSUE #1436: KeyboardEvent::key() and code() don't work in wasm-bindgen 0.2.41 in Firefox
### Describe the Bug

In wasm-bindgen 0.2.41, `KeyboardEvent::key()` and `::code()` always return the empty string. In 0.2.40 they worked fine. I'm not sure whether this bug is in wasm-bindgen, web-sys, or another related crate.

### Steps to Reproduce

1. Compile and run https://github.com/nstoddard/webgl-gui-demo
2. Press one of the arrow keys

### Expected Behavior

The arrow keys should move the diamond in the middle.

### Actual Behavior

The arrow keys do nothing. If logging of `pressed_keys` is added to `Demo::handle_pressed_keys`, it shows that the keys are always empty strings (they should be "ArrowLeft" or similar).

### Additional Context

I can only repro this in Firefox, not Chrome. I'm using Linux; I don't know if it happens on other platforms.

If you downgrade the wasm-bindgen dependency to 0.2.40 (which also requires downgrading some dependencies in [webgl-wrapper](https://github.com/nstoddard/webgl-wrapper) and [webgl-gui](https://github.com/nstoddard/webgl-gui)) the arrow keys work.

alexcrichton:
The last write accidentally wasn't accounted for in the returned length
of the string and we unfortunately don't have any test coverage of
`encodeInto` since it requires Firefox nightly right now (and doesn't
work in Node yet).

Closes #1436

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1364,11 +1364,11 @@ impl<'a> Context<'a> {
                 while (true) {{
                     const view = getUint8Memory().subarray(ptr + writeOffset, ptr + size);
                     const {{ read, written }} = cachedTextEncoder.encodeInto(arg, view);
+                    writeOffset += written;
                     if (read === arg.length) {{
                         break;
                     }}
                     arg = arg.substring(read);
-                    writeOffset += written;
                     ptr = wasm.__wbindgen_realloc(ptr, size, size += arg.length * 3);
                 }}
                 WASM_VECTOR_LEN = writeOffset;


RReverser:
(facepalm) sorry! I thought it was covered...

alexcrichton:
I'm realizing now we do actually have coverage of this nowadays, I think we just historically didn't! Yesterday I was confused that [tests failed on the release PR](https://github.com/rustwasm/wasm-bindgen/pull/1434) and was going to investigate more today, but I realized today that `encodeInto` is in stable Firefox and the test failures look like this bug, so I think the recent release of Firefox 66 broke our CI and gave us coverage :)

PR #1447, opened 2019-04-11
Improve Boolean/Number/JsString consistency

### LINKED ISSUE #1446: Soft-deprecate `Number::new` and `Boolean::new`
### Motivation

Boxed primitives are extremely rare and unidiomatic in real-world JavaScript, and cause more problems to support than worth it.

### Proposed Solution

As discussed in #1405, we would want to deprecate `fn new` on `Number` and `Boolean` to discourage use of boxed primitives, probably in favour of adding `From<f64>` and `From<bool>` implementations correspondingly (which will create real primitives).

### Alternatives

Keep the methods, but not support them actively in other parts of the stack. This might be confusing though.

### Additional Context

For what it's worth, we don't provide `fn new` on e.g. `JsString` so this aligns primitives nicely to a common format.

alexcrichton:
* Ensure `PartialEq` is implemented from these types to native Rust types
* Implement `From` between these type and native Rust types
* Deprecated `Number::new` and `Boolean::new` to discourage use of the
  object forms, recommending the `from` constructors instead.

Closes #1446

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -467,13 +467,14 @@ extern "C" {
 #[wasm_bindgen]
 extern "C" {
     #[wasm_bindgen(extends = Object)]
-    #[derive(Clone, Debug)]
+    #[derive(Clone)]
     pub type Boolean;
 
     /// The `Boolean()` constructor creates an object wrapper for a boolean value.
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean)
     #[wasm_bindgen(constructor)]
+    #[deprecated(note = "recommended to use `Boolean::from` instead")]
     pub fn new(value: &JsValue) -> Boolean;
 
     /// The `valueOf()` method returns the primitive value of a `Boolean` object.
@@ -481,6 +482,35 @@ extern "C" {
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/valueOf)
     #[wasm_bindgen(method, js_name = valueOf)]
     pub fn value_of(this: &Boolean) -> bool;
+}
+
+impl From<bool> for Boolean {
+    #[inline]
+    fn from(b: bool) -> Boolean {
+        Boolean::unchecked_from_js(JsValue::from(b))
+    }
+}
+
+impl From<Boolean> for bool {
+    #[inline]
+    fn from(b: Boolean) -> bool {
+        b.value_of()
+    }
+}
+
+impl PartialEq<bool> for Boolean {
+    #[inline]
+    fn eq(&self, other: &bool) -> bool {
+        self.value_of() == *other
+    }
+}
+
+impl Eq for Boolean {}
+
+impl fmt::Debug for Boolean {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.value_of().fmt(f)
+    }
 }
 
 // DataView
@@ -1406,7 +1436,7 @@ extern "C" {
 #[wasm_bindgen]
 extern "C" {
     #[wasm_bindgen(extends = Object)]
-    #[derive(Clone, Debug)]
+    #[derive(Clone)]
     pub type Number;
 
     /// The Number.isFinite() method determines whether the passed value is a finite number.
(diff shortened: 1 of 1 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -467,13 +467,14 @@ extern "C" {
 #[wasm_bindgen]
 extern "C" {
     #[wasm_bindgen(extends = Object)]
-    #[derive(Clone, Debug)]
+    #[derive(Clone)]
```
RReverser: Why is `Debug` removed? Looks like an unnecessary breaking change.

crates/js-sys/src/lib.rs
```diff
@@ -483,6 +484,30 @@ extern "C" {
     pub fn value_of(this: &Boolean) -> bool;
 }
 
+impl From<bool> for Boolean {
+    fn from(b: bool) -> Boolean {
+        Boolean { obj: JsValue::from(b) }
```
RReverser: This looks a bit like dark magic :)

I didn't know these types have an internal `obj` field and I guess it might change in the future. Would `JsValue::from_bool(b).unchecked_into()` maybe look better?

crates/js-sys/src/lib.rs
```diff
@@ -1500,6 +1526,35 @@ extern "C" {
...
+        }
+
+        impl PartialEq<$x> for Number {
+            fn eq(&self, other: &$x) -> bool {
+                self.value_of() == f64::from(*other)
+            }
+        }
+    )*)
+}
+number_from!(i8 u8 i16 u16 i32 u32 f32 f64);
```
RReverser: Perhaps a silly suggestion, but why not use transitive generics instead of a custom macro? Something like:

```rust
impl<T> From<T> for Number where f64: From<T> {
  // ...
}
```

The only downside it has is that it makes it impossible to implement custom `From<SomeType> for Number` but I guess we don't really want to support source types that are not convertible to `f64` too?

crates/js-sys/src/lib.rs
```diff
@@ -1500,6 +1526,35 @@ extern "C" {
     pub fn value_of(this: &Number) -> f64;
 }
 
+macro_rules! number_from {
+    ($($x:ident)*) => ($(
+        impl From<$x> for Number {
+            fn from(x: $x) -> Number {
+                Number { obj: x.into() }
```
RReverser: Same regarding using `unchecked_into` perhaps.

crates/js-sys/src/lib.rs
alexcrichton: Good point!

crates/js-sys/src/lib.rs
alexcrichton: That's possible yeah, but I think I'd prefer to stick with the macro strategy to avoid blanket impls unnecessarily blocking out other impls.

crates/js-sys/src/lib.rs
```diff
@@ -483,6 +484,30 @@ extern "C" {
...
+    }
+}
+
+impl From<Boolean> for bool {
+    fn from(b: Boolean) -> bool {
+        b.value_of()
+    }
+}
+
+impl PartialEq<bool> for Boolean {
```
Pauan: Shouldn't we impl `Eq` as well?

Also, shouldn't all of these simple methods be `#[inline]`?

crates/js-sys/src/lib.rs
```diff
@@ -483,6 +484,30 @@ extern "C" {
...
 
+impl From<bool> for Boolean {
+    fn from(b: bool) -> Boolean {
+        Boolean::unchecked_from_js(JsValue::from(b))
+    }
+}
+
+impl From<Boolean> for bool {
+    fn from(b: Boolean) -> bool {
+        b.value_of()
```
Pauan: This can probably be implemented in a more optimized way, but we can save that for later.

crates/js-sys/src/lib.rs
alexcrichton: Indeed!

RReverser:
@alexcrichton I see that tests are now giving deprecation warnings due to usage of `Boolean::new` / `Number::new`.

Maybe it's worth to either remove these tests or mark them with `allow(deprecated)` to silence these warnings?

alexcrichton:
Ah oops yeah, it should be fine to just `#[allow(deprecated)]` them

PR #1449, opened 2019-04-12
Support [NoInterfaceObject] in `web-sys`

### LINKED ISSUE #893: Figure out how to support interfaces with `NoInterfaceObject` attribute

### LINKED ISSUE #1257: Cannot access ANGLE_instanced_arrays or other extensions
I'm trying to use WebGL extensions within web_sys, but cannot seem to access the ANGLE_instanced_arrays interface that exists in this webidl - https://github.com/rustwasm/wasm-bindgen/blob/master/crates/web-sys/webidls/enabled/WebGLRenderingContext.webidl

Here is what I've tried...
let ext = context.get_extension("ANGLE_instanced_arrays").unwrap().unwrap();
let inst_arrays = ext.dyn_into::<web_sys::ANGLE_instanced_arrays>();

I've opened a related [StackOverflow question here](https://stackoverflow.com/questions/54699650/how-can-i-use-webgl-extensions-from-web-sys-in-rust): but wasn't sure that would get your attention.

Apologies if this is something obvious, today is my first day writing Rust.

### LINKED ISSUE #1315: Anchor tag does not contain href property
The `href` attribute is inherited by the `<a>` tag via `HTMLHyperlinkElementUtils` as per
[here](https://github.com/rustwasm/wasm-bindgen/blob/f161717afefd87d1cfea9ce830d055186169a70f/crates/web-sys/webidls/enabled/HTMLAnchorElement.webidl#L38). However for the life of me I can't find the `href` getter anywhere in the docs, nor any mention of `HTMLHyperlinkElementUtils`. `href` is of course the most useful property of `<a>`.

alexcrichton:
This commit enables `[NoInterfaceObject]` annotated interfaces in
`web-sys`. The `NoInterfaceObject` attribute means that there's not
actually a JS class for the object, but all of its properties and such
can still be accessed structually and invoked. This should help provide
more bindings for some more common types on the web!

Note that this builds on recent features to ensure that `dyn_into` and
friends always fail for `NoInterfaceObject` objects because they don't
actually have a class.

Closes #893
Closes #1257
Closes #1315


RReverser:
> Note that this builds on recent features to ensure that dyn_into and
friends always fail for NoInterfaceObject objects because they don't
actually have a class.

I'm not very familiar with `NoInterfaceObject` semantics to be honest, but is that behaviour really intentional? Or would it require some structural comparison instead for duck-typing?

alexcrichton:
[According to the WebIDL spec](https://heycam.github.io/webidl/#NoInterfaceObject) it looks like it's just duck typing for these interfaces (and intentional that it is). It's also noted though that this is a legacy feature which shouldn't be used by anything new, so it's just compat with older APIs for now.

RReverser:
> it looks like it's just duck typing for these interfaces (and intentional that it is)

I see. So that's why I wonder if `false` is correct or one has to actually check the keys to make sure that the object corresponds to the duck interface. Although that would be too expensive I guess.

Another alternative is maybe panicking in `is_type_of` to make it clear that one can convert to/from these objects only using `unchecked_` APIs?

Pauan:
> Although that would be too expensive I guess.

Since it's only run on `dyn_into` (and similar), I think expensive-but-correct is better than fast-but-useless.

> Another alternative is maybe panicking in is_type_of to make it clear that one can convert to/from these objects only using unchecked_ APIs?

I think that's a reasonable compromise. I like it.

alexcrichton:
FWIW I think we in general just need to expose the bindings found in the WebIDL as opposed to ensuring that all the conversions all make sense one way or another. It's already the case that calling `is_instance_of` will cause a JS exception to be thrown because most of these types aren't actually defined in JS so `x instanceof ANGLE_instance_array` will fail.

I think that gracefully failing here is probably better than panicking because it's true that there's no instances of this class. If a panic happens it'd be due to an `unwrap` which would still be loud and in theory easy to spot. 

I don't think we can probe for the structure of an object to see if it's of the right type, because although it may have properties there's no guarantee they're functions with the right arguments and right return values. I think that unconditionally returning `false` and failing all conversions is the right behavior here.

RReverser:
Your arguments make sense to me. Neither approach seems actually 100% correct, just implementing different tradeoffs, so I don't have a strong opinion on which one to choose.

Perhaps the only good thing we could do in the future is actually separating types that have underlying named class vs those that don't, into own traits, so that these cases would be forbidden at compile-time.

crates/webidl/src/first_pass.rs
```diff
@@ -45,6 +45,7 @@ pub(crate) struct FirstPassRecord<'src> {
 pub(crate) struct InterfaceData<'src> {
     /// Whether only partial interfaces were encountered
     pub(crate) partial: bool,
+    pub(crate) has_no_interface: bool,
```
fitzgen: the negative is a bit funky (vs `has_interface: bool`) but I guess it makes sense in this case

alexcrichton: Nah seems reasonable to me to switch!

PR #1450, opened 2019-04-12
Leverage new `is_type_of` for iterator protocol

RReverser:
Treats any object of shape `{ next: function }` as an iterator via new `is_type_of` method. This is consistent with JavaScript iteration protocol which we already respect.

Also fixes a minor issue that `is_function` was unnecessarily called twice (once explicitly and once as part of `dyn_into` which now does the same check).


alexcrichton:
Thanks!

PR #1451, opened 2019-04-13
Transform Math, Reflect and JSON into static namespaces

RReverser:
Inspired by `WebAssembly` and `Intl` definitions, I though it would be good to convert these other imports from types to static imports too.

They can't be called or constructed, so, just like in existing cases, they're already pretty useless as regular types. Hence, it shouldn't break any real-world code depending on wasm-bindgen, but still, as shown by removed tests, technically it might be considered a breaking change.


RReverser:
Fair warning: best to review in no-whitespace diff mode.

alexcrichton:

PR #1456, opened 2019-04-14
js-sys: add Object.fromEntries

### LINKED ISSUE #1453: Stage 4 JS proposal: Object.fromEntries
[This proposal](https://github.com/tc39/proposal-object-from-entries) got stage 4 at the [January 2019 meeting](https://github.com/tc39/tc39-notes/blob/master/es9/2019-01/jan-29.md#objectfromentries-for-stage-4), and is not yet implemented in `js-sys`.

See [spec text](https://tc39.github.io/ecma262/#sec-object.fromentries) and [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries).

bakkot:
Fixes #1453.

I couldn't actually get tests to pass, possibly because Node does not yet support this function.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -2176,6 +2176,13 @@ extern "C" {
     #[wasm_bindgen(static_method_of = Object)]
     pub fn freeze(value: &Object) -> Object;
 
+    /// The Object.fromEntries() method transforms a list of key-value pairs
+    /// into an object.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries)
+    #[wasm_bindgen(static_method_of = Object, catch, js_name = fromEntries)]
+    pub fn from_entries(iterable: &JsValue) -> Result<Object, JsValue>;
+
     /// The Object.getOwnPropertyDescriptor() method returns a
     /// property descriptor for an own property (that is, one directly
     /// present on an object and not in the object's prototype chain)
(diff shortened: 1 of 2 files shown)


bakkot:
Yeah, looks like CI has the same problem as I did locally. I'd appreciate input from a maintainer here - to my (uneducated) eye, it looks like tests in CI are being run in Firefox, but Firefox has had support for `Object.fromEntries` [since version 63](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/63#JavaScript) (released in October 2018) and Azure's Ubuntu agent claims [claims](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md) to be running Firefox 66.

alexcrichton:
Oh so `crates/js-sys/tests/wasm/*` is actually all executed in Node.js which may explain the discrepancy here, but can you try adding the tests to `crates/js-sys/tests/headless.rs` as those should be run in Firefox on Azure?

bakkot:
Hm. I believe Node 12 [will include](https://github.com/nodejs/node/blob/v12.x-staging/deps/v8/include/v8-version.h) v8 7.4, which [includes](https://v8.dev/blog/v8-release-73) `Object.fromEntries`. It's due to be cut on [April 23](https://github.com/nodejs/Release). So maybe it'd be simpler to just wait for that?

alexcrichton:
Seems reasonable to me!

alexcrichton:
I just updated to node 12 in https://github.com/rustwasm/wasm-bindgen/pull/1484, want to rebase this and see if it works?

bakkot:
@alexcrichton looks like it worked! (Once I fixed the silly bug in the test, anyway.)

PR #1459, opened 2019-04-15
Fix more DataView documentation mistakes

RReverser:
I fixed few copy-paste mistakes in https://github.com/rustwasm/wasm-bindgen/pull/1372, but haven't noticed that they go even deeper :(


alexcrichton:
Thanks!

Issue #1054, opened 2018-11-27
Add `JSVALUE_NULL` static, remove `JsValue::null()` function and `JsValue::NULL` constant


alexcrichton:
Ideally we want to have an invariant that `JsValue` (as a bare type) represents an owned value which is "heap allocated" and tracks and index that needs to be removed. Unfortunately though having constructors like `JsValue::null()` violate this invariant because they return a bare `JsValue` that isn't an owned value, but rather a transient one.

This in turns [requires a check on `drop`](https://github.com/rustwasm/wasm-bindgen/blob/e3b628689fab5c1038617b8b43b6a498bab9fa97/src/lib.rs#L541-L545) unconditionally. This required branch is a bummer in an otherwise quite streamlined type!

It'd be great if we could solve this problem by actually providing this guarantee. That would entail removing the `JsValue::NULL` (and other associated) constant as well as removing the `JsValue::null()` constructor (although we could keep the constructor if it returned `&'static JsValue`). By removing these we can remove entirely all branches in the destructor, ensuring that when `JsValue::drop` is executed we unconditionally execute the relevant destructor.

This is a breaking change, however, so we can't even do this just yet! I'm filing this issue so we can track a future implementation of it.

PR #1462, opened 2019-04-15
Remove repetitive branching on value drop

RReverser:
This was unnecessarily checking that the value is not reserved twice (on both sides).

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -854,7 +854,7 @@ impl<'a> Context<'a> {
         // remove usages of this.
         self.bind("__wbindgen_object_drop_ref", &|me| {
             me.expose_drop_ref();
-            Ok(String::from("function(i) { dropObject(i); }"))
+            Ok(String::from("dropObject"))
         })?;
 
         Ok(())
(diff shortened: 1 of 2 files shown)


alexcrichton:
Yeah the branch exists in Rust to ensure that we can constant-propagate and learn that `drop(JsValue::null())` is a no-op. Note though that it also needs to be present because if you call `some_imported_function(JsValue::null())` then JS is the one the drops the reference and needs to understand that it can't free everything.

https://github.com/rustwasm/wasm-bindgen/issues/1054 is an issue about removing these branches, both in Rust and in JS. With that issue implemented I believe both can be safely removed.

RReverser:
> Yeah the branch exists in Rust to ensure that we can constant-propagate and learn that drop(JsValue::null()) is a no-op.

Yeah I assumed so but then thought that it's unlikely to be a frequent case, and in most cases this just performs an unnecessary check instead?

Feel free to close if you think otherwise.

alexcrichton:
Ok in the absence of data one way or another I think I'd prefer to leave it as is (with a true fix in #1054 eventually). If it's determined to be more beneficial to remove the branch though we certainly can!

PR #1463, opened 2019-04-15
added Atomics and SharedArrayBuffer 

### LINKED ISSUE #1443: Atomics and SharedArrayBuffer support
### Motivation

There is no `Atomics` and `SharedArrayBuffer` objects in `js_sys` yet. Adding them will be helpful for #1379

ibaryshnikov:
Added `Atomics` and `SharedArrayBuffer` to `js-sys`.
Ready for review.

Closes #1443

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -624,18 +624,18 @@ pub mod Atomics {
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
         #[wasm_bindgen(static_method_of = Atomics, catch)]
-        pub fn wait(typed_array: &JsValue, index: u32, value: i32) -> Result<JsValue, JsValue>;
+        pub fn wait(typed_array: &JsValue, index: u32, value: i32) -> Result<String, JsValue>;
 
         /// Like `wait()`, but with timeout
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
-        #[wasm_bindgen(static_method_of = Atomics, catch)]
+        #[wasm_bindgen(static_method_of = Atomics, catch, js_name = wait)]
         pub fn wait_with_timeout(
             typed_array: &JsValue,
             index: u32,
             value: i32,
             timeout: f64,
-        ) -> Result<JsValue, JsValue>;
+        ) -> Result<String, JsValue>;
 
         /// The static `Atomics.xor()` method computes a bitwise XOR
         /// with a given value at a given position in the array,
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -624,7 +624,7 @@ pub mod Atomics {
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
         #[wasm_bindgen(static_method_of = Atomics, catch)]
-        pub fn wait(typed_array: &JsValue, index: u32, value: i32) -> Result<String, JsValue>;
+        pub fn wait(typed_array: &JsValue, index: u32, value: i32) -> Result<JsString, JsValue>;
 
         /// Like `wait()`, but with timeout
         ///
@@ -635,7 +635,7 @@ pub mod Atomics {
             index: u32,
             value: i32,
             timeout: f64,
-        ) -> Result<String, JsValue>;
+        ) -> Result<JsString, JsValue>;
 
         /// The static `Atomics.xor()` method computes a bitwise XOR
         /// with a given value at a given position in the array,
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -438,6 +438,47 @@ extern "C" {
     pub fn slice_with_end(this: &ArrayBuffer, begin: u32, end: u32) -> ArrayBuffer;
 }
 
+
+// SharedArrayBuffer
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(extends = Object)]
+    #[derive(Clone, Debug)]
+    pub type SharedArrayBuffer;
+
+    /// The `SharedArrayBuffer` object is used to represent a generic,
+    /// fixed-length raw binary data buffer, similar to the `ArrayBuffer`
+    /// object, but in a way that they can be used to create views
+    /// on shared memory. Unlike an `ArrayBuffer`, a `SharedArrayBuffer`
+    /// cannot become detached.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer)
+    #[wasm_bindgen(constructor)]
+    pub fn new(length: u32) -> SharedArrayBuffer;
+
+    /// The byteLength accessor property represents the length of
+    /// an `SharedArrayBuffer` in bytes. This is established when
+    /// the `SharedArrayBuffer` is constructed and cannot be changed.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer/byteLength)
+    #[wasm_bindgen(method, getter, js_name = byteLength)]
+    pub fn byte_length(this: &SharedArrayBuffer) -> u32;
+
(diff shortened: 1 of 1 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -463,6 +504,134 @@ extern "C" {
     pub fn values(this: &Array) -> Iterator;
 }
 
+// Atomics
+#[wasm_bindge]
```
alexcrichton: I thin kthere may be a typo here

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +504,134 @@ extern "C" {
...
+    pub type Atomics;
+
+    /// The static `Atomics.add()` method adds a given value at a given
+    /// position in the array and returns the old value at that position.
+    /// This atomic operation guarantees that no other write happens
+    /// until the modified value is written back.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/add)
+    #[wasm_bindgen(static_method_of = Atomics, catch)]
+    pub fn add(typed_array: &JsValue, index: u32, value: JsValue) -> Result<JsValue, JsValue>;
```
alexcrichton: Note that idiomatically in our bindings we try to take `&JsValue` wherever possible instead of `JsValue`, but I wonder if perhaps this should take/return `f64`? I believe you can only pass in numbers here, right?

alexcrichton: (the point of `JsValue` vs `f64` vs something else applies to most methods here I think)

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +504,134 @@ extern "C" {
...
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
+    #[wasm_bindgen(static_method_of = Atomics, catch)]
+    pub fn wait(typed_array: &JsValue, index: u32, value: JsValue) -> Result<JsValue, JsValue>;
+
+    /// Like `wait()`, but with timeout
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
+    #[wasm_bindgen(static_method_of = Atomics, catch)]
+    pub fn wait_with_timeout(typed_array: &JsValue, index: u32, value: JsValue, timeout: u32) -> Result<JsValue, JsValue>;
```
alexcrichton: I think the `timeout` argument here may be best represented as `f64` as well perhaps?

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +503,152 @@ extern "C" {
...
+        pub type Atomics;
+
+        /// The static `Atomics.add()` method adds a given value at a given
+        /// position in the array and returns the old value at that position.
+        /// This atomic operation guarantees that no other write happens
+        /// until the modified value is written back.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/add)
+        #[wasm_bindgen(static_method_of = Atomics, catch)]
+        pub fn add(typed_array: &JsValue, index: u32, value: f64) -> Result<f64, JsValue>;
```
ibaryshnikov: @alexcrichton since it returns the old value, is it fine to use `Result<f64, JsValue>` as a function return argument?
Another question: `typed_array` argument can be of any type, for example `Int8Array`. Is it fine to keep `value` type as `f64`?

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +503,152 @@ extern "C" {
...
+        /// The static `Atomics.wait()` method verifies that a given
+        /// position in an `Int32Array` still contains a given value
+        /// and if so sleeps, awaiting a wakeup or a timeout.
+        /// It returns a string which is either "ok", "not-equal", or "timed-out".
+        /// Note: This operation only works with a shared `Int32Array`
+        /// and may not be allowed on the main thread.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
+        #[wasm_bindgen(static_method_of = Atomics, catch)]
+        pub fn wait(typed_array: &JsValue, index: u32, value: i32) -> Result<JsValue, JsValue>;
```
ibaryshnikov: @alexcrichton it works only with `Int32Array`. Is it fine to keep `value` type as `i32`?

crates/js-sys/src/lib.rs
alexcrichton: Yeah I think `f64` is appropriate here since it's largely just "some JS number" which becomes an f64 anyway. Additionally returning `Result<f64, JsValue>` sounds good to me as well!

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +503,152 @@ extern "C" {
...
+
+        /// Like `wait()`, but with timeout
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait)
+        #[wasm_bindgen(static_method_of = Atomics, catch)]
```
alexcrichton: Does this need `js_name = wait`?

crates/js-sys/src/lib.rs
alexcrichton: Yeah if this is spec'd as only supporting a `Int32Array` I think it's fine to use an `i32` here. Should the `Ok` value of the `Result` returned also be `i32`?

ibaryshnikov: It returns "ok" | "not-equal" | "timed-out". Should it be `String` or `JsString` instead?

alexcrichton: Well then, that's news to me! In that case yeah let's change this to return `JsString`

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +503,152 @@ extern "C" {
...
+/// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics)
+#[allow(non_snake_case)]
+pub mod Atomics {
+    use super::*;
+
+    #[wasm_bindgen]
+    extern "C" {
+        #[derive(Clone, Debug)]
+        #[wasm_bindgen(extends = Object)]
+        pub type Atomics;
```
alexcrichton: Oh so I think since we're using `pub mod Atomics` we can remove this `pub type`

crates/js-sys/src/lib.rs
```diff
@@ -463,6 +503,152 @@ extern "C" {
...
+        /// The static `Atomics.load()` method returns a value at a given
+        /// position in the array.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/load)
+        #[wasm_bindgen(static_method_of = Atomics, catch)]
+        pub fn load(typed_array: &JsValue, index: u32) -> Result<f64, JsValue>;
+
+        /// The static `Atomics.notify()` method notifies up some agents that
+        /// are sleeping in the wait queue.
+        /// Note: This operation works with a shared `Int32Array` only.
```
alexcrichton: Since this only works with `Int32Array` I wonder if it would be useful to take `Int32Array` as the first argument actually? (same with `wait` and such I believe)

alexcrichton:
Looking great @ibaryshnikov!

Thinking some more on this, what do you think of changing `f64` to `i32`? It looks like it's documented that you can only pass integer arrays up to 32 bits, so using `i32` may actually cover everything here and may be easier to use than dealing with floats (and perhaps faster since floats sometimes have a bit of a cost to them).

ibaryshnikov:
Oh, I didn't notice it in the beginning. I thought that `f32` and `f64` typed arrays were allowed for operations like `add`. Using `i32` looks fine to me. What about `u32`?

alexcrichton:
Both u32 and i32 work but i32 doesn't need a job shim to switch the number to unsigned, so I think I'd lean towards i32 for this

ibaryshnikov:
Fine. I also need to change `static_method_of = Atomics` to `js_namespace = Atomics` after changing from `pub type Atomics` to `pub mod Atomics`, right?

alexcrichton:
That sounds right yeah, I sort of forget the specifics but if it works it works!

ibaryshnikov:
I'd like to write tests, but I can do it in a separate pr, so we can merge this I think.

ibaryshnikov:
@alexcrichton thank you for very useful comments!

PR #1465, opened 2019-04-16
Add env vars to filter `wasm-bindgen-test-runner` tests

### LINKED ISSUE #822: Gracefully handle attempts to run tests configured for headless browsers in node and vice versa
This will be used (perhaps indirectly?) by `wasm-pack test`. I expect users will do (or have CI configured to do) something like this:

```
wasm-pack test --node
wasm-pack test --firefox
```

with a tests directory like

```
my-crate/
  |- tests/
  |    |- node.rs        # tests for node.js
  |    `- browsers.rs    # tests for browsers
```

and we should ultimately do the Right Thing and just skip tests that aren't configured for the given target, but this requires some sort of introspection/communication in the test runner and/or test `.wasm` binary itself.

Perhaps a custom section in the test binary that identifies what it is configured to target?

cc @alexcrichton @ashleygwilliams

### LINKED ISSUE #1458: Running tests through `wasm-pack` doesn't work if JS snippets are used
### Describe the Bug

Using JS snippets prevents `wasm-pack` from being able to run tests.

### Steps to Reproduce

1. Create a new crate.
2. Paste this into `Cargo.toml`:

```toml
[dependencies]
wasm-bindgen = "0.2"

[dev-dependencies]
wasm-bindgen-test = "0.2"
```
3. Paste this into `src/lib.rs`:

```rust
extern crate wasm_bindgen;

#[cfg(test)]
extern crate wasm_bindgen_test;

use wasm_bindgen::prelude::*;

#[cfg(test)]
use wasm_bindgen_test::*;

#[cfg(test)]
#[wasm_bindgen_test]
fn foobar() {
    #[wasm_bindgen(inline_js = "export function foo() {}")]
    extern "C" {
        fn foo();
    }

    foo();
}
```

4. Run:

```
$ wasm-pack test --chrome --headless
```

### Actual Behavior

```
$ wasm-pack test --chrome --headless
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running target/wasm32-unknown-unknown/debug/deps/foobar-27ec9db4a368c0d0.wasm
error: executing `wasm-bindgen` over the wasm file
	caused by: failed to generate bindings for JS import `foo`
	caused by: local JS snippets are not supported with `--target nodejs`; see rustwasm/rfcs#6 for more details, but this restriction will be lifted in the future
error: test failed, to rerun pass '--lib'
Error: Running Wasm tests with wasm-bindgen-test failed
Caused by: failed to execute `cargo test`: exited with exit code: 1
```

### Expected Behavior

I expect the tests to run.

### Additional Context

I need this to be able to run `stdweb`'s test suite. I could probably hack around it by having a custom test runner, but I'd rather not do that and just use standard tooling.

Also, it seems somewhat counterintuitive that even though I want to run the tests under headless Chrome they're being built for Node.js by `wasm-pack`? Is this expected?

```
$ wasm-pack --version
wasm-pack 0.8.1
$ wasm-bindgen --version
wasm-bindgen 0.2.42
```

alexcrichton:
This is intended to handle #1458 and #822. These issues stem from
behavior where:

    wasm-pack test --node

will actually run both Node.js and browser tests! Two new env vars are
read here, `WASM_BINDGEN_TEST_ONLY_{NODE,WEB}`, which control which
tests are executed by `wasm-bindgen-test-runner`. The intention is that
these will be set by `wasm-pack` to configure which tests are run, and
if test suites are skipped due to the env vars we'll print an
informational message indicating how they can be run.

Closes #822
Closes #1458

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -103,6 +103,30 @@ fn rmain() -> Result<(), Error> {
     let headless = env::var("NO_HEADLESS").is_err();
     let debug = env::var("WASM_BINDGEN_NO_DEBUG").is_err();
 
+    // Gracefully handle requests to execute only node or only web tests.
+    if env::var_os("WASM_BINDGEN_TEST_ONLY_NODE").is_some() {
+        if !node {
+            println!("this test suite is only configured to run in a browser, \
+                      but we're only testing node.js tests so skipping");
+            return Ok(());
+        }
+    }
+    if env::var_os("WASM_BINDGEN_TEST_ONLY_WEB").is_some() {
+        if node {
+            println!("\
+This test suite is only configured to run in node.js, but we're only running
+browser tests so skipping. If you'd like to run the tests in a browser
+include this in your crate when testing:
+
+    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+You'll likely want to put that in a `#[cfg(test)]` module or at the top of an
+integration test.\
+");
+            return Ok(());
+        }
+    }
+
     // Make the generated bindings available for the tests to execute against.
     shell.status("Executing bindgen...");
     let mut b = Bindgen::new();


RReverser:
There is no reason for Node.js not to work in experimental-modules mode though, right?

alexcrichton:
Yes when node natively supports ES modules then wasm-bindgen can just require that version of node or later when snippets are required. We don't currently export functionality to use the experimental modules because they're a moving target we haven't been able to keep up with

RReverser:
> We don't currently export functionality to use the experimental modules 

Ah interesting, I was pretty sure I used them with wasm-bindgen, but I guess they were removed in the meanwhile?

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -103,6 +103,26 @@ fn rmain() -> Result<(), Error> {
...
+            return Ok(());
+        }
+    }
+    if env::var_os("WASM_BINDGEN_TEST_ONLY_WEB").is_some() {
+        if node {
+            println!("this test suite is only configured to run in node.js, \
+                      but we're only running browser tests so skipping");
+            println!("");
+            println!("if you'd like to run the tests in a browser you can \
+                      use the `wasm_bindgen_test_configure!(run_in_browser);` \
```
fitzgen: Might as well add the import and everything in a 4-space indented code block, since people will likely want to just copy paste this immediately.

alexcrichton:
@RReverser I added support some time ago to toy around with it but it's likely regressed in the meantime and it wasn't ever really intended to be official

PR #1467, opened 2019-04-17
Generate bindings for indexed struct properties

RReverser:
This allows to export fields of tuple-like structs as indexed JS properties.

While we could rename it to `_0`, `$0` or something else, this feels like the most natural mapping of "indexed" properties to JavaScript concepts, and just works out of the box with current JS and TypeScript codegens.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -227,7 +227,7 @@ pub struct Struct {
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
 #[derive(Clone)]
 pub struct StructField {
-    pub name: Ident,
+    pub name: syn::Member,
     pub struct_name: Ident,
     pub readonly: bool,
     pub ty: syn::Type,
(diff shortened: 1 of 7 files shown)


alexcrichton:
Seems reasonable to me, thanks @RReverser! Could you ensure that this is documented in the reference as well?

RReverser:
Ah yes. I'm leaving for the Easter holiday now but can back to it afterwards.

RReverser:
@alexcrichton Hmm, I might need help figuring out where best to put it. There currently doesn't seem to be a dedicated section describing how named properties are bound on the JS side? (I've found only `exported_types.rs`, but it doesn't demonstrate properties usage, only the class itself)

alexcrichton:
Ah that's true yeah, mind starting a page which describes the JS api created for exported rust structures? It could go in the area for attributes on exports perhaps in the reference

RReverser:
I'm thinking maybe just adding some property usage to the same existing `exported-rust-types.md` + `exported_types.rs` + `exported_types.js`?

RReverser:
I've added property examples to files mentioned above, let me know how it looks.

alexcrichton:
Thanks!

PR #1469, opened 2019-04-17
Fix or suppress {primitive}::new warnings in tests

RReverser:
Constructing boxed primitives was deprecated in #1447.

Some tests have been still using these methods, so this PR either updates them to use newly added {primitive}::from implementations or adds `#[allow(deprecated)]` where necessary.

r? @alexcrichton


alexcrichton:
Thanks!

PR #1473, opened 2019-04-18
Add JsString <-> char conversions

### LINKED ISSUE #1363: Consider adding JsValue::as_char
### Motivation

Currently it's not too easy to convert a single char, which is represented in JS as a string, and involves either getting a heap-allocated `String` and then extracting a char, or doing manual conversions that use `char_code_at` / `code_point_at` which currently don't have convenient types (see #1362), and also checking the `length` to make sure it's indeed a char and not an arbitrary string.

### Proposed Solution

Add `as_char` on `JsValue` that would make such conversions easier by performing `length` check and `.codePointAt` on JS side, passing `Option<u32>` back to Rust and converting back to char using `char::from_u32` internally.

### Alternatives

Wait for fixes to #1362, and then ask users to use js-sys and perform conversion manually, but the length check is still non-trivial to implement correctly (because JS `length` can be `1` or `2` for a single-codepoint string) and feels like belongs to the core of wasm-bindgen.

RReverser:
These are pretty common and already supported via ABI conversions, yet pretty easy to get wrong when converting them manually.

Fixes #1363.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3641,11 +3641,47 @@ impl JsString {
     ) -> impl ExactSizeIterator<Item = u16> + DoubleEndedIterator<Item = u16> + 'a {
         (0..self.length()).map(move |i| self.char_code_at(i) as u16)
     }
+
+    /// If this string consists of a single Unicode code point, then this method
+    /// converts it into a Rust `char` without doing any allocations.
+    ///
+    /// If this JS value is not a valid UTF-8 or consists of more than a single
+    /// codepoint, then this returns `None`.
+    ///
+    /// Note that a single Unicode code point might be represented as more than
+    /// one code unit on the JavaScript side. For example, a JavaScript string
+    /// `"\uD801\uDC37"` is actually a single Unicode code point U+10437 which
+    /// corresponds to a character '𐐷'.
+    pub fn as_char(&self) -> Option<char> {
+        let len = self.length();
+
+        if len == 0 || len > 2 {
+            return None;
+        }
+
+        // This will be simplified when definitions are fixed:
+        // https://github.com/rustwasm/wasm-bindgen/issues/1362
+        let cp = self.code_point_at(0).as_f64().unwrap_throw() as u32;
+
+        let c = std::char::from_u32(cp)?;
+
+        if c.len_utf16() as u32 == len {
+            Some(c)
+        } else {
+            None
+        }
+    }
 }
 
 impl PartialEq<str> for JsString {
     fn eq(&self, other: &str) -> bool {
         String::from(self) == other
+    }
+}
+
+impl PartialEq<char> for JsString {
+    fn eq(&self, other: &char) -> bool {
+        self.as_char() == Some(*other)
     }
 }
 
@@ -3676,6 +3712,13 @@ impl<'a> From<&'a str> for JsString {
 impl From<String> for JsString {
     fn from(s: String) -> Self {
         From::from(&*s)
+    }
+}
+
+impl From<char> for JsString {
+    #[inline]
+    fn from(c: char) -> Self {
+        JsString::from_code_point1(c as u32).unwrap_throw()
     }
 }
 
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3679,12 +3679,6 @@ impl PartialEq<str> for JsString {
     }
 }
 
-impl PartialEq<char> for JsString {
-    fn eq(&self, other: &char) -> bool {
-        self.as_char() == Some(*other)
-    }
-}
-
 impl<'a> PartialEq<&'a str> for JsString {
     fn eq(&self, other: &&'a str) -> bool {
         <JsString as PartialEq<str>>::eq(self, other)
--- a/crates/js-sys/tests/wasm/JsString.rs
+++ b/crates/js-sys/tests/wasm/JsString.rs
@@ -553,3 +553,14 @@ fn is_valid_utf16() {
     assert!(!JsString::from_char_code1(0xd800).is_valid_utf16());
     assert!(!JsString::from_char_code1(0xdc00).is_valid_utf16());
 }
+
+#[wasm_bindgen_test]
+fn as_char() {
+    assert_eq!(JsString::from('a').as_char(), Some('a'));
+    assert_eq!(JsString::from('🥑').as_char(), Some('🥑'));
+    assert_eq!(JsString::from("").as_char(), None);
+    assert_eq!(JsString::from("ab").as_char(), None);
+    assert_eq!(JsString::from_char_code1(0xd800).as_char(), None);
+    assert_eq!(JsString::from_char_code1(0xdc00).as_char(), None);
+    assert_eq!(JsString::from_char_code1(0xdfff).as_char(), None);
+}


alexcrichton:
Seems reasonable to me, thanks @RReverser! Just to confirm with a second opinion, @Pauan mind checking as well?

crates/js-sys/src/lib.rs
```diff
@@ -3641,6 +3641,31 @@ impl JsString {
     ) -> impl ExactSizeIterator<Item = u16> + DoubleEndedIterator<Item = u16> + 'a {
         (0..self.length()).map(move |i| self.char_code_at(i) as u16)
     }
+
+    /// If this string consists of a single UTF-8 codepoint, then this function
+    /// returns a Rust representation of that character without any allocations.
+    ///
+    /// If this JS value is not a valid UTF-8 or consists of more than a single
+    /// codepoint, then this returns `None`.
```
Pauan: I think it would be useful to clarify to the user what exactly a "UTF-8 codepoint" is, since it's easy to confuse "code point" with "scalar value" or "character" or "char code".

Also, "UTF-8 codepoint" is technically incorrect, since UTF-8 is an encoding, but code points transcend encodings.

So I would rename it to "Unicode code point", and I would also add a small section describing what it is:

> A JavaScript string that contains multiple characters might actually be a single Unicode code point. For example, `"\uD801\uDC37"` is the single Unicode code point `U+10437` (which is the same as the Rust character `'𐐷'`).

crates/js-sys/src/lib.rs
```diff
@@ -3641,6 +3641,31 @@ impl JsString {
     ) -> impl ExactSizeIterator<Item = u16> + DoubleEndedIterator<Item = u16> + 'a {
         (0..self.length()).map(move |i| self.char_code_at(i) as u16)
     }
+
+    /// If this string consists of a single UTF-8 codepoint, then this function
```
Pauan: Nit: function -> method

crates/js-sys/src/lib.rs
```diff
@@ -3641,6 +3641,31 @@ impl JsString {
...
+    /// returns a Rust representation of that character without any allocations.
```
Pauan: I think it's a bit awkward to call this a "Rust representation". Perhaps it's better to say something like "converts it into a Rust `char` (without doing any allocations)."

crates/js-sys/src/lib.rs
```diff
@@ -3641,6 +3641,31 @@ impl JsString {
...
+            return None;
+        }
+
+        // This will be simplified when definitions are fixed:
+        // https://github.com/rustwasm/wasm-bindgen/issues/1362
+        let cp = self.code_point_at(0).as_f64().unwrap_throw() as u32;
+
+        let c = std::char::from_u32(cp)?;
+
+        if c.len_utf16() as u32 == len {
```
Pauan: Is this needed? My understanding is that `std::char::from_u32` already returns `None` for invalid characters (including surrogate pairs).

If you just want it for extra safety, couldn't it be a `debug_assert`?

crates/js-sys/src/lib.rs
```diff
@@ -3649,6 +3674,12 @@ impl PartialEq<str> for JsString {
     }
 }
 
+impl PartialEq<char> for JsString {
```
Pauan: This feels pretty awkward. It's similar to implementing `PartialEq<u32> for Vec<u32>` (which would obviously be wrong).

`String` and `&str` in Rust do not define `PartialEq` for `char`, so this is unprecedented.

On the other hand, I can see the practical use for it, so I'm undecided on whether it's a good idea or not.

crates/js-sys/src/lib.rs
```diff
@@ -3679,6 +3710,12 @@ impl From<String> for JsString {
     }
 }
 
+impl From<char> for JsString {
+    fn from(c: char) -> Self {
```
Pauan: Should probably be `#[inline]`

crates/js-sys/tests/wasm/JsString.rs
```diff
@@ -553,3 +553,13 @@ fn is_valid_utf16() {
...
 }
+
+#[wasm_bindgen_test]
+fn as_char() {
+    assert_eq!(JsString::from('a').as_char(), Some('a'));
+    assert_eq!(JsString::from('🥑').as_char(), Some('🥑'));
+    assert_eq!(JsString::from("").as_char(), None);
+    assert_eq!(JsString::from("abc").as_char(), None);
+    assert_eq!(JsString::from_char_code1(0xd800).as_char(), None);
+    assert_eq!(JsString::from_char_code1(0xdc00).as_char(), None);
```
Pauan: Can we also get a test for `U+DFFF`?

crates/js-sys/src/lib.rs
RReverser: This is needed not for invalid characters, but to ensure this entire string is this character. E.g. without this check `"ab"` string would silently decode to just `'a'`.

crates/js-sys/src/lib.rs
RReverser: > String and &str in Rust do not define PartialEq for char, so this is unprecedented.

I think this is a bit different, because, unlike with str/char, with JsString we have a single type representation for both, so it seems to make sense to define traits for both too.

RReverser:
@Pauan All your other suggestions make sense to me.

Any chance you could submit them using "Github suggestion" comments which I could easily accept (including from the phone)? 

Otherwise I won't get around to this until after the Easter holiday (later next week).

crates/js-sys/src/lib.rs
Pauan: In JS, sure, but this is comparing a Rust `char` to a JS string, so they're actually not a single type. It's a tricky situation.

RReverser: Yeah, but where is `From`, it seems right to have `PartialEq` as well. 

Idk, if you feel strongly opposed, I can remove it, but it seems useful to have these symmetrical.

crates/js-sys/src/lib.rs
RReverser: I'm usually strictly against using `#[inline]` as an anti-pattern, but given how widespread it is in the codebase already, I guess doesn't harm to add here too.

crates/js-sys/src/lib.rs
RReverser: > since it's easy to confuse "code point" with "scalar value" or "character" or "char code"

I'd normally expect people to be familiar with "code point" vs "code unit" / "char code" distinction if they work with JS strings, but I guess doesn't harm adding an explanation.

crates/js-sys/src/lib.rs
RReverser: I've added a test to use such string to make it explicit.

crates/js-sys/src/lib.rs
RReverser: Fixed.

crates/js-sys/src/lib.rs
RReverser: Done.

crates/js-sys/tests/wasm/JsString.rs
RReverser: Done.

crates/js-sys/src/lib.rs
RReverser: Fixed.

crates/js-sys/src/lib.rs
RReverser: Done, although I've reworded it since "character" is an ambiguous term, especially in this context.

RReverser:
Ok fixed most nits (except the `PartialEq` thing, which I still think is useful in this context).

alexcrichton:
I think I would agree with @Pauan here actually, could the `From` be removed as well as the `PartialEq`? They're compatible to add later but currently inconsistent with Rust's libstd

RReverser:
Hmm, `From` is actually half of the PR... I don't think comparing with libstd makes sense, as it doesn't have string -> char conversion either, because, again, it has a separate type to represent it.

crates/js-sys/src/lib.rs
Pauan: I think most programmers in general (whether JavaScript or not) have no idea how Unicode works, and frequently get the terms mixed up.

Most programmers tend to just assume everything is ASCII.

Pauan: Nit: "corresponds to a character '𐐷'." -> "corresponds to the character '𐐷'."

crates/js-sys/src/lib.rs
Pauan: Ah, right, I hadn't considered that case.

crates/js-sys/src/lib.rs
Pauan: Why do you think it's an anti-pattern? What's wrong with improving performance in non-LTO builds? Keep in mind that LTO is disabled by default even in release mode.

Obviously inlining can be overused, but in the case where the function/method is known to be small (like in this case), why not?

Pauan:
Sorry for the delay, I recently caught a nasty cold.

I'm a bit wary of `PartialEq`, but I'm not strongly against it.

I feel like `From` is quite reasonable, so most of my concerns are with `PartialEq`.

I think they do seem rather useful, since JS itself doesn't make a distinction between strings and chars.

RReverser:
> Sorry for the delay, I recently caught a nasty cold.

No worries and get well!

RReverser:
Alright, I removed `PartialEq<char>` for now to avoid controversy around the PR and, indeed, we can still add it in the future.

alexcrichton:
Thanks @RReverser!

PR #1483, opened 2019-04-25
Fix optional arguments in TypeScript

### LINKED ISSUE #1129: Support for named and optional parameters in .d.ts

dbrgn:
For a function like this:

```rust
#[wasm_bindgen]
pub fn opt_fn(_a: Option<i32>) -> Option<i32> {
    None
}
```

...the currently generated typescript looks like this:

```typescript
/**
* @param {number | undefined} _a
* @returns {number | undefined}
*/
export function opt_fn(_a: number | undefined): number | undefined;
```

Support for `undefined` in the parameter type was added in #1201. However, while type `number | undefined` is technically optional at runtime, typechecking tools will still consider the parameter to be required. (CC @rhysd, please correct me if this information is actually incorrect.)

Instead, the TypeScript should look like this, with a `?` appended to the parameter name:

```typescript
/**
* @param {number | undefined} _a
* @returns {number | undefined}
*/
export function opt_fn(_a?: number): number | undefined;
```

Reference: https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters

This PR should actually fix #1129.

Some notes: I also adjusted `crates/typescript-tests/src/opt_args_and_ret.ts`, but I could not find out whether and where this file is actually used. I think it's dead code, but I might be wrong.

### DIFF
--- a/crates/cli-support/src/js/js2rust.rs
+++ b/crates/cli-support/src/js/js2rust.rs
@@ -2,6 +2,22 @@ use crate::descriptor::{Descriptor, Function};
 use crate::js::Context;
 use failure::{bail, Error};
 
+pub struct JsArgument {
+    pub optional: bool,
+    pub name: String,
+    pub type_: String,
+}
+
+impl JsArgument {
+    fn required(name: String, type_: String) -> Self {
+        Self { optional: false, name, type_ }
+    }
+
+    fn optional(name: String, type_: String) -> Self {
+        Self { optional: true, name, type_ }
+    }
+}
+
 /// Helper struct for manufacturing a shim in JS used to translate JS types to
 /// Rust, aka pass from JS back into Rust
 pub struct Js2Rust<'a, 'b: 'a> {
@@ -12,7 +28,7 @@ pub struct Js2Rust<'a, 'b: 'a> {
     rust_arguments: Vec<String>,
 
     /// Arguments and their types to the JS shim.
-    pub js_arguments: Vec<(String, String)>,
+    pub js_arguments: Vec<JsArgument>,
 
     /// Conversions that happen before we invoke the wasm function, such as
     /// converting a string to a ptr/length pair.
@@ -173,7 +189,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
 
         if let Some(kind) = arg.vector_kind() {
             self.js_arguments
-                .push((name.clone(), kind.js_ty().to_string()));
+                .push(JsArgument::required(name.clone(), kind.js_ty().to_string()));
 
             let func = self.cx.pass_to_wasm_function(kind)?;
             let val = if optional {
@@ -221,7 +237,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
         }
 
         if arg.is_anyref() {
-            self.js_arguments.push((name.clone(), "any".to_string()));
+            self.js_arguments.push(JsArgument::required(name.clone(), "any".to_string()));
             if self.cx.config.anyref {
                 if optional {
                     self.cx.expose_add_to_anyref_table()?;
@@ -250,7 +266,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
 
             if arg.is_wasm_native() {
                 self.js_arguments
-                    .push((name.clone(), "number | undefined".to_string()));
+                    .push(JsArgument::optional(name.clone(), "number".to_string()));
 
                 if self.cx.config.debug {
                     self.cx.expose_assert_num();
@@ -272,7 +288,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
 
             if arg.is_abi_as_u32() {
                 self.js_arguments
-                    .push((name.clone(), "number | undefined".to_string()));
+                    .push(JsArgument::optional(name.clone(), "number".to_string()));
 
                 if self.cx.config.debug {
                     self.cx.expose_assert_num();
@@ -299,7 +315,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
                 };
                 self.cx.expose_uint32_memory();
                 self.js_arguments
-                    .push((name.clone(), "BigInt | undefined".to_string()));
+                    .push(JsArgument::optional(name.clone(), "BigInt".to_string()));
                 self.prelude(&format!(
                     "
                         {f}[0] = isLikeNone({name}) ? BigInt(0) : {name};
@@ -320,7 +336,7 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
             match *arg {
                 Descriptor::Boolean => {
                     self.js_arguments
-                        .push((name.clone(), "boolean | undefined".to_string()));
+                        .push(JsArgument::optional(name.clone(), "boolean".to_string()));
                     if self.cx.config.debug {
                         self.cx.expose_assert_bool();
                         self.prelude(&format!(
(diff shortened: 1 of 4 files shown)


c410-f3r:
Nice work!

> Some notes: I also adjusted crates/typescript-tests/src/opt_args_and_ret.ts, but I could not find out whether and where this file is actually used. I think it's dead code, but I might be wrong.

It is used for local and [CI](https://github.com/rustwasm/wasm-bindgen/blob/cc891090046d07134c1f034a1bf03c43f3df9bad/azure-pipelines.yml#L134) tests.

crates/cli-support/src/js/js2rust.rs
```diff
@@ -2,6 +2,22 @@ use crate::descriptor::{Descriptor, Function};
...
+
+impl JsArgument {
+    fn required(name: String, type_: String) -> Self {
+        Self { optional: false, name, type_ }
+    }
+
+    fn optional(name: String, type_: String) -> Self {
+        Self { optional: true, name, type_ }
+    }
+}
```
c410-f3r: Should the names of these methods be changed to new_(required|optional)?

dbrgn:
> It is used for local and CI tests.

I assumed so, but when I run those tests (with `run.sh`) with bad TS type declarations, nothing fails...

c410-f3r:
> I assumed so, but when I run those tests (with `run.sh`) with bad TS type declarations, nothing fails...

Maybe because optional arguments are automatically coerced into `undefined`? For example, [this](https://www.typescriptlang.org/play/#src=let%20opt_fn%3A%20(_a%3F%3A%20number)%20%3D%3E%20number%20%7C%20undefined%20%3D%20function%20opt_fn(a)%20%7B%0D%0A%20%20%20%20return%20a%3B%0D%0A%7D%0D%0A%0D%0Aconst%20fn%3A%20(_a%3F%3A%20number)%20%3D%3E%20number%20%7C%20undefined%20%3D%20opt_fn%3B%0D%0A%0D%0Aconsole.log(fn())%3B) and [this](https://www.typescriptlang.org/play/#src=let%20opt_fn%3A%20(_a%3F%3A%20number)%20%3D%3E%20number%20%7C%20undefined%20%3D%20function%20opt_fn(a)%20%7B%0D%0A%20%20%20%20return%20a%3B%0D%0A%7D%0D%0A%0D%0Aconst%20fn%3A%20(_a%3A%20number%20%7C%20undefined)%20%3D%3E%20number%20%7C%20undefined%20%3D%20opt_fn%3B%0D%0A%0D%0Aconsole.log(fn(undefined))%3B) are identical.

Maybe with `ts-node` we could test things like this at run-time.

crates/cli-support/src/js/js2rust.rs
```diff
@@ -790,7 +810,11 @@ impl<'a, 'b> Js2Rust<'a, 'b> {
         let ts_args = self
             .js_arguments
             .iter()
-            .map(|s| format!("{}: {}", s.0, s.1))
+            .map(|s| if s.optional {
+                format!("{}?: {}", s.name, s.type_)
+            } else {
+                format!("{}: {}", s.name, s.type_)
+            })
```
rhysd: I considered that optional parameter is available or not. But my conclusion was 'available but not easy'. `foo: T | undefined` and `foo?: T` are not the same in terms of the point that the parameter can be omitted.

For example, TypeScript does not allow following function.

```typescript
function foo(a?: number, b: boolean) {}
```

Here is an error output:

```
foo.ts:1:26 - error TS1016: A required parameter cannot follow an optional parameter.

1 function foo(a?: number, b: boolean) {}
                           ~


Found 1 error.
```

So the parameter which can be optional is 'a parameter which is not followed by required parameter'.

For this reason, I used `T | undefined` always.

dbrgn: Oh, that is a very good point... I guess it's complicated :confused: 

One possibility would be to add a wasm-bindgen wrapper type called `Omittable` which results in an optional field (and an error when a non-omittable field follows an omittable field). I'm not sure whether I like that though.

Another option would be to add a function attribute called `#[optional_omittable]` or something like that. If set, then all optional fields that are not followed by non-optional fields would be marked as omittable (using `?`).

rhysd: I wrote 'not easy' but if we only use optional parameter for parameters not followed by required parameters, I think following function would be enough.

```rust
let mut omittable = true;
let mut ts_args = vec![];
for arg in self.js_arguments.iter().rev() {
    if arg.optional && omittable {
        if omittable {
            ts_args.push(format!("{}?: {}", name, type_)
        } else {
            ts_args.push(format!("{}: {} | undefined", name, type_)
        }
    } else {
        omittable = false;
        ts_args.push(format!("{}: {}", name, type_)
    }
}
ts_args.reverse();
```

alexcrichton:
Thanks for this @dbrgn! In addition to what @rhysd is mentioning, could the test in in `crates/typescript-tests` be updated with various flavors of optional arguments to make sure this works? (e.g. some methods, some cases about mixed optional arguments, etc)

crates/cli-support/src/js/js2rust.rs
dbrgn: Thank you @rhysd, I integrated this logic into the PR!

dbrgn:
PR updated!

@alexcrichton I added some tests, but note that the tests are a bit flaky with regards to ensuring correct output for optional and/or omittable parameters. I found several combinations that pass the test but are not actually correct. For example, this works:

```ts
// typescript_tests.d.ts
export function opt_fn_mixed(_a: number | undefined, _b: number, _c?: number): number | undefined;

// opt_args_and_ret.ts
const opt_fn_mixed: (a: number, b: number, c: number) => number | undefined = wbg.opt_fn_mixed;
```

The reason is that `number` is a type subset of `number | undefined`, and narrowing the set of allowed types is always possible.

I think a test function or macro that asserts the correctly generated TS output string/file would be the better approach than using TypeScript's type system for verification. But that's out of scope for this PR.

crates/cli-support/src/js/js2rust.rs
dbrgn: Are there any guidelines / recommendations on this topic? I couldn't find anything in https://github.com/rust-dev-tools/fmt-rfcs/.

I personally prefer the shorter version, it's clear from the signature that it's a constructor.

alexcrichton:
Ok no worries, this looks good to me! Sorry for the delay in reviewing :(

I'm fine testing more in further PRs as well, this certainly looks good enough for now! @dbrgn want to open a follow-up issue with your thoughts on testing?

PR #1487, opened 2019-04-26
Added tests for SharedArrayBuffer

ibaryshnikov:
Changes:
- Added tests for `SharedArrayBuffer`
- Fixed wrong usage of `#[test]` in `ArrayBuffer` tests

### DIFF
--- a/crates/js-sys/tests/wasm/SharedArrayBuffer.js
+++ b/crates/js-sys/tests/wasm/SharedArrayBuffer.js
@@ -1,0 +1,3 @@
+exports.is_shared_array_buffer_supported = function () {
+    return typeof SharedArrayBuffer === 'function';
+};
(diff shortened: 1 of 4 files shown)


crates/js-sys/tests/wasm/SharedArrayBuffer.js
```diff
@@ -0,0 +1,3 @@
+exports.is_shared_array_buffer_supported = function () {
```
ibaryshnikov: @alexcrichton There is `exports` here. Does it mean that the tests in `wasm` folder run only for `node.js`?

alexcrichton: Indeed yeah these tests only run in node.js

alexcrichton:
Thanks @ibaryshnikov!

PR #1490, opened 2019-04-26
Recommend to use dyn_ref on funcs and strings

RReverser:
Now that functions and strings work properly with custom typechecks, these custom methods seem obsolete, so I'd recommend to use standard dyn_ref instead.


alexcrichton:
👍

Issue #1460, opened 2019-04-15
JsString::from_char_code1 and family should take u16 not u32


RReverser:
`String.prototype.charCodeAt` can accept only UTF-16 code units which translate to `u16`, but current Rust bindings take `u32` instead.

This would be a breaking change.

PR #1491, opened 2019-04-26
Add missing variadic bindings for string methods

RReverser:
- `JsString::from_code_point` - allows to create JS strings using slice of codes in WASM memory.
 - `JsString::from_char_code` - same as above, but also uses updated signature with `u16` instead of `u32` (partially helps with #1460 at least for the new binding).

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3400,6 +3400,14 @@ extern "C" {
     ///
     /// There are a few bindings to `from_char_code` in `js-sys`: `from_char_code1`, `from_char_code2`, etc...
     /// with different arities.
+    ///
+    /// Additionally, this function accepts `u16` for character codes, but
+    /// fixing others requires a breaking change release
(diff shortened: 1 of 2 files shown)


RReverser:
```
127 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::traits::IntoWasmAbi` is not implemented for `&[wasm_bindgen::JsValue]`
```

Ah hmm... I should have checked these.

Is there any particular reason these slices are not allowed? (while `Box<[JsValue]>` is)

RReverser:
I've added support for `[JsValue]` slices since it seemed easy to do so (resulting in removing more code than added), but let me know if that's not desired for some reason.

RReverser:
Oh, hmm, currently bindings assume that they can drop every item of the slice. I see, it's not as easy as seemed initially.

RReverser:
Removed the problematic bindings for now, and kept just `String::fromCharCode` and `String::fromCodePoint` instead.

alexcrichton:
👍

PR #1492, opened 2019-04-26
Add bindings for `new Function(args, body)`

RReverser:
We don't support variadic args in front, but, luckily for us, `new Function` accepts comma-separated args as a single string as well (see updated JSON test for an example).


alexcrichton:
👍

PR #1493, opened 2019-04-26
Simplify JS iteration bindings

RReverser:
This change aligns handling of JS iterators closer to Rust iterators - just like in Rust, for most of them, calling `.next()` after the iteration has ended will automatically keep returning `{ done: true, value: undefined }` aka `None`, but, just like in Rust, this is not always guaranteed, so if the user needs this to be enforced, they can use `iter.fuse()` manually.

This removes separate `IterState`, simplifying bookkeeping and somewhat reducing the generated code.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1247,6 +1247,19 @@ impl Iterator {
 
         it.next().is_function()
     }
+
+    fn next_value(&self) -> Option<Result<JsValue, JsValue>> {
+        match self.next() {
+            Ok(next) => {
+                if next.done() {
+                    None
+                } else {
+                    Some(Ok(next.value()))
+                }
+            }
+            Err(e) => Some(Err(e)),
+        }
+    }
 }
 
 /// An iterator over the JS `Symbol.iterator` iteration protocol.
@@ -1254,7 +1267,6 @@ impl Iterator {
 /// Use the `IntoIterator for &js_sys::Iterator` implementation to create this.
 pub struct Iter<'a> {
     js: &'a Iterator,
-    state: IterState,
 }
 
 /// An iterator over the JS `Symbol.iterator` iteration protocol.
@@ -1262,11 +1274,6 @@ pub struct Iter<'a> {
 /// Use the `IntoIterator for js_sys::Iterator` implementation to create this.
 pub struct IntoIter {
     js: Iterator,
-    state: IterState,
-}
-
-struct IterState {
-    done: bool,
 }
 
 impl<'a> IntoIterator for &'a Iterator {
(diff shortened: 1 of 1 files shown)


alexcrichton:
@fitzgen I think you were the original proponent of the extra `bool`, so mind reviewing this?

fitzgen:
I think the fusing is important because if the JS iterator keeps returning `Err(..)`s then calling `iter.collect()` will infinite loop.

RReverser:
> if the JS iterator keeps returning Err(..)

This is not something a generator or any other well-behaved iterable can do, they "fuse" after an error automatically:

```js
> function *f() { yield 1; throw new Error(2) }
undefined
> i = f()
Object [Generator] {}
> i.next()
{ value: 1, done: false }
> i.next()
Thrown:
Error: 2
    at f (repl:1:32)
    at f.next (<anonymous>)
> i.next()
{ value: undefined, done: true }
> i.next()
{ value: undefined, done: true }
> ...
```

If you mean a custom-crafted iterable that intentionally does it, then yeah, it's possible, but then, it's possible to dynamically violate many other constraints that wasm-bindgen assumes.

fitzgen:
Hm ok, I suppose that is pretty convincing.

Want to add a warning to the `js_sys::Iterator` doc comment that this is not fused and that JS which keeps throwing errors will cause `.collect()` to infinite loop unless the iterator has `.fuse()` called on it? Happy with this change if we add that.

RReverser:
> and that JS which keeps throwing errors will cause .collect() to infinite loop unless the iterator has .fuse() called on it

Just to be sure, could you please clarify which particular example you have in mind?

fitzgen:
I was thinking either https://docs.rs/js-sys/0.3.20/js_sys/struct.Iterator.html and/or https://docs.rs/js-sys/0.3.20/js_sys/struct.Iter.html docs.

RReverser:
Oh no, sorry, I mean which JS example you had in mind by that.

fitzgen:
Oh I don't have a particular snippet in mind -- I don't even think a JS snippet is necessary to include. Just important to mention that this pathological case is possible, and how to deal with it.

RReverser:
I just wanted to confirm, because, if you're talking about that specially crafted iterable that can throw infinitely, then `.fuse()` on its own won't help - it helps only with emitting more values past the end of the iteration.

On the other hand, this is pretty hard to get wrong, and I can just include an example of the right usage?

fitzgen:
> I just wanted to confirm, because, if you're talking about that specially crafted iterable that can throw infinitely, then `.fuse()` on its own won't help - it helps only with emitting more values past the end of the iteration.

Ah yes, I'm brain farting today.

The only reason I'm a curmudgeon for this is because we hit `collect()` et al not working right for iterators of `Result`s in `gimli` pretty badly, and this is what motivated our switch over to the `fallible-iterator` crate.

That it is difficult to create in JS is definitely alleviating. All I want is docs that say something to the effect of

> If the JS iterable keeps throwing exceptions indefinitely, then `collect` will infinite loop. You can use the `fallible-iterator` crate or `take_until(|r| r.is_err())` to avoid this pathological case.

alexcrichton:
I'm going to close this since it's a very old PR at this point, but feel free to resubmit!

Issue #1501, opened 2019-05-01
Closures Example is not working.


sahil-bansal-sb:
link:
[https://rustwasm.github.io/wasm-bindgen/exbuild/closures/](https://rustwasm.github.io/wasm-bindgen/exbuild/closures/) 
is not working.

It is showing the following error on console
```
RuntimeError: unreachable
    at __rust_start_panic (wasm-function[673]:1)
    at rust_panic (wasm-function[594]:31)
    at std::panicking::rust_panic_with_hook::h5d8808384a53a826 (wasm-function[130]:304)
    at std::panicking::continue_panic_fmt::h821bed92a14cf5d5 (wasm-function[324]:116)
    at rust_begin_unwind (wasm-function[664]:3)
    at core::panicking::panic_fmt::hfa7141b9630aa10b (wasm-function[487]:70)
    at core::option::expect_failed::he814056c2f8215e8 (wasm-function[396]:91)
    at <core::option::Option<T>>::expect::h33643d65ce465d73 (wasm-function[161]:170)
    at closures::setup_clock::hcd8ed9afea8a32fa (wasm-function[44]:115)
    at closures::run::h0707c66cb6048ff9 (wasm-function[31]:1014)
```
and a warning:
```
[Deprecation] Element.createShadowRoot is deprecated and will be removed in M73, around March 2019. Please use Element.attachShadow instead. See https://www.chromestatus.com/features/4507242028072960 for more details.
```

alexcrichton:
cc @xtuc when trying to debug this with `--target web` it looks like the error goes away, so I'm curious if you know if this was an accidental regression in webpack?

xtuc:
I'm having an issue building the example locally, when running `wasm-pack`:
```
thread 'main' panicked at 'index 114 out of range for slice of length 39', src/libcore/slice/mod.rs:2413:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::_print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::slice::slice_index_len_fail
   9: <&'src str as wasm_bindgen_cli_support::decode::Decode<'src>>::decode
[...]

PR #1502, opened 2019-05-01
Update webpack.config.js

xtuc:
Fixes https://github.com/rustwasm/wasm-bindgen/issues/1501

The `#loading` element is not found in the page, which is triggering the `expect()`, the panic is not clear tho.

I didn't check the other example.


alexcrichton:
Oh dear how embarassing! Thanks for the fix @xtuc!

PR #1507, opened 2019-05-02
Adding in Futures 0.3 support to wasm-bindgen-futures

Pauan:
I haven't tested this yet, but it probably works.

I put everything behind the `nightly` feature. I'm open to changing the name.

Unfortunately I can't really add unit tests, because the `wasm-bindgen-test` crate requires Futures 0.1


Pauan:
I tried to test this, but I keep running into https://github.com/rustwasm/wasm-bindgen/issues/1501#issuecomment-488313543 and I wasn't able to compile any of the examples.

crates/futures/src/nightly.rs
```diff
@@ -0,0 +1,252 @@
...
+///
+/// # Panics
+///
+/// This function has the same panic behavior as `future_to_promise`.
+pub fn spawn_local<F>(future: F)
+where
+    F: Future<Output = ()> + 'static,
+{
+    struct Task {
+        future: Mutex<Option<Pin<Box<dyn Future<Output = ()> + 'static>>>>,
```
alexcrichton: Since the `Future` here is fundamentally non-`Send`, is the `Mutex`/`Atomic*` required or could they be replaced with non-threadsafe variants?

crates/futures/src/nightly.rs
```diff
@@ -0,0 +1,252 @@
...
+                future: Mutex::new(Some(Box::pin(future))),
+                is_queued: AtomicBool::new(false),
+            })
+        }
+    }
+
+    impl ArcWake for Task {
+        fn wake_by_ref(arc_self: &Arc<Self>) {
+            // TODO can this be more relaxed ?
+            if !arc_self.is_queued.swap(true, Ordering::SeqCst) {
```
alexcrichton: Could this be swapped for `if swap(true) { return }` to avoid indentation where we can?

crates/futures/src/nightly.rs
```diff
@@ -0,0 +1,252 @@
...
+            Arc::new(Self {
+                future: Mutex::new(Some(Box::pin(future))),
+                is_queued: AtomicBool::new(false),
+            })
+        }
+    }
+
+    impl ArcWake for Task {
+        fn wake_by_ref(arc_self: &Arc<Self>) {
+            // TODO can this be more relaxed ?
```
alexcrichton: Technically the ordering doesn't matter at all today in the sense that all orderings are equivalent when you're dealing with one thread of code. For multithreaded code wasm has no concept of orderings, so we'll want to stick `SeqCst` everywhere

crates/futures/src/nightly.rs
alexcrichton: Also, orthogonally, I wasn't quite sure why I initially saw `Option` here, but is the idea that you can drop the future at any time but there may still be lingering handles to the `Task` for awhile longer while events are processed throughout the system?

crates/futures/src/lib.rs
```diff
@@ -103,300 +103,14 @@
...
-/// `futures` crate and will either succeed or fail depending on what happens
-/// with the JavaScript `Promise`.
-///
-/// Currently this type is constructed with `JsFuture::from`.
-pub struct JsFuture {
-    resolved: oneshot::Receiver<JsValue>,
-    rejected: oneshot::Receiver<JsValue>,
-    callbacks: Option<(Closure<FnMut(JsValue)>, Closure<FnMut(JsValue)>)>,
-}
+#[cfg(not(feature = "nightly"))]
```
alexcrichton: I think that we'll probably want to have the standard library futures support in its own module. We don't want enabling the `nightly` feature to break this crate (features should be additive in terms of functionality) for existing users. That does mean that if `nightly` is enabled it'll pull in two implementations of futures, but I think that's ok since the linker will eliminate them if they're not used.

crates/futures/src/nightly.rs
Pauan: The atomic stuff isn't needed, it's just a holdover from when I tried to make it thread safe (and then later gave up).

The reason for the `Option` is because the Future might return `Poll::Ready` and also call `waker.wake()`. That would end up causing an infinite loop, so the `Option` lets us avoid that.

Sure, that's technically against the Future contract, but I like to make the code rock solid so it can survive even bad Futures.

It also has the benefit that it can drop the Future immediately when it returns `Poll::Ready` (rather than at some potentially far-future time when the `Task` is eventually dropped).

Because the `Waker` contains an `Arc<Task>`, it's theoretically possible that the Future will *never* get dropped (if someplace somewhere keeps a hold of the `Waker`). So the `Option` lets us fix that.

crates/futures/src/nightly.rs
Pauan: That's good to know about multi-threaded wasm.

crates/futures/src/lib.rs
Pauan: I'm okay with that.

Pauan:
Okay, I addressed all the feedback.

alexcrichton:
👍 nice!

PR #1508, opened 2019-05-02
Add `bufferSubData` and `clearBuffer[fiuv]` to immutable slice whitelist

RSSchermer:
Adds `bufferSubData` and `clearBuffer[fiuv]` to immutable slice whitelist.

Edit: `bufferData` -> `bufferSubData` (typo)

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -189,6 +189,7 @@ fn immutable_slice_whitelist() -> BTreeSet<&'static str> {
         "vertexAttrib3fv",
         "vertexAttrib4fv",
         "bufferData",
+        "bufferSubData",
         "texImage2D",
         "texSubImage2D",
         "compressedTexImage2D",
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I'm personally pretty unfamiliar with WebGL, but @chinedufn perhaps you'd be willing to take a look to verify this is a correct transformation to apply?

RSSchermer:
Here are some links the the relevant MDN documentation to perhaps help verify that these function indeed don't require and exclusive reference:

- [bufferSubData](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/bufferSubData)
- [clearBuffer](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext/clearBuffer)

alexcrichton:
Ok yeah sounds like @chinedufn is busy right now so looks good to me!

PR #1512, opened 2019-05-03
Add a page of microbenchmarks for wasm-bindgen

alexcrichton:
This commit starts to add a page of microbenchmarks for wasm-bindgen
which we can hopefully track and compare over time. Right now it's
primarily focused on data collection, making it easy to collect data
across a number of benchmarks for comparison. It doesn't currently do
much in the way of actually comparing the results for you (aka drawing
pretty graphs), so let's left for a future step.

It's hoped though that we can use this to track performance improvements
as well as ensuring that they work over time!


alexcrichton:
This isn't deploying correctly and I'm still trying to figure that out... will report back when it's hopefully working.

alexcrichton:
Ok [now live](https://rustwasm.github.io/wasm-bindgen/benchmarks/)!

PR #1514, opened 2019-05-04
Threadsafe futures

### LINKED ISSUE #1379: Implement a threadsafe version of wasm-bindgen-futures
Currently the wasm-bindgen-futures crate provides a number of `unsafe` implementations to implement a runtime to execute futures within. These `unsafe` blocks, however, are predicated on the application being single threaded. The crate isn't safe in a threaded environment!

We should implement a threadsafe version of `wasm-bindgen-futures`, gated on the `target_feature = "atomics"` feature of WebAssembly!

ibaryshnikov:
Closes #1379 

Changes:
- Added extern function `wait_async`, which does the same as `Atomics::waitAsync`
- Removed `impl Notify for Package`, also removed `Send + Sync` markers for `Package`
- Added `Waker` struct instead, which implements `Notify`
- Used `Waker` instead of `Package` to notify the executor


crates/js-sys/src/lib.rs
```diff
@@ -478,6 +478,9 @@ extern "C" {
     pub fn slice_with_end(this: &SharedArrayBuffer, begin: u32, end: u32) -> SharedArrayBuffer;
 }
 
+unsafe impl Send for SharedArrayBuffer {}
```
ibaryshnikov: @alexcrichton is it fine to mark `SharedArrayBuffer` as `Send + Sync` manually? I tried to remove `#[wasm_bindgen(extends = Object)]`, but the compiler still thought that `SharedArrayBuffer` had a `PhantomData` from `JsValue`

crates/futures/src/lib.rs
```diff
@@ -339,47 +352,48 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
...
-                let myself = slot2.borrow_mut().take();
-                debug_assert!(myself.is_some());
-                Package::poll(&me);
-            }) as Box<FnMut(JsValue)>);
-            promise.then(&closure);
-            *slot.borrow_mut() = Some(closure);
-        }
-    }
+    //    impl Notify for Package {
+    //        fn notify(&self, _id: usize) {
```
ibaryshnikov: @alexcrichton I added a `Waker` struct with a simple `Notify` implementation, but I don't know yet where to put all the logic from `impl Notify for Package`. Can you suggest a good place for it?

crates/js-sys/src/lib.rs
alexcrichton: Unfortunately this can't be done because the `JsValue` underlying can't actually be sent between threads (inherently). The `SharedArrayBuffer` value that you'll be notifying through will need to be `wasm_bindgen::memory()` I believe

crates/futures/src/polyfill.js
```diff
@@ -0,0 +1,93 @@
+const helperCode = `
```
alexcrichton: Could this file have a comment about where it comes from?

ibaryshnikov: Oh, thanks!

crates/js-sys/src/lib.rs
ibaryshnikov: It's an interesting point. So I can remove `unsafe impl` from here?
I have some more questions about implementation.
- WebAssembly memory could be just an ArrayBuffer, not the shared one. Is it fine?
- In docs there is `pub fn memory() -> JsValue`, which refers to the whole linear memory. Do I need to use `std::alloc` here to obtain a proper memory location?

ibaryshnikov:
Does it mean that the state with `Polling | Notified | Waiting` will be removed, and there will be only `true` or `false`?

alexcrichton:
Oh what I gisted was just a straw man, it may be that two bits are still needed rather than one for information exchange.

ibaryshnikov:
@alexcrichton the pr is not tested yet, I pushed changes to check the design.
There are several places to check: first, I removed `impl Notify for Package`. Then, instead of `&me`, there is `&me.waker` passed to `poll_future_notify`.
And I moved code from `impl Notify for Package` to `fn poll_again`
Inside `fn poll_again` `Promise::resolve` was replaced with `wait_async`
Did I get your idea right?

crates/futures/src/lib.rs
```diff
@@ -289,10 +296,68 @@ fn _future_to_promise(future: Box<Future<Item = JsValue, Error = JsValue>>) -> P
         Waiting(Arc<Package>),
     }
 
-    // No shared memory right now, wasm is single threaded, no need to worry
-    // about this!
-    unsafe impl Send for Package {}
-    unsafe impl Sync for Package {}
+    struct Waker {
+        buffer: SharedArrayBuffer,
```
alexcrichton: This I don't think will store `SharedArrayBuffer` but rather use `wasm_bindgen::memory().buffer()` as the `SharedArrayBuffer` accessor

ibaryshnikov: oh, fine! but how do I allocate it?

ibaryshnikov:
@alexcrichton I think I hit the wall here: https://github.com/rustwasm/wasm-bindgen/issues/1525
There's a need to either make js snippets work with node target or replace the snippet with polyfill with the wasm equivalent. What do you think is better?

alexcrichton:
Yeah it may be some time before snippets work on Node, so it's probably best to not block this on that. Perhaps the polyfill could be implemented in Rust though?

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,209 @@
...
+        console_log!("polyfill, not-equal");
+        return Ok(Promise::resolve(&JsString::from("not-equal")));
+    }
+
+    // General case, we must wait.
+
+    console_log!("polyfill, general case");
+
+    Ok(Promise::new(
+        &mut Box::new(move |resolve: Function, reject: Function| {
```
alexcrichton: I think this box may be able to be removed?

crates/futures/src/polyfill.js
```diff
@@ -0,0 +1,130 @@
+/*
+ * The polyfill was kindly borrowed from https://github.com/tc39/proposal-atomics-wait-async
```
alexcrichton: I think this file may no longer be necessary?

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,209 @@
...
+                    }
+                    // it's not specified in the proposal yet
+                    _ => (),
+                }
+            }) as Box<dyn FnMut(MessageEvent)>);
+            helper
+                .borrow()
+                .set_onmessage(Some(onmessage_callback.as_ref().unchecked_ref()));
+
+            onmessage_callback.forget();
```
alexcrichton: I think we'll want to avoid `forget` here since it'll leak memory quickly, perhaps the `Closure::once` constructor could be used?

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+    } catch (e) {
+        console.log('Exception in wait helper');
+        postMessage(['error', 'Exception']);
+    }
+}
+";
+
+thread_local! {
+    static HELPERS: RefCell<Vec<Rc<RefCell<Worker>>>> = RefCell::new(vec![]);
```
alexcrichton: Is the `RefCell` inside the `Rc` needed here to mutate the `Worker`?

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+fn free_helper(helper: &Rc<RefCell<Worker>>) {
+    HELPERS.with(move |helpers| {
+        helpers.borrow_mut().push(helper.clone());
+    });
+}
+
+pub fn wait_async(indexed_array: Int32Array, index: u32, value: i32) -> Result<Promise, JsValue> {
+    let timeout = 0.0;
```
alexcrichton: I think this default timeout may be causing a spin loop?

crates/js-sys/src/lib.rs
alexcrichton: Oh sorry forgot to follow-up but yes it should be safe to remove these.

examples/raytrace-parallel/Cargo.toml
```diff
@@ -13,7 +13,7 @@ futures = "0.1"
 js-sys = "0.3.22"
 raytracer = { git = 'https://github.com/alexcrichton/raytracer', branch = 'update-deps' }
 wasm-bindgen = { version = "0.2.45", features = ['serde-serialize'] }
-wasm-bindgen-futures = "0.3.22"
+wasm-bindgen-futures = { path = '../../crates/futures', version = "0.3.22" }
```
alexcrichton: If this `path` was necessary, mind avoiding this change and instead updating `/Cargo.toml` and the `[patch]` section?

crates/futures/src/lib.rs
```diff
@@ -110,14 +110,35 @@ pub mod futures_0_3;
...
 
 use futures::executor::{self, Notify, Spawn};
 use futures::future;
 use futures::prelude::*;
 use futures::sync::oneshot;
 use js_sys::{Function, Promise};
+#[cfg(target_feature = "atomics")]
+use js_sys::{Atomics, Int32Array, SharedArrayBuffer, WebAssembly};
 use wasm_bindgen::prelude::*;
+#[cfg(target_feature = "atomics")]
```
alexcrichton: Organizationally could there perhaps be:

```
src/stable.rs // what lib.rs currently is
src/atomics.rs // what this file is currently
src/futures03.rs // implementation with futures 0.3
```

and eventually we can add `src/futures03_atomics.rs`, but for now we'd just go with what's here. I don't think we'll benefit too too much from shared types, but the shared pieces can always be in `src/lib.rs`

crates/futures/src/lib.rs
```diff
@@ -110,14 +110,35 @@ pub mod futures_0_3;
...
+mod polyfill;
+
+macro_rules! console_log {
+    ($($t:tt)*) => (log(&format_args!($($t)*).to_string()))
+}
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = console)]
+    fn log(s: &str);
```
alexcrichton: Just a reminder that we'll want to remove these before landing.

crates/futures/src/lib.rs
```diff
@@ -289,9 +316,108 @@ fn _future_to_promise(future: Box<dyn Future<Item = JsValue, Error = JsValue>>)
...
+
+                let _ = Atomics::notify(&array, id as u32);
+            }
+        }
+    }
+
+    #[cfg(target_feature = "atomics")]
+    fn poll_again(package: Arc<Package>, id: usize) {
+        console_log!("poll_again called");
+        let me = match package.notified.replace(State::Notified) {
```
alexcrichton: This `notified` check will want to be the same as the `notified` field in `Waker` to ensure all the races align well

crates/futures/src/lib.rs
```diff
@@ -289,9 +316,108 @@ fn _future_to_promise(future: Box<dyn Future<Item = JsValue, Error = JsValue>>)
...
+
+        // Use `Promise.then` on a resolved promise to place our execution
+        // onto the next turn of the microtask queue, enqueueing our poll
+        // operation. We don't currently poll immediately as it turns out
+        // `futures` crate adapters aren't compatible with it and it also
+        // helps avoid blowing the stack by accident.
+        //
+        // Note that the `Rc`/`RefCell` trick here is basically to just
+        // ensure that our `Closure` gets cleaned up appropriately.
+        let promise = polyfill::wait_async(array, id as u32, 0)
```
alexcrichton: I don't think that this is quite right because `id` is arbitrarily passed in. I think it's safe to effectively ignore the `id` in the trait method and everywhere else, and this should be a `waitAsync` on the memory location of `notified` in the `Waker` (which can be switched to `AtomicI32` to ensure it's a 32-bit integer)

crates/futures/src/lib.rs
```diff
@@ -289,9 +316,108 @@ fn _future_to_promise(future: Box<dyn Future<Item = JsValue, Error = JsValue>>)
...
+        };
+
+        let memory_buffer = wasm_bindgen::memory()
+            .dyn_into::<WebAssembly::Memory>()
+            .expect("Should cast a memory to WebAssembly::Memory")
+            .buffer();
+
+        let array_location = package.waker.array.as_ptr() as u32 / 4;
+        let array = Int32Array::new(&memory_buffer)
+            .subarray(array_location, array_location + package.waker.array.len() as u32);
```
alexcrichton: I think it's safe to avoid the `subarray` here and just use the `Int32Array` with a larger offset that is the pointer that we're waiting on.

crates/futures/src/lib.rs
```diff
@@ -289,9 +316,108 @@ fn _future_to_promise(future: Box<dyn Future<Item = JsValue, Error = JsValue>>)
...
+        // onto the next turn of the microtask queue, enqueueing our poll
+        // operation. We don't currently poll immediately as it turns out
+        // `futures` crate adapters aren't compatible with it and it also
+        // helps avoid blowing the stack by accident.
+        //
+        // Note that the `Rc`/`RefCell` trick here is basically to just
+        // ensure that our `Closure` gets cleaned up appropriately.
+        let promise = polyfill::wait_async(array, id as u32, 0)
+            .expect("Should create a Promise");
+        let slot = Rc::new(RefCell::new(None));
```
alexcrichton: This dance may be a bit easier today with `Closure::once` perhaps?

crates/futures/src/lib.rs
```diff
@@ -289,9 +316,108 @@ fn _future_to_promise(future: Box<dyn Future<Item = JsValue, Error = JsValue>>)
...
+                let memory_buffer = wasm_bindgen::memory()
+                .dyn_into::<WebAssembly::Memory>()
+                .expect("Should cast a memory to WebAssembly::Memory")
+                .buffer();
+
+                let array_location = self.array.as_ptr() as u32 / 4;
+                let array = Int32Array::new(&memory_buffer)
+                    .subarray(array_location, array_location + self.array.len() as u32);
+
+                let _ = Atomics::notify(&array, id as u32);
```
alexcrichton: I think this can be the [notify intrinsic in Rust](https://doc.rust-lang.org/core/arch/wasm32/fn.atomic_notify.html) perhaps?

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,209 @@
...
+    })
+}
+
+fn free_helper(helper: &Rc<RefCell<Worker>>) {
+    HELPERS.with(move |helpers| {
+        helpers.borrow_mut().push(helper.clone());
+    });
+}
+
+pub fn wait_async(indexed_array: Int32Array, index: u32, value: i32) -> Result<Promise, JsValue> {
```
alexcrichton: FWIW we can probably simplify this polyfill if it's written in Rust to simply take a pointer here instead of an `Int32Array` plus an `index`

crates/futures/src/polyfill.rs
ibaryshnikov: `Closure::once` still produces `Error: closure invoked recursively or destroyed already` if I remove `forget`. May be `Closure::once_into_js` instead?

crates/futures/src/lib.rs
ibaryshnikov: I moved current implementation to the `atomics.rs` file. Do I need to go farther and move contents of `lib.rs` to `stable.rs`?

crates/futures/src/lib.rs
ibaryshnikov: I don't get it yet. Adding the following check after the `match` statement fixes performance problems, but breaks the whole thing
```rust
if package.waker.notified.swap(true, Ordering::SeqCst) {
    return;
}
```

crates/futures/src/polyfill.rs
ibaryshnikov: So I can just remove `indexed_array` here, and our index will be the pointer? Since pointers in wasm32 are simply u32 numbers

crates/futures/src/atomics.rs
```diff
@@ -0,0 +1,351 @@
...
+            }
+        }
+    }
+
+    impl Notify for Waker {
+        fn notify(&self, id: usize) {
+            console_log!("Waker notify");
+            if !self.notified.swap(true, Ordering::SeqCst) {
+                console_log!("Waker, inside if");
+                let _ = unsafe { core::arch::wasm32::atomic_notify(&self.value as *const AtomicI32 as *mut i32, 0) };
```
ibaryshnikov: @alexcrichton I have hard time trying to figure out how to use intrinsics here, may be you know a better way?

crates/futures/src/atomics.rs
```diff
@@ -0,0 +1,351 @@
...
+        // Note that the self-reference here is an Arc-cycle that will leak
+        // memory unless the future completes, but currently that should be ok
+        // as we'll have to stick around anyway while the future is executing!
+        //
+        // This state is removed as soon as a notification comes in, so the leak
+        // should only be "temporary"
+        Waiting(Arc<Package>),
+    }
+
+    struct Waker {
```
ibaryshnikov: @alexcrichton so we can use only one field here, `notified`, and make it `AtomicI32`?

crates/futures/src/lib.rs
```diff
@@ -101,12 +101,22 @@
 //! }
 //! ```
 
+#![feature(stdsimd)]
```
ibaryshnikov: @alexcrichton it's not correct to leave it here, but when I put it inside `atomics.rs`, the example fails to compile, with error mentioning this feature

crates/futures/src/polyfill.rs
ibaryshnikov: Wow, changing default timeout to 10.0 helped! In spec it's a number of milliseconds. But I'm not sure which number to chose as default. What do you think?

crates/futures/src/atomics.rs
```diff
@@ -0,0 +1,350 @@
...
+            // continue polling. For us, though, there's nothing else to do,
+            // so we bail out.
+            // later see
+            State::Polling => {
+                console_log!("poll_again Polling");
+                return;
+            }
+        };
+
+        let value_location = &package.waker.value as *const AtomicI32 as u32 / 4;
```
ibaryshnikov: @alexcrichton here is another place where I'd like to simplify things, is there a better way of obtaining a pointer as u32 number?

ibaryshnikov:
@alexcrichton thank you for a great review! I fixed most of the places, but still have some questions, like usage of `AtomicI32`, or simplifying `Waker` struct. Btw, it looks working, and after increasing default timeout from **0** to 1**0ms** the performance looks fine. So it's probably a time to start making changes to raytrace_parallel example itself?

crates/futures/src/lib.rs
alexcrichton: Oh no this is correct yeah since the atomic intrinsics aren't stable yet, but you'll want to put this behind a `cfg_attr` to only compile it in when the `atomics` feature is enabled. That way this can continue to compile on stable.

crates/futures/src/lib.rs
```diff
@@ -101,12 +101,21 @@
...
 #![deny(missing_docs)]
 
 #[cfg(feature = "futures_0_3")]
```
alexcrichton: While you're at it it's probably good to remove this if `"atomics"` is enabled (since it's not compatible).

The `cfg-if` crate may be able to help clean up the `#[cfg]` here a bit too.

crates/futures/src/lib.rs
```diff
@@ -101,12 +101,21 @@
...
 
+#[cfg(target_feature = "atomics")]
+/// Contains a thread-safe version of this crate, with Futures 0.1
+pub mod atomics;
+
+#[cfg(target_feature = "atomics")]
+/// Polyfill for `Atomics.waitAsync` function
+mod polyfill;
+
 use std::cell::{Cell, RefCell};
```
alexcrichton: Oh so actually the existing implementation in this file should be removed when `atomics` is enabled. With atomics the current implementation is unsound, but the `atomics` module contains the same interface anyway so the idea is that we'd just swap out for different implementations depending on `atomics` or not.

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+    HELPERS.with(|helpers| {
+        if let Some(helper) = helpers.borrow_mut().pop() {
+            return helper;
+        }
+
+        let mut initialization_string = "data:application/javascript,".to_owned();
+        let encoded: String = encode_uri_component(HELPER_CODE).into();
+        initialization_string.push_str(&encoded);
+
+        return Rc::new(Worker::new(&initialization_string).expect("Should create a Worker"));
```
alexcrichton: Sorry this may have gotten lost in the previous review comments, but I think that the `Rc` here shouldn't be necessary because a `Worker` can be natively cloned?

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+    })
+}
+
+fn free_helper(helper: &Rc<Worker>) {
```
alexcrichton: This might be a bit more appropriate to take `helper` by value rather than behind a reference

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+    HELPERS.with(move |helpers| {
+        helpers.borrow_mut().push(helper.clone());
+    });
+}
+
+pub fn wait_async(index: u32, value: i32) -> Result<Promise, JsValue> {
+    wait_async_with_timeout(index, value, DEFAULT_TIMEOUT)
```
alexcrichton: I think that instead of a `DEFAULT_TIMEOUT` you can pass `-1` here which indicates "wait forever"

Or at least I think that's the intented API of  `Atomics.waitAsync`

crates/futures/src/atomics.rs
alexcrichton: I actually think it's best to simply cahnge `wait_async` to take `&AtomicI32`, and then you can do different things internally there.

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+    })
+}
+
+fn free_helper(helper: &Rc<Worker>) {
+    HELPERS.with(move |helpers| {
+        helpers.borrow_mut().push(helper.clone());
+    });
+}
+
+pub fn wait_async(index: u32, value: i32) -> Result<Promise, JsValue> {
```
alexcrichton: FWIW I'd still recommend changing `index: u32` to `ptr: &AtomicI32`

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    let indexed_array = Int32Array::new(&memory_buffer);
+
+    if !indexed_array.buffer().has_type::<SharedArrayBuffer>() {
+        console_log!("polyfill, not a SharedArrayBuffer");
+        return Err(Error::new("Indexed array must be created from SharedArrayBuffer").into());
+    }
+
+    // Optimization, avoid the helper thread in this common case.
+    if Atomics::load(&indexed_array, index)? != value {
```
alexcrichton: If `index` is changed to `&AtomicI32`, you don't need to call a JS API here and can instead simply do `ptr.load(SeqCst)`

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+// rejected with an error string.
+pub fn wait_async_with_timeout(index: u32, value: i32, timeout: f64) -> Result<Promise, JsValue> {
+    let memory_buffer = wasm_bindgen::memory()
+        .dyn_into::<WebAssembly::Memory>()
+        .expect("Should cast a memory to WebAssembly::Memory")
+        .buffer();
+
+    let indexed_array = Int32Array::new(&memory_buffer);
+
+    if !indexed_array.buffer().has_type::<SharedArrayBuffer>() {
```
alexcrichton: This I think is fine to elide since it's always going to be true :)

crates/futures/src/polyfill.rs
```diff
@@ -0,0 +1,208 @@
...
+                            .expect("Should successfully call a reject callback");
+                    }
+                    // it's not specified in the proposal yet
+                    _ => (),
+                }
+            })
+                as Box<dyn FnMut(MessageEvent)>);
+            helper.set_onmessage(Some(onmessage_callback.as_ref().unchecked_ref()));
+
+            // onmessage_callback.forget();
```
alexcrichton: The instance of `Closure` here needs to be persisted with the `Worker` to ensure it's not deallocated on the Rust side of things, but this is basically stored with `helper_ref` I think

crates/futures/src/polyfill.rs
ibaryshnikov: In the docs it says that it defaults to `Infinity`, so I just made it as default. I didn't find any info about `-1` in the docs, and desided not to use it

crates/futures/src/polyfill.rs
ibaryshnikov: Do you mean to create something like
```rust
struct WorkerData {
  worker: Worker,
  closure: Closure,
}
```
?

ibaryshnikov:
@alexcrichton I fixed most of the places, but still have some where I'm not sure, like what to do with closures, and the contents of the `Waker` struct, which is currently
```rust
struct Waker {
    value: AtomicI32,
    notified: AtomicBool,
}
```

crates/futures/src/atomics.rs
```diff
@@ -0,0 +1,326 @@
...
+        };
+
+        // Use `Promise.then` on a resolved promise to place our execution
+        // onto the next turn of the microtask queue, enqueueing our poll
+        // operation. We don't currently poll immediately as it turns out
+        // `futures` crate adapters aren't compatible with it and it also
+        // helps avoid blowing the stack by accident.
+        let promise =
+            crate::polyfill::wait_async(&package.waker.value).expect("Should create a Promise");
+        let closure = Closure::once(Box::new(move |_| {
```
alexcrichton: FWIW I think [`once_into_js`](https://docs.rs/wasm-bindgen/0.2.47/wasm_bindgen/closure/struct.Closure.html#method.once_into_js) is what you want here

ibaryshnikov: I don't get it, once_into_js returns JsValue, and promise.then expects Closure. Am I missing something?

ibaryshnikov:
Updated to the latest master (I hope I didn't miss anything). Removed `notified` field, there's only `value` now inside `Waker`.
Can we summarize, what's left to do in this pr? I'm kind of lost in the comments.

crates/futures/src/atomics.rs
alexcrichton: Ah this needed a redefinition of a few functions, but I've handled that and pushed a few commits.

alexcrichton:
Ok thanks for the update @ibaryshnikov! 

I've added a few more commits which cleaned up a few things (good chunks of the old `future_to_promise` implementation aren't necessary when using `waitAsync`) and reorganized some files a bit. I also hooked up the parallel raytrace example and can confirm it's all plumbed together and at least rudimentarily working.

If CI is green and you're good with this I think it's good to merge!

ibaryshnikov:
@alexcrichton I'm good with this. Thanks for moving this forvard! And thanks for your comments, I've learned a lot from this pr!)

alexcrichton:
Ok let's merge then, thanks so much again @ibaryshnikov!

PR #1520, opened 2019-05-09
Update init() typings

### LINKED ISSUE #1517: Typescript definitions don't include declaration for default function import.
### Describe the Bug

wasm-pack build for a simple app generates a ts file that looks like:

```ts
/* tslint:disable */
/**
* @returns {void} 
*/
export function init_panic_hook(): void;

/**
* If `module_or_path` is {RequestInfo}, makes a request and
* for everything else, calls `WebAssembly.instantiate` directly.
*
* @param {RequestInfo | BufferSource | WebAssembly.Module} module_or_path
*
* @returns {Promise<any>}
*/
export function init (module_or_path: RequestInfo | BufferSource | WebAssembly.Module): Promise<any>;
```

Unfortunately, the init function isn't labeled as the module's default export, so trying to use it that way from a .ts file:

```ts
import init, { init_panic_hook } from './my_mod';

init_panic_hook(); // TS compiler is happy with this.
init("/path"); // TS compiler hates this!
```

...will fail with a semi-cryptic `TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof import(...)' has no compatible call signatures`.

The workaround I've found for this (thanks to the Typescript gitter channel) is to append `export default init;` to the end of the generated type file. I'm not a typescript expert, though, so I'm not sure if this is the ideal solution, or if it's preferable to document that the init function should be imported in some other way? I spent a while banging my head against the wall about it though :)

Thank you!

darthtrevino:
The JavaScript output emits a 'init' function that is the default export. This PR updates the TypeScript output so that it marks init as the default export instead of a named export.

Fixes #1517
Fixes rustwasm/wasm-pack/issues/644

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -922,7 +922,7 @@ impl<'a> Context<'a> {
             *\n\
             * @returns {{Promise<any>}}\n\
             */\n\
-            export function init \
+            export default function init \
                 (module_or_path: RequestInfo | BufferSource | WebAssembly.Module{}): Promise<any>;
         ",
             memory_doc, memory_param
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #1542, opened 2019-05-17
Update dependencies and use new walrus custom sections APIs

fitzgen:


fitzgen:
Ah crap `cargo upgrade --all` removed the `path = ...` from all the cargo.tomls :(

PR #1544, opened 2019-05-19
expose forget to wasm

### LINKED ISSUE #1543: error on init (__wbindgen_cb_forget)
### Describe the Bug

In browser the wasm fails to load with:
```
LinkError: import object field '__wbindgen_cb_forget' is not a Function
```

### Steps to Reproduce

1. Update `wasm-bindgen-cli` and `wasm-pack` to latest release
1. `git clone https://github.com/derekdreery/dodrio-realworld-starter-kit`
2. `cd dodrio-realworld-starter-kit`
3. `wasm-pack build --target web`
4. `python -m http.server` (or your fave static file server)
5. go to `http://localhost:<your_port>`

### Expected Behavior

You see a web page with some stuff in it

### Actual Behavior

A blank page plus the aforementioned error message.

### Additional Context
I tried adding in the export of `__wbindgen_cb_forget` manually and it all works then! So I guess the error is to do with it not getting exported. I tried to understand the code and failed.

ibaryshnikov:
Closes #1543 

I made some research on the issue

`forget` is exposed here:
https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli-support/src/js/mod.rs#L792

Then there is a line `if contents.starts_with("function") {` here
https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli-support/src/js/mod.rs#L229

Well, it's not a class and not a function, that's why we go here:
https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli-support/src/js/mod.rs#L252

Probably, exposing to wasm every exported const is not the best idea, but it depends on the usage. If only function aliases are bound in such way, it's fine enough.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -249,7 +249,7 @@ impl<'a> Context<'a> {
                     format!("{}{}\n", export, contents)
                 } else {
                     assert_eq!(export_name, definition_name);
-                    format!("{}const {} = {};\n", export, export_name, contents)
+                    format!("{}const {name} = {};\n__exports.{name} = {name};", export, contents, name = export_name)
                 }
             }
         };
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks great to me, thanks @ibaryshnikov!

PR #1545, opened 2019-05-19
Work around rust-lang/rust#58516

infinity0:
Actually it looks like the loop could be dropped completely in the "not-strict-macro" case but I wasn't sure if used.get() might panic or something.


alexcrichton:
Hm weird! So stylistically this is definitely a step backwards, but it seems reasonable if it's working around an LLVM bug. Can you be sure to leave comments explaining the duplication though so we know what to do with this when encountering it in the future?

infinity0:
Will do. Do you want me to drop the redundant loop as well, or shall I keep it so the two versions are closer to each other?

alexcrichton:
Seems reasonable!

infinity0:
Done!

PR #1548, opened 2019-05-20
Fix closure example compilation errors

lzubiaur:
Fixing compilation errors but not sure about the `'static` lifetime though.


alexcrichton:
thanks!

PR #1553, opened 2019-05-22
Add note to the guide about `serde-wasm-bindgen`

fitzgen:


guide/src/reference/arbitrary-data-with-serde.md
```diff
@@ -105,3 +105,23 @@ example.field2.push([5,6]);
...
+
+## An Alternative Approach: `serde-wasm-bindgen`
+
+[The `serde-wasm-bindgen`
+crate](https://github.com/cloudflare/serde-wasm-bindgen) serializes and
+deserializes Rust structures directly to `JsValue`s, without going through
+temporary JSON stringification. This approach has both advantages and
+disadvantages.
+
+The primary advantage is smaller code size: going through JSON entrenches code
```
RReverser: I would probably call out support for more types in advantages too - `Map`, `Set`, `ArrayBuffer` etc. are all unsupported in the current implementation, but work in the new one.

PR #1561, opened 2019-05-29
update wasm-pack-plugin to 0.4.1

xtuc:
Improves logging and will avoid the "no ouput since 10min".

### DIFF
--- a/_package.json
+++ b/_package.json
@@ -4,7 +4,7 @@
     "serve": "webpack-dev-server -p"
   },
   "devDependencies": {
-    "@wasm-tool/wasm-pack-plugin": "0.2.7",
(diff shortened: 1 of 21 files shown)


alexcrichton:
Thanks! Is it possible that this is the cause of the TypeScript error with something being upgraded?

xtuc:
Here are the changes https://github.com/wasm-tool/wasm-pack-plugin/compare/ffa26bfa77f165780efadddc97f0b2f596e23ec8...master. I doubt this would cause the TypeScript errors.

Could you please try to re-run a maser build? Maybe some unpinned dependecies changed?

alexcrichton:
Ok, we'll need to figure that out on master then.

PR #1573, opened 2019-06-03
Add Array#flat and Array#flatMap to js-sys (fixes #1454)

### LINKED ISSUE #1454: Stage 4 JS proposal: Array.prototype.{flat,flatMap}
[This proposal](https://github.com/tc39/proposal-flatMap) got stage 4 at the [January 2019 meeting](https://github.com/tc39/tc39-notes/blob/master/es9/2019-01/jan-29.md#arrayprototypeflatflatmap-for-stage-4), and is not yet implemented in `js-sys`.

See spec text for [`flat`](https://tc39.github.io/ecma262/#sec-array.prototype.flat) and [`flatMap`](https://tc39.github.io/ecma262/#sec-array.prototype.flatmap) and MDN docs for [`flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) and [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap).

not-an-aardvark:
This adds [`Array#flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) and [`Array#flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) to `js-sys`. These JS proposals both reached stage 4 in January, as noted in https://github.com/rustwasm/wasm-bindgen/issues/1454.

This is my first time working with `wasm-bindgen`, so please review thoroughly.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -188,6 +188,23 @@ extern "C" {
     #[wasm_bindgen(method, js_name = findIndex)]
     pub fn find_index(this: &Array, predicate: &mut dyn FnMut(JsValue, u32, Array) -> bool) -> i32;
 
+    /// The flat() method creates a new array with all sub-array elements concatenated into it
+    /// recursively up to the specified depth.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)
+    #[wasm_bindgen(method)]
+    pub fn flat(this: &Array, depth: i32) -> Array;
+
+    /// The flatMap() method first maps each element using a mapping function, then flattens
+    /// the result into a new array.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)
+    #[wasm_bindgen(method, js_name = flatMap)]
+    pub fn flat_map(
+        this: &Array,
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -188,6 +188,23 @@ extern "C" {
...
+    pub fn flat(this: &Array, depth: i32) -> Array;
+
+    /// The flatMap() method first maps each element using a mapping function, then flattens
+    /// the result into a new array.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)
+    #[wasm_bindgen(method, js_name = flatMap)]
+    pub fn flat_map(
+        this: &Array,
+        callback: &mut dyn FnMut(JsValue, u32, Array) -> Vec<JsValue>,
```
not-an-aardvark: I wasn't sure of the best type to use for the callback here. In JS, it's possible for the `flatMap` callback to return a non-array value `v`, which has the same effect as returning the one-element array `[v]`. But it seems cleaner to be more strict in the Rust type.

I think `Vec<JsValue>` is the most appropriate return value here (as opposed to e.g. a slice), but it might be worth double-checking this; I don't have a lot of experience with lifetimes.

alexcrichton:
Looks great to me and I agree this is a good signature for `flat_map` for now, thanks @not-an-aardvark!

PR #1577, opened 2019-06-06
Fix getter and setter for TS

c410-f3r:
Some setters were being included without types and this PR also fix some formatting.
I modified the code based on the fact that a getter (`get foo()`) **must** have a return and a setter (`set foo(a)`) **must** have one argument, at least for TS.


alexcrichton:
Thanks!

PR #1579, opened 2019-06-07
added default module path inside init function when target is web

### LINKED ISSUE #1559: Support init function without parameters
### Motivation

When used without a bundler, there is a need to call init with a `.wasm` file name:

```javascript
await init('../pkg/my_project_bg.wasm');
```

and again

```javascript
await init('../pkg/another_project_bg.wasm');
```

### Proposed Solution

Make it possible to use
```javascript
await init();
```
which will default to path with generated `.wasm` file

ibaryshnikov:
Closes #1559

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -450,6 +450,17 @@ impl<'a> Context<'a> {
         } else {
             ""
         };
+
+        let default_module_path = match self.config.mode {
+            OutputMode::Web => {
+                "\
+                    if (typeof module === 'undefined') {
+                        module = import.meta.url.replace(/\\.js$/, '_bg.wasm');
+                    }"
+            }
+            _ => "",
+        };
+
         let ts = Self::ts_for_init_fn(mem.import.is_some());
 
         // Initialize the `imports` object for all import definitions that we're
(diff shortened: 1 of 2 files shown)


alexcrichton:
Does `import.meta` work in all browsers today that support `import`? It sounded like on chat it was a somewhat experimental feature, so while it's fine to support we may want to hold off on changing default examples until it's supported everywhere.

Pauan:
@alexcrichton It [doesn't work on Edge](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta#Browser_compatibility) (I tested it), but it works on all other modern browsers.

It obviously won't work in IE or older browsers, but maybe Webpack can help polyfill `import.meta`?

Pauan:
(Also, it's [stage 3 in TC39](https://github.com/tc39/proposal-import-meta/#importmeta), so it's unlikely to change at this point)

There's also [some discussion about Webpack polyfilling it](https://github.com/webpack/webpack/issues/6719).

alexcrichton:
Ok nice! I think in that case this is probably fine to go ahead and land and switch the defaults. FWIW this probably won't benefit much though from a webpack polyfill because this is the `--target web` target whereas the default bundler target already doesn't have to deal with this and webpack takes care of it!

Thanks @ibaryshnikov and @Pauan!

PR #1583, opened 2019-06-10
Adding in AsRef impl for all wasm_bindgen types

### LINKED ISSUE #1582: Types should implement AsRef for themself
### Motivation

I'm working on porting [dominator](https://crates.io/crates/dominator) to use wasm-bindgen, and I ran into this big issue.

Dominator contains a `struct DomBuilder<A>` which allows it to work on any type. Then I implement methods only for certain types:

```rust
impl<A> DomBuilder<A> where A: AsRef<Node> {
    pub fn children(...) { ... }
    ...
}

impl<A> DomBuilder<A> where A: AsRef<Element> {
    pub fn class(...) { ... }
    ...
}

impl<A> DomBuilder<A> where A: AsRef<HtmlElement> {
    pub fn style(...) { ... }
    ...
}
```

The idea is that you can use `DomBuilder` on any type, but at compile-time it will statically restrict which methods you can use depending on the type.

So you can't use `HtmlElement` methods on a `Node`, for example. But you can use `Node` methods on a `Node`, `Element`, or `HtmlElement`.

This works great in stdweb, but it doesn't work in wasm-bindgen, because wasm-bindgen implements these...

```rust
impl AsRef<JsValue> for HtmlElement
impl AsRef<Object> for HtmlElement
impl AsRef<EventTarget> for HtmlElement
impl AsRef<Node> for HtmlElement
impl AsRef<Element> for HtmlElement
```

...but it does **not** implement `impl AsRef<HtmlElement> for HtmlElement`

That's a big problem, since it means that you cannot write code which is generic over the type.

### Proposed Solution

Implement `impl AsRef<Foo> for Foo` for all types created by the `#[wasm_bindgen]` macro.

### Alternatives

* Create a new trait that is similar to `AsRef` (seems redundant).

* Don't do this (makes writing generic code impossible).

### Additional Context

There is precedence for this, since the stdlib [defines these](https://doc.rust-lang.org/nightly/std/convert/trait.AsRef.html):

```rust
impl AsRef<CStr> for CStr
impl AsRef<OsStr> for OsStr
impl AsRef<Path> for Path
impl<T> AsRef<Vec<T>> for Vec<T>
```

And wasm-bindgen itself defines `impl AsRef<JsValue> for JsValue`

Pauan:
Fixes #1582

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -703,6 +703,11 @@ impl ToTokens for ast::ImportType {
                     fn as_ref(&self) -> &JsValue { self.obj.as_ref() }
                 }
 
+                impl AsRef<#rust_name> for #rust_name {
+                    #[inline]
+                    fn as_ref(&self) -> &#rust_name { self }
+                }
+
 
                 impl From<#rust_name> for JsValue {
                     #[inline]


alexcrichton:
Thanks!

Issue #1524, opened 2019-05-10
Moving `wasm-bindgen` from "Future Compatible with" to "Polyfill for" WebIDL Bindings


fitzgen:
## Where we are Now

Here is a simplified view of our Rust and Wasm toolchain, its artifacts and the processes that input and output these artifacts:

```text
+--------------+
| Rust Sources |
+--------------+
       |
       |
     rustc
       |
       |
       V
+----------------------------------------+
| Initial .wasm w/ wasm-bindgen metadata |
+----------------------------------------+
       |
       |
  wasm-bindgen CLI ----------.
       |                     |
       |                     |
       V                     V
+-------------+       +------------------+
| Final .wasm |       | JS Bindings Glue |
+-------------+       +------------------+
```

Right now the initial `.wasm` contains two things that are used by the `wasm-bindgen` CLI, and stripped from the final `.wasm` binary:

* A `wasm-bindgen`-specific custom section. This contains a summary of all the imported and exported functions, structs, fields, and methods. It doesn't include functions' and methods' ABI types, however.

* "Descriptor" functions. These are evaluated by the CLI and communicate the types being sent and received over the ABI boundary for a given function or method. We do this because we want the ABI representation of types to be trait-based, so we can customize it for different types, and it was the best way at the time to support that.

The generated JS bindings glue is responsible for the following things:

* Translating JS values going to Wasm into their ABI representation.

* Translating values coming out of Wasm to JS from their ABI representation into JS values.

* Instantiating the Wasm module and hooking up imported functions.

## Vision

We leverage WebIDL bindings and `anyref` to completely remove the `wasm-bindgen`-specific metadata and 99% of the JS bindings glue:

```text
+--------------+
| Rust Sources |
+--------------+
       |
       |
     rustc
       |
       |
       V
+---------------------------------------------------+
| Final .wasm w/ WebIDL Bindings Section and anyref |
+---------------------------------------------------+
       |
       |
  wasm-bindgen CLI
       |
       |
       V
+-----------------------+
| JS Instantiation Glue |
+-----------------------+
```

`rustc` has first-class support for `anyref` and emitting the WebIDL bindings custom section.

The JS bindings glue is no longer responsible for translating between ABI representations and JS values:

* The ABI representation of most JS values is simply `anyref`. In JS hosts, an `anyref` *is* a JS value, or at least the conversion is so transparent that they are effectively the same from our perspective.

[...]

PR #1588, opened 2019-06-11
Communicate exceptions through global memory

alexcrichton:
Instead of allocating space on the stack and returning a pointer we
should be able to use a single global memory location to communicate
this error payload information. This shouldn't run into any reentrancy
issues since it's only stored just before returning to wasm and it's
always read just after returning from wasm.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -974,28 +974,12 @@ impl TryToTokens for ast::ImportFunction {
         }
 
         let mut exceptional_ret = quote!();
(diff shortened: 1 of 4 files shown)


alexcrichton:
Note that this is part of the webidl work for https://github.com/rustwasm/wasm-bindgen/issues/1524 where we ideally want to alter the ABIs of functions as little as possible. This is basically just a small refactoring which I think we want regardless and should make implementing https://github.com/rustwasm/wasm-bindgen/issues/1524 a bit simpler.

PR #1598, opened 2019-06-15
Fixing panic if the Future wakes up after returning Poll::Ready

Pauan:
Some things like `futures::channel::mpsc::UnboundedSender` will trigger wakeups after returning `Poll::Ready`, so this fixes it so it ignores the spurious wakeups, rather than panicking.

The diff looks scary, but it's really just a tiny bit of reorganization (changing `future.as_mut().unwrap_throw()` into `if let Some(future) = borrow.as_mut() { ... }`)

It looks a lot less scary if you [disable whitespace in the diffs](https://github.com/rustwasm/wasm-bindgen/pull/1598/files?w=1).

### DIFF
--- a/crates/futures/src/futures_0_3.rs
+++ b/crates/futures/src/futures_0_3.rs
@@ -233,27 +233,27 @@ where
 
                     match lock.pop_front() {
                         Some(task) => {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Good catch!

PR #1599, opened 2019-06-16
Make the argument to init optional in the Typescript declaration too

marienz:
Pull request #1579 made the argument to the generated init() function optional (when the target is "web"), but not in the .ts.d file. This brings the two back in sync.

Tested locally, using a .ts file that mostly contains the same code the "without a bundler" example includes in its index.html.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -396,7 +396,7 @@ impl<'a> Context<'a> {
         Ok(imports)
     }
 
-    fn ts_for_init_fn(has_memory: bool) -> String {
+    fn ts_for_init_fn(has_memory: bool, has_module_or_path_optional: bool) -> String {
         let (memory_doc, memory_param) = if has_memory {
             (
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -396,7 +396,7 @@ impl<'a> Context<'a> {
         Ok(imports)
     }
 
-    fn ts_for_init_fn(has_memory: bool) -> String {
+    fn ts_for_init_fn(has_memory: bool, has_module_or_path_optional: bool) -> String {
```
marienz: I'm not convinced taking a second bool flag as argument is the best interface (maybe these should be enums so it's harder to get them mixed up), but there's only two callers so maybe it's ok? If you prefer a different interface please let me know what it is and I'll change it.

(Note that I'm new to all of Rust, WebAssembly and Typescript, please don't assume I know what I'm doing :)

alexcrichton:

PR #1605, opened 2019-06-17
Forbid duplicated getter/setter names in fields and methods

c410-f3r:
It is not possible to add an `ui` test because this is not a parsing error and it is also not possible to create an ordinary `wasm32` test, therefore, I tested the intended behaviour locally and empirically.


alexcrichton:
Looks good to me, thanks!

PR #1612, opened 2019-06-22
Initial interning implementation

Pauan:
This is just a rough draft so I can make sure I'm going in the right direction.

The way I designed it, you can change (at runtime) the max str len and whether the interning is enabled or disabled.

In addition, there's a `disabled` Cargo feature to completely wipe out the implementation at compile-time, so there's 0 overhead.

The reason I went with a `disabled` feature (and not an `enabled` feature) is that I expect the interning to be enabled by default.

The cache size is hard-coded at 1,024 elements (I just guessed at what seemed like a reasonable number), but we can make that configurable later.

I took a look at the existing LRU crates, and I decided on uluru because it seemed to be the fastest and lightest weight (based on looking at the source code).

@alexcrichton does this seem reasonable?


alexcrichton:
I think this is a good start yeah but I'm personally mostly interested in how this hooks up into the wasm-bindgen crate and CLI moreso than the cache itself. It seems like we can easily go hog wild with configuration settings and implementations of the cache but hammering out the details of how the CLI/ABIs work I think will be pretty important too

Pauan:
Right, I'm just taking it one step at a time. The next step is to hook it into the `#[wasm_bindgen]` macro.

alexcrichton:
mk makes sense!

I figure this may make the most sense on the wasm-bindgen crate itself where a feature would basically just change the implementation of `IntoWasmAbi` for strings?

Pauan:
Yeah, that's what I was thinking. Basically, have `IntoWasmAbi` call the `intern::str` function.

And then `wasm-bindgen` could have a `disable-interning` feature which disables the interning for the `wasm-bindgen-cache` crate.

Pauan:
Okay, so because of circular dependencies I can't make `wasm-bindgen-cache` into a separate crate, so I had to move it into `wasm-bindgen`.

Here are the profiling results. First, without interning:

DevTools 007

Now with interning:

DevTools 006

The performance went from `1256.8ms` to `1108.9ms`, an improvement of `147.9ms`! As you can see from the charts, decoding went from `387.2ms` to `67.0ms`.

Unfortunately, the increase wasn't as big as I was hoping for. I need to investigate why the `cache_str` function is taking up a whopping `154.3ms`.

By the way, I had to implement `OptionIntoWasmAbi` for `JsValue`, and I'm not sure if the implementation is correct.

Pauan:
Okay, so the issue is that it was just trying to cache too many useless strings, so I took a different approach.

Rather than automatically caching all strings, instead the user has to manually call an `intern("foo")` function, which has the signature `fn intern(s: &str) -> &str`

This will add the string to the cache, and so when the string is sent to JS, it will then pick it up from the cache.

By carefully choosing which strings go into the cache and which don't, I was able to get incredible performance gains.

**With interning completely disabled:** (`1363.8ms`)

DevTools 008

**With interning enabled but not used:** (`1485.0ms`)

DevTools 009

**With all strings interned:** (`1403.1ms`)

DevTools 010

**With specific strings interned:** (`915.0ms`)

DevTools 011

So in the end, by interning only the strings which are commonly used, it gained `448.8ms` in performance!

And as you can see in the last chart, decoding is very fast (`69.8ms`), and the cache itself is also very fast (`37.5ms`).

In fact, because everything is so fast, the top 7 items are all *DOM built-ins*, which means the app is going as fast as possible, and the browser is now the bottleneck.

In addition, the Major GC doesn't even include decoding, which means the browser is doing 0 garbage collection of strings. Amazing.

alexcrichton:
Sorry I've been pretty busy but this all looks great to me! Some high-level thoughts I might have are:

* I think we'll probably want to have this as an off-by-default feature of `wasm-bindgen`. Anything on-by-default is very difficult to turn off since `wasm-bindgen` is so heavily depended on, and this also has some code-size impact with the interning infrastructure. Having it as a switch is perfect but I think the better default will be off-by-default. I also think that having the switch being disable vs enable makes it sort of go against how Cargo features work and may lead to surprising results.

* I'd ideally prefer to avoid `impl OptionIntoWasmAbi for JsValue` for now if we can, would it be possible to just do raw bits under the hood to avoid adding the impl just yet?

* Could the main test suite be run on CI with interning explicitly both enabled and disabled?

Pauan:
> Anything on-by-default is very difficult to turn off since wasm-bindgen is so heavily depended on

Actually, any crate in the dependency graph (including the root/application crate) can very easily turn it off, they just need to use `features = ["disable-interning"]`

> I also think that having the switch being disable vs enable makes it sort of go against how Cargo features work and may lead to surprising results.

I don't think so. It's true that the typical way of doing things (using `default-features = false`) is broken and almost unusable, but I'm not doing that.

And there isn't an "enable-interning" feature, so it's not possible for there to be any conflicts. And the API is exactly the same between the interning and not interning modes, so there's no compile errors.

So could you explain what sorts of surprising results you think might happen?

> would it be possible to just do raw bits under the hood to avoid adding the impl just yet?

Sure, but what bits should it use? Maybe it should just return `JsValue::UNDEFINED`?

Pauan:
(Originally I intended to have full automatic string interning on by default, but after seeing the benchmarks it's clear that manual interning is much faster, so I'm okay with making the feature off-by-default, but I still want to know why you think it would be problematic)

alexcrichton:
Ah yeah so I was halfway done writing my original comment before I realized the feature was to disable, not an on-by-default feature which enabled. So to be clear *if* the feature were `enable-intern` and it was on-by-default, I don't think that's tenable since it could effectively never be turned off. 

The feature is, however, off-by-default and instead worded as disabling interning. Cargo features are meant to be an additive set, however, due to the unioning behavior to compile crates only once. This means that if a crate relies on interning for reasonable performance, another crate may actually disable interning and the original crate has no course of action. In other words, the "disable" form is removing a feature rather than adding it which goes against how shared dependencies work in Cargo today.

That, plus the benchmark data, is basically why I think I'd prefer if this was an opt-in feature rather than on-by-default.

For returning `JsValue::UNDEFINED`, I think that'd work yeah!

Pauan:
> So to be clear if the feature were enable-intern and it was on-by-default, I don't think that's tenable since it could effectively never be turned off.

Right, I learned that the hard way with `futures-preview`, where you cannot turn off their features at all, not even with `default-features = false`

> This means that if a crate relies on interning for reasonable performance, another crate may actually disable interning and the original crate has no course of action.

That's absolutely true, but I don't think that's a problem, in fact I think that's *desirable*.

I don't see any reason why a crate would *want* to disable interning, *unless* they are the root application crate.

And in that case the root should definitely have the power to disable interning (if they find that it improves the performance of their app).

So, although it's quite rare in general, I think in this *very* specific case a "negative feature" can be appropriate.

> That, plus the benchmark data, is basically why I think I'd prefer if this was an opt-in feature rather than on-by-default.

Yeah, I've come around to that as well.

alexcrichton:
Ah ok well in any case, want to update this to opt-in and we can go ahead and merge?

Pauan:
@alexcrichton So, while working on this, I noticed something: the `IntoWasmAbi` impl for `JsValue` uses `mem::forget`, so doesn't that mean that the object will stay in the JS heap forever, and never be cleaned up?

alexcrichton:
Oh right yeah I was thinking about that when reading this but forgot to say we should update it. I think that this is always returning (effectively) `&JsValue`, right? If `str` and `String` report `ANYREF` that may mean that `&str` is ok (since it'll show up as `REF ANYREF`) but for `String` I think we'll need to insert the `REF` there ourselves.

Pauan:
If the string exists in the cache, then yeah it can return `&JsValue` (which is fine).

The problem is that if the string *isn't* in the cache then the `into_abi` method uses `JsValue::from` to convert the `&str` into a `JsValue`, and then calls `into_abi` on that.

And since `JsValue` leaks (at least, if I'm understanding it correctly), that's quite bad. Maybe it's best we discuss this on Discord, since it seems to have some pretty deep implications?

alexcrichton:
Ah that's a good point! I'll be around sort of only spottily this week but feel free to ping me when you're around and we can hope it aligns! I'll be around-ish 9am-5pm CST.

Otherwise though it's true we don't have any precedent for returning a "maybe owned" `JsValue` from any APIs. That being said it doesn't seem entirely unresonable to invent one for strings. In the `anyref` world it will have no cost because the apis will still take `anyref` at the boundary, and for now it just means a bit more JS glue. I would imagine that the high bit (or something like that) would indicate whether it was owned or not and whether `free` needs to be called.

crates/cli-support/src/js/outgoing.rs
```diff
@@ -408,4 +429,15 @@ impl<'a, 'b> Outgoing<'a, 'b> {
         ));
         self.cx.require_internal_export("__wbindgen_free")
     }
+
+    fn prelude_free_cached_string(&mut self, ptr: &str, len: &str) -> Result<(), Error> {
+        self.js.prelude(&format!(
+            "if ({ptr} !== 0) {{ wasm.__wbindgen_free({ptr}, {len} * {size}); }}",
+            ptr = ptr,
+            len = len,
+            size = VectorKind::String.size(),
```
alexcrichton: This is probably safe to omit since it's a constant 1

Cargo.toml
```diff
@@ -38,6 +39,8 @@ xxx_debug_only_print_generated_code = ["wasm-bindgen-macro/xxx_debug_only_print_
 wasm-bindgen-macro = { path = "crates/macro", version = "=0.2.48" }
 serde = { version = "1.0", optional = true }
 serde_json = { version = "1.0", optional = true }
+uluru = "0.3.0"
```
alexcrichton: This'll want to be an optional dep I think, right?

src/cache/intern.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        thread_local! {
+            static CACHE: Cache = Cache {
+                entries: RefCell::new(LRUCache::default()),
+            };
+        }
+
+        fn get_js_string<'a>(cache: &'a mut Entries, key: &str) -> Option<&'a JsValue> {
+            cache.find(|p| p.key == key).map(|x| &x.value)
```
alexcrichton: Is this doing a linear search over all entries in the cache each time? I'm actually sort of surprised that's faster than copying the string across the boundary...

src/cache/intern.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        pub(crate) fn get_str(s: &str) -> Option<JsValue> {
```
alexcrichton: Instead of `_unsafe_clone` could this perhaps have a closure-like API or return a raw index?

alexcrichton: (basically somehow deferring the "oh the use-after-free possibility here is ok" to the callers of this which have to uphold that invariant)

Cargo.toml
Pauan: Ah, good catch!

src/cache/intern.rs
Pauan: Yes, but there's three things to keep in mind: the entries are kept in "most recently used" order, so more recently used strings will be found faster. And once it finds a match it immediately stops.

Secondly, the cache is quite small (only 1,024 items), which makes linear search fast. BTreeMap also uses linear search for small node sizes, because that's actually faster than doing something fancy.

Thirdly, doing a full `O(n)` copy of the string *and* encoding from UTF-8 to UTF-16 is *quite* slow (as shown in the real world benchmarks). If sending strings from Rust to JS was fast enough then I wouldn't have created an interning system in the first place.

That's all implementation stuff though, we can always optimize it to be faster later.

src/cache/intern.rs
Pauan: Keep in mind this is an internal API, so the only callers are ourselves.

As for changing the API, that is possible, but the reason I went with this approach is because later it calls `into_abi()` on the `JsValue`. So that means if the `into_abi` implementation ever changes in the future, everything will continue to work fine.

Is there any benefit to deferring the unsafety to the caller? How would the caller be able to handle it differently?

Pauan:
Now that I can build `wasm-bindgen-cli`, I was able to actually test this.

So, I significantly changed the implementation (before your review). Now the implementation is almost identical to the current string impl, except if the `ptr` is `0`, then it assumes that the `len` is a pointer to the `JsValue`.

What that means is that there is now basically no performance difference regardless of whether interning is enabled or not:

```
Without interning:
   interned   0.016361628484616742 ms
   normal     0.016440500294726636 ms  (100.48205354487818 %)

With interning:
   interned   0.006101373767490259 ms
   normal     0.017310679540981640 ms  (283.7177363763148 %)
```

~183% increase in performance sounds pretty great to me. And it doesn't regress the performance when not interning (which was a big problem with the old implementation). And the memory leak has been fixed.

src/cache/intern.rs
alexcrichton: I think it's just sort of a weird API that is basically incorrect to use *except* in this exact construction. Everything here is fine in the sense that it all technically works, but you don't have an owned `JsValue` and this function isn't the one that's "maybe this is a use-after-free" but it's later when the index is passed to JS. 

What I think I'm getting at is that the technical construction here is fine, but stylistically I think this is a bit weird and it'd be best if this did something like take an `FnOnce(Option<&JsValue>)` or just return the raw u32 index (or maybe even `*const JsValue` or something like that

crates/cli-support/src/js/mod.rs
```diff
@@ -1088,6 +1088,28 @@ impl<'a> Context<'a> {
...
+            return Ok(());
+        }
+
+        self.expose_get_object();
+        self.expose_get_string_from_wasm()?;
+
+        self.global("
+            function getCachedStringFromWasm(ptr, len) {
+                if (ptr === 0) {
+                    if (len !== 0) {
```
alexcrichton: Is this `if` necessary? I don't think ptr/len will ever be both zero, right?

Pauan: It is there to support `Option<String>` and `Option<&str>`.

Using the same function for optional and non-optional basically has no performance cost, but it makes the implementation a *lot* easier, and also reduces the code size.

src/cache/intern.rs
alexcrichton: My worry isn't so much for cache hits but cache misses. What happens if 1024 kilobyte strings are interned and then you pass `"a"` along the boundary? Every single string needs to be checked every single time for strings which aren't hits in the cache.

Like I agree encoding is slow, but this sort of interning is still slower than it would otherwise be because it's doing some linear lookup at least with respect to the length of the string, and this is also with respect to the cache.

My main point is that this seems like a case where we want some sort of LRU hash table, not just an LRU list.

crates/cli-support/src/js/mod.rs
alexcrichton: Ah ok I think I missed that, if that's the case I think this could go two ways: it could either still remove this `if` because we can guarantee that `getObject(0)` is always undefined or this I think should have an explicit `return undefined;` to be clear

src/cache/intern.rs
Pauan: Hmmm, originally I was going to say that the API is completely safe to use (in every situation), because we never remove items from the cache.

But after thinking about it more, once the cache reaches maximum capacity it automatically removes items. So I need to think hard about how to solve that.

src/cache/intern.rs
Pauan: String equality *also* short-circuits if it doesn't match. So if you pass `"a"`, it will check the length of `"a"` and the cached string, and they won't match, so it'll immediately move onto the next string.

That means you have exactly 1,024 very fast operations (just comparing string lengths) before you get a cache miss.

And even if the length of the strings are the same, it will also abort as soon as even a single char is different.

So the only time it would be an issue is if you're passing a lot of strings which are prefixes of cached strings, and the prefix is really long.

I have no problem with improving the performance, but I think it should be done in a new PR, with proper benchmarks to make sure it doesn't reduce the performance.

(Also keep in mind that all of this only matters if you enable the `enable-interning` feature, if you don't enable that feature then there's 0 cost, the cache doesn't even exist at all).

crates/cli-support/src/js/mod.rs
Pauan: I don't really feel comfortable relying upon `getObject(0) === undefined`, feels like it's leaking internal details.

I guess it's okay if we guarantee that `0` will always be `undefined` (even with `anyref` support).

Pauan: I went ahead and removed the `if`, since I think it's reasonable to guarantee that `getObject(0) === undefined` (both for pragmatic and theoretical reasons, since `undefined` is the closest thing JS has to NUL).

src/cache/intern.rs
Pauan: Okay, so I changed it so it returns the raw index (but internally it still uses `into_abi`).

I'm pretty sure that this is safe, since it should always immediately return the string from the heap, so use-after-free cannot occur.

Pauan:
I rebased, and I also created a benchmark with the 1,000 most common English words. These are the results:

```
No interning (113,556 bytes):
   normal str            0.015406891304873488 ms

LRU Cache (139,951 bytes):
   interned str (hit)    0.001420249078550068 ms
   normal str            0.015683253470461032 ms  + 1004.2607742067371 %
   interned str (miss)   0.065211217786988060 ms  + 4491.5338916158410 %

BTreeMap (119,950 bytes):
   interned str (hit)    0.0021928430482934585 ms
   normal str            0.0148238589445778490 ms  + 576.0109418735763 %
   interned str (miss)   0.0159713026124058200 ms  + 628.3376995373748 %

HashMap (121,198 bytes):
   interned str (hit)    0.0017783651625431442 ms
   normal str            0.0157200559689065660 ms  + 783.9610840344038 %
   interned str (miss)   0.0159029148265826120 ms  + 794.2434974288807 %
```

The cache is full and it has a lot of similar words, so it's a reasonably accurate stress test. In practice it will be a lot faster.

The interesting things to note:

* The LRU cache is actually the biggest in terms of file size, which is surprising. Maybe it's because it statically allocates the entire cache upfront?

* The LRU cache is indeed slower during misses (about 4.5x slower). Not too bad for the worst-case situation, but still not good.

   I tested it on lighter work-loads and it didn't have any cost during misses, so this is specific to heavy work-loads.

* Meanwhile, the BTreeMap and HashMap (from the Rust stdlib) perform exceptionally well: misses are the same performance as not interning.

* On the other hand, the LRU cache is fastest for hits (10x faster). The BTreeMap and HashMap are *very* slightly slower on hits, but are much faster on misses.

Given the results, I switched it to use HashMap. This gives a 7.8x speed boost for hits, no speed penalty for misses, infinite scalability, and smaller file size than the LRU cache.

Pauan:
I just added in docs for the `intern` function. The only thing left is adding in CI testing for the `enable-interning` feature, but I'm not sure how to do that.

Pauan:
Also, the CI failures seem really weird, I have no clue what's going on with those.

crates/cli-support/src/js/outgoing.rs
```diff
@@ -130,6 +130,27 @@ impl<'a, 'b> Outgoing<'a, 'b> {
...
+            NonstandardOutgoing::CachedString {
+                offset,
+                length,
+                owned,
+            } => {
+                let ptr = self.arg(*offset);
+                let len = self.arg(*length);
+                let tmp = self.js.tmp();
+
+                self.js.typescript_optional("string");
```
alexcrichton: I think this needs to be `typescript_required` because this is a required string argument, not an optional one

src/cache/intern.rs
```diff
@@ -0,0 +1,78 @@
...
+            })
+        }
+
+        fn intern_str(key: &str) {
+            CACHE.with(|cache| {
+                let mut cache = cache.entries.borrow_mut();
+
+                // Can't use `entry` because `entry` requires a `String`
+                if !cache.contains_key(key) {
+                    cache.insert(key.to_owned(), JsValue::from(key));
```
alexcrichton: I think that we're going to want some form of eviction for this, because otherwise the interning feature may not be too useful as the Rust memory would just grow without bound. Could we perhaps have something like a 1MB string limit or maybe a maximum amount limit for interning?

(it may also be good to skip interning huge strings over a threshold as well)

src/cache/intern.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        thread_local! {
+            static CACHE: Cache = Cache {
+                entries: RefCell::new(HashMap::new()),
+            };
+        }
+
+        /// This returns the raw index of the cached JsValue, so you must take care
+        /// so that you don't use it after it is freed.
+        pub(crate) fn unsafe_get_str(s: &str) -> Option<<JsValue as IntoWasmAbi>::Abi> {
```
alexcrichton: Note that this is somewhat misleading here and I think this should probably just return `u32`. Below the `into_abi` call isn't called on `JsValue` but is instead called on `&JsValue`. The two trait impls though have the same `Abi` type which is just a `u32`

src/cache/intern.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// When you send that cached string to JS, it will look it up in the cache,
+/// which completely avoids the `O(n)` copy and encoding. This has a significant
+/// speed boost (as high as 783%)!
+///
+/// However, there is a small cost to this caching, so you shouldn't cache every
+/// string. Only cache strings which have a high likelihood of being sent
+/// to JS multiple times.
+#[inline]
+pub fn intern(s: &str) -> &str {
```
alexcrichton: Perhaps this could be renamed to `insert_in_intern_cache` and not have a return value? (or something like that name)

I would expect an `intern` function to return some sort of different type which is zero cost to pass around (aka `JsValue::from(&str)`) so here it's more of "populate the cache so the next time it can be faster". 

It may also be worth pointing out that using `intern` is always going to be slower than just using `JsValue::from` because we still have to do some sort of lookup in the intern cache.

src/describe.rs
```diff
@@ -75,7 +76,7 @@ simple! {
     f64 => F64
     bool => BOOLEAN
     char => CHAR
-    str => STRING
+    str => if cfg!(feature = "enable-interning") { CACHED_STRING } else { STRING }
```
alexcrichton: I believe this is where the CI issues are coming from. In debug mode this is literally emitting an if/else block and some other assorted instructions, but our "poor man's interpreter" doesn't actually implement if/else/etc. I think you'll want to use `#[cfg]` instead of `if cfg!(...)` in this scenario to ensure the codegen is improved enough in debug mode that we can still handle it.

(same for the block below with `String`)

src/cache/intern.rs
alexcrichton: It might also be good to throw "maybe" in the name somewhere since if the feature isn't turned on this doesn't actually do anything.

crates/cli-support/src/js/outgoing.rs
Pauan: Ah, good catch, I need to pass in a boolean indicating whether it's required or not.

src/cache/intern.rs
Pauan: Do you think that will be an issue in practice? Strings are only interned if the user explicitly and manually calls the `intern` function. They are never automatically interned.

And the documentation for `intern` makes it clear that you should only use `intern` if there's a high probability that the string will be sent to JS multiple times.

So, what if the user actually *wants* to intern a large string which is sent to JS multiple times?

src/cache/intern.rs
Pauan: > I would expect an intern function to return some sort of different type which is zero cost to pass around (aka JsValue::from(&str)) so here it's more of "populate the cache so the next time it can be faster".

Yes, that's very intuitive, but that's not actually possible. It wouldn't be the first wasm-bindgen API which is a bit unintuitive for technical reasons.

> Perhaps this could be renamed to insert_in_intern_cache and not have a return value? (or something like that name)

I had considered that, but that's a *really* bad UX. The choice is between these two options:

```rust
document.create_text_node(intern("foo"));
```

```rust
let s = "foo";
insert_in_intern_cache(s);
document.create_text_node(s);
```

And it gets even worse when you consider things like dominator, which provide a convenient high-level API on top of the DOM:

```rust
html!("div", {
    .style("foo", intern("bar"))
})
```

```rust
html!("div", {
    .style("foo", {
        let s = "bar";
        insert_into_intern_cache(s);
        s
    })
})
```

Basically, every time you want to intern a string you *also* want to immediately use it.

The current API is designed for that, rather than the *incredibly* rare case of "I want to intern this string but *not* immediately use it".

> It might also be good to throw "maybe" in the name somewhere since if the feature isn't turned on this doesn't actually do anything.

I don't think that's necessary. I view it as being a performance hint, it's not guaranteed that the strings will be cached anyways (e.g. they might get evicted if the cache grows too large).

src/describe.rs
Pauan: Interesting, that seems like quite the limitation indeed.

Pauan:
> For the cached case though we'd need to check both the pointer and the length to see if they're both zero.

I don't think there will be any issue: it's guaranteed that non-undefined `JsValue` will never have `0` as their idx, so the length will only be `0` if it is in fact `None`.

I agree that we should have tests for it.

src/cache/intern.rs
Pauan: Ah, right, good catch.

Pauan:
Wait, nevermind, it doesn't support *receiving* cached strings in the first place, they use completely different systems, so there shouldn't be any problem with `OptionFromWasmAbi`.

src/cache/intern.rs
alexcrichton: Oh wow I've been thinking about this wrong the entire time, I thought everything automatically went through interning!

Nah in that case this is fine for now.

alexcrichton:
Oh I'm mostly worried about [this block](https://github.com/rustwasm/wasm-bindgen/blob/efacd8b74d73efe1b7ccc6069a0207ca9342e2c3/crates/cli-support/src/js/outgoing.rs#L273-L288) which handles outgoing optional slices (including strings). That's only checking `{ptr}` when to handle cached strings it'd also need to handle `{len}` being zero

src/cache/intern.rs
alexcrichton: Yeah I definitely agree there's different designs here, and I don't really mind one way or another. Looks like you feel stronger about this than I, so leaving it as `intern` seems fine.

Pauan:
> Oh I'm mostly worried about this block which handles outgoing optional slices (including strings).

That code path shouldn't ever be called. When caching is enabled, it will *always* go through `NonstandardOutgoing::CachedString` for all types (`&str`, `String`, `Option<&str>`, and `Option<String>`).

src/cache/intern.rs
Pauan: Yeah, it *used* to automatically cache, but when that caused performance issues [I changed it to manual caching](https://github.com/rustwasm/wasm-bindgen/pull/1612#issuecomment-504700943).

Pauan:
Okay, I think this is ready for merging. The CI failures seem spurious, and I added in CI tests for `enable-interning`.

alexcrichton:
Works for me, thanks again for pushing on this!

PR #1616, opened 2019-06-23
Simplify instructions for CI testing of wasm code

### LINKED ISSUE #1611: Consistent way to test crates on CI (without hardcoding a wasm-bindgen version in CI configs)
### Motivation

I'm trying to add CI testing to my crate following the explanation here:
https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/continuous-integration.html

As you can see there is a lot of hardcoded version numbers for wasm-bindgen, nvm, geckodriver, chromedriver. This means that whenever wasm-bindgen is updated we would have to manually modify configuration files like `.travis.yml`. 

This also means that the book itself is outdated all the time.

### Proposed Solution

I'm wondering if we can make a simpler way to set this up which would just use the latest version of everything. I'm suspecting that `wasm-bindgen` can be installed with `cargo install`.

In either case, the book needs an update (wasm-bindgen-0.2.21), and maybe it's a good moment to consider streamlining this workflow?

najamelan:
After discussion on #1611, this replaces the complicated setup with the automated install of wasm-pack.

I have added a note on caching, which I think would be useful to users, since the installer is really stuborn about not wanting to be automated. eg. `yes | installer` won't work. You have to remove all wasm-pack binaries before running it in order for it to be automated.

Closes #1611.

### DIFF
--- a/guide/src/wasm-bindgen-test/continuous-integration.md
+++ b/guide/src/wasm-bindgen-test/continuous-integration.md
@@ -29,22 +29,6 @@ script:
   - wasm-pack test --chrome  --headless
 ```
 
-Note that the wasm-pack installer will ask for confirmation before overwriting the wasm-pack binary. So if you use caching you can make sure it's not caching wasm-pack by adding this to your `travis.yml`:
-```yml
-# Need to cache the whole `.cargo` directory to keep .crates.toml for cargo-update to work
-#
-cache:
-  directories:
-    - /home/travis/.cargo
-
-# But don't cache the cargo registry
-# and remove wasm-pack binary to avoid the installer asking confirmation for overwriting it.
-#
-before_cache:
-  - rm -rf /home/travis/.cargo/registry
-  - rm -rf /home/travis/.cargo/bin/wasm-pack
-```
-
 ## AppVeyor
 
 ```yaml
--- a/guide/src/wasm-bindgen-test/continuous-integration.md
+++ b/guide/src/wasm-bindgen-test/continuous-integration.md
@@ -10,37 +10,39 @@ request!](https://github.com/rustwasm/wasm-bindgen)
 
 ```yaml
 language: rust
-rust: nightly
+rust    : nightly
 
 addons:
   firefox: latest
-  chrome: stable
+  chrome : stable
 
 install:
-  - rustup target add wasm32-unknown-unknown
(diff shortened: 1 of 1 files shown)


guide/src/wasm-bindgen-test/continuous-integration.md
```diff
@@ -10,37 +10,39 @@ request!](https://github.com/rustwasm/wasm-bindgen)
...
+
+  # this will test the non wasm targets if your crate has those, otherwise remove this line.
+  #
+  - cargo test
+
+  - wasm-pack test --firefox --headless
+  - wasm-pack test --chrome  --headless
+```
+
+Note that the wasm-pack installer will ask for confirmation before overwriting the wasm-pack binary. So if you use caching you can make sure it's not caching wasm-pack by adding this to your `travis.yml`:
```
alexcrichton: I think this note can be removed by passing `-y` to the installer script?

najamelan: I'll try that, when looking over the script I didn't see any argument parsing, so I didn't try that. I also didn't see any usage information.

I tried several things without success: 
```
curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf > wasm-pack-installer.sh
chmod +x wasm-pack-installer.sh
./wasm-pack-installer.sh -y
./wasm-pack-installer.sh -h
yes | ./wasm-pack-installer.sh
```

alexcrichton: I think `curl https://rustwasm.github.io/wasm-pack/installer/init.sh | sh -s -- -y` should do the trick

alexcrichton: Er sorry looks like the script doesn't actually support `-y` (now I'm actually reading it)

Hm for now can you open an issue in the wasm-pack repo? I'd prefer to avoid weird gotchas in the documentation like this

najamelan: I have opened rustwasm/wasm-pack#668.

Personally I feel it's more honest towards users to remove gotchas from the documentation once the issue is resolved, but it's your call.

alexcrichton: Thanks for that! I agree it's good to be truthful but I don't think that this is the place to have exhaustive documentation about everything. You already have to opt-in to caching and caching already has its own set of issues independent of this, so I would prefer that we delete this section with caveats about caching.

Issue #1619, opened 2019-06-24
Creating Uint8Array from Vec<u8> using sub_array() gives diffferent values compared to view()


nrayamajhee:
I am working on this WebGL example from: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Tutorial/Using_textures_in_WebGL
in Rust.

I tried creating a Uint8Array similarly to how its done for f32 at: https://rustwasm.github.io/docs/wasm-bindgen/examples/webgl.html

But the result produced by `Uint8Array(..).sub_array(..)` produces different output than produced by the unsafe `view(..)` method:

```rust
let data = vec![255,0,255,255];
let memory_buffer = wasm_bindgen::memory()
	.dyn_into::<WebAssembly::Memory>()?
	.buffer();
let buffer_loc = data.as_ptr() as u32;
let array = Uint8Array::new(&memory_buffer)
	.subarray(buffer_loc, buffer_loc + data.len() as u32)
log!("{:?}", array.to_string()); //displays "255,0,0,0"
let array = unsafe {
       js_sys::Uint8Array::view(&data[..])
};
log!("{:?}", array.to_string()); //displays "255,0,255,255"
```

nrayamajhee:
Never mind, It seems to work if I do it on a separate empty project. Must be something wrong with the rest of my code!

nrayamajhee:
Hmm.. This is weird. I just tried commenting the bottom block `let array  = unsafe{...` from the above code and the problem still exists. Here's my entire code. I just spun up a new wasm-pack project and edited the lib.rs file:

```rust
use wasm_bindgen::prelude::*;

macro_rules! log {
    ( $( $t:tt )* ) => {
        web_sys::console::log_1(&format!( $( $t )* ).into());
    }
}

#[wasm_bindgen(start)]
pub fn main() {
		use wasm_bindgen::JsCast;
		let data = vec![255,0,255,255];
		let memory_buffer = wasm_bindgen::memory()
			.dyn_into::<js_sys::WebAssembly::Memory>().unwrap()
			.buffer();
		let buffer_loc = data.as_ptr() as u32;
		let array = js_sys::Uint8Array::new(&memory_buffer)
			.subarray(buffer_loc, buffer_loc + data.len() as u32);
		log!("AN ARRAY: {:?}", array.to_string()); //displays "255,0,0,0"
//		let another_array = unsafe {
//			   js_sys::Uint8Array::view(&data[..])
//		};
[...]

Issue #1206, opened 2019-01-25
How to coerce Rust memory into Javascript generically


Thomspoon:
Hello,

I created a trait that would allow me to coerce slices into JS datatypes:

```rust
pub trait FromJsSys {
    fn from_js_sys(memory: &JsValue, data: Self) -> Uint8ClampedArray;
}

macro_rules! from_js_sys {
    ($type:ty, $id:ident) => {
        impl FromJsSys for $type {
            fn from_js_sys(memory: &JsValue, data: $type) -> Uint8ClampedArray {
                let data_size = data.len();
                let data_pointer = data.as_ptr() as u32;
                let data_normalization = size_of_val(&data[0]) as u32;
                let data_location = data_pointer / data_normalization;
                $id::new(&memory)
                    .slice(data_location, data_location + data_size as u32)
            }
        }
    };
}

from_js_sys!(&[f32], Float32Array);

//...

let vertices: [f32; 9] = [-0.7, -0.7, 0.0, 0.7, -0.7, 0.0, 0.0, 0.7, 0.0];
FromJsSys::from_js_sys(&vertices[..]);
```
However, when I look at the console I have the error: `wasm-bindgen: imported JS function that was not marked as `catch` threw an error: TypeError: "attempting to access detached ArrayBuffer"`

Is this due to the memory being resized?

fitzgen:
> Is this due to the memory being resized?

Most likely yes. Whenever wasm memory grows (eg someone called `malloc`), any existing array buffer views of the memory get detached.

If wasm can be called again between uses the `Uint8ClampedArray`, it is more robust to recreate the `Uint8ClampedArray` on each use.

Thomspoon:
So you're saying construct the Float32Array, then immediately construct a new Uint8ClampedArray using the data from the former? Is there a way to detect when a memory expansion happened? Lastly, is this going to force WebAssembly to have global variable structs where you have to reset all of them everytime you detect a memory change?

alexcrichton:
You should be able to detect when changes happen by using `===` on two `ArrayBuffer` instances. For example in `wasm-bindgen` you'll see code that [looks like this](https://github.com/rustwasm/wasm-bindgen/blob/c56dff8edecad4219ee59d4825d0072943e97944/crates/cli-support/src/js/mod.rs#L1438-L1444) where it only recreates typed array views if the underlying buffer changed, intending to reuse instances where possible

fitzgen:
[...]

Issue #1134, opened 2018-12-28
convert Vec<[u8]> to JsValue


lcnr:
I am using `fetch` in combination with `bincode`. As `bincode::deserialize` expects a byte array and `bincode::serialize` creates a byte array. To do this I currently use the following code: 

```
let data = bincode::serialize(content).unwrap()
opts.body(Some(&JsValue::from_str(unsafe { std::str::from_utf8_unchecked(data) })));
```

Is there a better way to do this?

alexcrichton:
Currently there's not a great way to do this inline in Rust, but it's covered by https://github.com/rustwasm/wasm-bindgen/issues/811 so I'm going to close this in favor of that. Thanks for the report though!

rylev:
@alexcrichton I have a `Vec<f64>` and I need a `&JsValue`. I believe it's not possible to do this conversion without doing the serialization trick. Am I missing something?

alexcrichton:
@rylev I think [`Float64Array::view`](https://docs.rs/js-sys/0.3.10/js_sys/struct.Float64Array.html#method.view) might do the trick for you?

alexcrichton:
I believe if you `slice()` just afterwards you can make a safe wrapper as that'll do a copy

PR #1620, opened 2019-06-24
Add From impl for TypedArrays

Pauan:
There's been quite a few issues ([1](https://github.com/rustwasm/wasm-bindgen/issues/1134), [2](https://github.com/rustwasm/wasm-bindgen/issues/1206), [3](https://github.com/rustwasm/wasm-bindgen/issues/1619)) about TypedArrays and Rust slices.

Right now the only way to convert a Rust slice into a TypedArray is to use the unsafe `view` function. But it's common for people to make mistakes when using it, and those mistakes are quite bad (since it's `unsafe`).

So this PR adds in safe `From` impls for all of the TypedArrays, allowing for easy, safe, and fast conversion from Rust slices.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -4593,11 +4593,6 @@ macro_rules! arrays {
         impl<'a> From<&'a [$ty]> for $name {
             #[inline]
             fn from(slice: &'a [$ty]) -> $name {
-                // TODO if this was written in JS it could avoid passing the `view` TypedArray to Rust,
-                //      which would be more efficient
-
-                // TODO measure if it's faster to use `slice` instead of `new`
-
                 // This is safe because the `new` function makes a copy if its argument is a TypedArray
                 unsafe { $name::new(&$name::view(slice)) }
             }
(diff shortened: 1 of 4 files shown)


alexcrichton:
Seems reasonable to me! Did you want to measure some of the TODO before merging though?

Pauan:
I ran some benchmarks, and the differences are negligible:

```
Chrome:
  new           309.59499999880790ms  (100%)
  slice len     326.13338709568546ms  (105.34194640191212%)
  slice length  328.74645161322286ms  (106.18597372547494%)
  raw js        330.33241935434840ms  (106.698245848423%)
  slice         331.26806451248063ms  (107.00046191439563%)

Firefox:
  raw js        343.23333333333335ms  (100%)
  slice         344.03333333333336ms  (100.23307780432312%)
  slice len     346.96551724137930ms  (101.08736153827171%)
  slice length  355.10344827586210ms  (103.4583254098343%)
  new           361.75000000000000ms  (105.3947800125618%)
```

So I went ahead and removed the TODOs.

alexcrichton:
👍

PR #1633, opened 2019-06-27
Attempted to tackle #1622

### LINKED ISSUE #1622: Function::bind2, 3, 4 Are Missing
`js_sys::Function` exports

```rust
pub fn bind(&self, context: &JsValue) -> Function
```

but the official calling convention (copied from MDN) is actually

```js
function.bind(thisArg[, arg1[, arg2[, ...]]])
```

So there's no way right now to supply more parameters to the function binding. It would need a similar treatment as `call0`, `call1`, `call2` and `call3` in the same struct.

Warfields:
Trying to fix #1622

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1077,6 +1077,14 @@ extern "C" {
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind)
     #[wasm_bindgen(method, js_name = bind)]
+    pub fn bind(this: &Function, context: &JsValue) -> Function;
+
+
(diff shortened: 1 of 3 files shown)


alexcrichton:
Looks great, thanks! Can you be sure to add a smoke test or two as well?

I also think for now we'll probably want to leave `bind0` as `bind` to avoid a breaking change, but we can file an issue about changing it on the next major breaking version?

Warfields:
> Looks great, thanks! Can you be sure to add a smoke test or two as well?
> 
> I also think for now we'll probably want to leave `bind0` as `bind` to avoid a breaking change, but we can file an issue about changing it on the next major breaking version?

Sounds good

Warfields:
@alexcrichton The tests have been written, I've added declarations for bind() and bind0()

alexcrichton:
Thanks!

PR #1645, opened 2019-07-06
Add in (unsafe and incorrect) impls of Send/Sync that are now required.

derekdreery:
To fix a soundness hole in futures, extra Send/Sync bounds were added to `ArcWake`. Since we already assume we are single-threaded, (see `crates/futures/src/futures_0_3.rs L224`), this patch impls `Send` and `Sync` so the futures 0.3 feature compiles again. This means we can continue to experiment while a full thread-safe solution is developed.


derekdreery:
There were a couple of warnings about trait object being labelled `dyn`, which I fixed. And my editor automatically runs rustfmt on the codebase when I save (I need to think about disabling it). I can revert out the formatting changes if you like.

alexcrichton:
thanks!

PR #1646, opened 2019-07-08
Added explicit extension for imported .wasm file for --target bundler

### LINKED ISSUE #1441: Explicitly add ".wasm" extension when importing wasm file in bundle target
## 💡 Feature description
When Parcel resolves generated pkg as a dependency, it cannot recognize "generated_pkg_bg" as a wasm file like what Webpack can do. Adding wasm extension explicitly can archive the compatibility.

There are plugins and built-in support in Parcel to resolve rs file or "Cargo.toml" directly, but this is fit for building components in a project, not for using package. Besides, Parcel encapsulate the pack so WebAssembly memory cannot be manipulated (I can still come back and import the generated bindgen file).

#### 💻 Basic example
In generated "my_pkg.js" (bundle target)
```js
import * as wasm from './my_pkg_bg';
```
⬇️
```js
import * as wasm from './my_pkg_bg.wasm';
```

ibaryshnikov:
Closes #1441 

This update is more in line with esm-module integration. At the moment there's a comment near this code:
```
// With Bundlers and modern ES6 support in Node we can simply import
// the wasm file as if it were an ES module and let the
// bundler/runtime take care of it.
```
If I get it right, without extension this will not work in node.js

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -284,7 +284,7 @@ impl<'a> Context<'a> {
             | OutputMode::Node {
                 experimental_modules: true,
             } => {
-                imports.push_str(&format!("import * as wasm from './{}_bg';\n", module_name));
+                imports.push_str(&format!("import * as wasm from './{}_bg.wasm';\n", module_name));
                 for (id, js) in sorted_iter(&self.wasm_import_definitions) {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks!

I don't think this works for some more esoteric use cases but since this is scoped to *just* bundlers I think it should work out alright.

PR #1649, opened 2019-07-08
Update futures implementation to not destroy callbacks

### LINKED ISSUE #1637: Issue with JSFuture::from(promise) imported from local source js file
# WASM-bindgen-futures aint workin' for me

This repo exemplifies my issue.

# Clone & Run it

```bash
git clone https://github.com/CallumJHays/wasm-bindgen-futures-test
cd wasm-bindgen-futures-test
yarn # install
rimraf dist pkg && webpack-dev-server --open -d # run
```

# The Code

[src/lib.rs](https://github.com/CallumJHays/wasm-bindgen-futures-test/blob/master/src/lib.rs):

```rust
use wasm_bindgen::prelude::*;
use js_sys::Promise;
use wasm_bindgen_futures::JsFuture;

// This is like the `main` function, except for JavaScript.
#[wasm_bindgen(start)]
pub fn main_js() -> Result<(), JsValue> {
    // This provides better error messages in debug mode.
    // It's disabled in release mode so it doesn't bloat up the file size.
    #[cfg(debug_assertions)]
    console_error_panic_hook::set_once();


    // Your code goes here!
    JsFuture::from(promise_true());

    Ok(())
}

#[wasm_bindgen(module = "/src/promise_true.js")]
extern "C" {
    fn promise_true() -> Promise;
}
```

[src/promise_true.js](https://github.com/CallumJHays/wasm-bindgen-futures-test/blob/master/src/promise_true.js):

```javascript
export const promise_true = async () => {
  console.log("promise_true() successfully called from Rust");
  return true;
};
```

It all compiles fine.

# Output

This is the runtime error I experience:

Help? 🐱

alexcrichton:
JS engines guarantee that at least one of our `then` callbacks are
invoked, so that means if we destroy them prematurely they're guaranteed
to log an exception to the console! Instead to prevent exceptions from
happening tweak how the completion callbacks for JS futures are managed
and ensure that the closures stay alive until they're invoked later.

Closes #1637

### DIFF
--- a/crates/futures/src/futures_0_3.rs
+++ b/crates/futures/src/futures_0_3.rs
@@ -3,6 +3,7 @@ use std::collections::VecDeque;
 use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
+use std::rc::Rc;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
@@ -24,10 +25,7 @@ use wasm_bindgen::prelude::*;
 ///
 /// Currently this type is constructed with `JsFuture::from`.
 pub struct JsFuture {
-    resolved: oneshot::Receiver<JsValue>,
-    rejected: oneshot::Receiver<JsValue>,
-    _cb_resolve: Closure<dyn FnMut(JsValue)>,
-    _cb_reject: Closure<dyn FnMut(JsValue)>,
+    rx: oneshot::Receiver<Result<JsValue, JsValue>>,
 }
 
 impl fmt::Debug for JsFuture {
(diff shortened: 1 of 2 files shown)


tomaka:
I'm trying this patch locally because I was getting these benign `closure invoked recursively ...` errors.

After this PR, I'm getting more serious errors:

Screenshot from 2019-07-10 11-15-32

None of the backtraces are useful. The lowest readable symbol of the backtrace is a function in the API that returns a `Promise`.

tomaka:
False alarm, I'm getting these errors on master as well compared to versions 0.2.47 and 0.3.24.

PR #1654, opened 2019-07-10
Skip generating JS import shims when unnecessary

### LINKED ISSUE #1636: Restore optimization to import functions directly
In the recent refactorings to get based on WebIDL bindings an optimization within wasm-bindgen-the-CLI was lost. In many cases we need a JS shim for imported functions, but in some cases we don't need any shim at all and can import the function directly. We should import the relevant function directly into the module if this is the case rather than creating a shim which simply forwards arguments.

It's worth noting as well that as wasm standards progress, we'll get more features for more types which makes this a more important optimization!

Additionally we should also apply a second optimization (which I don't think was ever done) which is to directly import the function from the wasm module if the imported path is a vanilla ES path, not requiring a JS shim for us to move, for example, a global JS type into a module's namespace.

fitzgen:
After this change, any import that only takes and returns ABI-safe numbers (signed integers less than 64 bits and unrestricted floating point numbers) will be a direct import, and will not have a little JS shim in the middle.

We don't have a great mechanism for testing the generated bindings' contents -- as opposed to its behavior -- but I manually verified that everything here does the Right Thing and doesn't have a JS shim:

```rust
#[wasm_bindgen]
extern "C" {
    fn trivial();

    fn incoming_i32() -> i32;
    fn incoming_f32() -> f32;
    fn incoming_f64() -> f64;

    fn outgoing_i32(x: i32);
    fn outgoing_f32(y: f32);
    fn outgoing_f64(z: f64);

    fn many(x: i32, y: f32, z: f64) -> i32;
}
```

Furthermore, I verified that when our support for emitting native `anyref` is enabled, then we do not have a JS shim for the following import, but if it is disabled, then we do have a JS shim:

```rust
#[wasm_bindgen]
extern "C" {
    fn works_when_anyref_support_is_enabled(v: JsValue) -> JsValue;
}
```

Fixes #1636.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2001,7 +2001,11 @@ impl<'a> Context<'a> {
                 if assert_no_shim {
                     panic!(
                         "imported function was annotated with `#[wasm_bindgen(assert_no_shim)]` \
-                         but we need to generate a JS shim for it"
+                         but we need to generate a JS shim for it:\n\n\
+                         \timport = {:?}\n\n\
+                         \tbinding = {:?}\n\n\
+                         \twebidl = {:?}",
+                        import, binding, webidl,
                     );
                 }
 
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1917,7 +1917,7 @@ impl<'a> Context<'a> {
         let js_doc = builder.js_doc_comments();
         let docs = format_doc_comments(&export.comments, Some(js_doc));
 
-        // Once we've got all the JS then put it in the right location dependin
+        // Once we've got all the JS then put it in the right location depending
         // on what's being exported.
         match &export.kind {
             AuxExportKind::Function(name) => {
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1,5 +1,6 @@
 use crate::descriptor::VectorKind;
 use crate::intrinsic::Intrinsic;
+use crate::webidl;
 use crate::webidl::{AuxEnum, AuxExport, AuxExportKind, AuxImport, AuxStruct};
 use crate::webidl::{AuxValue, Binding};
 use crate::webidl::{JsImport, JsImportName, NonstandardWebidlSection, WasmBindgenAux};
@@ -721,6 +722,15 @@ impl<'a> Context<'a> {
         // Accessing a heap object is just a simple index operation due to how
         // the stack/heap are laid out.
         self.global("function getObject(idx) { return heap[idx]; }");
+    }
+
+    fn expose_not_defined(&mut self) {
+        if !self.should_write_global("not_defined") {
+            return;
+        }
+        self.global(
+            "function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }"
+        );
     }
 
     fn expose_assert_num(&mut self) {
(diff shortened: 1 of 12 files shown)


alexcrichton:
Nice! 

This looks great to me, I'm gonna poke around with it a bit more locally and see what's what tomorrow.

crates/cli-support/src/js/mod.rs
```diff
@@ -1971,16 +1984,59 @@ impl<'a> Context<'a> {
...
             .get::<ast::WebidlFunction>(binding.webidl_ty)
             .unwrap();
-        let mut builder = binding::Builder::new(self);
-        builder.catch(catch)?;
-        let js = builder.process(&binding, &webidl, false, &None, &mut |cx, prelude, args| {
-            cx.invoke_import(&binding, import, bindings, args, variadic, prelude)
-        })?;
-        let js = format!("function{}", js);
+        let js = match import {
+            AuxImport::Value(AuxValue::Bare(js))
```
alexcrichton: I think that this may not be quite right in terms of matching where *technically* any value imported is a candidate for not requiring any glue. For example the other values, getters/setters, can all be hooked up directly in theory. They today, however, all require a `method` invocation style and will fail the `Static` check below very quickly, so it effectively ends up being the same. 

Perhaps though it'd be good to refactor this in a forward-facing way where we match `AuxImport::Value(v)` here and then have shared code between way below in here about how to turn `v` into a JS snippet?

alexcrichton:
Oh one other thought I had while reviewing this, in terms of tests it might be nift to assert that a JS shim is "meaningful"? Could we perhaps have an assert of some form when we generate a JS shim that it's somewhat nontrivial? That way we don't necessarily need to test this directly but we could rely on if that assert trips we forgot to handle something, but if the assert always passes then we're good to go. That would also help us cover future cases such as when anyref is on/off or WebIDL bindings get more fleshed out.

fitzgen:
> [functions returning boolean](https://github.com/rustwasm/wasm-bindgen/blob/c21c85a7ff1dc58873661a7e38ad6c25f4df7f86/tests/wasm/imports.rs#L53)

For incoming bools we can rely on the auto conversion turning this into `0` or `1`.

For outgoing bools that are *actually* going into a a Web IDL function, we can also rely on the auto conversion from a numbre to bool.

For outgoing bools that are going into random user JS functions, there won't be any auto conversions going on, and so we could pass the number `0` or `1` to functions that *might* be doing `typeof` reflection and could branch differently based on `1` vs `true`. Although, I suppose we already suffer from this today... I think ideally we would emit glue code to do `arg != 0` to convert from the `i32` into a boolean.

> [functions returning `u32`](https://github.com/rustwasm/wasm-bindgen/blob/c21c85a7ff1dc58873661a7e38ad6c25f4df7f86/tests/wasm/imports.rs#L33-L34)

I'm not sure if incoming unsigned numbers are safe. They get converted to signed integers by the JS Wasm interface, and although positive two's complement has the same representation as unsigned, I'm not sure about the edge cases here. Although, again, if our current behavior doesn't do anything, then I guess at most we are preserving bugs...

> [functions taking `u8`](https://github.com/rustwasm/wasm-bindgen/blob/c21c85a7ff1dc58873661a7e38ad6c25f4df7f86/tests/wasm/imports.rs#L31)

I guess we are assuming that the wasm-level `i32` won't be outside the range of the `u8` so this should be safe. I think I need to update some `wasm-webidl-bindings` methods to assume that outgoing wasm values are in the range of the Web IDL type that they claim to match.

fitzgen:
> For outgoing bools that are going into random user JS functions, there won't be any auto conversions going on, and so we could pass the number `0` or `1` to functions that _might_ be doing `typeof` reflection and could branch differently based on `1` vs `true`. Although, I suppose we already suffer from this today... I think ideally we would emit glue code to do `arg != 0` to convert from the `i32` into a boolean.

Heh... yeah actually this breaks if we don't have a shim that does `arg0 !== 0`:

```
---- wasm::no_shims::no_shims output ----
    error output:
        wasm-bindgen: imported JS function that was not marked as `catch` threw an error: assert_eq failed: 1 != true
        
        Stack:
        Error: assert_eq failed: 1 != true
            at assert_eq (/home/fitzgen/wasm-bindgen/target/wasm32-unknown-unknown/wbg-tmp/snippets/wasm-bindgen-3dff2bc911f0a20c/inline0.js:4:19)
            at module.exports.outgoing_bool (/home/fitzgen/wasm-bindgen/target/wasm32-unknown-unknown/wbg-tmp/snippets/wasm-bindgen-3dff2bc911f0a20c/inline0.js:35:51)
            at wasm::no_shims::outgoing_bool::h2a763567395adc3a (wasm-function[5492]:74)
            at wasm::no_shims::no_shims::hdbc99c282a7dac93 (wasm-function[349]:3016)
```

alexcrichton:
Oh sorry so in general the functions I gisted above all had a shim in JS but the shim didn't actually do anything, so I figure that we just want to optimize them to ensure that the shim is gone. That's also where I think it'd help to have an assert that if we generate a shim it's nontrivial in one way or another.

For booleans for example yeah for outgoing booleans we need a shim, but for incoming booleans I think it's a missing statement [here](https://github.com/rustwasm/wasm-webidl-bindings/blob/1bb746fdd7492a196b4c0af39a234079cb9d7ab0/src/ast.rs#L1325-L1382) which catches booleans going to i32.

> Although, again, if our current behavior doesn't do anything, then I guess at most we are preserving bugs...

I think this actually works out for incoming u32. It looks like the coercion to `i32` in wasm recognizes that it's big and positive so it fits all the bits. I made a roundtrip function in wasm (takes i32 returns the same) and passing in `4294967295` spits out `-1` on the other end. (according to JS).

In that sense I think we're actually covered for `u32` here and I don't think there's existing bugs? (not that I'm confident)

alexcrichton:
If it's possible I think it's still worthwhile to have a default always-on assert for all generated bindings that something is nontrivial about the JS shim that we generate. That I think is a much stronger assert than an annotation b/c we're still playing whack-a-mole with the annotation. I think it's of course though fine to add this stronger assert later.

fitzgen:
> If it's possible I think it's still worthwhile to have a default always-on assert for all generated bindings that something is nontrivial about the JS shim that we generate. That I think is a much stronger assert than an annotation b/c we're still playing whack-a-mole with the annotation.

It is stronger, yes.

However, I am a bit less confident in our ability to live up to its guarantees. By adding an assertion, we are explicitly saying that we expect these incoming and outgoing types not to require glue and that we are confident that we can guarantee that forever. I'm less certain about catching every single case that might happen not to require glue for every combination of bindings configurations and types that we support.

Additionally, implementing the check seems like it would have to be fairly ad-hoc unless we change the way the builders work a bit, and I'm not certain that the assertion itself wouldn't be super buggy.

alexcrichton:
Heh good points as well :)

In any case let's stick with a tag for now, and maybe add the tag to a more exhaustive set of types going in/out as well?

fitzgen:
> In any case let's stick with a tag for now, and maybe add the tag to a more exhaustive set of types going in/out as well?

yeah working on it now

fitzgen:
Ok, I think I got everything -- want to take another look @alexcrichton?

alexcrichton:
r=me on everything here, but I think [this line](https://github.com/rustwasm/wasm-webidl-bindings/blob/fabb287f69d4af92d14aca3e3b2c1e0566a7f4b5/src/ast.rs#L1203) may be wrong? While I think 0/1 have the same "truthy value" as `true` and `false` I think we still want a wrapper to generate honest-to-god boolean values?

fitzgen:
> r=me on everything here, but I think [this line](https://github.com/rustwasm/wasm-webidl-bindings/blob/fabb287f69d4af92d14aca3e3b2c1e0566a7f4b5/src/ast.rs#L1203) may be wrong?

Fixed! That was supposed to only be for incoming expressions >_<

PR #1665, opened 2019-07-13
Add Element::get_elements_by_class_name

ia0:


alexcrichton:
Thanks for this! While you're at it, mind removing the comment markers entirely from this block with this method? The comments AFAIK were added quite some time ago when web-sys was first being developed, but they shouldn't be necessary any more and we should implement all the methods!

ia0:
Sure, no problem!

alexcrichton:
👍

PR #1670, opened 2019-07-16
Check for use-after-move in JS glue when `--debug` is enabled again

### LINKED ISSUE #1669: Regression: methods that take `self` by value no longer assert against use-after-move with --debug
This used to generate "attempt to use moved value" checks in the glue for `MyThing::take`, but no longer does:

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct MyThing {}

#[wasm_bindgen]
impl MyThing {
    pub fn new() -> MyThing {
        MyThing {}
    }

    pub fn take(self) {}
}
```

Free functions do still generate the check. Will dig in a little bit.

See also: https://github.com/rustwasm/wasm-pack/issues/677

fitzgen:
Fixes #1669

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -138,15 +138,20 @@ impl<'a, 'b> Builder<'a, 'b> {
             // method, so the leading parameter is the this pointer stored on
             // the JS object, so synthesize that here.
             match self.method {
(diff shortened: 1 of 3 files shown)


alexcrichton:
👍 sorry for the regression!

PR #1673, opened 2019-07-18
Adding in PartialEq and Eq to web-sys types

### LINKED ISSUE #1672: impl PartialEq for web-sys
### Motivation

PartialEq is implemented for things in js_sys: https://github.com/rustwasm/wasm-bindgen/pull/1444

However, we still can't check for equality between `web_sys::Element`'s for example.

### Proposed Solution

Not sure if this is automatic via a single `derive` or in a macro or what, sorry! But on Discord @Pauan indicated that it's somehow doable.

### Additional Context

For a use case - we currently can't check if `Document.pointerLockElement` is equal to the element that initiated getting pointer lock. Workarounds are possible (like setting an id and comparing that), but it's not idiomatic and it would be much nicer to have direct equality checks.

Pauan:
Fixes #1672

### DIFF
--- a/crates/web-sys/tests/wasm/element.rs
+++ b/crates/web-sys/tests/wasm/element.rs
@@ -14,6 +14,8 @@ fn element() {
     */
     let element = new_div();
 
+    assert_eq!(element, element);
+
     assert_eq!(element.prefix(), None, "Shouldn't have a prefix");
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #1675, opened 2019-07-19
Update 'threads-xform' for LLVM 9

### LINKED ISSUE #1631: Updating LLD to use bulk memory for threads
In the near future I plan to update LLD to emit passive segments when a shared memory is requested. This will prevent memory from being reinitialized every time a thread is spawned. Currently toolchains need to do something like separate the contents of memory into a separate file and manually initialize memory on the main thread. With this update to LLD, the linker will synthesize a function called `__wasm_init_memory`. This function will automatically be called at the beginning of `__wasm_call_ctors`, but @fitzgen reminded me that Rust does not have any global constructors, so you will probably be able to call `__wasm_init_memory` directly.

The existing memory separation code will probably break since the data segments will no longer declare their own offsets, so I just wanted to give a heads up that this change is coming. It also means that enabling shared memory in Firefox will be insufficient to test multithreaded modules, which may affect your workflow or testing. I filed a Firefox bug about this at https://bugzilla.mozilla.org/show_bug.cgi?id=1561747.

I will update this issue once the change has landed upstream in LLD and when I update the tool-conventions repo with details about this arrangement.

alexcrichton:
In LLVM 9 LLD has been updated to emit shared memory and passive
segments by default for threaded code, and `__wasm_init_memory` is a
function exported used to initialize memory. Update our
transform/runtime here to hook up all those wires correctly.

Closes #1631

### DIFF
--- a/crates/threads-xform/src/lib.rs
+++ b/crates/threads-xform/src/lib.rs
@@ -78,25 +78,57 @@ impl Config {
     ///
     /// More and/or less may happen here over time, stay tuned!
     pub fn run(&self, module: &mut Module) -> Result<(), Error> {
-        let memory = update_memory(module, self.maximum_memory)?;
-        let segments = switch_data_segments_to_passive(module, memory)?;
         let stack_pointer = find_stack_pointer(module)?;
-
+        let memory = find_memory(module)?;
+        let addr = inject_thread_id_counter(module, memory)?;
         let zero = InitExpr::Value(Value::I32(0));
         let globals = Globals {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Note that this doesn't actually update the compiler used to produce the raytrace-parallel example just yet. This will require rust-lang/rust changes at https://github.com/rust-lang/rust/pull/62809 to make it so it's not too burdensome to produce it here, but once that lands only a small patch will be necessary to update the example's nightly compiler.

PR #1683, opened 2019-07-29
Improve leniency of `--no-modules` output

alexcrichton:
Instead of assuming names like `URL` and `Request` are defined, instead
check to see if they exist first and otherwise skip the checks that
reference them.


alexcrichton:
cc https://github.com/rustwasm/wasm-pack/issues/689, the original motivator

Issue #1662, opened 2019-07-13
Date methods missing support for negative arguments


portstrom:
Methods of the `Date` object such as [`setDate`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate) officially allow negative arguments, but this isn't currently supported by web-sys.

cburgos:
Is this still available? If so I would like to take a stab at it. Initially my thinking is this is just a type update for the affected methods which should be straightforward but a breaking change.

Issue #1664, opened 2019-07-13
Date constructors with more than one argument missing


portstrom:
The [`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) constructor takes up to seven arguments. web-sys currently only supports zero and one arguments.

alexcrichton:
Added in https://github.com/rustwasm/wasm-bindgen/pull/1759!

PR #1684, opened 2019-07-30
Bugfix/date arg types

cburgos:
Addresses #1662

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1978,7 +1978,7 @@ extern "C" {
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate)
     #[wasm_bindgen(method, js_name = setDate)]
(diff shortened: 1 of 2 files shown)


cburgos:
U

portstrom:
Notice that other methods such as `setHours` also need to allow negative arguments.

cburgos:
Noted, going through and updating the rest of the methods. Update should be ready shortly.

alexcrichton:
Thanks! FWIW as-is this is a breaking change so we won't be able to merge this quite just yet, we're still in the process of batching breaking changes. If you'd prefer though perhaps new methods with new names could be added, and then in the next major version release we can delete all the methods and change the existing signatures?

cburgos:
Added the other updates. @alexcrichton I think holding off and batching these with other breaking changes makes sense. Also worth noting the above commits didn't include the setFullYear and setUTCFullYear as support for optional parameters have not yet been implemented. It might make more sense to include that with #1664 for overall optional param support.

alexcrichton:
Ok sounds good to me! I'll tag this as a breaking change and we'll merge this when we get to doing the next round of breaking changes.

PR #1692, opened 2019-08-01
Fix a number of `anyref` related bugs in our passes

alexcrichton:
This commit goes through and fixes a number of bugs around the `anyref` pass in `wasm-bindgen`. Most of the issues were around initialization and tests, but after this commit the full test suite passes in Node.js. Yay!

### DIFF
--- a/crates/cli-support/src/intrinsic.rs
+++ b/crates/cli-support/src/intrinsic.rs
@@ -145,7 +145,7 @@ intrinsics! {
         #[symbol = "__wbindgen_anyref_heap_live_count"]
         #[signature = fn() -> I32]
         AnyrefHeapLiveCount,
-        #[symbol = "__wbindgen_init_nyref_table"]
+        #[symbol = "__wbindgen_init_anyref_table"]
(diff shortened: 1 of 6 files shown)


crates/cli-support/src/intrinsic.rs
```diff
@@ -145,7 +145,7 @@ intrinsics! {
         #[symbol = "__wbindgen_anyref_heap_live_count"]
         #[signature = fn() -> I32]
         AnyrefHeapLiveCount,
-        #[symbol = "__wbindgen_init_nyref_table"]
```
fitzgen: lolsob...

fitzgen:
So now that node has proper `anyref` support, can we actually run our test suite (or a subset of it) with anyref enabled in CI?

alexcrichton:
Soon we can I think! The support *just* (like 24 hours ago) landed on the master branch of nodejs, so I think we'll need to wait for the next release and then we should be able to download a recent verison of Node to test with.

PR #1697, opened 2019-08-03
fix instiatestreaming fallback

### LINKED ISSUE #1696: JS InstantiateStreaming fallback does not work
### Describe the Bug

In the generated JS file there's a fallback for when `InstantiateStreaming` doesn't work. This code will never succeed because the `fetch` body has already been consumed:
https://github.com/rustwasm/wasm-bindgen/blob/732b691bb85dbcba23c31482733b69b484a2a3c2/crates/cli-support/src/js/mod.rs#L538-L546

### Steps to Reproduce

I've got a bad wasm setup right now, and not sure why it's bad (I do have the mime set to `application/wasm`, but yeah, not sure what's going on).

### Expected Behavior

The fallback should be working.

### Actual Behavior

The fallback breaks because the request body is already consumed. Not working on both Chrome and Firefox.

__chrome__
2019-08-04-000620_1920x1080

__firefox__
2019-08-04-000630_1920x1080

### Additional Context

I think the way to fix it is by sending out another `fetch` request. It's wasteful, but it's probably better than straight up throwing.

Also I should figure out how to make streaming work from my server, but that's on me, haha. Thanks!

yoshuawuyts:
Closes #1696. Thanks!

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -541,7 +541,7 @@ impl<'a> Context<'a> {
                                                     `application/wasm` MIME type. Falling back to \
                                                     `WebAssembly.instantiate` which is slower. Original \
                                                     error:\\n\", e);
-                                    return response
+                                    return fetch(module)
                                         .then(r => r.arrayBuffer())
                                         .then(bytes => WebAssembly.instantiate(bytes, imports));
                                 }});


alexcrichton:
Thanks! I think this may not be the solution we want to commit to though since it implies a second network request. I'm hoping there's a way we could perhaps inspect the error and not try the fallback if the error isn't related to the mime type

yoshuawuyts:
@alexcrichton cool, yeah that makes sense! Going to go ahead and close this then (:

PR #1699, opened 2019-08-04
Use idiomatic syntax of Rc::clone()

UnHumbleBen:
This syntax makes it easier to see that this code is creating a new reference rather than copying the whole content of `sched`.


alexcrichton:
Thanks!

PR #1702, opened 2019-08-04
Fixes description of Closure lifetime in comments.

### LINKED ISSUE #1701: web-sys: Closures example contains inaccurate comment
The description of the lifetime of `Closure` is incorrect.
```
fn setup_clock(window: &Window, document: &Document) -> Result<(), JsValue> {
    // --snip--

    let a = Closure::wrap(Box::new(move || update_time(&current_time)) as Box<dyn Fn()>);
    window

    // --snip--

    // The instances of `Closure` that we created will invalidate their
    // corresponding JS callback whenever they're dropped, so if we were to
    // normally return from `run` then both of our registered closures will
    // raise exceptions when invoked.
    //
    // Normally we'd store these handles to later get dropped at an appropriate
    // time but for now we want these to be global handlers so we use the
    // `forget` method to drop them without invalidating the closure. Note that
    // this is leaking memory in Rust, so this should be done judiciously!
    a.forget();

    // --snip--

}
```
The comment claims that `a` will be dropped when `run` returns; however, `a` is declared inside of `setup_clock`, so `a` is actually dropped at the end of `setup_clock`, not `run`.

UnHumbleBen:
Fixes #1701

### DIFF
--- a/examples/closures/src/lib.rs
+++ b/examples/closures/src/lib.rs
@@ -72,14 +72,14 @@ fn setup_clock(window: &Window, document: &Document) -> Result<(), JsValue> {
         ));
     }
 
-    // The instances of `Closure` that we created will invalidate their
-    // corresponding JS callback whenever they're dropped, so if we were to
-    // normally return from `setup_clock` then both of our registered closures will
-    // raise exceptions when invoked.
+    // The instance of `Closure` that we created will invalidate its
+    // corresponding JS callback whenever it is dropped, so if we were to
+    // normally return from `setup_clock` then our registered closure will
+    // raise an exception when invoked.
     //
-    // Normally we'd store these handles to later get dropped at an appropriate
-    // time but for now we want these to be global handlers so we use the
-    // `forget` method to drop them without invalidating the closure. Note that
+    // Normally we'd store the handle to later get dropped at an appropriate
+    // time but for now we want it to be a global handler so we use the
+    // `forget` method to drop it without invalidating the closure. Note that
     // this is leaking memory in Rust, so this should be done judiciously!
     a.forget();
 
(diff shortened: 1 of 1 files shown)


examples/closures/src/lib.rs
```diff
@@ -74,7 +74,7 @@ fn setup_clock(window: &Window, document: &Document) -> Result<(), JsValue> {
 
     // The instances of `Closure` that we created will invalidate their
     // corresponding JS callback whenever they're dropped, so if we were to
-    // normally return from `run` then both of our registered closures will
+    // normally return from `setup_clock` then both of our registered closures will
     // raise exceptions when invoked.
```
limira: I think the comment is from the first version of the example when everything was put in a single `fn run`. Now it is split into several `fn`s, and you want to fix them, then I think other places should be fixed too. The comment is now applied to a single instance of Closure `a` at L66, then things should be fixed, e.g.: "instances" (become "instance"), "their", "they"... and finally add a note to `setup_clicker` (below) say that people may want to see why `a.forget()` as notes in `setup_clock` above.

UnHumbleBen: @limira I added the changes you suggested. I was not sure about whether to change "exceptions" to "exception". I am assuming that a single closure can throw at most one exception?

limira: To me it's good enough. But others may provide a more precise comments.

UnHumbleBen:
@limira Any ideas why the 2 checks are failing? It seems that other recent pull requests are failing these 2 checks as well.

limira:
> @limira Any ideas why the 2 checks are failing? It seems that other recent pull requests are failing these 2 checks as well.

You can follow the links in the test results to find out why. I follow one of them and it say that:

    error: Could not compile `unicode-xid`.

`unicode-xid` is a crate outside of `wasm-bindgen`, and may be it need to be upgraded.

alexcrichton:
Thanks for this! The failures are unrelated to this PR and I'll be debugging them soon hopefully

PR #1707, opened 2019-08-06
Removed unnecessary and confusing clone of String

### LINKED ISSUE #1705: How can JsValue implement From<&'a String>?
### Summary

I am confuse about whether calling `from` implemented for `From<&'a String>` ultimately moves the `String` or not. Does it create a clone?

### Additional Details

I was looking at the todomvc example and notice 2 peculiar lines of code
```
            let s = item.title.clone();
            child.push(&JsValue::from(&s));
```
`item.title` is a `String` and `child` is a `js_sys::Array`.

This code attempts to clone the String before it is passed into `from`, but this is unnecessary since `from` just borrows the string slice. So I tried to simplify the code as follows
```
            let s = &item.title;
            child.push(&JsValue::from(s));
```
I took out the clone and there are no compile errors!

 I am confused about how this could work because surely a clone needs to take place somewhere for `child` to `push` a new `JsValue`. Perhaps the trait `From<&'a String>` is implemented in a way such that `clone` is called for you? In which case, the code example's call to `clone` is unnecessary. Is there something I am not seeing here? Perhaps it will lead to a runtime error?

UnHumbleBen:
Fixes #1705

### DIFF
--- a/examples/todomvc/src/store.rs
+++ b/examples/todomvc/src/store.rs
@@ -60,8 +60,7 @@ impl Store {
         let array = js_sys::Array::new();
         for item in self.data.iter() {
             let child = js_sys::Array::new();
-            let s = item.title.clone();
-            child.push(&JsValue::from(&s));
+            child.push(&JsValue::from(&item.title));
             child.push(&JsValue::from(item.completed));
             child.push(&JsValue::from(item.id.to_string()));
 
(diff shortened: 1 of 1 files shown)


UnHumbleBen:
Please excuse the typo in the commit message for https://github.com/rustwasm/wasm-bindgen/pull/1707/commits/fffcc5ef35d4c109b1f20712a4a5af40c1f2563d
Should have been "Makes code clearer" not "Make code clearing"

alexcrichton:
👍 Thanks!

Issue #1694, opened 2019-08-02
Support USB API


MarcAntoine-Arnaud:
Is it possible to support Web USB API ?
Examples of required features can be found here: https://developers.google.com/web/updates/2016/03/access-usb-devices-on-the-web

Only chrome as implemented that API, but it can be great to expose it.

alexcrichton:
Thanks for the report! We currently support stage 3 and above proposals and we support them by copying in the official WebIDL file into our webidl folder, do you know what stage this proposal is at and where it's WebIDL can be found?

MarcAntoine-Arnaud:
Thanks for the reply.
And I learn something, I know nothing about IDL, but right now more ;-)

I have found https://wicg.github.io/webusb/#idl-index (very recent update...)

MarcAntoine-Arnaud:
Do you think it can be safe to add that one:
https://github.com/tidoust/reffy-reports/blob/master/whatwg/idl/webusb.idl ?

alexcrichton:
I'm going to close this in favor of https://github.com/rustwasm/wasm-bindgen/issues/1950 which is a general strategy of how to support unstable WebIDL files.

PR #1712, opened 2019-08-08
add WebUSB WebIDL

MarcAntoine-Arnaud:
issue #1694


alexcrichton:
Thanks for this! We in general only add stage 3 and beyond proposals to this repository. I can't seem to find documentation, but do you know at what stage in the standards process this proposal is at?

MarcAntoine-Arnaud:
I don't know where we can find the stage of the proposal...
But this patch don't work, I cannot access to USB feature (in the documentation generated with: `cargo doc --all-features`. Do you have any idea why ?

Pauan:
@alexcrichton I don't know what stage it is, but it is marked as [Experimental](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Conventions_definitions#Experimental) on MDN, and only Chrome and Opera have support for it right now.

Pauan:
I wonder if maybe we should have support for unstable APIs, perhaps behind an `unstable-api` feature or something like that.

And we would reserve the right to break anything behind `unstable-api` at any time, for any reason, even in minor releases.

alexcrichton:
Ok so it sounds like we shouldn't add this as-is because it's not an official spec yet, but I like the idea of having experimental specs here that we can break at any time. Prior art for that I'm aware of isn't to use Cargo features but rather `RUSTFLAGS=--cfg=web_sys_unstable_apis` or something like that.

That way it's a bit more experimental in that it forces consumers to always compile with the right flags. How's that sound to enable this?

Pauan:
> How's that sound to enable this?

That sounds reasonable to me. It means that libraries will have a harder time relying upon unstable features, so they will be mostly used by applications, which is how it should be.

MarcAntoine-Arnaud:
@Pauan @alexcrichton thanks for comments.
I'm agreed it make sense to put it in unstable api.

alexcrichton:

alexcrichton:
FWIW I've opened https://github.com/rustwasm/wasm-bindgen/issues/1950 to track longer-term about how we plan to support unstable WebIDL files.

PR #1717, opened 2019-08-12
Update with list IR from `walrus` 

alexcrichton:
This commit updates `wasm-bindgen` to the latest version of `walrus`
which transforms all internal IR representations to a list-based IR
instead of a tree-based IR. This isn't a major change other than
cosmetic for `wasm-bindgen` itself, but involves a lot of changes to the
threads/anyref passes.

This commit also updates our CI configuration to actually run all the
anyref tests on CI. This is done by downloading a nightly build of
node.js which is theorized to continue to be there for awhile until the
full support makes its way into releases.


alexcrichton:
Ok should be good for review! Current CI failure is a fluke, and otherwise I'd still like to land this without `[patch]` so I've started that at https://github.com/rustwasm/walrus/pull/117 to kick off the process.

PR #1728, opened 2019-08-20
Update browser support for `BigInt`

nasso:
`BigInt` is supported in Firefox as of Firefox 68!

Edge doesn't support it for now but it will when EdgeHTML will be replaced with Chromium.
I replaced Firefox with Safari as a second example of an unsupported browser (I still swapped them because "Edge or Safari" sounds better than "Safari or Edge" to me :p).

As a further improvement (not included here), we could add a link to the `Can I use...` support table for `BigInt`.


alexcrichton:
Nice! I'd be totally down for a link to [caniuse.com](https://caniuse.com/#feat=bigint), want to include that here?

nasso:
Sure!

PR #1737, opened 2019-08-27
Prototype inheritance

### LINKED ISSUE #210: Add ability to import and extend JS classes
JS classes are heavily used building block in UI development.

It's common for a library like ReactJS or Polymer to provide a component class to extend from.

Thus, there should be an easy way to interoperate with classes, extend them in Rust, and export them back to JS.

I don't know about Rust to know what's possible yet. But I imagine we could use attributes to annotate when a Rust struct inherits from a JS class?

### LINKED ISSUE #1721: Enable exported Rust structs to specify prototypal inheritance
### Motivation

Because the JS classes generated for exported Rust structs do not currently specify a parent class, they are implicit subclasses of `Object`—that is, their prototype objects are `Object.prototype`.  However, there are many situations where the parent class (prototype object) of one's exported structs should be something other than this default.

### Proposed Solution

I propose a new `super = Type` attribute for the `wasm_bindgen` macro, as follows:

```
#[wasm_bindgen]
pub struct Parent { ... }
#[wasm_bindgen(super=Parent)]
pub struct Child { ... }

// the superclass might be imported from JS
#[wasm_bindgen]
extern "C" {
    pub type ImportedParent;
}
#[wasm_bindgen(super=ImportedParent)]
pub struct ChildOfImportedParent { ... }

// or might be a built-in published by the `js_sys` crate
#[wasm_bindgen(super=js_sys::Date)]
pub struct CustomDate { ... }

// or might be a Web IDL published by the `web_sys` crate
#[wasm_bindgen(super=web_sys::XmlHttpRequest)]
pub struct MyAjaxRequest { ... }
```

In each case, the generated JavaScript class would include the relevant `extends` clause.

eggyal:
Prototype implementation for [draft RFC](https://github.com/rustwasm/rfcs/pull/11).  Fixes #210 and fixes #1721.

Enables one to do stuff like this—

```rust
#[wasm_bindgen(prototype=web_sys::HtmlElement)]
struct MyCustomElement {}

#[wasm_bindgen]
impl MyCustomElement {
    #[wasm_bindgen(constructor)]
    fn new() -> WasmType<MyCustomElement> {
        instantiate! { MyCustomElement{} }
    }
}

// ...

web_sys::window()
    .unwrap()
    .custom_elements()
    .define("my-custom-element", &js_sys::JsFunction::of::<MyCustomElement>())?;

// ...
```

See [Exporting a struct to JS](https://github.com/eggyal/wasm-bindgen/blob/prototype-inheritance/guide/src/contributing/design/exporting-rust-struct.md) for detailed information, and the [examples](https://github.com/eggyal/wasm-bindgen/tree/prototype-inheritance/examples) prefixed `web-components-` for use-cases.


crates/backend/src/codegen.rs
```diff
@@ -136,6 +138,17 @@ impl TryToTokens for ast::Program {
     }
 }
 
+fn gen_id() -> [u8; 8] {
+    use ::std::time::{SystemTime, UNIX_EPOCH};
+
+    let duration = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
+    
+    let mut bytes = duration.as_secs().to_be_bytes();
+    bytes[0..4].copy_from_slice(&duration.subsec_nanos().to_be_bytes());
```
eggyal: I think that the CI server is failing due to identifier collisions that are generated here: I don't see these when building locally, presumably because my machine is much slower and/or much less parallel than Azure such that it can generate two separate IDs in the same nanosecond whereas I cannot?  Or perhaps [`Duration::subsec_nanos`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.subsec_nanos) is returning 0 on Windows?

eggyal: Turns out that this is precisely what the issue that `backend::util::ShortHash` was built to resolve, however I still witnessed collisions when building `web_sys`: adding a timestamp to each package's (cached) hash seed solved that, but raises the question of whether the seed ought not just be entirely random?

alexcrichton:
I'm going to close this while the RFC makes progress to help clean out PRs.

Issue #1732, opened 2019-08-22
Connect to remote webdriver


Kirguir:
### Motivation

Have the ability to run tests on a remote webdriver.

### Proposed Solution

Pass in the environment variables (CHOMEDRIVER) URL of the remote webdriver. If the URL is passed, and not the PATH, do not start the webdriver locally, but connect to the remote one.

PR #1744, opened 2019-08-29
Add ability to run tests on remote webdriver.

### LINKED ISSUE #1731: Options for webdriver
### Motivation

Drivers for different browsers are configured differently - for the chromedriver "switches", which allow the use of fake media devices, are passed in arguments as a command line, for the geckodriver the same is done through preferences (about:config). 
```
let request = json!({
       "capabilities": {
            "alwaysMatch": {
                "moz:firefoxOptions": {
                    "prefs": {
                        "media.navigator.streams.fake": true,
                        "media.navigator.permission.disabled": true,
                    },
                    "args": args,
                }
            }
        }
    });
```
There is currently no way to pass preferences.

### Proposed Solution

Add the ability to pass preferences through environment variables

### Alternatives
Configure the client for the web driver through a file, where you can describe all the options - arguments and preferences.

Kirguir:
Resolve #1731, #1732

- added parsing env vars like `GECKODRIVER_REMOTE`  for address of remote webdriver
- removed parsing env vars like `GECKODRIVER_CLIENT_ARGS` 
- added parsing a file `webdriver.json` for configure browser's capabilities, like:
```
{
  "moz:firefoxOptions": {
    "prefs": {
      "media.navigator.streams.fake": true,
      "media.navigator.permission.disabled": true,
    },
    "args": []
  },
  "goog:chromeOptions": {
    "args": [
      "--use-fake-device-for-media-stream",
      "--use-fake-ui-for-media-stream"
    ]
  }
}
```


alexcrichton:
Thanks for this and sorry for the delay in reviewing! This is getting complicated enough that would you mind writing up a page in `guide/*` about all the various options to configure this? Also is the `webdriver` dependency here necessary, or could we perhaps continue to use raw json blobs using just serde?

Kirguir:
You can use a raw JSON blob, as shown in the example - each browser will only process its own section of options.
Of course I will add a configuration description to this [guaide](https://github.com/rustwasm/wasm-bindgen/blob/master/guide/src/wasm-bindgen-test/browsers.md).

alexcrichton:
Thanks! And I'm just thinking that the `webdriver` dependency isn't really buying us much since it's just a few structures, so would it be possible to remove the dependency on it? Additionally do you know if there's documentation for safari's supported options?

Kirguir:
For Safari, unfortunately, did not find anything. Do you mean remove webdriver from [here](https://github.com/rustwasm/wasm-bindgen/blob/f5f946721134da54f61f48bea43801092a152acb/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs#L260)?

alexcrichton:
Ah ok, seems fine to start out with chrome/firefox in any case. And yeah by removing `webdriver` I mean the crate used as a dependency.

alexcrichton:
Looks great to me, thanks!

PR #1746, opened 2019-08-29
WIP: Support WebIDL modules

### LINKED ISSUE #1740: Support for WebIDL modules?
Has there been a discussion about what it would take to modify `wasm-bindgen-webidl` to support potential WebIDL modules (see https://github.com/WICG/kv-storage/pull/68 and the issues mentioned there) or something that would provide similar functionality?

I've been experimenting with making webidls for at least parts of the node and electron apis (because doing it with wasm-bindgen is too much work) but it doesn't seem possible to do properly without namespaced interfaces or modules.

The `LegacyNamespace` attribute on interfaces could probably be used as a (temporary!) workaround in some cases, but `wasm-bindgen-webidl` doesn't support it, and adding support for it would require at least some of the same machinery as for modules.

I've been poking around the source for `weedle` and `wasm-bindgen-webidl` trying to get an idea for how they work to see if I could make the modifications myself and I think I'm starting to understand what's going on.

It looks like the changes could be extensive though since most of the machinery in `first_pass.rs` deals with names as `&str` (potentially within a non-nested namespace) and would need to be generalized to handle qualified names instead.

If there is interest in supporting this functionality now (either through experimental support for parsing webidl modules per the PR, or supporting `LegacyNamespace`, or some other interface) I'd be willing to try and implement it given some guidance.

freebroccolo:
Fixes #1740 (eventually); adds support for [WebIDL modules](https://pr-preview.s3.amazonaws.com/littledan/webidl/pull/675.html#idl-modules).

Related: rustwasm/weedle#31

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,3 +92,4 @@ wasm-bindgen = { path = '.' }
 wasm-bindgen-futures = { path = 'crates/futures' }
 js-sys = { path = 'crates/js-sys' }
 web-sys = { path = 'crates/web-sys' }
+weedle = { path = '../weedle' }
--- a/crates/webidl/Cargo.toml
+++ b/crates/webidl/Cargo.toml
@@ -20,4 +20,7 @@ proc-macro2 = "1.0"
 quote = '1.0'
 syn = { version = '1.0', features = ['full'] }
 wasm-bindgen-backend = { version = "=0.2.50", path = "../backend" }
-weedle = "0.10"
+
+[dependencies.weedle]
+git = "https://github.com/freebroccolo/weedle"
+branch = "webidl-modules"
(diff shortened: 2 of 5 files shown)


alexcrichton:
Looking reasonable to me! Could you expand a bit on the expected impact of web-sys as well?

freebroccolo:
@alexcrichton thanks! I will open a companion PR for `weedle` in a bit.

Regarding the expected impact on `web-sys`, I haven't thought too much about that yet. The goal is that code generation for webidls that don't use modules should remain unchanged. There should be opportunities to benefit from supporting modules going forward though.

The main motivation for this right now is to bring more flexibility for writing webidls for things like electron and node and other larger libraries that are difficult to fully cover with `wasm-bindgen-cli` and not currently possible to model accurately with `wasm-bindgen-webidl`.

If y'all have thoughts about how this might impact `web-sys` or other parts of the ecosystem, I'd be interested to here more about that.

freebroccolo:
I'm going to close this since I've put the parser rewrite on hold to focus on other things I'm not sure if/when I'll be able to get back to it. I still think it would be a good idea to support webidl module syntax at some point. Thanks for the feedback!

PR #1749, opened 2019-09-03
Adding in more methods for Array

Pauan:
Adds in the following methods:

* `new_with_length` to create an Array with a specific length
* `get` to retrieve the element at an index
* `set` to set the element at an index
* `delete` to delete the element at an index

Also adds a `FromIterator` impl, which means it's now easy to create an `Array` from an `Iterator` by using `collect`:

```rust
vec![ ... ].iter().collect::<js_sys::Array>()
```

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -130,9 +130,29 @@ extern "C" {
     #[derive(Clone, Debug, PartialEq, Eq)]
     pub type Array;
 
-    /// Creates a new empty array
+    /// Creates a new empty array.
     #[wasm_bindgen(constructor)]
     pub fn new() -> Array;
+
+    /// Creates a new array with the specified length (elements are initialized to `undefined`).
+    #[wasm_bindgen(constructor)]
+    pub fn new_with_length(len: u32) -> Array;
+
+    /// Retrieves the element at the index (returns `undefined` if the index is out of range).
+    #[wasm_bindgen(method, structural, indexing_getter)]
+    pub fn get(this: &Array, index: u32) -> JsValue;
+
+    /// Sets the element at the index (auto-enlarges the array if the index is out of range).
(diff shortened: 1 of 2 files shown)


Pauan:
The only controversial thing is the `FromIterator` impl. In particular, there's a few different ways it could be implemented:

* Accept `AsRef<JsValue>` (which is what this PR does)

* Accept `Into<JsValue>`

* Accept something else?

Using `Into<JsValue>` is good, it allows for some native types like `&str`, `u32`, etc. But it doesn't work for certain things, like `Closure`

Using `AsRef<JsValue>` is very good, but it does require the user to manually call `JsValue::from` for things like `&str`, `u32`, etc. This isn't a big deal, they can just do `.map(JsValue::from)`

Another issue with `AsRef<JsValue>` is that it doesn't (currently) work with `#[wasm_bindgen]` structs, because they don't implement `AsRef<JsValue>`. Is there any technical reason they can't impl `AsRef<JsValue>`?

alexcrichton:
Thanks! This all seems great to me, and I'm fine with `AsRef` chosen here. 

> Is there any technical reason they can't impl AsRef<JsValue>?

Currently they don't actually have the `JsValue` on-hand anywhere in the struct to implement `AsRef` with, we don't actually modify the struct we just generate a wrapper around the `struct`. I saw a comment from you about this on another thread though that I haven't gotten to, so I may be missing something as well.

Pauan:
> Currently they don't actually have the JsValue on-hand anywhere in the struct to implement AsRef with, we don't actually modify the struct we just generate a wrapper around the struct.

Ahhh, right. So in that case if a user wants to send a struct they'll have to use `.map(JsValue::from)`, just like with native types. That's fine, after all Rust structs *are* a native type.

alexcrichton:
👍

PR #1754, opened 2019-09-05
Add support for `#[wasm_bindgen]` on `async fn`

alexcrichton:
This commit adds support to attach `#[wasm_bindgen]` on an `async fn`
which will change the return value into a `Promise` in JS. This in
theory has the exact same semantics as an `async` function in JS where
you call it with all the arguments, nothing happens and you get a
promise back, and then later the promise actually resolves.

This commit also adds a helper trait, `IntoJsResult`, to allow `async`
functions with multiple kinds of return values instead of requiring
everything to be `Result<JsValue, JsValue>`.

### DIFF
--- a/guide/src/reference/js-promises-and-rust-futures.md
+++ b/guide/src/reference/js-promises-and-rust-futures.md
@@ -1,4 +1,76 @@
-# Converting Between JavaScript `Promise`s and Rust `Future`s
+# Working with a JS `Promise` and a Rust `Future`
+
+Many APIs on the web work with a `Promise`, such as an `async` function in JS.
+Naturally you'll probably want to interoperate with them from Rust! To do that
+you can use the `wasm-bindgen-futures` crate as well as Rust `async`
+functions.
+
+The first thing you might encounter is the need for working with a `Promise`.
+For this you'll want to use [`js_sys::Promise`]. Once you've got one of those
+values you can convert that value to `wasm_bindgen_futures::JsFuture`. This type
+implements the `std::future::Future` trait which allows naturally using it in an
+`async` function. For example:
+
+[`js_sys::Promise`]: https://docs.rs/js-sys/*/js_sys/struct.Promise.html
+
+```rust
+async fn get_from_js() -> Result<JsValue, JsValue> {
+    let promise = js_sys::Promise::resolved(&42.into());
+    let result = wasm_bindgen_futures::JsFuture::from(promise).await?;
+    Ok(result)
+}
+```
+
+Here we can see how converting a `Promise` to Rust creates a `impl Future<Output
+= Result<JsValue, JsValue>>`. This corresponds to `then` and `catch` in JS where
+a successful promise becomes `Ok` and an erroneous promise becomes `Err`.
+
+Next up you'll probably want to export a Rust function to JS that returns a
+promise. To do this you can use an `async` function and `#[wasm_bindgen]`:
+
+```rust
+#[wasm_bindgen]
+pub async fn foo() {
+    // ...
+}
+```
+
+When invoked from JS the `foo` function here will return a `Promise`, so you can
+import this as:
+
+```js
+import { foo } from "my-module";
+
+async function shim() {
+    const result = await foo();
+    // ...
+}
+```
+
+## Return values of `async fn`
+
+When using an `async fn` in Rust and exporting it to JS there's some
+restrictions on the return type. The return value of an exported Rust function
+will eventually become `Result<JsValue, JsValue>` where `Ok` turns into a
+successfully resolved promise and `Err` is equivalent to throwing an exception.
+
+The following types are supported as return types from an `async fn`:
+
+* `()` - turns into a successful `undefined` in JS
+* `T: Into<JsValue>` - turns into a successful JS value
+* `Result<(), E: Into<JsValue>>` - if `Ok(())` turns into a successful
+  `undefined` and otherwise turns into a failed promise with `E` converted to a
+  JS value
+* `Result<T: Into<JsValue>, E: Into<JsValue>>` - like the previous case except
+  both data payloads are converted into a `JsValue`.
+
+Note that many types implement being converted into a `JsValue`, such as all
+imported types via `#[wasm_bindgen]` (aka those in `js-sys` or `web-sys`),
+primitives like `u32`, and all exported `#[wasm_bindgen]` types. In general,
+you should be able to write code without having too many explicit conversions,
+and the macro should take care of the rest!
+
+## Using `wasm-bindgen-futures`
 
 The `wasm-bindgen-futures` crate bridges the gap between JavaScript `Promise`s
 and Rust `Future`s. Its `JsFuture` type provides conversion from a JavaScript
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -213,6 +213,7 @@ pub struct Function {
     pub ret: Option<syn::Type>,
     pub rust_attrs: Vec<syn::Attribute>,
     pub rust_vis: syn::Visibility,
+    pub r#async: bool,
 }
 
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -442,13 +442,30 @@ impl TryToTokens for ast::Export {
         if let syn::Type::Reference(_) = syn_ret {
             bail_span!(syn_ret, "cannot return a borrowed ref with #[wasm_bindgen]",)
         }
-        let ret_ty = quote! {
-            -> <#syn_ret as wasm_bindgen::convert::ReturnWasmAbi>::Abi
+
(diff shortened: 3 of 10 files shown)


guide/src/reference/js-promises-and-rust-futures.md
```diff
@@ -1,4 +1,76 @@
...
+values you can convert that value to `wasm_bindgen_futures::JsFuture`. This type
+implements the `std::future::Future` trait which allows naturally using it in an
+`async` function. For example:
+
+[`js_sys::Promise`]: https://docs.rs/js-sys/*/js_sys/struct.Promise.html
+
+```rust
+async fn get_from_js() -> JsValue {
+    let promise = js_sys::Promise::resolved(&42.into());
+    let future = wasm_bindgen_futures::JsFuture::from(promise);
```
Pauan: I think we should be encouraging best practices here, so I think it should be like this instead:

```rust
async fn get_from_js() -> Result<JsValue, JsValue> {
    let promise = js_sys::Promise::resolved(&42.into());
    let result = wasm_bindgen_futures::JsFuture::from(promise).await?;
    Ok(result)
}
```

alexlapa:
@alexcrichton ,

Is this a known issue?

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct Foo{

}
#[wasm_bindgen]
impl Foo {
    pub async fn bar(&self) -> Result<i32,i32> {
        Ok(42)
    }
}
```

```
error[E0597]: `me` does not live long enough
 --> src/lib.rs:8:1
  |
8 | #[wasm_bindgen]
  | ^^^^^^^^^^^^^^-
  | |             |
  | |             `me` dropped here while still borrowed
  | argument requires that `me` is borrowed for `'static`

```

```toml
[dependencies]
wasm-bindgen = "0.2.51"
wasm-bindgen-futures = "0.4.1"
```

alexcrichton:
@alexlapa although not a great error message, that's a legitimate error message because futures shared with JS need to have the `'static` lifetime unfortunately, which means `&self` can't work

alexlapa:
@alexcrichton ,

I see. Is it possible to make functions like this valid?

```
#[wasm_bindgen]
impl Foo {
    pub fn bar(&self) -> impl Future<Output=Result<Foo, JsValue>> + 'static {
        async {
            Ok(Foo {})
        }
    }
}
```

Or to add some kind of `#[wasm_bindgen(async_static)]` that would generate all necessary glue?

alexcrichton:
@alexlapa that should work yeah!

For that though you'll have to return `-> js_sys::Promise` and use `wasm_bindgen_futures::future_to_promise` directly.

alexcrichton:
As for automatically doing glue, I'm not entirely sure how it would work, but PRs and ideas are always welcome!

Pauan:
@alexlapa That isn't an issue with wasm-bindgen, it's an issue with how Rust Futures work in general:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=30a32014c0d66fb5e6cccb5d5bb5b541

Because Futures are heap allocated and run in the background, they cannot contain stack references. And `&self` is a stack reference.

Instead, you should take `self` by value:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=31f05428313900f131e4eaaa5fdf1daf

If you need to access `self` in multiple places, then `Rc<Foo>` is the way to do that. And if you need to mutate `Foo`, then `Rc<RefCell<Foo>>` is the way to do that.

alexlapa:
@Pauan ,

Thanks for detailed explanation. So there is no way to `#[wasm_bindgen] async fn foo(&self) -> _`, and i have to use promises. Huge inconvenience is that promises erase concrete types which harms readability and testability. E.g.:

```
#[wasm_bindgen]
pub struct Foo;

#[wasm_bindgen]
impl Foo {
    pub fn new() -> Promise {
        future_to_promise(async {
            Ok(Foo.into())
        })
    }

    pub fn i_want_to_test_this_function(&self) -> bool {
        true
    }
}

#[wasm_bindgen_test]
async fn test() {
    let foo = JsFuture::from(Foo::new()).await.unwrap();

    assert!(foo.i_want_to_test_this_function()); // nope, not gonna happen
}
```

It would be awesome if `impl Future<_> + 'static` would be valid return type for exported functions, and `future_to_promise` conversion would happen in generated code in such case.

I can implement this feature, if it makes sense :)

Pauan:
> @alexlapa Huge inconvenience is that promises erase concrete types which harms readability and testability.

That's fairly easy to workaround by simply having two methods:

```rust
impl Foo {
    pub async fn new_() -> Result<Foo, JsValue> {
        Ok(Foo)
    }
}

#[wasm_bindgen]
impl Foo {
    pub fn new() -> Promise {
        future_to_promise(Self::new_())
    }
}
```

Now you can test the `new_` method in Rust (with proper typing), while still exposing a Promise to JS.

This is a common technique when using wasm-bindgen, because sometimes you need to do some type conversions when sending to JS, so that's what the above allows you to do.

> It would be awesome if impl Future<_> + 'static would be valid return type for exported functions

That already is valid. The problem is that `&self` is *not* `'static`, it's a shorthand for `&'a self`. So you can't have a `&self` method return a `'static`, because that implies that the return value outlives `self`, which would cause use-after-free, so Rust doesn't allow that:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=4bdb2d5abd8db1abb74cb0d885dac852

You can specify a method using `&'static self`...

```rust
impl Foo {
    pub async fn bar(&'static self) { ... }
}
```

...but now you can only put `Foo` inside of a `static`, you can't have a stack-allocated `Foo` anymore.

These restrictions aren't caused by wasm-bindgen, they're caused by lifetimes in Rust, which is necessary to prevent undefined behavior. So it can't be "fixed" by wasm-bindgen, because "fixing" it would cause undefined behavior.

Like I said, the solution is to not use `&self`, instead use `self`. All of these problems are caused by *references*, because references are stack-allocated (except for `&'static`), and Futures are heap allocated, so they can't contain stack references. This isn't an issue with only wasm-bindgen, other Future systems like tokio have it as well, it's just how Futures work in general.

But if you avoid references (i.e. you use `self`), then all the problems go away, because now there aren't any more references, so there's no more lifetime issue (lifetimes only care about references, not owned values).

But... if you can't use references, what if you need to access `self` in multiple places? In that case you can use `self: Rc<Self>`. And if you need to mutate `self` you can use `self: Rc<RefCell<Self>>`. It's very common to use `Rc` and `RefCell` when dealing with JS stuff.

shijunti19:
@Pauan Is there a complete example of using self: RC < self > to call its own method

Pauan:
@shijunti19 I don't think there are any examples, but it's just standard Rust:

```rust
struct Inner {
    value: i32,
}

struct Foo {
    inner: RefCell<Inner>,
}

impl Foo {
    fn new(value: i32) -> Rc<Self> {
        Rc::new(Self {
            inner: RefCell::new(Inner { value }),
        })
    }

    async fn increment(self: Rc<Self>) {
        let mut borrow = self.inner.borrow_mut();
        borrow.value += 1;
    }
}
```

```rust
let x = Foo::new(5);

spawn_local(async move {
    x.increment().await;
});
```

Make sure that you never use `.await` while holding a borrow, or it will probably panic:

```rust
let mut borrow = self.inner.borrow_mut();

// Very bad, the borrow is still alive during the .await, don't do this
some_async().await;
```

```rust
{
    let mut borrow = self.inner.borrow_mut();
}

// This is okay, since it drops the borrow before the .await
some_async().await;

{
    // It can reborrow after the .await is finished
    let mut borrow = self.inner.borrow_mut();
}
```

flosse:
@Pauan thanks for your example.

I tried to combine this with wasm-bindgen but unfortunately that does not work.
Here is what I tried to do:

```rust
use std::{cell::RefCell, rc::Rc};
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct Wrapper {
    inner: RefCell<InnerCore>,
}

pub struct InnerCore {
    val: u32,
}

// TODO:
// Using `#[wasm_bindgen]` does not compile:
//
//     the trait `IntoWasmAbi` is not implemented for `std::rc::Rc<Wrapper>`
//
// #[wasm_bindgen]
impl Wrapper {
    //#[wasm_bindgen(constructor)]
    pub fn new() -> Rc<Wrapper> {
        Rc::new(Self {
            inner: RefCell::new(InnerCore { val: 33 }),
        })
    }

    pub async fn method_one(self: Rc<Wrapper>, _: u32) -> Result<JsValue, JsValue> {
        todo!()
    }

    pub async fn method_two(self: Rc<Wrapper>, _: String) -> Result<JsValue, JsValue> {
        todo!()
    }
}
```

Do you have any idea how to solve this?
Here is the code to play around with: https://github.com/flosse/wasm-bindgen-async-method-example

Pauan:
@flosse You have to move the `Rc` inside the struct:

```rust
#[wasm_bindgen]
#[derive(Clone)]
pub struct Wrapper {
    inner: Rc<RefCell<InnerCore>>,
}
```

PR #1758, opened 2019-09-07
Remove shell status update

cburgos:
Seems like this was a hard-coded update left over and no longer needed?


alexcrichton:
Sounds good to me!

PR #1760, opened 2019-09-08
Major improvements to wasm-bindgen-futures

Pauan:
Sorry I'm late, I struggled with the multi-threading atomic stuff.

This PR contains a few major improvements:

* Code duplication has been removed.

* Everything has been refactored so that the implementation is much easier to understand.

* `future_to_promise` is now implemented with `spawn_local` rather than the other way around (this means `spawn_local` is faster since it doesn't need to create an unneeded `Promise`).

* Both the single threaded and multi threaded executors have been rewritten from scratch:

   * They only create 1-2 allocations in Rust per Task, and all of the allocations happen when the Task is created.

   * The singlethreaded executor creates 1 Promise per tick, rather than 1 Promise per tick per Task.

   * Both executors do *not* create `Closure`s during polling, instead all needed `Closure`s are created ahead of time.

   * Both executors now have correct behavior with regard to spurious wakeups and waking up during the call to `poll`.

   * Both executors cache the `Waker` so it doesn't need to be recreated all the time.

I believe both executors are now optimal in terms of both Rust and JS performance.


Pauan:
(The build failures don't seem related to this PR, except for the `unix` tests, which seem... really weird. It almost seems like a bug with `anyref`)

crates/futures/src/lib.rs
```diff
@@ -7,46 +7,222 @@
...
-cfg_if! {
-    if #[cfg(target_feature = "atomics")] {
-        mod wait_async_polyfill;
-        mod multithread;
-        pub use multithread::*;
-    } else {
-        mod singlethread;
-        pub use singlethread::*;
-     }
+        // TODO change Promise::new to be FnOnce
```
alexcrichton: I don't think `#[wasm_bindgen]` supports this, so I'd prefer the `TODO` be removed an an issue filed (although I suspect there's already an issue)

crates/futures/src/task/multithread.rs
```diff
@@ -0,0 +1,201 @@
...
+    } else {
+        let mem = wasm_bindgen::memory().unchecked_into::<js_sys::WebAssembly::Memory>();
+        Atomics::wait_async(&mem.buffer(), ptr as *const AtomicI32 as i32 / 4, current_value)
+    }
+}
+
+fn notify(atomic: &AtomicI32, wakeup: u32) {
+    unsafe {
+        core::arch::wasm32::atomic_notify(
+            // TODO verify the safety of this
```
alexcrichton: For this the safety here looks fine by me, so this should be safe to remove the comment

crates/futures/src/task/multithread.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+impl AtomicWaker {
+    fn new() -> Arc<Self> {
+        Arc::new(Self {
+            state: AtomicI32::new(AWAKE),
+        })
+    }
+
+    fn done(&self) {
```
alexcrichton: Could these one-liner methods be inlined into where they're used? I find it somewhat difficult to jump back and forth when most of them are only used once anyway

crates/futures/src/task/multithread.rs
```diff
@@ -0,0 +1,201 @@
...
+
+            if let Poll::Ready(_) = poll {
+                // This prevents it from notifying again
+                self.atomic.done();
+
+                // Cleanup the Future and Closure immediately
+                *borrow = None;
+
+            } else {
+                // TODO this creates 2 Promises, figure out a way to avoid that
```
alexcrichton: I don't think this `TODO` is particularly actionable, we're sort of at the whims of web designers and a TODO here probably won't sway them much :)

crates/futures/src/task/multithread.rs
```diff
@@ -0,0 +1,201 @@
...
+            self.atomic.sleep();
+
+            let poll = {
+                let cx = &mut Context::from_waker(&self.waker);
+                inner.future.as_mut().poll(cx)
+            };
+
+            if let Poll::Ready(_) = poll {
+                // This prevents it from notifying again
+                self.atomic.done();
```
alexcrichton: I think this is fine to omit, it's saving at most one `atomic.notify` instruction and everything else here is so expensive I think it's safe to just omit this and deal with it later if necessary.

crates/futures/src/task/multithread.rs
```diff
@@ -0,0 +1,201 @@
...
+        if let Some(inner) = borrow.as_mut() {
+            // This is so that it will re-queue if poll calls waker.wake()
+            self.atomic.sleep();
+
+            let poll = {
+                let cx = &mut Context::from_waker(&self.waker);
+                inner.future.as_mut().poll(cx)
+            };
+
+            if let Poll::Ready(_) = poll {
```
alexcrichton: Could this `if let` be turned into a match since you're handling both variants?

alexcrichton: Also could `_` here be explicitly bound as `()`?

crates/futures/src/queue.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        lock.push_back(task);
+
+        // If we already scheduled the next tick then do nothing
+        if self.state.is_spinning.replace(true) {
+            return;
+        }
+
+        // The Task will be run on the next microtask event tick
+        // TODO replace with https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask
```
alexcrichton: I think it's fine to remove the "TODO" here and basically just comment that this is probably slightly more expensive than it needs to be and the proposed API here may be usable instead. I doubt we'll ever need to do it though since this is already so heavily amortized.

crates/futures/src/queue.rs
```diff
@@ -0,0 +1,90 @@
...
+    // This is a queue of Tasks which will be run in order
+    tasks: RefCell<VecDeque<A>>,
+}
+
+impl<A> QueueState<A>
+where
+    A: Task,
+{
+    fn run_all(&self) {
+        loop {
```
alexcrichton: Could a comment here be added that this greedily performs all work, but we may want to one day in the future have it capped at a fixed number of iterations to prevent an infinite loop in wasm from never returning to the event loop?

crates/futures/src/queue.rs
```diff
@@ -0,0 +1,90 @@
+use js_sys::Promise;
+use std::cell::{Cell, RefCell};
+use std::collections::VecDeque;
+use std::rc::Rc;
+use wasm_bindgen::prelude::*;
+
+pub(crate) trait Task {
```
alexcrichton: I would personally prefer to remove this `trait` and just use a type from either the singlethread/multithread modules, sort of like how `std::sys` works where it just has the same structure. The trait otherwise just sort of adds unnecessary complication.

alexcrichton: (there's also only ever one instantiation ever)

crates/futures/src/queue.rs
Pauan: That's how the microtask event queue works in general, you can definitely create infinite loops with JS Promises which never return to the event loop:

```js
function loop() {
    return Promise.resolve().then(() => {
        return loop();
    });
}

loop();
```

Do we want to diverge from Promise semantics?

(I also believe that's how non-wasm Executors work, they can also get into infinite loops with their event loop)

crates/futures/src/task/multithread.rs
Pauan: I was imagining a situation where all Rust threads run in workers, so they can use `Atomics.wait`, but that's pretty far in the future.

crates/futures/src/task/multithread.rs
Pauan: I did it that way to encapsulate everything inside the `AtomicWaker`, so that way it's easier to understand, rather than having internal state updates strewn around the file.

But I don't have particularly strong opinions on it one way or the other.

crates/futures/src/task/multithread.rs
Pauan: Since I'm not well versed in unsafe voodoo: why is it safe to cast a `&` to a `*mut`? Doesn't that violate Rust's memory aliasing?

crates/futures/src/queue.rs
Pauan: I definitely plan on replacing it with `queueMicrotask` once it stabilizes. I use TODOs to basically mean, "there is something we can't do now, but we should do it later", just as a reminder so I (or others) don't forget.

crates/futures/src/queue.rs
Pauan: I used a trait so we could reuse it in other situations, like the eventual multi-threading implementation. It could also be useful to allow users to use it as a fast Rust `queueMicrotask` polyfill.

But I guess the multi-threading implementation will need a different queue anyways, so I'm fine with changing this.

crates/futures/src/task/multithread.rs
alexcrichton: `&AtomicI32` has interior mutability through its usage of `UnsafeCell` and while it'd be best to have an inherent method to go from `&AtomicI32` to `*mut i32` this is basically the same. Note that `&` doesn't mean "readonly" it just means "shared"

crates/futures/src/queue.rs
alexcrichton: Er sorry I don't mean that we'll actively break anything, only try to make executions a bit more fair. If you write an infinite loop you'll write an infinite loop, but a typically thing with executors like this is to yield back to the system every so often to let other items progress if that's the case.

crates/futures/src/task/multithread.rs
Pauan: Yeah, but I mean, can't Rust make assumptions that the `*mut` doesn't alias (since it's `*mut`), and then make optimizations based on that, but then that will break since it's actually aliased?

crates/futures/src/queue.rs
Pauan: But doing any sort of yielding would diverge from Promise semantics. I'm not sure that's a good idea, I'd rather the user insert a manual `wait(0)` or similar if they want to yield.

alexcrichton: Er so just to be clear, you understand that I'm not asking for code to change, I'm just asking for a comment, right? This is something we're going to want to get to eventually because it's basically what every futures executor does, so I'd rather not debate its merits now but rather we can deal with that if the need for an implementation actually arises. I'd just like some documentation indicating that this is possibly a known problem in case someone runs into it in the future.

crates/futures/src/task/multithread.rs
alexcrichton: I won't pretend to really understand all the various aspects of the unsafe guidelines, but this pattern is basically required to work in one way or another.

Pauan:
(@alexcrichton I fixed all your feedback btw)

alexcrichton:
I pushed up some stylistic feedback and tweaks that I figured would be easiest to go ahead and do myself. I debugged the issue with anyref and I think we'll need to fix that somehow before this lands. Looks like `Promise.resolve` wasn't previously used in the wasm-bindgen test suite, and it's now used transitively through `wasm-bindgen-futures`. The anyref pass means that `Promise.resolve` is directly imported, as-is, into the wasm module. This is, however, invalid:

```
> Promise.resolve(1)
Promise { 1 }
> x = Promise.resolve
[Function: resolve]
> x(1)
Thrown:
TypeError: PromiseResolve called on non-object
    at resolve (<anonymous>)
```

We have to call `resolve` via dot-notation or somehow set the `this` parameter, so the `wasm-bindgen` optimization to directly import `Promise.resolve` is not actually valid. I'm not sure the best way to fix that at this time.

Pauan:
> I'm not sure the best way to fix that at this time.

Yeah, that's tricky. Can we somehow special-case `static_method_of` so it doesn't get the optimization? Just as a temporary solution until we get proper WebIDL integration.

alexcrichton:
Yeah that seems reasonable to me!

alexcrichton:
@Pauan have you had a chance to take a look at this again? Do you want some help in fixing importing `Promise`?

Pauan:
@alexcrichton I did take a look, but the anyref optimizations are over my head, so some help would be appreciated.

alexcrichton:
Sure thing, I've posted https://github.com/rustwasm/wasm-bindgen/pull/1795 to fix imports of `static_method_of` to ensure that `this` is set correctly.

alexcrichton:
👍 

All green now!

PR #1789, opened 2019-09-25
Fix return type of WebIDL indexing getters

### LINKED ISSUE #1756: DomStringMap::get() throws an unexpected error
### Describe the Bug

`DomStringMap::get()` throws an unexpected error when non-existent key string is passed as argument.

### Steps to Reproduce

1. Build the `todomvc` example:
```
$ cd examples/todomvc
$ npm i && npm run serve
```
2. Visit the locally served webpage, open dev console, and add some todo items.
3. Click on 'x' to remove one of the items.

### Expected Behavior

The remove operation should be successful without errors.

### Actual Behavior

The output is errors starting with:
```
index.js:98 wasm-bindgen: imported JS function that was not marked as `catch` threw an error: expected a string argument

Stack:
Error: expected a string argument
    at passStringToWasm (webpack:///./pkg/index.js?:225:45)
    at Module.__widl_f_get_DOMStringMap (webpack:///./pkg/index.js?:439:22)
    at __widl_f_get_DOMStringMap (http://localhost:8082/index.js:83:94)
    at web_sys::DomStringMap::get::h22a3e60caadf998d (wasm-function[617]:167)
    at todomvc::element::Element::dataset_get::h7e42e0c4503b93ca (wasm-function[174]:400)
    at todomvc::view::item_id::{{closure}}::h53639abff4334373 (wasm-function[141]:136)
    at core::option::Option<T>::map::h4dd41af953263be4 (wasm-function[240]:255)
    at todomvc::view::item_id::h3ead9cd27809c346 (wasm-function[931]:85)
    at todomvc::view::View::bind_remove_item::{{closure}}::h6f34d6986b964f1c (wasm-function[124]:308)
    at todomvc::element::Element::delegate::{{closure}}::h89df052d5a99b170 (wasm-function[127]:1150)
```
Screen Shot 2019-09-06 at 16 10 43

### Additional Context

- I have confirmed this issue exists since wasm-bindgen 0.2.33
- The error is being raised at `el.dataset().get(key)` when `el.dataset()` (which is a `DomStringMap`) doesn't have the corresponding `key`:
https://github.com/rustwasm/wasm-bindgen/blob/7fd6702c6dcfe01d38e0240157e57e0ed6f654d5/examples/todomvc/src/element.rs#L292-L303
Per [API](https://docs.rs/web-sys/0.3.27/web_sys/struct.DomStringMap.html), I think `.get()` should simply return `""` instead in that case.  (like wasm-bindgen <= 0.2.32)
- For clarification, I experimented with DomStringMap related test cases
```
    assert_eq!(element.dataset().get("id"), "", "Shouldn't have data-id");
    element.dataset().set("id", "123").unwrap();
    assert_eq!(element.dataset().get("id"), "123", "Should have data-id");
```
by modifying https://github.com/rustwasm/wasm-bindgen/blob/master/crates/web-sys/tests/wasm/html_element.rs

In my [local experiment](https://github.com/j-devel/wasm-bindgen/blob/48f32365501e66c3c5ae4561740fa1e5353a29e7/crates/web-sys/tests/wasm/html_element.rs#L36-L38) with the latest wasm-bindgen version, the first line `assert_eq!(element.dataset().get("id"), "", "Shouldn't have data-id");` is failing as
```
$ cd crates/web-sys
$ cargo +nightly  test --target wasm32-unknown-unknown --all-features
```

```
...
test wasm::html_element::test_html_element ... FAIL
...
failures:

---- wasm::html_element::test_html_element output ----
    error output:
        wasm-bindgen: imported JS function that was not marked as `catch` threw an error: expec
ted a string argument

        Stack:
        passStringToWasm@http://127.0.0.1:50399/wasm-bindgen-test:241:41
        init/imports.wbg.__widl_f_get_DOMStringMap@http://127.0.0.1:50399/wasm-bindgen-test:948
:26
        web_sys::DomStringMap::get::h7dbc1ad3c254416a@http://127.0.0.1:50399/wasm-bindgen-test_
bg.wasm:wasm-function[1786]:0xfb0d3
        wasm::html_element::test_html_element::h56a8b75625ff47a5@http://127.0.0.1:50399/wasm-bindgen-test_bg.wasm:wasm-function[533]:0x1d138
        core::ops::function::FnOnce::call_once::h2f109aedc6512641@http://127.0.0.1:50399/wasm-bindgen-test_bg.wasm:wasm-function[6316]:0x177fe3
...

```
- I'm novice at WebIDL details and not sure how to fix the above error.

j-devel:
This PR is to update the return signatures of WebIDL indexing getters.  All indexing getters will return optional values unless marked by `[Throws]` in their WebIDL definitions.

I summarize the impact of this change in the list below.  As a result, 14 indexing getters with originally unwrapped return type `T` (non-highlighted in 3rd column) will return `Option<T>` instead.

table-before-after-change-clean

Since this is a breaking change, I am including some commits that deal with affected examples as well.  It turned out that I found only the `todomvc` example (50fbb5c).

**Motivation**

In #1756,  we have observed the situations where indexing getters (such as `__widl_f_get_DOMStringMap` and `__widl_f_get_CSSStyleDeclaration`) can return `undefined` that breaks our assumptions based on their WebIDL definitions,  leading to unrecoverable runtime exceptions.  Following @Pauan 's [advice](https://github.com/rustwasm/wasm-bindgen/issues/1756#issuecomment-529680334), this PR is aimed at fixing the problem by updating the return signatures of indexing getters.

### DIFF
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -580,6 +580,20 @@ impl<'src> FirstPassRecord<'src> {
             let structural =
                 force_structural || is_structural(signature.orig.attrs.as_ref(), container_attrs);
             let catch = force_throws || throws(&signature.orig.attrs);
+            let ret_ty = if id == &OperationId::IndexingGetter {
+                // All indexing getters should return optional values (or
+                // otherwise be marked with catch).
+                match ret_ty {
+                    IdlType::Nullable(_) => ret_ty,
+                    ref ty @ _ => if catch {
+                        ret_ty
+                    } else {
+                        IdlType::Nullable(Box::new(ty.clone()))
+                    },
+                }
+            } else {
+                ret_ty
+            };
             let variadic = signature.args.len() == signature.orig.args.len()
                 && signature
                     .orig
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -297,23 +297,9 @@ impl<'src> FirstPassRecord<'src> {
             },
         };
 
-        // Some WebIDL indexing getters (e.g. __widl_f_get_DOMStringMap) return
-        // `undefined` although the WebIDL specification does not explicitly
-        // state so, which causes uncaught runtime exceptions
-        // (rustwasm/wasm-bindgen#1756). To fix this, we check the return
-        // types of the generated Rust bindings for indexing getters, and
-        // ensure that they are wrapped as either `Result<T, JsValue>` or
-        // `Option<T>`. For indexing getters with an unwrapped return type, we
-        // convert their return types to `Option<T>`.
-        //
-        // Here we compute the `is_indexing_getter_to_fix` flag that tells us
-        // whether the binding function we are generating is for an indexing
-        // getter and its return type is not originally wrapped (i.e. neither
-        // `Result<T, JsValue>` nor `Option<T>`) thus needing the conversion
-        // to `Option<T>`.
-        let is_indexing_getter_to_fix = is_indexing_getter && // exclude non indexing getters
-            !catch && // exclude indexing getters that return `Result<T, JsValue>`
-            if let Some(ref ty) = ret { // exclude indexing getters that return `Option<T>`
+        let is_indexing_getter_to_fix = is_indexing_getter && // exclude non index getters.
+            !catch && // exclude indexing getters that return `Result<T, JsValue>`.
+            if let Some(ref ty) = ret { // exclude indexing getters that return `Option<T>`.
                 !format!("{}", quote! { #ty })
                     .replace(" ", "")
                     .starts_with("Option<")
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -3,7 +3,6 @@ use std::ptr;
 
 use heck::{CamelCase, ShoutySnakeCase, SnakeCase};
 use proc_macro2::{Ident, Span};
-use quote::quote;
 use syn;
 use wasm_bindgen_backend::ast;
 use wasm_bindgen_backend::util::{ident_ty, leading_colon_path_ty, raw_ident, rust_ident};
@@ -228,7 +227,6 @@ impl<'src> FirstPassRecord<'src> {
         idl_arguments: impl Iterator<Item = (&'a str, &'a IdlType<'src>)>,
         ret: &IdlType<'src>,
         kind: ast::ImportFunctionKind,
-        is_indexing_getter: bool,
         structural: bool,
         catch: bool,
         variadic: bool,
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for the PR! It's been awhile since I looked at this, so can you describe in the commits/PR message what this is doing?

crates/webidl/src/util.rs
```diff
@@ -294,9 +296,24 @@ impl<'src> FirstPassRecord<'src> {
...
             },
         };
-        let js_ret = ret.clone();
+
+        let is_indexing_getter_to_fix = is_indexing_getter && // exclude non index getters.
+            !catch && // exclude indexing getters that return `Result<T>`.
+            if let Some(ref ty) = ret { // exclude indexing getters that return `Option<T>`.
+                !format!("{}", quote! { #ty })
+                    .replace(" ", "")
+                    .starts_with("Option<")
```
j-devel: Help: I think there should be a smarter way to tell whether `ty` is a `Option< ... >` or not.

j-devel:
@alexcrichton I have updated the description of PR/commits.

alexcrichton:
Thanks for this!

The impact here seems pretty minimal actually so I think it  might be reasonable for us to consider doing this in a point release, despite it being a breaking change.

In terms of implementation though, could this be done not by looking for `"Option<"` but instead by updating the webidl signatures themselves? It seems like we should basically be able to say that all indexing getters should return optional values (or otherwise be marked with `catch`)?

j-devel:
Thanks for the feedback!

> could this be done not by looking for `"Option<"` but instead by updating the webidl signatures themselves?

Yes, and I have arrived at a much more compact solution: 53b0cd0

> It seems like we should basically be able to say that all indexing getters should return optional values (or otherwise be marked with `catch`)?

Precisely.  I have updated the PR's description and implementation comments.

alexcrichton:
Great! Mind doing an analysis again to produce a list of which APIs are changing?

j-devel:
> Mind doing an analysis again to produce a list of which APIs are changing?

I have done an analysis as follows:

**1) set up API signature dumping**
Before the `let invocation` line in
https://github.com/rustwasm/wasm-bindgen/blob/74812bd2366af8c27128798d517202880c278ffc/crates/backend/src/codegen.rs#L1006-L1012
add
```rust
        {
            let js_fn_name = format!("{}", &self.shim);
            let rust_fn = quote! { #vis fn #rust_name(#me #(#arguments),*) };
            let rust_fn_ret = quote! { #ret };
            println!("analysis: | {} | {} | {} |", js_fn_name, rust_fn, rust_fn_ret);
        }
```
**2) collect API signature dumps**
2.1) dump signatures to *,log--before*
```
$ cd crates/web-sys
$ cargo test --target wasm32-unknown-unknown html_element --all-features
$ grep '^analysis: ' ../../target/wasm32-unknown-unknown/debug/build/web-sys-b50e021c48dde6fa/output > ,log--before
```
2.2) apply 53b0cd0
2.3) dump signatures to *,log--after*
```
$ cargo test --target wasm32-unknown-unknown html_element --all-features
$ grep '^analysis: ' ../../target/wasm32-unknown-unknown/debug/build/web-sys-b50e021c48dde6fa/output > ,log--after
```
**3) print the changing APIs**
```
$ git diff --no-index --unified=0 ,log--{before,after}
```
web-sys debug paths

alexcrichton:
Thanks for this!

@fitzgen how do you feel landing this in a point release? While strictly a breaking change it seems like it's sort of hard to use right without this

fitzgen:
I feel OK about it, given that these are pretty niche APIs. It seems like it is worth the trade off.

alexcrichton:
Ok, let's go ahead and merge this then. If issues come up we can yank, revert, and publish a new version, and then schedule it for the next breaking wasm-bindgen release.

PR #1791, opened 2019-09-25
Fix logic for `is_enabled` in the threads transform

alexcrichton:
The threads transform is implicitly enabled nowadays when the memory
looks like it's shared, so ensure that's taken into account in the
`is_enabled` check.


alexcrichton:
No worries!

Issue #1808, opened 2019-10-05
wasm-bindgen-backend await should be used as raw identifier


Zaba505:
### Describe the Bug

Compiler doesn't like keyword await being used as field name in Ident type.

### Steps to Reproduce

```shell
> cargo generate --git https:://github.com/rustwasm/wasm-pack-template
> cd project-name
> wasm-pack build # or cargo check
```

### Expected Behavior

It should build

### Actual Behavior

```shell
> wasm-pack build # or cargo check
[INFO]: Checking for the Wasm target...
[INFO]: Compiling to Wasm...
    Updating crates.io index
   Compiling proc-macro2 v1.0.5
   Compiling unicode-xid v0.2.0
   Compiling log v0.4.8
   Compiling wasm-bindgen-shared v0.2.51
   Compiling syn v1.0.5
   Compiling cfg-if v0.1.10
   Compiling lazy_static v1.4.0
   Compiling bumpalo v2.6.0
   Compiling wasm-bindgen v0.2.51
   Compiling quote v1.0.2
   Compiling wasm-bindgen-backend v0.2.51
error[E0721]: `await` is a keyword in the 2018 edition
   --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/wasm-bindgen-backend-0.2.51/src/codegen.rs:454:79
    |
454 |                         wasm_bindgen::__rt::IntoJsResult::into_js_result(#ret.await)
    |                                                                               ^^^^^ help: you can use a raw identifier to stay compatible: `r#await`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0721`.
error: Could not compile `wasm-bindgen-backend`.

To learn more, run the command again with --verbose.
Error: Compiling your crate to WebAssembly failed
Caused by: failed to execute `cargo build`: exited with exit code: 101
```

### Additional Context

Rust Version: 1.35.0

With 1.38 this doesn't occur.

alexcrichton:
Closing [due to this comment here](https://github.com/rustwasm/wasm-bindgen/pull/1809#issuecomment-539248817)

PR #1809, opened 2019-10-05
Accessing Ident.await now uses a raw identifier

Zaba505:
Fixs #1808


alexcrichton:
Thanks for the report and the PR! The minimum required Rust version for this crate though is now 1.36.0, since it looks like 1.35.0 doesn't have enough support to compile with this `await` keyword. Using a raw identifier actually changes the meaning of the code as well, as can be seen from the failing tests. I think we're not going to fix this at this time and instead recommend that you upgrade to 1.36.0, but if that's too difficult to do then we can explore other opportunities to add support for older compilers?

Issue #1522, opened 2019-05-10
Enable selectionStart/End on HtmlTextAreaElement


wayeast:
### Motivation

Wasm calls to these JS functions would be very handy!

### Proposed Solution

Uncomment the block [here](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/web-sys/webidls/enabled/HTMLTextAreaElement.webidl#L67-L72).

### Additional Context

Hello wasm-bindgen maintainers.  I just wanted to see if there was a good reason why these two functions are still prevented from being implemented in web-sys (?).  It looks like (the?) one compelling reason was lack of support for `Option<u32>` and the like in WebIDL ([this issue](https://github.com/rustwasm/wasm-bindgen/issues/624)), but this appears to have been resolved already [here](https://github.com/rustwasm/wasm-bindgen/pull/630).  Are these just the victims of oversight, or is there something else preventing them from being included in the crate?  Many thanks.

fitzgen:
Yeah, I think this is just an oversight -- want to submit a PR?

wayeast:
@fitzgen #1533 Thanks!

ibaryshnikov:
@fitzgen this may probably be closed since the merge of #1533?

alexcrichton:
indeed thanks!

PR #1811, opened 2019-10-08
Enable selectionStart and selectionEnd for HTMLInputElement

MartinKavik:
The same case like https://github.com/rustwasm/wasm-bindgen/issues/1522, but for `HTMLInputElement`.


alexcrichton:
👍

PR #1820, opened 2019-10-17
Improve worker.js reliability

RReverser:
While it doesn't happen right now in this particular example, `lastPtr` can be potentially overridden several times before the module is fully initialised.

Rather than having a boolean and a storage for one last argument, `await` a promise returned from `wasm_bindgen` itself in the new `onmessage` handler before executing actual command.

This way all the potential tasks will queue up naturally, wait for the initialisation, and then execute in a correct order.

### DIFF
--- a/examples/raytrace-parallel/worker.js
+++ b/examples/raytrace-parallel/worker.js
@@ -1,36 +1,25 @@
 // synchronously, using the browser, import out shim JS scripts
 importScripts('raytrace_parallel.js');
 
-let booted = false;
-let lastPtr = null;
-
 // Wait for the main thread to send us the shared module/memory. Once we've got
 // it, initialize it all with the `wasm_bindgen` global we imported via
 // `importScripts`.
 //
 // After our first message all subsequent messages are an entry point to run,
 // so we just do that.
-self.onmessage = function(args) {
-  self.onmessage = event => run(event.data);
-  const [module, memory] = args.data;
-  wasm_bindgen(module, memory)
-    .then(() => {
-      booted = true;
-      if (lastPtr)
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

RReverser:
@alexcrichton Thanks for merging!

By the way, I found this published crate that pretty much extracted code from this example in a reusable fashion: https://crates.io/crates/web_worker.

I believe it's slightly out of sync and misses e.g. this `worker.js`, but otherwise it seems like quite a good idea to have a ready-to-use adapter for other people (like me) playing with multithreading in Rust + Wasm on their own.

I wonder if you'd want to coordinate with author and maybe help to make sure that their code is up-to-date, or make your own crate, and then switch this example to use it as a dependency?

alexcrichton:
Sure yeah, refactorings here seem reasonable to me! When I made this example afaik it was the only thing that could work, so it was the minimal thing I got working quickly. If there's more official ways to do it nowadays seems good!

RReverser:
@alexcrichton Oh, I don't think there are more official ways - as I said, that crate is basically an extracted version of your code here.

What I'm asking is if you'd want to make that crate *the* official way and sync your code to it, or publish your own crate for others to use.

alexcrichton:
I think we probably don't want to farm this out just yet, I don't have the time to maintain a separate crate and the threads spec is volatile enough I think we want to retain the ability to make changes here if necessary.

RReverser:
Fair enough. I think I'll publish some small wrapper myself then, but for now am hitting some Chromium bugs that need to be fixed before it's released.

PR #1823, opened 2019-10-19
Add option to wasm_bindgen macro for automatic case conversion

dropletzz:
The idea behind this feature is that insead of writing this:
```rust
#[wasm_bindgen(js_name = "someFunkyFunction")]
fn some_funky_function() { }
```
we could be writing this:
```rust
#[wasm_bindgen(js_case)]
fn some_funky_function() { }
```
and the results should be exactly the same. This isn't a replacement for the `js_name` option, just a shorthand for it's most common use case. IMHO, this will improve developer experience quite a bit.

The name of the class/function that's being exported will be converted to CamelCase, if the fist letter of the original identifier is uppercase, or to mixedCase, otherwise (using the case naming of the [heck](https://github.com/withoutboats/heck) crate here).

So far I've only tested it manually on a simple application I created and it seems to work. If you like this and are willing to merge, I will write automated tests, docs and whatnot.

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -45,6 +45,7 @@ macro_rules! attrgen {
             (readonly, Readonly(Span)),
             (js_name, JsName(Span, String, Span)),
             (js_class, JsClass(Span, String, Span)),
+            (js_case, JsCase(Span)),
             (is_type_of, IsTypeOf(Span, syn::Expr)),
             (extends, Extends(Span, syn::Path)),
             (vendor_prefix, VendorPrefix(Span, Ident)),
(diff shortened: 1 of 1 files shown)


Pauan:
Thanks! Since both JS and Rust have rather restricted identifiers, I'd rather us use a simple hardcoded implementation rather than relying on heck. Something like this:

"If a `_` is followed by a lower-case letter (`a-z`) then remove the `_` and make the letter upper-case"

dropletzz:
I agree, we don't really need to bring in a dependency to do the case conversion.

I have a question then: what can I assume about the identifiers? (apart from the fact that they are Rust identifiers) Do I just assume that they follow the rust naming convention? Or do we want to handle all possible cases? For example:
```rust
struct why_onEarth___WouldYou_Do_this_ { }
```

alexcrichton:
Thanks for the PR! Can this also include some tests which exercise this feature?

I think it's fine to basically assume that we don't get weird identifiers like that for now, we're dealing mostly with handwritten code rather than generated code which greatly reduces the likelihood of an identifier like that arising. We can always fix it and tweak it over time, but "don't panic on it" for now I think is fine.

Pauan:
Since it's so common to do the case conversion, could we also get this as a top-level attribute which will apply it to all of the items? Like this:

```rust
#[wasm_bindgen(js_case)]
extern "C" {
    // This has js_case applied
    fn foo_bar() -> i32;

    // This has js_case applied
    fn qux_corge() -> i32;
}
```

This would be in addition to allowing `js_case` on individual items.

alexcrichton:

PR #1828, opened 2019-10-22
Adding in unintern function

Pauan:
See https://github.com/rustwasm/wasm-bindgen/issues/1386#issuecomment-544530828 for an example use-case for this.

### DIFF
--- a/src/cache/intern.rs
+++ b/src/cache/intern.rs
@@ -40,6 +40,14 @@ cfg_if! {
                 }
             })
         }
+
+        fn unintern_str(key: &str) {
+            CACHE.with(|cache| {
+                let mut cache = cache.entries.borrow_mut();
+
+                cache.remove(key);
+            })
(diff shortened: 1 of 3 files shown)


alexcrichton:
👍

src/cache/intern.rs
```diff
@@ -40,6 +40,14 @@ cfg_if! {
                 }
             })
         }
+
+        fn unintern_str(key: &str) {
```
RReverser: Should we have some sort of reference counter perhaps? That is, if several parts of code use `intern`, but then one does `unintern`, maybe we shouldn't evict it yet?

Pauan: I figured that if you want that, it's easy enough to do your own ref-counting in your program.

I'm a bit wary of introducing too much complexity, it seems to me that ref-counting is orthogonal to interning.

PR #1843, opened 2019-10-29
Add first-class support for binary crates

### LINKED ISSUE #1630: First-class support for bin crates
### Motivation

Right now, all crates must have `crate-type = ["cdylib"]`. This causes some issues:

* Even though you're creating an application, you need to use `src/lib.rs` (which goes against Rust idioms).

* You need to use this...

   ```rust
   #[wasm_bindgen(start)]
   pub fn main_js() {}
   ```

   ...rather than this:

   ```rust
   fn main() {}
   ```

* You can't (easily) have multiple binaries, since you can only have one lib per crate.

   So that means you need to create multiple crates, link them together with a workspace, then set up a build script to call wasm-pack multiple times (since it can only compile one crate at a time).

   This is a pretty big deal: you might want to develop both the client and server in the same crate, but right now that's quite clunky.

   Let's compare that with the standard Rust approach, which is to simply make multiple `src/bin/` subfolders.

### Proposed Solution

Both wasm-bindgen and wasm-pack should have native support for binaries. You should be able to create a `src/main.rs` file, slap a `fn main() {}` into it, and have it Just Work(tm).

Similarly, you should be able to create multiple `src/bin/` subfolders, and they should also Just Work(tm).

In other words, the user experience should be exactly the same as creating a non-wasm Rust binary.

That means wasm-bindgen and wasm-pack will need to compile multiple binaries at the same time and output them all together into the output folder.

### Additional Context

There might be a bit of size bloat caused by using bin crates rather than lib crates. We should investigate what we can do to reduce that.

boringcactus:
fixes #1630 

Essentially, "if nothing was tagged `#[wasm_bindgen(start)]` and there is an exported `fn main(i32, i32) -> i32` then process it as though it were tagged that way."

open questions before i'm confident this implementation is ready:
- [ ] should it refuse to autodiscover a `main` that is already tagged for `wasm_bindgen` in some other way?
- [x] can the `shim_idx` field on the function descriptor be safely set to 0?
- [x] can the arg names be justifiably set to `argc` and `argv`?
- [x] should there be an automated test for this behavior somewhere? if so, how and where should that test be written?

### DIFF
--- a/crates/cli-support/src/webidl/mod.rs
+++ b/crates/cli-support/src/webidl/mod.rs
@@ -627,59 +627,40 @@ impl<'a> Context<'a> {
         Ok(())
     }
 
+    // Discover a function `main(i32, i32) -> i32` and, if it exists, make that function run at module start.
     fn discover_main(&mut self) -> Result<(), Error> {
-        let main_info = self
+        // find a `main(i32, i32) -> i32`
+        let main_id = self
             .module
             .functions()
-            .find(|x| x.name.as_ref().map_or(false, |x| x == "main"))
-            .map(|x| (x.ty(), x.id()));
-        if let Some((main_type, main_id)) = main_info {
-            use walrus::ValType::I32;
-            // an entry point `main` has signature (i32, i32) -> i32
-            let type_matches = {
-                let ty = self.module.types.get(main_type);
-                ty.params() == [I32, I32] && ty.results() == [I32]
-            };
-            if !type_matches {
-                return Ok(());
-            }
-            let mut wrapper = walrus::FunctionBuilder::new(&mut self.module.types, &[], &[]);
-            const NAME: &str = "__wasm_bindgen_autodiscoveredmain_wrapper";
-            wrapper
-                .func_body()
-                .i32_const(0)
-                .i32_const(0)
-                .call(main_id)
-                .drop()
-                .return_();
-            let wrapper = wrapper.finish(vec![], &mut self.module.funcs);
-            let export = self.module.exports.add(NAME, wrapper);
-            self.function_exports
-                .insert(NAME.to_string(), (export, wrapper));
-            let descriptor = Descriptor::Function(Box::new(Function {
-                arguments: vec![],
-                shim_idx: 0, // TODO is this bad
-                ret: Descriptor::Unit,
-            }));
-            self.descriptors.insert(NAME.to_string(), descriptor);
+            .find(|x| {
+                use walrus::ValType::I32;
+                // name has to be `main`
+                let name_matches = x.name.as_ref().map_or(false, |x| x == "main");
+                // type has to be `(i32, i32) -> i32`
+                let ty = self.module.types.get(x.ty());
+                let type_matches = ty.params() == [I32, I32] && ty.results() == [I32];
+                name_matches && type_matches
+            })
+            .map(|x| x.id());
+        let main_id = match main_id {
+            Some(x) => x,
+            None => return Ok(()),
+        };
 
-            let export = decode::Export {
-                class: None,
-                comments: vec![],
-                consumed: false,
-                function: decode::Function {
-                    arg_names: vec![],
-                    name: NAME,
-                },
-                method_kind: decode::MethodKind::Operation(decode::Operation {
-                    is_static: true,
-                    kind: decode::OperationKind::Regular,
-                }),
-                start: true,
-            };
+        // build a wrapper to zero out the arguments and ignore the return value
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for this! I agree that pattern matching and using inference here is the way to go. I think though what we need to do here is to perhaps inject a new function to be `start` which calls `main` with zero'd out arguments?

Currently this I believe will export a binding for `main` but we want to be sure to hook it up to all the `start` infrastructure to make sure it's started automatically

boringcactus:
considering that `#[wasm_bindgen(start)]` emits a scary warning if applied to a function with arguments, hacking the planet to get the same effect attached to a function with two arguments is probably not ideal.

if i manually write `#[wasm_bindgen(start)] pub fn start() { main(); }` in Rust and then ask walrus what its body is, i get
```
function Id { idx: 22 } (Some("start"))
  GlobalGet(GlobalGet { global: Id { idx: 0 } })
  LocalSet(LocalSet { local: Id { idx: 182 } })
  Const(Const { value: I32(16) })
  LocalSet(LocalSet { local: Id { idx: 183 } })
  LocalGet(LocalGet { local: Id { idx: 182 } })
  LocalGet(LocalGet { local: Id { idx: 183 } })
  Binop(Binop { op: I32Sub })
  LocalSet(LocalSet { local: Id { idx: 184 } })
  LocalGet(LocalGet { local: Id { idx: 184 } })
  GlobalSet(GlobalSet { global: Id { idx: 0 } })
  Call(Id { idx: 21 } (Some("bin_crate_without_a_bundler::start::h2a2091f59a0e039a")))
  Call(Id { idx: 888 } (Some("<T as wasm_bindgen::convert::traits::ReturnWasmAbi>::return_abi::h4fdec286c4868b5e")))
  Const(Const { value: I32(16) })
  LocalSet(LocalSet { local: Id { idx: 185 } })
  LocalGet(LocalGet { local: Id { idx: 184 } })
  LocalGet(LocalGet { local: Id { idx: 185 } })
  Binop(Binop { op: I32Add })
  LocalSet(LocalSet { local: Id { idx: 186 } })
  LocalGet(LocalGet { local: Id { idx: 186 } })
  GlobalSet(GlobalSet { global: Id { idx: 0 } })
  Return(Return)
function Id { idx: 21 } (Some("bin_crate_without_a_bundler::start::h2a2091f59a0e039a"))
  Call(Id { idx: 20 } (Some("bin_crate_without_a_bundler::main::ha40db6ddc3c404a3")))
  Return(Return)
```

so hopefully replicating that verbatim is not actually necessary (especially bc a few pieces would be difficult or impossible to reliably replicate). zeroing out the arguments and dropping the return value is, unsurprisingly, far easier.

alexcrichton:
I think you'll want to avoid calling `self.program` here and synthesize a new function to place in the `start` function with `walrus`, and I think that'd help cut down on the verbosity here as well?

Additionally can you be sure to add some tests for this change?

boringcactus:
I'm already synthesizing a replacement with walrus, but per your suggestion I've skipped `program` and gone straight to `export` since that's the only part of `program` that actually runs anyway.

i wasn't quite sure where to write tests, but https://github.com/rustwasm/wasm-bindgen/blob/2e03961ca10ae80622d7a6012c25abbd4de19a0a/crates/cli/tests/wasm-bindgen/main.rs#L8-L9 pointed me in the right direction.

alexcrichton:
Oh right sorry good point! I think though that none of `export` or `program` should be needed here, it should be possible to just directly modify the state of the local context to reflect that the main function was found? Either that or this could share infrastructure with the detection of `#[wasm_bindgen(start)]`?

And yeah that `tests` directory should be good to house tests!

crates/cli-support/src/webidl/mod.rs
```diff
@@ -623,6 +627,63 @@ impl<'a> Context<'a> {
         Ok(())
     }
 
+    fn discover_main(&mut self) -> Result<(), Error> {
+        let main_info = self
+            .module
+            .functions()
+            .find(|x| x.name.as_ref().map_or(false, |x| x == "main"))
+            .map(|x| (x.ty(), x.id()));
+        if let Some((main_type, main_id)) = main_info {
```
alexcrichton: Could this use a `match` with an early return to de-indent the rest of this body?

boringcactus: it could indeed!

boringcactus:
`export` *is* the detection of `#[wasm_bindgen(start)]`, but it turns out only one line from `export` is actually needed for the new test to pass, so just using that one line cleans up a lot of other stuff.

alexcrichton:
Perfect, thanks!

PR #1844, opened 2019-10-29
Adding in to_vec method for typed arrays

Pauan:
I noticed that some people needed to copy from a typed array into a `Vec`, and currently this is a bit awkward to do, so this adds in a `to_vec` method which makes it easy.


alexcrichton:
Nice!

Issue #1643, opened 2019-07-05
Add an unsafe method `Uint8Array::view_mut_raw(ptr: *mut u8, length: usize)` so that JS can efficiently initialize a wasm buffer


mstange:
### Motivation

I am using the following pattern to efficiently initialize a large buffer inside wasm memory from JavaScript without copies (see #1079 for full motivation):

```rust
#[wasm_bindgen]
pub struct WasmMemBuffer {
    buffer: Vec<u8>,
}

#[wasm_bindgen]
impl WasmMemBuffer {
    #[wasm_bindgen(constructor)]
    pub fn new(byte_length: u32, f: &js_sys::Function) -> Self {
        let buffer = vec![0; byte_length as usize];
        unsafe {
            let array = js_sys::Uint8Array::view(&buffer);
            f.call1(&JsValue::NULL, &JsValue::from(array))
                .expect("The callback function should not throw");
        }
        Self { buffer }
    }
}

fn compute_buffer_hash_impl(data: &[u8]) -> u32 { ... }

#[wasm_bindgen]
pub fn compute_buffer_hash(data: &WasmMemBuffer) -> u32 {
    compute_buffer_hash_impl(&data.buffer)
}
```

```js
let buffer = new WasmMemBuffer(100000, array => {
  // "array" wraps a piece of wasm memory. Fill it with some values.
  for (let i = 0; i < array.length; i++) {
    array[i] = Math.floor(Math.random() * 256);
  }
});
let hash = compute_buffer_hash(buffer); // No buffer copy when passing this argument. Yay!
buffer.free();
console.log(hash);
```

There are two problems with this: 

 1. There's a mutation here that the rust compiler doesn't know about. `Uint8Array::view` takes an immutable slice ref, but then I go ahead and mutate the contents. Lying to the compiler is dangerous.
 2. The vec is zero-initialized before its contents are overwritten, which wastes almost 400ms in my use case: https://perfht.ml/2XvbnS9 

If I wanted to eliminate the zero initialization, I could do that using `reserve` + `set_len` on the vec. However, rust forbids wrapping uninitialized values in a slice - you have to use raw pointers when initializing uninitialized memory. But the current signature of `Uint8Array::view` requires me to make a slice.

### Proposed Solution

I propose adding a new method to `Uint8Array` (and maybe to the other typed array variants):

`pub unsafe fn view_mut_raw(ptr: *mut u8, length: usize)`

This would let me change my rust code to the following:

```rust
[...]

PR #1855, opened 2019-11-08
Add `copy_to_mem_raw` for the typed arrays (Int8Array, Int16Array, etc)

lshlyapnikov:
The 2nd part of the #1643: adding  `copy_to_mem_raw(&self, dst: *mut $ty, length: usize)`

Plus some refactoring, typed arrays's `view_mut_raw` test cases moved into `TypedArrays.rs` which looks like a better place for these tests.

Closes: #1643

### TODO:
- [x] Tests
- [x] Rustdoc


crates/js-sys/src/lib.rs
```diff
@@ -4841,6 +4841,26 @@ macro_rules! arrays {
...
+            /// It is up to the caller to guarantee that the `MaybeUninit<T>` really is in an initialized state.
+            /// Calling this when the content is not yet fully initialized causes immediate undefined behavior.
+            /// For more details please refer to the `std::mem::MaybeUninit` documentation.
+            ///
+            /// # Panics
+            ///
+            /// This function will panic if this typed array's length is
+            /// different than the length of the provided `dst` vector wrapped in `std::mem::MaybeUninit`.
+            pub unsafe fn copy_to_maybe_uninit(&self, dst: &mut mem::MaybeUninit<Vec<$ty>>) {
+                let vec: *mut Vec<$ty> = dst.as_mut_ptr();
```
lshlyapnikov: this implementation requires that `MaybeUninit` is initialized, and the wrapped destination vector has the same length as the source. Or do we actually want to initialize `MaybeUninit` in here? I could not find in the docs if there is a way to check if `MaybeUninit` is initialized or not.

lshlyapnikov: thinking about this, it kind of makes sense to change it, so the `copy_to_maybe_uninit` does the `MaybeUninit` initialization.

alexcrichton:
Thanks for this! Is there a way to perhaps do this without `MaybeUnint`? This feels like it's not quite the right usage of that where you'd likely want to deal with some sort of pointer/length here instead of a `Vec`

lshlyapnikov:
I will see what I can do.

lshlyapnikov:
@alexcrichton: @That3Percent  was asking for `dst: &mut [MaybeUninit<T>]`, however I believe: `dst: &mut mem::MaybeUninit<&mut [$ty]>` makes a bit more sense. Let me know how it looks.

If `&mut [MaybeUninit<T>]` is required, I would like to see a code snippet with the use case, so I can turn it into a test. Not sure how `copy_to_unsafe` can be used with the requested type signature, still learning.

lshlyapnikov:
>  Is there a way to perhaps do this without MaybeUnint?

I will have a version of the same function that works with a slice, without `MaybeUnint`.

lshlyapnikov:
See this comment: https://github.com/rustwasm/wasm-bindgen/issues/1643#issuecomment-554674323

I don't think we need any additional function, the existing `copy_to` works with uninitialized slice, with a bit of boiler plate.

alexcrichton:
I think I'd personally prefer to just avoid `MaybeUninit` altogether. I don't think there's really any strong idioms about how it's supposed to be used and it generally just feels like we're contorting ourselves trying to use it. Could this take a pointer/length pair instead/

That3Percent:
Responding to @lshlyapnikov :
> I don't think we need any additional function, the existing `copy_to` works with uninitialized slice, with a bit of boilerplate.

No. This is undefined behavior, and cannot be used safely even if it happens to work with a specific version of the rust compiler. From here: [https://doc.rust-lang.org/std/mem/union.MaybeUninit.html](.MaybeUninit.html)

> Moreover, uninitialized memory is special in that the compiler knows that it does not have a fixed value. This makes it undefined behavior to have uninitialized data in a variable even if that variable has an integer type, which otherwise can hold any fixed bit pattern

Responding to @alexcrichton:
> I think I'd personally prefer to just avoid MaybeUninit altogether. I don't think there's really any strong idioms about how it's supposed to be used and it generally just feels like we're contorting ourselves trying to use it. Could this take a pointer/length pair instead/

Taking a pointer/length is acceptable to me, but not ideal. Although there aren't clear idioms around MaybeUninit<T> right now, it's because it's a very new feature to the std lib. However, MaybeUninit<T> is specifically designed for this case and so should be idiomatic in the near future. See the following from: [https://doc.rust-lang.org/std/mem/union.MaybeUninit.html](.MaybeUninit.html)

> You can use MaybeUninit<T> to implement "out-pointers": instead of returning data from a function, pass it a pointer to some (uninitialized) memory to put the result into. This can be useful when it is important for the caller to control how the memory the result is stored in gets allocated, and you want to avoid unnecessary moves.

The contortions you describe are because of a misunderstanding here:

> @alexcrichton: @That3Percent was asking for dst: &mut [MaybeUninit<T>], however I believe: dst: &mut mem::MaybeUninit<&mut [$ty]> makes a bit more sense. Let me know how it looks.

That makes no sense at all. In fact, this implementation ensures that each call will leak the memory allocated in this function. I think I explained it well [https://github.com/rustwasm/wasm-bindgen/issues/1643#issuecomment-551963438](here), that the entire point is to move control of the allocation to the caller.

I'd be willing to take over the implementation here, but it may take some time as I'm presently traveling.

alexcrichton:
I would personally prefer to not be a pioneer in idioms here and stick to a pointer/length rather than type magic.

That3Percent:
> I would personally prefer to not be a pioneer in idioms here and stick to a pointer/length rather than type magic.

That works. The nice thing with this choice is that anyone who is working with MaybeUninit can still call the function with pointers.

That3Percent:
I think a good test case here would be to allocate an array on the stack and populate that stack allocated array with data from the TypedArray. That's the simplest thing I can think of that doesn't bring in a custom allocator.

lshlyapnikov:
> I would personally prefer to not be a pioneer in idioms here and stick to a pointer/length rather than type magic.

@alexcrichton can you write the exact function signature you want to see?

There is `pub fn copy_to(&self, dst: &mut [$ty])`
https://github.com/rustwasm/wasm-bindgen/blob/master/crates/js-sys/src/lib.rs#L4976-L4979

you want to have a similar call that takes a pointer + length instead of mutable slice reference, right? So something like this:
```
pub fn copy_to(&self, dst: *mut $ty, len: usize)
```

lshlyapnikov:
@alexcrichton I have added `copy_to_mem_raw` method (please suggest a better name):
https://github.com/rustwasm/wasm-bindgen/pull/1855/files#diff-c9d0ff464c43ea0811b849c3b468a955R4991-R4998

and corresponding tests for all typed js arrays:
https://github.com/rustwasm/wasm-bindgen/pull/1855/files#diff-c9d0ff464c43ea0811b849c3b468a955R4991-R4998

crates/js-sys/src/lib.rs
```diff
@@ -4978,6 +4978,25 @@ macro_rules! arrays {
...
+            /// memory, that might be unitialized.
+            ///
+            /// This function will efficiently copy the memory from a typed
+            /// array into this wasm module's own linear memory.
+            ///
+            /// # Panics
+            ///
+            /// This function will panic if this typed array's length is
+            /// different than the length argument passed into the function.
+            pub fn copy_to_mem_raw(&self, dst: *mut $ty, length: usize) {
```
RReverser, line 4991: First of all, this function should be marked as `unsafe` since it's writing to a raw pointer.

Second, it seems to pretty much duplicate `raw_copy_to` / `copy_to` above, except by accepting raw pointer instead of a slice. Thus, its body could be replaced with just

```rust
self.copy_to(std::slice::from_raw_parts_mut(dst, length))
```

at which point it's not very clear to me what value does it bring as a separate API.

lshlyapnikov: @RReverser

> First of all, this function should be marked as `unsafe` since it's writing to a raw pointer.

Agreed.

> Second, it seems to pretty much duplicate `raw_copy_to` / `copy_to` above, except by accepting raw pointer instead of a slice.

I brought this up in here: https://github.com/rustwasm/wasm-bindgen/pull/1855#discussion_r422647554

In any case, let me know what you want to do with this PR. `view_mut_raw`  has already been merged, so we can close the issue. This PR is a follow-up that started with idea of introducing a new `copy_to_xxx` that works with `MaybeUninit` and ended up with raw pointer.

Do you think it is worth converting the test cases to use `copy_to`? There is only one existing `copy_to` test case, it tests `Int32Array`.

That3Percent: > Second, it seems to pretty much duplicate `raw_copy_to` / `copy_to` above, except by accepting raw pointer instead of a slice. Thus, its body could be replaced with just
> 
> ```rust
> self.copy_to(std::slice::from_raw_parts_mut(dst, length))
> ```
> 
> at which point it's not very clear to me what value does it bring as a separate API.

No. Constructing a reference to uninitialized memory is illegal and undefined behavior. The entire point of this issue was to be able to work with uninitialized memory in a way that is performant and doesn't invoke undefined behavior.

I'm no longer using wasm-bindgen, so I no longer have any vested issue in the outcome of this PR. `view_mut_raw` probably does the job but I haven't verified this. I think it didn't exist when this issue was opened.

RReverser: > In any case, let me know what you want to do with this PR.

Oh, sorry, I'm not representative of the team, just someone fairly actively involved with Wasm and Wasm-bindgen. It's up to them anyway :)

> No. Constructing a reference to uninitialized memory is illegal and undefined behavior.

Right, that's a good point - the destination should be `&mut [MaybeUninit<$ty>]` instead. Maybe worth adding a function using that accepts such uninitialized slice instead, and then switching `copy_to` to use it under the hood?

lshlyapnikov: @That3Percent

The undefined behavior is pretty well document in here https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html

If you pass a pointer to aligned, continuous block of memory of proper size you are good :). The whole thing is unsafe. The caller should understand what they are doing.

I agree that `MaybeUninit` is the way to go for one reason, it is `safe`... however ` MaybeUninit::uninit_array()` is still nightly-only experimental API:
https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.uninit_array
So to allocate uninitialized array, callers will have to use `unsafe` Rust. So one way or another it has to be unsafe.

This issue should be reviewed again when `MaybeUninit::uninit_array()` is stabilized.

That3Percent: Hold up. No, this is still undefined behavior. It is not sufficient that the pointer is aligned, contain a continuous block of memory, etc. These are merely additional constraints on top the normal invariants that must be upheld for references (of which there are many!)

Please see the [documentation ](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html)on `MaybeUninit` which states:

> This makes it undefined behavior to have uninitialized data in a variable even if that variable has an integer type, which otherwise can hold any fixed bit pattern

Also please see the [section in the nomicon](https://doc.rust-lang.org/nomicon/unchecked-uninit.html) about unchecked uninitialized memory which states:

> It is illegal to construct a reference to uninitialized data

Saying "the whole thing is unsafe" is not a license to be fast and loose. "Unsafe" means that you are providing a guarantee to the compiler that there is no undefined behavior. But this is clearly undefined behavior and illegal. I'm not sure how I can state this more strongly.

That3Percent: Furthermore, the API `MaybeUninit::uninit_array` is irrelevant to the discussion. It is for creating an uninitialized array, not a slice.

Creating an uninitialized slice is straightforward and does not require any nightly APIs. Here is a link to the relevant [docs](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element)

RReverser: FWIW another alternative is to accept a pointer-based slice: `*mut [T]`. It can be constructed safely via [`std::ptr::slice::from_raw_parts`](https://doc.rust-lang.org/stable/std/ptr/fn.slice_from_raw_parts.html) - since it's a raw pointer, it's not a subject to the same UB, and can be even cast from either a regular slice or `MaybeUninit` one.

PR #1868, opened 2019-11-21
Pin the toolchain to nightly to fix build

darinmorrison:
This PR pins the toolchain to nightly for the `raytrace-parallel` example. Running `build.sh` fails otherwise since it uses compiler options only available on nightly.


darinmorrison:
Looks like the build is failing due to a missing rustup component. Maybe it's better to just mention in the docs that nightly is required to build the example?

PR #1869, opened 2019-11-21
Update .gitignore to ignore cargo-timing*.html

darinmorrison:
This PR updates the `.gitignore` to ignore the `cargo-timing*.html` files that are generated after building and running the example.


alexcrichton:
Oh oops! This was an extra flag that was never meant to be committed, so I've removed the flag in https://github.com/rustwasm/wasm-bindgen/commit/9768ec9cabecb5f17cc8737b548f8e6ab396409e. Thanks for the PR though!

PR #1870, opened 2019-11-21
Update alert message regarding Firefox version

darinmorrison:
This PR updates the alert message regarding the required Firefox version for running the demo. As far as I can tell, the demo doesn't actually work with the current Firefox nightly. It does work with the current stable release though.

(It seems like it should also work with Chrome since I think the required features are supported there now too [but also must be enabled]. However, when I tried with Chrome it didn't work; just gets stuck at "loading wasm" on the button or something.)


alexcrichton:
Thanks!

PR #1876, opened 2019-11-24
Allow for js property inspection

### LINKED ISSUE #1857: Unable to view properties of exported Rust structs
### Describe the Bug

When exporting a struct from Rust to JavaScript, calling `console.log` or `JSON.stringify` on an object of the class shows only a `ptr` property and none of the exported struct's properties

### Steps to Reproduce

I've created an example project at [codehearts/whoops](https://github.com/codehearts/whoops/tree/master/rust/debug-wasm-exported-rust-struct) to reproduce the issue

Consider the following exported Rust struct:

```rust
#[wasm_bindgen]
pub struct Series {
    name: String
}

#[wasm_bindgen]
impl Series {
    #[wasm_bindgen(js_name=okKo)]
    pub fn ok_ko() -> Self {
        Self {
            name: "OK K.O.! Let's Be Heroes".to_string(),
        }
    }

    #[wasm_bindgen(getter)]
    pub fn name(&self) -> String {
        self.name.clone()
    }
}
```

In the output from `wasm-pack build`, the `Series` object has the following properties:

```javascript
const series = Series.okKo();
console.log(series.name); // OK K.O.! Let's Be Heroes
```

However, when logging the object we see neither of these properties:

```javascript
console.log(series); // Series { ptr: 1114136 }
console.log(JSON.parse(JSON.stringify(series))); // { ptr: 1114136 }
```

### Expected Behavior

I would expect to see the `name` property exported from Rust, like how a native JavaScript class would look

### Actual Behavior

The object shows only a `ptr` property, which is not useful for debugging or logging state

### Additional Context

I found a solution that works for Node.js at least. The output of `console.log` can be customized by defining a `[util.inspect.custom]` method within the class, and the `JSON.stringify` result is based on an existing `toJSON` method if one is defined

wasm-bindgen would have to define the `[util.inspect.custom]` method because `util` is not imported by default and the brackets create an invalid symbol during compilation, but `toJSON` can be exported from the Rust

I'm wondering if wasm-bindgen should provide a default `[util.inspect.custom]`, `toJSON`, and `toString` which provides the exported properties. Would these be appropriate methods for wasm-bindgen to include when targeting Node.js (and possible just JS for the JSON and string representations?)

codehearts:
This resolves #1857 by introducing `#[wasm_bindgen(inspectable)]`. This annotation:

- Generates `toJSON`, which can be overridden if the user defines their own `toJSON`
- Generates `toString`, which can be overridden if the user defines their own `toString`
- Generates `[util.inspect.custom]` for the Node.js target only, which applies to `console.log` and friends

Its allows all public fields/getters of the Rust struct to be shown when logging or serializing the generated class in JavaScript (`console.log`, `JSON.stringify`, etc.)

Let me know if this needs any additional work! I thought about using "display," "debug," or "serialize" to mimic the similar derivations in Rust, but I decided "inspectable" was more clear

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -674,7 +674,7 @@ impl<'a> Context<'a> {
 
             if self.config.mode.nodejs() {
                 // `util.inspect` must be imported in Node.js to define [inspect.custom]
-                self.import_name(&JsImport {
+                let module_name = self.import_name(&JsImport {
                     name: JsImportName::Module {
                         module: "util".to_string(),
                         name: "inspect".to_string(),
@@ -685,13 +685,14 @@ impl<'a> Context<'a> {
                 // Node.js supports a custom inspect function to control the
                 // output of `console.log` and friends. The constructor is set
                 // to display the class name as a typical JavaScript class would
-                dst.push_str(
+                dst.push_str(&format!(
                     "
-                    [inspect.custom]() {
-                        return Object.assign(Object.create({constructor: this.constructor}), this.toJSON());
-                    }
-                    "
-                );
+                    [{}.custom]() {{
+                        return Object.assign(Object.create({{constructor: this.constructor}}), this.toJSON());
+                    }}
+                    ",
+                    module_name
+                ));
             }
         }
 
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -676,6 +676,29 @@ impl<'a> Context<'a> {
                 ",
                 readable_properties = readable_properties_js_object
             ));
+
+            if self.config.mode.nodejs() {
+                // `util.inspect` must be imported in Node.js to define [inspect.custom]
+                self.import_name(&JsImport {
+                    name: JsImportName::Module {
+                        module: "util".to_string(),
+                        name: "inspect".to_string(),
+                    },
+                    fields: Vec::new(),
+                })?;
+
+                // Node.js supports a custom inspect function to control the
+                // output of `console.log` and friends. The constructor is set
+                // to display the class name as a typical JavaScript class would
+                dst.push_str(
+                    &format!("
+                    [inspect.custom]() {{
+                        return Object.assign(Object.create({{constructor: this.constructor}}), {readable_properties});
+                    }}
+                    ",
+                    readable_properties = readable_properties_js_object
+                ));
+            }
         }
 
         dst.push_str(&format!(
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -57,6 +57,10 @@ pub struct ExportedClass {
     typescript: String,
     has_constructor: bool,
     wrap_needed: bool,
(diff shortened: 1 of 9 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -644,6 +648,59 @@ impl<'a> Context<'a> {
...
+            );
+
+            dst.push_str(&format!(
+                "
+                toJSON() {{
+                    return {readable_properties};
+                }}
+
+                toString() {{
+                    return JSON.stringify({readable_properties});
```
alexcrichton: Could this perhaps be `JSON.stringify(this.toJSON())` to avoid duplication?

crates/cli-support/src/js/mod.rs
```diff
@@ -644,6 +648,59 @@ impl<'a> Context<'a> {
...
+
+            if self.config.mode.nodejs() {
+                // `util.inspect` must be imported in Node.js to define [inspect.custom]
+                self.import_name(&JsImport {
+                    name: JsImportName::Module {
+                        module: "util".to_string(),
+                        name: "inspect".to_string(),
+                    },
+                    fields: Vec::new(),
+                })?;
```
alexcrichton: The return value of this is a `String` which I think should be used instead of `import.custom` below

crates/cli-support/src/js/mod.rs
```diff
@@ -644,6 +648,59 @@ impl<'a> Context<'a> {
...
+
+                // Node.js supports a custom inspect function to control the
+                // output of `console.log` and friends. The constructor is set
+                // to display the class name as a typical JavaScript class would
+                dst.push_str(
+                    &format!("
+                    [inspect.custom]() {{
+                        return Object.assign(Object.create({{constructor: this.constructor}}), {readable_properties});
```
alexcrichton: Could this use `this.toJSON()` instead of duplicating the readable properties?

crates/cli-support/src/js/mod.rs
codehearts: I implemented it that way originally, but I thought it may be confusing if a struct providing its own `toJSON` had the side effect of changing `toString` and `[util.custom.inspect]`. If it’s documented then it shouldn’t be an issue though. I’ll take your advice and mention it in the guide!

alexcrichton: Yeah sounds good to me!

crates/cli-support/src/js/mod.rs
codehearts: I'm not sure if I can, this import gives me "inspect" back but the symbol name has to be "inspect.custom"… Do you know if there's another method I should try to get the symbol name in a variable?

alexcrichton: Oh sorry what I mean is that below instead of `[inspect.custom]` you'll want to do `[{}.custom]` where `{}` is the return value of this function. That way if the identifier `inspect` is imported from elsewhere it'll still work.

codehearts: That makes sense! I've updated it

alexcrichton:
Thanks again for this!

PR #1895, opened 2019-12-04
Adding in support for async iterators

Pauan:
Async iterators were standardized in ES2018 and are available in all browsers except for Edge. This adds in the minimum amount of support for them.

Ideally we would want to have convenience functions for converting from a Rust Stream into a JS iterable (and vice versa), but maybe that would be better served by gloo.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1349,6 +1349,25 @@ impl Iterator {
     }
 }
 
+// Async Iterator
+#[wasm_bindgen]
+extern "C" {
+    /// Any object that conforms to the JS async iterator protocol. For example,
+    /// something returned by `myObject[Symbol.asyncIterator]()`.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)
+    #[derive(Clone, Debug)]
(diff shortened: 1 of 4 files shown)


alexcrichton:
Looks good to me! Could a few small tests be added if it's been added to Node as well?

I think that a test assertion may need an update in the UI test suite as well?

Pauan:
I think the test failures aren't related to this PR. Or at least, I don't see how they're related.

alexcrichton:
I believe the UI tests are caused by this PR, since other PRs are not suffering the same issues. You can update them by running the tests and setting `TRYBUILD=overwrite`

Pauan:
@alexcrichton Fixed. The fact that merely adding a new type in a completely unrelated crate causes test failures seems *incredibly* brittle.

Python fixes this problem by [allowing you to use `...` in UI tests](https://docs.python.org/2/library/doctest.html#what-about-exceptions). The `...` basically acts like a `.*` in regexps, saying "match anything".

So in this case, if [the `62`](https://github.com/rustwasm/wasm-bindgen/pull/1895/commits/0c73056079d9d5ef7717f020b3322ffee91089ca) was replaced with `...` (or equivalent) then it wouldn't break when adding new types.

alexcrichton:
Yeah UI tests aren't always the easiest to work with, but I'd be fine with any changes that makes them less brittle!

PR #1903, opened 2019-12-07
Changing wasm-in-wasm example to be async

Pauan:
This has three big benefits:

1. It removes an `unsafe` usage.

2. It uses the idiomatic async APIs rather than the soft deprecated sync APIs.

2. The old code would error with a `.wasm` module which is bigger than 4KB. But the new code works no matter how big the `.wasm` file is.


alexcrichton:
Nice!

PR #1905, opened 2019-12-09
Adding in async support for start

### LINKED ISSUE #1904: Support async functions for wasm_bindgen(start)
We currently support `async` functions in a variety of places, including in unit tests and in `pub` functions exposed to JS.

When writing applications, it's common to have a single "root" `async` function which kickstarts everything. Right now you have to do something like this:

```rust
async fn main_async() -> Result<(), JsValue> {
    // ...
}

#[wasm_bindgen(start)]
pub fn main_js() {
    spawn_local(async {
        main_async().await.unwrap_throw();
    });
}
```

But it would be great if we could instead do this:

```rust
#[wasm_bindgen(start)]
pub async fn main_js() -> Result<(), JsValue> {
    // ...
}
```

The semantics is that it would generate a new shim function which calls `spawn_local` and logs the `Err`.

It should support a return value of either `()` or `Result<(), JsValue>`.

Pauan:
Fixes #1904

One issue with this PR is that the `start-function` UI test is *supposed* to error for `foo5`/`foo6` (like how it does for `foo_async3`/`foo_async4`) but it doesn't error. I tried for several hours, but I couldn't figure out why.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -452,11 +452,7 @@ impl TryToTokens for ast::Export {
                     quote! { () },
                     quote! {
                         wasm_bindgen_futures::spawn_local(async move {
-                            let val = <#syn_ret as wasm_bindgen::__rt::StartReturn>::into_null_result(#ret.await);
-
-                            if let std::result::Result::Err(val) = val {
-                                wasm_bindgen::throw_val(val);
-                            }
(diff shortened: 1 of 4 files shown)


alexcrichton:
Nice!

On one hand it'd be good to document this, but on the other hand I also don't think it really needs documentation since this is just being consistent with the rest of `#[wasm_bindgen]`. I'm also ok with the UI tests getting fixed over time (if possible), thanks for adding them though!

PR #1909, opened 2019-12-14
Adding in Array::iter and Array::to_vec

Pauan:
There's a few reasons for this:

* Using `array.to_vec()` is more convenient and more efficient than using `array.values().into_iter().map(|x| x.unwrap_throw()).collect::<Vec<_>>()`

* Using `array.iter()` is much more efficient than using `array.values().into_iter()`

* The iterator returned by `array.iter()` supports `size_hint`, `ExactSizeIterator`, and `DoubleEndedIterator`, all of which *cannot* be supported by `array.values()`

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -425,6 +425,76 @@ extern "C" {
     pub fn unshift(this: &Array, value: &JsValue) -> u32;
 }
 
+/// Iterator returned by `Array::iter`
+#[derive(Debug, Clone)]
+pub struct ArrayIter<'a> {
+    start: u32,
+    end: u32,
+    array: &'a Array,
+}
+
+impl<'a> std::iter::Iterator for ArrayIter<'a> {
+    type Item = JsValue;
+
+    fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -425,6 +425,76 @@ extern "C" {
     pub fn unshift(this: &Array, value: &JsValue) -> u32;
 }
 
+/// Iterator returned by `Array::iter`
+#[derive(Debug, Clone)]
+pub struct ArrayIter<'a> {
+    start: u32,
+    end: u32,
```
alexcrichton: Instead of start/end could this perhaps use `Range<u32>` instead? That could also easily map methods to all of the iterator methods on `Range` as well

alexcrichton:
Looks great to me, thanks!

crates/js-sys/src/lib.rs
Pauan: Done.

Pauan: I wouldn't have thought of using `Range` like that, guess you really do learn something new every day.

alexcrichton:
👍

PR #1913, opened 2019-12-15
Add support for iterators using `js_sys::Iterator`

derekdreery:
I was exploring mapping javascript `Headers` to `http::HeaderMap` and found it not possible using `web-sys`, because the way to access all headers is through the iterable API.

This PR handles `iterable` types by adding an `iter` method to those types that creates a `js_sys::Iterator`.

Open questions

 - [ ] Is this the right way (an `iter` method) to expose the call to [Symbol.iterator]?
 - [ ] Is using the reflect api the best way to implement this?
 - [ ] I'd much prefer to do more of the type-fu within wasm-bindgen.

In particular, an iterator either returns values or an array of 2 values (key, value) in the case of named entries. It would be nice if there was a type like `js_sys::KeyValueIterator<K, V>` where the corresponding `next` method return a tuple `(K, V)`, and a `js_sys::TypedIterator<T>` where `next` returns `T`. I'm using unchecked_into because I know from webidl that the types are correct - this isn't the case for custom `js_sys::Iterators`.


derekdreery:
I've modified the implementation to use typed output values for the iterators.

derekdreery:
To make progress with this (assuming you want to merge it), I would need to know if creating a `TryFrom` in `wasm-bindgen` is OK, and if not, how to do the conversion within the iterator?

src/lib.rs
```diff
@@ -483,6 +536,39 @@ macro_rules! numbers {
...
 numbers! { i8 u8 i16 u16 i32 u32 f32 f64 }
 
+macro_rules! integers {
+    ($($n:ident)*) => ($(
+        impl JsTryFrom<JsValue> for $n {
+            type Error = TypeMismatch;
+
+            fn try_from(value: JsValue) -> Result<Self, Self::Error> {
+                let value = value.as_f64().ok_or(TypeMismatch)?;
+                if value > $n::max_value() as f64 || value < $n::min_value() as f64 {
```
derekdreery: TODO should also do some sort of `is_finite` check.

alexcrichton:
Thanks for this!

I wonder if we could perhaps get by though without dealing with the generics in webidl? We don't, for example, handle anything about promises/futures, we just sling around bland `JsValue` objects. I think that may help reduce the complexity of this PR?

In terms of implementation I think we're still currently in a mode of "if it works it works", no need to worry too too much about speed and such here I think.

derekdreery:
I've moved back to using `JsValue`s.

alexcrichton:
It looks like there's still some typing support and traits like `TryFromJsValue`, would it be possible to prune those as well?

derekdreery:
Hi yeah sorry This isn't ready for review yet. I'll ping once it is.

derekdreery:
Ready for review :).

crates/webidl/src/first_pass.rs
```diff
@@ -115,6 +116,22 @@ pub(crate) struct Arg<'src> {
     pub(crate) variadic: bool,
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum Iterable<'src> {
+    MapLike {
+        key: weedle::types::Type<'src>,
+        value: weedle::types::Type<'src>,
+    },
+    ArrayLike(weedle::types::Type<'src>),
```
alexcrichton: I think the types here can probably be removed?

derekdreery: Yeah I left them in so that in future they could be used to generate rust types, but yep they could be removed now. Shall I remove them?

crates/backend/src/codegen.rs
```diff
@@ -1496,3 +1511,91 @@ fn respan(input: TokenStream, span: &dyn ToTokens) -> TokenStream {
...
+                    Iter(#get_iterator)
+                }
+            }
+        }
+        ast::Iterable::No => {
+            quote! {}
+        }
+    };
+    // if we have something to output and rust_name was given, then wrap, otherwise don't
+    if let Some(rust_name) = rust_name {
```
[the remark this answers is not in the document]
derekdreery: The idea with this code is that either you wrap it in an `impl` yourself and pass `None`, or pass `Some(ident)` and and the function will wrap it in an `impl #ident`.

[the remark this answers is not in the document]
derekdreery: An alternative to what I'm doing currently would be to put the `iter` method in a separate `impl` block, then both cases would be the `Some` branch and could be unified. The downside is the generated docs are slightly less nice to read (since they have more impl blocks). Would you like me to do this?

Pauan:
It would be great if we could support proper static types, so `NodeList` would return an `Iterator<Node>`, for example. The WebIDL should have enough information to do that.

Pauan:
Ah, I see you already tried that before. I'm surprised at @alexcrichton 's reasoning: if we use `JsValue` then we cannot change it without a semver breaking change. And we already have enough information in the WebIDL to do the correct thing.

I think it's possible to support proper static types in a relatively easy way (not too different from the current implementation).

alexcrichton:
I'm not necessarily opposed to having a more typed interface, but originally it was relatively complicated with new traits and such which I felt was a bit overblown. If we want a typed interface I think we'll want to eschew complexity in favor of being pretty simple.

derekdreery:
> I'm not necessarily opposed to having a more typed interface, but originally it was relatively complicated with new traits and such which I felt was a bit overblown. If we want a typed interface I think we'll want to eschew complexity in favor of being pretty simple.

If you could sketch out your thinking, I can implement it.

Pauan:
@alexcrichton That's fair. I think it's possible to get proper static types with only some minor tweaks:

Since it's already generating a brand new `Iter` struct for each iterable type, all it needs to do is change the `Item` type and add in an `unchecked_into`. In other words, change it into this:

```rust
pub fn iter(&self)
    -> impl std::iter::Iterator<Item=(#key, #value)>
{
    struct Iter(js_sys::Iterator);

    impl Iterator for Iter {
        type Item = (#key, #value);
        fn next(&mut self) -> Option<Self::Item> {
            // ...
            let key = entry.get(0).unchecked_into();
            let value = entry.get(1).unchecked_into();
            // ...
        }
    }

    // ...
}
```

derekdreery:
@Pauan I think I need to copy whatever the webidl crate does to generate rust types from webidl types. I'm reading through that code now.

derekdreery:
What I think now would be the best way to do typed iter would be the following (ignoring the map case for simplicity)

```rust
fn iter(&self) -> impl Iterator<Item=Ty> {
    // ...
}
```
when called runs the following code in js
```js
return self[Symbol.iterator];
```
and returns the value to rust
```rust
fn iter(&self) -> impl Iterator<Item=Ty> {
    as_rust_owned(run_js_code(as_js(self)))
}
```
Then, say this value is called `MyIter` (the actually name would be chosen to be unique (it is never seen by the user of wasm-bindgen)

```rust
struct MyIter {
    inner: HandleToIterator
}

impl MyIter {
    #[wasm-bindgen]
    extern "C" fn next_js(&self) -> Option<Ty>;
}

impl Iterator for MyIter {
    type Item=Ty;
    fn next(&self) -> Option<Self::Item> {
        self.next_js()
    }
}
```
where in javascript next_js looks like
```js
function next_js(self) {
    const next = self.next();
    if next.done {
        return js_none;
    } else {
        return js_some(next.value);
    }
}
```

Given this I need the following js-rust glue

  - A function in rust that creates an object in js
  - That has a method `next` callable repeatedly from rust
  - That returns an Option<Ty> into rust.
  - where Ty is a type that can cross the rust-js boundary.

derekdreery:
I think I need a little bit of guidance on how to proceed with typing the output of the iterator.

Pauan:
@derekdreery Does my suggestion not work? You would just add back in [the type arguments to `Iterable`](https://github.com/rustwasm/wasm-bindgen/blob/cde783e510550d47ab0c7aa7da37fdd586f0b53b/crates/backend/src/ast.rs#L341-L343), then just change the code to use `unchecked_into`. It should only require ~6 lines of code to change the implementation.

crates/backend/src/codegen.rs
```diff
@@ -1496,3 +1517,64 @@ fn respan(input: TokenStream, span: &dyn ToTokens) -> TokenStream {
...
+    let item = match iterable {
+        ast::Iterable::MapLike { .. } => quote!((::wasm_bindgen::JsValue, ::wasm_bindgen::JsValue)),
+        ast::Iterable::ArrayLike { .. } => quote!(::wasm_bindgen::JsValue),
+    };
+    let get_values = match iterable {
+        ast::Iterable::MapLike { .. } => quote!{
+            let entry = next.value();
+            debug_assert!(entry.has_type::<js_sys::Array>());
+            let entry: js_sys::Array = entry.unchecked_into();
+            let key = entry.get(0);
```
Pauan:
```suggestion
            let key = entry.get(0).unchecked_into();
```

crates/backend/src/codegen.rs
```diff
@@ -1496,3 +1517,64 @@ fn respan(input: TokenStream, span: &dyn ToTokens) -> TokenStream {
...
+            let value = entry.get(1);
```
Pauan:
```suggestion
            let value = entry.get(1).unchecked_into();
```

crates/backend/src/codegen.rs
```diff
@@ -1496,3 +1517,64 @@ fn respan(input: TokenStream, span: &dyn ToTokens) -> TokenStream {
...
+
+/// Generate the code for an iterable.
+fn iterable(rust_name: &Ident, iterable: &ast::Iterable) -> TokenStream {
+    // Wrap our iterator implementation in a const
+    let const_name = format!("__wbg_generated_const_{}_iter", rust_name);
+    let const_name = Ident::new(&const_name, Span::call_site());
+    let iter_name = Ident::new(&format!("{}Iter", rust_name), rust_name.span());
+
+    let item = match iterable {
+        ast::Iterable::MapLike { .. } => quote!((::wasm_bindgen::JsValue, ::wasm_bindgen::JsValue)),
```
Pauan:
```suggestion
        ast::Iterable::MapLike { key, value } => quote!((#key, #value)),
```

crates/backend/src/codegen.rs
```diff
@@ -1496,3 +1517,64 @@ fn respan(input: TokenStream, span: &dyn ToTokens) -> TokenStream {
...
+        ast::Iterable::ArrayLike { .. } => quote!(::wasm_bindgen::JsValue),
```
Pauan:
```suggestion
        ast::Iterable::ArrayLike { value } => value,
```

crates/backend/src/codegen.rs
```diff
@@ -1496,3 +1517,64 @@ fn respan(input: TokenStream, span: &dyn ToTokens) -> TokenStream {
...
+    let get_values = match iterable {
+        ast::Iterable::MapLike { .. } => quote!{
+            let entry = next.value();
+            debug_assert!(entry.has_type::<js_sys::Array>());
+            let entry: js_sys::Array = entry.unchecked_into();
+            let key = entry.get(0);
+            let value = entry.get(1);
+            Some((key, value))
+        },
+        ast::Iterable::ArrayLike { .. } => quote!(Some(next.value())),
```
Pauan:
```suggestion
        ast::Iterable::ArrayLike { .. } => quote!(Some(next.value().unchecked_into())),
```

derekdreery:
So I tried your suggestions and got the errors

```
the trait `wasm_bindgen::cast::JsCast` is not implemented for `std::string::String`
```
and
```
the trait `wasm_bindgen::cast::JsCast` is not implemented for `std::option::Option<Node>`
```

The problem is that to generate good bindings we want to run the string import code and present the rust side as a string, rather than a `js_sys::JsString` or similar. `unchecked_into` is a no-op as I understand it - it simply reinterprets the js object that we have a handle to.

Pauan:
@derekdreery Hmm, that's annoying, I guess it should be using `JsString` instead of `String`.

As for `Option<Node>`, that is indeed a very tricky problem. To be honest, I really dislike the way that wasm-bindgen currently handles type conversions, it's very complicated with `IntoWasmAbi`, `FromWasmAbi`, `ReturnWasmAbi`, `JsCast`, etc. And none of the type conversion functions handle all use cases, only some use cases.

stdweb handles it so much better: it just has `TryFrom<JsValue>`, `Into<JsValue>`, and `AsRef<JsValue>`. I've been meaning to make an RFC about that, I just haven't had the time.

As for this specific issue, maybe we can ignore those types somehow? It means we'll be missing some `iter` impls, but we can always add them in a future PR.

derekdreery:
@Pauan I think maybe work on the types should be in a separate PR.
@alexcrichton ready for review, just using JsValue for now.

derekdreery:
squashed.

Pauan:
@derekdreery Well, I'm personally okay with having it done in a second PR, but that second PR would have to be done before we release, because changing the type is a breaking change.

derekdreery:
I think the CI fail was spurious.

derekdreery:
just for information, here are all the `iterable` instances in `web-sys` webild files.

```
enabled/FormData.webidl
27:  iterable<USVString, FormDataEntryValue>;

enabled/URLSearchParams.webidl
29:  iterable<USVString, USVString>;

enabled/NodeList.webidl
17:  iterable<Node?>;

enabled/Headers.webidl
29:  iterable<ByteString, ByteString>;

enabled/DOMTokenList.webidl
30:  iterable<DOMString?>;

enabled/MediaKeyStatusMap.webidl
24:  iterable<ArrayBuffer,MediaKeyStatus>;
```

so
 1. There aren't very many
 2. Most are strings

Also the `NodeList` is different in the latest version of the spec. It should be 

```
enabled/NodeList.webidl
17:  iterable<Node>;
```

so there wouldn't be an `Option` in any iterator. We could change this webidl without making a breaking change since iterators are not implemented yet.

Pauan:
@derekdreery Great, so if we use `JsString` and fix the issue with `?` then it should work with `unchecked_into`, that's good news.

derekdreery:
@Pauan yes - but there would be 2 remaining issues

 1. If new webidl files were added with optional types we need to correctly ignore those and not panic
 2. The macinery for converting idl types to syn types uses `String` not `JsString`, so we might have to recreate some of that stuff/create our own syn type.

Alternatively we could do whatever the rest of wasm-bindgen does to emit glue code and actually use `String`s in the signatures. The problem with this is that I don't know how to do it/understand that bit of code.

daxpedda:
Note that in the meantime we have addressed this without specific type support: #3962.

PR #1915, opened 2019-12-17
Consistent inline code formatting in js-sys docs

haveyaseen:
Also corrects the docstring of `Math.cbrt()` (the cubic root of x is not x^3).


alexcrichton:
Thanks!

Issue #1917, opened 2019-12-18
Why CanvasRenderingContext2d.get_image_data argument is float?


poccariswet:
### Summary

https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.CanvasRenderingContext2d.html#method.put_image_data
```rs
pub fn get_image_data(
    &self,
    sx: f64,
    sy: f64,
    sw: f64,
    sh: f64{
) -> Result<ImageData, JsValue>
```
webidl's difinition?
I think not bad u32 or i32 data format.

derekdreery:
In JavaScript all numbers are 64 bit floats. So rust generates `f64` bindings for them. The idea is that `web_sys` is a kind of "lowest-level" direct mirror of JavaScript functions as defined in webidl (automatically generated from the webidl definitions). We can then manually build higher-level libraries on top using the extra information we know (in this case that it makes more sense to take integers). There are some caveats here though - because math in JavaScript is floating point, sometimes you don't get the results you expect. Also, not all `u64` numbers are representable as `f64`, which is why `Number.MAX_SAFE_INTEGER` exists in JavaScript.

The [`gloo`](https://github.com/rustwasm/gloo) project aims to evenutally provide layers on top of `web_sys`. There are already some useful abstractions in there.

Pauan:
Hmm, according to [the spec](https://html.spec.whatwg.org/multipage/canvas.html#2dcontext:dom-context-2d-getimagedata) it should be `i32`, maybe our WebIDL files are wrong?

derekdreery:
From `wasm-bindgen/crates/web-sys/webidls/enabled/CanvasRenderingContext2D.webidl`:

```
254:  ImageData getImageData(double sx, double sy, double sw, double sh);
```

poccariswet:
@Pauan @derekdreery 
Thank you for your suggestions!
I also check [it](https://html.spec.whatwg.org/multipage/canvas.html#2dcontext:dom-context-2d-getimagedata).  I will request to change `wasm-bindgen/crates/web-sys/webidls/enabled/CanvasRenderingContext2D.webidl`

RReverser:
Rather than a question, this probably should be marked / tracked as a bug so that it's fixed on the next major bump (and this applies not just to `getImageData`, but also to `putImageData` and a bunch of other methods).

RReverser:
Ah I guess that's tracked in https://github.com/rustwasm/wasm-bindgen/pull/1920.

PR #1920, opened 2019-12-18
CanvasRenderingContext2d.get_image_data argument is float

poccariswet:
issue #1917

according to the [spec](https://html.spec.whatwg.org/multipage/canvas.html#2dcontext:dom-context-2d-getimagedata) it should be `i32` .

```
 ImageData getImageData(long sx, long sy, long sw, long sh);
```
It also fix putImageData, createImageData


alexcrichton:
Thanks for this! This is unfortunately a breaking change though and once that others are probably already relying on, so I've tagged this for the next breaking release of web-sys.

poccariswet:
@alexcrichton 
Will this request be merged in the next major version release?

alexcrichton:
Indeed!

guybedford:
I've rebased this, and altered the approach to use the unstable merge functionality for web-sys so it will be available behind the web_sys_unstable_apis cfg. This way we can land it now and stabilize for the next web-sys major.

guybedford (approved):

PR #1931, opened 2019-12-28
The example should output "Hello from Rust!"

fbitti:
Thanks for your examples, I'm learning a lot from them.
In this one in particular, the HTML says the user should see "Hello from Rust!" on the console, however it says "Hello, World!". 
This is a proposed fix.

### DIFF
--- a/examples/import_js/crate/defined-in-js.js
+++ b/examples/import_js/crate/defined-in-js.js
@@ -1,5 +1,5 @@
 export function name() {
-    return 'World';
+    return 'Rust';
 }
 
 export class MyClass {
--- a/examples/import_js/crate/src/lib.rs
+++ b/examples/import_js/crate/src/lib.rs
@@ -26,7 +26,7 @@ extern "C" {
 
 #[wasm_bindgen(start)]
 pub fn run() {
-    log(&format!("Hello, {}!", name()));
+    log(&format!("Hello from {}!", name())); // should output "Hello from Rust!"
 
     let x = MyClass::new();
     assert_eq!(x.number(), 42);


limira:
My thought: Leave the example's message unchanged, or change it to:

    Rust-wasm: Hello, JavaScript!

Where `JavaScript` is the value return by the function named `name` from JS.

fbitti:
> My thought: Leave the example's message unchanged, ...
Some change is needed to match the guidance on HTML with the actual output on the JS console, in my humble opinion, this would reduce the level of frustration of a learner who is following the examples ("is the output different because this example is not doing what it says it would, or am I doing something wrong?").
Either change the HTML text or the console output or both.

limira:
Ha, sorry for misreading your original PR's message. I see the reason of this PR now. I never ran the example or read it carefully so I didn't know that the messages are mismatched.

This PR fixes the mismatch. It's turn out that the name **Rust** has to be retrieved from a JS's function and I think it's a bit weird. That's the reason of my comment.

Now, I think the message in `index.html` can be modified to: 

    Open up the developer console and you should see messages that output by Rust.

Because "*Hello...*" is not the only message output there. Again, sorry about the confusion!

fbitti:
Oh, by all means, thanks for your comments and attention. I will also make sure my future PRs have a clear explanation for the motivation. 
As for your suggestion: I can revert the changes to the 2 configuration files and add this change to the HTML. Should I go ahead?

limira:
I am just a random contributor, not the one who make decisions in this project. I just voice my thought, you may do what I say if you agree with it. Otherwise, let's wait for response from the project's members.

fbitti:
Ok, I see... by all means your suggestions are welcome, but I will wait for the owner of this wasm-bindgen example to tell us the way forward.

alexcrichton:
Seems reasonable to me, I'm fine with tweaking this as necessary!

Issue #1926, opened 2019-12-21
Invalid d.ts produced after 0.2.56 upgrade


TomCrypto:
### Describe the Bug

After upgrading to 0.2.56, there are errors in the generated crate.d.ts (not the _bg one) that are due to invalid generated typescript.

### Steps to Reproduce

1. Use the https://github.com/rustwasm/wasm-pack-template or whatever to create a new project
2. Insert the following test case

```rust
#[wasm_bindgen]
pub struct Foo {}

#[wasm_bindgen]
impl Foo {
    pub fn test1(&mut self) -> bool {
        true
    }

    pub fn test2(&mut self, _foo: f32) {
        return;
    }
}
```

### Expected Behavior

The expected (and generated as of 0.2.55) d.ts looked like this:

```ts
export class Foo {
  free(): void;
/**
* @returns {boolean} 
*/
  test1(): boolean;
/**
* @param {number} _foo 
*/
  test2(_foo: number): void;
}
```

### Actual Behavior

When using `wasm-bindgen` 0.2.56 instead, the following invalid typescript is generated instead:

```ts
export class Foo {
  free(): void;
/**
* @returns {bool} 
*/
  test1(): bool;
/**
* @param {number} this.ptr 
*/
  test2(this.ptr: number): void;
}
```

### Additional Context

There might be other regressions; these are just the ones that I immediately found in my wasm project.

alexcrichton:
Thanks for the report here! Indeed `bool` -> `boolean` was one glaring error, and I can see where the other is coming from (the `this.ptr` weirdness), I'll try to work on a fix soon!

PR #1933, opened 2019-12-31
bool -> boolean in generated TypeScript code

paulgb:
[TypeScript uses](https://www.typescriptlang.org/docs/handbook/basic-types.html#boolean) the name `boolean` rather than `bool`. There is one instance of `bool` used instead in `wasm-bindgen`. This fixes the typo.

I believe this fixes the issue behind `test1` in #1926.


alexcrichton:
Oops, sorry about this mistake and thanks for the fix! Would you be up for adding a test for this as well? I think it would go in the `crates/typescript-tests` directory

alexcrichton:
Ah I went ahead and added a test, thanks again!

PR #1938, opened 2020-01-03
Add default module to `init` for `no-modules` output mode.

daxpedda:
Also checks if we are in a worker scope.

Related: #1559.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -437,6 +437,18 @@ impl<'a> Context<'a> {
                         module = import.meta.url.replace(/\\.js$/, '_bg.wasm');
                     }"
             }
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -437,6 +437,18 @@ impl<'a> Context<'a> {
                         module = import.meta.url.replace(/\\.js$/, '_bg.wasm');
                     }"
             }
+            OutputMode::NoModules { .. } => {
+                "\
+                    if (typeof module === 'undefined') {
+                        let src
```
alexcrichton: Could this and the lines below also include semicolons? Other than that looks good to me!

alexcrichton:
Thanks!

PR #1940, opened 2020-01-03
Automatically start execution on import.

### LINKED ISSUE #1939: Automatically start `main` on import.
### Motivation

Especially after #1843, I think it would make a lot of sense that on importing the `js` file, it would immediately start `init` and therefore execute `main`.

This would solve a couple of issues:
-  Less boilerplate: `<script src="wasm.js" type="module"></script>` would be enough.
   -  No need to call `init`
   -  No need to preeload the `js` file (the `wasm` still needs to be preloaded)
-  Can be directly used with `Worker`, because it starts itself.
-  It should only start itself when `main` or `wasm_bindgen(start)` is used in rust, which should be pretty intuitive.

### Proposed Solution

Call `init` after its definition if `needs_manual_start` is true in [`cli-support`](https://github.com/rustwasm/wasm-bindgen/blob/580daab1d33ab1ce86d9ff4b80a12673449bd843/crates/cli-support/src/js/mod.rs).
Additionally we should add a flag to disable this functionality.

### Alternatives

-  Potentially we could always call `init` to start loading the `wasm` file, even if there is no start function.
-  Hide this functionality behind a flag instead of making it default and disabling it by a flag.

### Additional Context

I am afraid there was some reason that this functionality wasn't implemented already, I just can't figure out why.

daxpedda:
This changes `wasm-bindgen` to start execution when importing the `js` file by default and adds a flag to disable this behaviour.

The `no-modules` part currently relies on #1938, I can change that if necessary.

Currently what this does is change `init`/`wasm_bindgen` from a function into a promise. The original function is now in an exported `init` or respectively in `wasm_bindgen` init.

Having worked on this PR a little, made me realize that there is some bigger design work necessary, I can't change this PR in a draft, but discussion is obviously in #1939.

I would like to get some feedback on the documentation/description of the added flag.

Fixes #1939.

### DIFF
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -161,7 +161,6 @@ function executeBenchmark(name, bm) {
 // Load wasm files and when they're done (plus the DOM) then we initialize
 // everything
 const wasms = [];
-wasms.push(wbindgen_init('./pkg/wasm_bindgen_benchmark_bg.wasm'));
 wasms.push(fetch('./raw.wasm')
   .then(r => r.arrayBuffer())
   .then(m => WebAssembly.instantiate(m, { './globals.js': globals }))
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -220,11 +220,16 @@ impl<'a> Context<'a> {
                 js.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, None)?;
                 footer.push_str(&format!(
-                    "self.{} = Object.assign(init, __exports);\n",
-                    global
+                    "self.{} = Object.assign(init{}, __exports);\n",
+                    global,
+                    if needs_manual_start && auto_start {
+                        "()"
+                    } else {
+                        ""
+                    }
                 ));
                 if needs_manual_start && auto_start {
-                    footer.push_str("self.wasm_bindgen();\n");
+                    footer.push_str(&format!("self.{}.init = init;\n", global));
                 }
             }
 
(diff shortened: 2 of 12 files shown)


alexcrichton:
Thanks for the PR here! I'm a bit wary to land this though because initialization can often require some form of configuration or tweaking, and I'm not sure we can reliable automatically start it on import all the time. It'd certainly be a nice goal to move towards though!

(this is what es6 module integration is all about fwiw)

daxpedda:
> I'm a bit wary to land this though because initialization can often require some form of configuration or tweaking, and I'm not sure we can reliable automatically start it on import all the time.

What kind of configuration or tweaking do you mean?
The only thing that comes to my mind is:
-  sometimes its not desired to start automatically
   -  doing some things in javascript first
   -  website isn't build around wasm, and wasm might only be initialized in certain user interactions
-  wasm files have been renamed, so automatic start can't determine the correct wasm file name

Basically just user reasons that can be disabled with the flag, I can't figure out if there is something more technical, is there?

In any case this PR enables automatic start by default, I don't think we should merge this because its going to break most builds, at least until we get more feedback.

alexcrichton:
Yes primarily that hosting reasons mean that we don't actually know the structure of files 100% of the time, it may be tweaked from time to time. I believe contexts like web workers and such may also throw a wrench into the issue.

daxpedda:
What if we make a flag called `--auto-start [file]` and `file` is supposed to be the `wasm` file that's automatically started.
This way, nothing is going to change by default, and anybody who is actually interested in automatic start can pass this flag with a proper path, even if the structure is changed.

This would be an ad-hoc solution, I don't know how else we would solve that.

The alternative is of course to just wait for ES module integration of wasm.

If this suggestion is not acceptable, and we can't come up with better alternatives, maybe we should do exactly that: leave everything as it is and wait for ES module integration.

alexcrichton:
I think I would personally prefer to wait for ES module integration to propagate more. The fact that you need to pass in a file is specific to the output type, and part of the output type is that simply you need to pass in a file of what to load.

daxpedda:
Alright.

Issue #1956, opened 2020-01-15
Inconsistent API and unnecessary mut qualifier.


wdanilo:
Hi! See the signatures of all `uniform_matrixXXX_with_f32_array` functions. Almost all of them consume `data: &[f32]`. However, the ones which are defined for non-square matrices, consume `data: &mut [f32]` instead. To keep the API consistent, all of them should consume `data: &[f32]`.

alexcrichton:
Thanks for the report!

By default we assume mutable access is needed (since we don't know otherwise) and we have [a whitelist](https://github.com/rustwasm/wasm-bindgen/blob/762bd0dabd32961552340c62f6ddae49cba744d2/crates/webidl/src/lib.rs#L185-L221) of methods that don't mutate. If you'd like a PR to extend that list would be most welcome!

wdanilo:
@alexcrichton sure thing, here it is:
https://github.com/rustwasm/wasm-bindgen/pull/1957

:)

wdanilo:
Oh, one more question!
There is another issue - the generated functions are not specialized for all possible variants. There is function `tex_image_2d_with_i32_and_i32_and_i32_and_format_and_type_and_opt_u8_array` but there are not functions for other array types (sure we can do it via the `JsObject` casting). According to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D) we can provide much more types here. 

Would you be so nice and tell me where can I fix it in the code? :)

alexcrichton:
Ah unfortunately that's a lot more complicated since it's all based on the WebIDL that we're reading. Want to open a dedicated issue for that?

wdanilo:
@alexandrestein of course!

PR #1957, opened 2020-01-15
Adding missing `uniformMatrix` binding rules for non-square matrices.

wdanilo:
Fixes https://github.com/rustwasm/wasm-bindgen/issues/1956


alexcrichton:
Thanks!

PR #1958, opened 2020-01-16
Add electron support via --omit-imports

darinmorrison:
This PR adds support for using `wasm-bindgen-cli` for Electron apps through the addition of two flags: `--omit-imports` and `--preloads`. See #1928 for context (the "Alternatives" section at the end of the OP).

@alexcrichton I think this is a better approach since it doesn't require adding a new target. I've also cleaned up the implementation and included your feedback from the original PR. If you think this looks good, we could go with this PR and I'll close the old one.

I'm also open to renaming the flags or modifying the documentation if you have thoughts about that.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -316,6 +316,11 @@ impl<'a> Context<'a> {
 
     fn js_import_header(&self) -> Result<String, Error> {
         let mut imports = String::new();
+
+        if self.config.omit_imports {
+            return Ok(imports)
+        }
+
         match &self.config.mode {
             OutputMode::NoModules { .. } => {
                 for (module, _items) in self.js_imports.iter() {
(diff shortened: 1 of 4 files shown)


darinmorrison:
Just realized I forgot to add the blurb about the switches being experimental. I'll do that tomorrow.

darinmorrison:
After further reflection, I've decided to just drop the `--preloads` flag and not include Electron preloads generation in this PR. The rationale is the same as what I described in the "alternatives" section in the original post of #1928.

At this point I'm not sure yet how useful automatic preloads generation will be so given the concern about long term maintenance I guess it's better to be on the conservative side.

This PR is now pretty simple and just adds a flag `--omit-imports` to disable generation of the import header section. This is enough to allow electron apps to work with either the `bundler` or `web` backend after writing a preloads file by hand.

alexcrichton:
Oh sorry for the delay, thanks for this though!

Can you expand the documentation a bit on this? Right now the docs sort of just say what the option itself already says. For example I don't really know easily what "imports" are in the js context, is this just like the section that brings names into the local scope?

darinmorrison:
@alexcrichton sorry for the delay on this. I've updated the description with more detail. Does that work?

alexcrichton:
Thanks!

PR #1969, opened 2020-01-21
fix: ignore non dependency keys in package json

### LINKED ISSUE #1921: [...] `/path/to/project/root/package.json` can currently [..] have [...] no other fields
Hey there!

I have a project directory, where both `cargo` and `npm` reside. The `package.json` has, among others, a `scripts` section to centralize custom commands related to the project.

When I add

```rust
#[wasm_bindgen(module = "/src/foo.ts")]
[...]
```

to a Rust file, I get the following error when compiling (in this case by invoking `wasm-pack`, but it should be unrelated):

```
error: NPM manifest found at `/path/to/project/root/package.json` can currently only have one key, `dependencies`, and no other fields
```

Can you help me understand why this limitation exists? Why is e.g. ignoring the other keys not an option?

Tarnadas:
resolves #1921

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2769,17 +2769,18 @@ impl<'a> Context<'a> {
             ),
         };
         let mut iter = object.iter();
-        let (key, value) = match iter.next() {
-            Some(pair) => pair,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #1973, opened 2020-01-22
[WIP] add parameter to async function --> error

ultrasaurus:
This change to the fetch example does not compile.
It would be great to include how to do this!

I got the example working locally:
```
cd examples/fetch
npm install
npm run serve
```
All is good!

Then I attempted to add a parameter to the async function, but the example fails to
compile with the following error:

```
$ cargo build
   Compiling fetch v0.1.0 (/Users/sallen/src/rust/wasm-bindgen/examples/fetch)
error[E0597]: `*arg0` does not live long enough
  --> examples/fetch/src/lib.rs:35:1
   |
35 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^-
   | |             |
   | |             `*arg0` dropped here while still borrowed
   | |             argument requires that `*arg0` is borrowed for `'static`

For more information about this error, try `rustc --explain E0597`.
error: could not compile `fetch`.
```

Perhaps I don't know the correct syntax for this.  
It would be fabulous if the example illustrated how to pass a parameter.  
If someone can give me a hint, I can add it :)

### DIFF
--- a/examples/fetch/index.js
+++ b/examples/fetch/index.js
@@ -2,7 +2,7 @@ const rust = import('./pkg');
 
 rust
   .then(m => {
-      return m.run().then((data) => {
+      return m.run("rustwasm/wasm-bindgen").then((data) => {
           console.log(data);
 
           console.log("The latest commit to the wasm-bindgen %s branch is:", data.name);
--- a/examples/fetch/src/lib.rs
+++ b/examples/fetch/src/lib.rs
@@ -33,7 +33,7 @@ pub struct Signature {
 }
 
 #[wasm_bindgen]
-pub async fn run(repo: &str) -> Result<JsValue, JsValue> {
+pub async fn run(repo: String) -> Result<JsValue, JsValue> {
     let mut opts = RequestInit::new();
     opts.method("GET");
     opts.mode(RequestMode::Cors);
--- a/examples/fetch/src/lib.rs
+++ b/examples/fetch/src/lib.rs
@@ -33,13 +33,15 @@ pub struct Signature {
 }
 
 #[wasm_bindgen]
-pub async fn run() -> Result<JsValue, JsValue> {
+pub async fn run(repo: &str) -> Result<JsValue, JsValue> {
     let mut opts = RequestInit::new();
     opts.method("GET");
     opts.mode(RequestMode::Cors);
 
+    let url = format!("https://api.github.com/repos/{}/branches/master", repo);
+
     let request = Request::new_with_str_and_init(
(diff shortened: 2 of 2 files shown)


Pauan:
The issue is that you can't spawn a Future which is using external references, because the Future outlives the stack frame, so that would be unsound.

All Future spawners have this restriction (not just wasm-bindgen). It's complicated because this restriction only applies to references. Owned values and `'static` references are completely fine.

So you have two options:

1. You can change it to accept `String` instead, which doesn't contain any references so it's safe.

2. You can instead use `async move` and return a `Promise`, like this:

```rust
#[wasm_bindgen]
pub fn run(repo: &str) -> Promise {
    let url = format!("https://api.github.com/repos/{}/branches/master", repo);

    future_to_promise(async move {
        let mut opts = RequestInit::new();
        opts.method("GET");
        opts.mode(RequestMode::Cors);

        let request = Request::new_with_str_and_init(
            &url,
            &opts,
        )?;

        request
            .headers()
            .set("Accept", "application/vnd.github.v3+json")?;

        let window = web_sys::window().unwrap();
        let resp_value = JsFuture::from(window.fetch_with_request(&request)).await?;

        // `resp_value` is a `Response` object.
        assert!(resp_value.is_instance_of::<Response>());
        let resp: Response = resp_value.dyn_into().unwrap();

        // Convert this other `Promise` into a rust `Future`.
        let json = JsFuture::from(resp.json()?).await?;

        // Use serde to parse the JSON into a struct.
        let branch_info: Branch = json.into_serde().unwrap();

        // Send the `Branch` struct back to JS as an `Object`.
        Ok(JsValue::from_serde(&branch_info).unwrap())
    })
}
```

This works because it's converting the `&str` into a `String` *outside* of the `async move`, and then using the `String` inside of the `async move`. So there's no references, so everything's okay.

ultrasaurus:
What a great explanation.  Thank you @Pauan !  

The problem was obvious once you pointed it out, but it was so helpful to see the solution, since it can be sometimes hard to construct the correct Rust syntax.

In any case, I've updated the PR using String for the param, since that seems like a simpler approach for this example.

alexcrichton:
Looks great to me, thanks!

shijunti19:
@alexcrichton Reference problem: I can't add a loading state before the request, assign true to the request to prevent concurrency, and assign false after the request

alexcrichton:
Hm I'm not sure I quite understand @shijunti19, can you file a separate issue?

shijunti19:
@alexcrichton  https://github.com/rustwasm/wasm-bindgen/issues/2354

PR #1975, opened 2020-01-23
Allow experimental features to be disabled by setting envvars to "0"

eminence:
It's surprising that the WASM_INTERFACE_TYPES feature (and other
experimental features) can be enabled by setting their environment
variables to "0".  With this change, these features will only be enabled
if their respective environment variable is set to "1".

This change was motivated because after setting `WASM_INTERFACE_TYPES=1` to do some testing, I tried to set them to "0" to disable them and it took me a little while to figure out why I wasn't seeing the results I expected.

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -86,9 +86,15 @@ pub enum EncodeInto {
 
 impl Bindgen {
     pub fn new() -> Bindgen {
-        let anyref = env::var("WASM_BINDGEN_ANYREF").is_ok();
-        let wasm_interface_types = env::var("WASM_INTERFACE_TYPES").is_ok();
-        let multi_value = env::var("WASM_BINDGEN_MULTI_VALUE").is_ok();
+        let anyref = env::var("WASM_BINDGEN_ANYREF")
+            .map(|val| val == "1")
+            .unwrap_or(false);
+        let wasm_interface_types = env::var("WASM_INTERFACE_TYPES")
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR!

While this makes sense to me in theory, how come these env vars are set but you also don't want their corresponding feature enabled?

eminence:
Sorry, I should have explained the motivation a little bit better.  It's basically boils down to "less typing".

In a build script, I've written something like this to do a test with interface types:

    env WASM_INTERFACE_TYPES=1 ./wasm-bindgen --out-dir out foo.wasm

Then later, I want to turn off interface types.   Instead of deleting the `env WASM_INTERFACE_TYPES=1` part, it would be convenient to just flip one byte from `1` to `0`.   Similarly when I want to re-enable interface types, it's again another byte flip instead of typing out the whole variable name.

I should also add that this change is technically a breaking change, since someone out there might be doing something like `WASM_INTERFACE_TYPES=yes`.

This is a fairly weak motivation and I wouldn't feel bad if it was rejected on those grounds :)

alexcrichton:
Ah ok, makes sense, thanks for explaining!

Since these are experimental features and not necessarily intended to be the most ergonomic I think I'd personally prefer to keep the code in wasm-bindgen as simple as possible when handling them, so I'm not sure I'd want to add in conditionals about the values to explicitly handle the `=0` case. Does that sound ok?

eminence:
OK with me!

PR #1978, opened 2020-01-23
feat(test): test CLI on windows

ashleygwilliams:
we're running into a geckodriver issue on windows (appears to be just a CI issue, but hard to say)- but i had the occasion to take a look at the azure-pipelines config here. it looks like the library portion of this project is tested on all 3 platforms, but the cli tool is not. 

i know it's a non-zero addition to ask ya'll to add tests, but it would be very help to have the cli tests run on windows so we can catch issues that might begin here first. my concern is mostly wasm-bindgen-test, since it's significantly more tenuous x-platform.

would love to hear your thoughts!

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -106,6 +106,30 @@ jobs:
       - script: cargo test -p wasm-bindgen-wasm-interpreter
         displayName: "wasm-bindgen-wasm-interpreter tests"
 
+  - job: test_cli_windows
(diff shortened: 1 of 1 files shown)


alexcrichton:
In general I've been wary to add extra tests here since the likelihood of regressions on just one platform are pretty rare and it's otherwise a pain to keep it up and running (not to mention expensive on resources). I think though it's best to evaluate the costs after seeing what it would look like to see things working, so I wouldn't mind taking another look once CI is green.

PR #1980, opened 2020-01-26
Update webidl files based on latest spec

derekdreery:
Update webidl files based on

 - https://dom.spec.whatwg.org/#interface-domtokenlist
 - https://dom.spec.whatwg.org/#interface-nodelist

This is a non-breaking change as iterables are currently ignored.


alexcrichton:
👍

Issue #1591, opened 2019-06-11
Setting Typescript types for returned duck-typed objects.


andreubotella:
### Motivation

As far as I understand, the only way to return an object with a duck-typed interface from Rust to JS for the time being is to build it by reflection and return it as a `JsValue`. The generated Typescript definition rightfully gives the returned type as `any`, which does its job, but it fails to convey the author's intentions. A better solution would be to add a way to declare a Typescript type for a returned JsValue.

### Proposed Solution

The proposed solution would be to add a `wasm_bindgen` attribute, perhaps called something like `return_typescript_type`, containing a Typescript type. Any named references contained in the type must be defined elsewhere in a `typescript_custom_section`, although parsing the TS to ensure that might not be a priority.

```rust
#[wasm_bindgen(js_name = jpegImageSize, return_typescript_type = {width: number, height: number}]
pub fn jpeg_image_size(image: &[u8]) -> JsValue {
   let (width, height): (u32, u32) = unimplemented!();
   let ret = Object::new();
   Reflect::set(&ret, &JsValue::from_str("width"), &JsValue::from_f64(width as f64);
   Reflect::set(&ret, &JsValue::from_str("height"), &JsValue::from_f64(height as f64);
   ret
}
```

### Alternatives

Hopefully sometime in the future, `wasm_bindgen` will offer a way to return duck-typed interfaces from Rust without reflection, which would provide the necessary types for the TS declaration to be reasonably complete by default.

alexcrichton:
Seems like a reasonable feature to me!

Pauan:
Currently you can just return a Rust struct, like so:

```rust
#[wasm_bindgen]
pub struct Foo {
    pub width: f64,
    pub height: f64,
}

#[wasm_bindgen]
pub fn jpeg_image_size(image: &[u8]) -> Foo {
   let (width, height): (u32, u32) = unimplemented!();
   Foo { width, height }
}
```

This also generates correct TypeScript types. The only issue is it creates a JS `class`, so I think it would be cool to have a `plain_object` or similar attribute that would cause it to instead generate a plain JS object:

```rust
#[wasm_bindgen(plain_object)]
pub struct Foo {
    pub width: f64,
    pub height: f64,
}
```

(Obviously methods wouldn't work on it, and there's some subtlety around `free`)

clearloop:
Any progress?

andreubotella:
> Any progress?

[...]

PR #1986, opened 2020-02-02
Add typescript_type attribute

clearloop:
## proc-macro attribute 

`#[wasn_bindgen(plain_object)]`

### Motivation

Ref to @Pauan 's [suggestion][0] in #1591, I just wrote an attribute named `plain_object`.

I am writing a wasm UI library these days, and I have to duplicate some types for 3 times(in my project, rust for 2 and ts for 1) without the ability exporting typed struct from rust...

I feel very excited about `wasm`, and never thinking about that I can contribute to `wasm-bindgen` one day, hope this pr can help.

### Solution

1. Add an attribute named `plain_object`
2. Not generating `constructor` with `palin_object`
2. Hide `free() void;` while generating `.d.ts`

### Result
Now we can export rust types like:

```rust
#[wasm_bindgen(plain_object)]
pub struct Ping {
    pong: bool
}
```

And in typescript file:

```ts
import * as r from "pkg";

const p: r.Ping = {
    pong: true,
};
```

### Confusions

I'm quite a newbie in programming, not sure about if just hide `free() void;` in `d.ts` is okay.

[0]: https://github.com/rustwasm/wasm-bindgen/issues/1591#issuecomment-501784387

### DIFF
--- a/crates/typescript-tests/src/lib.rs
+++ b/crates/typescript-tests/src/lib.rs
@@ -5,4 +5,5 @@ pub mod opt_args_and_ret;
 pub mod optional_fields;
 pub mod simple_fn;
 pub mod simple_struct;
+pub mod typescript_type;
 pub mod web_sys;
--- a/crates/typescript-tests/src/typescript_type.rs
+++ b/crates/typescript-tests/src/typescript_type.rs
@@ -1,0 +1,37 @@
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(typescript_custom_section)]
+const ITEXT_STYLE: &'static str = r#"
+interface ITextStyle {
+    bold: boolean;
+    italic: boolean;
+    size: number;
+}
+"#;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(typescript_type = "ITextStyle")]
+    pub type ITextStyle;
+}
+
+#[wasm_bindgen]
+#[derive(Default)]
+pub struct TextStyle {
+    pub bold: bool,
+    pub italic: bool,
+    pub size: i32,
+}
+
+#[wasm_bindgen]
+impl TextStyle {
+    #[wasm_bindgen(constructor)]
+    pub fn new(_i: ITextStyle) {
+        // parse JsValue
+    }
+
+    pub fn optional_new(_i: Option<ITextStyle>) -> TextStyle {
+        // parse JsValueo
+        TextStyle::default()
+    }
+}
--- a/crates/typescript-tests/src/typescript_type.ts
+++ b/crates/typescript-tests/src/typescript_type.ts
@@ -1,0 +1,9 @@
+import * as wbg from '../pkg/typescript_tests';
+
+const style: wbg.TextStyle = new wbg.TextStyle({
+  bold: true,
+  italic: true,
+  size: 42,
+});
+
+const optional_style: wbg.TextStyle = wbg.TextStyle.optional_new();
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -82,6 +82,7 @@
       - [`getter` and `setter`](./reference/attributes/on-rust-exports/getter-and-setter.md)
       - [`inspectable`](./reference/attributes/on-rust-exports/inspectable.md)
       - [`skip_typescript`](./reference/attributes/on-rust-exports/skip_typescript.md)
+      - [`typescript_type`](./reference/attributes/on-rust-exports/typescript_type.md)
 
 - [`web-sys`](./web-sys/index.md)
   - [Using `web-sys`](./web-sys/using-web-sys.md)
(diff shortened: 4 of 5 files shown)


clearloop:
This pr has some problems, `plain_object` can pass the type check, but can not generate the target struct, and...add 61 lines to delete 1 line, it's kind of stupid, I'll try to find a better way.

clearloop:
This pr has some problems, `plain_object` can pass the type check, but can not generate the target struct, and...add 61 lines to delete 1 line, it's kind of stupid, I'll try to find a better way.

alexcrichton:
Thanks for the PR! I'm not personally too familiar with the motivations here, but @Pauan would you be able to review this?

clearloop:
What about adding a `object_args` attribute to use with `typescript_custom_section`?

I just got a new idea to deal this with `JsValue`, for Example:

```rust
const ITEXT_STYLE: &'static str = r#"
interface ITextStyle  {
    bold: boolean,
    italic: boolean,
    size: i32,
}
"#;

#[wasm_bindgen]
struct TextStyle {
    pub bold: bool,
    pub italic: bool,
    pub size: i32,
}

#[wasm_bindgen]
impl TextStyle {
    #[wasm_bindgen(constructor, object_args="ITextStyle")]
    pub fn new(interface: JsValue) {
        // parse JsValue
    }
}
```

and in typescript, we can generate `TextStyle` like this:
```typescript
import { TextStyle } from "../pkg";

const style = new TextStyle({
    bold: true,
    italic: true,
    size: 8,
});
```

Pauan:
Although I originally suggested this, I'm not sure `plain_object` is the right approach. The problem with `plain_object` is that we're not really sending a Rust struct to JS, instead we're dealing with a JS object.

So I agree that something like `object_args` would be better. However, we can make it a lot more flexible, like this:

```rust
const ITEXT_STYLE: &'static str = r#"
interface ITextStyle {
    bold: boolean;
    italic: boolean;
    size: i32;
}
"#;

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(typescript_type = "ITextStyle")]
    pub type ITextStyle;
}

#[wasm_bindgen]
struct TextStyle {
    pub bold: bool,
    pub italic: bool,
    pub size: i32,
}

#[wasm_bindgen]
impl TextStyle {
    #[wasm_bindgen(constructor)]
    pub fn new(interface: ITextStyle) {
        // parse JsValue
    }
}
```

The `typescript_type` attribute allows you to attach a TypeScript type to the Rust type. Then this will be reflected in the generated `.d.ts`.

The benefit of `typescript_type` rather than `object_args` is that it will automatically work correctly in all places: struct fields, function arguments, and function returns.

The above code is rather verbose, however it's possible to create a macro which will generate it:

```rust
typescript! {
    interface ITextStyle  {
        bold: boolean;
        italic: boolean;
        size: i32;
    }
}
```

So I think it would be best to add in support for `typescript_type`, rather than `plain_object` or `object_args`.

clearloop:
Cool! I agree with `typescript_type` is the best,  I'll try to implement this!

Pauan:
@clearloop As a consequence of https://github.com/rustwasm/wasm-bindgen/pull/2012 I needed to add a `typescript_type` attribute (based on the already existing `typescript_name` attribute). I did some basic tests, and it seems to work correctly.

But this PR is still useful, since it adds in some unit tests. So could you please rebase this on master?

clearloop:
@Pauan  Yep, I'll do it right now!

clearloop:
It occurs lots of warnings in my fork, 

```
warning: conflicting implementations of trait `convert::traits::IntoWasmAbi` for type `&dyn for<'r> core::ops::Fn(&'r _) -> _`:
   --> src/convert/closures.rs:11:9
    |
11  | /         impl<'a, 'b, $($var,)* R> IntoWasmAbi for &'a (dyn Fn($($var),*) -> R + 'b)
12  | |             where $($var: FromWasmAbi,)*
13  | |                   R: ReturnWasmAbi
14  | |         {
...   |
22  | |             }
23  | |         }
    | |_________^ conflicting implementation for `&dyn for<'r> core::ops::Fn(&'r _) -> _`
```

even before rebase, I didn't see these before...should I try to fix these?

clearloop:
Waiting for your review @Pauan, I also have a big update for my project if `typescript_type` can merge into master~

Pauan:
> Waiting for your reply for a long long time...

Sorry, I'm usually busy with a lot of different things, so things sometimes slip through the cracks. If I'm taking too long to respond please do ping me.

> I also have a big update for my project if typescript_type can merge into master~

To be clear, https://github.com/rustwasm/wasm-bindgen/pull/2012 already added in `typescript_type`, so the only thing this PR needs to do is add some unit tests, and change the documentation so that it says that all TypeScript types are supported.

clearloop:
Thanks for offering me this opportunity, I'll remove my verbose code(except the test) right now~

Issue #1988, opened 2020-02-03
Support optional fields in struct reflect in typescript


clearloop:
### Motivation

Support optional fields in struct reflect in typescript, for example:

```rust
#[wasn_bindgen]
pub struct Ping {
    pong: Option<bool>,
}
```

in `.d.ts`

```ts
export class Ping {
    pong?: bool;
}
```

### Proposed Solution

Struct generation in `wasm_bindgen` separate fields into `getters` and `setters`, and the `ret_ty` for `getter` or `args` for `setter` which decide the final field type out are embedded in `JsBuilder`, we can't checkout if some fields' types are optional, but...optional fields generate some special docs like `@returns{{{}}}` and `@params{{{}}}` we can trace.

**If it is appropriate to render fields' types with checking if its doc contains `.. | undefined` ?**

For example, we can change this line

https://github.com/rustwasm/wasm-bindgen/blob/f507a2a5ff2fa84d3197909a4b6f095e90ee4b98/crates/cli-support/src/js/mod.rs#L3070

```rust
if docs.contains("undefined") {
    *ty = "?".to_string() + ret_ty
} else {
    *ty = ret_ty.to_string()
}
```

and this line 

https://github.com/rustwasm/wasm-bindgen/blob/f507a2a5ff2fa84d3197909a4b6f095e90ee4b98/crates/cli-support/src/js/mod.rs#L703

```rust
if &ty[0..1] == "?" {
    ts_dst.push_str("?: ");
    ts_dst.push_str(&ty[1..]);
} else {
    ts_dst.push_str(": ");
    ts_dst.push_str(&ty);
}
```

### Alternatives

I think the solution above is quite a mess, but I can't figure out a better one.

Pauan:
I like the idea of generating optional fields... however, I would suggest that whatever tool you are using *should* work with `| undefined`, because from JS's perspective there isn't much difference, and many TypeScript programs will use `| undefined`.

clearloop:
The `| undefined` checking just applies to `setter` and `getter` in `.d.ts`,  if this solution can make a pr, I'd like to do it, or any other way can add optional fields definition to Typescript...really need this.

btw, @Pauan, would you mind to have a look at #1986, I just wrote a pr about `#[wasm_bindgen(plain_object)]` yesterday.

alexcrichton:
Seems like a reasonable feature to me to support as well! A PR would be most welcome to implement this

PR #1990, opened 2020-02-06
Reflect optional struct fields in typescript

clearloop:
Addressed issue #1988 

Rust test file:
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct Fields {
    pub hallo: Option<bool>,
    pub spaceboy: bool,
}
```

Typescript test file:
```typescript
import * as wbg from '../pkg/typescript_tests';

const fields: wbg.Fields = { spaceboy: true, free: () => { } };
```

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -699,8 +699,13 @@ impl<'a> Context<'a> {
                 ts_dst.push_str("readonly ");
             }
             ts_dst.push_str(name);
-            ts_dst.push_str(": ");
-            ts_dst.push_str(ty);
+            if &ty[0..1] == "?" {
+                ts_dst.push_str("?: ");
+                ts_dst.push_str(&ty[1..]);
+            } else {
+                ts_dst.push_str(": ");
+                ts_dst.push_str(&ty);
+            }
             ts_dst.push_str(";\n");
         }
         dst.push_str("}\n");
@@ -772,9 +777,7 @@ impl<'a> Context<'a> {
         if !self.should_write_global("not_defined") {
             return;
         }
-        self.global(
-            "function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }"
-        );
+        self.global("function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }");
     }
 
     fn expose_assert_num(&mut self) {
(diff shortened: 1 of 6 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -3067,7 +3070,13 @@ impl ExportedClass {
             .typescript_fields
             .entry(field.to_string())
             .or_insert_with(Default::default);
-        *ty = ret_ty.to_string();
+
+        if docs.contains("| undefined") {
+            *ty = "?".to_string() + ret_ty;
+        } else {
+            *ty = ret_ty.to_string();
+        }
```
alexcrichton: This feels like it's a bit of a hack to me, could this be done perhaps in a more structure fashion? I'd prefer to not search around in the *documentation* for `| undefined` and have a leading `?` only to get popped off later.

clearloop: That's why I open an issue first...I'm not sure about checking "| undefined" just like what you said,  the other way to implement this is adding a leading `?` to the `ty` in fn `typescript_signature`, but it has to check if `?` is the first letter of the ty in getter's return type,  because typescript doesn't support optional return type either [#35577][1]

I'll try to implement the leading `?` way now.

[1]: https://github.com/microsoft/TypeScript/issues/35577

clearloop: What about checking if `| undefined` in the match getter? I've got two solutions and I think the first one might be the one.

1. `ty` of  class fields depends on getter's `ret_ty`, and setter's `arg_ty`, if it is optional, the direct way is checking out the getter's `ret_ty` or setter's ~~`jsdoc`~~

2. Go deeper, what if changing `arg_ty` in `binding.rs`? `binding.rs` doesn't know the passing `function` is a `setter`, `getter` or just a normal `function`, if we add a leading `?` to an optional `arg_ty`, it will apply to all functions, and we have to check and pop `?` in every of them.

alexcrichton: Hm I still don't think this PR should really do any sort of string inspection to figure out if a field is optional, we should instead plumb the `bool` through from wherever it originates to the final location that the field accessor is generated.

clearloop: 1. No string sort
2. A plumb bool

Got it, I'll try it again!

clearloop:
Sorry for the late, I got myself some rest last week, with my project blown up...

The latest version has no `String` checks, I think I probably made it, pleeeaaase check out it!

alexcrichton:
Thanks!

PR #1994, opened 2020-02-08
Installation instructions for wasm-bindgen CLI tool

canadaduane:
How to install the CLI tool was not obvious to me from the manual. I first tried `cargo install wasm-bindgen` to get the `wasm-bindgen` CLI tool, but was surprised by the error message:

```
error: specified package wasm-bindgen v0.2.58 has no binaries"
```

This patch instructs how to install `wasm-bindgen-cli`.

### DIFF
--- a/guide/src/reference/cli.md
+++ b/guide/src/reference/cli.md
@@ -11,9 +11,13 @@ always be listed via `wasm-bindgen --help`.
 
 ## Installation
 
-The command line tool is available through a separate crate, `wasm-bindgen-cli`:
+Although the command line tool is available through its own crate, `wasm-bindgen-cli`, the recommended way to install it is through installing the `wasm-pack` crate:
 
 ```
+# Recommended:
+cargo install wasm-pack
+
+# OR
 cargo install wasm-bindgen-cli
 ```
 
--- a/guide/src/reference/cli.md
+++ b/guide/src/reference/cli.md
@@ -9,6 +9,14 @@ always be listed via `wasm-bindgen --help`.
 
 [wasm-pack]: https://github.com/rustwasm/wasm-pack
 
+## Installation
+
+The command line tool is available through a separate crate, `wasm-bindgen-cli`:
+
+```
+cargo install wasm-bindgen-cli
+```
+
 ## Usage
 
 ```


Pauan:
Thanks! Though the recommendation is to use wasm-pack instead (which automatically installs `wasm-bindgen-cli`). So maybe we should also add an paragraph pointing to wasm-pack?

canadaduane:
That would be good to know, too! For my case, I don't think I need `wasm-pack`, but I'm new to wasm-bindgen in general, so I could be wrong!

Pauan:
What is your use-case? wasm-pack is intended for basically all use cases involving wasm-bindgen.

guide/src/reference/cli.md
```diff
@@ -9,6 +9,18 @@ always be listed via `wasm-bindgen --help`.
 
 [wasm-pack]: https://github.com/rustwasm/wasm-pack
 
+## Installation
+
+Although the command line tool is available through its own crate, `wasm-bindgen-cli`, the recommended way to install it is through installing the `wasm-pack` crate:
+
+```
+# Recommended:
+cargo install wasm-pack
```
Pauan: This isn't actually the recommended way to install wasm-pack. I think it would be best for it to just link to the [wasm-pack website](https://rustwasm.github.io/wasm-pack/installer/).

canadaduane:
I'm building a game using https://github.com/not-fl3/good-web-game and https://github.com/not-fl3/miniquad which implements its own wasm<->js interop. I'd like to add a few functions to my game using wasm-bindgen, but without disturbing the existing wasm<->js code or repackaging it.

guide/src/reference/cli.md
canadaduane: I think I'll lean on you to fix it from here. I know basically nothing!

Pauan:
@canadaduane wasm-pack does the following:

* It installs `wasm-bindgen-cli` with the correct version (which is very important, because otherwise things will break).

* It runs `cargo`, `wasm-bindgen`, and `wasm-opt` for you.

* *Optionally* it can do other things like create an npm package, but you don't need to do that.

Basically, wasm-pack is just a thin wrapper around `wasm-bindgen-cli` + `cargo`. So you should still be using it, even if you plan to use your own glue code.

Also, as a general rule wasm-bindgen *requires* its glue code, it is not optional. So if you're going to replace it, you would have to recreate all the things that wasm-bindgen does.

I'm curious: why are you using your own glue code? Why not just use the wasm-bindgen glue code? Using your own glue seems like a lot of extra work.

guide/src/reference/cli.md
Pauan: I would replace it with this:

````
The command line tool is available through its own crate, `wasm-bindgen-cli`.

The recommended way to install it is by using
[`wasm-pack`](https://rustwasm.github.io/wasm-pack/installer/).

However, if you want to install `wasm-bindgen-cli` by itself, you can do this:

```
cargo install wasm-bindgen-cli
```
````

canadaduane:
> I'm curious: why are you using your own glue code? Why not just use the wasm-bindgen glue code? Using your own glue seems like a lot of extra work.

I think that's mostly a question for @not-fl3 who built `miniquad` but for me the advantage was not having to learn another tool in the chain! Probably not a very satisfying answer :)

astraw:
On Windows, I just downloaded wasm-pack from [the installer page](https://rustwasm.github.io/wasm-pack/installer/) (via the link https://github.com/rustwasm/wasm-pack/releases/download/v0.9.1/wasm-pack-init.exe ) and this did not install `wasm-bindgen.exe` in my `.cargo/bin` directory. (It did successfully install `wasm-pack.exe`.) I could get `wasm-bindgen.exe` installed with `cargo install wasm-bindgen-cli`.

I agree that it would be helpful if [the `wasm-bindgen` Guide](https://rustwasm.github.io/wasm-bindgen/) mentioned how to get it.

Pauan:
@astraw When you run `wasm-pack build` it will then automatically download wasm-bindgen-cli.

The reason it doesn't happen during installation is because it has to synchronize the version of wasm-bindgen-cli with the version of wasm-bindgen. But it only knows the version of wasm-bindgen after reading the `Cargo.lock` file.

astraw:
> There isn't any need to run wasm-bindgen-cli directly

I was trying to follow [the script to build all the examples in yew](https://github.com/yewstack/yew/blob/master/examples/build_all.sh). FWIW, once I installed wasm-bindgen directly (with `cargo install wasm-bindgen-cli`), I could do so.

So maybe update the PR with the following?

To install the `wasm-bindgen` command-line tool, download and install wasm-pack and then run `wasm-pack build`. (It is also possible to install it directly with `cargo install wasm-bindgen-cli`, but this may have version conflicts with `wasm-pack`.)

Pauan:
@astraw We ***really*** do not recommend using wasm-bindgen directly, because the version of `wasm-bindgen-cli` must *exactly* match with the version of `wasm-bindgen` which is used by the project.

So our suggestion is for yew to change its scripts to use wasm-pack instead.

astraw:
Can you open an issue on the yew repo to phrase this correctly? You certainly understand the issue better than I do. Otherwise I will just paraphrase what I understand you to say and link here.

astraw:
Because #2076 was merged, this can be closed. I do not have the permissions.

Pauan:
@astraw Thanks!

PR #1995, opened 2020-02-10
Removing self from no-modules target

Pauan:
Some JS targets don't have `self`, so this removes `self` from the `no-modules` target.

The behavior should be the same, it still creates a global `wasm_bindgen` variable and assigns to it.

This is accomplished by creating the variable outside of the function and then assigning to it inside the function.


alexcrichton:
Thanks!

Issue #1950, opened 2020-01-08
Easily support non-standard WebIDL 


alexcrichton:
This is intended to be a tracking issue to pull together a number of requests from a few different locations.

The basic problem is that browsers will ship features before fully stabilized. Often these features are behind flags or some other gating mechanism, but once the gates are all unlocked some pages are able to use these features, and Rust/wasm often want to use the features as well!

Currently we are not adding these unstable WebIDL definitions to web-sys because they are almost guaranteed to change. This means that we'd have to ship breaking changes which we would prefer not to do. We do, however, want to make it easier to consume these WebIDL files!

Split across a number of PRs, there's two possible solutions I think we have right now for supporting this:

1. First we can use `--cfg` flags to support everything in `web-sys` itself. The general idea is that `web-sys` would grow all the unstable WebIDL files, but you'd have to compile with `RUSTFLAGS=--cfg=web_sys_unstable` to actually get anything to show up. This sort of "explicit opt in" would be required by everyone building a crate.
  * The pro of this approach is that it centralizes all the WebIDL into this repository, taking the maintenance burden away from users
  * The cons of this approach is that it's difficult to build for users and it puts more maintenance burden ehre.

2. An alternative to putting these WebIDL annotations in `web-sys` itself would be to encourage users to publish their own crates synthesized from WebIDL. The general idea here is that this repository would publish a stable crate (or tool) which can be used to synthesize a crate from WebIDL. That way users could publish their own `webxr-sys` (for example) and it would not be maintained in this repository. Once the specification stabilizes we'd move the WebIDL into `web-sys` itself.
  * The pro of this approach is that it doesn't have much maintenance burden on us in this repository. It also allows anyone to build their own crate with raw WebIDL in the same manner web-sys is built. This approach is also pretty easy for users to consume, they'd basically just depend on a crate.
  * The cons of this approach are pretty bad though. This involves a relatively significant amount of infrastructure which doesn't exists, so would be difficult to start out. For example these new crates probably want to depend on `web-sys` for basic types to start, which `web-sys` itself doesn't do at all (cross-crate dependencies). Additionally it puts the maintenance burden on users.

I'm still somewhat tempted to go with (1) myself, but we've had enough proposals running around here for awhile that I wanted to get this all written down myself!

kettle11:
Thanks for thinking this through!

Some thoughts:

[...]

PR #1997, opened 2020-02-12
[WIP] Add support for unstable WebIDL

grovesNL:
As a follow-up to #1896, #1950, and #1972, I started to look into supporting unstable WebIDL through `RUSTFLAGS=--cfg=web_sys_unstable_apis` as dementioned by @alexcrichton 

In this PR the idea is to have an attribute like `[WebSysUnstableApi]` on all experimental WebIDL definitions. During parsing, this attribute can be recognized, and we can attach `#[cfg(web_sys_unstable_apis)]` on any blocks which reference these unstable definitions.

For example, the definition of:
```webidl
[WebSysUnstableApi]
interface GPUBindGroup
```

becomes something like:

```rust
#[cfg(web_sys_unstable_apis)]
pub struct GpuBindGroupBinding
```

This PR tries to cover the common types of WebIDL definitions (enums, interfaces, partial interfaces, etc.). It doesn't currently handle exports used internally, such as
```rust
pub extern "C" fn __wbindgen_describe___widl_f_copy_texture_to_texture_with_gpu_extent_3d_dict_GPUCommandEncoder
```

~~(I believe these exports are harmless, but we could hide these too)~~ Edit: looks like this hiding these is necessary

Some possible pros/cons to this approach:
- We can mark certain parts of an API as stable or unstable which could be useful to extend existing APIs
- It's difficult to know whether this PR updates all parts of the parser and code-gen that we need, so this might be error-prone (i.e. we might accidentally leak some experimental API). The changes in general feel a bit invasive, so I'd be interested if there's a better way to approach these changes
- Annotating the WebIDL takes a bit of effort and could be error-prone, compared to moving it into a special "experimental" directory instead of "enabled"
  - Moving WebIDL into a separate "experimental" directory has some other challenges though – currently we parse all WebIDL files as a single `String` and splitting them apart might have some subtle effects (i.e. de-duplicating types which are shared)

I'd be interested in any feedback, especially whether there's a better solution or even a way to simplify passing these `unstable_api` flags around. In general this approach seems to work fine for me so far, but I'd be happy to rework this PR entirely.


Pauan:
As an alternative, what if we kept the experimental WebIDL in a separate `experimental` folder, and then added a new argument to `wasm_bindgen_webidl::compile`, so it would be like this:

```rust
fn compile(webidl_source: &str, experimental_source: &str, allowed_types: Option<&[&str]>) -> Result<String>
```

Now it can generate a single string as usual, but there is a separation between stable and unstable without needing a new `WebSysUnstableApi` attribute.

Pauan:
P.S. Thanks a lot for working on this! This has been bugging me for quite a while now.

grovesNL:
> As an alternative, what if we kept the experimental WebIDL in a separate `experimental` folder

I looked into this briefly, but I wasn't sure how to split both sources in `parse` and later combine their ASTs (`Program`s) back together. Maybe this change wouldn't be too difficult though.

For example, any time we see partial interfaces like `Navigator`, we would probably need to keep them separate during parsing, because it's possible for the interface to be stable but have some unstable attributes attached to it (`readonly attribute GPU gpu`)

alexcrichton:
Thanks for this! I would agree that it would be best if we could avoid editing the `*.webidl` that much (such as adding all the custom attributes). Would @Pauan's suggestion be workable? For inclusions like the fields and such, could mixins be edited into the `*.webidl` file?

grovesNL:
Alright sounds good, I'll try to make the separate directory work

grovesNL:
Ok it seems reasonably straightforward to move these into a separate directory and remove the attribute from the WebIDL definitions. I'll update this PR again once I work through a couple remaining issues

grovesNL:
Updated the PR again to use an `unstable` directory – let me know what you think.

The nightly failures seem unrelated to this PR (https://dev.azure.com/rustwasm/wasm-bindgen/_build/results?buildId=2042&view=logs&j=aef2040c-a7c2-5c59-fd4c-9a8df33dc1b6&t=529981cf-1b81-530f-0edf-a26fa1e08bdd&l=42)

alexcrichton:
Looks great to me, thanks!

I think that lots of `WebSysUnstableApi`  annotations are still present though? Could the `*.webidl` be directly imported now except for perhaps a small edit to include an extra mixin to make sure the global property is available?

grovesNL:
Thanks for taking a look! The annotations were left in by mistake while rebasing but should be fixed now.

alexcrichton:
Ok you can disregard my previous comment about editing webidl files, this looks great!

A few final asks before merging:

* Can you be sure to add tests to CI that we build web-sys with this feature flag?
* Can you configure docs.rs to pass this flag when generating documentation?
* Can all unstable APIs be documented that they need this `--cfg` flag to get passed when compiling? This could probably reference the documentation in the next bullet.
* Additionally, can you add documentation about the feature flag? For example how to pass it, when to pass it, how to use `RUSTFLAGS`, etc.

grovesNL:
@alexcrichton could you check the 4 latest commits to see if they address the points in your comment? I'm not sure if we can verify the docs.rs change until it's published

guide/src/web-sys/unstable-apis.md
```diff
@@ -0,0 +1,31 @@
...
+This causes some challenges for `web-sys` because it means `web-sys` would have
+to make breaking API changes whenever the WebIDL changes. It also means that
+previously published `web-sys` versions would be invalid, because the browser
+API may have been changed to match the updated WebIDL.
+
+To avoid frequent breaking changes for unstable APIs, `web-sys` hides all
+unstable APIs through an attribute that looks like:
+
+```rust
+#cfg[web_sys_unstable_apis]
```
alexcrichton: I think the `[` and `]` are misplaced here?

alexcrichton:
Looks great, thanks!

Mind also [updating the doc building](https://github.com/rustwasm/wasm-bindgen/blob/d26068dc6ccfc07d0ab1cd2ead85193ddc3880cd/azure-pipelines.yml#L320-L321) to document unstable APIs?

grovesNL:
Thanks! I've fixed the typo and updated the docs CI step

alexcrichton:
Thanks so much!

PR #1998, opened 2020-02-13
Allow web-sys to emit correct typescript declarations from webidl

bennetthardwick:
This change adds the `NamedAnyref` adapter-type / descriptor, which behaves the same as an `Anyref` but also contains some information about the name. This allows web-sys to output the correct names for types instead of falling back to `any`.

If this feature is the kind of thing that the rustwasm team would like to merge - I'll make sure it's test covered. At the moment I just wanted to get it working and make sure it's something worth working on.

This is currently a breaking change but could be hidden behind a feature flag in web-sys.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -181,6 +181,7 @@ pub struct ImportType {
     pub rust_name: Ident,
     pub js_name: String,
     pub attrs: Vec<syn::Attribute>,
+    pub typescript_name: Option<String>,
     pub doc_comment: Option<String>,
     pub instanceof_shim: String,
     pub is_type_of: Option<syn::Expr>,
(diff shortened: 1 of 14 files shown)


Pauan:
I don't think we should be worrying *too* much about TypeScript breaking changes, since TypeScript's type system is so loose, and it's always possible for people to use `as` to arbitrarily convert from one type to another.

bennetthardwick:
That makes sense :+1: My only concern that was since the types were `any` before, there's a high chance that existing projects have been providing the wrong types, but it sounds like that's not too bad.

alexcrichton:
Thanks for this! Could this be sure to add some tests as well to exercise it and ensure it works?

bennetthardwick:
Hey @alexcrichton, I've added a few typescript tests. I couldn't find anywhere that explicitly tests stuff like `ast::ImportKind` through `adapter2ts`, so I figured just making some in the `typescript-tests` crate would be best. 

Also since all of the webidl stuff is using this, the `webidl-tests` crate is making sure that it all works (at least in terms of wasm). 

Please let me know if there's anywhere else I should write some tests :smile:

alexcrichton:
Looks great to me, thanks!

Issue #1948, opened 2020-01-08
WebXR support?


kettle11:
The WebXR Device API has shipped and is now supported in Chrome! It'd be really cool to write Rust to build WebXR experiences.

### Additional Context

A previous issue was opened for this that was closed to wait until the WebXR Device API stabilized (and it's now stabilized [**Edit:** It is not stabilized, see below discussion] and shipped!): https://github.com/rustwasm/wasm-bindgen/issues/1774

I left some comments on that previous issue with my findings about what needs to be done to get the WebXR IDL added.

Pauan:
Could you give a link to the stabilized spec? I can only find the draft.

kettle11:
You're right. There's only the W3C Working Draft: https://www.w3.org/TR/webxr/

I was blending together "shipped" and "stabilized" in my phrasing, which is my mistake as it doesn't match to the official terminology.

The WebXR Device API has shipped so I don't think there are plans for substantial near term changes to the WebXR Device API.

Pauan:
I just tried accessing WebXR in 81.0.4000.3 (Official Build) dev (64-bit) but it doesn't seem to exist.

Are you sure that it's being shipped *without* needing to enable experimental flags in `chrome://flags`?

kettle11:
It is shipped and on by default for Windows Mixed Reality headsets with Chrome and it's on by default for the Oculus Quest (which is super popular now) with the Oculus Browser, which is based on Chrome.

Other desktop headsets require experimental flags to be enabled, which is unfortunate.

For Oculus on desktop the following flags need to be set:
Oculus hardware support: Enabled
XR device sandboxing: Disabled

Not sure about the ones for other headsets.

Pauan:
And just to clarify, it is enabled by default under `navigator.xr`? No browser prefixes?

kettle11:
Yes, that's correct.

Pauan:
It's disturbing to see Chrome turning into the new IE, pushing through non-standard APIs and bypassing the standards process without even bothering to use vendor prefixes.

But, that's off-topic. As far as this issue goes, I think we shouldn't officially support the API until it's implemented by more browsers.

However, I think we really should have support for unstable APIs (as described [here](https://github.com/rustwasm/wasm-bindgen/pull/1712#issuecomment-519618850)). So we'd accept a PR implementing that.

kettle11:
[...]

Issue #1774, opened 2019-09-19
WebXR WebIDL


kevthecoder:
### Summary

It looks like the upcoming W3C Standard WebXR is getting some support in browsers. I've been trying to implement the Web IDL based on the Draft Spec (which is the version being implemented in browsers, I believe): https://www.w3.org/TR/2019/WD-webxr-20190521/#idl-index. However I can't seem to produce the bindings - do you know anyone who's had success with this?

### Additional Details

I've been following the contribution guide for Web IDL (https://rustwasm.github.io/docs/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html) and have come up with this attempt: https://github.com/kevthecoder/wasm-bindgen/pull/1/files

I don't seem to be getting any bindings appearing in `bindings.rs`, but I'm not seeing any `Unsupported WebIDL interface` warnings when I run:
```
cd crates/web-sys
RUST_LOG=wasm_bindgen_webidl cargo build -vv
```

I appreciate it's early days for the standard so this is probably a low priority, but thought I'd be check if anyone could spot a problem with the attempt above, or if anyone has had success implementing WebXR.

Thanks!

alexcrichton:
If you use the `__WASM_BINDGEN_DUMP_FEATURES` feature of the build script I think the feature should be named `Xr` instead of `XR`, and that should do the trick?

kevthecoder:
Thanks very much @alexcrichton.  Looks like I've got the bindings now.

I'll keep working on a basic example then submit a Pull Request once WebXR stabilises.

kettle11:
WebXR has shipped in Chrome and for the Oculus Browser. It'd be great to see support added here. Any update on this?

kettle11:
To add to this it seems like some modifications need to be made to the IDL to have full WebXR support.

While I didn't dig into what it means (I'm not an expert here) the WebXR IDL features a number of "FrozenArrays" that seem to be not supported. Other IDLs have examples of how to modify the FrozenArray lines, but I'm not sure what the implications are.

This is something that would need to be updated in @kevthecoder 's pull request.

kevthecoder:
@kettle11  Sorry for the delay replying. The work above was based on the "VR-complete" Working Draft. I'll look to update it to the "Candidate Recommendation" draft over the next week so we can merge it in quickly if it gets standardised.

[...]

PR #2000, opened 2020-02-13
Webxrdevice

kevthecoder:
Adds WebIDL for "Candidate Recommendation" version of WebXR (https://www.w3.org/TR/2019/WD-webxr-20191010/).

Supported without flags in Chrome 80, but not supported in Firefox (as of 13 Feb 2020).

To update WebIDL bindings:
```
cd crates/web-sys/
RUST_LOG=wasm_bindgen_webidl __WASM_BINDGEN_DUMP_FEATURES=webxr cargo build -vv
# (had to run the command twice for some reason)
```

Linked issues:
https://github.com/rustwasm/wasm-bindgen/issues/1950
https://github.com/rustwasm/wasm-bindgen/issues/1774
https://github.com/rustwasm/wasm-bindgen/issues/1948


Pauan:
@kevthecoder Thanks, though that isn't a Candidate Recommendation, it's a Draft, so it's not standardized yet.

So this will need to wait for https://github.com/rustwasm/wasm-bindgen/pull/1997 to be implemented first.

kevthecoder:
Thanks @Pauan , sounds like a plan.

kevthecoder:
I've had to disable some properties in the WebIDL for `XRWebGLLayer` to get the bindings to work, otherwise the build command throws an error (with no further error message).

I've also had to apply a workaround for attributes marked with `FrozenArray`, see https://bugzilla.mozilla.org/show_bug.cgi?id=1236777 for more details.

All alterations are commented with a `TODO` to keep track of them.

Manishearth:
Note that it's a draft, but also it's been shipped in multiple browsers now.

Pauan:
@Manishearth Do you have any evidence for that? As far as I know, [only one engine supports it](https://caniuse.com/#feat=webxr), and even then it requires the user to enable an experimental flag, it is not enabled by default. I have verified this by testing in both Chrome and Firefox.

I'm very confused why everybody seems to think this is a widely implemented API.

Manishearth:
@Pauan You don't need a flag to use it in Chrome (but it only works for [some devices](https://immersiveweb.dev/chrome-support.html)). It's shipped on Oculus Browser and Chrome, and Firefox Reality is shipping the same API very soon. Firefox planned to ship in December along with Chrome, but some issues cropped up on the Firefox side delaying it a bit.

Either way, the editors consider the API surface done (and we're gunning for CR soon, but the process takes time), and it would be extremely unlikely for any further changes to be breaking given that there is real world content using this API now. New APIs may of course be tacked on, but that's true of all web specs.

Manishearth:
Basically: if the main concern is breaking changes, I wouldn't be _too_ concerned. If the main concern is just not having anything that's not gone through the full w3c standards pipeline, that's understandable.

Pauan:
@Manishearth Ah, I see, that explains why it wasn't working for me. In that case we should consider stabilizing XR.

alexcrichton:
Want to rebase this on #1997 and we should be able to merge then?

kevthecoder:
> Want to rebase this on #1997 and we should be able to merge then?

No problem, but might be a few days before I get the time. I've been trying to improve the example code too.

Pauan:
Just a note that the process of generating WebIDL will change after https://github.com/rustwasm/wasm-bindgen/pull/2012 lands.

Pauan:
Since https://github.com/rustwasm/wasm-bindgen/pull/2012 has landed now, the process for generating WebIDL is:

1. Put any unstable WebIDL files into the `crates/web-sys/webidls/unstable` folder.

2. Run this command:
   ```
   cargo run --release --package wasm-bindgen-webidl crates/web-sys/webidls crates/web-sys/src/features
   ```

3. Copy the `crates/web-sys/features` file into the `crates/web-sys/Cargo.toml` file (right underneath `[features]`).

4. Use `git add crates/web-sys` to add all the generated files into git.

kevthecoder:
@Pauan @alexcrichton the code is updated to compatible with #2012 now. I like the new system, found it much easier to debug a webidl issue I was having with XrWebGlLayer.

I've tested the new bindings on a WebXr game engine project and they seem to work fine. I tried to make a simplified example application in `examples/webxr`, but there's a small issue I can't seem to fix at the moment (something to do with calling `request_animation_frame`). I'll try to take another look at it with a fresh pair of eyes, but at least the bindings seem to be working ok.

For testing I've been using Chrome with the Mozilla WebXr emulator plugin - https://blog.mozvr.com/webxr-emulator-extension.

alexcrichton:
Nice, thanks for this! This all looks great to me, so any further iteration can happen in-repo!

kevthecoder:
Fantastic, thanks @alexcrichton.

PR #2001, opened 2020-02-14
Add get/set for `TypedArray`s.

derekdreery:
This provides the ability to do

```
let value = array[idx];
array[idx] = value;
```
for typed arrays. They are named with an underscore so thay do not clash with the `set` method, which does something different.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -4884,6 +4884,14 @@ macro_rules! arrays {
             /// input values from a specified array.
             #[wasm_bindgen(method)]
             pub fn set(this: &$name, src: &JsValue, offset: u32);
+
+            /// Gets the value at `idx`, equivalent to the javascript `my_var = arr[idx]`.
+            #[wasm_bindgen(method, structural, indexing_getter)]
+            pub fn get_(this: &$name, idx: u32) -> $ty;
+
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! This feels a bit odd though with the trailing underscores? Is there perhaps a different way we could expose this?

derekdreery:
Hmm, looking at [the mdn page for indexed collections](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections), it seems that currently the two main types for this are `Array` and the various typed arrays. So we can either bikeshed some names like `get_array` or `get_indexed` etc., or have a trait called `IndexedArray` (bikesheddable) with methods `get` and `set`. The `js_sys::Array` type just has the `get` and `set` methods already defined, but we can't do that on typed arrays because the `set` method means something else. Unfortunately, the typed arrays don't inherit from `Array`, so we can't just deref to `Array` either.

I don't think there is an obvious optimal solution, but I think personally I'd go for something like `get_array_like`, with documentation explaining the situation (e.g. saying

> equivalent to `val = arr[idx]` on the javascript side

). I'll implement this in the PR, but am happy to change it to something else if preferred.

Pauan:
I'd be in favor of `get_index` and `set_index`.

derekdreery:
Actually thinking about it again, I agree.

alexcrichton:
Thanks!

PR #2011, opened 2020-02-20
Propagate missing memory argument

### LINKED ISSUE #2010: Regression: can't instantiate wasm-bindgen with instance of WebAssembly.Module
### Describe the Bug

On master wasm-bindgen trying to instantiate wasm-bindgen with a custom instance of WebAssembly.Module (e.g. in a Cloudflare Worker) results in an error due to `maybe_memory` being undefined:

image

If you check the generated code with ESLint, it also warns that variable is not defined anywhere:

image

This appears to be a regression, perhaps caused by a refactor in #1996? (cc @Pauan)

RReverser:
Fixes #2010.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -510,7 +510,7 @@ impl<'a> Context<'a> {
 
         let js = format!(
             "\
-                async function load(module, imports) {{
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oops, thanks!

PR #2016, opened 2020-02-28
Add skip_typescript attribute to prevent .d.ts emit

### LINKED ISSUE #1691: Add an attribute to prevent emitting typescript definitions for a specific #[wasm_bindgen]'d item
### Motivation

Currently, TypeScript support in wasm-bindgen is all-or-nothing: either every single item annotated with #[wasm_bindgen] gets a TypeScript definition, or none of them do.
While the generated bindings are useful, they currently contain lots of `any`s, and it's unlikely wasm-bindgen will be able to produce strict typings for JsValues without lots of additional work like #1197.

This can be partially worked-around in the meantime, with `typescript_custom_section`:
```rust
[wasm_bindgen(typescript_custom_section)]
const TS_FOO_TYPE: &'static str = r#"
export function foo(bar: (x: number) => number): void;
"#;

#[wasm_bindgen]
pub fn foo(bar: js_sys::Function) {
    // call bar, handling errors if the wrong function signature is passed
}
```
However, this will generate *two* TypeScript definitions:
The automatic `export function foo(bar: any): void;` and the custom `export function foo(bar: (x: number) => number): void;`

This does still add utility, since you can get IDE suggestions for what types you should pass, but it doesn't cause TypeScript to *enforce* passing the correct types, since it will still accept `any` as a parameter.

### Proposed Solution

I propose an attribute to prevent emitting TypeScript definitions for a specific #[wasm_bindgen]'d item.
It could look something like `#[wasm_bindgen(typescript = false)]`
This would not autogenerate or emit any typescript bindings for that item.

This would allow developers to provide their own type definitions that replace the autogenerated ones, and allow third-parties to write a crate that provides something like #1197 without it having to be part of wasm-bindgen itself.

My only concern is that this feature encourages maintaining two copies of a function signature, where it would be easy to update either the Rust or TypeScript signature and forget to update the other. However, there's no solution to this problem without much more advanced TypeScript type generation than js_sys and wasm-bindgen currently support.

### Alternatives
A way to emit custom typescript for specific parameters or return types for a function could function in a similar way. I believe this would take more work to implement than simply not emitting TypeScript definitions per-item and letting developers inline the strict definition in a `typescript_custom_section`.

We could instead implement  #1197, but that's a big complicated issue - this feature would allow developers to start using more strict TypeScript types immediately.

joeywatts:
Fixes #1691 by adding support for a `skip_typescript` attribute which disables generation of type definitions for a particular Rust export.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -221,6 +221,7 @@ pub struct Function {
     pub rust_attrs: Vec<syn::Attribute>,
     pub rust_vis: syn::Visibility,
     pub r#async: bool,
+    pub generate_typescript: bool,
 }
 
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
@@ -231,6 +232,7 @@ pub struct Struct {
     pub fields: Vec<StructField>,
     pub comments: Vec<String>,
     pub is_inspectable: bool,
+    pub generate_typescript: bool,
 }
 
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
@@ -243,6 +245,7 @@ pub struct StructField {
     pub getter: Ident,
     pub setter: Ident,
     pub comments: Vec<String>,
+    pub generate_typescript: bool,
 }
 
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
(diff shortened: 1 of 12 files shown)


alexcrichton:
Thanks for this! I think though perhaps instead of `typescript = false` we could do something like `skip_typescript`? I'm not sure how many other things we'll want on the other side of the `=` for now perhaps?

It may also make sense to hold off on this for a bit while https://github.com/rustwasm/wasm-bindgen/pull/2012 is sorted out, since that's a pretty unwieldy PR and may be hard to rebase over this.

joeywatts:
`skip_typescript` seems good to me. If you prefer to wait on that PR, I don't mind rebasing/reworking this once that's in.

alexcrichton:
Thanks!

PR #2021, opened 2020-03-02
Fix `TypedArray::subarray` docs

RReverser:
Looks like these were mistakenly copy-pasted from the `TypedArray::set` method.


alexcrichton:
Thanks!

PR #2026, opened 2020-03-03
Bump to 0.2.59

alexcrichton:
Wow this is a big release!


Pauan:
(The changelog doesn't mention it, but https://github.com/rustwasm/wasm-bindgen/pull/2012 also added in a new `typescript_type` attribute for specifying the TypeScript type of an import.)

PR #2028, opened 2020-03-04
Adding in typescript_type to the js_sys types

Pauan:
This gives proper TypeScript types to all of the `js_sys` types, so they won't show up as `any`.

This is technically a breaking change, but I think that's okay because:

* I expect the breakage to be very minimal since it would only affect people who were passing in the wrong type.

* By default TypeScript doesn't halt compilation when an error happens, so TypeScript errors are really more like warnings.

* If the user's code does break, they can just use `foo as any` to fix it.


alexcrichton:
Nice!

PR #2038, opened 2020-03-10
Fix undefined error in worker.

daxpedda:
I'm unsure what happened here, last time I tried it worked, anyway, this fixes the issue that I introduced in #1938.


alexcrichton:
Thanks!

PR #2057, opened 2020-03-26
try to fix global / modulaized import ns conflict

### LINKED ISSUE #1853: Can't imported aliased names in some circumstances
### Describe the Bug

wasm-bindgen can sometimes deal with name conflicts, e.g. when two identically named items are imported from different modules. However, this behaviour does not trigger when importing items from two identically named namespaces from inside a module and outside it. I suggest to see the example for a better description.

### Steps to Reproduce

1. Create a standard rust wasm library(lib.rs, crate-type, wasm-bindgen dependency).
2. Enter Rust code given below.
3. Rust `wasm-pack build --target nodejs`.

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(module = "fs")]
extern "C" {
	#[wasm_bindgen(js_namespace = window)]
	fn f1();
}

#[wasm_bindgen]
extern "C" {
	#[wasm_bindgen(js_namespace = window)]
	fn f2();
}

#[wasm_bindgen]
pub fn yoyo() {
	f1();
	f2();
}
```

### Expected Behavior

The former import is correctly renamed so that both work. Naturally, `require('fs').window.f1` does not exist and this fails at runtime, but it would work if there were appropriate node modules, omitted for the sake of brevity.

### Actual Behavior

```
[INFO]: Compiling to Wasm...
    Finished release [optimized] target(s) in 0.01s
:-) [WARN]: origin crate has no README
error: failed to generate bindings for import `Value(Bare(JsImport { name: Global { name: "window" }, fields: ["f2"] }))`
    caused by: cannot import `window` from two locations
```

### Additional Context

I originally found the issue when porting a project. As of pustaczek/icie@e8f10b8, it can be also reproduced there by running `wasm-pack build --target nodejs`.

a1trl9:
Hi,

Tried to fix #1853 by prioritizing global imports.

Please let me know if I made any mistakes (:

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -42,6 +42,7 @@ pub struct Context<'a> {
     /// the number of times they've been used, used to generate new
     /// identifiers.
     defined_identifiers: HashMap<String, usize>,
+    global_defined_import_identifiers: HashMap<String, usize>,
 
     exported_classes: Option<BTreeMap<String, ExportedClass>>,
 
@@ -90,6 +91,7 @@ impl<'a> Context<'a> {
             imported_names: Default::default(),
             js_imports: Default::default(),
             defined_identifiers: Default::default(),
+            global_defined_import_identifiers: Default::default(),
             wasm_import_definitions: Default::default(),
             exported_classes: Some(Default::default()),
             config,
@@ -113,7 +115,7 @@ impl<'a> Context<'a> {
         contents: &str,
         comments: Option<&str>,
     ) -> Result<(), Error> {
-        let definition_name = generate_identifier(export_name, &mut self.defined_identifiers);
+        let definition_name = generate_identifier(export_name, &mut self.defined_identifiers, &mut self.global_defined_import_identifiers);
         if contents.starts_with("class") && definition_name != export_name {
             bail!("cannot shadow already defined class `{}`", export_name);
         }
@@ -1945,13 +1947,13 @@ impl<'a> Context<'a> {
 
         let mut name = match &import.name {
             JsImportName::Module { module, name } => {
-                let unique_name = generate_identifier(name, &mut self.defined_identifiers);
+                let unique_name = generate_identifier(name, &mut self.defined_identifiers, &mut self.global_defined_import_identifiers);
                 add_module_import(module.clone(), name, &unique_name);
                 unique_name
             }
 
             JsImportName::LocalModule { module, name } => {
-                let unique_name = generate_identifier(name, &mut self.defined_identifiers);
+                let unique_name = generate_identifier(name, &mut self.defined_identifiers, &mut self.global_defined_import_identifiers);
                 let module = self.config.local_module_name(module);
                 add_module_import(module, name, &unique_name);
                 unique_name
@@ -1965,7 +1967,7 @@ impl<'a> Context<'a> {
                 let module = self
                     .config
                     .inline_js_module_name(unique_crate_identifier, *snippet_idx_in_crate);
-                let unique_name = generate_identifier(name, &mut self.defined_identifiers);
+                let unique_name = generate_identifier(name, &mut self.defined_identifiers, &mut self.global_defined_import_identifiers);
                 add_module_import(module, name, &unique_name);
                 unique_name
             }
@@ -1996,11 +1998,7 @@ impl<'a> Context<'a> {
             }
 
             JsImportName::Global { name } => {
-                let unique_name = generate_identifier(name, &mut self.defined_identifiers);
-                if unique_name != *name {
-                    bail!("cannot import `{}` from two locations", name);
-                }
-                unique_name
+                if name == "default"  { format!("{}{}", name, 1) } else { name.clone() }
             }
         };
         self.imported_names
@@ -2053,6 +2051,7 @@ impl<'a> Context<'a> {
     }
 
     pub fn generate(&mut self) -> Result<(), Error> {
+        self.precache_global_import_identifiers()?;
         for (id, adapter) in crate::sorted_iter(&self.wit.adapters) {
             let instrs = match &adapter.kind {
                 AdapterKind::Import { .. } => continue,
@@ -2079,6 +2078,53 @@ impl<'a> Context<'a> {
             self.process_package_json(path)?;
         }
 
+        Ok(())
+    }
+
+    fn precache_global_import_identifiers(&mut self) -> Result<(), Error> {
+        for (_id, adapter) in crate::sorted_iter(&self.wit.adapters) {
+            let instrs = match &adapter.kind {
+                AdapterKind::Import { .. } => continue,
+                AdapterKind::Local { instructions } => instructions,
+            };
+            let mut call = None;
+            for instr in instrs {
+                match instr.instr {
+                    Instruction::CallAdapter(id) => {
+                        if call.is_some() {
+                            continue
+                        } else {
+                            call = Some(id);
+                        }
+                    }
+                    Instruction::CallExport(_)
+                    | Instruction::CallTableElement(_)
+                    | Instruction::Standard(wit_walrus::Instruction::CallCore(_))
+                    | Instruction::Standard(wit_walrus::Instruction::CallAdapter(_)) => {
+                        continue
+                    }
+                    _ => {}
+                }
+            }
+            if let Some(id) = call {
+                let js = match &self.aux.import_map[&id] {
+                    AuxImport::Value(AuxValue::Bare(js)) => Some(js),
+                    _ => None,
+                };
+                if let Some(js) = js {
+                    match &js.name {
+                        JsImportName::Global { name } => {
+                            let cnt = self.global_defined_import_identifiers.entry(name.clone()).or_insert(0);
+                            if *cnt == 1 {
+                                bail!("cannot import `{}` from two locations", name);
+                            }
+                            *cnt += 1;
+                        }
+                        _ => {}
+                    }
+                }
+            }
+        }
         Ok(())
     }
 
@@ -3165,9 +3211,14 @@ fn check_duplicated_getter_and_setter_names(
     Ok(())
 }
 
-fn generate_identifier(name: &str, used_names: &mut HashMap<String, usize>) -> String {
+fn generate_identifier(name: &str, used_names: &mut HashMap<String, usize>, global_names: &mut HashMap<String, usize>) -> String {
     let cnt = used_names.entry(name.to_string()).or_insert(0);
     *cnt += 1;
+    let global_cnt = (*global_names.entry(name.to_string()).or_insert(0)).clone();
+    // leave index for global import
+    if *cnt == 1 && global_cnt != 0 {
+        *cnt += 1;
+    }
     // We want to mangle `default` at once, so we can support default exports and don't generate
     // invalid glue code like this: `import { default } from './module';`.
     if *cnt == 1 && name != "default" {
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks for the PR!

I'm not super familiar with the issue here (a test would be great!) but it feels like the fix here should be with generating imports as opposed to reordering definitions?

a1trl9:
> Thanks for the PR!
> 
> I'm not super familiar with the issue here (a test would be great!) but it feels like the fix here should be with generating imports as opposed to reordering definitions?

@alexcrichton Thank you for your comment! I think you are right. Will recheck my code and add tests.

a1trl9:
Hi @alexcrichton,

Would like to share two approaches I have tried the last few days for this issue and please don't hesitate to point out my mistakes.

### About Issue
The minimal case, as presented in #1853, is like following:
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(module = "fs")]
extern "C" {
	#[wasm_bindgen(js_namespace = window)]
	fn f1();
}

#[wasm_bindgen]
extern "C" {
	#[wasm_bindgen(js_namespace = window)]
	fn f2();
}

#[wasm_bindgen]
pub fn yoyo() {
	f1();
	f2();
}
```
For now, an error would be thrown: `cannot import window from two locations`. But actually, as one of `window` is imported from module `fs`, we can generate code like:

```javascript
import { window as window2 } from 'fs';
import * as wasm from './wasm_test_bg.wasm';

function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }
/**
*/
export function yoyo() {
    wasm.yoyo();
}

export const __wbg_f1_5e436d22ea2eab51 = typeof window2.f1 == 'function' ? window2.f1 : notDefined('window2.f1');

export const __wbg_f2_a182d819e1c1a65f = typeof window.f2 == 'function' ? window.f2 : notDefined('window.f2');
```

That is, rename imported functions from module automatically, and leave global imports untouched.


### Approach 1
As you suggested, I tried to figure out if it is possible to generate imports which prioritize global imports, so that I don't need to resort it afterwards when generating adapters like I did last week. The main challenge for this approach is, it seems like all imports are processed linearly in a loop, and imports are stored separately in programs, which again are processed and generated linearly. The code structure is like:

```rust
// code from cli-support/src/wit/mod.rs

// get decoded programs    
let programs = extract_programs(module, &mut storage)?;

// try to process program one by one.
for program in programs {
      cx.program(program)?;
}

fn program(...) {
      // nested, process import in a loop
      for import in imports {
           self.import(import)?;
      }
}

fn import(&mut self, import: decode::Import<'_>) -> Result<(), Error> {
      match &import.kind {
            decode::ImportKind::Function(f) => self.import_function(&import, f),
            decode::ImportKind::Static(s) => self.import_static(&import, s),
            decode::ImportKind::Type(t) => self.import_type(&import, t),
            decode::ImportKind::Enum(_) => Ok(()),
     }
}
```
Adapter info and instructions are built by function `import_adapter`, which called by `import_function`, `import_static` and `import_type` (Hope I didn't miss something). 

So, overall, it looks like unless I rewrite functions of how programs decoded and processed (maybe even backend logic such as codegen, encode?), or how imports are processed (e.g. extract all imports first and process them alongside other logic in `program`), I can't get an import list, or adapter list which prioritizing global imports at the beginning. I feel it might be a big change for this project. For example, for now, in the `import_adapter` function, the multi-step process makes an adapter with `AdapterKind::Local {..}` always follows the one with `AdapterKind::Import {..}`. Thus, some refactoring like above might alter the order of the whole adapter list, not only the import part.

### Approach 2
For approach 2, I tried to leave all generation logic untouched, and not reorder adapters. Instead, build a new hashmap: `global_defined_import_identifiers` to store all global import keys ahead of import name generation. Now for function `generate_identifier`, two more parameters are required (one for `global_defined_import_identifiers` and another is simply a flag to show if the processed name is from a global import). Inside the function, two kinds of import name are handled **separately**. For non-global one, an index space would be left if any global imports with the same name exist. For global one, it now only considers conflicts with other global ones (but does this case exist now?) You can see my recent commits for more details.

This approach is not elegant. Any advice is welcomed if you think there are much better solutions, or this one actually misses some cases.

I also added one minimal test in `cli` crate, and tested some more cases manually, for example, with `default`, multiple global imports from the same namespace, etc., to double-check current behaviour is not broken:

#### default
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = Math)]
    fn random() -> f64;
}

#[wasm_bindgen(module = "fs")]
extern "C" {
	#[wasm_bindgen(js_namespace = window)]
	fn f1();

	#[wasm_bindgen(js_namespace = default)]
	fn f2();
}

#[wasm_bindgen]
pub fn test() {
    f1();
    f2();
}
```

```javascript
import { window, default as default1 } from 'fs';
import * as wasm from './wasm_test_bg.wasm';

function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }
/**
*/
export function test() {
    wasm.test();
}

export const __wbg_f1_5e436d22ea2eab51 = typeof window.f1 == 'function' ? window.f1 : notDefined('window.f1');

export const __wbg_f2_a4247dfb755d7558 = typeof default1.f2 == 'function' ? default1.f2 : notDefined('default1.f2');
```

#### multiple global imports with same namespace

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = Math)]
    fn random() -> f64;
}

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = Math)]
    fn log(a: f64) -> f64;
}

#[wasm_bindgen]
pub fn test() {
    random();
    log(3.2);
}
```

```javascript
import * as wasm from './wasm_test_bg.wasm';

function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }
/**
*/
export function test() {
    wasm.test();
}

export const __wbg_random_cba1f145a042ae34 = typeof Math.random == 'function' ? Math.random : notDefined('Math.random');

export const __wbg_log_7bde4e1bb51408eb = typeof Math.log == 'function' ? Math.log : notDefined('Math.log');

```

#### Same name between import namespace and export
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(module = "fs")]
extern "C" {
	#[wasm_bindgen(js_namespace = window)]
	fn f1();
}

#[wasm_bindgen]
pub fn window() {
    f1();
}
```


```javascript
import { window } from 'fs';
import * as wasm from './wasm_test_bg.wasm';

function notDefined(what) { return () => { throw new Error(`${what} is not defined`); }; }
/**
*/
function window2() {
    wasm.window();
}
export { window2 as window };

export const __wbg_f1_5e436d22ea2eab51 = typeof window.f1 == 'function' ? window.f1 : notDefined('window.f1');
```

I am looking forward to any advice.

alexcrichton:
Thanks for the update! I think that a solution along the lines of this is the way to go, either by reordering how imports are bound or precalculating information about imports rather than generating them on the fly.

Could this be made a method instead of a standalone function to avoid passing maps around? Additionally could you explain the counts on globals a bit more? There's some logic around numbering that I don't fully understand myself.

a1trl9:
@alexcrichton Thanks for your comment.

UPDATE:
- Now `generate_identifier` has been moved into `Context` so we don't need to pass either `defined_identifiers` or `global_defined_import_identifiers`. One issue here is since now `generate_identifier` is a private method, I didn't figure out how to test it directly like before. I added more related tests into `import.rs` / `import.js` to check if `js_namespace` and `js_name` conflicts could be handled correctly. Hope it could mitigate the drawback.
- Some comments added for numbering.

crates/cli-support/src/js/mod.rs
```diff
@@ -3118,6 +3163,37 @@ impl<'a> Context<'a> {
     fn adapter_name(&self, id: AdapterId) -> String {
         format!("__wbg_adapter_{}", id.0)
     }
+
+    fn generate_identifier(&mut self, name: &str, global_import: bool) -> String {
+        let cnt = (&mut self.defined_identifiers)
```
alexcrichton, line 3168: I think this leading `&mut` may not be necessary?

crates/cli-support/src/js/mod.rs
```diff
@@ -2082,6 +2084,49 @@ impl<'a> Context<'a> {
...
+            }
+            if let Some(id) = call {
+                let js = match &self.aux.import_map[&id] {
+                    AuxImport::Value(AuxValue::Bare(js)) => Some(js),
+                    _ => None,
+                };
+                if let Some(js) = js {
+                    match &js.name {
+                        JsImportName::Global { name } => {
+                            self.global_defined_import_identifiers
```
alexcrichton, line 2118: I wonder if this extra map is necessary? Could this just call `generate_identifier` for `name` here? 

Once we generate an identifier for each global name I think every other name naturally gets a numerical suffix?

crates/cli-support/src/js/mod.rs
```diff
@@ -1920,6 +1922,18 @@ impl<'a> Context<'a> {
         require_class(&mut self.exported_classes, name).wrap_needed = true;
     }
 
+    fn add_module_import(&mut self, module: String, name: &str, actual: &str) {
+        let rename = if name == actual {
+            None
+        } else {
+            Some(actual.to_string())
+        };
+        (&mut self.js_imports)
```
alexcrichton, line 1931: I think the leading `&mut` can be removed here

crates/cli-support/src/js/mod.rs
a1trl9: Thank you for your comment.

I feel it might not work in some cases. Since function `import_name` is called on the fly during iteration, some non-global import might be processed first and therefore produces and stores a name with a numerical suffix if name conflict with another global import occurs. Then afterwards when processing the global one it would be hard to figure out if the cached name in the hashmap comes from a global or non-global one (for the first case, an error would be thrown).

Handling global name conflicts directly here might not work as well. Since at this stage, we still don't know if the import is actually field by dot-access rather than name itself. I guess this is why the following if-condition is tested at the beginning of `import_name`:

```rust
        if let Some(name) = self.imported_names.get(&import.name) {
            let mut name = name.clone();
            for field in import.fields.iter() {
                name.push_str(".");
                name.push_str(field);
            }
            println!("{:?}", name);
            return Ok(name.clone());
        }
```

alexcrichton: Er sorry but what I mean is that you're already iterating over the global names explicitly first here, so instead of adding a second map to track this couldn't this pre-populate the map with only global names during this first pass?

a1trl9: Emm... my bad not make it clear. 
Assuming we have imports like [Module { module: 'xxx_module', name: "Math"}, xxx, xxx, xxx, Global {name: "Math"}].
If **here** we count the global name `Math` in `defined_identifiers` instead of a new one. Then during calling `import_name` one by one, the `Math` from the module `xxx_module` would be processed first and increment *cnt of `Math` again in the same map:

```Rust
            JsImportName::Module { module, name } => {
                // here, generate_identifier would increment *cnt for name `Math`
                let unique_name = self.generate_identifier(name, false);
                self.add_module_import(module.clone(), name, &unique_name);
                unique_name
            }

```


Afterwards, when processing the global import, the following code would break:

```Rust
            JsImportName::Global { name } => {
                // current code, for clarify.
                // this wouldn't work surely, we have generated global names before.
                let unique_name = self.generate_identifier(name, true);
                if unique_name != *name {
                    bail!("cannot import `{}` from two locations", name);
                }
                // but modifying to the following code still not work
                // since non-global import `Math`
                // has been inserted to the same map, *cnt = 2. However the name conlifcts
                // are not from two global imports, bail is not expected.
                let cnt = self.defined_identifiers.entry(name.to_string()).or_insert(0)
                if *cnt > 1 {
                   bail!("cannot import `{}` from two locations", name);
                }
                unique_name
            }
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2082,6 +2076,47 @@ impl<'a> Context<'a> {
...
+            }
+            if let Some(id) = call {
+                let js = match &self.aux.import_map[&id] {
+                    AuxImport::Value(AuxValue::Bare(js)) => Some(js),
+                    _ => None,
+                };
+                if let Some(js) = js {
+                    match &js.name {
+                        JsImportName::Global { name } => {
+                            self.generate_identifier(name);
```
alexcrichton, line 2110: I think this will still want to forward to `import_name` which does caching and error handling, it should just only call that method with `JsImportName::Global` values.

a1trl9: sry but I don't quite understand what you suggest by "it should just only call that method with `JsImportName::Global` values." Could you please explain a little bit more?

alexcrichton: Oh so my point is that I don't think we want to call `generate_identifier` as a bare function since it always generates a new identifier. This means that if the same global is referenced multiple times it'll throw off generation. Instead this should use `import_name` which has error handling and caching which should be used instead of the raw `generate_identifier`

alexcrichton:
Thanks again!

Issue #2053, opened 2020-03-25
Copied doc comments are escaped in .d.ts


0xd4d:
Useless backslashes are added to the copied doc comments, see below.

wasm-bindgen = "0.2.59"

build: `wasm-pack build`

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct MyStruct;

#[wasm_bindgen]
impl MyStruct {
    /// This "is" \ some text: it's
    /// ```
    /// let a = b"\x12\x34";
    /// ```
    pub fn meth(&self) {}
}
```

Note all the extra backslashes below:

```ts
/* tslint:disable */
/* eslint-disable */
export class MyStruct {
  free(): void;
/**
* This \"is\" \\ some text: it\'s
* ```
* let a = b\"\\x12\\x34\";
* ```
*/
  meth(): void;
}
```

alexcrichton:
Thanks for the report! I'm not too entirely sure where this is coming from myself. It seems like this has to do with some attribute escaping somewhere, and I think we'd just need to track down what's escaped where and see if we can get an unescaped version. Not really a great answer, but this is definitely a bug worth fixing!

Issue #2052, opened 2020-03-25
Missing doc comments in .d.ts file


0xd4d:
Many doc comments aren't copied to the .d.ts file, eg. enums, enum items, structs.

wasm-bindgen = "0.2.59"

build: `wasm-pack build`

```rust
use wasm_bindgen::prelude::*;

/// This is MyStruct
#[wasm_bindgen]
pub struct MyStruct {
    /// This is a field
    pub field1: u32,
}

#[wasm_bindgen]
impl MyStruct {
    /// This is the constructor
    #[wasm_bindgen(constructor)]
    pub fn new(field1: u32) -> Self {
        Self { field1 }
    }

    /// This is a method
    pub fn meth(&self) {}

    /// This is a getter
    #[wasm_bindgen(getter)]
    pub fn value(&self) -> bool {
        true
    }

    /// This is a setter
    #[wasm_bindgen(setter)]
    pub fn set_value(&mut self, value: bool) {}
}

/// This is MyEnum
#[wasm_bindgen]
pub enum MyEnum {
    /// This is Zero
    Zero,
    /// This is One
    One,
}
```

Note all the missing comments. I expect the getter's docs are copied if there's a getter, else the setters docs are copied.

- no enum items docs
- no getter docs
- no field docs
- no struct docs
- free() is auto-generated, maybe add some docs there too?

```ts
/* tslint:disable */
/* eslint-disable */
/**
* This is MyEnum
*/
export enum MyEnum {
  Zero,
  One,
}
export class MyStruct {
  free(): void;
/**
* This is the constructor
* @param {number} field1 
*/
  constructor(field1: number);
/**
* This is a method
*/
  meth(): void;
  field1: number;
  value: boolean;
}
```

alexcrichton:
Thanks for the great test case here! Not everything has the docs plumbed all the way through (it seems like we're definitely inconsistent!), but definitely would be a good bug fix to treat docs consistently in all these locations.

PR #2070, opened 2020-04-06
Copy more doc comments to JS/TS files, unescape comments

0xd4d:
Closes https://github.com/rustwasm/wasm-bindgen/issues/2053
Closes https://github.com/rustwasm/wasm-bindgen/issues/2052

Unescapes comments. Looks like it's coming from proc-macro2, it returns a quoted + escaped string. I couldn't find an unescape function so I had to write it myself.

Copies more comments from rust to js/ts files. I think all comments are copied now.

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -3,9 +3,8 @@ use crate::descriptors::WasmBindgenDescriptorsSection;
 use crate::intrinsic::Intrinsic;
 use crate::{decode, PLACEHOLDER_MODULE};
 use anyhow::{anyhow, bail, Error};
-use std::char;
 use std::collections::{HashMap, HashSet};
-use std::str::{self, Chars};
+use std::str;
 use walrus::MemoryId;
 use walrus::{ExportId, FunctionId, ImportId, Module};
 use wasm_bindgen_shared::struct_function_export_name;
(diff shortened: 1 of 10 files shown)


crates/cli-support/src/wit/mod.rs
```diff
@@ -1513,7 +1520,77 @@ fn verify_schema_matches<'a>(data: &'a [u8]) -> Result<Option<&'a str>, Error> {
...
     comments
         .iter()
-        .map(|s| s.trim_matches('"'))
+        .map(|&s| try_unescape(s).unwrap_or_else(|| s.to_owned()))
         .collect::<Vec<_>>()
         .join("\n")
 }
+
+// Unescapes a quoted string. char::escape_debug() was used to escape the text.
+fn try_unescape(s: &str) -> Option<String> {
```
alexcrichton, line 1529: I'm personally a bit confused on why we need a method like this, can you help me understand where this is coming from? Where do we acquire an escaped string from, and would it be possible to acquire an unescaped string instead?

0xd4d: https://github.com/rustwasm/wasm-bindgen/blob/b9f78aba5765aceac2e4cc3d67ba95a312f4c1cc/crates/macro-support/src/parser.rs#L1320-L1324

That's the code that gets a quoted + escaped string from `proc_macro2`. I couldn't find an unescape function anywhere. If you know of one, I'll replace this func with that one.

alexcrichton: Can the un-quoting/etc happen around there? I don't think it should be happening in the CLI but rather at the source here if possible.

0xd4d: It's been moved now. The old code used to unquote the string, but it's also been moved. Let me know if that's considered a breaking change and I'll restore it.

alexcrichton:
Thanks!

Issue #2072, opened 2020-04-06
Error when using js_name with both getter and setter


0xd4d:
### Describe the Bug

It's not possible to rename both getter & setter with js_name=xxx

### Steps to Reproduce

build: `wasm-pack build`

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct MyStruct;

#[wasm_bindgen]
impl MyStruct {
	#[wasm_bindgen(getter)]
	#[wasm_bindgen(js_name = "myProp")]
    pub fn my_prop(&self) -> u32 {
        0
    }

    #[wasm_bindgen(setter)]
	#[wasm_bindgen(js_name = "myProp")]
    pub fn set_my_prop(&mut self, _value: u32) {}
}
```

### Expected Behavior

I expect the property to be named `myProp`

### Actual Behavior

I get this compile error:

```
error: symbol `mystruct_myProp` is already defined
 --> src\lib.rs:6:1
  |
6 | #[wasm_bindgen]
  | ^^^^^^^^^^^^^^^

error: aborting due to previous error
```

Changing the setter's name to `set_myProp` creates two properties:

```rust
    #[wasm_bindgen(setter)]
	#[wasm_bindgen(js_name = "set_myProp")]
    pub fn set_my_prop(&mut self, _value: u32) {}
```

A workaround is to rename the rust methods to `myProp` and `set_myProp` respectively.

Pauan:
You have to use the same attribute, like this:

```rust
#[wasm_bindgen]
impl MyStruct {
    #[wasm_bindgen(getter, js_name = "myProp")]
    pub fn my_prop(&self) -> u32 {
        0
    }

    #[wasm_bindgen(setter, js_name = "myProp")]
    pub fn set_my_prop(&mut self, _value: u32) {}
}
```

0xd4d:
Did you try that code? I'm getting the exact same error message.

Pauan:
@0xd4d Hmm, odd, it did work when I tested it. It seems to work if you have a getter or setter, but not both. So that's definitely a bug with wasm-bindgen.

0xd4d:
Seems to have been fixed by https://github.com/rustwasm/wasm-bindgen/pull/2074

PR #2074, opened 2020-04-07
try to fix js_name error when both `getter` and `setter` used

a1trl9:
Related to #2072 

The error may come from Line 492 in codegen.rs:

```rust
        (quote! {
            #(#attrs)*
            #[allow(non_snake_case)]
            #[cfg_attr(
                all(target_arch = "wasm32", not(target_os = "emscripten")),
                export_name = #export_name, // this line...
            )]
            #[allow(clippy::all)]
            pub extern "C" fn #generated_name(#(#args),*) -> #projection::Abi {
                #start_check
                // Scope all local variables to be destroyed after we call the
                // function to ensure that `#convert_ret`, if it panics, doesn't
                // leak anything.
                let #ret = {
                    #(#arg_conversions)*
                    #receiver(#(#converted_arguments),*)
                };
                #convert_ret
            }
        })
        .to_tokens(into);
```
When both `getter` and `setter` are used, a name conflict occurs since `export_name` for both are the same.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -330,12 +330,6 @@ impl Function {
     pub fn infer_setter_property(&self) -> Result<String, Diagnostic> {
         let name = self.name.to_string();
 
-        // if `#[wasm_bindgen(js_name = "...")]` is used then that explicitly
-        // because it was hand-written anyway.
-        if self.renamed_via_js_name {
-            return Ok(name);
-        }
-
         // Otherwise we infer names based on the Rust function name.
         if !name.starts_with("set_") {
             bail_span!(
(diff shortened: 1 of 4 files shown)


alexcrichton:

a1trl9:
@alexcrichton  6 tests added for wasm::getters_and_setters as follows:
8 - no getter/setter field name, same js_name, different rust name.
9 - no getter/setter field name, same js_name, same rust name (except set prefix).
10 - same getter/setter field name, same js_name, different rust name.
11 - different getter/setter field name, samd js_name, different rust name.
12 - same getter/setter field name, same js_name, same rust name (except set prefix).
13 - different getter/setter field name, samd js_name, same rust name (except set prefix).

PR #2083, opened 2020-04-16
Update raytracing example browser support

chinedufn:
Raytracing example works fine for me on Chrome 80

### DIFF
--- a/guide/src/examples/raytrace.md
+++ b/guide/src/examples/raytrace.md
@@ -6,15 +6,13 @@
 [code]: https://github.com/rustwasm/wasm-bindgen/tree/master/examples/raytrace-parallel
 
 **This is an unstable and experimental example** of using threads with
-WebAssembly and Rust, culminating in a parallel raytracer demo. Current
-requirements for viewing this demo are:
+WebAssembly and Rust, culminating in a parallel raytracer demo. The browser requirements are:
 
-* Firefox Nightly - other browsers haven't implemented the proposed WebAssembly
-  features yet.
-* `SharedArrayBuffer` is enabled in `about:config` in Firefox
-
-This demo may also break over time as Firefox updates, but we'll try to keep it
-working!
+* Firefox Nightly 
+  - `SharedArrayBuffer` is enabled in `about:config` in Firefox
+* Google Chrome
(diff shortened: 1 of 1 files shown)


bennetthardwick:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#Browser_compatibility) `SharedArrayBuffer` has been available in Chrome / Chromium since version 68.

chinedufn:
Yo!

It's not just `SharedArrayBuffer` that's required. Also needs an implementation of the `WebAssembly` threads proposal.

For example, in Chrome 70 it was behind a flag.

image

via - https://developers.google.com/web/updates/2018/10/wasm-threads

chinedufn:
According to https://www.chromestatus.com/feature/5724132452859904 worked based threads have been enabled by default since Chrome 74 but I don't recall the demo working back then.

Definitely works in 80, _might_ work in 79 but I didn't try.

chinedufn:
Updated the wording to remove the version number so that I don't have to go through and check when the demo started working.

guide/src/examples/raytrace.md
```diff
@@ -6,15 +6,15 @@
...
+WebAssembly and Rust, culminating in a parallel raytracer demo. The browser requirements are:
 
-* Firefox Nightly - other browsers haven't implemented the proposed WebAssembly
-  features yet.
-* `SharedArrayBuffer` is enabled in `about:config` in Firefox
-
-This demo may also break over time as Firefox updates, but we'll try to keep it
+* Firefox Nightly 
+  - `SharedArrayBuffer` is enabled in `about:config` in Firefox
+  - This demo may also break over time as Firefox updates, but we'll try to keep it
```
alexcrichton, line 13: I think this can be removed now as well

chinedufn: Removed `This demo may also break over time as Firefox updates, but we'll try to keep it`

Didn't remove the shared array buffer bit since my Firefox 75 can't run the demo (I don't have any flags enabled)

alexcrichton:
👍

PR #2086, opened 2020-04-16
No longer fail if env vars do not exist.

crjohns:
This change prevents the ShortHash from failing if environment variables
CARGO_PKG_NAME and CARGO_PKG_VERSION are not present. This allows the
wasm_bindgen macro to work on platforms that do not support environment
variables and build systems that use rustc directly.

### DIFF
--- a/crates/backend/src/util.rs
+++ b/crates/backend/src/util.rs
@@ -136,12 +136,15 @@ impl<T: Hash> fmt::Display for ShortHash<T> {
         // hashing for a lot of symbols.
         if !HASHED.load(SeqCst) {
             let mut h = DefaultHasher::new();
+
+            // If the environment variables cannot be found, use defaults.
             env::var("CARGO_PKG_NAME")
-                .expect("should have CARGO_PKG_NAME env var")
+                .unwrap_or_else(|_| "UNKNOWN".to_string())
                 .hash(&mut h);
             env::var("CARGO_PKG_VERSION")
-                .expect("should have CARGO_PKG_VERSION env var")
+                .unwrap_or_else(|_| "0.0.0".to_string())
                 .hash(&mut h);
+
             // This may chop off 32 bits on 32-bit platforms, but that's ok, we
             // just want something to mix in below anyway.
             HASH.store(h.finish() as usize, SeqCst);


benbrittain:
I didn't look too closely at this, but I'm fairly certain this is a graceful degradation instead of breaking functionality

crjohns:
I think I missed another unwrap of env vars here: https://github.com/rustwasm/wasm-bindgen/blob/master/crates/backend/src/encode.rs#L54

Trying to see if similar defaults there would work OK.

alexcrichton:
Thanks for the PR! I don't mind doing this if it helps unlock some use cases for you, but FWIW these are non-optional for wasm-bindgen to work correctly in the fullest extent. If these lines are removed entirely you'll see a few linker errors with `cargo test --target wasm32-unknown-unknown --test wasm` because of the symbol conflicts that are avoided with these variables.

Ideally, if possible, we'd find a different way to solve the underlying issue without using these env vars.

crjohns:
Thanks for the feedback!

It looks like the issue in util.rs is simply uniqueness. Would it be acceptable to fall back to timestamp on the first run of the macro if the env variables are unset?

The issue in encode.rs seems like it may have more subtle implications. crate_name seems to be used only for uniqueness again, but the manifest dir is used to resolve absolute imports relative to the crate root for interning. What are the implications of setting that to a default if the env var is not set?

alexcrichton:
I'd personally still be pretty hesitant to switch to something like timestamps. While that would likely solve the original issue (although not guaranteed necessarily with concurrent compilations, just highly unlikely), it introduces a vector for nondeterministic builds because the build time is encoded into the artifact.

I'm not entirely sure what the other locations are used for, but I suspect largely for inclusion of JS itself. It might be fine to hobble along without them or with false/unknown values. TBH though this crate wasn't really designed for a non-Cargo build system...

benbrittain:
@crjohns The right thing to do is probably fork this repository so that it can have a contract with GN not cargo.

crjohns:
Ack. I have a workaround for now for our use case. I'll close this out and consider what could be upstreamed.

Thank you both for the help!

PR #2090, opened 2020-04-18
Fix #2089

### LINKED ISSUE #2089: Can't import aliasing workspace in a module and a global getter
### Describe the Bug

The first binding is a function inside a namespace inside a module, like trying to import [vscode.window.createOutputChannel]. The second binding is a global getter like [get_window from js-sys]. When the namespace and the getter have conflicting names, wasm-bindgen fails to generate bindings.

Bug also reproduces on git master, so it's not the case of #2057 not being released in 0.2.61 yet. I have used a git dependency both in Cargo.toml and during cargo install, and #1853 is indeed fixed, but this isn't.

### Steps to reproduce

1. Create a standard rust wasm library (lib.rs, crate-type, wasm-bindgen dependency).
2. Enter Rust code given below.
3. Rust `wasm-pack build --target nodejs`.

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(module = vscode)]
extern "C" {
    #[wasm_bindgen(js_namespace = window)]
    fn f1();
}

#[wasm_bindgen]
extern "C" {
    type Global;

    #[wasm_bindgen(getter, static_method_of = Global, js_class = window, js_name = window)]
    fn f2();
}

#[wasm_bindgen]
pub fn yoyo() {
    f1();
    Global::f2();
}
```

### Expected Behavior

Bindings to both functions are generated correctly. 

### Actual Behavior

During `wasm-pack build --target nodejs`, wasm-bindgen exists with the following error:

```
error: failed to generate bindings for import of `__wbindgen_placeholder__::__wbg_window_d7364fcea086330e`

Caused by:
    cannot import `window` from two locations
```

### Additional Context

In this case (vscode.window and window from js-sys), this can be worked around by forking wasm-bindgen and deleting the offending import from js-sys.

[vscode.window.createOutputChannel]: https://github.com/pustaczek/icie/blob/ac44353bce35c7e617189d5607851416391649ad/vscode-sys/src/lib.rs#L366-L367
[get_window from js-sys]: https://github.com/rustwasm/wasm-bindgen/blob/4900732f60e9f8e056b969001249c7c463f69fe4/crates/js-sys/src/lib.rs#L4743-L4744

a1trl9:
Fix #2089.

The case provided there could be reduced to:

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(module = vscode)]
extern "C" {
    #[wasm_bindgen(js_namespace = window)]
    fn f1();
}

#[wasm_bindgen]
extern "C" {
    type window;

    #[wasm_bindgen(getter, static_method_of = window)]
    fn f2();
}

#[wasm_bindgen]
pub fn yoyo() {
    f1();
    window::f2();
}
```

Seems `prestore_global_import_identifiers` didn't exhaustively match all possible `JsImport` so that `StructuralClassGetter` doesn't work here.

A test added.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2097,7 +2097,19 @@ impl<'a> Context<'a> {
     fn prestore_global_import_identifiers(&mut self) -> Result<(), Error> {
         for import in self.aux.import_map.values() {
             let js = match import {
-                AuxImport::Value(AuxValue::Bare(js)) => js,
+                AuxImport::Value(AuxValue::Bare(js))
+                | AuxImport::Value(AuxValue::ClassGetter(js, ..))
+                | AuxImport::Value(AuxValue::Getter(js, ..))
+                | AuxImport::Value(AuxValue::ClassSetter(js, ..))
+                | AuxImport::Value(AuxValue::Setter(js, ..))
+                | AuxImport::ValueWithThis(js, ..)
+                | AuxImport::Instanceof(js)
+                | AuxImport::Static(js)
+                | AuxImport::StructuralClassGetter(js, ..)
+                | AuxImport::StructuralClassSetter(js, ..)
+                | AuxImport::IndexingGetterOfClass(js)
+                | AuxImport::IndexingSetterOfClass(js)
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

PR #2095, opened 2020-04-21
Adding in some TryFrom impls

Pauan:
Ideally we would have `TryFrom<JsValue>` impls for everything, but we currently can't have them for `Option<T>` and structs.

It's also missing impls for `Vec<T>`, `Box<[T]>`, and `char`, because that requires some sort of integration with `js_sys` (which is a much bigger change).


Pauan:
Hmmm, it seems there's way too many conflicting impls, because of the way that `TryInto` and `From` are defined. So I'll need to refactor this a lot.

PR #2098, opened 2020-04-21
Improving the code generation for catch

Pauan:
Right now every `catch` function generates the minified wrapper `try{}catch(e){h(e)}`. This PR changes it so that it now generates the minified wrapper `h()`. This saves 16 bytes per `catch` function.

I did benchmark tests, the performance was identical (any differences seemed to be noise).


alexcrichton:
Nice idea!

Issue #2097, opened 2020-04-21
How to get navigator.clipboard?


vojta7:
### Summary

I would like to access [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard), but I don't see it mentioned anywhere in the documentation and the following code fails.

```rust
window().navigator().clipboard()
```

with this error

```
no method named `clipboard` found for struct `web_sys::features::gen_Navigator::Navigator` in the current scope
```

### Additional Details

I tried to enable experimentall features, but it had no effect.

Pauan:
@vojta7 `Clipboard` is an unstable API. We only add unstable APIs on an ad-hoc basis, so we would gladly accept a pull request adding it in.

PR #2100, opened 2020-04-22
Clipboard API

vojta7:
Adds Clipboard API (https://www.w3.org/TR/2019/WD-clipboard-apis-20190605/)

Linked issues:
#2097


vojta7:
Some tests are failing because there is already a specified subset of Clipboard API in [ClipboardEvent.webidl](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/web-sys/webidls/enabled/ClipboardEvent.webidl). To my understanding, it is part of the same working draft, so it should be unstable.

How should I resolve this conflict? Remove old specification and keep only content of my pull request (if so in unstable or enabled?), or just ignore it?

alexcrichton:
If it was an already-unstable API I think it's fine to delete the original WebIDL in favor of the new unstable WebIDL

vojta7:
I don't understand why is the remaining test failing. It seems just like the wrong order of declarations, can I fix it somehow?

alexcrichton:
This could perhaps be instability in the bindings generator (maybe a hash map somewhere?) or perhaps a merge that needs to happen before regenerating?

Pauan:
@vojta7 You probably just have a different version of Rust, make sure your Rust is up to date (`rustup update`) and that you're on the Stable channel. You should also try running `cargo update` in the wasm-bindgen repository.

vojta7:
@Pauan I tried your suggestions, but the generated code is still the same.

vojta7:
I did also try to rebase onto the current master (last force push), but the tests are still failing.

Pauan:
@vojta7 Okay, then it's probably caused by non-determinism in the bindings generator, let me take a look.

Pauan:
@vojta7 Since https://github.com/rustwasm/wasm-bindgen/pull/2101 was merged, could you rebase and try again?

vojta7:
Thank you for your time :), it did help.

I also think that #1978 may have failed due to the same nondeterminism.

Pauan:
Thanks!

PR #2103, opened 2020-04-26
No longer error for npm dependencies with web

### LINKED ISSUE #2065: web target should allow for importing npm modules
When using the following code with `--target web`...

```rust
#[wasm_bindgen(module = "foo")]
extern "C" {
    fn foo();
}
```

...it gives the following error:

```
error: NPM dependencies have been specified in `package.json` but this is only compatible with the `bundler` and `nodejs` targets
```

This error is confusing, but more importantly we should allow for npm modules with `--target web` because:

1. The [Rollup plugin](https://github.com/wasm-tool/rollup-plugin-rust#readme) uses `--target web` internally because it cannot use `--target bundler`. Rollup can handle npm modules perfectly fine, so this should be allowed.

2. There are plans to have web native modules, such as [`std::kv-storage`](https://developers.google.com/web/updates/2019/03/kv-storage) (which was ultimately removed, but there will be other native modules in the future).

3. By using [import maps](https://github.com/WICG/import-maps) it is possible to load npm modules in the browser.

Pauan:
Fixes #2065

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -3013,10 +3013,10 @@ impl<'a> Context<'a> {
     }
 
     fn process_package_json(&mut self, path: &Path) -> Result<(), Error> {
-        if !self.config.mode.nodejs() && !self.config.mode.bundler() {
+        if self.config.mode.no_modules() {
             bail!(
                 "NPM dependencies have been specified in `{}` but \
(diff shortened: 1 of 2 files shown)


Pauan:
@alexcrichton The build error isn't from this PR, it's actually because of a new version of Stable which added in some new warnings. It looks like we'll need to change the implementation of `IntoWasmAbi` for closures.

alexcrichton:
👍

PR #2108, opened 2020-04-28
be more precise how to open the example in the browser (needs serving)

mike-kfed:
just opening the index.html file from the local filesystem is not enough, it needs to be served from a webserver, updated docs accordingly


alexcrichton:
Thanks!

PR #2109, opened 2020-04-28
Websockets binary msgs

mike-kfed:
I struggled to understand how binary messages over websockets are supposed to be handled using web_sys::WebSocket - after a while I found all info necessary, so I figured I add a compilation of text and binary message type sending/handling to the websockets example, hopefully helping others. Had to add js-sys as dependency I hope that is okay.


alexcrichton:
Looks great, thanks!

PR #2115, opened 2020-04-29
Fixing require to be ignored by Webpack

Pauan:
Some bundlers (such as Webpack) will process `require` even if it's inside of a conditional branch. Since we only want the `require` to run in Node, we want Webpack to [completely ignore it](https://github.com/webpack/webpack/issues/8826). Using `module.require` seems to be the best way to do that.

I would really rather just remove the `require` completely, but [NodeJS 10 is supported until 2021-04-01](https://en.wikipedia.org/wiki/Node.js#Releases), so we can't remove it until then.

Fixes https://github.com/rustwasm/wasm-bindgen/issues/2116

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1247,7 +1247,7 @@ impl<'a> Context<'a> {
             self.global(&format!(
                 "
                     const l{0} = typeof {0} === 'undefined' ? \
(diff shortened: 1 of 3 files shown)


BenoitDel:
If this PR is included in was-bindgen v0.2.61. The webpack compilation error is still there.

Pauan:
@BenoitDel It is not included in 0.2.61

PR #2124, opened 2020-05-06
Support name alias on type directly for methods

### LINKED ISSUE #2040: static_method_of shows unexpected behavior
### Describe the Bug

Take this code:

```Rust
#[wasm_bindgen(module = "my-great-module")]
extern "C" {
    pub type Module;

    #[wasm_bindgen(static_method_of = Module)]
    pub fn init();
}
```

### Expected Behavior

Generated index.js:

```js
import { default as default1 } from 'my-great-module';
```

### Actual Behavior

```js
import { Module } from 'my-great-module';
```

### Additional Context

The problem is that `static_method_of` binds to the type name given instead of the `js_name` defined for that type. `static_method_of = default` also doesn't work, because the type `default` is not known.

This produces the expected behavior:

```Rust
#[wasm_bindgen(module = "my-great-module")]
extern "C" {
    pub type Module;

    #[wasm_bindgen(static_method_of = Module, js_class = "default")]
    pub fn init();
}
```

but this double specification shouldn't be necessary.

### LINKED ISSUE #2048: js-sys v0.3.36 - error[E0603]: struct `Memory` is private
### Summary

I am getting below error when I try to compile my wasm program with **wasm-pack build**
Not  sure where is the issue !

```
 Compiling js-sys v0.3.36
error[E0603]: struct `Memory` is private
    --> /home/naveen/.cargo/registry/src/github.com-1ecc6299db9ec823/js-sys-0.3.36/src/lib.rs:4920:60
     |
4920 |                 let mem = buf.unchecked_ref::<WebAssembly::Memory>();
     |                                                            ^^^^^^ this struct is private
     |
note: the struct `Memory` is defined here
    --> /home/naveen/.cargo/registry/src/github.com-1ecc6299db9ec823/js-sys-0.3.36/src/lib.rs:3554:5
```

A clear and concise summary of your question.

### Additional Details

Provide any additional details here.

I am using below version:

wasm-pack 0.8.1
cargo 1.43.0-nightly (e57bd0299 2020-02-21)
rustc 1.43.0-nightly (436494b8f 2020-02-22)

a1trl9:
This PR tries to fix #2040.

It could also help to address the following case:

```rust
#[wasm_bindgen(module = "my-great-module")]
extern "C" {

    #[wasm_bindgen]
    pub type MModule;

    #[wasm_bindgen(constructor, js_class = "JKModule")]
    pub fn new() -> MModule;

    #[wasm_bindgen(static_method_of = MModule)]
    pub fn init();
}
```

Currently, for snippet above, `wasm-bindgen` imports both `JKModule` and `MModule`, which I don't think is perfect -- in rust code, both methods actually belong to the same type. Probably just importing `JKModule` would be clearer (or is it by design?).

Now a global name-alias map will be built before determining imports so that for each import we can check if any name-alias exists and if we should import aliased name instead.

Not sure if this modification follows the initial design of `js_class` -- for `globalThis` in `js-sys`, multiple `js_class` for type `global` is actually used.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -139,6 +139,7 @@ pub struct ImportFunction {
 pub enum ImportFunctionKind {
     Method {
         class: String,
+        rust_class_str: String,
         ty: syn::Type,
         kind: MethodKind,
     },
@@ -185,6 +186,7 @@ pub struct ImportStatic {
 pub struct ImportType {
     pub vis: syn::Visibility,
     pub rust_name: Ident,
+    pub rust_name_str: String,
     pub js_name: String,
     pub attrs: Vec<syn::Attribute>,
     pub typescript_type: Option<String>,
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -140,6 +140,7 @@ pub enum ImportFunctionKind {
     Method {
         class: String,
         rust_class_str: String,
+        aliased_by_js_class: bool,
         ty: syn::Type,
         kind: MethodKind,
     },
@@ -187,6 +188,7 @@ pub struct ImportType {
     pub vis: syn::Visibility,
     pub rust_name: Ident,
     pub rust_name_str: String,
+    pub aliased_by_js_name: bool,
     pub js_name: String,
     pub attrs: Vec<syn::Attribute>,
     pub typescript_type: Option<String>,
--- a/crates/backend/src/encode.rs
+++ b/crates/backend/src/encode.rs
@@ -263,9 +263,18 @@ fn shared_import_function<'a>(
     intern: &'a Interner,
 ) -> Result<ImportFunction<'a>, Diagnostic> {
     let method = match &i.kind {
-        ast::ImportFunctionKind::Method { class, kind, .. } => {
+        ast::ImportFunctionKind::Method {
+            class,
+            rust_class_str,
+            kind,
+            ..
+        } => {
             let kind = from_ast_method_kind(&i.function, intern, kind)?;
-            Some(MethodData { class, kind })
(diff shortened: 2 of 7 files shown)


a1trl9:
Hmm seems it breaks both web-sys and js-sys and path segments are not handled well during parsing. Would re-open it if it could be solved.

alexcrichton:
Sorry for the delay in getting to this, but I'm personally sort of hesitant to merge this. This feels like it's a bit of a complicated interpretation of the attributes and requiring a preprocessing pass doesn't really feel great to me. Does this work if you put `js_class` on all the attributes? Or are there perhaps other alternatives to getting this working?

a1trl9:
@alexcrichton Thank you very much for your comment. 

Well yeah, it does rely too much on current attributes' behaviour and probably not scalable if we wanna add some other features one day. Also, it isn't very efficient to do a heavy preprocessing just for this one purpose. My bad not thinking it very thoroughly and after a few weeks calming down, I also don't feel it is good enough to be merged. 

It feels like in `wasm-bindgen` some tasks (e.g.encoding, code generation) are done one-way which means we don't always know global information of the whole program. For #2040 it is because we can't tell how `js_class` was (or will be) handled for `type` when processing `methods`.  Issue #1853 may be another case. So probably adding more passes to collect global information in `backend` would be an acceptable long-term solution?

alexcrichton:
I'm fine with adding more analysis as necessary in the CLI, the speed isn't really what I'm worried about. I'm mostly worried about the maintenance burden about how there's all these weird interactions between all these attributes. No one ever really designed the whole system here, it was just added piecemeal over time and we have what we have. I'm hesitant to continuously do that because it ends up being an unmaintainable mess.

PR #2140, opened 2020-05-16
web-sys: Add Blob.stream() method

str4d:
This enables a ReadableStream to be obtained for a file, which can then
be handled with e.g. https://github.com/MattiasBuelens/wasm-streams

```rust
let file: web_sys::File = ...;
let stream = wasm_streams::readable::ReadableStream::from(
    file.stream()
        .unchecked_into::<wasm_streams::readable::sys::ReadableStream>(),
);
```

WebIDL source: https://w3c.github.io/FileAPI/#blob-section
Docs: https://developer.mozilla.org/en-US/docs/Web/API/Blob/stream

### DIFF
--- a/crates/web-sys/src/features/gen_Blob.rs
+++ b/crates/web-sys/src/features/gen_Blob.rs
@@ -212,6 +212,14 @@ extern "C" {
         end: f64,
         content_type: &str,
     ) -> Result<Blob, JsValue>;
+    #[cfg(feature = "ReadableStream")]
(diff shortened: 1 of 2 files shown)


Pauan:
The build failures don't seem related to this PR, it seems to be related to the [recent change to rename `anyref` to `externref`](https://github.com/WebAssembly/reference-types/pull/87).

alexcrichton:
Thanks!

PR #2144, opened 2020-05-18
Explictly drop instead of relying on RAII.

TannerRogalsky:
This is functionally identical to the previous implementation but it subverts a linter warning on structs annotated with `#[must_use]`. 

An alternative solution might be to add an annotation to the generated function to `#[allow(unused)]` but this seems more focused and explicit.


alexcrichton:
Thanks!

Although on second though now a few moments after merging, mind adding a test for this?

TannerRogalsky:
Yep! I will open a separate PR for it.

PR #2145, opened 2020-05-18
Add a test that fails to compile if generated code triggers unused lint warning.

TannerRogalsky:
This test validates #2144.


alexcrichton:
Thanks!

PR #2157, opened 2020-05-26
Add another example to js_namespace

mibac138:
I tried to use `js_namespace` on a type instead of a function which confused me. This PR adds another example which demonstrates how to properly use it with a type.

### DIFF
--- a/guide/src/reference/attributes/on-js-imports/js_namespace.md
+++ b/guide/src/reference/attributes/on-js-imports/js_namespace.md
@@ -11,11 +11,16 @@ name (like a class or function name) it'll be accessed through this namespace.
 extern "C" {
     #[wasm_bindgen(js_namespace = console)]
     fn log(s: &str);
+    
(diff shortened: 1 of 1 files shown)


guide/src/reference/attributes/on-js-imports/js_namespace.md
```diff
@@ -11,11 +11,16 @@ name (like a class or function name) it'll be accessed through this namespace.
 extern "C" {
     #[wasm_bindgen(js_namespace = console)]
     fn log(s: &str);
+    
+    type Foo;
+    #[wasm_bindgen(constructor, js_namespace = Bar]
```
alexcrichton, line 16: I think there's a missing `)` here?

mibac138: Oops, my bad

alexcrichton:
👍

PR #2159, opened 2020-05-26
Handle the possibility that the class name is in its own Group

### LINKED ISSUE #2158: `js-sys` fails to build with nightlies on and after 2020-05-25
### Describe the Bug

`js-sys` fails to build with `rustc` nightlies on and after `nightly-2020-05-25`.

```bash
error: return value of constructor must be a bare path
    --> <snipped>/.cargo/registry/src/github.com-1ecc6299db9ec823/js-sys-0.3.39/src/lib.rs:4795:13
     |
4795 | /             /// The
4796 | |             #[doc = $ctor]
4797 | |             /// constructor creates an array of unsigned 8-bit integers.
4798 | |             ///
...    |
4802 | |             #[wasm_bindgen(constructor)]
4803 | |             pub fn new(constructor_arg: &JsValue) -> $name;
     | |___________________________________________________________^
...
4999 | / arrays! {
5000 | |     /// `Int8Array()`
5001 | |     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Int8Array
5002 | |     Int8Array: i8,
...    |
5034 | |     Float64Array: f64,
5035 | | }
     | |_- in this macro invocation
     |
     = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: return value of constructor must be a bare path
    --> <snipped>/.cargo/registry/src/github.com-1ecc6299db9ec823/js-sys-0.3.39/src/lib.rs:4805:13
     |
4805 | /             /// An
4806 | |             #[doc = $ctor]
4807 | |             /// which creates an array with an internal buffer large
4808 | |             /// enough for `length` elements.
...    |
4813 | |             #[wasm_bindgen(constructor)]
4814 | |             pub fn new_with_length(length: u32) -> $name;
     | |_________________________________________________________^

```

### Steps to Reproduce

1. Build `js-sys` with a toolchain newer than `nightly-2020-05-24`:
  - `rustup install nightly-2020-05-25`
  - `cargo +nightly-2020-05-25 build -p js-sys`

### Expected Behavior

The build succeeds!

### Actual Behavior

The build does not succeed.

### Additional Context

Running [cargo-bisect-rustc](https://github.com/rust-lang/cargo-bisect-rustc) pins it down to [this rollup](https://github.com/rust-lang/rust/commit/7726070fa755f660b5da3f82f46e07d9c6866f69) and from there it's fairly straightforward to pick out rust-lang/rust#72388.

Here's the output from `cargo-bisect-rustc` if it helps:

<details>
<summary>bisected with <a href='https://github.com/rust-lang/cargo-bisect-rustc'>cargo-bisect-rustc</a> v0.5.1</summary>

searched nightlies: from nightly-2020-05-24 to nightly-2020-05-26
regressed nightly: nightly-2020-05-25
searched commits: from https://github.com/rust-lang/rust/commit/8970e8bcf6153d1ead2283f1a0ed7b192230eca6 to https://github.com/rust-lang/rust/commit/46e85b4328fe18492894093c1092dfe509df4370
regressed commit: https://github.com/rust-lang/rust/commit/7726070fa755f660b5da3f82f46e07d9c6866f69

Host triple: x86_64-unknown-linux-gnu
Reproduce with:
```bash
cargo bisect-rustc -- check --target x86_64-unknown-linux-gnu 
```
</details>
 
The cause of this error, as best as I can tell, is that `$name` in [this macro](https://github.com/rustwasm/wasm-bindgen/blob/1e1cab6202519dec2c30314e26aa3c8f601f3506/crates/js-sys/src/lib.rs#L4787-L4997) goes from being passed to `#[wasm_bindgen]` as a [`syn::Type::Path`](https://docs.rs/syn/1.0.25/syn/enum.Type.html#variant.Path) to being passed in as a [`syn::Group`](https://docs.rs/syn/1.0.25/syn/enum.Type.html#variant.Path) containing the `syn::Type::Path` that the macro wants (as in the span that `$name` is in isn't discarded). `wasm_bindgen`'s parser doesn't know to handle this ([here](https://github.com/rustwasm/wasm-bindgen/blob/1e1cab6202519dec2c30314e26aa3c8f601f3506/crates/macro-support/src/parser.rs#L469-L474) and [here](https://github.com/rustwasm/wasm-bindgen/blob/1e1cab6202519dec2c30314e26aa3c8f601f3506/crates/macro-support/src/parser.rs#L428-L434)) and so we get the errors above.

There's already an issue open in rust-lang/rust for this: rust-lang/rust#72545. For that specific issue the crate in question was using `proc-macro-hack` but as [shown](https://github.com/rust-lang/rust/issues/72545#issuecomment-633293353) the same change can also break proc macros that don't use it.

rrbutani:
Closes #2158.

I'm not sure if this is the best way to go about fixing this or if there are other places where this kind of thing should be added.

If it's helpful, @dtolnay's fix for `proc-macro-hack` is [here](https://github.com/dtolnay/proc-macro-hack/pull/57/files).

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -875,7 +875,11 @@ impl<'a> MacroParse<BindgenAttrs> for &'a mut syn::ItemImpl {
                 "#[wasm_bindgen] generic impls aren't supported"
             );
         }
-        let name = match *self.self_ty {
+        let self_ty = match &*self.self_ty {
+            syn::Type::Group(syn::TypeGroup { elem, .. }) => &**elem,
+            otherwise => &otherwise,
+        };
+        let name = match self_ty {
             syn::Type::Path(syn::TypePath {
                 qself: None,
                 ref path,
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -376,6 +376,14 @@ impl<'a> ConvertToAst<BindgenAttrs> for &'a mut syn::ItemStruct {
     }
 }
 
+fn get_ty(mut ty: &syn::Type) -> &syn::Type {
+    while let syn::Type::Group(g) = ty {
+        ty = &g.elem;
+    }
+
+    ty
+}
+
 impl<'a> ConvertToAst<(BindgenAttrs, &'a ast::ImportModule)> for syn::ForeignItemFn {
     type Target = ast::ImportKind;
 
@@ -414,7 +422,7 @@ impl<'a> ConvertToAst<(BindgenAttrs, &'a ast::ImportModule)> for syn::ForeignIte
             let class = wasm.arguments.get(0).ok_or_else(|| {
                 err_span!(self, "imported methods must have at least one argument")
             })?;
-            let mut class = match &*class.ty {
+            let class = match get_ty(&class.ty) {
                 syn::Type::Reference(syn::TypeReference {
                     mutability: None,
                     elem,
@@ -425,10 +433,7 @@ impl<'a> ConvertToAst<(BindgenAttrs, &'a ast::ImportModule)> for syn::ForeignIte
                     "first argument of method must be a shared reference"
                 ),
             };
-            if let syn::Type::Group(syn::TypeGroup { elem, .. }) = class {
-                class = elem;
-            }
-            let class_name = match *class {
+            let class_name = match get_ty(class) {
                 syn::Type::Path(syn::TypePath {
                     qself: None,
                     ref path,
@@ -465,14 +470,11 @@ impl<'a> ConvertToAst<(BindgenAttrs, &'a ast::ImportModule)> for syn::ForeignIte
 
             ast::ImportFunctionKind::Method { class, ty, kind }
         } else if opts.constructor().is_some() {
-            let mut class = match js_ret {
+            let class = match js_ret {
                 Some(ref ty) => ty,
                 _ => bail_span!(self, "constructor returns must be bare types"),
             };
-            if let syn::Type::Group(syn::TypeGroup { elem, .. }) = class {
-                class = elem;
-            }
-            let class_name = match *class {
+            let class_name = match get_ty(class) {
                 syn::Type::Path(syn::TypePath {
                     qself: None,
                     ref path,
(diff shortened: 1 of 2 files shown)


rrbutani:
I think the remaining `#[wasm_bindgen]` invocation that fails under the latest nightly is breaking because of `syn`.

`syn` seems to assume that entire expressions will be a part of a single `Group` (`trailer_expr` in `expr.rs`) which leads to the bit that parses the body of functions (`Block::parse_within` in `stmt.rs`) checking if the fragment of the expression that was returned from `parse_stmt` (this calls `stmt_expr` which calls `expr_early` → `unary_expr` → `trailer_expr`) needs a semicolon (using `requires_terminator` in `expr.rs`). `requires_terminator` isn't `Group` aware but it doesn't matter; either way the function would correctly report that expressions _do_ need to be followed by a semicolon when they're the last thing in a block. And so, we get an error.

I think the right thing to do here is to update the expr parse machinery to not stop on Groups (without just throwing away the Groups). Just ditching the check for Groups in `trailer_expr` seems to do the trick for me, but I'm not sure if this is the best way to do things/actually fixes the issue properly (I'm not at all familiar with `syn`'s internals).

alexcrichton:
Thanks for this! Poking around a bit I think it might be best to be [a bit more general with these fixes](https://gist.github.com/alexcrichton/3c93ab2547d45d9caa3b72309cd4262b) rather than being so case-specific in these few locations. I'm a bit lost though in the final error and how we could work around and/or fix it...

alexcrichton:
If this parse error seems to originate from `syn` though then it seems reasonable to file an issue upstream about that.

rrbutani:
@alexcrichton I made the suggested changes.

That last error doesn't seem to happen anymore in CI, but I have no idea why. The [fix that was pushed to syn](https://github.com/dtolnay/syn/pull/832/files#diff-180bb3ad456cc10517d878fb7795341cL1548-L1549) doesn't seem like it fixes our case (we don't have a `::` in the function name).

It actually also still fails with the same error for me locally.

rrbutani:
Oh whoops; just noticed you pinned the nightly version.

I'll open an issue on `syn`.

alexcrichton:
In any case this seems like a good change to land regardless, so I'm going to merge this.

PR #2176, opened 2020-06-01
add deno target

jakobhellermann:
adds a new `OutputMode` for [Deno](https://deno.land/).

Exports and js-imports are generated like `OutputMode::Web` or `OutputMode::Node { experimental_modules: true }`.

`wasm`-files cannot be imported directly, so the file is read at runtime. If embedding the file as a base64 string or similar is preferable, I can try to do that.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -143,7 +143,8 @@ impl<'a> Context<'a> {
             | OutputMode::Node {
                 experimental_modules: true,
             }
-            | OutputMode::Web => {
+            | OutputMode::Web
+            | OutputMode::Deno => {
                 if contents.starts_with("function") {
                     let body = &contents[8..];
                     if export_name == definition_name {
@@ -269,6 +270,40 @@ impl<'a> Context<'a> {
         reset_indentation(&shim)
     }
 
+    // generates somthing like
+    // ```js
+    // const imports = {
+    //   __wbindgen_placeholder__: {
+    //     __wbindgen_throw: function(..) { .. },
+    //     ..
+    //   }
+    // }
+    // ```
+    fn generate_deno_imports(&self) -> String {
+        let mut imports = "const imports = {\n".to_string();
+        imports.push_str(&format!("  {}: {{\n", crate::PLACEHOLDER_MODULE));
+
+        for (id, js) in crate::sorted_iter(&self.wasm_import_definitions) {
+            let import = self.module.imports.get(*id);
+            imports.push_str(&format!("{}: {},\n", &import.name, js.trim()));
+        }
+
+        imports.push_str("\t}\n};\n\n");
+
+        imports
+    }
+
+    fn generate_deno_wasm_loading(&self, module_name: &str) -> String {
+        format!(
+            "const file = new URL(import.meta.url).pathname;
+            const wasmFile = file.substring(0, file.lastIndexOf(Deno.build.os === 'windows' ? '\\\\' : '/') + 1) + '{}_bg.wasm';
+            const wasmModule = new WebAssembly.Module(Deno.readFileSync(wasmFile));
+            const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
+            const wasm = wasmInstance.exports;",
+            module_name
+        )
+    }
+
     /// Performs the task of actually generating the final JS module, be it
     /// `--target no-modules`, `--target web`, or for bundlers. This is the very
     /// last step performed in `finalize`.
@@ -325,6 +360,15 @@ impl<'a> Context<'a> {
                         module_name
                     ))),
                 );
+
+                if needs_manual_start {
+                    footer.push_str("wasm.__wbindgen_start();\n");
+                }
+            }
+
+            OutputMode::Deno => {
+                footer.push_str(&self.generate_deno_imports());
+                footer.push_str(&self.generate_deno_wasm_loading(module_name));
 
                 if needs_manual_start {
                     footer.push_str("wasm.__wbindgen_start();\n");
@@ -443,7 +487,8 @@ impl<'a> Context<'a> {
             | OutputMode::Node {
                 experimental_modules: true,
             }
-            | OutputMode::Web => {
+            | OutputMode::Web
+            | OutputMode::Deno => {
                 for (module, items) in crate::sorted_iter(&self.js_imports) {
                     imports.push_str("import { ");
                     for (i, (item, rename)) in items.iter().enumerate() {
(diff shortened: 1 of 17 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -1247,7 +1292,7 @@ impl<'a> Context<'a> {
                 fields: Vec::new(),
             })?;
             self.global(&format!("let cached{} = new {}{};", s, name, args));
-        } else if !self.config.mode.always_run_in_browser() {
+        } else if !self.config.mode.always_run_in_browser() && !self.config.mode.deno() {
```
jakobhellermann, line 1295: the function `always_run_in_browser` seems to only be used to determine whether `const lTextEncoder = typeof TextEncoder === 'undefined' ? (0, module.require)('util').TextEncoder : TextEncoder;` will be used instead of the text processor directly, so maybe renaming the method to `text_processor_available` and including `OutputMode::Deno` may be cleaner.

alexcrichton: This function is sort of turning into spaghettit with its conditionals, mind just throwing in an explicit `match` over the output mode for this?

crates/cli-support/src/js/mod.rs
```diff
@@ -269,6 +270,40 @@ impl<'a> Context<'a> {
...
+        imports.push_str("\t}\n};\n\n");
+
+        imports
+    }
+
+    fn generate_deno_wasm_loading(&self, module_name: &str) -> String {
+        format!(
+            "const file = new URL(import.meta.url).pathname;
+            const wasmFile = file.substring(0, file.lastIndexOf(Deno.build.os === 'windows' ? '\\\\' : '/') + 1) + '{}_bg.wasm';
+            const wasmModule = new WebAssembly.Module(Deno.readFileSync(wasmFile));
```
alexcrichton, line 300: It looks like deno support for native imports was removed in https://github.com/denoland/deno/pull/5135, and perhaps https://github.com/denoland/deno/issues/5609 is a tracking issue for that returning? Mind leaving a comment referring to that issue here in case we need to check up on it in the future?

jakobhellermann:
looks like the test finally passes.
I copied the `import_js` example, because that way imports and exports are being tested, and modified it a little bit to include passing a string to a function.
If anything else should be tested, I can add that.

crates/cli-support/src/js/mod.rs
```diff
@@ -1238,27 +1285,31 @@ impl<'a> Context<'a> {
...
-            ));
-            self.global(&format!("let cached{0} = new l{0}{1};", s, args));
-        } else {
-            self.global(&format!("let cached{0} = new {0}{1};", s, args));
-        }
+                    s
+                ));
+                self.global(&format!("let cached{0} = new l{0}{1};", s, args));
+            }
+            _ => self.global(&format!("let cached{0} = new {0}{1};", s, args)),
```
alexcrichton, line 1311: Could the other variants be explicitly mentioned here to avoid the catch-all? That way if new variants are added they can be considered here as well

jakobhellermann: done

jakobhellermann:
I'm not sure what would need to be done to run the test suit in deno.
Add a new file similar to `crates/cli/src/bin/wasm-bindgen-test-runner/node.rs` for deno and add `wasm_bindgen_test_configure!(deno)`?

alexcrichton:
Something like that would work, yeah, although since it's so similar to node it might be best to control with an env var rather than in the code itself. Either way though it's not the most urgent thing in the world. Running the full test suite would provide a lot of confidence that the wrapper generation here is working for all use cases, but since this is a new feature it doesn't have to do everything right out of the gate.

jakobhellermann:
also all the tests use `exports.foo = ...` and `require`.

examples/deno/crate/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
alexcrichton, line 1: The other examples all have `examples/foo/Cargo.toml`, so could this follow the same pattern?

guide/src/reference/deployment.md
```diff
@@ -14,6 +14,7 @@ should behave the same way in this respect. The values possible here are:
 | [`bundler`]     | Suitable for loading in bundlers like Webpack              |
 | [`web`]         | Directly loadable in a web browser                         |
 | [`nodejs`]      | Loadable via `require` as a Node.js module                 |
+| [`Deno`]        | Loadable using imports from Deno modules                   |
```
alexcrichton, line 17: I think this wants to be a lower-case `d`, and it looks like the destination of this link needs to be filled in?

jakobhellermann: it looks like the branding is an uppercase D: https://deno.land/
I will fix the link.

jakobhellermann: though this is not the branding but the flag, so nevermind

alexcrichton:
All looks great to me, thanks!

Pauan:
I know this has already been merged, but why does `--target web` not work?

Why can't you just do `await init(Deno.readFileSync("path/to/foo_bg.wasm"));`?

jakobhellermann:
To be honest I didn't quite understand why the init function was needed and thought there were some changes to be done anyway because for Deno the file wouldn't be `fetch`ed but `Deno.readFile`ed instead. But I just tried `await init(Deno.readFileSync("path/to/foo_bg.wasm"));` and it works perfectly...

jakobhellermann:
maybe wasm-pack can instead have a deno target using `--target web` which generates a `mod.ts` with 
```ts
// @deno-types=foo.d.ts
export * from "./foo.js";
import init from "./foo.js";
init(Deno.readFileSync(import.meta.url logic));
```

Pauan:
@jakobhellermann The `init` function accepts string, `Request`, `Url`, `ArrayBuffer`, `Uint8Array`, or `WebAssembly.Module`.

If you pass in an `ArrayBuffer`, `Uint8Array` or `WebAssembly.Module` then it will not fetch. This allows you to handle all of the loading yourself (such as with Deno).

> maybe wasm-pack can instead have a deno target using --target web which generates a mod.ts 

That won't work, because then you cannot access the exports of the Rust code:

```js
import init, * as exports from "./foo.js";

init(Deno.readFileSync("./foo_bg.wasm")).then(() => {
    // Use exports...
});
```

Instead, I recommend using [our Rollup plugin](https://github.com/wasm-tool/rollup-plugin-rust#readme), which will automatically call `init` for you. You can customize the loading behavior like this:

```js
rust({
    importHook: function (path) {
        return "Deno.readFileSync(" + JSON.stringify(path) + ")";
    },
})
```

alexcrichton:
I think it might still makes sense to have a deno target for "it works similarly to node" where it auto-inits, but perhaps the generated could could be refactored to use top-level `await` like @Pauan mentions so it's basically `--target web` but with one extra line?

Pauan:
@alexcrichton I'm okay with that, though I'm a bit concerned that it might create a precedent for other targets.

alexcrichton:
It's true yeah I don't want to create a proliferation of too many targets, and @Pauan if you'd like I think it's fine if we revert this and continue discussion on an issue (sorry should have consulted first!)

Issue #2206, opened 2020-06-16
tests fail if start attribute is present


Urhengulas:
### Describe the Bug

Testing of wasm-bindgen lib fails if `#[wasm_bindgen(start)]` is present (for both browser and node).

<details>
<summary>Error Message</summary>

```bash
$ wasm-pack test
[INFO]: Checking for the Wasm target...
   Compiling rav1e_js v0.1.0 (/home/urhengulas/Documents/github.com/xiph/rav1e/rav1e_js)
error: linking with `rust-lld` failed: exit code: 1

### ----- stripped ----- ###

  = note: rust-lld: error: function signature mismatch: main
          >>> defined as () -> void in /path/to/rav1e/target/wasm32-unknown-unknown/debug/deps/librav1e_js.rlib(rav1e_js.e16cnisiuzwaayg.rcgu.o)
          >>> defined as (i32, i32) -> i32 in /path/to//rav1e/target/wasm32-unknown-unknown/debug/deps/web-17ed0dcdfd20c376.2kwhsv3emygdoj04.rcgu.o

### ----- stripped ---- ###

error: entry symbol `main` declared multiple times
  |
  = help: did you use `#[no_mangle]` on `fn main`? Use `#[start]` instead

error: aborting due to previous error

error: could not compile `rav1e_js`.

### ----- stripped ----- ###

Caused by: failed to execute `cargo build`: exited with exit code: 101
  full command: "cargo" "build" "--tests" "--target" "wasm32-unknown-unknown"
```
</details>

### Steps to Reproduce

```bash
# 1. clone repo
git clone https://github.com/Urhengulas/rav1e.git
# 2. checkout commit
git checkout 46283554cfc093fb6721557068b3242794544226
# 3. go into subfolder
cd rav1e_js/
# 4. run tests
wasm-pack test --node
```

### Expected Behavior
I expect `wasm-pack test --node` to run all wasm-bindgen tests for node.

And since the wasm-bindgen books says [here](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/start.html):
> The start function will not be executed when testing.

I also expect it to not be executed. (Maybe not even part of the build).

### Actual Behavior

The command fails and doesn't run any tests.

### Additional Context

`wasm-pack build` runs without any problems and also the `pkg/` works.

alexcrichton:
[...]

PR #2207, opened 2020-06-16
Rename fn to main_js in guide/src/reference/attributes/on-rust-exports/start.md

Urhengulas:
Because the tests fail, when a function annotated with `#[wasm_bindgen(start)]` is named `main`, this PR changes the name in the example in the wasm_bindgen book to `main_js` and says, that it can't be named `main`.

The information that it can't be named `main` is from https://github.com/rust-gamedev/wg/issues/51, and solved #2206.

## Changes
* rename function from `main` to `main_js`
* add hint that function can't be named `main`

### DIFF
--- a/guide/src/reference/attributes/on-rust-exports/start.md
+++ b/guide/src/reference/attributes/on-rust-exports/start.md
@@ -6,7 +6,7 @@ soon as the wasm module is instantiated.
 
 ```rust
 #[wasm_bindgen(start)]
-pub fn main() {
+pub fn main_js() {
     // executed automatically ...
 }
 ```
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I don't think this is quite right though. You can indeed have a function called `main`. Is it perhaps specific to the context of the gamedev area that they've already got a symbol called `main` somewhere else? It may be the case that a different crate you're using already defines `main`, which means you can't define `main` yourself when using that crate.

Urhengulas:
If you say so that is probably true and in my experience it is also for the `wasm-pack build`.

But it doesn‘t work for `wasm-pack test —node` (or browser) and the only dev-dependency I am having is `wasm-bindgen-test = "0.3.13"`.

17cupsofcoffee:
> The information that it can't be named main is from rust-gamedev/wg#51

To add some context: I raised that issue before https://github.com/rustwasm/wasm-bindgen/issues/1630 was created/merged, so the info in the first post of rust-gamedev/wg#51 is now outdated. I was referring to the fact that you previously had to wrap your existing `main` in a `#[wasm_bindgen(start)]` annotated function in order to get a binary crate running on WASM, but that's no longer the case.

Urhengulas:
Thank you @17cupsofcoffee for clarifying.

So it seems I had some weird edge-case problem, which accidentally could be resolved with this outdated information. Nice 😎

I am a bit worried other people might also encounter this, but I am going to close the issue.

Thanks everyone :)

PR #2210, opened 2020-06-18
Trims trailing whitespaces in doc comments in generated d.ts

sl4m:
The doc comments in the generated d.ts file contains trailing whitespaces. This simply removes the whitespaces.


alexcrichton:
Thanks!

PR #2191, opened 2020-06-06
Implement `IntoWasmAbi` for `Option<&Closure>.

teymour-aldridge:
I think this is the correct way to do it. I'm not sure about that though (my knowledge about WebAssembly is very limited).
This should resolve #2173.

Issue #2173, opened 2020-05-31
implement OptionIntoWasmAbi for Option<&Closure>


ctaggart:
### Motivation
In Monico Editor, you configure the environment with callbacks. This is the TypeScript:
``` ts
    export interface Environment {
        getWorkerUrl?(workerId: string, label: string): string;
    }
```
It is not `readonly`, so it is settable. It has a `?`, so it may be set with [undefined](https://basarat.gitbook.io/typescript/recap/null-undefined).

This works if i map it using an optional js_sys::Function:
``` rust
#[wasm_bindgen()]
extern "C" {
    pub type Environment;
    # [ wasm_bindgen ( method , setter , js_name = getWorkerUrl ) ]
    pub fn set_get_worker_url(this: &Environment, value: Option<&Function>>);
}
```

In order to have it be more type safe, I want to use Closure<T>:
``` rust
#[wasm_bindgen()]
extern "C" {
    pub type Environment;
    # [ wasm_bindgen ( method , setter , js_name = getWorkerUrl ) ]
    pub fn set_get_worker_url(this: &Environment, value: Option<&Closure<dyn FnMut(String, String) -> String>>);
}
```

But then I get this error:
``` rust
   Compiling monaco-editor-hello-world v0.1.0 (/Users/cameron/rs/ts2rs/monaco-editor-hello-world)
error[E0277]: the trait bound `&wasm_bindgen::closure::Closure<dyn std::ops::FnMut(std::string::String, std::string::String) -> std::string::String>: wasm_bindgen::convert::traits::OptionIntoWasmAbi` is not satisfied
  --> monaco-editor-hello-world/src/monaco.rs:10:51
   |
10 |     pub fn set_get_worker_url(this: &Environment, value: Option<&Closure<dyn FnMut(String, String) -> String>>);
   |                                                   ^^^^^ the trait `wasm_bindgen::convert::traits::OptionIntoWasmAbi` is not implemented for `&wasm_bindgen::closure::Closure<dyn std::ops::FnMut(std::string::String, std::string::String) -> std::string::String>`
   |
   = note: required because of the requirements on the impl of `wasm_bindgen::convert::traits::IntoWasmAbi` for `std::option::Option<&wasm_bindgen::closure::Closure<dyn std::ops::FnMut(std::string::String, std::string::String) -> std::string::String>>`

[...]

PR #2211, opened 2020-06-20
test for OptionIntoWasmAbi for Option<&Closure>

ctaggart:
I added a test for #2191, but it shows it not working yet. Running:
```
cargo test --target wasm32-unknown-unknown --features nightly --test wasm
```
I get the error from #2173.:
```
error[E0277]: the trait bound `&wasm_bindgen::prelude::Closure<dyn std::ops::FnMut(u32) -> u32>: wasm_bindgen::convert::OptionIntoWasmAbi` is not satisfied
   --> tests/wasm/closures.rs:115:21
    |
115 |     fn option_works(a: Option<&Closure<FnMut(u32) -> u32>>) -> u32;
    |                     ^ the trait `wasm_bindgen::convert::OptionIntoWasmAbi` is not implemented for `&wasm_bindgen::prelude::Closure<dyn std::ops::FnMut(u32) -> u32>`
```

### DIFF
--- a/tests/wasm/closures.js
+++ b/tests/wasm/closures.js
@@ -136,3 +136,11 @@ exports.js_store_forgotten_closure = f => {
 exports.js_call_forgotten_closure = () => {
   FORGOTTEN_CLOSURE();
 };
+
+exports.option_works = a => {
+  if (a) {
+    a(2);
+  } else {
+    1;
+  }
+};
--- a/src/convert/closures.rs
+++ b/src/convert/closures.rs
@@ -136,6 +136,28 @@ where
     }
 }
 
+impl<'a, 'b, A, R> IntoWasmAbi for Option<&'a (dyn Fn(&A) -> R + 'b)>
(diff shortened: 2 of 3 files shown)


tests/wasm/closures.js
```diff
@@ -136,3 +136,11 @@ exports.js_store_forgotten_closure = f => {
...
   FORGOTTEN_CLOSURE();
 };
+
+exports.option_works = a => {
+  if (a) {
+    a(2);
+  } else {
+    1;
+  }
+};
```
teymour-aldridge, line 146: The POSIX specification for files requires that they should have new lines. I think this is breaking the tests.
```suggestion
};

```

alexcrichton:
Ah I think a few more `impl` blocks are needed. It looks like there's one for stack closures directly (`&dyn Fn(...)`) but not for `&Closure<T>`. It'd be good to add tests for all of them!

Issue #2045, opened 2020-03-18
Imperfect enum of TypeScript.


s-shin:
### Describe the Bug

`enum` definition is generated in `d.ts` from `enum` type of rust, but the actual code in JS is not compatible with TS. This can cause unexpected errors for example in [reverse mappings](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings).

### Steps to Reproduce

```rust
#[wasm_bindgen]
pub enum Foo {
    A,
    B,
    C,
}
```

This code will be converted to JS:

```js
export const Foo = Object.freeze({ A:0,B:1,C:2, });
```

and TS (in `d.ts`):

```ts
export enum Foo {
  A,
  B,
  C,
}
```

but these are not completely compatible.

### Expected Behavior

Consistency should be kept.

### Actual Behavior

We can make JS code compatible with TS like this:

```js
export var Foo;
(function (Foo) {
    Foo[Foo["A"] = 0] = "A";
    Foo[Foo["B"] = 1] = "B";
    Foo[Foo["C"] = 2] = "C";
})(Foo || (Foo = {}));
```

or can also do TS definition compatible with JS.

PR #2240, opened 2020-07-15
add reverse mappings from value to name on enums exported from rust/wasm

Treeki:
This PR adds reverse mappings to enums exported from Rust, so that the string corresponding to a particular integer value can be obtained easily. This matches what TypeScript does for enums defined directly by TypeScript code.

I’ve updated the corresponding tests to take this into account.

This goes some way towards resolving #2045, albeit the approach is still a little different - I’ve kept the `Object.freeze` call that wasm-bindgen performs on its enums (whereas TypeScript does not seem to return a frozen object at all). Here’s an example of how they compare:

```javascript
// TypeScript output
var Shape;
(function (Shape) {
    Shape[Shape["Circle"] = 1] = "Circle";
    Shape[Shape["Triangle"] = 3] = "Triangle";
    Shape[Shape["Rectangle"] = 4] = "Rectangle";
})(Shape || (Shape = {}));

// wasm-bindgen output
module.exports.Shape = Object.freeze({ Circle:1,"1":"Circle",Triangle:3,"3":"Triangle",Rectangle:4,"4":"Rectangle", });
```

Nevertheless, having reverse mappings still makes it more compatible with TypeScript code.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -3068,6 +3068,7 @@ impl<'a> Context<'a> {
                 variants.push_str(&variant_docs);
             }
             variants.push_str(&format!("{}:{},", name, value));
+            variants.push_str(&format!("\"{}\":\"{}\",", value, name));
             if enum_.generate_typescript {
                 self.typescript.push_str("\n");
                 if !variant_docs.is_empty() {
--- a/tests/wasm/enums.js
+++ b/tests/wasm/enums.js
@@ -5,7 +5,10 @@ exports.js_c_style_enum = () => {
     assert.strictEqual(wasm.Color.Green, 0);
     assert.strictEqual(wasm.Color.Yellow, 1);
     assert.strictEqual(wasm.Color.Red, 2);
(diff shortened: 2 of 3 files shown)


alexcrichton:
Sounds reasonable to me! CI thinks this may need a rebase though?

Treeki:
I am very confused - it's just one commit based off the current tip of the master branch, GitHub claims it's mergeable, and my other PR didn't have this issue (albeit it did fail on the rustfmt phase, which I've just fixed).

This page claims it can be fixed by closing and reopening the PR, so I suppose I'll try that... https://help.appveyor.com/discussions/problems/4830-stuck-on-appveyor-was-unable-to-build-non-mergeable

Treeki:
This has gotten further, but one of the jobs has failed with a SSL error involving fetching the crates.io index, which feels like it’s unrelated to the PR itself.

I’ve not used AppVeyor before; is there a way to fix this that doesn’t involve pushing an unnecessary new commit or doing another ugly close/re-open cycle?

alexcrichton:
Ah those are just spurious errors, no need to worry about them!

Issue #2154, opened 2020-05-25
Typescript enum definitions are not used in function signatures


Vlad-Shcherbina:
### To reproduce

1. Run `wasm-bindgen --target web` on the wasm generated from the following library:
    ```rust
    use wasm_bindgen::prelude::*;

    #[wasm_bindgen]
    pub enum Hand {
        Left = 0,
        Righ = 1,
    }

    #[wasm_bindgen]
    pub fn f(h: Hand) -> Hand { h }
    ```
2. Inspect generated .d.ts

### Expected result

```typescript
export enum Hand {
  Left,
  Righ,
}

export function f(h: Hand): Hand;
```

### Actual result

```typescript
export enum Hand {
  Left,
  Righ,
}

export function f(h: number): number;
```

### Version info

wasm-bindgen 0.2.62

alexcrichton:
I unfortunately don't have the time to implement a new feature like this, but a PR would be much appreciated!

ChristianIvicevic:
> I unfortunately don't have the time to implement a new feature like this, but a PR would be much appreciated!

The requested feature is really nice to have but the PR opened 3.5 years ago is stale. What are the chances that you pick up the work on this feature? Sadly I can't personally contribute myself.

daxpedda:
I'm not aware of any current maintainer wanting to implement this.
But I'm happy to review a potential PR.

ChristianIvicevic:
@daxpedda I just opened a PR that is ready to be reviewed :)

PR #2242, opened 2020-07-16
use enums in function signatures within typescript definitions

Treeki:
This is an attempt at fixing #2154 by generating TypeScript definitions that use the correct type for enums within function arguments and return types rather than just `number`.

I’m quite unfamiliar with this codebase so it’s very possible that my approach will require further tweaks, but I hope it’s at least a good start.

I’ve adapted the backend crate’s `codegen` module to output the `js_name` for Enums within the descriptor and likewise adapted the cli-support crate to load that information. To have it appear in the TS definitions, I’ve added `AdapterType::Enum(String)` which compiles to the same adapter code but carries the name alongside it.

I’ve also added a new pair of files to the typescript_tests crate to handle these cases.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -1156,6 +1156,9 @@ impl<'a> ToTokens for DescribeImport<'a> {
 impl ToTokens for ast::Enum {
     fn to_tokens(&self, into: &mut TokenStream) {
         let enum_name = &self.rust_name;
+        let name_str = self.js_name.to_string();
+        let name_len = name_str.len() as u32;
+        let name_chars = name_str.chars().map(|c| c as u32);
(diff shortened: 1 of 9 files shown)


SebastienGllmt:
Heads up this would conflict with https://github.com/rustwasm/wasm-bindgen/pull/2631/files#diff-c631726c09f890be2c14b2da764ef112342a81effe4cfbf15d415497a7fc0db7L138

daxpedda:
I'm happy to review this after a rebase.

thomasetter:
The underlying issue #2154  seems to have been fixed in #3647.

This PR should therefore probably be closed as it is obsolete.

daxpedda:
Indeed, thanks!

PR #2248, opened 2020-07-22
Update support for weak references

alexcrichton:
This commit updates the `WASM_BINDGEN_WEAKREF` feature support to the
latest version of the weak references proposal in JS. This also updates
the `Closure` type to use finalization registries to ensure closures are
deallocated with the JS gc. This means that `Closure::forget` will not
actually leak memory in a weakref-using application.

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -356,22 +356,28 @@ where
         }
     }
 
-    /// Leaks this `Closure` to ensure it remains valid for the duration of the
-    /// entire program.
+    /// Release memory management of this closure from Rust to the JS GC.
     ///
-    /// > **Note**: this function will leak memory. It should be used sparingly
-    /// > to ensure the memory leak doesn't affect the program too much.
+    /// When a `Closure` is dropped it will release the Rust memory and
+    /// invalidate the associated JS closure, but this isn't always desired.
+    /// Some callbacks are alive for the entire duration of the program or for a
+    /// lifetime dynamically managed by the JS GC. This function can be used
+    /// to drop this `Closure` while keeping the associated JS function still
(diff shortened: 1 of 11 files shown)


src/closure.rs
```diff
@@ -356,22 +356,28 @@ where
...
-        }
+    /// By default this function will leak memory. This can be dangerous if this
+    /// function is called many times in an application because the memory leak
+    /// will overwhelm the page quickly and crash the wasm.
+    ///
+    /// If the browser, however, supports weak references, then this function
+    /// will not leak memory. Instead the Rust memory will be reclaimed when the
+    /// JS closure is GC'd. Weak references are not enabled by default since
+    /// they're still a proposal for the JS standard. They can be enabled with
+    /// `WASM_BINDGEN_WEAKREF=1` when running `wasm-bindgen`, however.
```
fitzgen, line 376: Maybe time to make an actual CLI flag, now that these are pretty far down the standards pipeline.

alexcrichton: Do you think it's worth turning this on by default and requiring a flag to turn it off? I'm not actually sure at what point we should do that...

I agree though that at least this should be a CLI flag now.

fitzgen: I don't think we are at default-on yet. Probably not for a year or so, I'd guess, if ever :-/

kellpossible:
Just a heads up, this change broke the gloo-events crate: rustwasm/gloo#120

alexcrichton:
Thanks for the report! I'll publish a new release with https://github.com/rustwasm/wasm-bindgen/pull/2258 to fix this.

PR #2249, opened 2020-07-22
Pass actual stack pointers around instead of address 8

### LINKED ISSUE #2218: Generated JS binding is not thread-safe
### Describe the Bug

I am working on a Rust project that we compile to wasm32 which uses atomics. The compiled wasm is then used in a browser by multiple web workers. I was seeing strange issues with return values from different workers where the return values got mixed up, e.g. worker 1 received a return value from another worker. After drilling into this I found that the JS binding that is generated by wasm-bindgen is not thread-safe. In particular, I ran into a lot of issues when returning Option's to JS. The generated code for this is the following:

```js
var r0 = getInt32Memory0()[8 / 4 + 0];
var r1 = getInt32Memory0()[8 / 4 + 1];
```

When using synchronization primitives the order in which threads / workers are unlocked is not always guaranteed, and that's why this can easily mix up return values because all threads use the same stack space for return values.

### Possible Solution

After talking to @alexcrichton, one possible solution could be to increment a stack pointer per thread and have each thread allocate a unique stack space (a thread-local area).

alexcrichton:
This commit updates the implementation of passing return pointers from
JS to wasm to actually modify the wasm's shadow stack pointer instead of
manufacturing the arbitrary address of 8. The purpose of this is to
correctly handle threaded scenarios where each thread will write to its
own area instead of everyone trying to compete at address 8.

The implementation here will lazily, if necessary, export the stack
pointer we find the module and modify it as necessary.

Closes #2218

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -211,11 +211,12 @@ jobs:
     displayName: "Build benchmarks"
     steps:
       - template: ci/azure-install-rust.yml
-      - template: ci/azure-install-wasm-pack.yml
-      - script: wasm-pack build --target web benchmarks
+      - script: rustup target add wasm32-unknown-unknown
+        displayName: "add target"
+      - script: cargo build --manifest-path benchmarks/Cargo.toml --release --target wasm32-unknown-unknown
         displayName: "build benchmarks"
-      - script: rm -f benchmarks/pkg/.gitignore
-        displayName: "remove stray gitignore"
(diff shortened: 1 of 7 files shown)


d3lm:
Ah, this is awesome! Thanks @alexcrichton for implementing this. I totally did not get around it and had too much on my plate. Sorry I couldn't get involved earlier and work on a PR.

alexcrichton:
No worries!

I basically do releases when requested, no schedule currently. Would you like a release to be made?

d3lm:
Yep, that would be amazing, so I could bumped the version in my project and make sure the js binding is thread safe.

alexcrichton:
Ok, I'm doing the bump at https://github.com/rustwasm/wasm-bindgen/pull/2256

PR #2257, opened 2020-07-28
Add a `--reference-types` CLI flag

alexcrichton:
This proposal is now shipping in browsers!


guide/src/reference/reference-types.md
```diff
@@ -0,0 +1,59 @@
...
+
+```js
+const heap = new Array(32).fill(undefined);
+
+heap.push(undefined, null, true, false);
+
+let stack_pointer = 32;
+
+function addBorrowedObject(obj) {
+    if (stack_pointer == 1) throw new Error('out of js stack');
```
chinedufn, line 29: Hey quick question.

How come this isn't `if (stack_pointer == 0)`?

What's the issue with assigning an object to `heap[0]`.

Just for my own understanding.

Thanks!

alexcrichton: I think at some point we were reserving the first slot (index 0) as a sort of null pointer or something like that. I don't believe those plans ever panned out though so this is likely a bug that we can fix nowadays.

chinedufn: Sweet thanks for explaining!

Urhengulas:
Hello @alexcrichton :wave: 

Sorry if this question is unrelated, but how can I enable this option with `wasm-pack`?

I've tried to add following to my `Cargo.toml`, but this doesn't seem to do the trick...
```toml
[package.metadata.wasm-pack.profile.release.wasm-bindgen]
reference-types = true
```

alexcrichton:
Sorry I'm not faimliar enough with wasm-pack to know how to pass this argument. This may need a patch to wasm-bindgen.

gilesv:
@Urhengulas There is an open PR which adds the `--reference-types` option to wasm-pack. https://github.com/rustwasm/wasm-pack/pull/888/
We will be able to use it as soon as it gets merged.

PR #2266, opened 2020-08-03
Update `Map.for_each` docs

pauldorehill:
Hello,

I'm currently taking a deep dive into wasm + rust and had this unexpected expected behavior happen to me: Javascript `foreach` methods are reversed with respect to the ordering of `key` and `value` compared to the `normal`. So I thought that it might be nice to explicitly state this in the docs.


alexcrichton:
Thanks!

PR #2267, opened 2020-08-03

grovesNL:
Update to latest WebGPU WebIDL as of July 29.

There are some changes from the upstream WebIDL (mostly to match Gecko's snapshot of the WebIDL/implementation status):
- Define `GPUCode` as `typedef (USVString or Uint32Array) GPUCode;` to allow both SPIR-V and WGSL for now
- Continue to use the old signature for `set_index_buffer` because that's what Gecko is currently using
- Continue to use the old signatures for buffer mapping because that's what Gecko is currently using


alexcrichton:
Thanks!

Issue #2200, opened 2020-06-12
memory import undefined


jregistr:
I've been going through the [rust wasm book](https://rustwasm.github.io/docs/book/game-of-life/) and I've gotten stuck on a problem.
### Summary

I'm have an issue where `memory`
```ts
import {memory} from "wasm-rust/index_bg";
```
is undefined when I import it. So when I try to access buffer, it ofc errors out. Are there any known situations that would cause this?

### Additional Details

I'm using the Webpack [Wasm-pack plugin](https://www.npmjs.com/package/@wasm-tool/wasm-pack-plugin).
The project is can be found here: https://github.com/jregistr/wasm-game-of-life

alexcrichton:
I believe you'll need to import `index_bg.wasm` since `index_bg.js` also exists, but that should do the trick!

noloerino:
I'm running into a similar issue in a project, and importing `*_bg.wasm` does indeed do the trick in vanilla JS. I'd like to have my project in Typescript though, and when I try to import with the .wasm extension (`import { memory } from "duna_frontend/duna_frontend_bg.wasm"`), it complains about being unable to find the module or its type definitions - if I try to import without the .wasm extension (`import { memory } from "duna_frontend/duna_frontend_bg"`), then webpack builds but warns that `"export 'memory' was not found in 'duna_frontend/duna_frontend_bg'`. I'm having trouble determining where on my toolchain the issue is coming from (webpack? yarn? typescript? wasm-pack? Cargo config?), so I'm wondering if anyone has any insight into this.

For what it's worth, my code can be found here: https://github.com/noloerino/duna/tree/reg-display/duna_frontend (JS stuff is in `www/`).

alexcrichton:
I think that may be a bug in our configuration right now in that case. We're generating `foo_bg.js` and `foo_bg.wasm`, but we're also generating `foo_bg.d.ts` which I think corresponds to just the `foo_bg.wasm` file. We'll probably need to rename something to get everything working in concert.

jregistr:
FWIW: I noticed I couldn't import the .wasm file in TS so I changed to a javascript file and memory is still undefined. I also tried using wasm-pack over the webpack plugin version and still same issue. Something else must be off with my configuration....

henriiik:
Hello!

We are running into this as well. We have found a workaround that works for us, change the import like @alexcrichton said:

[...]

PR #2283, opened 2020-08-13
ts declaration file name for wasm import

alxkhr:
Fixes import of wasm memory in a typescript setup with type safety #2200


alexcrichton:
Thanks!

PR #2288, opened 2020-08-19
Add userVisibleOnly property to PushSubscriptionOptionsInit

### LINKED ISSUE #2287: Add userVisibleOnly to web_sys::PushSubscriptionOptionsInit
### Motivation

I would like to subscribe to push notifications using a service worker. This can easily be achieved by calling the `subscribe` function on the PushManager. The subscribe function takes an object with the following properties:
1. userVisibleOnly
1. applicationServerKey
* See mdn docs: https://developer.mozilla.org/en-US/docs/Web/API/PushManager/subscribe

web-sys currently supports the `applicationServerKey` but does not include support for the `userVisibleOnly` property:
https://docs.rs/web-sys/0.3.44/web_sys/struct.PushSubscriptionOptionsInit.html

If this property is not set to true, subscribing with an application server key will fail.

### Proposed Solution

I would like to add the `userVisibleOnly` property to `web-sys::PushSubscriptionOptionsInit`.

### Alternatives

I do not have any alternatives at this time.

### Additional Context

Without setting this property to true, the subscription will fail. Here is the error message I receive:

> Chrome currently only supports the Push API for subscriptions that will result in user-visible messages. You can indicate 
> this by calling pushManager.subscribe({userVisibleOnly: true}) instead. See https://goo.gl/yqv4Q4 for more details.
> index_bg.js?6119:336 panicked at 'Couldn't subscribe to push manager.: JsValue(NotAllowedError: Registration failed - 
> permission denied

Here is an example of how I am subscribing to push notifications with service worker:

```
fn url_b64_to_uint_8_array(base_64_string: &str) -> js_sys::Uint8Array {
    let padding = std::iter::repeat('=').take((4 - (base_64_string.len() % 4)) % 4).collect::<String>();
    let base64 = format!("{}{}", base_64_string, &padding).replace('-', "+").replace('_', "/");
    let window = web_sys::window().expect("Couldn't get window.");
    
    let raw_data: String = window.atob(&base64).expect("couldn't call atob.");
    let output_array: js_sys::Uint8Array = js_sys::Uint8Array::new_with_length(raw_data.chars().count() as u32);
    

    let mut pos = 0;
    for c in raw_data.chars() {
        log!("Setting value", c, "in position", pos);
        output_array.set_index(pos, c as u8);
        pos = pos + 1;
    }

    output_array
}

        let window = web_sys::window().expect("Couldn't get window.");
        let sw_container = window.navigator().service_worker();

        let p = sw_container.register("/service-worker.js");
        let reg = wasm_bindgen_futures::JsFuture::from(p).await;

        if let Ok(reg) = reg {
            let sw_reg: web_sys::ServiceWorkerRegistration = reg.into();
            let manager: web_sys::PushManager = sw_reg.push_manager().unwrap();

            let mut subscription_options = web_sys::PushSubscriptionOptionsInit::new();
            let key = "<key>";
            let encoded_key = url_b64_to_uint_8_array(key);

            subscription_options.application_server_key(
                Some(&encoded_key)
            );
            //subscription_options.user_visible_only(Some(true));   <-- This is where I would like to set the userVisibleOnly property


            let subscribe = manager.subscribe_with_options(&subscription_options).unwrap();
            wasm_bindgen_futures::JsFuture::from(subscribe).await.expect("Couldn't subscribe to push manager.");

            if let Some(x) = sw_reg.installing() {
                log!("Service worker installing.");
            } else if let Some(x) = sw_reg.waiting() {
                log!("Service worker waiting.");
            } else if let Some(x) = sw_reg.active() {
                log!("Service worker active.");
            }
        } else {
            log!("Error registering service worker.");
        }
```

johnsonw:
Fixes #2287.

The PushSubscriptionOptionsInit is currently missing the
`userVisibleOnly` property.

Signed-off-by: johnsonw <wjohnson@whamcloud.com>

### DIFF
--- a/crates/web-sys/src/features/gen_PushSubscriptionOptionsInit.rs
+++ b/crates/web-sys/src/features/gen_PushSubscriptionOptionsInit.rs
@@ -36,4 +36,21 @@ impl PushSubscriptionOptionsInit {
         let _ = r;
         self
     }
+    #[doc = "Change the `userVisibleOnly` field of this object."]
+    #[doc = ""]
+    #[doc = "*This API requires the following crate features to be activated: `PushSubscriptionOptionsInit`*"]
+    pub fn user_visible_only(&mut self, val: bool) -> &mut Self {
+        use wasm_bindgen::JsValue;
+        let r = ::js_sys::Reflect::set(
+            self.as_ref(),
+            &JsValue::from("userVisibleOnly"),
+            &JsValue::from(val),
+        );
+        debug_assert!(
+            r.is_ok(),
+            "setting properties should never fail on our dictionary objects"
+        );
+        let _ = r;
+        self
+    }
 }
--- a/crates/web-sys/webidls/enabled/PushManager.webidl
+++ b/crates/web-sys/webidls/enabled/PushManager.webidl
@@ -8,7 +8,7 @@
 */
 
 dictionary PushSubscriptionOptionsInit {
-  // boolean userVisibleOnly = false;
+  boolean userVisibleOnly = false;
   (BufferSource or DOMString)? applicationServerKey = null;
 };
 


johnsonw:
The errors seem to be related to the following:
```
error: failed to run custom build command for `curl-sys v0.4.35+curl-7.71.1`
```

alexcrichton:
Thanks!

PR #2294, opened 2020-08-21
Changed externref_table to use geometric resizing, giving amortized constant time additions

### LINKED ISSUE #2291: Adding JsValues to Vec<JsValue> takes quadratic time with reference types enabled for first run.
### Describe the Bug

I wanted to try out reference types. In testing, it sped up my code, except on first run, where it is slowed it down dramatically. It appears to take quadratic time with respect to number of items added.

I've isolated it to adding items to a vec.

Addin 1.2 million `JsValue`s to a vec, using reference types made the first run go from ~.2-~.3 seconds (not using them) to ~150 seconds. 

(the JsValue was always 1)

 Subsequent runs after removing all items were faster than the first. (Only <0.7 seconds for 1.2 million)

Removing all items from first, and then re-adding 1 million items to a second object was also faster.

### Steps to Reproduce

1. [Install wasm-pack with reference type support using this branch (remember to switch branches) using cargo install while in the directory](https://github.com/Tarnadas/wasm-pack/tree/feature/anyref)
2. [Clone this minimal reproduction repo](https://github.com/richardlett/perf-test), `
3. run` wasm-pack build --release --ref-types --target web` to build
4.  Create a static http server at root, using `python3 -m http.server` or similar, and open in firefox. After a while, in the console you will see
image
 
wasm-push is 1.2 million items to a vec. wasm-pop is removing them all. Wasm-pushpop does both.

Expected behavior: 

Take linear time to add N items to  a vec.

Actual behavior: 

It seems to take quadratic time. Here's a table of time taken for first N items added vs a chart.

image

Sorry if this has already been reported.

richardlett:
Previously `__wbindgen_externref_table_grow` was called in `alloc `adding a constant size each time.

This results in taking Theta(N) time amortized to pass a single new `JsValue `to wasm, where N is number/size of existing items.

**Changes**

This pull request resizes geometrically, giving amortized constant additions to the externref_table.  

Specifically, `max(128, curr_len)`, so that we the larger of 128*usize, or the size of the current array.

This results in a growth factor of 2, the same as Rust's `Vec` growth factor

**Testing**:

It fixes #2291, where it took over 150 seconds to add 1.2 million `JsValues`'s to a Vec on my PC with reference types enabled.   
Now, it takes only 0.7-0.9 seconds.

----

**Pros**
Reduces complexity from` Theta(N^2)` to pass` N` `JsValue`s to wasm to `Theta(N)`
**Cons**
Slightly more memory required -- After resizing, the table will be half full.

I'm setting this as draft, since this is my first PR! :) 
All tests passed for me. Tell me if I missed anything!

### DIFF
--- a/src/externref.rs
+++ b/src/externref.rs
@@ -5,6 +5,7 @@ use std::mem;
 use std::ptr;
 use std::slice;
 use std::vec::Vec;
+use std::cmp::max;
 
 externs! {
     #[link(wasm_import_module = "__wbindgen_externref_xform__")]
@@ -32,8 +33,9 @@ impl Slab {
     fn alloc(&mut self) -> usize {
         let ret = self.head;
         if ret == self.data.len() {
-            if self.data.len() == self.data.capacity() {
-                let extra = 128;
+            let curr_len = self.data.len();
+            if curr_len == self.data.capacity() {
+                let extra = max(128, curr_len);
                 let r = unsafe { __wbindgen_externref_table_grow(extra) };
                 if r == -1 {
                     internal_error("table grow failure")


richardlett:
No problem! I think its all good - Submitted as draft, as I was unfamiliar with the pull request system. 👍

alexcrichton:
Ok no worries, thanks again!

PR #2297, opened 2020-08-26
Try to fix #2296 and support to get wasm file through http protocol

### LINKED ISSUE #2296: Deno target on windows error.
### Describe the Bug

When I use the deno target, there will be an error that {name}_bg.wasm cannot be found

### Steps to Reproduce

1. Run `wasm-pack build --target=deno --out-name=deno_test`
2. `import { ... } from './pkg/deno_test.js'`

### Actual Behavior

`error: Uncaught NotFound: 系统找不到指定的文件。 (os error 2)`(The system can not find the file specified)

### Additional Context
os: Windows10 19042.450 (20H2)
`import.meta.url`: file:///C:/Users/soeur/CLionProjects/deno-deno_test/pkg/deno_test.js
`const file = new URL(import.meta.url).pathname`: /C:/Users/soeur/CLionProjects/deno-deno_test/pkg/deno_test.js

### I think...
The reason for this problem is that in my case, os==windows but the separator is `/` instead of `\`, and the first character of URL.pathname is `/`

juzi5201314:
Fix #2296


alexcrichton:
Thanks! Looks like the deno CI is failing though?

juzi5201314:
> Thanks! Looks like the deno CI is failing though?

Yes, I made a stupid mistake. Now it has been fixed and passed ci.

juzi5201314:
In deno, the usual method is to get the library directly from remote.
e.g. `import {...} from'https://deno.land/x/lib/mod.ts'`,
But `Deno.readFileSync` can only read wasm files from the local file system.
So I use `fetch` to complete the support for getting wasm files from http.

alexcrichton:
cc @jakobhellermann mind taking a look at this?

crates/cli-support/src/js/mod.rs
```diff
@@ -318,12 +318,29 @@ impl<'a> Context<'a> {
...
+                }}
+                wasmCode = Deno.readFileSync(file)
+            }} else if (url.protocol.includes('http')) {{
+                const wasm_url = import.meta.url.substring(0, import.meta.url.lastIndexOf('/') + 1) + '{module_name}_bg.wasm'
+                wasmCode = await (await fetch(wasm_url)).arrayBuffer()
+            }} else {{
+                console.error(`Unsupported protocol: ${{url.protocol}}`)
+                Deno.exit(0)
+            }}
+            const wasmModule = new WebAssembly.Module(wasmCode);
```
RReverser, line 340: You should almost never use synchronous constructors, as they can be very slow for large Wasm. Instead, use `await WebAssembly.instantiate(wasmCode, imports)`.

zebp:
Can this get some movement? This issue makes it almost impossible to publish a Deno library that uses wasm-bindgen without modifying the generated output.

RReverser:
I'll admit I don't fully understand the original issue. Given that Deno supports all the required Web APIs, doesn't `--target web` just work in Deno?

zebp:
> I'll admit I don't fully understand the original issue. Given that Deno supports all the required Web APIs, doesn't `--target web` just work in Deno?

It almost works with the web target. Deno's `fetch` doesn't support fetching local files, so it can't fetch the wasm binary.

RReverser:
> Deno's fetch doesn't support fetching local files, so it can't fetch the wasm binary.

Oh I see, that's unfortunate.

juzi5201314:
Glad someone saw this problem.  Is there any resistance to this problem at the moment? 
 How can I help?

zebp:
It'd be great to get some movement on this PR, this issue is a blocking one for me.

crates/cli-support/src/js/mod.rs
RReverser: ^ this issue still stands. (I'm not a maintainer, but I don't think we should ever block)

crates/cli-support/src/js/mod.rs
```diff
@@ -318,12 +318,29 @@ impl<'a> Context<'a> {
         // Deno removed support for .wasm imports in https://github.com/denoland/deno/pull/5135
         // the issue for bringing it back is https://github.com/denoland/deno/issues/5609.
         format!(
-            "const file = new URL(import.meta.url).pathname;
-            const wasmFile = file.substring(0, file.lastIndexOf(Deno.build.os === 'windows' ? '\\\\' : '/') + 1) + '{}_bg.wasm';
-            const wasmModule = new WebAssembly.Module(Deno.readFileSync(wasmFile));
+            "const url = new URL(import.meta.url)
+
+            let wasmCode = ''
+            if (url.protocol.includes('file')) {{
```
RReverser, line 324: Better to check protocols explicitly instead of "includes". Maybe change this to a switch-case?

crates/cli-support/src/js/mod.rs
juzi5201314: ok! https://github.com/rustwasm/wasm-bindgen/pull/2297/commits/f71515d2c7bf601b80df832388fdabccf70b926f

crates/cli-support/src/js/mod.rs
juzi5201314: https://github.com/rustwasm/wasm-bindgen/pull/2297/commits/4920e21c7ab857e75707b5ac305c798236267ca4

crates/cli-support/src/js/mod.rs
```diff
@@ -341,8 +341,7 @@ impl<'a> Context<'a> {
                 console.error(`Unsupported protocol: ${{url.protocol}}`)
                 Deno.exit(0)
             }}
-            const wasmModule = new WebAssembly.Module(wasmCode);
-            const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
+            const wasmInstance = await WebAssembly.instantiate(await WebAssembly.compile(wasmCode), imports);
```
RReverser, line 349: You don't need `compile` separately.

```suggestion
            const wasmInstance = (await WebAssembly.instantiate(wasmCode, imports)).instance;
```

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
+                    let file = new URL(import.meta.url).pathname;
+                    if (Deno.build.os === 'windows') {{
+                        if (file.startsWith('/')) file = file.substr(1);
+                        file = await Deno.realPath(file + '/../{module_name}_bg.wasm');
+                    }} else {{
+                        file = file.substring(0, file.lastIndexOf('/') + 1) + '{module_name}_bg.wasm';
+                    }}
+                    wasmCode = await Deno.readFile(file);
+                    break
+                case 'http:':
```
RReverser, line 339: I'm assuming you want it to work for `https` too?

```suggestion
                case 'http:':
                case 'https:':
```

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
+                    const wasm_url = import.meta.url.substring(0, import.meta.url.lastIndexOf('/') + 1) + '{module_name}_bg.wasm';
+                    wasmCode = await (await fetch(wasm_url)).arrayBuffer();
+                    break
+                default:
+                    console.error(`Unsupported protocol: ${{url.protocol}}`);
+                    Deno.exit(0);
```
RReverser, line 345: Better to just throw an error and give a caller a chance to handle it, rather than forcefully close the program.

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
+                    }}
+                    wasmCode = await Deno.readFile(file);
+                    break
+                case 'https:':
+                case 'http:':
+                    const wasm_url = import.meta.url.substring(0, import.meta.url.lastIndexOf('/') + 1) + '{module_name}_bg.wasm';
+                    wasmCode = await (await fetch(wasm_url)).arrayBuffer();
+                    break
+                default:
+                    throw `Unsupported protocol: ${{url.protocol}}`;
```
RReverser, line 345: Throwing strings is a bad practice, since they don't have any stacktrace attached. Best to throw a proper `Error` instance:

```suggestion
                    throw new Error(`Unsupported protocol: ${{url.protocol}}`);
```

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
-            "const file = new URL(import.meta.url).pathname;
-            const wasmFile = file.substring(0, file.lastIndexOf(Deno.build.os === 'windows' ? '\\\\' : '/') + 1) + '{}_bg.wasm';
-            const wasmModule = new WebAssembly.Module(Deno.readFileSync(wasmFile));
-            const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
+            "const url = new URL(import.meta.url);
+
+            let wasmCode = '';
+            switch (url.protocol) {{
+                case 'file:':
+                    let file = new URL(import.meta.url).pathname;
```
RReverser, line 330: You already have `url` instance from above:

```suggestion
                    let file = url.pathname;
```

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
+                        if (file.startsWith('/')) file = file.substr(1);
+                        file = await Deno.realPath(file + '/../{module_name}_bg.wasm');
+                    }} else {{
+                        file = file.substring(0, file.lastIndexOf('/') + 1) + '{module_name}_bg.wasm';
+                    }}
+                    wasmCode = await Deno.readFile(file);
+                    break
+                case 'https:':
+                case 'http:':
+                    const wasm_url = import.meta.url.substring(0, import.meta.url.lastIndexOf('/') + 1) + '{module_name}_bg.wasm';
```
RReverser, line 341: It's more reliable to use real URL resolution instead of string manipulation. You can probably reuse this code: https://github.com/rustwasm/wasm-bindgen/blob/8594cae64b435cb5ab4d5233336fd15b3cb8cec2/crates/cli-support/src/js/mod.rs#L648

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
+
+            let wasmCode = '';
+            switch (url.protocol) {{
+                case 'file:':
+                    let file = new URL(import.meta.url).pathname;
+                    if (Deno.build.os === 'windows') {{
+                        if (file.startsWith('/')) file = file.substr(1);
+                        file = await Deno.realPath(file + '/../{module_name}_bg.wasm');
+                    }} else {{
+                        file = file.substring(0, file.lastIndexOf('/') + 1) + '{module_name}_bg.wasm';
```
RReverser, line 335: Same here about URL manipulation.

crates/cli-support/src/js/mod.rs
```diff
@@ -322,12 +322,33 @@ impl<'a> Context<'a> {
...
-            const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
+            "const url = new URL(import.meta.url);
+
+            let wasmCode = '';
+            switch (url.protocol) {{
+                case 'file:':
+                    let file = new URL(import.meta.url).pathname;
+                    if (Deno.build.os === 'windows') {{
+                        if (file.startsWith('/')) file = file.substr(1);
+                        file = await Deno.realPath(file + '/../{module_name}_bg.wasm');
```
RReverser, line 333: Why do we use `../` on Windows but not elsewhere?

crates/cli-support/src/js/mod.rs
juzi5201314: Sorry and thank you.
I'm a newbie, I didn't know that URL has this kind of usage before this. So some codes used bad methods to try to get the `_bg.wasm` file in the same directory.
Now the code has become clean🎉

RReverser:
@juzi5201314 Didn't see the updates following the last round of reviews, thanks, code looks a lot better!

PR #2310, opened 2020-09-15
cli-support: Remove Node.js specific passStringToWasm

josephlr:
Reverts #1391 (CC #1313)

Now that Node has encodeInto it can use the same path as the normal Web
Browser code (that avoids calling into the C++ code, see #1470 ). Node 12,13, and 14
have this method, so it is now fairly well supported.

@alexcrichton @RReverser, what benchmarks should I run to see if this is actually a speedup for ASCII strings? It looks like the existing benchmarker framework is intended for the Web Broswer.

Signed-off-by: Joe Richey <joerichey@google.com>

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1003,37 +1003,6 @@ impl<'a> Context<'a> {
             ""
         };
 
-        // If we are targeting Node.js, it doesn't have `encodeInto` yet
-        // but it does have `Buffer::write` which has similar semantics but
-        // doesn't require creating intermediate view using `subarray`
-        // and also has `Buffer::byteLength` to calculate size upfront.
-        if self.config.mode.nodejs() {
-            let get_buf = self.expose_node_buffer_memory(memory);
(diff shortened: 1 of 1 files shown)


alexcrichton:
Sounds reasonable to me, thanks!

RReverser:
> what benchmarks should I run to see if this is actually a speedup for ASCII strings?

I don't think it should, because we already used `Buffer.write` in Node.js. While it was Node-specific, it provided same functionality as the experimental `encodeInto` in browsers - writing to an existing memory view.

josephlr:
> > what benchmarks should I run to see if this is actually a speedup for ASCII strings?
> 
> I don't think it should, because we already used `Buffer.write` in Node.js. While it was Node-specific, it provided same functionality as the experimental `encodeInto` in browsers - writing to an existing memory view.

Oh I mainly meant the initial ASCII loop (that's just pure JS) that occurs before `encodeInto` that was added in #1470, but not for Node.

RReverser:
Ah right, that might have some impact.

Issue #2314, opened 2020-09-20
Link to webaudio example is 404


guest271314:
### Describe the Bug

Link to webaudio example is 404

### Steps to Reproduce

1. Go to https://github.com/rustwasm/wasm-bindgen/tree/master/examples/webaudio 
2. Click on 'View compiled example online'

If applicable, add a link to a test case (as a zip file or link to a repository we can clone).

https://rustwasm.github.io/wasm-bindgen/exbuild/web-audio/

### Expected Behavior

Link to example to not be 404

### Actual Behavior

Link to example is 404

Xuanwo:
It's a typo and I made a PR to fix it.

PR #2315, opened 2020-09-20
example/webaudio: Fix typo for the link to compiled example

Xuanwo:
Fix https://github.com/rustwasm/wasm-bindgen/issues/2314

---

Maybe we should normalize all `web-audio` to `webaudio` instead?


alexcrichton:
Thanks! And yeah using just one form should be fine.

PR #2316, opened 2020-09-20
Add dataview attribute output to WebIDL Codegen

qdot:
WebBluetooth and some other specs use DataView as an attribute type. Tested against my project using WebBluetooth, seems to work fine (Wrapped resulting buffer in UInt8Array to go to &[u8], everything seems to work). Not sure if there were other tests I need to add here.


alexcrichton:
Thanks!

PR #2342, opened 2020-11-03
lib: take_last_exception; remove return

Horki:
@alexcrichton removed return from `take_last_exception`; please test

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1080,7 +1080,7 @@ pub mod __rt {
             };
             GLOBAL_EXNDATA[0] = 0;
             GLOBAL_EXNDATA[1] = 0;
-            return ret;
+            ret
         }
     }
 


alexcrichton:
Thanks for the PR! Could you describe in a bit more detail though what the purpose of this is?

Horki:
@alexcrichton to implement the `implicit return`, AFAIK rust convection is to use return statement *only* for "early returns"
(from "Programming Rust" book - Jim Blandy & Jason Orendorff)
- fell free to close branch, due to insignificant change

alexcrichton:
Ah ok, thanks for the info! I personally prefer the style in the source for returning a bare variable name, so I'm going to close in that case.

Horki:
Thanks for the quick feedback, oddly this is the only place where explicit return is consumed; tough it would look cleaner and more idiomatic.

PR #2345, opened 2020-11-06
Added WebUSB API

hajifkd:
I added WebUSB IDL file. I took the spec from [here](https://wicg.github.io/webusb/).

In addition to adding the WebUSB spec, I added `DataView` as one of the builtin types.
I guess usually generated features in `crates/web-sys/feature` are to be copied to `crates/web-sys/Cargo.toml`. Since `DataView` is actually a builtin type but not dealt so, `DataView` feature is generated in Rust codes but not in `crates/web-sys/feature`. Thus, one need to add it to `Cargo.toml` by hand. This causes future confusion and to be fixed, I think. (IMHO, this should have been done in #2316.)
One drawback is that existing projects using WebBluetooth API via wasm-bindgen are broken, since `DataView` feature must be used there but it is now removed. But since WebBluetooth API is unstable feature of web-sys, it should be acceptable.


alexcrichton:
Looks good to me, thanks for this! Also agreed on the action taken with `DataView`, and we can always tweak it if necessary and breakage comes up.

hajifkd:
I noticed that `FrozenArray` is not implemented in `webidl` crate and some API does not work. The same problem actually exists for other APIs, like `Clipboard` and `Bluetooth`.
In my understanding, `FrozenArray` is a immutable array in Javascript. Since the mutability is not completely reflected in `web-sys` / `js-sys` crates (as long as I guess - is it correct?), I feel just generating the same codes as `Sequence` sounds fine. (I guess unifying `Sequence` and `FrozenArray` in https://github.com/rustwasm/wasm-bindgen/blob/3c0c2b39a44a584fa9ed984c663a59d08561d179/crates/webidl/src/idl_type.rs#L601-L607 would work.) 

Or, it might be also fine to intepret any type for which [`to_syn_type`](https://github.com/rustwasm/wasm-bindgen/blob/28d4575839b55dad58426353d06b67e342b9180c/crates/webidl/src/idl_type.rs#L498) returns `Err` as `JsValue`  in Rust although it partially spoils the advantage of the typed language.

What do you think?

alexcrichton:
Ah yeah makes sense! I think that's the JS type appearing there anyway so seems reasonable to just do the same thing as `Sequence`. Thanks for catching this!

hajifkd:
OK, will do!
Also I am going to add tests for `DataView` and `FrozenArray` in webidl-tests.
I cannot find any test for `Sequence` and `Promise` so I will add them as well.

hajifkd:
Is there any reason to ignore generics arguments of `Sequence`? It seems to be possible to append it. The compatibility for [these APIs](https://github.com/rustwasm/wasm-bindgen/search?l=IDL&q=Sequence) would be broken, though. In addition, `Promise` can be also typed.

alexcrichton:
Currently we don't have a great way to thread it through to the Rust definition right now and doing so would probably be a pretty large breaking change for web-sys as well. That being said it'd be better if we could do it so I'd love to have it on the docket for a future release. Was there anything else you wanted to do here before landing though?

hajifkd:
That makes sense. Let me squash all commits before merging.

alexcrichton:
Thanks again!

Issue #2202, opened 2020-06-13
ImageCapture API is not consistent with MDN


lwansbrough:
### Describe the Bug

https://github.com/rustwasm/wasm-bindgen/blob/master/crates/web-sys/webidls/enabled/ImageCapture.webidl

The WebIDL file for the ImageCapture API description provided by MDN (here: https://developer.mozilla.org/en-US/docs/Web/API/ImageCapture/takePhoto) is outdated/incorrect.

Specifically: `takePhoto` is a `Promise<Blob>`. The Rust wasm bindings provide the following method signature: `pub fn take_photo(&self) -> Result<(), JsValue>` as well as a number of non-standard (old/deprecated?) events.

### Expected Behavior

take_photo() returns a Promise which resolves a blob.

### Actual Behavior

ImageCapture API is unusable as there is no way to retrieve the Blob.

JAicewizard:
Shouldnt this one be closed now?

alexcrichton:
Indeed!

PR #2348, opened 2020-11-06
Fix ImageCapture API (#2347)

### LINKED ISSUE #2347: Expose and fix all ImageCapture APIs
### Describe the Bug

Not all ImageCapture API are exposed, maybe it is just because of Gecko not currently supporting some of them. The webidl contains some (maybe obsolete?) API that are different from the [current working draft](https://www.w3.org/TR/image-capture/).

### Steps to Reproduce

Open [ImageCapture.webidl](https://github.com/rustwasm/wasm-bindgen/blob/f8354b3a88de013845a304ea77d8b9b9286a0d7b/crates/web-sys/webidls/enabled/ImageCapture.webidl) to notice some attributes and methods are commented out. Compare them with the official specification to notice some differences.

### Expected Behavior

These features should be fixed and exposed in order to make them usable in `web-sys`, even if the implementation in Gecko is lacking.

### Actual Behavior

The fact that some features are commented out make them unavailable in `web-sys`, which a bit painful because using _extension traits_ is not doable when `wasm_bindgen` is involved -- if there is some workaround for this other than using a newtype, let me know. Obviously, having them available but not correctly implemented is bad as well.

dodomorandi:
I literally copy-pasted the webidl [from the working draft](https://www.w3.org/TR/image-capture/#imagecaptureapi) then run the `wasm-bindgen-webidl` to re-generate the bindings.

It looks fine, even if it could be considered breaking change because the methods depend on different set of features.

Closes #2347


alexcrichton:
Thanks! To confirm, were all these APIs basically broken and not working before? (asking because there's a bit of breakage)

dodomorandi:
I think that the API in the webidl file is very old (before December 2016), because I tried going back to a bunch of versions, without encountering something like the `onerror` attribute. Still being a working draft, there is a good possibility that this API was not used in production, even because the MDN shows a totally different interface respect to the webidl. I did not explicitly try all the APIs, but MDN shows a compatibility table that matches the _new_ API. 

In any case, I just realized that the issue #2202 was a subset of this problem.

alexcrichton:
If this WebIDL is still a draft, perhaps we should move it to the `unstable` directory?

dodomorandi:
You are probably right about this: the specs is in working draft, and it's not been touched since 21 June 2017. Surely I would rather have the current spec in the `unstable` than the old one in `enabled`. :blush:

If you want, I can move the file in this PR, it is a breaking change (even if probably irrelevant) in any case.

alexcrichton:
The current WebIDL is so old it doesn't work right? If so it seems fine to delete since it doesn't work anywhere anyway.

alexcrichton:
Thanks!

PR #2351, opened 2020-11-08
feat: cast JSValue from/to u64

### LINKED ISSUE #2350: Better support for u64 through JSValue::as_u64() and u64::into() -> JsValue
### Motivation

`BigInt` is now supported in stable releases of all modern browsers. Currently interfacing with `BigInt` generically in wasm-bindgen with JSValue is quite painful. `JSValue` is not aware of `BigInt` (and thus u64) at all.

### Proposed Solution

- Add a `impl From<u64> for JSValue`, that can be used to cast u64 to `BigInt`.
- Add `JSValue::as_u64(&self) -> Option<u64>` to cast `BigInt` to u64.

### Alternatives

This solution aligns closely to current implementation of primitives handling in `JSValue`. I am not aware of alternative implementation ideas, or if this can be built on top of wasm-bindgen.

lucacasonato:
This PR tries to add support for `JSValue::as_u64`, and `JSValue::from_u64`. I can't seem to get it to a functioning state though.

In Chrome 86 (with experimental wasm features enabled) my attempts fail the error below. I don't understand why this happens though. I would understand this error if this occurred on a exported function (because of the disallowed `i64` type in the function signature). But as that is not the case (function never directly touches JS), I don't get it.

```
TypeError: wasm function signature contains illegal type
            at wasm_bindgen::JsValue::from_u64::h34ae9b79359a621f (<anonymous>:wasm-function[8442]:0x1e2046)
            at <wasm_bindgen::JsValue as core::convert::From<u64>>::from::h1c482d09df9df836 (<anonymous>:wasm-function[9106]:0x1ec14c)
            at wasm::api::api_faz::h498cfaecbd4acab5 (<anonymous>:wasm-function[12674]:0x211e08)
            at api_faz (<anonymous>:wasm-function[8813]:0x1e7ba9)
            at Object.module.exports.api_faz (/mnt/f9/Projects/github.com/rustwasm/wasm-bindgen/target/wasm32-unknown-unknown/wbg-tmp/wasm-bindgen-test.js:504:20)
            at exports.js_works (/mnt/f9/Projects/github.com/rustwasm/wasm-bindgen/tests/wasm/api.js:13:29)
            at wasm::api::js_works::haa9d61ac0376a8a3 (<anonymous>:wasm-function[10685]:0x200618)
            at wasm::api::works::h78bff9502dc41e95 (<anonymous>:wasm-function[12841]:0x212712)
            at core::ops::function::FnOnce::call_once::h363f2b57eb52eb28 (<anonymous>:wasm-function[10732]:0x200def)
            at wasm_bindgen_test::__rt::Context::execute_sync::{{closure}}::h098a2e306ed5fa39 (<anonymous>:wasm-function[2625]:0x14384d)
```

This is my first time touching wasm-bindgen, or low level wasm in general, so it is also highly likely I am just completely misunderstanding how this should work. Could anyone point me in the right direction?

Closes #2350

### DIFF
--- a/crates/cli-support/src/intrinsic.rs
+++ b/crates/cli-support/src/intrinsic.rs
@@ -79,6 +79,10 @@ fn opt_f64() -> Descriptor {
     Descriptor::Option(Box::new(Descriptor::F64))
 }
 
+fn opt_u64() -> Descriptor {
+    Descriptor::Option(Box::new(Descriptor::U64))
+}
+
 intrinsics! {
     pub enum Intrinsic {
         #[symbol = "__wbindgen_jsval_eq"]
@@ -117,6 +121,9 @@ intrinsics! {
         #[symbol = "__wbindgen_number_new"]
         #[signature = fn(F64) -> Externref]
         NumberNew,
+        #[symbol = "__wbindgen_bigint_new"]
+        #[signature = fn(U64) -> Externref]
+        BigIntNew,
         #[symbol = "__wbindgen_string_new"]
         #[signature = fn(ref_string()) -> Externref]
         StringNew,
@@ -129,6 +136,9 @@ intrinsics! {
         #[symbol = "__wbindgen_number_get"]
         #[signature = fn(ref_externref()) -> opt_f64()]
         NumberGet,
+        #[symbol = "__wbindgen_bigint_get"]
+        #[signature = fn(ref_externref()) -> opt_u64()]
+        BigIntGet,
         #[symbol = "__wbindgen_string_get"]
         #[signature = fn(ref_externref()) -> opt_string()]
         StringGet,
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks for the PR!

I'm not precisely sure where this is coming from but it seems like the node being used might not support `i64` being used in the interface of a wasm module? I think that's where the bit about the illegal type is coming from.

00ff0000red:
Now, I may very well be completely wrong, but I belive the problem is that the Wasm imports `__wbindgen_bigint_new` as a JS function with the call signature of `(i64) -> JsValue`, so it proceeds to actually call it like that, with a Wasm `i64`, throwing the exception when invoking the host function.

As for why this could fail with experimental feature flags enabled, that is beyond me.

trevyn:
Is this closable now that https://github.com/rustwasm/wasm-bindgen/pull/2629 is merged, supporting `BigInt` conversions?

RReverser:
See https://github.com/rustwasm/wasm-bindgen/issues/2350#issuecomment-929563421, I think `JsValue::as_u64` part is still missing, so this PR or the like of it would be useful.

PR #2353, opened 2020-11-11

grovesNL:
Update to latest WebGPU WebIDL as of November 7.

Like in #2267, there are some temporary changes from the upstream WebIDL for Gecko:
- Define GPUCode as typedef (USVString or Uint32Array) GPUCode; to allow both SPIR-V and WGSL for now
- Continue to use the old signature for set_index_buffer because that's what Gecko is currently using

I have the latest rustfmt as far as I can tell, but regenerating the bindings seems to shift things around by adding extra spaces in attributes. Is there a specific version of rustfmt I should be using? I currently have a formatting commit but I'll remove it if I can fix this.


alexcrichton:
Thanks! According to CI though it looks like your local rustfmt may be doing something differently? Maybe a global config file? We should just be using the default rustfmt settings for the stable channel to format everything

crates/web-sys/Cargo.toml
```diff
@@ -632,7 +634,7 @@ IirFilterOptions = []
 ImageBitmap = []
 ImageBitmapFormat = []
 ImageBitmapRenderingContext = []
-ImageCapture = ["EventTarget"]
+ImageCapture = []
```
grovesNL, line 637: (this wasn't related to the WebGPU changes, but it was removed when I regenerated the bindings)

grovesNL:
Thanks for taking a look! Yeah I noticed the CI check failing and looked into this a bit more. I freshly cloned the repository and the formatting issue disappeared, so it turns out it wasn't related to rustfmt at all.

Instead it seems to be caused by the output of proc-macro2 when it's formatting the attribute initially. I had an older version in my Cargo.lock, so I updated proc-macro2 (`cargo update -p proc-macro2`) and the formatting is alright again. It was probably a small change in the proc-macro2 output (maybe alexcrichton/proc-macro2/pull/246).

I think everything should be alright now.

alexcrichton:
Ah right yes forgot about that!

PR #2360, opened 2020-11-19
Add support for renaming a struct field with js_name

### LINKED ISSUE #2331: js_name does not seem to effect struct values
### Describe the Bug

`js_name` does not affect public values in structs like it affects methods.

### Steps to Reproduce

Create a struct
```rust
#[wasm_bindgen]
pub struct FooBar {
    #[wasm_bindgen(js_name = "spamEggs")]
    pub spam_eggs: u32,
}
```

### Expected Behavior

Should generate JS/TS with name `spamEggs` just like for methods.

OR throw an error that this is not possible.

### Actual Behavior

Generates JS/TS with name `spam_eggs` as if the js_name attribute would not have been set.

CraftSpider:
fixes #2331

Added a test in `classes.js`/`.rs` to ensure that this behaves nicely, and that something else with the same name overrides it correctly.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -238,7 +238,8 @@ pub struct Struct {
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
 #[derive(Clone)]
 pub struct StructField {
-    pub name: syn::Member,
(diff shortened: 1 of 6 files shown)


alexcrichton:
Perfect, thanks so much for this!

PR #2365, opened 2020-11-20
Document the backend

CraftSpider:
Makes the backend crate `#![deny(missing_docs)]`, and documents all public values. Motivation is making contribution and development a little easier.

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -153,11 +153,13 @@ pub struct ImportFunction {
     pub variadic: bool,
     /// Whether the function should use structural type checking
     pub structural: bool,
-    /// For testing purposes
+    /// Causes the Builder (See cli-support::js::binding::Builder) to error out if
+    /// it finds itself generating code for a function with this signature
     pub assert_no_shim: bool,
     /// The kind of function being imported
     pub kind: ImportFunctionKind,
-    /// The shim name to use in the generated code
+    /// The shim name to use in the generated code. The 'shim' is a function that appears in
+    /// the generated JS as a wrapper around the actual function to import.
     pub shim: Ident,
     /// The doc comment on this import, if one is provided
     pub doc_comment: Option<String>,
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -1,3 +1,7 @@
+//! A representation of the Abstract Syntax Tree of a Rust program,
+//! with all the added metadata necessary to generate WASM bindings
+//! for it.
+
 use crate::Diagnostic;
 use proc_macro2::{Ident, Span};
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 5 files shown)


crates/backend/src/ast.rs
```diff
@@ -110,91 +123,144 @@ impl Hash for ImportModule {
...
 #[cfg_attr(feature = "extra-traits", derive(Debug))]
 #[derive(Clone)]
 pub enum ImportKind {
+    /// Importing a function
     Function(ImportFunction),
+    /// Importing a static value
     Static(ImportStatic),
+    /// Importing a type/class
     Type(ImportType),
+    /// Importing a JS enum
```
alexcrichton, line 136: I've personally never been a huge fan of docs like on this enum, the documentation doesn't really expand any more than the english name of the type itself, e.g. that `ImportKind::Function` is "importing a function" I think is pretty obvious. I feel that docs like this tend to clutter things up without adding much value unfortuantely.

crates/backend/src/ast.rs
```diff
@@ -110,91 +123,144 @@ impl Hash for ImportModule {
...
     pub rust_name: Ident,
+    /// The type being returned
     pub js_ret: Option<syn::Type>,
+    /// Whether to catch JS exceptions
     pub catch: bool,
+    /// Whether the function is variadic on the JS side
     pub variadic: bool,
+    /// Whether the function should use structural type checking
     pub structural: bool,
+    /// For testing purposes
```
alexcrichton, line 156: Could this be expanded with for what testing purposes this is used for?

(this is a public-facing attribute IIRC so it's not entirely just testing)

crates/backend/src/ast.rs
```diff
@@ -110,91 +123,144 @@ impl Hash for ImportModule {
...
     pub assert_no_shim: bool,
+    /// The kind of function being imported
     pub kind: ImportFunctionKind,
+    /// The shim name to use in the generated code
```
alexcrichton, line 160: FWIW the "shim" here I think could use more explanation. It's not 100% clear from nothing what the shim even is here I think.

alexcrichton (commented):
Thanks for the PR! Always nice seeing things documented!

I'm a bit worried though in that most of the documentation here just sort of looks at the struct/field name and doesn't document much more than the meaning behind that. I think it would be great to document with background knowledge how these fields interact, but if it's just reiterating the english name of the struct and the field then I'm worried that this is just checking off a box of "everything documented" without actually being too useful.

CraftSpider:
I at least tried to make sure to look through the 'usage' of each field, to make sure I didn't say anything false. But I agree some of them aren't much more useful than the name, and will admit to being motivated partially by a desire for completionism. Overall, at least with the struct names, I think it straddles the line but leans towards useful. EG with `ImportFunction`, 'A function being imported from JS' is mostly similar, but provides at least one useful confirmation: That 'import' means 'from JS'.
I'll make a second pass through and see if there are any subtleties I can mention (EG, on the ImportKind it's probably worth it to mention how it looks for that object / what makes it different from the others), and generally try to make sure they are useful for development. If you would like, I can strip out comments that don't end up providing more than a repetition of the name, and remove the `deny`. I personally prefer the style, but not so much that I would block the rest on it :P

crates/backend/src/ast.rs
CraftSpider: Practically, it causes the Builder for the JS to bail if it finds itself generating code for something with `assert_no_shim`. I'll add an explanation of the behavior.

alexcrichton:
Ok well in any case the words aren't hurting anything! I'll go ahead and merge, and thanks for the PR!

CraftSpider:
Thank you. I'll try to make a follow-up with any improvements, once I have the time (With the holiday, we'll see whether it happens this week or not).

Issue #2252, opened 2020-07-26
web-sys: ReadableStream is Empty


anlumo:
### Motivation

The [ReadableStream API](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) is [not exposed](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.ReadableStream.html) via web-sys. The struct itself does exist, but it does not implement any functions.

This API is used in the Fetch API for reading responses as a stream, rather than loading them completely before operating on them.

### Proposed Solution

Expose the API for the ReadableStream feature.

alexcrichton:
I think this probably  just needs an update to the WebIDL, and PRs are welcome to do that!

DesmondWillowbrook:
I'd be willing to do it, if someone could point me in the right direction!
The ReadableStream feature should provide Read by the very least, right?

DesmondWillowbrook:
Having a bit of trouble getting it to compile. Hope someone can help.

anlumo:
Yes, you're right.

PR #2373, opened 2020-11-27
Add complete WebIDL for ReadableStream 

DesmondWillowbrook:
To solve: #2252 
Found [this source](https://streams.spec.whatwg.org/#rs-class-definition), which is the official Streams spec.


alexcrichton:
Thanks for this! I think what's next is to follow [step 2 here](https://rustwasm.github.io/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html) and regenerate the bindings, and then this should be good to go!

DesmondWillowbrook:
Hey @alexcrichton! I tried following step 2 and came up with errors which I have written down in comments in the WebIdl file, as per this [commit](https://github.com/rustwasm/wasm-bindgen/pull/2373/commits/8c60b9a87a9b06169bd797c6827ff6207dc27d7d).

The build fails and the tool points the first square bracket '[' as the source of the error. I don't understand why that's happening, since as far as I can tell, there is nothing special about this file, and many other files also include similar annotations for the interfaces.

alexcrichton:
Hm can you gist the full error message you're receiving?

DesmondWillowbrook:
[Here it is.](https://gist.github.com/DesmondWillowbrook/7a1cbe6017d576b3cebd9189a06b2978)

alexcrichton:
Ah ok I see sorry I though you meant Rust compile error, but this is a WebIDl parse error. Unfortunately I'm not familiar enough with WebIDL syntax to know how to fix that.

DesmondWillowbrook:
The off-putting thing is that as far as I can see, there is nothing syntactically different about this file. Many other WebIDL files which look very similar compile normally, while this doesn't.

olanod:
Mmm wondering if someone knows about WebIDL to make this work, I'm kind of in need of this feature 🤔

MartinKavik:
The problem is actually on the line `ReadableStream.webidl:24`:
```webidl
async iterable<any>(optional ReadableStreamIteratorOptions options = {});
``` 
`async iterable` is a relatively new WebIDL feature, however it has been already integrated into the Rust WebIDL parser: https://github.com/rustwasm/weedle/pull/37.
But the change is still only in the [Weedle](https://crates.io/crates/weedle)'s `master` branch.
So to make the command for generating Rust items and features from WebIDLs work, you need to:
1. Change `weedle` dependency version in `crates/webidl/Cargo.toml` from `"0.11"` to `{ git = "https://github.com/rustwasm/weedle" }`. (Yeah, ideally, the new `weedle` version would be released and linked to this project to remove the `{ git = ".."}` workaround.)
2. Add something like the code below to `crates/webidl/src/first_pass.rs` to satisfy the Rust compiler.
```rust
InterfaceMember::AsyncIterable(_iterable) => {
    log::warn!("Unsupported WebIDL async iterable interface member: {:?}", self);
    Ok(())
}
```

DesmondWillowbrook:
So I deleted the fork, and seeing as Weedle's version has now been [updated](https://github.com/rustwasm/wasm-bindgen/pull/2427#issuecomment-765004027), I'll open a new pull request and try integrating the ReadableStream WebIDL now!

PR #2377, opened 2020-11-30
Bump to 0.2.69

### LINKED ISSUE #2371: New release?
### Summary

Could we create a new release sometime soon-ish? We're hitting the same problem as mentioned in #1374 and it would be nice if we could update to a new release to avoid this.

### Additional Details

We're trying to use an unpublished version of web-sys/wasm-bindgen from git that contains updated WebGPU bindings, but another (transitive) dependency depends on published releases. This is a problem because two versions of `wasm-bindgen-shared` can't be used currently:
> the package `wasm-bindgen-shared` links to the native library `wasm_bindgen`, but it conflicts with a previous package which links to `wasm_bindgen` as well

(this fails intentionally in this case due to https://github.com/rustwasm/wasm-bindgen/blob/master/crates/shared/Cargo.toml#L15-L18)

We've thought about a couple options to workaround it in the meantime, but neither work well for us:
- Use `patch` to patch all transitive dependencies to use the git version instead. This works alright, but decreases usability because all root crates need to do it too (cc gfx-rs/wgpu-rs#637, gfx-rs/wgpu-rs#644).
- Change all dependencies to use the git version temporarily. This is problematic because we can't easily maintain forks of the full dependency chain.

alexcrichton:
Closes #2371


alexcrichton:
@grovesNL to confirm I didn't see any WebGPU-specific changes in the commits, is that expected?

grovesNL:
@alexcrichton thank you! WebGPU changes should be located under commit https://github.com/rustwasm/wasm-bindgen/commit/316c5a70fdc4a052fb65ef82bf02d52107b5671b (basically anything prefixed with `Gpu`), so I think they're included?

alexcrichton:
Aha I see it now in `git log`, thanks for confirming! I apparently need to read things twice!

PR #2385, opened 2020-12-02
Adjust Bluetooth IDL for Rust

### LINKED ISSUE #2381: web-sys: Bluetooth usability issues
### Motivation

I'm porting Bluetooth using code over [from Javascript](https://gitlab.com/chrysn/coap-over-gatt/), and am running into a few warts around how web-sys exposes them.

As I'm relatively new to web-sys usage, I don't have a clear plan forward with them, and they may or may not be related; in this issue I'd like to hash out how they are related, and what can reasonably be done within being a -sys crate.

### Issues

* navigator().bluetooth() may fail: Not all browsers have Bluetooth support, and Mozilla probably won't until it has vastly changed.

  Currently this causes a panic at with `wasm-bindgen: imported JS function that was not marked as `catch` threw an error: getObject(...) is undefined`; making `.bluetooth()` fallible may help here.

* The [request_device](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Bluetooth.html#method.request_device) method is unusable as the default empty argument for options is not accepted, at least by Chrome. It either needs a concrete filter set, or acceptAllDevices set to true, in the [options object](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice#Syntax).

* Building the actual options object with a filter as a JsObject is comparatively verbose. To express

    ```js
    navigator.bluetooth.requestDevice({
            filters: [{services: [serviceUuid]}]
            });
    ````

    I had to go through a serde object and wrote

    ```rust
    #[derive(serde::Serialize)]
    struct ServiceFilter {
        services: [&'static str; 1]
    }
    let services = [ServiceFilter { services: [UUID_US] }];
    b.request_device_with_options(
        web_sys::RequestDeviceOptions::new().filters(
            &wasm_bindgen::JsValue::from_serde(&services).unwrap()
    )).then([...]
    ```

    Making what MDN calls BluetoothScanFilters available in Rust might help, but so would an example of how to do this more in-line in the JsValue documentation. (I'm pretty sure there's a way to write `RequestDeviceOptions::new().filters(js! [ {"services": [UUID_US]} ])`, I just didn't find it). If that's accessible easily, it may even be worth considering to drop RequestDeviceOptions in favor of going all JavaScript object (given most of the possible options incur going to a JsValue right again), possibly through a TryInto (but that wouldn't be zero-cost; in a regular API one could probably go from accepting RequestDeviceOptions to accepting any Into<JsValue> which RequestDeviceOptions could then implement for compatibility, but a) this is all unstable anyway, and b) given there's codegen involved, such stunts may be hard).

chrysn:
* navigator.bluetooth is optional -- it may be manatory by the
  WebBluetooth spec, but the Rust code should still be able to handle
  the errors resulting from a browser not implementing the spec at all
  without panicking.

* requestDevice: While it's marked optional upstream, practically it's
  mandatory (the default value, an empty dictionary, is not accepted).

Contributes-To: https://github.com/rustwasm/wasm-bindgen/issues/2381

---

These two don't fully resolve #2381, but the easy parts. The navigator.bluetooth part is necessary in the sense that without this, panics on unsupported browsers need manual workarounds to avoid. The latter is more of a usability thing (keeping name clutter low).

### DIFF
--- a/crates/web-sys/src/features/gen_Bluetooth.rs
+++ b/crates/web-sys/src/features/gen_Bluetooth.rs
@@ -204,17 +204,6 @@ extern "C" {
     #[doc = "[described in the `wasm-bindgen` guide](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html)*"]
     pub fn get_devices(this: &Bluetooth) -> ::js_sys::Promise;
     #[cfg(web_sys_unstable_apis)]
-    # [wasm_bindgen (method , structural , js_class = "Bluetooth" , js_name = requestDevice)]
-    #[doc = "The `requestDevice()` method."]
-    #[doc = ""]
-    #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice)"]
-    #[doc = ""]
-    #[doc = "*This API requires the following crate features to be activated: `Bluetooth`*"]
-    #[doc = ""]
-    #[doc = "*This API is unstable and requires `--cfg=web_sys_unstable_apis` to be activated, as"]
-    #[doc = "[described in the `wasm-bindgen` guide](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html)*"]
-    pub fn request_device(this: &Bluetooth) -> ::js_sys::Promise;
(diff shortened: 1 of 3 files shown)


alexcrichton:
Looks good to me, thanks!

PR #2390, opened 2020-12-03
Update `html_root_url` to point to version 0.3

thomaseizinger:
I believe this should be pointing to version 0.3!


alexcrichton:
Oops, indeed!

Issue #2386, opened 2020-12-03
Add the ability to avoid emitting mutable globals.


arilotter:
### Motivation

Exported mutable globals were not a thing in MVP WebAssembly, but are now part of the spec. 
Since version `0.2.66`, `wasm-bindgen` has generated WebAssembly that uses mutable globals, but not all WebAssembly engines in the wild support the mutable globals part of the spec:
e.g. `iOS 12` by virtue of shipping WebKit with OS updates (so every iPhone <= iPhone 6), `Firefox ESR 60,` and `Node.js 10` all don't support mutable globals. They run binaries made with `wasm-bindgen` <= `0.2.65` without mutable globals perfectly fine.

This means that you'll never be able to run WebAssembly binaries compiled with recent and future versions of `wasm-bindgen` on those platforms.

My company would like to support the iPhone 6, and I know that we're not the only ones using Rust that would like to run WebAssembly binaries on engines without the mutable globals feature, and thus being forced to pin to `wasm-bindgen` version `0.2.65`:
https://github.com/cormacrelf/citeproc-rs/commit/e0a3bddb9308c9877685e90819635f9a71b7aec4
https://github.com/cloudflare/rustwasm-worker-template/commit/fe5ce45da2415f3ac08b8cf2c6f337b109138914
https://github.com/chitalian/wasm-pack-template/commit/46f988e07495ffb6de59ad496ab15890ca5d3339
https://github.com/zakhenry/mesh-to-svg/commit/9597aecd0a67e1e3b300ac9cf81515d45555e9db
https://github.com/Emurgo/cardano-serialization-lib/commit/02432d5ae2712c26caf3113c111d749ff2f81735
https://github.com/p2panda/sesamoid/commit/43d81b3a709044986ecd7436fe57586a04d4fe15

### Proposed Solution

Add a flag `--disable-mutable-globals` to explicitly disable emitting mutable globals. If `wasm-bindgen` needs to export a variable in a way that's globally writable, `wasm-bindgen` should emit a setter instead.

At least one other project outside the Rust ecosystem implemented a similar workaround (emitting setters instead of mutable globals) to maintain backwards compatibility with older implementations of the WebAssembly spec:
https://github.com/AssemblyScript/assemblyscript/pull/1084/commits/e03b71d0cb49c9da9ab07df993d1d7d4552044e0

### Alternatives

[...]

PR #2391, opened 2020-12-04
Don't emit mutable globals

arilotter:
A proposed fix for https://github.com/rustwasm/wasm-bindgen/issues/2386, based on @alexcrichton's comment here: https://github.com/rustwasm/wasm-bindgen/issues/2386#issuecomment-738266000

We emit a shim function that sets the stack pointer, so we can avoid emitting mutable globals.
Mutable globals aren't supported everywhere Wasm is.

This PR seems to pass tests, but I believe it doesn't actually avoid exporting the stack pointer yet.
I'm not sure where to look for that, and I imagine we should add a test to ensure we actually don't export mutable globals.

This is my first foray into the `wasm-bindgen` codebase, so I expect there to be some mistakes in this code.

I think I might still need to modify `export_shadow_stack_pointer` to not export it to the wasm module...

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -11,7 +11,7 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt;
 use std::fs;
 use std::path::{Path, PathBuf};
-use walrus::{FunctionId, ImportId, MemoryId, Module, TableId};
+use walrus::{FunctionId, ImportId, MemoryId, Module, TableId, ValType};
 
 mod binding;
 
@@ -53,6 +53,9 @@ pub struct Context<'a> {
     /// names.
     memory_indices: HashMap<MemoryId, usize>,
     table_indices: HashMap<TableId, usize>,
+
+    /// A flag to track if the stack pointer setter shim has been injected.
+    stack_pointer_shim_injected: bool,
 }
 
 #[derive(Default)]
@@ -100,6 +103,7 @@ impl<'a> Context<'a> {
             aux,
             memory_indices: Default::default(),
             table_indices: Default::default(),
+            stack_pointer_shim_injected: false,
         })
     }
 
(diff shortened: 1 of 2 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -2179,6 +2184,10 @@ impl<'a> Context<'a> {
         pairs.sort_by_key(|(k, _)| *k);
         check_duplicated_getter_and_setter_names(&pairs)?;
 
+        if self.should_generate_stack_pointer_shim {
+            self.inject_stack_pointer_shim()?;
```
alexcrichton, line 2188: It's fine if this happens during `require_stack_pointer_shim`, and then the context's boolean prevents us from adding it twice.

crates/cli-support/src/js/mod.rs
```diff
@@ -3305,6 +3314,47 @@ impl<'a> Context<'a> {
...
+        // Create a shim function that mutate the stack pointer
+        // to avoid exporting a mutable global.
+        body.local_get(arg)
+            .global_get(stack_pointer)
+            .binop(BinaryOp::I32Add)
+            .local_tee(local)
+            .global_set(stack_pointer)
+            .local_get(local);
+
+        let add_to_stack_pointer_func = builder.finish(Vec::new(), &mut self.module.funcs);
```
alexcrichton, line 3351: I think the `Vec::new()` here is the list of arguments, right? And `arg` should be the one-element vector here?

crates/cli-support/src/js/mod.rs
```diff
@@ -3305,6 +3314,47 @@ impl<'a> Context<'a> {
...
+        };
+
+        // Create a shim function that mutate the stack pointer
+        // to avoid exporting a mutable global.
+        body.local_get(arg)
+            .global_get(stack_pointer)
+            .binop(BinaryOp::I32Add)
+            .local_tee(local)
+            .global_set(stack_pointer)
+            .local_get(local);
```
alexcrichton, line 3349: You can elide the local here by just doing a `global.set` followed by a `global.get` probably?

alexcrichton (commented):
Looks great to me, thanks for tackling this!

crates/cli-support/src/js/mod.rs
arilotter: Done, the local is removed, and we just do `global.set` then`global.get`. Funny that there's `local.tee` but no `global.tee`.

arilotter:
I've resolved all the comments from your preview review :)

alexcrichton:
Thanks!

stuartZhang:
Which version will include the improvement? I need it as well. @alexcrichton

alexcrichton:
I've done a version bump at https://github.com/rustwasm/wasm-bindgen/pull/2435

Issue #693, opened 2018-08-13
Add `namespace` and `function` to .d.ts file when in --no-modules mode


LukasKalbertodt:
I tried using the TS declaration file in the `--no-modules` mode. Example:

```rust
#[wasm_bindgen]
pub fn get_color(x: u8, y: u8) -> Color {
    Color { r: x, g: y, b: 0 }
}

#[wasm_bindgen]
pub struct Color {
    pub r: u8,
    pub g: u8,
    pub b: u8,
}
```

Results in this file:

```
/* tslint:disable */
export function get_color(arg0: number, arg1: number): Color;

export class Color {
r: number
g: number
b: number
free(): void;
}
```

But since this isn't a module anymore, I can't import it in my TS file via `import ...;`. Instead, I think, the correct was is to use namespaces. If I modify the file from above:

```
declare namespace wasm_bindgen {
    /* tslint:disable */
    export function get_color(arg0: number, arg1: number): Color;

    export class Color {
    r: number
    g: number
    b: number
    free(): void;
    }
}

declare function wasm_bindgen(path: string): Promise<void>;
```

I can simply place this modified file next to my real TS file and it works (as in I can call `wasm_bindgen("foo")` and `wasm_bindgen.get_color()`).

Currently, I use a build script to apply these modifications to the file emitted by `wasm-bindgen`. But I think this is what `wasm-bindgen` should always emit in `--no-modules` mode, right? But if I'm wrong somewhere (which could very well be), please just tell me ^_^

alexcrichton:
Thanks for the report! I definitely agree that `--no-modules` emits broken TS, and it'd be great to fix!

Would you like to send a PR for this?

LukasKalbertodt:
@alexcrichton I'm not sure if I'll find time to work on this. But regardless: could you maybe list the main files and functions involved in generating the TS declarations? Knowing where to start looking makes it way easier to start working on this -- for me or anyone else :)

alexcrichton:
Sure! Almost all the JS generation [happens here](https://github.com/rustwasm/wasm-bindgen/blob/master/crates/cli-support/src/js/mod.rs). Right now the `typescript` field of the context builds up the Typescript that's generated. It's all a bit of a mess in the sense that it's not too structured, but it shouldn't be to hard to add some structures and postprocessing around there

mizkichan:
For those who are looking around for a way how to load a `wasm-bindgen`-generated module from a worker. Following code at least just works anyway.

[...]

PR #2396, opened 2020-12-13
Getting the `no-modules` TypeScript `*.d.ts` Files Working

aaron-human:
Partially based on https://github.com/rustwasm/wasm-bindgen/issues/693.

The current `--target no-modules` TypeScript declaration files don't correctly describe what the JS interface looks like:
* The functions exported via `extern {` and `#[wasm_bindgen]` are stored in the `wasm_bindgen` namespace, but the TypeScript declaration files don't put them in any namespace.
* Everything else at the root level of the TypeScript file was `export`d. As of TypeScript 3.8.3, that meant it wasn't visible. So those were switched over to `declare` statements.

This tries to fix the above by adjusting the output for just `no-modules`. It also removes the bunch of extra spaces at the end of the declaration file.

As an example, the output went from:
```typescript
/* tslint:disable */
/* eslint-disable */
/**
* @param {number} id
*/
export function call_in(id: number): void;

export type InitInput = RequestInfo | URL | Response | BufferSource | WebAssembly.Module;

export interface InitOutput {
  readonly call_in: (a: number) => void;
}

/**
* If `module_or_path` is {RequestInfo} or {URL}, makes a request and
* for everything else, calls `WebAssembly.instantiate` directly.
*
* @param {InitInput | Promise<InitInput>} module_or_path
*
* @returns {Promise<InitOutput>}
*/
export default function init (module_or_path?: InitInput | Promise<InitInput>): Promise<InitOutput>;
        
```
To:
```typescript
declare namespace wasm_bindgen {
	/* tslint:disable */
	/* eslint-disable */
	/**
	* @param {number} id
	*/
	export function call_in(id: number): void;
	
}

declare type InitInput = RequestInfo | URL | Response | BufferSource | WebAssembly.Module;

declare interface InitOutput {
  readonly call_in: (a: number) => void;
}

/**
* If `module_or_path` is {RequestInfo} or {URL}, makes a request and
* for everything else, calls `WebAssembly.instantiate` directly.
*
* @param {InitInput | Promise<InitInput>} module_or_path
*
* @returns {Promise<InitOutput>}
*/
declare function wasm_bindgen (module_or_path?: InitInput | Promise<InitInput>): Promise<InitOutput>;

```
## Testing

When I ran `cargo test` and `cargo test --target wasm32-unknown-unknown` on my local setup, it reported all passes.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -335,7 +335,7 @@ impl<'a> Context<'a> {
         module_name: &str,
         needs_manual_start: bool,
     ) -> Result<(String, String, Option<String>), Error> {
-        let mut ts = self.typescript.clone();
+        let mut ts;
         let mut js = String::new();
         let mut start = None;
 
@@ -435,6 +435,16 @@ impl<'a> Context<'a> {
                 init = self.gen_init(needs_manual_start, Some(&mut imports))?;
                 footer.push_str("export default init;\n");
             }
+        }
+
+        // Before putting the static init code declaration info, put all existing typescript into a `wasm_bindgen` namespace declaration.
+        // Not sure if this should happen in all cases, so just adding it to NoModules for now...
+        if let OutputMode::NoModules { global : _ } = &self.config.mode {
+            ts = String::from("declare namespace wasm_bindgen {\n\t");
+            ts.push_str(&self.typescript.replace("\n", "\n\t"));
+            ts.push_str("\n}\n");
+        } else {
+            ts = self.typescript.clone();
         }
 
         let (init_js, init_ts) = init;
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -602,8 +602,7 @@ impl<'a> Context<'a> {
             * @returns {{Promise<InitOutput>}}\n\
             */\n\
             {setup_function_declaration} \
-                (module_or_path{}: InitInput | Promise<InitInput>{}): Promise<InitOutput>;
-        ",
+                (module_or_path{}: InitInput | Promise<InitInput>{}): Promise<InitOutput>;\n",
             memory_doc, arg_optional, memory_param,
             output = output,
             declare_or_export = declare_or_export,
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -546,13 +546,14 @@ impl<'a> Context<'a> {
         Ok(imports)
     }
 
-    fn declare_or_export(&self) -> String {
-        // With TypeScript 3.8.3, I'm seeing that any "export"s at the root level cause TypeScript to ignore all "declare" statements.
-        // So using "declare" everywhere for at least the NoModules option.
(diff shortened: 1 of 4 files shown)


alexcrichton:
Thanks for this! Would it be possible to add tests for this specifically too? I believe `crates/typescript-tests` doesn't currently use `--no-modules`, but perhaps a separate entry point and/or a separate test for `--no-modules` could be added?

Other than that the changes look great here!

aaron-human:
Ok, I'll see if I can setup some tests.

aaron-human:
Ok, the tests should now try to build with `no-modules` and then verify that TypeScript doesn't reject the built `typescript_tests.d.ts`. That's at least the file I usually use. Not sure if the `*.bg.d.ts` files also need to be tested. Or if the files need to be built against something (i.e. to produce an actual JS output file). What's in now was enough to catch when I manually dumped a junk string into `typescript_tests.d.ts`, so it's at least validating the syntax.

aheadmode:
@alexcrichton

alexcrichton:
Thanks! For the test though it looks like this just asserts that it parses correctly, but could some JS be written which consumes the wasm-bindgen generated JS? That way we can ensure the expected idioms all typecheck and information is all listed as we'd expect.

aaron-human:
@alexcrichton The tests now build the (no-modules) `*.d.ts` files against `*.ts` files very similar to those in `src`. The exact `*.ts` files were modified to try and remove the `import` statements and otherwise switch things away from using JS modules. I was aiming to avoid copying/committing the modified `*.ts` files in `src`; didn't want to make future maintainers have to manually keep the two variants in sync.

aaron-human:
Any chance this will get merged? It should be ready to review.

alexcrichton:
Er apologies I missed that this was ready!

I'm not sure what the test is doing though, what's with all the `sed`? I would hope that we wouldn't have to doctor the output, otherwise that seems like a bug that needs fixing in wasm-bindgen.

aaron-human:
The testing changes try to re-use the existing typescript testing that have been setup for the "--target web" case. The `sed` and `grep` calls in the script are to copy TypeScript files from "src" into a new (temporary) "src_no_modules" directory while also renaming things inside the files so they'll work with `--target no-modules` rather than `--target web`. I've added more comments so it's clearer that this is going on. Though the overview is that:

* The files have dropped any lines with "import". Since "no-modules" has no modules, imports aren't needed.
* The "wbg" namespace (which is where the tests imported the API to) has been replaced with "wasm_bindgen" (the namespace that "no-modules" exports into).
* I skipped all the TypeScript test files that mentioned importing the "../pkg/typescript_tests_bg.wasm" (by excluding files with "_bg.wasm") as I don't know what that file does.

I was aiming to re-use the existing tests rather than writing up new ones, as I'm not too well versed in all of the binding generation functionality. (The project that got me to set this up really only used "call Rust from JS" and "call JS from Rust" with fairly simple argument/return types.) I also figured it'd be nice to not have separate files for testing "--target web" vs "--target no-modules".

alexcrichton:
Ah I see, thanks for clarifying! Sounds good to me!

PR #2400, opened 2020-12-19
Allow constructors to be immutable and add it for ImageData

### LINKED ISSUE #2364: `ImageData.html::new_with_u8_clamped_array` should accept immutable `u8_clamped_array`
### Motivation

This requites piet-web to clone the entire image even when it is already in the right format, this is costly because images are big.

### Proposed Solution

`ImageData.html::new_with_u8_clamped_array` should accept an imitable `u8_clamped_array` because I suspect it is not needed (I'm sure you know that better then I do). 

This was originally mentioned in in #1005 by @raphlinus but that has since been closed.

JAicewizard:
This fixes #2364 and makes it easy to any constructor to me immutable.

Also includes a test to make sure that there is no magic things going on and that the original array is not modified(at least form the WASM side of things)


alexcrichton:
Thanks!

Issue #2399, opened 2020-12-19
Stream from AsyncIterator in wasm-bindgen-futures


olanod:
`AsyncIterator` in JS is the native equivalent of Promise to represent multiple future values instead of one just as in Rust we have `Future` and `Stream`(soon in the standard library). Instead of creating a separate utility crate that converts async iterators to Rust streams I propose having that in function(e.g. `iter_to_stream`) in the `wasm-bindgen-futures` crate since the feature(can be behind a flag) is very "futures" related.

mattgibb:
This would be awesome! Would it be possible to have the opposite too i.e. subscribe to a Rust `Stream` from Javascript via the registration of a callback/RxJS Observable or similar? Apologies in advance if this has an obvious solution, I am at the beginning of my journey with Rust...

I know already that you can call pre-existing global/static JS functions from Rust via an `extern "C"` block. However, it would be great to be able to manage subscriptions dynamically. I tried passing in a `&js_sys::Function` callback, but I couldn't get the callback to survive longer than the registration function:

```rust
use std::collections::HashMap;
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
#[derive(Default)]
pub struct VecU32<'a> {
  xs: Vec<u32>,
  subscribers: HashMap<u32, &'a js_sys::Function>,
  next_id: u32,
}

#[wasm_bindgen]
impl VecU32 {
  pub fn new() -> Self {
    Self {
      xs: vec![1, 2, 3],
      subscribers: HashMap::new(),
      next_id: 0,
    }
  }

  pub fn each(&self, callback: &js_sys::Function) {
    let this = JsValue::null();

    for &x in &self.xs {
      let x = JsValue::from(x);
      callback.call1(&this, &x);
    }
  }

  pub fn subscribe(&mut self, subscriber: &js_sys::Function) -> u32 {
    let id = self.next_id;
    self.subscribers.insert(id, subscriber);
    self.next_id += 1;
    id
  }

  pub fn unsubscribe(&self, subscription: u32) {
    self.subscribers.remove(&subscription);
  }
}
```

sisou:
@mattgibb Hi! Did you ever figure out how to do this? I'm at the same point now, having streams in Rust that I want to react to in JS, without a separate global listener function import. Either with callbacks like your code above, or by converting to an AsyncIterator.

hezhaozhao-git:
[...]

PR #2401, opened 2020-12-20
Convert async iterators to streams

olanod:
An initial implementation of what I proposed in https://github.com/rustwasm/wasm-bindgen/issues/2399 
It would be nice to have the ability to turn async iterators into Rust streams in the same way we can turn promises into futures. Instead of creating a separate crate for this utility I think it fits well in the scope of `wasm-bindgen-futures`.

### DIFF
--- a/crates/futures/Cargo.toml
+++ b/crates/futures/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2018"
 cfg-if = "1.0.0"
 js-sys = { path = "../js-sys", version = '0.3.46' }
 wasm-bindgen = { path = "../..", version = '0.2.69' }
+futures-core =  { version = '0.3.8', default-features = false }
 
 [target.'cfg(target_feature = "atomics")'.dependencies.web-sys]
 path = "../web-sys"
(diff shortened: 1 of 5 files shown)


alexcrichton:
Thanks for this! I'm a bit hesitant to do this at this  time though since it introduces `futures-core` as a public dependency. AFAIK `Stream` is intended at some point to make its way into the standard library, right? If so perhaps this could be gated behind an off-by-default Cargo feature?

olanod:
Yeah it looks like https://github.com/rust-lang/rust/pull/79023 could land anytime which would also remove the need for the dev-dependency. I'll put this under a `streams` feature if that's ok?

olanod:
@alexcrichton Added the feature under the `stream` flag even for the test because it seemed I would have to change the version of `wasm-bindgen-test` all over the place. Also mentioned the existence of feature in the Readme as experimental which can be made "stable" later when there is `core::stream::Stream` and perhaps also adding the feature of `Stream` to `AsyncIterator`([`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)?) conversion.

alexcrichton:
Looks good, thanks! For a small amount of extra future-proofing, though, could this perhaps be named `futures-core-03-stream`?

alexcrichton:
Thanks!

PR #2403, opened 2020-12-24
Add documentation for importing `export default`

codingkoi:
Added some documentation to the `js_name` page about using it to import
JavaScript modules that use `export default`.

I couldn't find this in the documentation, and only found out about it by finding a closed issue about it. I'm not sure if this is the best place for it. Maybe it could use its own page? It seems like a common enough pattern in JavaScript modules that calling it out seemed like it would be helpful. It certainly would have saved me some time.


alexcrichton:
Thanks!

PR #2408, opened 2020-12-29
Fix canvas is not a self closing tag.

ltzz:
I'm sorry I'm no good in English.
I got in trouble with the sample. It happened in Firefox 84.0.
canvas is not a self closing tag. As a result, the element written after canvas is not displayed in the browser.


alexcrichton:
Thanks!

PR #2411, opened 2021-01-03
cli-support: Remove default_module_path

diceride:
wasm-bindgen `target --web` generates code that expects the target to be an ES2015 module.

Generated code (`default_module_path`)

```js
async function init(input) {
        input = import.meta.url.replace(/\.js$/, '_bg.wasm');
    }

    ...
```

Error

```sh
Uncaught SyntaxError: Cannot use 'import.meta' outside a module
```

ES2015 modules for Web Workers are currently implemented in Chromium only ([constructor `type` option](https://developer.mozilla.org/en-US/docs/Web/API/Worker#Browser_compatibility)).

Example

```js
    new Worker("worker.js", { type: "module" });
```

List of supported browsers

- [Mozilla Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687): ~~Priority: Not set~~
- Chromium:
    - [Dedicated Workers](https://bugs.chromium.org/p/chromium/issues/detail?id=680046): Available since version 80
    - [Shared Workers](https://bugs.chromium.org/p/chromium/issues/detail?id=680046): Available since version 83
    - [Service Workers](https://bugs.chromium.org/p/chromium/issues/detail?id=824647): ~~Priority: P3~~
- [Safari](https://bugs.webkit.org/show_bug.cgi?id=164860): ~~Priority: P2 Normal~~
- Android: Available since version 80
- iOS: ~~Unknown~~

In general, providing a default path or expecting it to be of a certain filename or extension should be considered bad practice. 

Users should always provide a `module_or_path`, either a of type `InitInput` or `Promise<InitInput>` (as documented).

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -540,11 +540,7 @@ impl<'a> Context<'a> {
         Ok(imports)
     }
 
-    fn ts_for_init_fn(
-        &self,
-        has_memory: bool,
-        has_module_or_path_optional: bool,
-    ) -> Result<String, Error> {
+    fn ts_for_init_fn(&self, has_memory: bool) -> Result<String, Error> {
         let output = crate::wasm2es6js::interface(&self.module)?;
 
         let (memory_doc, memory_param) = if has_memory {
@@ -555,7 +551,7 @@ impl<'a> Context<'a> {
         } else {
             ("", "")
         };
-        let arg_optional = if has_module_or_path_optional { "?" } else { "" };
+
         Ok(format!(
             "\n\
             export type InitInput = RequestInfo | URL | Response | BufferSource | WebAssembly.Module;\n\
@@ -573,9 +569,9 @@ impl<'a> Context<'a> {
             * @returns {{Promise<InitOutput>}}\n\
             */\n\
             export default function init \
-                (module_or_path{}: InitInput | Promise<InitInput>{}): Promise<InitOutput>;
+                (module_or_path: InitInput | Promise<InitInput>{}): Promise<InitOutput>;
         ",
-            memory_doc, arg_optional, memory_param,
+            memory_doc, memory_param,
             output = output,
         ))
     }
@@ -610,29 +606,7 @@ impl<'a> Context<'a> {
             }
         }
 
-        let default_module_path = match self.config.mode {
-            OutputMode::Web => {
-                "\
-                    if (typeof input === 'undefined') {
-                        input = import.meta.url.replace(/\\.js$/, '_bg.wasm');
-                    }"
-            }
-            OutputMode::NoModules { .. } => {
-                "\
-                    if (typeof input === 'undefined') {
-                        let src;
-                        if (typeof document === 'undefined') {
-                            src = location.href;
-                        } else {
-                            src = document.currentScript.src;
-                        }
-                        input = src.replace(/\\.js$/, '_bg.wasm');
-                    }"
-            }
-            _ => "",
-        };
-
-        let ts = self.ts_for_init_fn(has_memory, !default_module_path.is_empty())?;
+        let ts = self.ts_for_init_fn(has_memory)?;
 
         // Initialize the `imports` object for all import definitions that we're
         // directed to wire up.
@@ -721,7 +695,6 @@ impl<'a> Context<'a> {
                 }}
 
                 async function init(input{init_memory_arg}) {{
-                    {default_module_path}
                     const imports = {{}};
                     {imports_init}
 
@@ -738,7 +711,6 @@ impl<'a> Context<'a> {
                 }}
             ",
             init_memory_arg = init_memory_arg,
-            default_module_path = default_module_path,
             init_memory1 = init_memory1,
             init_memory2 = init_memory2,
             start = if needs_manual_start {


aheadmode:
The CI is fundamentally broken.

I'm not familiar with Microsoft Windows. I can't debug this Microsoft Azure thing you're using here. It's all so tiresome ...

aheadmode:
Some CI pipelines are fundamentally broken

aheadmode:
@alexcrichton

alexcrichton:
I believe this feature was added specifically to avoid the need for supporting the URL, but since it's causing issues for you is there a way to fix your issue without removing the feature? (since that would be a breaking change)

aheadmode:
You are right, I am open to change the PR. Instead of deleting the code, the inclusion of the `default_module_path` generated code could be configurable with a new [wasm-bindgen option
](https://github.com/rustwasm/wasm-bindgen/blob/9554beef5fa929d7b3c8d68e582ee2084f945fae/guide/src/reference/cli.md#options)

Proposals

- **`--default-module-path`** - The `default_module_path` generated code is disabled by default, users can enable it

- **`--no-default-module-path`** - The `default_module_path` generated code is enabled by default, users can disable it

Let me know what you think.

In my opinion, the `default_module_path` generated code is not sound. A breaking change might be reasonable. 

Here are more objections

- The `default_module_path` generated code is biased and disfavors other solutions, eg. Fetch API, XMLHttpRequest (XHR).

- The `default_module_path` generated code expects the target to be an ES2015 module. If the target is not an ES2015 module, the code won't load. This is even the case when the wasm-bindgen generated code is bundled, eg. by [rollup.js](https://rollupjs.org/guide/en/).

- The `default_module_path` generated code expects the **wasm binary** to have the exact same URL as the ES2015 module by parsing the context-specific metadata exposed by the `import.meta` object. This is even the case when the wasm-bindgen generated code is bundled, eg. by [rollup.js](https://rollupjs.org/guide/en/).

alexcrichton:
Sorry I don't have a ton of domain knowledge in this area to know what solution is the best. All I know is that I would prefer to not have a breaking change at this time. If you're able to implement a change which works for you and isn't a breaking change for existing users I can take a look at it.

behunin:
@alexcrichton Can you or the team start a dev branch or repo? Please!
I understand that you don't want breaking changes but I also see quite a few breaking changes in the current open pull requests and not having some sort of dev branch or repo is hindering development.

alexcrichton:
I'm currently the only maintainer of this repository, and I do not have the time or effort to maintain two branches of wasm-bindgen. In the future if I get enough time I can try to figure out an 0.3 release, but that's unfortunately not the most likely thing in the world to happen. I think we're in general stuck with wasm-bindgen as-is unless someone wants to step up and help maintaining and help drive feature and future development of the project.

behunin:
I would be happy to maintain a development branch. 😃 Email me the details if you like.

saleemadat:
Me too count me in

alexcrichton:
Thanks for the offers, but to be clear, I'm not too interested in having someone to simply maintain a new branch. What I would be looking for is for someone to help out in-depth with the project to move the whole thing forward. A wasm-bindgen-next branch or something like that is fine as a temporary state of affairs but it shouldn't ever be permanent.

If an alternative "dev" branch is maintained that's effectively just a fork of the project which you can already do today at any time.

behunin:
That sounds wonderful @alexcrichton  
I have experience in rewriting generated Javascript/Typescript client code [See Commits](https://github.com/behunin/protobuf)
I see lots of room for improvement in this project.

1. I would like to split the typescript code from the javascript code 
    -- no more js + definition files
    -- .ts or .js files generated
2. Resolve importing and instantiation in all the various environments 
    -- CommonJs, EsNext, Deno(when it supports wasm)
3. Implement WebAssembly::{Memory, Table, Global} in generated code.
    -- Utilize these classes for better reference and memory management  
4. The list goes on...

I would like to see a dev crate for wasm-bindgen(wasm-bindgen-dev) so that new features and fixes can be tested there before moving to the mainline crate. So however you would like the github.com -> crates.io pipeline to run is fine with me.

alexcrichton:
Ok great! I think a number of those items may be PR-able today (which would be a good starting place), and otherwise if you want to get serious about another major version of wasm-bindgen I think it would be good to write up a plan for how it will work. For example I think it would be good to have a list of breakages in mind. I think there will also need to be some sort of guide or explanation of how to handle breakages within a project. Additionally it'd be good to know how it's expected to be used in the interim and what the final end-state would look like (etc).

behunin:
Right now with all the changes I have in mind and the with the changes I have seen in the open PRs, I don't really see an easy way to transition from the current version to a stable 1.0 release.

The way I see it is that implementing WebAssembly::{Memory, Table, Global} classes should be done inside a generated class.
```javascript 
class initYourClass  {
    global; // WebAssembly global(s)
    memory; // WebAssembly memory
    table; // WebAssembly table

    instance; 
    module;

    constuctor(path, options) {
      // init class properties with options
      // load wasm from path or use default import.meta.url
    }

    // helper functions
}

```

Users would create a new instance of their class and call the exports.instance.functions, globals, and table from that class instead of initializing a promise.

This is the main reason why I am asking for a dev repo. Let's have a place where we can tear this project apart and make it into the elegant solution that fits all the needs of the community. 

Tomroww I will start a new issue with the plans I have in mind so that we can get feedback from the community on my ideas and/or how they would like initialization and usage to be handled. Then we should have a good idea on how to move forward.

alexcrichton:
FWIW you can already create a dev repo with a fork, and you can already publish `*-dev` crates to crates.io. I'm still not sure whether that's something I'd like to manage myself.

I'm also not sure I feel too great about the idea of "tearing apart this project" and trying to rebuild it from scratch. I fear that sort of mentality means that you're not too willing to learn from past design decisions in this crate and it seeems like your main goal is to just restructure things to suit your needs?

behunin:
> FWIW you can already create a dev repo with a fork, and you can already publish `*-dev` crates to crates.io. I'm still not sure whether that's something I'd like to manage myself.
> 
I understand. As long as we are working together towards a stable 1.0 release here, I am happy.
  
> I'm also not sure I feel too great about the idea of "tearing apart this project" and trying to rebuild it from scratch. I fear that sort of mentality means that you're not too willing to learn from past design decisions in this crate and it seeems like your main goal is to just restructure things to suit your needs?

Please don't let my enthusiasm deter you, I want to make big changes and that is how I felt about the process. I think that a restructuring is in order but would never try to just push for just what I want, which is why I created a new issue to try and gather consensus on how to move forward. By no means do I want to start from scratch or forgo the current design all together.

diceride:
https://github.com/rustwasm/wasm-bindgen/pull/2519

PR #2416, opened 2021-01-07
mod blob_url

exellife:
an idea how to keep Web Workers script in compiled code, so we can avoid fetching script N times.


alexcrichton:
Seems plausible to me! Do you want to hook this up into the example itself?

exellife:
Hi Alex,

I have a full working example with some modifications in source code. I'll
create a new branch and make a pull request, so it will make more sense to
have a working example.
The file that I added is not gonna work by itself. It just gives an idea
how to solve a problem with `Worker('./script.js')`.

PR #2427, opened 2021-01-20
Update WebIDL to use `undefined` instead of `void`

grovesNL:
We updated weedle to use `undefined` instead of `void` in rustwasm/weedle#43, so now the plan is to update all WebIDL too.

The changes here are split into two commits for reviewing. The first commit updates all of the WebIDL and the second commit updates how the tokens are used from weedle.

Overall the changes seem to be fine because the generated bindings don't show any differences.

### DIFF
--- a/crates/webidl/Cargo.toml
+++ b/crates/webidl/Cargo.toml
@@ -21,7 +21,7 @@ proc-macro2 = "1.0"
 quote = '1.0'
 syn = { version = '1.0', features = ['full'] }
 wasm-bindgen-backend = { version = "=0.2.69", path = "../backend" }
-weedle = "0.11"
+weedle = "0.12"
 lazy_static = "1.0.2"
 sourcefile = "0.1"
 structopt = "0.3.9"
--- a/crates/webidl/src/first_pass.rs
+++ b/crates/webidl/src/first_pass.rs
@@ -503,6 +503,13 @@ impl<'src> FirstPass<'src, (&'src str, ApiStability)> for weedle::interface::Int
                 log::warn!("Unsupported WebIDL Setlike interface member: {:?}", self);
                 Ok(())
             }
+            InterfaceMember::AsyncIterable(_iterable) => {
+                log::warn!(
+                    "Unsupported WebIDL async iterable interface member: {:?}",
+                    self
+                );
+                Ok(())
+            }
         }
     }
 }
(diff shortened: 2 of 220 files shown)


crates/webidl/Cargo.toml
```diff
@@ -21,7 +21,7 @@ proc-macro2 = "1.0"
 quote = '1.0'
 syn = { version = '1.0', features = ['full'] }
 wasm-bindgen-backend = { version = "=0.2.69", path = "../backend" }
-weedle = "0.11"
+weedle = { git = "https://github.com/rustwasm/weedle" }
```
grovesNL, line 24: We'll need to use a published version here, so this PR can wait for that

crates/webidl/src/first_pass.rs
```diff
@@ -503,6 +503,10 @@ impl<'src> FirstPass<'src, (&'src str, ApiStability)> for weedle::interface::Int
                 log::warn!("Unsupported WebIDL Setlike interface member: {:?}", self);
                 Ok(())
             }
+            InterfaceMember::AsyncIterable(_iterable) => {
```
grovesNL, line 506: This is unrelated to undefined/void changes, it's just a new variant added between weedle 0.11 and weedle `master` also noticed in https://github.com/rustwasm/wasm-bindgen/pull/2373#issuecomment-762931655

crates/webidl/src/idl_type.rs
```diff
@@ -490,7 +490,7 @@ impl<'a> IdlType<'a> {
             }
 
             IdlType::Any => dst.push_str("any"),
-            IdlType::Void => dst.push_str("void"),
+            IdlType::Undefined => dst.push_str("undefined"),
```
grovesNL, line 493: Note this is technically able to change the function names of generated bindings (i.e. a breaking change), but it doesn't happen with any of the WebIDL in this repository.

So this change seems ok unless there's a path where downstream crates are able to provide their own WebIDL, but I don't think we expose that anywhere.

alexcrichton: Nah seems fine to me!

alexcrichton:
Looks great to me, thanks! I think the only thing this needs now is a weedle publish, right?

grovesNL:
Yeah I think we just need to publish a new version of weedle and use it here, which will also unblock progress in #2373. Created a PR at rustwasm/weedle#44

alexcrichton:
Ok published!

grovesNL:
Thanks! This should be ready now

PR #2438, opened 2021-01-29
Avoid errors if vendor-prefixed APIs do not exist (fix #2437)

### LINKED ISSUE #2437: ReferenceError in Tor browser when AudioContext doesn't exist
### Describe the Bug

When using `AudioContext`, wasm-bindgen generates code like this:
`const lAudioContext = (typeof AudioContext !== 'undefined' ? AudioContext : webkitAudioContext);`
In the Tor browser, this code will error immediately and prevent any Wasm execution, because Web Audio is not enabled in Tor, and neither `AudioContext`/`webkitAudioContext` exist.

Ideally this would not error, `lAudioContext` could be set to `null`/`undefined` and allow the module to run. The Rust code can then catch errors itself when it tries to use `AudioContext`.

I suspect this could apply to all vendor-prefixed APIs.

### Steps to Reproduce

1. Create a Rust library using `AudioContext` (or try the [web-sys WedAudio example](https://rustwasm.github.io/wasm-bindgen/exbuild/webaudio/))
2. Navigate to the sample in the Tor Browser
3. Observe the error, and the module does not run.

### Expected Behavior

The module should still run even if the API/vendor-prefixed API doesn't exist. The Rust code should have a chance to catch the error when the API is actually used.

### Actual Behavior
The following error occurs immediately, and the wasm module is not loaded:
```
Uncaught ReferenceError: webkitAudioContext is not defined
    240 test_web.js:3
    Webpack 4
ruffle_web.js:3:47
    240 test_web.js:3
    Webpack 4
        n
        <anonymous>
        <anonymous>
        <anonymous>
```

Herschel:
Avoid errors in the generated vendor-prefix code if neither the
standard API nor the vendor-prefixed API exists. Instead, fall back
to undefined, which will give the Rust module a chance to run and
handle the error when it tries to use the API.

Fixes #2437.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2095,10 +2095,6 @@ impl<'a> Context<'a> {
                 self.imports_post.push_str(";\n");
 
                 fn switch(dst: &mut String, name: &str, prefix: &str, left: &[String]) {
-                    if left.len() == 0 {
-                        dst.push_str(prefix);
-                        return dst.push_str(name);
-                    }
                     dst.push_str("(typeof ");
                     dst.push_str(prefix);
                     dst.push_str(name);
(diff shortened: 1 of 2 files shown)


alexcrichton:
Seems reasonable to me, thanks! Could you be sure to add a test for this as well?

Herschel:
Done!

PR #2441, opened 2021-02-02
Update browser support caveats

AminArria:
Most of the caveats can be removed:
- Edge 79+ supports TextEncoder and TextDecoder APIs  (https://caniuse.com/textencoder)
- Edge 79+ and Safari 14+ supports BigInt (https://caniuse.com/bigint)

### DIFF
--- a/guide/src/reference/browser-support.md
+++ b/guide/src/reference/browser-support.md
@@ -19,6 +19,45 @@ also like to be aware of it!
   an example of doing this too](../examples/wasm2js.html)). Note
   that at this time no bundler will do this by default, but we'd love to
   document plugins which do this if you are aware of one!
+  
+* **Edge before 79+** - the `TextEncoder` and `TextDecoder` APIs where not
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Instead of deleting the sections though could they perhaps be re-worded? Something along the lines of "browsers before X" need consideration?

AminArria:
@alexcrichton made the changes, let me know if the wording is fine or it needs some more tweaking

alexcrichton:
Thanks!

tash-2s:
I think Safari 14+ isn't still supported. It has `BigInt` but doesn't have `BigUint64Array`.
```
ReferenceError: Can't find variable: BigUint64Array
```
https://caniuse.com/?search=BigUint64Array
https://bugs.webkit.org/show_bug.cgi?id=190800

Firaenix:
Can also confirm BigUint64Array does not work on Safari.

I do believe there is potential for this to be polyfilled however.
We would need to amend the existing u64 polyfill.

PR #2443, opened 2021-02-04
Iterator and related additions

silvanshade:
This PR just adds and refines a few things related to `AsyncIterator`, `Iterator`, and `Generator`.


alexcrichton:
Thanks for the PR! Unfortunately though I think this is a breaking change? Currently there's no plan of when or how to make breaking changes to js-sys.

hvithrafn:
I see. Well, I guess I can just leave this open in case that situation changes. Thanks for letting me know.

Issue #2447, opened 2021-02-05
[weak-refs] UAF in cli-generated JS shims when dealing with by-value receivers


danielhenrymantilla:
### Describe the Bug

When the `--weak-refs` feature is enabled, the JS shim code that `wasmbindgen-cli` generates (`out_dir/libname_bg.js` file) is incorrect and leads to UAFs (_e.g._, double free) when a decorated `[wasm_bindgen]` method uses a by-value receiver.

The very short technical description of the bug is that in the JS glue wrapping such a method is missing a call to `ThatClassFinalization.unregister(this);`.

### Steps to Reproduce

 1. ```rust
    //! src/lib.rs
    use ::wasm_bindgen::prelude::*;

    #[wasm_bindgen]
    pub struct Foo { /* … */ }

    #[wasm_bindgen]
    impl Foo {
        #[wasm_bindgen(constructor)]
        pub fn new(/* … */) -> Self {
            Self { /* … */ }
        }

        pub fn identity(self) -> Self {
            self
        }
    }
    ```

 1. ```toml
    # Cargo.toml
    [lib]
    crate-type = ["cdylib"]

    [package]
    name = "example"
    version = "0.0.0"
    edition = "2018"

    [dependencies]
    wasm-bindgen = "0.2.70"
    ```

 1. ```shell
    cargo build --target wasm32-unknown-unknown
    wasm-bindgen --out-dir out_dir --out-refs ./target/wasm32-unknown-unknown/debug/example.wasm
    ```

    The generated `out_dir/example_bg.js` will then contain, among other things:

    ```js
    const FooFinalization = new FinalizationRegistry(ptr => wasm.__wbg_foo_free(ptr));
    /**
    */
    export class Foo {
        static __wrap(ptr) {
            const obj = Object.create(Foo.prototype);
            obj.ptr = ptr;
            FooFinalization.register(obj, obj.ptr, obj);
            return obj;
        }
    
        free() {
            const ptr = this.ptr;
            this.ptr = 0;
            FooFinalization.unregister(this);
            wasm.__wbg_foo_free(ptr);
        }
        /**
        */
        constructor() {
            var ret = wasm.foo_new();
            return Foo.__wrap(ret);
        }
        /**
        * @returns {Foo}
        */
        identity() {
            var ptr = this.ptr;
            this.ptr = 0;
            var ret = wasm.foo_identity(ptr);
            return Foo.__wrap(ret);
        }
    }
    ```

[...]

PR #2448, opened 2021-02-05
[cli] Fix the UAF with by-value receivers and `--weak-refs`

danielhenrymantilla:
Fixes https://github.com/rustwasm/wasm-bindgen/issues/2447


alexcrichton:
Oh dear, thanks for the report and the quick fix though!

PR #2454, opened 2021-02-12
Add a setter for the list property of Select

vthg2themax:
Let me know if there's anything else you need.


vthg2themax:
Sorry, it appears there is no setter available for list. Thanks anyway!

PR #2457, opened 2021-02-18
Update the Windows Documentation

vthg2themax:
I am currently trying to get my webIDL changes done, but cannot get GitAhead to work with my 2FA, so I am reverting to windows to the Github Desktop because they don't have a Linux version.

### DIFF
--- a/guide/src/contributing/web-sys/supporting-more-web-apis.md
+++ b/guide/src/contributing/web-sys/supporting-more-web-apis.md
@@ -35,5 +35,8 @@
    cd crates/web-sys
    cargo run --release --package wasm-bindgen-webidl webidls src/features
    ```
+   
+   If you are on Windows, you need to use the following command:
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I don't think this is a Windows-only thing, though. Perhaps this is a bug in Cargo?

vthg2themax:
Could I put a disclaimer that if you encounter an issue (I used Linux first with no problems), that you could try this alternate command?

alexcrichton:
Er sorry, my point is that something weird happened here. If this doesn't work for you that may mean this was an accidental breaking change in Cargo or some other Rust tooling. The unix/windows behavior should not be different here. If this needs a fix there's nothing Windows-specific about the fix, it just needs to be fixed. Ideally there'd be a little bit of digging to figure out what changed as well to see if this is a bug in Cargo.

PR #2458, opened 2021-02-18
Update scrollHeight and scrollTop

vthg2themax:


vthg2themax:
I did this all in one go. Can you tell me what might be wrong?

alexcrichton:
No worries! I think this is a formatting "error" due to a more recent version of rustfmt, but otherwise looks great. Thanks!

vthg2themax:
Thank you very much for approving this! Did the other version of this pull request have the same problem then? I just want to make sure that I have good code outputs.

alexcrichton:
Probably? I didn't check.

PR #2469, opened 2021-02-26
Make maybe_memory truly optional

### LINKED ISSUE #2133: Parameter `maybe_memory` should be optional
### Describe the Bug

I have noticed that when the compilation target is `web` that the `maybe_parameter` of the `initi(input, maybe_memory)` is required. The name implies that it might be missing and especially for the first call to `init` it's most likely missing if `input` is a string and the module is fetched. Only subsequent calls might have memory. Maybe we change the type of this to also accept `null` or `undefined` or mark it as optional.

I think it would make sense to have a sanity check in `load` that, if `module` is not a function nor a `Response` object, meaning it's a WebAssembly module, then we add a check to ensure that `maybe_memory` is **not** `undefined`.

Alternatively, we can pass `null!` for `memory_memory` but to me this feels hacky.

### Steps to Reproduce

1. Compile Rust to `wasm32-unknown-unknown` with target being `web`

---

If there is a solution to this and a clear path, I am volunteering to work on a PR to get this resolved.

RReverser:
As the name implies, it's already optional, but wasn't marked as such in TS.

We could put some more complicated / stricter types here depending on type of the first argument, but at least this fixes the issue for TS consumers.

Also fixes an issue where `maybe_memory` wouldn't be initialised if the `input` (module) is given as an `ArrayBuffer`.

Fixes #2133

### DIFF
--- a/examples/raytrace-parallel/rust-toolchain
+++ b/examples/raytrace-parallel/rust-toolchain
@@ -1,1 +1,0 @@
-nightly
--- a/examples/raytrace-parallel/rust-toolchain
+++ b/examples/raytrace-parallel/rust-toolchain
@@ -1,0 +1,1 @@
+nightly
--- a/tests/wasm/variadic.rs
+++ b/tests/wasm/variadic.rs
@@ -37,9 +37,9 @@ extern "C" {
     //                          rest: Vec<String>) -> String;
     #[wasm_bindgen(variadic)]
     fn variadic_sum_rest_vec(first: u8, second: u8, rest: Vec<u8>) -> u8;
-//#[wasm_bindgen(variadic)]
-//fn variadic_compare_pairs(first: JsValue, second: JsValue, rest: &[JsValue]);
-//TODO imported type
+    //#[wasm_bindgen(variadic)]
+    //fn variadic_compare_pairs(first: JsValue, second: JsValue, rest: &[JsValue]);
+    //TODO imported type
 }
 
 // ints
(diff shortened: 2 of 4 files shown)


RReverser:
Just noticed the issue with passing in ArrayBuffer too; please don't merge yet, fixing it as well.

RReverser:
Done. /cc @alexcrichton @Pauan

tests/wasm/variadic.rs
```diff
@@ -37,9 +37,9 @@ extern "C" {
     //                          rest: Vec<String>) -> String;
     #[wasm_bindgen(variadic)]
     fn variadic_sum_rest_vec(first: u8, second: u8, rest: Vec<u8>) -> u8;
-    //#[wasm_bindgen(variadic)]
-    //fn variadic_compare_pairs(first: JsValue, second: JsValue, rest: &[JsValue]);
-    //TODO imported type
+//#[wasm_bindgen(variadic)]
```
RReverser, line 40: Looks like this was also touched by `cargo fmt`. Should I revert?

RReverser: But at the same time `cargo fmt` on CI complains about this new formatting :/

I'll try to update my local copy.

examples/raytrace-parallel/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
alexcrichton, line 1: Could this be removed? I'd prefer that nightly is explicitly used to build this example

alexcrichton (commented):

examples/raytrace-parallel/rust-toolchain
RReverser: It's mostly a convenience - I found it a bit annoying to always remember to run `rustup run nightly ./build.sh` for build to work in this folder, but certainly can remove if you're strongly opposed.

RReverser: Reverted.

PR #2476, opened 2021-03-02
raytrace-parallel: use rust nightly explicitly

mattgibb:
Hey Rust Wasm peeps,

I cannot get over how awesome this project is. I have learned so much rust and wasm going through all the documentation, and would never have been able to get close to using webassembly without this project. Thanks so much for your efforts!

Ok so this is just a cheeky little PR to get my feet wet. I _think_ it makes sense, since at the moment I believe you can't use the `-Z` flag in stable Rust.

What do you think?

### DIFF
--- a/examples/raytrace-parallel/build.sh
+++ b/examples/raytrace-parallel/build.sh
@@ -13,7 +13,7 @@ set -ex
 #   shared memory, passive segments, etc.
 
 RUSTFLAGS='-C target-feature=+atomics,+bulk-memory' \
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR, but I personally sometimes use a specific nightly with this script, and unfortunately with hardcoding that prevents doing so :(

mattgibb:
Ah roger that, thanks for the feedback @alexcrichton looks like I need to have a tinker with how cargo works.

mattgibb:
Closing as the change clobbers specific nightly builds

PR #2482, opened 2021-03-06

grovesNL:
Update to the latest WebGPU WebIDL as of March 5.

There is one manual change to the WebIDL to accept `GPUCode` (as `typedef (USVString or Uint32Array) GPUCode;`) instead of `USVString` for shaders to allow both SPIR-V and WGSL (as mentioned in #2267).


alexcrichton:
👍

PR #2490, opened 2021-03-14
Fix typo in web-sys contributing docs

eminence:


alexcrichton:
Thanks for the report. I don't think this is a typo but rather a bug in Cargo -- https://github.com/rust-lang/cargo/issues/9267 -- but it seems fine to update this for now until that's fixed.

PR #2492, opened 2021-03-15
Add TypedArray::copy_from

MattiasBuelens:
This introduces `TypedArray::copy_from(src)`, as the dual of `TypedArray::copy_to(dst)`.

While this *can* already be achieved today with `TypedArray::set()`, it requires either unsafe code or an additional allocation:
```rust
let x = [1, 2, 3];
let mut array = Int32Array::new(&3.into());

unsafe { array.set(Int32Array::view(&x), 0); } // view() is unsafe
// or
array.set(Int32Array::from(&x), 0); // from() allocates a new Uint8Array to copy the contents of x
```
The new helper method provides a safe interface to perform this copy:
```rust
let x = [1, 2, 3];
let mut array = Int32Array::new(&3.into());
array.copy_from(&x);
```
Internally, it uses the unsafe implementation listed above. We know that this is safe, because:
* the `TypedArray` view on the Rust slice lives only for the duration of the `copy_from()` call
* `set()` does not grow the WebAssembly linear memory, so it cannot invalidate the created view

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -4997,7 +4997,7 @@ macro_rules! arrays {
             ///
             /// This function will panic if this typed array's length is
             /// different than the length of the provided `src` array.
-            pub fn copy_from(&mut self, src: &[$ty]) {
+            pub fn copy_from(&self, src: &[$ty]) {
                 assert_eq!(self.length() as usize, src.len());
                 unsafe { self.set(&$name::view(src), 0) }
             }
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -4987,6 +4987,21 @@ macro_rules! arrays {
...
+            /// This function will efficiently copy the memory from within
+            /// the wasm module's own linear memory to this typed array.
+            ///
+            /// # Panics
+            ///
+            /// This function will panic if this typed array's length is
+            /// different than the length of the provided `src` array.
+            pub fn copy_from(&mut self, src: &[$ty]) {
+                assert_eq!(self.length() as usize, src.len());
+                unsafe { self.set(&$name::view(src), 0) }
```
MattiasBuelens, line 5002: I suppose this needs a comment explaining with this `unsafe` block is actually safe? Something like:
```rust
// This is safe because the `set` function copies from its TypedArray argument
```
Suggestions are welcome! 😁

alexcrichton:
Looks good to me, thanks! I think that the method should probably take `&self` instead of `&mut self` as well since it's all JS, but otherwise seems good to go.

Issue #2496, opened 2021-03-17
`js_sys::Function::new_with_args` should catch syntax errors


s9ferech:
### Motivation

The functions `new_with_args` and `new_no_args` on `js_sys::Function` currently panic in case of syntax errors. That makes it impossible to handle such errors and hides the fact that those two constructors can fail.

### Proposed Solution

Both functions should catch errors and return a `Result`, as in
```
#[wasm_bindgen(constructor, catch)]
pub fn new_with_args(args: &str, body: &str) -> Result<Function, JsValue>;
```

### Alternatives

Since the proposed solution would be a breaking change, it might be better to define new functions with the desired behaviour instead, for example called `try_new_with_args`.

But this would make the API more complex and not solve the problem that the type of `new_with_args` is slightly misleading.

PR #2497, opened 2021-03-17
Catch syntax errors in constructors of `js_sys::Function`

s9ferech:
This is a potential solution to issue #2496.

The constructors `new_with_args` and `new_no_args` will catch exceptions that are caused by syntax errors or similar problem.

This change would require to add an explicit `unwrap` in all places where the constructors are already used.


alexcrichton:
Thanks! Since this is a breaking change it can't land yet, but I'll tag it.

wdanilo:
@alexcrichton are there any plans to merge it one day?

alexcrichton:
As mentioned this is a breaking change, and no there is no plans of when a breaking change will land. Adding new constructors is fine.

Issue #2489, opened 2021-03-11
web_sys::InputEvent missing important fields


lunabunn:
### Describe the Bug

`web_sys::InputEvent` is missing multiple fields/getters (`data`, `dataTransfer`, `inputType`).

### Steps to Reproduce

See [the docs](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.InputEvent.html) for `web_sys::InputEvent`, then see the [MDN docs for InputEvent](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent).

### Expected Behavior

I expected there to be getters on `web_sys::InputEvent` for the fields mentioned above.

### Actual Behavior

There were no such getters.

### Additional Context

None

PR #2499, opened 2021-03-17
Add missing APIs for InputEvent

lunabunn:
Fix https://github.com/rustwasm/wasm-bindgen/issues/2489.


lunabunn:
I got the WebIDL definitions from https://hg.mozilla.org/mozilla-central/file/tip/dom/webidl/InputEvent.webidl, is that the right thing to do?

alexcrichton:
That's what we've done historically, yeah, thanks!

PR #2506, opened 2021-03-24
Fix js_sys::Intl::DateTimeFormat::format

glennsl:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format) 'Intl.DateTimeFormat.format' takes a `Date` and returns a `String`.

I'm not sure why the current binding was made as it is, but the fact that there's a test for it that checks that it does the wrong thing makes it look deliberate. This was originally added in #691, but I see no rationale for it there either. Perhaps there's some obscure historical reason for it. I do see several similar mistakes in the `Intl` namespace however, but figured I'd start with something small and simple to make sure I'm not misunderstanding something.

Also, the value returned apparently isn't an instance of `JsString`, but `is_string()` still returns `true` so I'm hoping that's fine.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -4488,8 +4488,8 @@ pub mod Intl {
         /// Intl.DateTimeFormat object.
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/format)
-        #[wasm_bindgen(method, getter, js_class = "Intl.DateTimeFormat")]
-        pub fn format(this: &DateTimeFormat) -> Function;
+        #[wasm_bindgen(method, js_class = "Intl.DateTimeFormat")]
+        pub fn format(this: &DateTimeFormat, date: &Date) -> JsString;
 
         /// The `Intl.DateTimeFormat.prototype.formatToParts()` method allows locale-aware
         /// formatting of strings produced by DateTimeFormat formatters.
--- a/crates/js-sys/tests/wasm/Intl.rs
+++ b/crates/js-sys/tests/wasm/Intl.rs
@@ -55,7 +55,7 @@ fn date_time_format() {
     let epoch = Date::new(&JsValue::from(0));
 
     let c = Intl::DateTimeFormat::new(&locales, &opts);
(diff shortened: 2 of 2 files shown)


glennsl:
According to the [the specification](https://tc39.es/ecma402/#sec-intl.datetimeformat.prototype.format) it actually is a getter that returns a [DateTime format function](https://tc39.es/ecma402/#sec-datetime-format-functions). That happens to work out well in JavaScript where it's indistinguishable from a method. It's not so useful as a rust binding though.

Perhaps an argument can be made for a `format_function` binding in addition to this, but I don't really see it. I reckon you could just as well make a closure of your own to call this.

alexcrichton:
Thanks for the PR! Unfortunately this is a breaking change so it can't land just yet, but I'll tag it as such for now.

glennsl:
No problem. Would you like me to do the same with the other similar APIs in the `Intl` namespace? As they'd also be breaking changes it would be good to get them sooner rather than later.

Also, is there any kind of ETA on a breaking release? Judging by the history it seems it might take a while. And I do need these APIs, so perhaps it would be better to make a local copy.

alexcrichton:
If you'd like feel free!

Unfortunately there is no eta on a new breaking release

glennsl:
Done.

glennsl:
Would it also be good to make the options objects strongly typed? I've tried to mimic the way it's done in `web-sys`, but I'm getting ``failed to resolve: could not find `js_sys` in `{{root}}` `` on `wasm_bindgen` attributes attached to `enum`s. Perhaps it wasn't designed to be used from within `js-sys` itself. And I'd rather not waste a lot of time looking into this if it wouldn't be all that beneficial anyway.

alexcrichton:
I think it's probably fine to leave as-is for now, but thanks!

MendyBerger:
Since it doesn't look like this will be fixed any time soon.
Here's a quick snippit the get the current version to work.
```rust
let locales = js_sys::Array::of1(&JsValue::from("en-US"));
let options = js_sys::Object::new();
let intl = js_sys::Intl::NumberFormat::new(&locales, &options);
let num = JsValue::from_f64(140.90);
intl.format().call1(&intl, &num).unwrap().as_string().unwrap()
```

Issue #2343, opened 2020-11-04
Incompatibility when using web_sys along with asyncWebAssembly in webpack 5


Herschel:
### Describe the Bug

When using webpack 5, enabling `asyncWebAssembly: true` in the webpack config will cause errors when trying to load the mdoule:

```
ReferenceError: Cannot access '__wbindgen_string_new' before initialization
```

The Rust code in question:
```
#[wasm_bindgen]
pub fn foo(n: &web_sys::HtmlElement) {
    web_sys::console::log_1(&n.tag_name().into());
}
```
Commenting out this code, or replacing it with a `fn foo(n: i32) { // ... }` allows the code to function. Using `syncWebAssembly: true` also works as it did in webpack 4.

#2110 seems to indicate that `asyncWebAssembly` should work. 

### Steps to Reproduce

Minimal example repo:
1. `git clone https://github.com/Herschel/webpack-test`
2. `npm install`
3. `npm run start`
4. Navigate to the page in the browser and view the JS console.

### Expected Behavior

The code should run successfully.

### Actual Behavior

The following error is thrown:
```
ReferenceError: Cannot access '__wbindgen_string_new' before initialization
```

### Additional Context
```
rustc 1.49.0-nightly (ffa2e7ae8 2020-10-24)
wasm-bindgen = "0.2.68"
web-sys = "0.3.45"
npm@6.14.4
wasm-pack 0.9.1
"webpack": "^5.4.0"
```

alexcrichton:
Thanks for the report! I'm not entirely sure what changed in Webpack 5, but it at least looks like we need to account for something! I'd need to dig into how things changed and also any recent updates on the ESM proposal, if any. I don't have a ton of time for this but I'll try to make some time in the near future.

JohnForster:
FYI, this also happens with wasm-bindgen's built in JsValue.

hansihe:
I am having a very similar issue, I believe the same thing.

When the single `my_rust.js`, webpack wraps it in something like the following boilerplate:

```
((module, __webpack_exports__, __webpack_require__) => {
    module.exports = async () => {
        [...]
        __webpack_require__.d(__webpack_exports__, {
            [...],
            "__wbindgen_throw": () => __wbindgen_throw,
        });
        [...]

        var _my_rust_bg_wasm__WEBPACK_IMPORTED_MODULE = __webpack_require__(119);
[...]

PR #2512, opened 2021-03-29
Work around for building with webpack5

nokotan:
This pr will be solution to #2343.
In this PR, binding functions are declared with "function" instead of "const".

[Hansihe showed](https://github.com/rustwasm/wasm-bindgen/issues/2343#issuecomment-766363602):

```js
((module, __webpack_exports__, __webpack_require__) => {
    module.exports = async () => {
        [...]
        __webpack_require__.d(__webpack_exports__, {
            [...],
            "__wbindgen_throw": () => __wbindgen_throw,
        });
        [...]

        var _my_rust_bg_wasm__WEBPACK_IMPORTED_MODULE = __webpack_require__(119);
        _my_rust_bg_wasm__WEBPACK_IMPORTED_MODULE = await Promise.resolve(_my_rust_bg_wasm__WEBPACK_IMPORTED_MODULE);

        [...]

        const __wbindgen_throw = function([..]) { [..] };

        [...]
        return __webpack_exports__;
    }();
})
```

Webpack tries to export "__wbindgen_throw", but "__wbindgen_throw" is undefined because "__wbindgen_throw" is declared below with "const" and undefined will be exported.

```js
        __webpack_require__.d(__webpack_exports__, {
            [...],
            "__wbindgen_throw": () => __wbindgen_throw,
        });
```

Now webpack start to load .wasm, but imported object "__wbindgen_throw" is undefined, which leads silent promise rejection.

```js
        var _my_rust_bg_wasm__WEBPACK_IMPORTED_MODULE = __webpack_require__(119);
```

"__wbindgen_throw" will be hoisted and it is not "undefined" when webpack exports "__wbindgen_throw".

```js
        __webpack_require__.d(__webpack_exports__, {
            [...],
            "__wbindgen_throw": () => __wbindgen_throw,
        });
        [...]
        function __wbindgen_throw([..]) { [..] };
```

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1765,7 +1765,7 @@ impl<'a> Context<'a> {
                     "\
                     function handleError(f, args) {{
                         try {{
-                            return f.apply(this, arguments);
+                            return f.apply(this, args);
                         }} catch (e) {{
                             wasm.{}(addHeapObject(e));
                         }}
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -419,11 +419,19 @@ impl<'a> Context<'a> {
                 for (id, js) in crate::sorted_iter(&self.wasm_import_definitions) {
                     let import = self.module.imports.get_mut(*id);
                     import.module = format!("./{}_bg.js", module_name);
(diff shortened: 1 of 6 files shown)


alexcrichton:
Sorry for the delay in getting to this but thanks for the PR!

PR #2519, opened 2021-04-07
cli-support: Make default_module_path optional

diceride:
Adds a new CLI flag `--omit-default-module-path`, to control the `default_module_path` generated JavaScript code.

The `default_module_path` generated JavaScript code is **enabled** by default, users need to explicitly disable it by setting the `--omit-default-module-path` CLI flag

```sh
$ wasm-bindgen --omit-default-module-path
```

Documentation

```txt
$ wasm-bindgen --help

Options:

...

--omit-default-module-path       Don't add WebAssembly fallback imports in generated JavaScript
```

Affected JavaScript code

`$ wasm-bindgen --target web`

```js
async function init(input) {
        input = import.meta.url.replace(/\.js$/, '_bg.wasm');
    }
    ...
```

`$ wasm-bindgen --target no-modules`

```js
async function init(input) {
        let src;
            src = location.href;
        } else {
            src = document.currentScript.src;
        }
        input = src.replace(/\\.js$/, '_bg.wasm');
    }
    ...
```

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -632,15 +632,16 @@ impl<'a> Context<'a> {
             }
         }
 
-        let default_module_path = match self.config.mode {
-            OutputMode::Web => format!(
-                "\
+        let default_module_path = if !self.config.omit_default_module_path {
+            match self.config.mode {
+                OutputMode::Web => format!(
+                    "\
                     if (typeof input === 'undefined') {{
                         input = new URL('{stem}_bg.wasm', import.meta.url);
                     }}",
-                stem = self.config.stem()?
-            ),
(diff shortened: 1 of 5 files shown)


diceride:
@alexcrichton

alexcrichton:
Thanks for this! I think though due to docopt the default is still `false`? Can a test be added for this as well?

diceride:
I changed the CLI option to the flag `--omit-default-module-path` and added tests to the CLI test suite.

diceride:
I changed the CLI option to the flag `--omit-default-module-path` and added tests to the CLI test suite.

diceride:
The Windows tests are flaky, I had to force-push it multiple times.

alexcrichton:
Ok thanks again for the PR!

Issue #2502, opened 2021-03-18
Target no-modules failed to find its .wasm file automatically


yume-chan:
### Describe the Bug

When building with `--target no-modules`, the output js file should be able to find its `.wasm` file.

### Steps to Reproduce

1. Clone the `without-a-bundler-no-modules` example and build
2. Modify `index.html` as:
    ```diff
    - await wasm_bindgen('./pkg/without_a_bundler_no_modules_bg.wasm');
    + await wasm_bindgen();
    ```
3. Open it in a browser (I'm using Microsoft Edge 89.0.774.54)

### Expected Behavior

It loads the `.wasm` file and runs.

### Actual Behavior

```text
Uncaught (in promise) CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 68 74 6d @+0
```

### Additional Context

I treat it as a bug because the generated code did try to detect the correct url for its `.wasm` file, but it failed.

Related code:

```js
    async function init(input) {
        if (typeof input === 'undefined') {
            let src;
            if (typeof document === 'undefined') {
                src = location.href;
            } else {
                src = document.currentScript.src;
            }
            input = src.replace(/\.js$/, '_bg.wasm');
        }
//...
```

The problem is that, `document.currentScript` refers to the current processing script, so as most of time `init` was been called from another script (and in this case, a script tag without `src`), this code fails.

The correct method is caching `document.currentScript.src` when it's initializing, for example:

```diff
+     const cachedSrc = document.currentScript.src;
      async function init(input) {
          if (typeof input === 'undefined') {
              let src;
              if (typeof document === 'undefined') {
                  src = location.href;
              } else {
-                 src = document.currentScript.src;
+                 src = cachedSrc;
            }
            input = src.replace(/\.js$/, '_bg.wasm');
        }
```

I verified it works after the modification:

![image](https://user-images.githubusercontent.com/1330321/111612544-895b8080-8818-11eb-85da-29c558deb001.png)

CrimsonCodes0:
We can't literally cache it like you had shown, because the code needs to check for the existence of the document object.

What we can do instead is cache the entire expression.

If document doesn't exist, and is created later, then I doubt that we'd want that document object, and the location won't change without reloading the scripts, so the entire thing can be a constant.

```js
const cached_source = ("document" in globalThis
    ? document.currentScript.src
    : location.href
).replace(/\.js$/, '_bg.wasm');

async function init(input = cached_source) {
[...]

PR #2520, opened 2021-04-10
Cache document.currentScript for non-ESM output

CrimsonCodes0:
Fixes issue https://github.com/rustwasm/wasm-bindgen/issues/2502.

Currently, when using non-ESM module output (i.e. scripts output), the generated JavaScript accesses the `Document#currentScript` getter at runtime, which will always have a different (incorrect) value by the time that the user runs their code to load their Wasm.

With this change, the default source for ESM and web script output is cached in an immutable variable, and accessed when the user doesn't pass an argument.

Other output targets should remain untouched.

As an added side-effect, the default source is always constructed, this should add a small amount of extra overhead to loading the glue JS, but if the user loads the Wasm multiple times (for some reason?), it is only constructed once, which should improve performance.


CrimsonCodes0:
Oh wait, Deno and Node don't get default values, I'll just use the constant variable.

CrimsonCodes0:
Oh wait, Deno and Node don't get default values, I'll just use the constant variable.

alexcrichton:
Thanks! Could you leave a comment in the source for why it's done this way?

CrimsonCodes0:
@alexcrichton Two questions: by "source," are you referring to the JS or Rust? And what about the way that it's done would you like me to leave a comment about?

alexcrichton:
This comment looks fine, thanks for that! I think this needs a rebase now after a previous merge, want to do that and I can merge?

CrimsonCodes0:
Sure, I'll rebase and deal with the conflicts soon.

crates/cli-support/src/js/mod.rs
```diff
@@ -640,35 +640,43 @@ impl<'a> Context<'a> {
...
+                EMPTY
             }
-        } else {
-            String::from("")
         };
 
         let ts = self.ts_for_init_fn(
             has_memory,
-            !self.config.omit_default_module_path && !default_module_path.is_empty(),
+            !self.config.omit_default_module_path && !default_module_path_creator.is_empty(),
```
CrimsonCodes0, line 679: If I'm following this logic correctly, the entire expression could be shortened to the following:
```rust
!default_module_path_creator.is_empty()
```

cc @diceride

If someone omits the default path, and chooses, say, Node.js, there still won't be a default path, therefore we can just check if there is a default path via `.is_empty()`.

CrimsonCodes0:
@alexcrichton It seems like a bunch of runtime tests are now failing after rebasing from the recent merges.

Could you take a look into it and help me debug this?

alexcrichton:
I think you need to run the tests locally with `BLESS=1` to regenerate the expected output since I believe it's changing here

alexcrichton:
(or the test expectations of what the output is just need updating)

CrimsonCodes0:
More recently, I've been responding to people and writing code on mobile, I don't have a PC to download rustc & friends on at the moment, so it would be some time before I could attempt to run those tests locally.

If someone else could run them, we should be able to fix the test failure sooner

PR #2528, opened 2021-04-15
Fix typo wasm_bidngen to wasm_bindgen

bluk:
Fixes a couple of macro error messages which have `wasm_bidngen` instead of `wasm_bindgen`.


alexcrichton:
Thanks!

PR #2546, opened 2021-05-04
Use a hash as the SCHEMA_VERSION

### LINKED ISSUE #2544: Relax same schema version requirement between wasm-bindgen-cli and wasm-bindgen
## Background / Motivation

After getting lots of awesome mileage out of it, I'm looking to move away from wasm-pack (for reasons that I don't think are important for this issue).

The main problem that wasm-pack solves for me that would be difficult to make up for after moving away is the handling of version conformance between wasm-bindgen-cli (WB-CLI) and wasm-bindgen (WB).

As in, if WB in your Cargo.toml is 0.2.57 and your WB-CLI binary is 0.2.58, wasm-pack will download a different WB-CLI binary at 0.2.57 and use that instead. Nifty.

## Current Status

The reason that you're forced to use the same version is because, historically, the schema has been unstable.

Forcing people to use the same version guaranteed that WB and WB-CLI would be using the same data format.

An alternative would be to simply maintain a schema version by hand, however this is prone to human error. Maintainers could forget to bump the schema version. So, I'm all for something human-error proof.

## Potential Better Approach

Rather than use the Cargo.toml version as the schema version, it would be nice if the schema version was tied to the actual schema.

This could be accomplished with a build script.

1. Move all of the relevant schema stuff into one or a few files. Isolated in this way so that when they change we know it's a schema change.

2. Build script hashes these files and writes the hash to the OUT_DIR.

3. [`pub const SCHEMA_VERSION: &str = include_str!(concat!(env!("OUT_DIR"), "/schema-hash"));`](https://github.com/rustwasm/wasm-bindgen/blob/87663c6d2a442d98b3d8ea6242f20c5c21fc0174/crates/shared/src/lib.rs#L5)

## End State

By moving towards using a hash of what can actually change instead of an unrelated version, there would be much less churn in the WB and WB-CLI compatibility story.

This would make it feasible to simply depend on WB-CLI directly, and just upgrade the binary when you need to.

For example, as of May 2021, the last time that the [shared crate](https://github.com/rustwasm/wasm-bindgen/tree/87663c6d2a442d98b3d8ea6242f20c5c21fc0174/crates/shared) changed was November 2020. If the trend is that this will continue to change less and less frequently, I'm very fine with occasionally needing to upgrade my WB-CLI binary.

## Questions

- Rather than look for everything that would need to be hashed.. I'd rather just ask someone who already knows. Is it really just some of the stuff in [shared's lib.rs](https://github.com/rustwasm/wasm-bindgen/blob/87663c6d2a442d98b3d8ea6242f20c5c21fc0174/crates/shared/src/lib.rs), or is there much more to it?

- If it's just that file or just a few files, this seems like a no brainer to me. However, I might be missing or overlooking something. Please let me know if I am?

---

Thanks!

chinedufn:
Closes #2544

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -1457,15 +1457,16 @@ fn extract_programs<'a>(
                     "
 
 it looks like the Rust project used to create this wasm file was linked against
-a different version of wasm-bindgen than this binary:
+version of wasm-bindgen that uses a different bindgen format than this binary:
 
-  rust wasm file: {}
-     this binary: {}
+  rust wasm file schema version: {}
+     this binary schema version: {}
 
-Currently the bindgen format is unstable enough that these two version must
-exactly match, so it's required that these two version are kept in sync by
-either updating the wasm-bindgen dependency or this binary. You should be able
-to update the wasm-bindgen dependency with:
+Currently the bindgen format is unstable enough that these two schema versions
+must exactly match. You can accomplish this by either updating the wasm-bindgen
+dependency or this binary.
+
+You should be able to update the wasm-bindgen dependency with:
 
     cargo update -p wasm-bindgen
 
--- a/guide/src/reference/cli.md
+++ b/guide/src/reference/cli.md
@@ -4,34 +4,11 @@ The `wasm-bindgen` command line tool has a number of options available to it to
 tweak the JavaScript that is generated. The most up-to-date set of flags can
 always be listed via `wasm-bindgen --help`.
 
-> Note: usually, one should use a [`wasm-pack`-based workflow][wasm-pack] rather
-> than running the `wasm-bindgen` command line tool by hand.
-
-[wasm-pack]: https://github.com/rustwasm/wasm-pack
-
 ## Installation
 
-The recommend way to install the `wasm-bindgen` command line tool is with the
-`wasm-pack` installer described
-[here](https://rustwasm.github.io/wasm-pack/installer/). After installing
-`wasm-pack`, you are ready to build project invoking `wasm-pack build`.
-This command installs apropriate version of the `wasm-bindgen` command-line
-tool. The version of `wasm-bindgen` installed by `wasm-pack` is not available
- to be used directly via command line.
-
-It is not recommended to install `wasm-bindgen-cli` as its version must match
-_exactly_ the version of `wasm-bindgen` that is specified in the project's
-cargo.lock file. Using `wasm-pack` for building simplifies the build process
-as `wasm-pack` ensures that the proper version of `wasm-bindgen` command-line
-tool is used. That means that `wasm-pack` may install many different versions
-of `wasm-bindgen`, but during the build `wasm-pack` will always make sure to
-use the correct one.
-
-Note: if, for any reason, you decide to use wasm-bindgen directly (this is
-not recommended!) you will have to manually take care of using exactly the
-same version of wasm-bindgen command-line tool (wasm-bindgen-cli) that
-matches the version of wasm-bingden in cargo.lock.
-
+```
+cargo install -f wasm-bindgen-cli
+```
 
 ## Usage
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -915,8 +915,7 @@ Released 2019-04-10.
 
 ### Added
 
-* Initial support for transitive NPM dependencies has been added, although
-  support has not fully landed in `wasm-pack` yet so it's not 100% integrated.
+* Initial support for transitive NPM dependencies has been added.
   [#1305](https://github.com/rustwasm/wasm-bindgen/pull/1305)
 
 * The `constructor` property of `Object` is now bound in `js-sys`.
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -20,7 +20,8 @@ $ cp ./benchmarks /some/other/directory
 Next, `cd` into that directory and execute:
 
 ```
-$ wasm-pack build --target web
+$ cargo build --release --target wasm32-unknown-unknown
+$ wasm-bindgen --target web ./target/wasm32-unknown-unknown/release/crate.wasm
 ```
 
 Next, use your favorite static file server to host the current directory. For
--- a/crates/shared/build.rs
+++ b/crates/shared/build.rs
@@ -1,6 +1,11 @@
+use std::collections::hash_map::DefaultHasher;
+use std::hash::Hasher;
+use std::path::PathBuf;
 use std::process::Command;
 
 fn main() {
+    set_schema_version_env_var();
+
     let rev = Command::new("git")
         .arg("rev-parse")
         .arg("HEAD")
@@ -13,4 +18,14 @@ fn main() {
             println!("cargo:rustc-env=WBG_VERSION={}", &rev[..9]);
         }
     }
+}
+
+fn set_schema_version_env_var() {
+    let schema_file = PathBuf::from(concat!(env!("CARGO_MANIFEST_DIR"), "/src/lib.rs"));
+    let schema_file = std::fs::read(schema_file).unwrap();
+
+    let mut hasher = DefaultHasher::new();
+    hasher.write(&schema_file);
+
+    println!("cargo:rustc-env=SCHEMA_FILE_HASH={}", hasher.finish());
 }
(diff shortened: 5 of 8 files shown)


guide/src/reference/cli.md
```diff
@@ -15,23 +15,23 @@ The recommend way to install the `wasm-bindgen` command line tool is with the
...
-This command installs apropriate version of the `wasm-bindgen` command-line
+This command installs a compatible version of the `wasm-bindgen` command-line
 tool. The version of `wasm-bindgen` installed by `wasm-pack` is not available
- to be used directly via command line.
+to be used directly via command line.
 
-It is not recommended to install `wasm-bindgen-cli` as its version must match
-_exactly_ the version of `wasm-bindgen` that is specified in the project's
-cargo.lock file. Using `wasm-pack` for building simplifies the build process
+It is not recommended to install `wasm-bindgen-cli` as its internal schema
```
chinedufn, line 22: To be fair, this is much less of an issue now so I think this warning can be lifted especially since the schema hasn't changed in over 6 months. 

But I kept it intact for now in case you still want this here.

alexcrichton: Actually yeah would you be ok rewording these docs indicating that using the CLI directly is recommended nowadays? Given the current state of things I think it'd be fine to either remove references to `wasm-pack` or mention it only in a subsection below perhaps too

crates/cli-support/src/wit/mod.rs
```diff
@@ -1457,15 +1457,16 @@ fn extract_programs<'a>(
                     "
 
 it looks like the Rust project used to create this wasm file was linked against
-a different version of wasm-bindgen than this binary:
+version of wasm-bindgen that uses a different bindgen format than this binary:
 
   rust wasm file: {}
      this binary: {}
```
alexcrichton, line 1463: FWIW This was intended to be a human readable error message where the versions here were human readable and could give an indicator of what to do. With the versions being hashes now, though, this message isn't quite as appropriate as it was before.  Perhaps these indicators could be less emphasized? Or could the version still be serialized somewhere and it's only used for these purposes?

Mostly I'm just thinking that hashes shouldn't be too prominently displayed here.

chinedufn: Ah right, I forgot that the versions are now hashes when I was editing this. 

I'll expose the Cargo.toml version in the schema so that be used here and reword the message to more or less suggest that the easiest solution is to make the CLI and wasm-bindgen versions match.

guide/src/reference/cli.md
chinedufn: Cool, I'll remove all wasm-pack recommendations for build related things.

While we're at it, I think we can replace the [`wasm-pack` test recommendation](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/usage.html#appendix-using-wasm-bindgen-test-without-wasm-pack) with just an example of how to download the webdriver yourself. 

I'll go ahead and do that, but if you happen to see this before I dive in and you think it's a bad idea just let me know.

I'll also include a an example of [setting the runner with an environment variable](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner) instead of needing to manage the config file.

chinedufn: Now that we're no longer recommending wasm-pack I don't think this quick little hash solution is good enough anymore. A little too brittle. 

For example, if rustfmt got a new rule and we ran `cargo fmt` and the schema file got touched we'd have a new version for no good reason.

I'm going to change the approach to something along the lines of:

1. Generate a `SCHEMA_HASH` in the build script.
2. Set `SCHEMA_VERSION` by hand to the current `Cargo.toml` version
3. Add a test case to the bottom of `shared/src/lib.rs` that asserts that the current hash is equal to the last human approved hash.
4. Whenever the hash changes tests will fail. Then a human can either decide to keep the `SCHEMA_VERSION` the same, or bump it to the latest `Cargo.toml` version if the schema has indeed changed. This would be documented above the test.

---

For my own needs just bumping versions every now and then would've been fine, even if sometimes unnecessary.

I'm only going this far because a lot of people use wasm-bindgen so I don't want the unnecessary churn across the entire ecosystem.

The most robust would probably be to pull the `shared_api!`  macro out into a crate, have the `shared` crate re-export it and have the `shared/build.rs` depend on that `shared_api!` macro directly along with the encode_api! and decode_api! macros and run them all, hash the results and then use that as our source of truth hash.... But I'll hold off on that since the approach suggested above seems like it will meet our needs in practice just fine.

alexcrichton: For changes to testing, sounds good! For human verification the hash changes, also sounds good!

I'm fine for restructuring things too, but it's probably not too necessary to go too overboard. Honestly things don't change enough nowadays that manually changing the schema version/hash is probably fine.

crates/cli-support/src/wit/mod.rs
```diff
@@ -1457,15 +1457,16 @@ fn extract_programs<'a>(
                     "
 
 it looks like the Rust project used to create this wasm file was linked against
-a different version of wasm-bindgen than this binary:
+version of wasm-bindgen that uses a different bindgen format than this binary:
 
-  rust wasm file: {}
-     this binary: {}
+  rust wasm file schema version: {}
```
chinedufn, line 1462: I suppose this could be more descriptive and make it clear why the schema version is different from the actual version of the crate/binary that you're using, but in practice since the schema seems to be fairly stable at this point I don't think we need to come up with the perfect language here.

chinedufn:
@alexcrichton alright I edited some sections in the guide to focus on wasm-bindgen-cli, but I didn't get all of them. Notably the testing section still revolves around wasm-pack. 

I think that these remaining sections can just be refreshed over time.

Otherwise, I think we're good to go here, so just let me know if there's anything wrong or missing and I'll get it fixed.

chinedufn:
Hmm the failing UI tests don't seem related to my changes but I see that tests are passing on the master branch.

Here's an example failure:

```
EXPECTED:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error[E0277]: the trait bound `std::result::Result<wasm_bindgen::JsValue, wasm_bindgen::JsValue>: FromWasmAbi` is not satisfied
 --> $DIR/missing-catch.rs:6:9
  |
6 |     pub fn foo() -> Result<JsValue, JsValue>;
  |            ^^^ the trait `FromWasmAbi` is not implemented for `std::result::Result<wasm_bindgen::JsValue, wasm_bindgen::JsValue>`
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈

ACTUAL OUTPUT:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error[E0277]: the trait bound `Result<wasm_bindgen::JsValue, wasm_bindgen::JsValue>: FromWasmAbi` is not satisfied
 --> $DIR/missing-catch.rs:6:9
  |
6 |     pub fn foo() -> Result<JsValue, JsValue>;
  |            ^^^ the trait `FromWasmAbi` is not implemented for `Result<wasm_bindgen::JsValue, wasm_bindgen::JsValue>`
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
note: If the actual output is the correct output you can bless it by rerunning
      your test with the environment variable TRYBUILD=overwrite

```

Looks like the actual output isn't using the fully qualified path for `std::result::Result`.

alexcrichton:
For the UI test failures are you using the same compiler version? A release just happened and I think it's using stable rustc, which often tweaks error messages slightly. I can also fix this up after merging.

alexcrichton:
Thanks!

Issue #2549, opened 2021-05-09
Running *part* of your wasm in a `web_sys::Worker`?


sgasse:
### Summary

What is the best architectural approach to have a 'pure' wasm web application that runs part of the calculations in a web worker?

### Additional Details

I am still fairly new to Rust and especially web_sys. However I find the concept of wasm really exciting and I am currently trying out to build an easy web app. It needs to react to clicks on a canvas, some buttons etc., but then there are heavier calculations which should run in the background.

Initially, I followed the [Wasm Game of Life](https://rustwasm.github.io/docs/book/) tutorial and found a good first entry point there. After reading this guide, I designed my app to handle the user interaction part in JS. The background calculation was running actually in the foreground but returned very regularly, giving a false impression of responsiveness.

Next I read about web_sys and especially [building wasm apps that can be deployed without a bundler](https://rustwasm.github.io/wasm-bindgen/examples/without-a-bundler.html). Now I rewrote the user interaction part almost purely in Rust and web_sys. Knowing that there is probably [no native multiprocessing](https://users.rust-lang.org/t/wasm32-multithreading-co-op-is-fine/32564/2) in wasm as of now, web workers are the only alternative.

My dream scenario would be to spawn a `web_sys::Worker` and hand it a struct with a method or a function callback to run. However it looks like a web worker expects a URL to execute ([1](https://docs.rs/web-sys/0.3.27/web_sys/struct.Worker.html#method.new), [2](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker)). Loading in any local JS via the URL works, but not a wasm module.

If I create a worker in my main wasm app
```
pub fn setup_worker() {
    Worker::new("worker.js");
    console::log_1(&"Created a new worker".into());
}
```

and have this minimal `worker.js`:
```
import Dummy from './pkg/my_wasm_app.js';
```

I get the error `SyntaxError: import declarations may only appear at top level of a module`. This might have JS reasons (not so familiar with JS ^^). Though maybe someone could point me in the right direction? Which one of the following options is possible? And which one is best?

1. Setup a separate wasm project with target so that I can load an `init`. I think it is not possible to have two wasm targets in the same `cargo.toml` so it would be annoying to maintain two project folders for essentially one projects. Also I would need to figure out a better way to load it given that the above approach at least for _some_ imported module was not allowed in the worker.
[...]

PR #2556, opened 2021-05-14
Add wasm in web worker example

sgasse:
This commit adds a parallel execution example in which we spawn a web
worker with `web_sys`, run WASM code in the web worker and interact
between the main thread and the web worker.

It is intended to add an easy starting point for parallel execution
with WASM, complementing the more involved parallel raytrace example.

Related to GitHub issue #2549

### DIFF
--- a/guide/src/examples/wasm-in-web-worker.md
+++ b/guide/src/examples/wasm-in-web-worker.md
@@ -1,0 +1,71 @@
+# WASM in Web Worker
+
+[View full source code][code]
+
+[code]: https://github.com/rustwasm/wasm-bindgen/tree/master/examples/wasm-in-web-worker
+
+A simple example of parallel execution by spawning a web worker with `web_sys`,
+loading WASM code in the web worker and interacting between the main thread and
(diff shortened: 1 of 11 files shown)


guide/src/examples/wasm-in-web-worker.md
```diff
@@ -0,0 +1,71 @@
+# WASM in Web Worker
+
+[View full source code][code]
```
sgasse, line 3: I saw that other examples not only point to the code but also to a compiled demo - but that is not autogenerated for each example, right?

alexcrichton:
Thanks!

I think CI needs to be tweaked though when building this example?

sgasse:
Ah yes, I missed that another test failed ^^
If I understood the error correctly, then it failed to build my example because there is no `package.json`. Though that should also not be there since I made an example to be built with `--target no-modules`. So I added an ignore case (just like for `without-a-bundler` - is that right?

alexcrichton:
Yep that's fine!

PR #2575, opened 2021-06-07
Enable additional TextMetrics properties

paulgb:
When the webidl for `TextMetrics` was originally created, no properties except `width` had widespread browser support. Today, a number of additional metrics have reached widespread browser support ([caniuse link](https://caniuse.com/?search=TextMetrics)).

This PR enables the additional properties that have widespread browser support. I leave `emHeight{Ascent,Descent}` and `{hanging,alphabetic,ideographic}Baseline` commented as they currently have only Safari support.


alexcrichton:
Thanks!

PR #2580, opened 2021-06-10
3 Features were missing in Cargo.toml

halzy:
The features existed and were referred to elsewhere, they were missing from the Cargo.toml file.


alexcrichton:
Oops, thanks!

Issue #1793, opened 2019-09-25
Add getters for WebIDL's dictionaries.


alexlapa:
### Motivation

Currently `web-sys` wrappers do not have any getters for WebIDL dictionaries (e.g. [MediaStreamConstraints](https://docs.rs/web-sys/0.3.27/web_sys/struct.MediaStreamConstraints.html)). 

Such fields can be accessed using [js_sys::Reflect](https://docs.rs/js-sys/0.3.6/js_sys/struct.Reflect.html), which is quite awkward.

### Proposed Solution

Add getters for WebIDL dictionaries fields.

### Alternatives

Some generic getter like: `fn get(name:&str) -> JsValue`

PR #2584, opened 2021-06-11
Enabling more Stream functionality

halzy:
This PR is rebased it off of the PR #2586 so that it's easier to see what that PR does.

This is useful when compiling to WASM for the Cloudflare Workers:
https://developers.cloudflare.com/workers/runtime-apis/streams

Added Unstable WebIDL:
   unstable/ReadableStreamDefaultReader.webidl                                                       
   unstable/ReadableStreamGenericReader.webidl                                                       
   unstable/QueuingStrategy.webidl                                                                  
   unstable/ReadableStreamBYOBReader.webidl                                                         
   unstable/TransformStream.webidl                                                                  
   unstable/WritableStream.webidl                                                                   
   unstable/WritableStreamDefaultWriter.webidl


crates/web-sys/Cargo.toml
```diff
@@ -1237,6 +1243,7 @@ TouchInit = []
 TouchList = []
 TrackEvent = ["Event"]
 TrackEventInit = []
+TransformStream = ["ReadableStream", "WritableStream"]
```
halzy, line 1246: TransformStream is useless without these.

crates/web-sys/src/features/gen_ReadableStream.rs
```diff
@@ -73,6 +73,28 @@ extern "C" {
         transform: &ReadableWritablePair,
         options: &StreamPipeOptions,
     ) -> ReadableStream;
+    #[cfg(web_sys_unstable_apis)]
```
halzy, line 76: I'm not certain this is correct. `pipe_to` takes a `WritableStream` (unstable) as a parameter.

crates/web-sys/src/features/gen_ReadableStreamByobReadResult.rs
```diff
@@ -0,0 +1,64 @@
...
+        use wasm_bindgen::JsValue;
+        let r = ::js_sys::Reflect::set(self.as_ref(), &JsValue::from("value"), &JsValue::from(val));
+        debug_assert!(
+            r.is_ok(),
+            "setting properties should never fail on our dictionary objects"
+        );
+        let _ = r;
+        self
+    }
+}
```
halzy, line 64: This type isn't created by the API user. It would benefit from #1793 and the setters are not useful.

crates/web-sys/src/features/gen_ReadableStreamDefaultReadResult.rs
```diff
@@ -0,0 +1,48 @@
...
+        use wasm_bindgen::JsValue;
+        let r = ::js_sys::Reflect::set(self.as_ref(), &JsValue::from("value"), &JsValue::from(val));
+        debug_assert!(
+            r.is_ok(),
+            "setting properties should never fail on our dictionary objects"
+        );
+        let _ = r;
+        self
+    }
+}
```
halzy, line 48: Same here, not constructed by user, needs getters not setters.

alexcrichton:
Seems reasonable to me! I don't know anything about these APIs myself so I'd basically just be trusting you about the stable/unstable. Was there anything else you wanted to do here before merging though?

halzy:
> Seems reasonable to me! I don't know anything about these APIs myself so I'd basically just be trusting you about the stable/unstable. Was there anything else you wanted to do here before merging though?

@alexcrichton I've had a hard time determining how stable they are. The spec is a living standard and I've been relying on MDN to know if they're 'experimental' or not. After reviewing them again, I'll move them all to unstable as MDN lists the methods of `ReadableStreamDefaultReader` as experimental.

halzy:
I believe this is in a good state for merging now.

alexcrichton:
This is unfortunately a breaking change to the `get_reader` and `get_reader_with_options` APIs. Is there a way to avoid that? Or otherwise do you think it's ok to break things?

halzy:
> This is unfortunately a breaking change to the `get_reader` and `get_reader_with_options` APIs. Is there a way to avoid that? Or otherwise do you think it's ok to break things?

`get_reader` and `get_reader_with_options` are returning `unstable` types, so this change seems reasonable. We may want to consider a future change that would make these functions `unstable` due to the return type, but that would cause users to have to enable 'unstable' when they didn't have it before.

In the ReadableStream webidl it has the return type for getReader defined as a `ReadableStreamReader`:
`ReadableStreamReader getReader(optional ReadableStreamGetReaderOptions options = {});`

The typedef for `ReadableStreamReader` is:
`typedef (ReadableStreamDefaultReader or ReadableStreamBYOBReader) ReadableStreamReader;`

These 2 types didn't exist before and I expect it's choosing `Object` as the return type because it's the nearest to both in the type hierarchy.

I think it's OK to break this as it's more correct. Users were likely casting from JsValue into either of those unstable types that they hand-rolled in their own project.

alexcrichton:
Ok seems reasonable to me! This defintely seems like a bug in the code generator though...

PR #2586, opened 2021-06-12
Enable unstable APIs for stable types.

halzy:
When `unstable` types are used in stable methods, those methods should be marked as `unstable`.

For example:
`ReadableStream::pipe_to(writer: WritableStream)` (`WritableStream` is `unstable`)

`ReadableWritablePair` is a dictionary of a `ReadableStream` and `WritableStream`. It's `constructor` and setter for the `WritableStream` should be `unstable`.


alexcrichton:
Thanks!

PR #2590, opened 2021-06-17
Add TSDoc for the `imports`

umidbekk:
Fixes:

```node
 error TS2345: Argument of type '{}' is not assignable to parameter of type 'Imports'.
  Index signature is missing in type '{}'.
```

Here is the repro:
https://github.com/umidbekk/prettier-plugin-prisma/runs/2845989557?check_suite_focus=true

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -203,6 +203,7 @@ impl<'a> Context<'a> {
 
         let mut shim = String::new();
 
+        shim.push_str("/** @type {WebAssembly.Imports} */\n");
         shim.push_str("let imports = {};\n");
 
         if self.config.mode.nodejs_experimental_modules() {


alexcrichton:
Thanks for the PR! I don't think that this is emitted into a typescript file though? You might need to add a suppression for the bindgen-generated files.

umidbekk:
> Thanks for the PR! I don't think that this is emitted into a typescript file though? You might need to add a suppression for the bindgen-generated files.

Hey @alexcrichton, thanks for the review! I'm new to `rust-bindgen` and `rust` itself, so I might not get you right :) 
But as I understood, TS files do not have an actual implementation, and there is no local variables.

I'm using TypeScript to type check JavaScript project, and it start failing on that line 
```js
const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
``` 
It's because variable `let imports = {};` was defined without types and TypeScript automatically inferred `{}` type.
___
On the other hand, I've tried to build with changes from my forked branch 
```toml
wasm-bindgen = { git = "https://github.com/umidbekk/wasm-bindgen", branch = "fix-imports" }
```
And there were no output changes, so I guess I did something wrong, but I can't really point a finger to it 🤔

alexcrichton:
Oh sure yeah, but the `*.js` output of `wasm-bindgen` is not intended to be fed through `tsc`. The `*.d.ts` files generated should describe the interface of the `*.js` files, but it's not a guarantee that this tool provides that the `*.js` files themselves will successfully get fed through `tsc`

umidbekk:
> Oh sure yeah, but the `*.js` output of `wasm-bindgen` is not intended to be fed through `tsc`. The `*.d.ts` files generated should describe the interface of the `*.js` files, but it's not a guarantee that this tool provides that the `*.js` files themselves will successfully get fed through `tsc`

I was using that way only because JS output already had all required types (first introduced in #1412), that was main reason why I could simplify build pipeline by removing TypeScript transpiration step.

But if you're suggesting to emit TS files for checks, that's fine by me too. After some time I found that removing `--no-typescript` was actually fixing my issue. 

Anyway, thanks for the review.

PR #2595, opened 2021-06-21
Added WebAssembly.Global

syrusakbary:
Added WebAssembly.Global

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3577,7 +3577,7 @@ pub mod WebAssembly {
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global)
         #[wasm_bindgen(constructor, js_namespace = WebAssembly, catch)]
-        pub fn new(global_descriptor: &Object, value: Option<&JsValue>) -> Result<Global, JsValue>;
+        pub fn new(global_descriptor: &Object, value: &JsValue) -> Result<Global, JsValue>;
 
         /// The value prototype property of the `WebAssembly.Global` object
         /// returns the value of the global.
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3577,7 +3577,7 @@ pub mod WebAssembly {
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global)
         #[wasm_bindgen(constructor, js_namespace = WebAssembly, catch)]
-        pub fn new(global_descriptor: &Object, value: Option<&JSValue>) -> Result<Global, JsValue>;
+        pub fn new(global_descriptor: &Object, value: Option<&JsValue>) -> Result<Global, JsValue>;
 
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -3561,6 +3561,33 @@ pub mod WebAssembly {
...
+        #[wasm_bindgen(js_namespace = WebAssembly, extends = Object, typescript_type = "WebAssembly.Global")]
+        #[derive(Clone, Debug, PartialEq, Eq)]
+        pub type Global;
+
+        /// The `WebAssembly.Global()` constructor creates a new `Global` object
+        /// of the given type and value.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global)
+        #[wasm_bindgen(constructor, js_namespace = WebAssembly, catch)]
+        pub fn new(global_descriptor: &Object, value: Option<usize>) -> Result<Global, JsValue>;
```
syrusakbary, line 3581: I'm unsure if here value should be `Option<f64>` instead

crates/js-sys/src/lib.rs
```diff
@@ -3561,6 +3561,33 @@ pub mod WebAssembly {
...
+
+        /// The value prototype property of the `WebAssembly.Global` object
+        /// returns the value of the global.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global)
+        #[wasm_bindgen(method, getter, js_namespace = WebAssembly)]
+        pub fn value(this: &Global) -> JsValue;
```
syrusakbary, line 3588: Should this return `f64` ?

crates/js-sys/src/lib.rs
alexcrichton: I think `Object` here should probably be a `GlobalDescriptor` dictionary with typed field accesses, and `value` should be `JsValue` most likely to support `externref` globals.

crates/js-sys/src/lib.rs
alexcrichton: I think `JsValue` here is fine for `externref`, but can this also add a setter?

alexcrichton (commented):
Looks good to me, but I think CI is failing?

crates/js-sys/src/lib.rs
syrusakbary: Done. Added a setter as well.
Note that I also added the `structural` attr to value, but it would be great to assure that's the correct way

crates/js-sys/src/lib.rs
syrusakbary: Done. I set `value` to `Option<&JsValue>`.

For the `global_descriptor` I used the same pattern as `WebAssembly.Table` or `WebAssembly.Module`: using `Object`.

I do agree that having a `GlobalDescriptor` dictionary would be safer, but I'm not clear on how to do it looking at the code. Any tips here would be appreciated.

syrusakbary:
Tests are failing (`Run UI tests` more specifically), but I have no idea how to resolve them or why they are failing.

Any help here would be appreciated!

alexcrichton:
Thanks!

PR #2600, opened 2021-06-23
Pass-through file urls to Deno.readFile

caspervonb:
This removes the conversion from URL to string and just passes through the file urls directly to `Deno.readFile`.
We support taking URLs as an overload in the runtime and when you have one resolved from import.meta, it's better to just pass it through.

See https://github.com/rustwasm/wasm-bindgen/issues/2296


alexcrichton:
Thanks!

PR #2609, opened 2021-07-03
Upgraded examples/webgl to use WebGL2 instead of WebGL1

mmulli:
Since all recent major browser releases (With the possible exception of Safari iOS) now support WebGL2, I thought it may help others if the example was upgraded to support WebGL2. 

Although, I completely understand if you would like to keep it at webgl 1 to support the widest range of browsers / devices.


alexcrichton:
Thanks!

PR #2616, opened 2021-07-12
Fix webrtc_datachannel example panic in Safari

matthew-ch:
In Safari, the `webrtc_datachannel` example would panic at `dc2.send_with_str("Ping from pc2.dc!").unwrap()`, with the underlying error being `InvalidStateError: The object is in an invalid state`.
This PR fixes this issue by placing `send_with_str` call inside `onopen` callback of `dc2`.


alexcrichton:
Thanks! I think though a `cargo fmt` is needed?

matthew-ch:
> Thanks! I think though a `cargo fmt` is needed?

Done! Glad I can help!

PR #2625, opened 2021-07-17
Implement FromStr for JsString

Jules-Bertholet:


src/lib.rs
```diff
@@ -395,6 +397,13 @@ impl<'a> From<&'a str> for JsValue {
     }
 }
 
+impl FromStr for JsValue {
+    type Err = Infallible;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(Self::from_str(s))
+    }
+}
```
alexcrichton, line 405: I'm a bit hesitatnt to have this impl since `JsValue` is so generic. Often `FromStr` is used in parsing contexts so one might expect this to parse a float, for example, rather than producing a string value. 

Is this required by a use case of yours or would it be ok to leave it out for now?

Jules-Bertholet: No, I don't need this for anything, I just noticed while adding other `impl`s that this one would be easy to add in as well. I'll remove it

PR #2626, opened 2021-07-17
Implement Default for JS types

Jules-Bertholet:
- Implement the `Default` trait for types in `js-sys` where it makes sense
- Generate a `Default` impl for WebIDL types with zero-argument constructors

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1098,12 +1098,6 @@ extern "C" {
     pub fn to_string(this: &Error) -> JsString;
 }
 
-impl Default for Error {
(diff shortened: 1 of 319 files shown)


alexcrichton:
Thanks for this!

I think the 0-argument ctor makes sense for web-sys, although some of the default implementations in js-sys seem like they're a bit ambitiuos, e.g. creating an error without a string and such. Could the initial set of ctors in js-sys be a bit more conservative to basically match `Default` semantics in Rust? (e.g. errors typically don't implement `Default`)

Jules-Bertholet:
Ready for review

PR #2629, opened 2021-07-19
Add math-related intrinsics/functions for `JsValue`s

Jules-Bertholet:
Adds handling for [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), among other things.


alexcrichton:
Thanks! Overall though I'd prefer to keep the set of intrinsics and such to a minimum unless there's a good reason to expand things. Could you elaborate a bit more on the motivation for this? It seems, for example, that some of the conversions (e.g. `From<JsValue> for f64` aren't necessarily idiomatic in Rust since they can be fallible. If it's applying to a specific JS operation then ideally that'd either be done with the `Reflect` structure from `js-sys` but if necessary it seems fine to have an intrinsic.

Jules-Bertholet:
@alexcrichton JS operators generally don't have corresponding methods, so `Reflect` isn't an option in those cases. I could remove the `Object.is` intrinsic, though. `From` definitely shouldn't be fallible, I'll fix that. But do you think it's a problem if traits like `Add` are fallible?

crates/cli-support/src/intrinsic.rs
```diff
@@ -102,9 +105,75 @@ intrinsics! {
         #[symbol = "__wbindgen_is_string"]
         #[signature = fn(ref_externref()) -> Boolean]
         IsString,
+        #[symbol = "__wbindgen_is_bigint"]
+        #[signature = fn(ref_externref()) -> Boolean]
```
CrimsonCodes0, line 109: Why special-case this? __wbindgen_typeof can be used to implement this.

Jules-Bertholet: The other types are already special-cased, so I special-cased this one too for consistency

crates/cli-support/src/intrinsic.rs
```diff
@@ -102,9 +105,75 @@ intrinsics! {
...
+        #[signature = fn(ref_externref()) -> F64]
+        AsNumber,
+        #[symbol = "__wbindgen_try_into_number"]
+        #[signature = fn(ref_externref()) -> Externref]
+        TryIntoNumber,
+        #[symbol = "__wbindgen_neg"]
+        #[signature = fn(ref_externref()) -> Externref]
+        Neg,
+        #[symbol = "__wbindgen_bit_and"]
+        #[signature = fn(ref_externref(), ref_externref()) -> Externref]
```
CrimsonCodes0, line 130: These bitwise operators always return an i32, besides unsigned_shr, which yields u32.
They're currently typed as yielding Externref.

But, on another note, these are already native Rust operators, why use the significantly slower, morr unsound JS operators? Is there anything that can't be done without these?

Jules-Bertholet: These operators only return 32-bit ints when passed JS `number`s. When passed `bigint`s they return `bigint`s. (The only exception is `>>>` (`unsigned_shl`), which does not accept `bigint`s; I included it for completeness but if it's deemed unnecessary I can remove it

Jules-Bertholet:
I've removed the `Object.is` intrinsic and made the `From<JsValue>` `f64` impl into a `TryFrom`

alexcrichton:
Can you elaborate more on the motivation for this? This is a lot of library support with a lot of traits and a lot of intrinsics. It's a good deal to maintain in what should ideally be a pretty lean core library.

Jules-Bertholet:
The motivation for this is mostly to allow performing mathematical operations on JS `bigint`s. There currently isn't any other method in `wasm-bindgen` for handling these values. And the only way to add that functionality is to allow accessing the JS math operators, because there aren't any equivalent methods. Note that plan to make a follow-up PR to add a `BigInt` type to `js-sys` to expose this functionality in a more strongly-typed manner, so I'm not particularly attached to the current API in this PR.

The only two intrinsics this PR adds that aren't useful for `bigint`s are the unary `+`,  `>>>`, and `in`. I added these for completeness (if every other JS operator is going to be made available, why leave out these three?), but I'm not attached to them. I suppose `+` and `>>>` could be marginally useful when dealing with objects with a custom `Symbol.toPrimitive`, but I don't think that's a very common use case.

Jules-Bertholet:
> follow-up PR to add a BigInt type to js-sys to expose this functionality in a more strongly-typed manner

I've been working on this, but I've run into a roadblock: the [`BigInt()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/BigInt) doesn't use `new`, and as far as I can tell `wasm-bindgen` doesn't have any facilities to account for this.

alexcrichton:
Ah ok, if possible then I think it would be best to bind `BigInt` in `js-sys` without adding new features to wasm-bindgen itself. Perhaps we could try to work through the issues that you're seeing there?

For example for the `BigInt` constructor you could do something like:

```rust
impl BigInt {
    pub fn new(arg: &JsValue) -> BigInt {
        // manually call the `BigInt` name bound as a function
    }
}
```

where it's not necessarily pure `extern "C" { ... }` blocks but that's ok.

Jules-Bertholet:
Ready for review

trevyn:
This `BigInt` support is working perfectly for me! 👍🏼

Cargo.toml
```diff
@@ -40,6 +40,7 @@ wasm-bindgen-macro = { path = "crates/macro", version = "=0.2.76" }
 serde = { version = "1.0", optional = true }
 serde_json = { version = "1.0", optional = true }
 cfg-if = "1.0.0"
+num-traits = { version = "0.2.14", optional = true }
```
alexcrichton, line 43: This is a small enough addition could this be a method instead of a trait implementation?

alexcrichton (commented):
Sorry for the delay. I'm not overly in love with how this is such a large increase in the API surface area, but I dunno I don't really have anything against it. I would prefer to avoid the `num-traits` dep but other than that seems fine I guess?

Jules-Bertholet:
@alexcrichton The `num-traits` dep is just there to allow easier integration with other Rust crates that also use `num-traits`, you don't miss out on features by not enabling it. There's a standalone `pow` method you can use instead of its `Pow` trait, and all of the other trait impls are trivial.

alexcrichton:
Makes sense, yeah, but I'd prefer to leave it out and keep wasm-bindgen at least somewhat slim in deps.

crates/js-sys/Cargo.toml
```diff
@@ -18,8 +18,13 @@ edition = "2018"
 test = false
 doctest = false
 
+[features]
+rust-num = ["num-bigint", "num-traits"]
+
 [dependencies]
 wasm-bindgen = { path = "../..", version = "0.2.76" }
+num-bigint = { version = "0.4.1", optional = true }
+num-traits = { version = "0.2.14", optional = true }
```
alexcrichton, line 27: Sorry but personally I do not want to add new public dependencies to these crates. There is no story for landing breaking changes to js-sys at this time and having new public dependencies increases the risk that a breaking change may be needed (in case these deps have a major-version-bump)

I realize these are optional but it's still part of the public API and something that I'm not willing to commit to.

Issue #1775, opened 2019-09-19
How do you use a Rust struct with a `String` field?


garrettmaring:
### Summary

How do you use a Rust struct with a `String` field using wasm-bindgen?

The `String` type seems to be supported for function parameters and return values. https://rustwasm.github.io/docs/wasm-bindgen/reference/types/string.html

### Additional Details

```rust
#[wasm_bindgen]
pub struct Data {
    id: String,
}
```

alexcrichton:
Thanks for the report! For this you'll want to use [getters and setters](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/getter-and-setter.html), and that shoul dod the trick!

garrettmaring:
That worked! Thank you.

andimarek:
Hi @garrettmaring can you share some details how exactly you solved it with getters and setters? thanks

garrettmaring:
Sure!

```rust
// doesn't work...
#[wasm_bindgen]
struct Data {
    pub id: String,
}
```

You'll get the error **error[E0277]: the trait bound std::string::String: std::marker::Copy is not satisfied**.

Since, the `String` type in Rust [isn't implicitly copyable](https://stackoverflow.com/a/38305901/3599414). I had to read up on the [difference between `Copy` and `Clone`](https://doc.rust-lang.org/std/marker/trait.Copy.html) to understand that I couldn't just implement `Copy` but rather needed to use `.clone()` to explicitly copy it.

Thankfully, _wasm-bindgen_ gives us a simple way to do it. 

```rust
#[wasm_bindgen]
struct Data {
    id: String, // ensure that the field is private
}

#[wasm_bindgen]
impl Data {
    #[wasm_bindgen(getter)]
    pub fn id(&self) -> String {
        self.id.clone()
    }

    #[wasm_bindgen(setter)]
    pub fn set_id(&mut self, id: String) {
        self.id = id;
    }
}
```

There are some interesting things that you can do with getters and setters that are documented [here](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/getter-and-setter.html).

garrettmaring:
@alexcrichton would it be feasible for _wasm-bindgen_ to generate this code if a `struct` implements `Clone`?

alexcrichton:
[...]

PR #2633, opened 2021-07-26
Add `#[wasm_bindgen(getter_with_clone)]` attribute

trevyn:
https://github.com/rustwasm/wasm-bindgen/issues/1775#issuecomment-534144161 :

>it'd be a neat feature to do something like `#[wasm_bindgen(getter_setter_with_clone)]` or something like that so the boilerplate could be drastically reduced

```rust
// doesn't work: "the trait bound std::string::String: std::marker::Copy is not satisfied"
#[wasm_bindgen]
struct Data {
    pub id: String,
}

// now does work
#[wasm_bindgen(getter_with_clone)]
struct Data {
    pub id: String,
}
```

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -343,6 +343,8 @@ pub struct StructField {
     pub comments: Vec<String>,
     /// Whether to generate a typescript definition for this field
     pub generate_typescript: bool,
+    /// Whether to use .clone() in the auto-generated getter for this field
+    pub getter_with_clone: bool,
 }
 
 /// Information about an Enum being exported
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -274,8 +274,19 @@ impl ToTokens for ast::StructField {
         let getter = &self.getter;
         let setter = &self.setter;
 
-        let assert_copy = quote! { assert_copy::<#ty>() };
-        let assert_copy = respan(assert_copy, ty);
+        let maybe_assert_copy = if self.getter_with_clone {
+            quote! {}
+        } else {
+            quote! { assert_copy::<#ty>() }
+        };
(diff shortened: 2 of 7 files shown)


alexcrichton:
Thanks! Can you add both documentation and tests for this? Also I would naively expect this to be at the field level rather than the struct level?

trevyn:
@alexcrichton Done. It works at both the field level and the struct level, to reduce boilerplate in the case where it needs to be applied to many fields.

alexcrichton:
Thanks!

fosskers:
Hah, just ran into the need for this today!

brandonros:
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(getter_with_clone)]
pub struct Input {
    pub name: String
}

#[wasm_bindgen]
extern {
    pub fn alert(s: &str);
}

#[wasm_bindgen]
pub fn greet(input: Input) {
    alert(&format!("Hello, {}!", input.name));
}
```

```html
<!doctype html>
<html>
  <head>
  </head>
  <body>
    <script type="module">
      import * as wasm from '../pkg/wasm_driver.js'

      const run = async () => {
        await wasm.default()
        const input = new wasm.Input()
        input.name = 'James'
        wasm.greet(input)
      }

      run()
    </script>
  </body>
</html>
```

```
(index):14 Uncaught (in promise) Error: null pointer passed to rust
    at imports.wbg.__wbindgen_throw (wasm_driver.js:179:15)
    at 554b666e:0x38a8
    at 554b666e:0x3890
    at 554b666e:0x342d
    at Input.set name [as name] (wasm_driver.js:134:14)
    at run ((index):12:20)
```

trevyn:
@brandonros I think you have to do something like this:

```rust
#[wasm_bindgen]
impl Input {
    #[wasm_bindgen(constructor)]
    pub fn new() -> Self {
        Self::default()
    }
}
```

PR #2649, opened 2021-08-07
Implement To/FromWasmAbi for arrays

Jules-Bertholet:
Allows using Rust arrays as arguments and return values for `wasm-bindgen` functions.

### DIFF
--- a/src/convert/slices.rs
+++ b/src/convert/slices.rs
@@ -146,6 +146,8 @@ cfg_if! {
 }
 
 if_std! {
+    use std::convert::TryFrom;
+
     impl<T> IntoWasmAbi for Vec<T> where Box<[T]>: IntoWasmAbi<Abi = WasmSlice> {
         type Abi = <Box<[T]> as IntoWasmAbi>::Abi;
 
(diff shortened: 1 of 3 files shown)


src/convert/slices.rs
```diff
@@ -174,6 +176,38 @@ if_std! {
...
+
+    impl<T, const N: usize> FromWasmAbi for [T; N] where Box<[T]>: FromWasmAbi<Abi = WasmSlice> {
+        type Abi = <Box<[T]> as FromWasmAbi>::Abi;
+
+        #[inline]
+        unsafe fn from_abi(js: Self::Abi) -> Self {
+            let result = <[T; N]>::try_from(<Vec<T>>::from(<Box<[T]>>::from_abi(js)));
+            return match result {
+                Ok(arr) => arr,
+                Err(_) => panic!("JS array length does not match Rust array")
```
Jules-Bertholet, line 201: This is not ideal. I see a few alternatives here:
1. Accept the panic
2. Add a `T: Default` bound and fill missing values with `default()`
3. Add a `T: JsCast` bound and fill missing values with `JsValue::UNDEFINED`

Jules-Bertholet: I've gone ahead and implemented alternative 2

alexcrichton:
Thanks for the PR! I'd prefer to not implement this via this strategy though becuase this can be a hidden performance trap relative to allocations and transferring/copying memory. I don't necessarily know of an alternative way to implement this that doesn't suffer from such an issue, but I would recomend consumers instead work with `Vec<T>` and then convert to `[T; N]` themselves (e.g. with the `array-init` crate you found) instead of baking that directly into `wasm-bindgen`

Issue #2639, opened 2021-07-29
Typed array -> Vec conversion appears to unnecessarily preinitialize Vec contents


adrian17:
Referring to the code here:

https://github.com/rustwasm/wasm-bindgen/blob/b97837b0989b58595f3ef72bd5691ce420591204/crates/js-sys/src/lib.rs#L5133-L5138

Rust appears to be unable to optimize out the first line, zero-initializing the Vec. This makes the compiler generate a `memset`, which can be slower than actual action of copying the data. Here's an example profile of running `to_vec()` on an 80MB `Uint8Array` on Firefox:

![image](https://user-images.githubusercontent.com/4729533/127533178-0ed93a02-be23-4cd4-9c43-d77e15cf13ff.png)

The other available APIs also make it impossible to avoid - `copy_to` and `raw_copy_to` take a slice, and creating a slice requires elements to be already initialized (otherwise it's UB, in my understanding). And the slice is only really needed for a length check, which `to_vec` already guarantees will pass.

So an extra API taking a pointer (and refactored `to_vec()` using `unsafe`) could make this cheaper without UB... at least on paper - the pointer is never directly written into, so I'm hoping the aliasing rules correctly handle this case of pointer escaping as `usize`.

```rs
/// Requires `dst` to point to a buffer large enough to fit this array's contents.
pub unsafe fn raw_copy_to_ptr(&self, dst: *mut $ty) {
    let buf = wasm_bindgen::memory();
    let mem = buf.unchecked_ref::<WebAssembly::Memory>();
    let all_wasm_memory = $name::new(&mem.buffer());
    let offset = dst as usize / mem::size_of::<$ty>();
    all_wasm_memory.set(self, offset as u32);
}
/// Efficiently copies the contents of this JS typed array into a new Vec.
pub fn to_vec(&self) -> Vec<$ty> {
    let mut output = Vec::with_capacity(self.length() as usize);
    unsafe {
        self.raw_copy_to_ptr(output.as_mut_ptr());
        output.set_len(self.length() as usize);
    }
    output
}
```
And the memset is gone:
![image](https://user-images.githubusercontent.com/4729533/127536071-b9e9fe4e-a8d1-40cc-94e1-5994810e8fbc.png)

Anyway, if you think that's a good idea, I can make a PR with this change? Or you can take it over.

alexcrichton:
[...]

PR #2650, opened 2021-08-08
Improve TypedArray::to_vec performance by not zero-initializing buffer

adrian17:
You can see the rationale and results in https://github.com/rustwasm/wasm-bindgen/issues/2639 . 

Changes:

- changed `raw_copy_to` to take pointers instead of slices (as slices require data to be pre-initialized)
- made it `unsafe`, as even though technically we don't use any unsafe APIs, invalid use can modify random memory
- made it `pub` (this part is optional, let me know if you don't want this PR to extend public API)
- use it in `to_vec()`. The general pattern of `with_capacity->modify through pointer->set_len` is similar to the official example in https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len , and looks sound in general. Still, this does add a significant amount of `unsafe` uses.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -5090,11 +5090,23 @@ macro_rules! arrays {
                 )
             }
 
-            fn raw_copy_to(&self, dst: &mut [$ty]) {
+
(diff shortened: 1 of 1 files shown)


alexcrichton:

PR #2651, opened 2021-08-09
Upgrade the webpack-based examples' npm dependencies (move to webpack 5)

thomasetter:
This is mostly:
```
$ fd webpack.config.js examples/ | xargs sed "s/mode: 'development'/mode: 'development',\n    experiments: {\n        asyncWebAssembly: true\n   }/" -i

$ fd package.json examples/ | xargs sed 's/serve": "webpack-dev-server/serve" : "webpack serve/' -i
```
And in each folder touched by the above:
```
$ ncu -u && npm update && npm update --dev
```
The `weather_report` and the `webxr` examples needed minor adjustments, but besides that everything seems to 

The examples seem to work, except `ducktype` and `guide-supported-types` which build, but I could not really see anything in the browser.
This was the same for me before this change, please let me know if this is not expected.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ members = [
   "examples/wasm-in-wasm-imports",
   "examples/wasm-in-web-worker",
   "examples/wasm2js",
+  "examples/weather_report",
   "examples/webaudio",
   "examples/webgl",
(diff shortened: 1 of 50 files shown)


alexcrichton:
Thanks!

PR #2665, opened 2021-08-22
Generate TypeScript return types for `async` functions

### LINKED ISSUE #2394: `async` functions don't have proper type signatures in Typescript
### Describe the Bug

As soon as a function is `async`, the return type in the Typescript definition is always `any`.

### Steps to Reproduce

1. Annotate an `async` function with `#[wasm_bindgen]`:

```rust
#[wasm_bindgen]
pub fn greet() -> String {
    "Hello World!".to_owned()
}

#[wasm_bindgen]
pub async fn async_greet() -> String {
    "Hello World!".to_owned()
}
```

2. Run `wasm-pack build`

3. Observe the generated typescript types:

```typescript
/* tslint:disable */
/* eslint-disable */
/**
* @returns {string}
*/
export function greet(): string;
/**
* @returns {any}
*/
export function async_greet(): any;
```

### Expected Behavior

The type definition for `async_greet` should be:

```typescript
export function async_greet(): Promise<string>;
```

### Actual Behavior

Function has return type `any`.

trevyn:
Threads through an `asyncness` value and an "inner" type descriptor for functions, then generates `Promise<...>` return types for `async` functions.

Fixes #2394

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -459,9 +459,10 @@ impl TryToTokens for ast::Export {
         // For an `async` function we always run it through `future_to_promise`
         // since we're returning a promise to JS, and this will implicitly
         // require that the function returns a `Future<Output = Result<...>>`
-        let (ret_ty, ret_expr) = if self.function.r#async {
+        let (ret_ty, inner_ret_ty, ret_expr) = if self.function.r#async {
             if self.start {
                 (
+                    quote! { () },
                     quote! { () },
                     quote! {
                         wasm_bindgen_futures::spawn_local(async move {
@@ -472,6 +473,7 @@ impl TryToTokens for ast::Export {
             } else {
                 (
                     quote! { wasm_bindgen::JsValue },
+                    quote! { #syn_ret },
                     quote! {
                         wasm_bindgen_futures::future_to_promise(async move {
                             <#syn_ret as wasm_bindgen::__rt::IntoJsResult>::into_js_result(#ret.await)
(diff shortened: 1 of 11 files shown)


trevyn:
These tests are now failing, where before they were being covered up by the return type elision for `async` functions:

```rust
#[wasm_bindgen]
pub async fn async_return_7() -> Result<AsyncCustomReturn, u32> {
    Ok(AsyncCustomReturn { val: 7 })
}

#[wasm_bindgen]
pub async fn async_return_8() -> Result<AsyncCustomReturn, AsyncCustomReturn> {
    Ok(AsyncCustomReturn { val: 8 })
}

#[wasm_bindgen]
pub async fn async_throw() -> Result<(), js_sys::Error> {
    Err(js_sys::Error::new("async message"))
}
```

```
| the trait `wasm_bindgen::describe::WasmDescribe` is not implemented for `Result<AsyncCustomReturn, u32>`
|
= help: the following implementations were found:
<Result<T, wasm_bindgen::JsValue> as wasm_bindgen::describe::WasmDescribe>
```

Per the docs, they never should have been supported, and indeed, they don't work when removing the `async`:

```markdown
The `Result` type can be returned from functions exported to JS as well as
closures in Rust. Only `Result<T, JsValue>` is supported where `T` can be
converted to JS. Whenever `Ok(val)` is encountered it's converted to JS and
handed off, and whenever `Err(error)` is encountered an exception is thrown in
JS with `error`.
```

Shall I just remove these tests?

alexcrichton:
Thanks for the PR! The implementation here looks pretty reasonable to me. I think it's fine to update those tests as well, yeah.

Can you add some specific tests for this new feature though? Ideally there'd be reference tests to see the output, tests in the main tests suite to exercise that things work, and tests in the typescript area to ensure that everything typechecks as expected.

trevyn:
@alexcrichton

I added a test in the typescript-tests area.

I don't see any existing reference tests at all for `.d.ts` output; am I missing something, or shall I add reference tests for `.d.ts` to the cli/tests/reference section? (Right now they just test the `.js` and `.wat` output)

Also, the only intended output change is to the `.d.ts` files, so there shouldn't be any new behavior outside of TypeScript, and existing async fn tests should still apply. Not sure if there was something specific you wanted me to add in the main tests suite?

PR #2666, opened 2021-08-24
web-sys: add `WorkerType` feature

wngr:
PR #2656 added support for setting `type` and `request_credentials` on
`WorkerOptions`. However, the `WorkerType` struct` was not exposed, as
the feature was not added to Cargo.toml.


alexcrichton:
A bit delayed but I'm getting around to it now at https://github.com/rustwasm/wasm-bindgen/pull/2675

PR #2672, opened 2021-09-04
[WIP] Generics

Jules-Bertholet:
Allow importing generic structs and functions


Jules-Bertholet:
@alexcrichton This isn't quite done (generic closures still need to be implemented, exported Rust functions need to be able to use the imported types, and documentation and tests need to be written), but this is a big feature, and it's at a stage where I would like some feedback on the design before I continue.

With this PR, you can write something like this:
```Rust
#[wasm_bindgen]
extern "C" {
    pub type Generic<T, U = JsValue>;

    #[wasm_bindgen(constructor)]
    pub fn new<T, U>() -> Generic<T, U>;

    #[wasm_bindgen(method)]
    pub fn do_foo<T, U>(this: &Generic<T, U>, index: u32, value: &T) -> Generic<U, T>;
}

fn test<T: JsCast>(gen: &Generic<T>, arg: &T) {
    gen.do_foo(17, arg);
}
```

The motivation is to be able to represent generic JS collection types in Rust, without having everything return a JsValue (as an example, I've made the collections in `js-sys` generic)

Some notes/caveats:
1. Every type parameter in an imported `type` declaration (but not in imported functions) has a `: 'static` bound added (I think this is needed so the drop checker doesn't think it needs to drop something)
2. Every parameter and return type in an imported function that references a generic parameter has a `: JsCast` bound added to it (every imported generic function definition corresponds to a single concrete generated `extern fn`)
3. For whatever reason `rust-analyser` really doesn't like this at all and will crash repeatedly when analyzing code that uses this.

alexcrichton:
I think I probably don't have the capacity in me to help with this right now. This feels like a major new feature to `wasm-bindgen` that would need to be both carefully designed and implemented. I don't really have the time/energy to do either of those (or even help review a change like this at this point).

The intention, though, was that `js-sys` is low-level and ripe for abstractions to be built on top. I don't want to stymie progress with my own lack of energy but at the same time we tried to design `wasm-bindgen` for an ecosystem on top instead of everything going into `wasm-bindgen`. In that sense is it possible to build these sorts of abstractions externally on `wasm-bindgen` instead of adding them directly inside? I realize how nice it is to have it built-in but at some point I also need to push back and say that I'm barely able to maintain `wasm-bindgen` as-is and something else needs to exist in this ecosystem besides solely me maintaining `wasm-bindgen`.

PR #2674, opened 2021-09-07
Fix the "extra-traits" feature of macro-support, allowing it to build again

Jules-Bertholet:
Allows this feature (which is useful for debugging) to build again


alexcrichton:
Oops, thanks!

PR #2677, opened 2021-09-10
Fix weak-ref by-value function parameter UAF 

mikialex:
The <https://github.com/rustwasm/wasm-bindgen/issues/2447> not considered the by-value function parameter case. This MR try fix that. 

All `xxx.ptr = 0;` will now called by `__destroy_into_raw` and unregister the finalizer.

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -641,8 +641,7 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
             js.assert_class(&val, &class);
             js.assert_not_moved(&val);
             let i = js.tmp();
-            js.prelude(&format!("var ptr{} = {}.ptr;", i, val));
-            js.prelude(&format!("{}.ptr = 0;", val));
+            js.prelude(&format!("var ptr{} = {}.__destroy_into_raw();", i, val));
             js.push(format!("ptr{}", i));
         }
 
@@ -661,8 +660,7 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
             js.prelude(&format!("if (!isLikeNone({0})) {{", val));
             js.assert_class(&val, class);
             js.assert_not_moved(&val);
-            js.prelude(&format!("ptr{} = {}.ptr;", i, val));
-            js.prelude(&format!("{}.ptr = 0;", val));
+            js.prelude(&format!("ptr{} = {}.__destroy_into_raw();", i, val));
             js.prelude("}");
             js.push(format!("ptr{}", i));
         }


mikialex:
Could someone help checking why ci failed. Maybe the error not related to this change

danielhenrymantilla (approved):
Indeed, the CI failure seems to stem from a change in the compiler diagnostics; UI tests are generally brittle because of this. Someone ought to take the top of `master`, run the UI tests with the override setting enabled, so as to update the error message expectations, and submit a PR with those changes, so that you can afterwards rebase and fix that.

Regarding this PR, as the author of the one that fixed the `self` case (sorry I did not notice it occurred for non `self` values as well), the changes LGTM ✅

alexcrichton:
Thanks for this! Can you add some tests specifically for this bug as well?

mikialex:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry, the time when FinalizationRegistry cleanup logic triggered is unpredictable. So I think it's hard to add test logic into CI, because your can't await GC happens.

For reference I create a repo simply for reproducing this bug and test the above fix: https://github.com/mikialex/wasm-bindgen-reproduce-2677

mikialex:
I'm trying to add some test but have some issues when figuring out how to run test suite on my machine.

Accroding to this <https://rustwasm.github.io/docs/wasm-bindgen/contributing/testing.html#wasm-tests-on-node-and-headless-browsers> and this <https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/browsers.html#appendix-testing-in-headless-browsers-without-wasm-pack>, I tried `SAFARIDRIVER=/usr/bin/safaridriver cargo test --target wasm32-unknown-unknown` and `cargo test --target wasm32-unknown-unknown` they produce this error:

```
Running `target/debug/wasm-bindgen-test-runner /Users/mikialex/dev/wasm-bindgen/target/wasm32-unknown-unknown/debug/deps/headless-4f54a9baec924af2.wasm`
Set timeout to 20 seconds...
Running headless tests in Safari on `http://127.0.0.1:51646/`
Ok
driver status: signal: 9                          
Error: non-200 response code: 500                 
<html>
<head>
 <title>500 Internal Privoxy Error</title>
 <link rel="shortcut icon" href="http://config.privoxy.org/error-favicon.ico" type="image/x-icon"></head>
<body>
<h1>500 Internal Privoxy Error</h1>
<p>Privoxy encountered an error while processing your request:</p>
<p><b>Could not load template file <code>no-server-data</code> or one of its included components.</b></p>
<p>Please contact your proxy administrator.</p>
<p>If you are the proxy administrator, please put the required file(s)in the <code><i>(confdir)</i>/templates</code> directory.  The location of the <code><i>(confdir)</i></code> directory is specified in the main Privoxy <code>config</code> file.  (It's typically the Privoxy install directory, or <code>/etc/privoxy/</code>).</p>
</body>
</html>

error: test failed, to rerun pass '--test headless'

```
so, did I miss something important? I assum this will execute the tests in tests folder.

alexcrichton:
I'd recommend running the `--test wasm` tests which should only require `node` installed with a somewhat recent version

alexjg:
Am I correct in thinking that this issue is fixed by 4458587c50291a933e808bf75125fb4ed5fb5eed ?

Liamolucko:
> Am I correct in thinking that this issue is fixed by [4458587](https://github.com/rustwasm/wasm-bindgen/commit/4458587c50291a933e808bf75125fb4ed5fb5eed) ?

Yes, I think so.

PR #2680, opened 2021-09-14
Run schema tests in CI

### LINKED ISSUE #2679: cli panics with out of bounds error
### Describe the Bug

Trying to just compile my project to the web, I stumbled upon this panic message.
I think this may happen because of version mismatch between the cli app and the dependency.
After running `cargo update` everything works fine.

### Steps to Reproduce

I can not really show the code. I may try to provide a simplified example if needed

OS: Windows
Rust version: rustc 1.55.0 (c8dfcfe04 2021-09-06)
`wasm-bindgen-cli` version: 0.2.77
`wasm-bindgen` dependency version: 0.2.76

Running command: `wasm-bindgen --target=web --no-typescript --out-dir tmp target/wasm32-unknown-unknown/release/theproject.wasm`

### Expected Behavior

`wasm-bindgen` does its magic without any errors

### Actual Behavior

`wasm-bindgen` panics with following message:

```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', C:\Users\myusername\.cargo\registry\src\github.com-1ecc6299db9ec823\wasm-bindgen-cli-support-0.2.77\src\descriptor.rs:205:15
```

### Additional Context

As I said, looks like problems with different versions. I remember `wasm-bindgen` would just panic instantly with message saying that versions should match. I guess it is not the case anymore? So this took me a while to even think about running `cargo update`

chinedufn:
Fixes #2679

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,6 +41,8 @@ jobs:
         displayName: "Futures test suite on wasm"
       - script: cargo test -p wasm-bindgen-multi-value-xform
         displayName: "multi-value xform tests on native"
+      - script: cargo test -p wasm-bindgen-shared
+        displayName: "Shared test suite on native"
       # TODO: re-enable when binary decoding for reference types is updated in
       # Node.
       # - script: |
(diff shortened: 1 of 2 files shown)


chinedufn:
@alexcrichton mind putting out `0.2.78` after this lands?

alexcrichton:
Oh nice, thanks for checking for this!

alexcrichton:
Sure yeah I can do the bump later tomorrow

PR #2681, opened 2021-09-14
Implement `Extend<A> for Array where A: AsRef<JsValue>`

arnauorriols:
Collections are conventionally expected to implement both `FromInterator` and `Extend` traits. More specifically, I need `Array` to implement `Extend` in order to build one with [`futures::StreamExt::collect()`](https://docs.rs/futures/0.3.17/futures/stream/trait.StreamExt.html#method.collect)


alexcrichton:
Thanks!

PR #2682, opened 2021-09-15
Add Crypto.randomUUID API

littledivy:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

Implemented in Node.js, Deno and Chrome/Edge 92

### DIFF
--- a/crates/web-sys/src/features/gen_Crypto.rs
+++ b/crates/web-sys/src/features/gen_Crypto.rs
@@ -39,4 +39,11 @@ extern "C" {
         this: &Crypto,
         array: &mut [u8],
     ) -> Result<::js_sys::Object, JsValue>;
+    # [wasm_bindgen (method , structural , js_class = "Crypto" , js_name = randomUUID)]
+    #[doc = "The `randomUUID()` method."]
+    #[doc = ""]
+    #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)"]
+    #[doc = ""]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! If this isn't implemented in Firefox or Safari though I think it needs to be unstable?

littledivy:
Landed in Webkit too https://trac.webkit.org/changeset/281284/webkit
Firefox is working on it https://bugzilla.mozilla.org/show_bug.cgi?id=1723674
The specification is also going to be merged with the WebCrypto spec. https://github.com/WICG/uuid/issues/27

I think it's safe to be in stable?

alexcrichton:
"going to be merged" and "working on it" means that it's not finalized, and web-sys only supports finalized APIs by default.

yisibl:
Firefox 95 has shipped.

Dav1dde:
@alexcrichton is there an update on this or is this still considered unstable?

alexcrichton:
I do not follow web standards to the degree that I know whether this is ready to ship or not.

littledivy:
Sorry for the delay in switching to unstable, I tried earlier but it didn't work...will give it a shot again.

hamza1311 (approved):
According to [caniuse](https://caniuse.com/?search=randomUUID()), this is implemented by all major browsers and there is no mentioning of it being unstable in either [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) or [W3C docs](https://w3c.github.io/webcrypto/#Crypto-method-randomUUID)

@alexcrichton can you merge this?

alexcrichton:
Ok great!

PR #2687, opened 2021-09-21
Fix invalid ts definitions for many param fns

nickbabcock:
Currently if one has a function that takes over 26 arguments, the
generated typescript definitions contain invalid typescript identifiers
for the function parameters. The invalid code will cause the typescript
compiler to fail (even if one specifies `skipLibCheck`). To give a
contrived example:

```rust
pub fn param_silliness(
    _data0: u8,
    _data1: u8,
    _data2: u8,
    _data3: u8,
    _data4: u8,
    _data5: u8,
    _data6: u8,
    _data7: u8,
    _data8: u8,
    _data9: u8,
    _data10: u8,
    _data11: u8,
    _data12: u8,
    _data13: u8,
    _data14: u8,
    _data15: u8,
    _data16: u8,
    _data17: u8,
    _data18: u8,
    _data19: u8,
    _data20: u8,
    _data21: u8,
    _data22: u8,
    _data23: u8,
    _data24: u8,
    _data25: u8,
    _data26: u8,
    _data27: u8,
) -> usize {
    0
}
```

Generates a `_bg.wasm.d.ts` that uses `{` as an parameter name which is invalid.

A non-contrived example is bundling the brotli crate:

```rust
pub fn brotli_compress(data: &[u8]) -> Vec<u8> {
    let out = Vec::with_capacity(data.len() / 10);
    let mut reader = Cursor::new(data);
    let cursor = Cursor::new(out);
    let mut compressor = brotli::CompressorWriter::new(cursor, 4096, 9, 22);
    std::io::copy(&mut reader, &mut compressor).unwrap();
    compressor.into_inner().into_inner()
}
```

Which generates a function (`BroccoliDestroyInstance`) with a cool 121
parameters and a lot of invalid identifiers.

The commit fixes this issue by appending additional English letters to
the parameter names when a function exceeds 26 arguments. Functions with
less than 26 arguments will see no difference in output. Functions
greater than 26 arguments will see an output like:

```ts
export function param_silliness(
    // ...
    z: number, aa: number, ba: number
): number
```

### DIFF
--- a/crates/cli-support/src/wasm2es6js.rs
+++ b/crates/cli-support/src/wasm2es6js.rs
@@ -1,5 +1,6 @@
 use anyhow::{bail, Error};
 use std::collections::HashSet;
+use std::fmt::Write;
 use walrus::Module;
 
 pub struct Config {
@@ -44,6 +45,16 @@ impl Config {
     }
 }
 
+// Function to ensure we always append a valid typescript parameter name based
+// on parameter index
+fn push_index_identifier(i: usize, s: &mut String) {
+    let letter = b'a' + ((i % 26) as u8);
+    s.push(letter as char);
+    if i >= 26 {
+        write!(s, "{}", i / 26).unwrap();
+    }
+}
+
 pub fn interface(module: &Module) -> Result<String, Error> {
     let mut exports = String::new();
 
@@ -68,7 +79,8 @@ pub fn interface(module: &Module) -> Result<String, Error> {
             if i > 0 {
                 args.push_str(", ");
             }
-            args.push((b'a' + (i as u8)) as char);
+
+            push_index_identifier(i, &mut args);
             args.push_str(": number");
         }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh dear, what a weird bug but thanks for the fix! I think though it's fine to avoid trying to do a, b, c, ... and instead do something like a1, a2, a3, ... which should be easier to implement. Otherwise can you also add a test for this?

nickbabcock:
@alexcrichton, I tweaked the output a bit based on your suggestion, let me know if further tweaking is wanted. It now looks like:

```ts
a1: number, b1: number, ... a2: number, b2: number
```

Unit tests have been added. I looked for some integration tests that involved _bg.wasm.d.ts files, but I couldn't find any. So feel free to point me in their direction.

alexcrichton:
Works for me!

Issue #1742, opened 2019-08-28
Replace Result<T, JsValue> with Result<T, Error>


Pauan:
### Motivation

As discussed in #1735, it's not idiomatic to throw JS values, like `throw "foo"`, because that loses stack traces.

Instead it's *highly* recommended to always use `throw new Error("foo")`. In wasm-bindgen, this corresponds to `Err(js_sys::Error::new("foo"))`

Internally, wasm-bindgen itself follows this recommendation, however, because the wasm-bindgen (and js-sys and web-sys) APIs return `Result<T, JsValue>`, this encourages *user* functions to return `Result<T, JsValue>` as well, which makes it easy for users to return non-`Error` values:

```rust
fn foo() -> Result<(), JsValue> {
    call_some_wasm_bindgen_api()?;
    Err(JsValue::from("hi!"))
}
```

### Proposed Solution

All wasm-bindgen APIs which return `Result<T, JsValue>` should instead return `Result<T, Error>`.

This includes `wasm_bindgen(catch)` (which should have a `debug_assert` to verify that it actually is an `Error` object)

This uses the static type system to make it impossible to *accidentally* throw non-`Error` types. It also is more "Rust-like", which makes the APIs easier to understand.

This is obviously a very large breaking change.

### Alternatives

* Do nothing. This is a lot less work, and isn't a breaking change, but it does mean that wasm-bindgen is encouraging bad behavior.

* Keep `Result<T, JsValue>`, but put in runtime checks which verify that the `Err` is a `js_sys::Error`. This is also a lot less work, but it does have a small runtime cost, and it's surprising behavior for the user.

alexcrichton:
This is somewhat similar to https://github.com/rustwasm/wasm-bindgen/issues/1017 and https://github.com/rustwasm/wasm-bindgen/issues/1004 as well I think. I'd love to start immediately making progress on this if we can instead of delaying it to a batch of breaking changes though, that'd help us get our feet wet and experiment with with possible ideas before we set it all in stone.

One thing we might want to do to start out is to support returning custom error types, converting them with `Display` to a `new Error(rust_error.to_string())` (basically). Next it might be prudent to go ahead and start warning about `Err(JsValue::from("hi"))` in some capacity, or maybe even moving `Error` into `wasm-bindgen` the crate itself to make it more easily available

Pauan:
[...]

PR #2710, opened 2021-10-27
Fix `fn () -> Result<T, JsValue>` leaking stack space

### LINKED ISSUE #1963: Support for `Result<T, JsValue>` can leak stack space
Given code that looks like this:

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn bar(val: JsValue) -> Result<u64, JsValue> {
    foo(&mut 0);
    Err(val)
}

#[inline(never)]
pub fn foo(a: &mut u128) {
    *a = 1;
}
```

we can build and run it with:

```
$ wasm-pack build --target nodejs
```

and run it with:

```js
const { bar } = require('./pkg/wasm_repro.js');

const err = new Error();
for (let j = 0; j < 100000; j++) {
  try {
    bar(err);
  } catch (e) {
    if (e === err) {
      continue;
    }
    console.log('iteration', j);
    throw e;
  }
}
```

to yield:

```
iteration 65536
/home/alex/code/wasm-memory-test/index.js:12
    throw e;
    ^

RuntimeError: memory access out of bounds
    at bar (wasm-function[1]:0x94)
    at module.exports.bar (/home/alex/code/wasm-memory-test/pkg/wasm_repro.js:50:10)
    at Object.<anonymous> (/home/alex/code/wasm-memory-test/index.js:6:5)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1103:10)
    at Module.load (internal/modules/cjs/loader.js:914:32)
    at Function.Module._load (internal/modules/cjs/loader.js:822:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1143:12)
    at internal/main/run_main_module.js:16:11
```

The problem here is apparent when we diassemble the wasm output:

```wat
(module
  (type (;0;) (func (param i32)))
  (type (;1;) (func (param i32 i32)))
  (import "./wasm_repro.js" "__wbindgen_rethrow" (func $__wbindgen_rethrow (type 0)))
  (func $bar (type 1) (param i32 i32)
    (local i32)
    global.get 0
    i32.const 16
    i32.sub
    local.tee 2
    global.set 0
    local.get 2
    i64.const 0
    i64.store offset=8
    local.get 2
    i64.const 0
    i64.store
    local.get 2
    call $wasm_repro::foo::h2ba1f0de78e4c442
    local.get 1
    call $wasm_bindgen::throw_val::h4d534fc8bf695b07
    unreachable)
  (func $wasm_repro::foo::h2ba1f0de78e4c442 (type 0) (param i32)
    local.get 0
    i64.const 0
    i64.store offset=8
    local.get 0
    i64.const 1
    i64.store)
  (func $wasm_bindgen::throw_val::h4d534fc8bf695b07 (type 0) (param i32)
    local.get 0
    call $__wbindgen_rethrow
    unreachable)
  (memory (;0;) 17)
  (global (;0;) (mut i32) (i32.const 1048576))
  (export "memory" (memory 0))
  (export "bar" (func $bar))
  (data (;0;) (i32.const 1048576) "\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00"))
```

Here the problem we can see is that the export and function `bar` allocate 16 bytes of stack space, but this stack space is never released. This means that our global stack pointer is constantly decreasing, and will never get bumped back up.

The **tl;dr;** is that every time an error is thrown/return from a Rust/wasm function it runs the risk of not reclaiming used stack space, and this is clearly a bug we need to fix!

cormacrelf:
Fixes #1963

### Overview

- No longer implements `-> Result<T, JsValue>` by throwing it from within the wasm stack frame
- Instead, defers the throw until the outer glue code's return path
- Expands the acceptable return-position types to `Result<T, E> where T: IntoWasmAbi, E: Into<JsValue>`, so you can easily throw your own custom error classes imported from JS
- Adds a new `JsError` that makes the common need for `throw new Error("message")` easier to use. It converts automatically from `std::error::Error` for use with `?`.

Basically, while you can continue using `-> Result<T, JsValue>` and it will no longer leak, this is overall much better at its job:

```diff
 #[wasm_bindgen]
-fn exported_fn() -> Result<i32, JsValue> {
+fn exported_fn() -> Result<i32, JsError> {
     let v = [0u8; 1000];
-    Err("string errors are terrible anyway".into())
+    Err(JsError::new("error message"))
 }
```

You can also use `?` with any `std::error::Error` implementation, e.g. `anyhow::Error`, using `thiserror` on your own types, or just adding `impl core::fmt::Display for MyType { ... } impl std::error::Error for MyType {}`.

The remaining questions are

- [x] What to do about existing `Result<T, JsValue>` where `T: IntoWasmAbi`. That's distinct from `T: Into<JsValue>` (for example, `#[wasm_bindgen] enum A { ... }` does not implement `Into<JsValue>`), so you can't just change the impl to `T: Into<JsValue>`.
  - You could just deprecate it. It leaks. Throwing non-Error objects is a terrible anti-pattern anyway.
- [x] What to do about async. I think just supporting `JsError` in exported async function signatures with minimal conversion should be enough.
- [x] How to support custom error classes.
  - You probably want this to make your errors catchable separately from eg a wasm `RuntimeError`.
  - You would have to expose a `JsError` constructor that converts from a `JsValue` without implementing `From<JsValue>` lest `?` convert any old `JsValue` without you meaning to.
  - You might want a way to annotate on an imported class that it happens to subclass Error.
  - Very hard to get `?` conversion perfectly on these while also allowing conversion from `T: std::error::Error`. You could pick one or the other. Or you could add a second type `JsErrorSubclass` or something, but that's getting a bit much.
- [x] Documentation on the new types
- [ ] Figure out how it might fit in with the [Exception Handling proposal](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md). ([Implementations](https://webassembly.org/roadmap/))

<details>
<summary>Obsolete</summary>

### Overview (obsolete)

- new type `JsError` equivalent to `js_sys::Error` that implements conversions from `T` where `T: std::error::Error` using `T::to_string()`
- `impl ReturnWasmAbi for Result<T, JsError>` distinct from the `Result<T, JsValue>`, that just converts the `JsError` to a `JsValue` and returns it instead of using `wasm_bindgen::throw_val(e)`
- adds a `function throwIfError(x) { if (x instanceof Error) { throw x } }`
- Modifies the glue code to `let tmp0 = takeObject(...); throwIfError(tmp0); return tmp0`

### Notes

This is more lightweight than the `WasmResult` I described in the issue. It only works for Result types with `T: Into<JsValue> + WasmDescribe` and `E = JsError`. The outcome is similar to `WasmResult` in that every result return involves creating a `JsValue` (`ok.into()` or `JsError::from` something, e.g. via `?`) but it does not require a whole `WasmResult` class in the glue code. The glue code just checks if the returned heap object happens to be an `instanceof Error` and throws the error if so. You get a reasonably good stack trace because the Error was created in a Rust stack frame.

</details>


cormacrelf:
Thanks for having a look. I owe a lot to this library (my job!) and I hope I can give back.

As it stands, this is not a breaking change. It only applies when you use JsError instead of JsValue, and JsError is a new Rust type so nobody is using it yet. The added test file notes that the old style `Result<T, JsValue>` will still leak stack space, but it does still "work" unchanged. The new style has some differences in terms of what you can return, but it's a voluntary upgrade.

However, fixing the old style results **as well** would require first adding the original WasmResult idea instead of checking `x instanceof Error`, and then either:

1. A convoluted system for converting these `T: IntoWasmAbi` into JsValues; or
2. Changing the API, so you can only return `T: Into<JsValue>`.

Like the JsError version, both options involve converting to JsValue before returning, they only differ in how. If anyone wants to build a solution using multiple return values, they are welcome to and it would probably be the best implementation by miles, but my own requirements include targeting an ancient version of Firefox, so I can't use that and really neither can anyone else yet.

I have built most of option 1, will probably PR separately so you can see it and weigh up whether a breaking change is worthwhile. Some general observations about them:

- With option 1, you'd be adding a lot of generated JS shims. It's quite awkward from a macro / codegen perspective. You can avoid some of these by hard coding a list of `syn::Type`a that are known to implement `Into<JsValue>`. That can't work for any other types eg RustStructs.
- Option 1's treatment of RustStruct in the glue code is difficult and I'm still working on that.
- Option 1 appears to imply the obsolescence of the ReturnWasmAbi trait, because the codegen module has to special case Results in order to generate the into-jsvalue shims, and you cannot do that inside a trait impl / default function. It would be lovely to have one shim per *type*, but this is a Rust macro, the best you can really do is one shim per exported function.
- With option 2, The resulting difference between the type bounds on plain returned values and result-wrapped ones is a bit of an unpleasant implementation detail / limitation. (But also, Result is never going to perfectly match JS exceptions.)
- Performance is likely a wash, both options and even the JsError solution all involve one more WASM -> JS call than the leaky version, only difference is amount of JS glue spewed out. Option 1 spews most, JsError (which like option 2, uses `Into<JsValue>`) spews least.

gthb:
👏 I wish more people explained their work so meticulously.

alexcrichton:
Ok thanks for the explanation! It's good to know that this doesn't break things. I realize that the previous design isn't great but I would prefer to not break existing users.

Naively though what I would expect is that this:

```rust
impl<T: IntoWasmAbi> ReturnWasmAbi for Result<T, JsValue>
```

would probably change to 

```rust
impl<T: IntoWasmAbi, U: Into<JsValue>> ReturnWasmAbi for Result<T, U>
```

and the "abi" would be something like:

```rust
#[repr(C)]
pub struct ResultAbi<T> {
    is_ok: u32,
    abi: ResultAbiUnion<T>,
}

#[repr(C)]
pub union ResultAbiUnion<T> {
    ok: T,
    err: u32,
}
```

where `ok` is determined by `T` and `err` is the index of the `JsValue` if present (and its what gets `throw`n). The changes you have here of altering `Describe for Result<T, U>` would stay where it becomes a more first-class thing that the generated glue handles (it'd read the result and use the integer discriminant to figure out what happened).

One of the things I'm worried about with the system you describe is something like `Result<JsError, JsError>` which isn't possible since it unconditionally gets thrown based on type information. I also think it would be best to allow the `T` of `Result<T, _>` to be anything as opposed to only other JS values.

cormacrelf:
That's an interesting idea, it pretty neatly addresses the shim problem and the RustStruct problem at the same time. Nicely done.

Regarding the edge cases with JsError, I think simply removing any custom implementation and relying on the generic one would resolve them. With the change to the constraints you mention there, JsError would be covered; it would then simply be a friendly helper type for creating Errors with no effect on ABI.

alexcrichton:
Ah yeah I was hoping that with an explicit discriminant of sorts we don't have to check types and that way `Result<JsError, JsError>` would work well.

I think if this works out we can update most documentation and examples to having `Result<T, js_sys::Error>` instead of `JsValue` since `JsValue` is rarely what you'd want anyway.

cormacrelf:
You reckon ditch the `JsError` entirely in favour of `js_sys::Error`? Move the convenient `From<T> where T: std::Error::Error` impl there? (Maybe it already has one.)

Also, as for custom JS error subclasses, with the error bound expanded to `Into<JsValue>` it becomes very easy to just implement that on your custom error enum and instantiate one there. That'll be really good too.

Final thing to address is async. `Result<_, JsValue>` is all over those definitions. Promise, async imports, async exports. Maybe the bound expansion could be applied elsewhere too. Probably best I have a crack implementing before I get carried away thinking about that, it'll be clearer once it's working.

alexcrichton:
Hm that's true, given how fundamental errors are I can see having them in `wasm-bindgen` itself would be quite useful. Ideally we could just move `js_sys::Error` into `wasm-bindgen` and then reexport it from `js-sys`, but I'm not sure if that would work out. I do think it would be best to avoid having `wasm_bindgen::JsError` and `js_sys::Error`, though, that sounds a little too confusing.

For `async` I would naively hope that the macro could have syntactic sugar which accepts `Result<T, U>` but all the runtime internals deal with `Result<T, JsValue>` or w/e they are today. I forget precisely how all the expansions work though and whether that is workable.

cormacrelf:
Is it correct that `JsValue` could implement `OptionIntoWasmAbi` by returning `0`? If you generate a rust struct, then it [implements it this way](https://github.com/rustwasm/wasm-bindgen/blob/e4477293c15df8500330b295cb9624dccb2a063b/crates/backend/src/codegen.rs#L662-L674). Is there a reason why `JsValue` doesn't already?

cormacrelf:
Alright, this works overall pretty well! Assuming that I'm right about JsValue/OptionIntoWasmAbi. But there's a bug in the existing stack return reading (`LoadRetptr`) that's been revealed.

https://github.com/cormacrelf/wasm-bindgen/blob/e4477293c15df8500330b295cb9624dccb2a063b/crates/cli-support/src/wit/mod.rs#L1347-L1358

https://github.com/cormacrelf/wasm-bindgen/blob/e4477293c15df8500330b295cb9624dccb2a063b/crates/cli-support/src/js/binding.rs#L598-L611

This gives you:

```rust
#[wasm_bindgen]
fn return_option_err() -> Result<Option<f64>, JsError> { ... }
```

```js
module.exports.return_option_err = function() {
    try {
        const retptr = wasm.__wbindgen_add_to_stack_pointer(-32);
        wasm.return_option_err(retptr);
        var r0 = getInt32Memory0()[retptr / 4 + 0];
        var r1 = getFloat64Memory0()[retptr / 8 + 1];
        var r2 = getInt32Memory0()[retptr / 4 + 2];
        var err0 = r2;
        if (err0 !== 0) { throw takeObject(err0); }
        return r0 === 0 ? undefined : r1;
    } finally {
        wasm.__wbindgen_add_to_stack_pointer(32);
    }
};
```

Those offsets are nonsensical. We're going to need a miniature version of the C struct packing algorithm.

cormacrelf:
Ah, I will flip it back to use a separate discriminant then.

> the discriminant could be forced to a u64 to force 8-byte alignment of the rest of the fields

I chose to place the JsValue's u32 at the end and reduce the problem of getting its value off the stack down to `pop()`, so that was the field being misread. This turned out to simplify the struct unpacking too. The simple field aligner/unpacker I implemented can handle WasmAbi structs that satisfy:

- repr(C) of course
- u32/i32/f32/f64 leaf fields only of course
- layout equivalent to a completely flattened struct, constructed by a depth first in order traversal of all the leaf fields in it. (This means that you can't embed `struct A(u32, f64)` as `struct B(u32, A);` because the "completely flattened" `struct AB(u32, u32, f64)` would miss the 4 byte padding within B and then as a consequence also miss the 4 byte padding within A that repr(C) inserts.)

The type in question (not having fixed the JsValue zero thing yet) is `ResultAbi<OptionalF64>`, which boils down to `{ { u32, f64 }, u32 }` (in particular, not `{ u32, { u32, f64 } }`) and can therefore be unpacked by the limited implementation. If there are no WasmAbis that don't have this property, then this would be technically good enough, but it should at least be documented on the WasmAbi trait for later reference. I haven't spotted any counterexamples so far. If there aren't, that is probably due in part to the limited recursion (eg Option can't contain more Options, wasm-bindgen won't do it.)

I agree full-on struct packing is overboard. We only need to generate aligned glue code for WasmAbis we know about, so the instruction builder has perfect knowledge of where the fields should be, and it would be unnecessary (however cool!) to expose this via eg WasmDescribe. All you need to do is communicate this information from the instruction builder to the stack reader, for each field you know needs tweaking. I don't think you would need to actually use a u64, for instance, just annotate the adapter types.

- You could sprinkle some `AdapterType::Padding` and simply advance the offset in the unpacker without emitting a LoadRetptr 
- Similarly `AdapterType::AlignTo(usize)` to align the next field read, slightly less thinking required than padding.

Review the use of f64 within other structures (basically Option), a bit of fiddling with the sanity checks on inputs.len or whatever, and I think that would do it.

alexcrichton:
Hm so another option is something like `#[repr(packed)]` on the Rust side of things so there's never any padding and no alignment issues, it's just one field to the next. Would that help the decoding side?

I also think it's fine to place any field wherever, what you're doing already it seems for whatever's easiest feels like the best way to go to me.

cormacrelf:
Nah, the Float64Array view of the wasm memory can only do aligned reads. You'd have copy from a UInt32Array view into a buffer and reinterpret as f64.

alexcrichton:
Oh I since learned awhile back that there's `DataView` which supports unaligned reads which the bindings could switch to using instead of various views with typed arrays (only needs one `DataView` for all reads as opposed to a view-per-type)

cormacrelf:
I looked into DataView, and it just touches too much. It may be worth doing if you add a lot more structs that need this, but for the ones we have now, it's too much work. If anyone finds some exotic alignment needs, they have this treasure trove of a thread to refer to :)

I've undone the OptionIntoWasmAbi thing, using a discriminant instead. I've fixed the UI tests (thanks to whoever got github actions going again) and added some docs based on one of my comments above. There's a test using `T: Into<JsValue>` in an outgoing `async fn` signature. Most of the above is a rebase over main.

Is there anything else I've forgotten?

cormacrelf:
This last thing, I now have absolutely no idea what I'm doing, but:

1. takeObject is now exposed, but this failed when you had externref enabled in one of the cli reference tests (anyref-import-catch) since exposing takeObject fails when it is
2. I basically copied the way `Instruction::I32FromOptionExternref` does the `addHeapObject`/`addToExternrefTable` switching, because I am a robot coder and that looked like it would work
3. it did, and I added a `WASM_BINDGEN_EXTERNREF=1 NODE_ARGS="--experimental-wasm-reftypes"` version of the test suite to CI

crates/cli-support/src/wit/outgoing.rs
```diff
@@ -337,6 +345,88 @@ impl InstructionBuilder<'_, '_> {
...
+                self._outgoing(arg)?;
+                // then push an instruction just before that one with two extra inputs
+                self.instruction_unwrap();
+                // then we end up with
+                //
+                //     LoadRetptr [f64, u32, u32, u32, u32]
+                //     UnwrapResult { popped: 2 } (the top two u32s)
+                //     Decode { popped: 3 } (f64, u32, u32)
+                //
+                // and the code for UnwrapResult is simple.
```
alexcrichton, line 407: This block here feels a bit brittle, can you add some more asserts here that everything is the shape as expected? For example this looks like it assumes that `self._outgoing(arg)` pushes 2 instructions unconditionally, so could this assert that 2 instructions are indeed pushed and the first one pushed is `LoadRetptr`?

crates/cli-support/src/js/mod.rs
```diff
@@ -2194,6 +2194,29 @@ impl<'a> Context<'a> {
...
+            "
+                function {}(idx) {{
+                    return wasm.{}.get(idx);
+                }}
+            ",
+            view, table,
+        ));
+
+        Ok(view)
+    }
```
alexcrichton, line 2218: I think it'd be fine to avoid an extra intrinsic function for this, I think it could just inline the `wasm.table.get(idx)` at each call-site using `self.export_name_of(table)`

crates/cli-support/src/js/binding.rs
```diff
@@ -783,6 +787,45 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
...
+                    dereference,
+                    i = i,
+                    j = j,
+                ));
+            } else {
+                js.cx.expose_take_object();
+                js.prelude(&format!(
+                    "
+                    if (is_ok{i} === 0) {{
+                        throw takeObject(err{j});
```
alexcrichton, line 820: I'm a bit confused on why one branch reaches into a table and one branch uses `takeObject`, presumably the branch that reaches into the table needs to also "deallocate" the slot in the table, right? I would naively expect both branches here to use `takeObject` or otherwise a slot of the table is leaked.

crates/cli-support/src/js/binding.rs
```diff
@@ -596,16 +596,20 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
...
             };
+            let size = quads * 4;
+            // Separate the offset and the scaled offset, because otherwise you don't guarantee
+            // that the variable names will be unique.
+            let scaled_offset = offset / quads;
             // If we're loading from the return pointer then we must have pushed
             // it earlier, and we always push the same value, so load that value
             // here
-            let expr = format!("{}()[retptr / {} + {}]", mem, size, offset);
+            let expr = format!("{}()[retptr / {} + {}]", mem, size, scaled_offset);
```
alexcrichton, line 612: Are the changes in this block still necessary?

crates/cli-support/src/js/binding.rs
alexcrichton: Aha ok rereading your comments I see you are also as stumped about this as I am. I think the bug is that we need to deallocate the index in the `externref` case. I believe that the `__externref_table_dealloc` intrinsic defined in `src/externref.rs` is what needs to be invoked after the table slot is loaded from. I imagine if you poke around the various bits and pieces of the JS glue code generation for other usages of that I think it should be relatively easy to pattern-match on what is already done and doing roughly the same thing here too.

alexcrichton (commented):
This is looking really good to me, I think the only semi-major thing is what you've pointed out, but otherwise I'm quite happy with how this is shaping up!

crates/cli-support/src/wit/outgoing.rs
cormacrelf: LoadRetptr should just be called the inputs list, it's prepended automatically later on. But you were right, this was assuming 1 instruction would be pushed, and String for example pushes 2 (one is a deferred free) so I managed to get Node to crash with SIGBUS freeing a null pointer because it popped the wrong values.

crates/cli-support/src/js/binding.rs
cormacrelf: Yes. The offsets placed in the generated LoadRetptr instructions have changed from `.enumerate()` indexes, into correct quad offsets. So before, Optionalf64 worked by a fluke, because the f64 LoadRetptr had offset=1 size=8, and the f64 field was aligned to 8 bytes, and that happened to line up. Now, the same situation gives you `offset=2`. So this code has to change at least a little bit, in order to get f64 LoadRetptrs to divide that quad offset by 2 to get an offset into the Float64Array.

The variable names uniqueness is the rest of it, we introduce `scaled_offset` to keep `offset` for use in the `var r{} = ...` below, as `offset` is unique to each LoadRetptr but `scaled_offset` is not (sometimes being divided by two).

crates/cli-support/src/js/binding.rs
cormacrelf: Yep, this worked nicely.

crates/cli-support/src/js/mod.rs
cormacrelf: We did end up using it for calling `drop` as well.

crates/cli-support/src/wit/outgoing.rs
cormacrelf: Oop, called that one too early.

cormacrelf: Oh no, this is another thing. That bit is fixed.

cormacrelf:
Alright there is ONE MORE.... consider this:

```js
module.exports.return_string_err = function() {
    try {
        const retptr = wasm.__wbindgen_add_to_stack_pointer(-16);
        wasm.return_string_err(retptr);
        var r0 = getInt32Memory0()[retptr / 4 + 0];
        var r1 = getInt32Memory0()[retptr / 4 + 1];
        var r2 = getInt32Memory0()[retptr / 4 + 2];
        var r3 = getInt32Memory0()[retptr / 4 + 3];
        if (r3) {
            throw takeObject(r2);
        }
        return getStringFromWasm0(r0, r1);
    } finally {
        wasm.__wbindgen_add_to_stack_pointer(16);
        wasm.__wbindgen_free(r0, r1);
    }
};
```

If you `throw takeObject(r2);` you still hit the finally block, and it frees `r0, r1`, which is an undefined pointer/len combo as we didn't write to them at all. So I think we need either

- a conditional deferred call instruction, i.e. `if (r3) { wasm.__wbindgen_free(r0, r1); }`, or
- an equivalent intrinsic i.e. `wasm.__wbindgen_free_conditional(r3, r0, r1);`  to use with the normal DeferredCallCore.

cormacrelf:
Or actually, __wbindgen_free is fine to call with len=0. We can just set `r1 = 0`!

crates/cli-support/src/js/binding.rs
alexcrichton: Hm sorry so discuss this further, I don't really understand why the term "quad" is being used in wasm-bindgen. I think it would make more sense for everything to be byte-based instead of quad-based. I realize that wasm-bindgen only works with 4 and 8-byte things but I think it makes more sense to still use bytes internally instead of bytes some places and quads/etc other places

alexcrichton:
This looks fantastic to me, thanks again for your work on this! I've got one remaining nit about terminology of quads vs bytes, but that's pretty minor and can be fixed in a follow-up if truly necessary. Otherwise I'm gonna go ahead and merge this.

crates/cli-support/src/js/binding.rs
cormacrelf: Fair enough, it would be better with only bytes. I'll submit a separate PR, it requires modifying this function alongside the `StructUnpacker`. Fortunately doing so is only a code change + glue code variable name change (as opposed to an ABI change).

cormacrelf:
It was a pleasure working with you! If there are any issues related to this or other ABI composition problems, feel free to ping me.

(Also, you can probably also close #1742 now.)

SebastienGllmt:
This stack space leak has been a plague on so many projects for years. Thank you so much for helping to solve this issue. I am eternally grateful ❤

PR #2719, opened 2021-11-08
Adjust attributes for linters in generated code

tyranron:
This PR replaces `#[allow(clippy::all)]` usage in the generated code with `#[automatically_derived]` ~~+ `#[allow(clippy::nursery)]`~~.

## Motivation

If the project using `wasm-bindgen` hardens its rustc and Clippy lints having `#[allow(clippy::all)]` in the generated code is not enough and linter complains about many stuff, which cannot be supressed from the user code.

## Solution

`#[automatically_derived]` attribute allows rustc and Clippy to omit linting items with stylish lints and is intended exactly for that purpose.

~~Additionally, `#[allow(clippy::nursery)]` is used, because some lints from that group don't consider `#[automatically_derived]` attribute well in some edge cases, as are kinda unstable.~~   
UPD:  Considered to omit this.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -527,25 +527,30 @@ impl TryToTokens for ast::Export {
             quote! {}
         };
 
+        let generated_name_const =
+            Ident::new(&format!("{}__const", generated_name), generated_name.span());
         (quote! {
-            #[allow(nonstandard_style)]
-            #[allow(clippy::all, clippy::nursery, clippy::pedantic, clippy::restriction)]
-            #(#attrs)*
-            #[cfg_attr(
-                all(target_arch = "wasm32", not(target_os = "emscripten")),
-                export_name = #export_name,
-            )]
-            pub extern "C" fn #generated_name(#(#args),*) -> #projection::Abi {
-                #start_check
-                // Scope all local variables to be destroyed after we call the
-                // function to ensure that `#convert_ret`, if it panics, doesn't
-                // leak anything.
-                let #ret = {
-                    #(#arg_conversions)*
-                    #receiver(#(#converted_arguments),*)
-                };
-                #convert_ret
-            }
+            #[automatically_derived]
+            const #generated_name_const: () = {
+                #(#attrs)*
+                #[cfg_attr(
+                    all(target_arch = "wasm32", not(target_os = "emscripten")),
+                    export_name = #export_name,
+                )]
+                pub extern "C" fn #generated_name(#(#args),*) -> #projection::Abi {
+                    #start_check
+                    // Scope all local variables to be destroyed after we call
+                    // the function to ensure that `#convert_ret`, if it panics,
+                    // doesn't leak anything.
+                    let #ret = {
+                        #(#arg_conversions)*
+                        #receiver(#(#converted_arguments),*)
+                    };
+                    #convert_ret
+                }
+
+                ()
+            };
         })
         .to_tokens(into);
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm so on one hand CI needs fixing (independent of this PR, it's just an ambient issue I haven't gotten around to fixin).

On the other hand if `#[allow(clippy::all)]` isn't working I'm tempted to simply remove clippy annotations. If there isn't a one-liner to turn off clippy lints for a block of code then I think it's probably too much hassle to maintain otherwise.

tyranron:
@alexcrichton 

> Hm so on one hand CI needs fixing (independent of this PR, it's just an ambient issue I haven't gotten around to fixin).

Does it has sense to migrate onto GitHub Actions?

> On the other hand if `#[allow(clippy::all)]` isn't working I'm tempted to simply remove clippy annotations. If there isn't a one-liner to turn off clippy lints for a block of code then I think it's probably too much hassle to maintain otherwise.

Actually, `#[automatically_derived]` is sort of this one-liner. I've used it a lot in many codegen crates and have had almost zero issues with it. The only exception I've described above, when `clippy::nursery` lints don't respect it correctly due to implementation issues. That's a very niche case though, so I don't see much problems to remove `#[allow(clippy::pedantic)]` and suppress additional possible complaints one-by-one __only if__ they will appear.

alexcrichton:
Yes I think the best solution would be to migrate to github actions at this point, I just haven't gotten around to doing it. The build-the-release-tarballs logic is somewhat nontrivial.

I ideally would do nothing related to clippy in wasm-bindgen. I personally feel like clippy just wastes my time otherwise trying to talk about it and get everyone else to work around clippy. If there's a one-liner that's better than the current one I guess this can change, otherwise I would be inclined to say this is a clippy problem.

tyranron:
@alexcrichton 

> If there's a one-liner that's better than the current one I guess this can change, otherwise I would be inclined to say this is a clippy problem.

As stated above: `#[automatically_derived]` is the one. Except Clippy it also considered by rustc lints, if someone hardens them (we actually do).

So, I'll remove `#[allow(clippy::nursery)]` then and hope this will be it.

> The build-the-release-tarballs logic is somewhat nontrivial.

I'll try to look tomorrow into it. If I'll be able, I'll send a PR with migration to GitHub Actions.

tyranron:
@alexcrichton done. Now there is only `#[automatically_derived]` used on the generated code. Adjusted PR description as well.

tyranron:
@alexcrichton regarding CI migration: I've looked into it and didn't see any major troubles/problems. It's only just the whole CI pipeline is somewhat quite big itself, so rewriting it carefully will consume some time. I'll try to schedule it for the next week for myself, and hope will come with a goof PR.

alexcrichton:
Ok thanks, and no the CI stuff is my problem so I'll try to get around to it today and get it done

alexcrichton:
Ok I've redone CI in https://github.com/rustwasm/wasm-bindgen/pull/2720 if you'd like to rebase.

tyranron:
@alexcrichton you're quick 😅  
Done.

alexcrichton:
I think that CI is failing, but additionally looking at the successful CI runs it looks like there ~hundreds of new warnings from generated code as well.

tyranron:
@alexcrichton meh... need to poke with it a little bit more.

tyranron:
@alexcrichton well, the problem was that `#[automatically_derived]` attribute, unfortunately, cannot be placed on bare fns complaining with a warning:
```
warning: `#[automatically_derived]` only has an effect on items
```

So, I've worked-around such places by enclosing `extern "C" fn`s  in a `const` block, which accepts `#[automatically_derived]` attribute.

But, unfortunately, there is the only single place, where it cannot be done, because the generated function may not be enclosed in an `impl` block and should be accessible by its name for the code outside. I left there `#[allow(clippy::*)]` attributes along with the comment explaining why. I think this way it shouldbn't be a burden, while resolves any issues with linters in user code.

alexcrichton:
Ok this seems fine, thanks for the PR

tyranron:
@alexcrichton can we hope for a patch release with thus in near future?

alexcrichton:
Personally I'm not inclined to make a release purely just for lints, but when other features crop up I don't mind doing a release.

PR #2727, opened 2021-11-27
Fix u64 and i64 in Safari 14

Firaenix:
This PR removes the use of BigUint64Array and BigInt64Array as Safari 14 does not support them.
I have also run all the u64 and i64 tests and they work as expected.


alexcrichton:
Thanks for the PR here on this, but I think that much of this can be simplified. It looks like this is having numbers travel through the u32 array but I don't think that's necessary if the numbers are already manually split. Additionally I don't think the u64 conversion "shims" should really be necessary any more.

Firaenix:
Sorry for the late response, life gets in the way sometimes.

I think you're right about the U32 Array, I can remove those.
In regards to the u64 conversion shims, I can remove those too, just need to do some testing.

Thanks for the feedback.

akramhussein:
Hi all - this PR looks very close to be doing done, is there anything I can help with here to get it over the line?

Liamolucko:
This got fixed in #3037.

PR #2728, opened 2021-11-27
Add list of current unstable APIs to docs

epompeii:
The motivation behind making this change is to increase the discoverability of unstable APIs.
The current list has been populated using the contents of `wasm-bindgen/crates/web-sys/webidls/unstable/`.
Going forward, MRs that add or remove unstable APIs will need to update this list accordingly.

### DIFF
--- a/guide/src/web-sys/unstable-apis.md
+++ b/guide/src/web-sys/unstable-apis.md
@@ -29,3 +29,21 @@ to do this. For example:
 ```bash
 RUSTFLAGS=--cfg=web_sys_unstable_apis cargo run
 ```
+
+## List of Current Unstable APIs
+- `Bluetooth`
+- `Clipboard`
+- `ImageCapture`
+- `QueuingStrategy`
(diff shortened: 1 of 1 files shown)


alexcrichton:
Is there a way this list could be auto-generated? I otherwise will surely forget to ever remind folks to update this when WebIDL changes.

epompeii:
Yeah, that would be best!
One option is to create an `mdbook` preprocessor that auto-generates the markdown list based on a given file path.
Or I could add a check in the `doc_book` CI step that validates the list against the contents of `wasm-bindgen/crates/web-sys/webidls/unstable/`. At the very least this would catch a missed update and prompt the appropriate change.

alexcrichton:
Either way is fine by me, either generating this during the doc build or validating on CI the list is correct. Whichever is easiest to compute seems fine by me!

daxpedda:
Automating this should be easily implementable in the Webidl generator.

Issue #2156, opened 2020-05-25
support newtype pattern in bindings 


ctaggart:
### Motivation

It would be more type-safe if we could wrap primitives using the [newtype pattern](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html). The API would then become:
``` rust
pub struct KeyCode(pub i32);

extern {
  # [ wasm_bindgen ( method , getter , js_name = keyCode ) ] 
  pub fn key_code(this: &IKeyboardEvent) -> KeyCode; 
}
```
Currently, `KeyCode` is defined as an alias and the enum values becomes constants in a separate type.

https://github.com/taggartsoftware/ts2rs/blob/6d357315c915758f1e5a3c9af2d7e94b203e9fab/monaco-editor/src/monaco_extern.rs#L219-L220

https://github.com/taggartsoftware/ts2rs/blob/6d357315c915758f1e5a3c9af2d7e94b203e9fab/monaco-editor/src/monaco_help.rs#L22-L171

### Proposed Solution

Is there way to see if it is a single value struct & wrap it accordingly?

### Alternatives

Leave as is, which is less type safe.

### Additional Context

This would be helpful for all TypeScript enums translated to wasm-bindgen. https://github.com/taggartsoftware/ts2rs/issues/4

Pauan:
The user API would look something like this:

```rust
#[wasm_bindgen(transparent)]
pub struct KeyCode(i32);
```

This would apply a `#[repr(transparent)]` and generate the various trait impls.

nasso:
Do we really need a `wasm_bindgen(transparent)` attribute if we already have `#[repr(transparent)]` (#1474)?

PR #2751, opened 2021-12-28
Add support for `#[repr(transparent)]` on exported structs

### LINKED ISSUE #1474: Respect #[repr(transparent)]
### Describe the Bug

I'm not 100% sure whether this is a bug or a feature request, but, in theory, wasm-bindgen ABI can be seen as any other, and so `repr(...)` should have an effect.

### Steps to Reproduce

1. Define a struct with a single field (or a single field and a zero-size type like `PhantomData`).
2. Add `#[repr(transparent)]` to signify that its ABI should be the same as the inner field.
2. Add `#[wasm_bindgen]` attribute to export struct to JS.

### Expected Behavior

Struct is represented on the JS side as a type alias to the inner field.

### Actual Behavior

A wrapper class is generated.

nasso:
This PR adds support for [`#[repr(transparent)]`][RFC] on exported types:

```rust
#[repr(transparent)]
#[wasm_bindgen]
pub struct TransparentU32(u32);

#[wasm_bindgen]
pub fn make_transparent_u32(value: u32) -> TransparentU32 {
    TransparentU32(value)
}

#[wasm_bindgen]
pub fn get_transparent_u32(value: TransparentU32) -> u32 {
    value.0
}
```

```js
const r = wasm.make_transparent_u32(42);
assert.strictEqual(r, 42);
assert.strictEqual(wasm.get_transparent_u32(r), 42);
assert.strictEqual(r + 1, 43);
assert.strictEqual(wasm.get_transparent_u32(r) + 1, 43);
assert.strictEqual(typeof r, "number");
```

See #2156 and #1474 for the motivation behind this.

- [x] Transparent primitives (`u8`, `u16`, `u32`, `f32`...)
- [x] Transitive transparency (`Transparent(Transparent(a))` == `a`)
- [x] Zero-sized field(s) after the only non-zero-sized field (`Transparent(u32, ())`)
- [ ] Zero-sized field(s) before the only non-zero-sized field (`Transparent((), u32)`)
- [ ] Transparent `Option<T>`, `Box<T>`, `Result<T, E>`...
- [ ] Transparent classes, enums...
- [ ] Typescript type definitions

Fixes #1474.

[RFC]: https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md


crates/macro-support/src/parser.rs
```diff
@@ -377,11 +377,17 @@ impl<'a> ConvertToAst<BindgenAttrs> for &'a mut syn::ItemStruct {
             .unwrap_or(self.ident.to_string());
         let is_inspectable = attrs.inspectable().is_some();
         let getter_with_clone = attrs.getter_with_clone().is_some();
+        let transparent = self.attrs.iter().any(|attr| {
+            matches!(attr.style, AttrStyle::Outer)
+                && attr.path.is_ident("repr")
+                && attr.tokens.to_string() == "(transparent)"
```
alexcrichton, line 383: I think that this match should be done with something more structured than calling `.to_string` on the tokens

crates/macro-support/src/parser.rs
```diff
@@ -377,11 +377,17 @@ impl<'a> ConvertToAst<BindgenAttrs> for &'a mut syn::ItemStruct {
...
+        });
         for (i, field) in self.fields.iter_mut().enumerate() {
-            match field.vis {
-                syn::Visibility::Public(..) => {}
-                _ => continue,
+            let public = matches!(field.vis, syn::Visibility::Public(..));
+            if !transparent && !public {
```
alexcrichton, line 387: Can you add a comment for why `transparent` is factored in here?

crates/backend/src/codegen.rs
```diff
@@ -146,131 +146,180 @@ impl ToTokens for ast::Struct {
...
+                impl wasm_bindgen::convert::IntoWasmAbi for #name {
+                    type Abi = u32;
 
-                fn into_abi(self) -> u32 {
-                    use wasm_bindgen::__rt::std::boxed::Box;
-                    use wasm_bindgen::__rt::WasmRefCell;
-                    Box::into_raw(Box::new(WasmRefCell::new(self))) as u32
+                    fn into_abi(self) -> u32 {
+                        // SAFETY: repr(transparent) means, by definition, that we can transmute
+                        let value: #field_ty = unsafe { std::mem::transmute(self) };
```
alexcrichton, line 166: Instead of using `transmute` can this construct an instance of the type directly? (same for accessing the internal value)

crates/backend/src/codegen.rs
```diff
@@ -146,131 +146,180 @@ impl ToTokens for ast::Struct {
...
-                    #(inform(#name_chars);)*
                 }
-            }
 
-            #[automatically_derived]
-            impl wasm_bindgen::convert::IntoWasmAbi for #name {
-                type Abi = u32;
+                #[automatically_derived]
+                impl wasm_bindgen::convert::IntoWasmAbi for #name {
+                    type Abi = u32;
```
alexcrichton, line 162: I don't think that `u32` is right here, instead I think it should be `<#field_ty as IntoWasmAbi>::Abi` (and similar for other occurrences below)

alexcrichton (commented):
Thanks for this! One thing that makes me hesitant though is that newtypes in Rust often have extra invariants over the underlying type. For example you could imagine `struct char(u32)` but only some `u32` value are valid for `char`. By using this implementation it forces structs like `char` to consider everything valid because we can't control what JS gives Rust.

crates/backend/src/codegen.rs
nasso: That's what I did first, but this becomes complicated when the type has one or more zero-sized fields. I can't really think of a simple way to construct any type without bloating `ast::Struct`, but maybe I'm missing something?

crates/macro-support/src/parser.rs
nasso: Yeah, it might not stay like this, because I'm not sure if this is indeed the best place to handle the `transparent` case. [I mentioned it in `#wg-wasm` on Discord](https://discord.com/channels/442252698964721669/443151097398296587/925143307368157205):

> [...] im not sure if i should just add a `transparent: bool` field to `ast::Struct` or make a new type to represent anything that's `#[repr(transparent)]`
> the implementation of `ToTokens` for `ast::Struct` would be quite different for transparent types, im not sure if branching there is a good idea

nasso:
How does that usually work across the FFI? Is it a non-issue because the FFI is `unsafe` anyway?

alexcrichton:
Yeah I would imagine that most other FFI is already "unsafe enough" that this concern doesn't come up. That being said I also don't personally see this much with FFI since the goal isn't to make it as nice as possible but rather get something working first.

nasso:
One of the original motivations behind supporting `repr(transparent)` was just to provide a safer API on the JS/TS side (instead of manipulating the wrapped type directly). This could be achieved with some [Typescript magic](https://kubyshkin.name/posts/newtype-in-typescript/).

Another major motivation (much more important, imo) is to allow some types to be represented as JavaScript primitives to allow some operations such as `===` equality checks or comparisons to be performed more naturally. It also allows values to be serialised (sometimes necessary, e.g. when using the Drag and Drop API).

If the invariants introduced by some `repr(transparent)` types are a pain point, maybe a similar bindgen-specific attribute could be used to provide similar functionality? Another solution to explore would be to allow the user to define their own validation/conversion functions between the wrapped and the new type, but I'm not sure if this is desirable

crates/macro-support/src/parser.rs
```diff
@@ -377,11 +377,26 @@ impl<'a> ConvertToAst<BindgenAttrs> for &'a mut syn::ItemStruct {
...
+            Ok(Meta::List(MetaList { path, nested, .. })) => {
+                path.is_ident("repr")
+                    && nested.iter().any(|nested| match nested {
+                        NestedMeta::Meta(Meta::Path(path)) => path.is_ident("transparent"),
+                        _ => false,
+                    })
+            }
+            _ => false,
+        }) {
+            self.fields.iter().next().map(|field| field.ty.clone())
```
alexcrichton, line 390: Given that transparent structures can have multiple fields so long as only one has nonzero size, I don't think that this is the correct algorithm to find the type that we're being transparent over?

alexcrichton (commented):
I think a way to work around the unsafe transmutes is to do something like requiring `From` to be implemented perhaps?

nasso:
To me that only seems fair if `From` was already required when targeting platforms other than the Web

I don't know if forcing the user to write anything more than `#[repr(transparent)]` is a good idea. If `wasm-bindgen` needs more than that, I think it's better to have it as a separate bindgen attribute like `#[wasm_bindgen(transparent)]` or `#[wasm_bindgen(transparent = Foo)]`. This solution would solve the "which is the non-zero sized field?" problem by emulating type transparency at the bindgen level instead (with `From`/`Into` conversions at the FFI boundary).

crates/macro-support/src/parser.rs
nasso: Yeah I don't think this is either. But I don't think there's a way to know if a given field has a non-zero size.

Maybe we will have to force the user to implement a trait like `TransparentBind` so that we can refer to the type as `TransparentBind::NonZeroSizedFieldType` (with a more acceptable name of course haha)

But, just like the `From` implementation, I feel like this is a lot of extra work to get `#[repr(transparent)]` to compile when targeting the Web vs other platforms.

alexcrichton: That's true yeah, and I also feel like this is perhaps getting a bit too complicated. If this can't be done easily then I think it would be best to shelf the idea for now perhaps?

prideout:
This PR is quite old but it adds something valuable to wasm-bindgen that I would find quite useful.

Issue #2767, opened 2022-01-13
Allow `Option<&Closure<T>>` to be passed to JS


sinking-point:
### Motivation

Sometimes a JS function accepts an optional callback. In such cases it would be good to be able to import the function from JS with `Option<&Closure<T>>` as the argument type, so that if `None` is passed then the JS function would receive `undefined`.

E.g.

```rust
            #[wasm_bindgen(js_namespace = ["chrome", "storage", "sync"], js_name = set)]
            fn _set(data: JsValue, callback: Option<&Closure<dyn FnMut()>>);
```

### Proposed Solution

Implement `OptionIntoWasmAbi` for `&Closure<T>`. 

### Alternatives

There aren't really any viable alternatives that I can think of, but a workaround would be to import the function twice, e.g.

```rust
            #[wasm_bindgen(js_namespace = ["chrome", "storage", "sync"], js_name = set)]
            fn _set(data: JsValue);

            #[wasm_bindgen(js_namespace = ["chrome", "storage", "sync"], js_name = set)]
            fn _set_and_then(data: JsValue, callback: &Closure<dyn FnMut()>);
```

This only works if the callback is the final argument.

cataggar:
It looks like #2768 resolved this.

PR #2768, opened 2022-01-13
Implement OptionIntoWasmAbi for Closure references

sinking-point:
See https://github.com/rustwasm/wasm-bindgen/issues/2767

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -491,6 +491,15 @@ where
     }
 }
 
+impl<'a, T> OptionIntoWasmAbi for &'a Closure<T>
+where
+    T: WasmClosure + ?Sized,
+{
+    fn none() -> Self::Abi {
+        0
+    }
+}
+
 fn _check() {
     fn _assert<T: IntoWasmAbi>() {}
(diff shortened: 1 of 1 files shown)


alexcrichton (commented):
Can you add some tests where `None` is used? Otherwise I think it would be useful to add a test or double-check that use of the closure from JS after the function has returned throws a well-defined exception

sinking-point:
@alexcrichton Thanks for your feedback.

I meant to uncomment the `None` test before but it seems I forgot, so I've done that now and tightened it a bit too. Since there are no variations on `None`, I think this one test is sufficient.

Unlike `Fn`, `FnMut` and `FnOnce`, `Closure`s are not invalidated after the imported function returns. I have instead added a test for the behaviour after `.drop()`ping the `Closure`. I manually checked the message of the exception:

> Error: closure invoked recursively or destroyed already

This is as expected.

alexcrichton:
Ah yes, my mistake! Thanks!

sinking-point:
No worries. Thanks for merging.

Issue #2744, opened 2021-12-23
cant compile


tyoc213:
### Describe the Bug

Can't run hello_world example

### Steps to Reproduce

1. git clone 
2. `cd examples/hello_world`
3. `pm audit fix --force`
4. `npm run serve`
5. 

```
$ npm run serve

> serve
> webpack serve

🧐  Checking for wasm-pack...

✅  wasm-pack is installed at /home/tyoc213/.cargo/bin/wasm-pack. 

ℹ️  Compiling your crate in development mode...

<i> [webpack-dev-server] Project is running at:
<i> [webpack-dev-server] Loopback: http://localhost:8080/
<i> [webpack-dev-server] On Your Network (IPv4): http://192.168.1.149:8080/
<i> [webpack-dev-server] On Your Network (IPv6): http://[fe80::9a99:375:9c0d:783c]:8080/
<i> [webpack-dev-server] Content not from webpack is served from '/home/tyoc213/webassembly/wasm-bindgen/examples/hello_world/public' directory
[INFO]: Checking for the Wasm target...
[INFO]: Compiling to Wasm...
   Compiling hello_world v0.1.0 (/home/tyoc213/webassembly/wasm-bindgen/examples/hello_world)
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
[INFO]: Installing wasm-bindgen...
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', crates/cli-support/src/descriptor.rs:205:15
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Running the wasm-bindgen CLI
Caused by: failed to execute `wasm-bindgen`: exited with exit status: 101
  full command: "/home/tyoc213/.cache/.wasm-pack/wasm-bindgen-e63eccbd87203048/wasm-bindgen" "/home/tyoc213/webassembly/wasm-bindgen/target/wasm32-unknown-unknown/debug/hello_world.wasm" "--out-dir" "/home/tyoc213/webassembly/wasm-bindgen/examples/hello_world/pkg" "--typescript" "--target" "bundler" "--out-name" "index" "--debug"
asset index.js 243 KiB [emitted] (name: main)
asset index.html 232 bytes [emitted]
runtime modules 26.9 KiB 12 modules
modules by path ./node_modules/ 161 KiB
  modules by path ./node_modules/webpack-dev-server/client/ 56.8 KiB 12 modules
[...]

PR #2772, opened 2022-01-19
Version bump

mtKeller:
This is to resolve: https://github.com/rustwasm/wasm-bindgen/issues/2744
Just need to publish as the current version of wasm-bindgen is from 4 months ago


mtKeller:
test: build_examples - is not self referential as its installing the old wasm-bindgen from crates
see line 30 and compare with issue

Looking into the windows test now

alexcrichton:
Thanks for the PR! For version bumps though this requires running `./publish bump` after compilling the `publish.rs` script in this repo. Can you also add an entry to `CHANGELOG.md` for the release date?

mtKeller:
My pleasure, thank you all for your hard work.

MichealZwen:
wasm-bindgen maybe version-0.2.78 is incompatible with version-0.2.79, when I use asnycify for await/async.
but i cannot fix it. So I eventually went back to old version, everything becomes fine.

mtKeller:
@MichealZwen Can you provide a test case/error log?

MichealZwen:
@mtKeller Sorry, there's too much detail.
can you provide the update lists from 0.2.78 to 0.2.79?

Basically what I did was compile c/c ++ programs into WebAssembly via Emscripten and package them into static libraries. Finally rust links to the static library and binds javascript interfaces using wasm-bindgen.

As it turns out, the pure Rust version is fine. But c/c ++ programs using Asyncify failed the tescases in 0.2.79.

mtKeller:
I'll see about doing a test. Version bump was a release of all changes from September to current. So there is a bit a room for something to break

https://github.com/rustwasm/wasm-bindgen/compare/0.2.78...0.2.79

PR #2773, opened 2022-01-19
Fix missing wasm_bindgen attribute

abi:
The generated JS code in the following code snippet has a constructor. The corresponding Rust code did not have the attribute marking the `new` as the constructor.


alexcrichton:
Thanks!

PR #2788, opened 2022-02-03
Temporal api

### LINKED ISSUE #2783: Temporal API
[Temporal](https://github.com/tc39/proposal-temporal) is a stage 3 proposal for fixing date problems in JS. Would you accept PR for implementing it in `js_sys`, for example under `cfg(js_sys_unstable)` ?

HKalbasi:
fix #2783

Temporal is relatively big, so I want to check if I'm doing it correctly, before adding all of them. I currently only added `Temporal.PlainDate` and `Temporal.Duration`.

My specific questions:
* Are doc comments ok? I cherry picked and copy pasted them from [official docs](https://tc39.es/proposal-temporal/docs/), so they are js-centeric and low quality for wasm bindgen propose. And they don't have mdn links because there are no mdn page yet. If comments are not deal breaker, they can be omitted and added when mdn pages become appear.
* Some functions pass their parameter into some type from's method. Should we use the type in rust side signature, or `JsValue` is prefered? And similarly some functions are weirdly polymorphic, should we limit the function to its meaningful usage and select proper type?
* I added (copy pasted) an entire temporal polyfill in the tests, is it fine? It can be removed when support lands in node, I think.
* I putted temporal module in its own file, is there any reason that all `js_sys` crate is in a single file?
* `static_method_of` doesn't accept `["Temporal", "Duration"]` and simple `Duration` doesn't work (it thinks `Duration` is in top level), so I added them with `js_namespace`. From [docs](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-js-imports/static_method_of.html) functions added in this way should go under `Temporal`, but they are under their own type. I believe some action is needed here, but I don't know what action.
* There is a new `js_sys_unstable_apis` cfg, so some change in CI testing is probably needed. Or we should go with the same `web_sys_unstable_apis`?

### DIFF
--- a/crates/js-sys/tests/wasm/Temporal.js
+++ b/crates/js-sys/tests/wasm/Temporal.js
@@ -1,0 +1,3 @@
+exports.is_temporal_supported = function () {
+    return typeof Temporal === 'object';
+};
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -161,6 +161,9 @@ jobs:
         node-version: '16'
     - uses: ./.github/actions/setup-geckodriver
     - run: cargo test -p js-sys --target wasm32-unknown-unknown
+    - run: cargo test -p js-sys --target wasm32-unknown-unknown
+      env:
+        RUSTFLAGS: --cfg=js_sys_unstable_apis
 
   test_webidl:
     name: "Run wasm-bindgen-webidl crate tests"
--- a/crates/js-sys/src/Temporal.rs
+++ b/crates/js-sys/src/Temporal.rs
@@ -1,0 +1,1057 @@
+use super::*;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_namespace = Temporal, extends = Object)]
+    #[derive(Clone, Debug)]
+    pub type Instant;
+
+    #[wasm_bindgen(static_method_of = Instant, js_namespace = Temporal)]
+    pub fn from(val: &JsValue) -> Instant;
+
+    #[wasm_bindgen(static_method_of = Instant, js_namespace = Temporal)]
+    pub fn compare(one: &Instant, two: &Instant) -> i32;
+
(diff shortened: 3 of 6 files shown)


crates/js-sys/tests/wasm/Temporal.js
```diff
@@ -0,0 +1,9035 @@
+exports.is_temporal_supported = function () {
```
alexcrichton: Is this a polyfill of `Temporal`? If so I think it's fine to leave this out and have the tests be opt-in and used only if `Temporal` actually exists (or disabled at compile-time by default or something like that)

alexcrichton (commented):
> Are doc comments ok? 

Yeah they're fine, we can always refine over time as necessary.

> Some functions pass their parameter into some type from's method. 

I don't know anything about `Temporal` or the JS API, so I don't think I'll be of much help on the API-design side of things here. This is all unstable though so it seems fine to add and refine over time

>    I added (copy pasted) an entire temporal polyfill in the tests, is it fine?

I commented below too but I think it'd be best to leave this out for now.
>  I putted temporal module in its own file, is there any reason that all js_sys crate is in a single file?

Yeah that's ok!

> static_method_of doesn't accept ["Temporal", "Duration"] ...

Sorry I don't really know what you mean here. Again though I know nothing of the temporal API in JS so I may not be of much help here in general. (tbh most of the `js_namespace` and `static_method_of` interactions are all long paged out of my memory as well)

> There is a new js_sys_unstable_apis cfg, so some change in CI testing is probably needed. 

Yeah testing this in some form or another on CI seems good. If it's too hard though that's fine, it's unstable and can get figured out later.

HKalbasi:
It should be complete now if I didn't miss anything. I removed the polyfill (so tests are now no-op because no platform currently implemented it) and my problem with `static_method_of` is solved (it accepts an additional `js_namespace`)

alexcrichton:
Thanks!

PR #2797, opened 2022-02-09
Fix deno lint warnings

egfx-notifications:
This PR modifies the JS snippets generated by `wasm-bindgen` to satisfy a linting check with `deno lint`.

Please note that I only adjusted code snippets which were actually in use in my projects, there are more candidates in the source files. I want to see if you would be interested in merging these changes before putting in more effort than necessary for my current needs.

I also did not check if those changes are compatible with targets other than Deno, though I expect this should be fine as `const` is already in use in other parts of the bindings generation. I did run the tests for the crate, though.


alexcrichton:

PR #2832, opened 2022-03-07
Reference types fixes

### LINKED ISSUE #2830: `wasm-bindgen` sometimes creates duplicate exports with reference types enabled
### Describe the Bug

With reference types enabled, `wasm-bindgen` sometimes exports multiple shims with the same name if there are two closures with the same number of arguments but with different types.

### Steps to Reproduce

1. Compile the following code for wasm:

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern "C" {
    // For some reason the error doesn't occur without a `typescript_type`.
    #[wasm_bindgen(typescript_type = "foo")]
    type Foo;

    type Bar;
}

fn main() {
    Closure::wrap(Box::new(|_: Foo| {}) as Box<dyn FnMut(Foo)>);
    Closure::wrap(Box::new(|_: Bar| {}) as Box<dyn FnMut(Bar)>);
}
```

2. Run `wasm-bindgen` on it with `--reference-types` enabled
3. Run the resulting wasm through some kind of validator (e.g. `wasm-validate`)

If applicable, add a link to a test case (as a zip file or link to a repository we can clone).

### Expected Behavior

The result should be valid WebAseembly.

### Actual Behavior

You get an error something like this:

```
error: duplicate export "closure7_externref_shim"
```

### Additional Context

This bug is triggered by `winit`.

### LINKED ISSUE #2860: Panic on assertion failed when using `--reference-types`
### Describe the Bug

`wasm-bindgen` v0.2.80 panics when using `--reference-types` option.

### Steps to Reproduce

Having a simple code like this seems to trigger the panic:
```rust
#[wasm_bindgen(start)]
pub fn start() {
    std::panic::set_hook(Box::new(console_error_panic_hook::hook));
}
```

### Expected Behavior

`wasm-bindgen` succeeds.

### Actual Behavior

`wasm-bindgen` panics with the following error:
```
thread 'main' panicked at 'assertion failed: !self.config.externref', ~/.cargo/registry/src/github.com-1ecc6299db9ec823/wasm-bindgen-cli-support-0.2.80/src/js/mod.rs:986:9
```

### Additional Context

```sh
wasm-bindgen --target web --no-typescript --reference-types --omit-default-module-path --debug --keep-debug --out-dir out --out-name index file.wasm
```

The command succeeds when I remove the `--reference-types` option.

Liamolucko:
Fixes #2830, as well as some other bugs I found while working on it.
See the commit messages for more detailed descriptions of each fix.
Edit: Also fixes #2860.

### DIFF
--- a/crates/cli-support/src/descriptors.rs
+++ b/crates/cli-support/src/descriptors.rs
@@ -22,7 +22,6 @@ use wasm_bindgen_wasm_interpreter::Interpreter;
 pub struct WasmBindgenDescriptorsSection {
     pub descriptors: HashMap<String, Descriptor>,
     pub closure_imports: HashMap<ImportId, Closure>,
-    cached_closures: HashMap<Descriptor, FunctionId>,
 }
 
 pub type WasmBindgenDescriptorsSectionId = TypedCustomSectionId<WasmBindgenDescriptorsSection>;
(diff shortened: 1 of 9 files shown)


Dragonink:
Also fixes #2860.

alexcrichton:
Sorry it took me so long to get to this, but this all seems reasonable enough to me!

Unfortunately the `--reference-types` argument was added pretty early in wasm-bindgen's history and I never went back and re-tested it and got it fully working with 100% of features, so it's great to see that it's getting some usage and bugs are getting fixed!

PR #2835, opened 2022-03-10
fixed `UnwrapThrowExt` formatting to align with `core`

### LINKED ISSUE #2732: `UnwrapThrowExt` should propagate the error for `Result<T, E>`
### Describe the Bug

I was debating between creating this issue as a feature request or as a bug, but I do believe it is a bug, as it is not what I would believe to be the intended behavior.

Currently, when calling `unwrap_throw()` or `expect_throw()` on an `Err(E)` variant, where `E: Debug`, only the message is printed, and not the debug value of `E`, as it does for the regular `expect()`/`unwrap()` counterparts. This makes the thrown error much less usable in diagnostics, especially since the `unwrap_throw` does not provide the `std::panic:Location` to hint as to what went wrong and where.

If this is, perhaps, a security concern, a feature flag might make this functionality available, and keep the old behavior when this flag is not set. Although, I am behind the idea of mimicking the original functions behavior when it comes to the thrown message.

### Steps to Reproduce

1. Try the following code:
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen(start)]
fn start() {
  let res: Result<(), String> = Err("This message should be displayed when thrown".to_string());
  
  res.expect_throw("Throwing error...");
  // or
  res.unwrap_throw();
}
```

### Expected Behavior

The `unwrap_throw()`/`expect_throw()` trait methods should mimic the behavior of the `core`/`std` library, displaying a debug view of the contained error value.

### Actual Behavior

The `expect_throw()` method only displays the message in the thrown error, and neither `unwrap_throw()` nor `expect_throw()` display a `std::panic::Location` to trace the origin of the thrown error.

jquesada2016:
This pull request closes #2732

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,8 @@
 #![doc(html_root_url = "https://docs.rs/wasm-bindgen/0.2")]
 #![cfg_attr(feature = "nightly", feature(unsize))]
 
+extern crate alloc;
+
 use core::convert::TryFrom;
 use core::fmt;
 use core::marker;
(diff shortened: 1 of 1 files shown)


alexcrichton:
I believe it was originally intended that the `E` was ignored since the purpose of using these methods was to avoid the panicking infrastructure and have a lightweight cost in terms of code size. Additionally these methods are also somewhat broken where destructors for values on the stack are not run yet the instance can still be reentered. This isn't actually memory safe in Rust and is a long outstanding but in wasm-bindgen.

jquesada2016:
I do get the desire to shave off the baggage that comes with the `panicking` infrastructure, but it is very useful and makes debugging much more painful then it needs to be.

How about feature gating this?

Also, since the `E` is ignored, it would also be prudent to drop the `E: Debug` constraint.

PR #2840, opened 2022-03-19
Added WebCodecs API

Sixshaman:
https://www.w3.org/TR/webcodecs/

### DIFF
--- a/crates/web-sys/src/features/mod.rs
+++ b/crates/web-sys/src/features/mod.rs
@@ -76,6 +76,12 @@ mod gen_AllowedUsbDevice;
 #[cfg(feature = "AllowedUsbDevice")]
 pub use gen_AllowedUsbDevice::*;
 
+#[cfg(feature = "AlphaOption")]
+#[allow(non_snake_case)]
+mod gen_AlphaOption;
+#[cfg(feature = "AlphaOption")]
+pub use gen_AlphaOption::*;
+
 #[cfg(feature = "AnalyserNode")]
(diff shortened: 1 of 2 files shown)


Pascualex:
Hi! Thanks for your contribution. 

I've tried to use this API but I've noticed that the `new()` function is missing for the `VideoEncoder` and `VideoDecoder` (and probably more structs).

Am I missing something? Is there an alternative way to construct these?

Sixshaman:
> I've tried to use this API but I've noticed that the `new()` function is missing for the `VideoEncoder` and `VideoDecoder` (and probably more structs).
It's a long problem in wasm-bindgen - most of the newer standards use the new syntax for constructors, and the Rust code generator doesn't support it, The support was added in #2842, but it's not merged yet. If you really need to test it right now, you can fork the branch and generate the bindings yourself.

Pascualex:
Thanks a lot for the explanation!

maxcountryman:
@Sixshaman how did you go about generating the feature flags for this? I tried to add the share API a few weeks back, but was unable to figure out how to auto generate the proper feature flags.

Sixshaman:
> @Sixshaman how did you go about generating the feature flags for this? I tried to add the share API a few weeks back, but was unable to figure out how to auto generate the proper feature flags.

When you run `cargo run --release --package wasm-bindgen-webidl -- webidls src/features`, the flags get written to the "src/features" file. Simply copy them to Cargo.toml afterwards.

PR #2852, opened 2022-04-04
Mark some set_* fns as potentially throwing.

derekdreery:
Make attributes marked `[SetterThrows]` use a `set_*` function that returns a `Result`. Currently webidl only looks for a `[Throws]` attribute.

I found this problem while writing an indexed DB wrapper. The function `web_sys::IdbObjectStore::set_name` should return a `Result`, because setting the name can go wrong in a number of ways. When I looked at the IDL files, the attribute had a `[SetterThrows]` mark rather than the more general `[Throws]`.


derekdreery:
Ready for review

derekdreery:
It is a breaking change, but currently the generated code is incorrect
(some attribute setters can throw and agent caught by web_sys). Do you
allow breaking changes to correct errors?

alexcrichton:
Sorry I'm not merging breaking changes at this time because I'm not personally prepared to deal with the fallout. I wish I had a better answer for you but I do not at this time.

derekdreery:
Ok no worries shall I leave the pr open?

Issue #2806, opened 2022-02-15
Generated JS code can end up with (reserved) keywords as identifiers


y21:
It seems wasm-bindgen does not check that identifiers that are valid in Rust code might not be valid in JavaScript. 

Building a project containing a function like this:
```rs
use wasm_bindgen::prelude::wasm_bindgen;

#[wasm_bindgen]
pub fn test(class: u8) {}
```
... generates this piece of JS code:
```js
/**
* @param {number} class
*/
export function test(class) {
    wasm.test(class);
}
```
which is invalid (`class` is a keyword and can't be used for identifiers in JS)

### Steps to Reproduce
1. `wasm-pack new bug`
2. Paste the code above in src/lib.rs
3. `wasm-pack build`
4. Check `pkg/bug_bg.js` (filename depends on the project name)

### Expected Behavior
Prefix identifier with `_` so it can't clash with keywords (or mangle identifiers in some other way)

### Actual Behavior
The identifier is not changed and clashes with the JavaScript "class" keyword

petrstudynka:
I am not sure whether generating idents with prefixes is the "right" approach. It breaks the Rust -> JS interface mapping contract.

I'd suggest issue compiler warnings once user uses idents that are keywords in js.

Nyrox:
I guess I will tack on to this issue, but a slightly more insidious version of the same problem is that names in your rust code can also *shadow* the same names in the generated JS code.

I just ran into this with a function with the signature:
`pub fn process_frame(mut wasm: WasmEngine, time: u64)`
which then generated this piece of JS shim:
```ts
export function process_frame(wasm, time) {
    _assertClass(wasm, WasmEngine);
    if (wasm.ptr === 0) {
        throw new Error('Attempt to use a moved value');
    }
    var ptr0 = wasm.ptr;
    wasm.ptr = 0;
    uint64CvtShim[0] = time;
    const low1 = u32CvtShim[0];
    const high1 = u32CvtShim[1];
    wasm.process_frame(ptr0, low1, high1);
}
```
which in quite hilarious fashion breaks, because the parameter called 'wasm' shadows the fairly important import: 
`import * as wasm from './[REDACTED]_wasm_bg.wasm';`
at the top of the file. :p

Ideally I guess all generated JS code should prefix their names with something, but some warnings for names like 'wasm' would be a good idea. 😂

daxpedda:
Duplicate of #2338.

PR #2855, opened 2022-04-06
#2806 bugfix: Added checks for JS keywords in signatures.

petrstudynka:
Added checks in macro crate for those cases downstream crate uses idents, which would result in invalid JS code generation.

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -15,6 +15,29 @@ use syn::spanned::Spanned;
 
 thread_local!(static ATTRS: AttributeParseState = Default::default());
 
+/// Javascript keywords which are not keywords in Rust.
+const JS_KEYWORDS: [&str; 20] = [
+    "class",
+    "case",
+    "catch",
+    "debugger",
+    "default",
+    "delete",
+    "export",
+    "extends",
+    "finally",
+    "function",
+    "import",
+    "instanceof",
+    "new",
+    "null",
+    "switch",
+    "this",
+    "throw",
+    "var",
+    "void",
+    "with",
+];
 #[derive(Default)]
 struct AttributeParseState {
     parsed: Cell<usize>,
@@ -462,6 +485,7 @@ impl<'a> ConvertToAst<(BindgenAttrs, &'a ast::ImportModule)> for syn::ForeignIte
             self.vis.clone(),
             false,
             None,
+            false,
         )?
         .0;
         let catch = opts.catch().is_some();
@@ -704,13 +728,19 @@ impl ConvertToAst<BindgenAttrs> for syn::ItemFn {
             self.vis,
             false,
             None,
+            false,
         )?;
         attrs.check_used()?;
         Ok(ret.0)
     }
 }
 
+pub(crate) fn is_js_keyword(keyword: &str) -> bool {
+    JS_KEYWORDS.contains(&keyword)
+}
+
 /// Construct a function (and gets the self type if appropriate) for our AST from a syn function.
+#[allow(clippy::too_many_arguments)]
 fn function_from_decl(
     decl_name: &syn::Ident,
     opts: &BindgenAttrs,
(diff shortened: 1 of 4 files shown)


crates/macro-support/src/parser.rs
```diff
@@ -9,12 +9,35 @@ use backend::Diagnostic;
...
+    "catch",
+    "debugger",
+    "default",
+    "delete",
+    "export",
+    "extends",
+    "finally",
+    "function",
+    "import",
+    "instaceof",
```
y21, line 31: typo: `instanceof`

crates/macro-support/src/parser.rs
```diff
@@ -9,12 +9,35 @@ use backend::Diagnostic;
...
 use shared;
-use syn;
 use syn::parse::{Parse, ParseStream, Result as SynResult};
 use syn::spanned::Spanned;
+use syn::{self, FnArg, Pat, Signature};
 
 thread_local!(static ATTRS: AttributeParseState = Default::default());
 
+/// Javascript keywords which are not keywords in Rust.
+const JS_KEYWORDS: [&str; 19] = [
```
y21, line 19: `null` is missing

y21, line 19: `null` is missing

y21:
Not sure if it's just a visual bug on my side or if it actually submitted the same review twice. No idea how that happened. Sorry for the spam if it did.

alexcrichton:
Personally I think this would be best solved in the JS code generation rather than by forbidding this in Rust. Would it be possible to move the fix there instead?

petrstudynka:
Sure, I will change that.

petrstudynka:
I found out those keywords do not conflict when there are in defined as method name. So thats the reason for the if-else in the last commit. Otherwise those idents are mangled with underscore.

alexcrichton:
Can some tests be added for this to ensure that it works?

alexcrichton:
Ok looks great, thanks!

crates/macro-support/src/parser.rs
```diff
@@ -754,11 +785,21 @@ fn function_from_decl(
         })
     };
 
+    let replace_colliding_arg = |i: &mut syn::PatType| {
+        if let syn::Pat::Ident(ref mut i) = *i.pat {
+            let ident = i.ident.to_string();
+            if is_js_keyword(ident.as_str()) {
+                i.ident = Ident::new(format!("_{}", ident).as_str(), i.ident.span());
```
WorldSEnder, line 792: This should not silently replace the name of exported and imported identifiers where they make a semantic difference. Accessing them by bracket instead of dot syntax sounds like a reasonable workaround. I.e. `wasm["class"]` instead of silently changing to `wasm._class` in

https://github.com/rustwasm/wasm-bindgen/pull/2855/files#diff-f7d340b325e0e9f30bd18d30036698d1edfece6974682dd3480457e4383806d3R22

crates/macro-support/src/parser.rs
```diff
@@ -784,21 +825,29 @@ fn function_from_decl(
...
+    let (name, name_span, renamed_via_js_name) = if let Some((js_name, js_name_span)) =
+        opts.js_name()
+    {
+        let kind = operation_kind(opts);
+        let prefix = match kind {
+            OperationKind::Setter(_) => "set_",
+            _ => "",
+        };
+        let name = if prefix.is_empty() && opts.method().is_none() && is_js_keyword(js_name) {
+            format!("_{}", js_name)
```
WorldSEnder, line 837: This should not silently replace the name of exported and imported identifiers where they make a semantic difference. Accessing them by bracket instead of dot syntax sounds like a reasonable workaround if the user actually does use a js keyword. I.e. `wasm["class"]` instead of silently changing to `wasm._class` against explicit annotation in

https://github.com/rustwasm/wasm-bindgen/pull/2855/files#diff-f7d340b325e0e9f30bd18d30036698d1edfece6974682dd3480457e4383806d3R17

daxpedda: That sounds like a reasonable solution that should be easy to implement, would you like to make a PR? Happy to review it.

PR #2867, opened 2022-04-15
Add note about using custom WebDrivers

thisisrandy:
This updates the docs to reflect the change made in rustwasm/wasm-pack#1127 and discussed in rustwasm/wasm-pack#1126.

### DIFF
--- a/guide/src/wasm-bindgen-test/browsers.md
+++ b/guide/src/wasm-bindgen-test/browsers.md
@@ -46,6 +46,10 @@ test` with the appropriate browser flags and `--headless`:
 wasm-pack test --headless --chrome --firefox --safari
 ```
 
+If you wish to use a custom WebDriver, e.g. to test an older version of a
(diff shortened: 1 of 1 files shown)


thisisrandy:
@alexcrichton this change actually isn't accurate until rustwasm/wasm-pack#1127 is merged. Could you take a look at that as well please?

thisisrandy:
I see. I saw your fingerprints on something over there and assumed, but I see now that I was mistaken. How would you prefer I handle this? It's a bit awkward to have two contingent PRs to different repos pending.

alexcrichton:
Sorry but I don't help out with wasm-pack any more so I don't know how to coordinate the two repos.

thisisrandy:
Okay then. I think what I'll do is wait for the wasm-pack PR to be resolved (it's been quite some time at this point, but hopefully someone looks at it eventually), and then resubmit this PR to update the docs. Thanks for your help.

Issue #2714, opened 2021-11-02
Question: How to test code that relies on Duck-typed interface?


vimmerru:
### Summary

Documentation suggests usage of Duck-typed interfaces

```rust
#[wasm_bindgen]
extern "C" {
    pub type Quacks;

    #[wasm_bindgen(structural, method)]
    pub fn quack(this: &Quacks) -> String;
}

/// Next, we can export a function that takes any object that quacks:
#[wasm_bindgen]
pub fn make_em_quack_to_this(duck: &Quacks) {
    let _s = duck.quack();
    // ...
}
```

But it is quite unobvious how to write unit test for `make_em_quack_to_this` function in Rust. Is there any way how to build instance of `Quacks` in Rust for passing to the tested function?

alexcrichton:
I think the "best" way to test this would be to write a Rust function that calls to JS which calls back into Rust's `make_em_quack_to_this` function.

vimmerru:
@alexcrichton What is good place to write js code in this case? JS snippets have problems with imports and nodejs target.

alexcrichton:
Ah unfortunately I don't think there's a good place. The tests in this crate itself do some wonky things with `NODE_PATH` to get things to work, but otherwise it's not super well supported.

Issue #1197, opened 2019-01-20
How to get more detailed type information in `.d.ts`


dlesl:
Hi!
At the moment it's possible to pass pretty much anything between Rust and js as a `JsValue`, or using  js-sys, however it would be nice to have a way to provide more detailed information in the .d.ts files.
For example, I can return an array of rust structs as `Box<[JsValue]>`, which in typescript becomes `any[]`. If I return `js_sys::Promise`, I get `Promise<any>`.

Would it be a good idea to add type parameters to the types in js-sys, so that you could return `Array<SomeStruct>` in rust? Of course, js arrays can contain mixed types, but that could still be represented by `Array<JsValue>`.

Would this be making things unnecessarily complex?

ThomasdenH:
Ideally an `extern "C"` type would generate a corresponding Typescript `interface`/`type` when it's used at code boundaries.

Michael-F-Bryan:
> Ideally an `extern "C"` type would generate a corresponding Typescript `interface`/`type` when it's used at code boundaries.

I was thinking the same. For example, my code accepts a `JavaScriptCallbacks` object with several methods, like so:

```rust
#[wasm_bindgen]
extern "C" {
    pub type JavaScriptCallbacks;

    #[wasm_bindgen(method)]
    fn unknown_content(this: &JavaScriptCallbacks, text: &str, span: Span);

    #[wasm_bindgen(method)]
    fn gcode_buffer_overflowed(
        this: &JavaScriptCallbacks,
        mnemonic: char,
        major_number: u32,
        minor_number: u32,
        span: Span,
    );

    ...
}

#[wasm_bindgen]
impl Parser {
    #[wasm_bindgen(constructor)]
    pub fn new(text: String, callbacks: JavaScriptCallbacks) -> Parser {
        ...
    }
}
```

This currently (as of v0.2.59) generates the following `index.d.ts`:

```ts
export class Parser {
  free(): void;
  constructor(text: string, callbacks: any);
}
```

However to prevent errors due to missing methods or bad signatures at runtime it'd be awesome if the type declarations also contained a `JavaScriptCallbacks` interface. `index.d.ts` might then look something like this:

```ts
export interface JavaScriptCallbacks {
  unknown_content(text: string, span: Span): void;

  gcode_buffer_overflowed(
    mnemonic: string,
    major_number: number,
    minor_number: number,
    span: Span,
  ): void;
}

export class Parser {
  free(): void;

[...]

PR #2871, opened 2022-04-20
Add support for traits

### LINKED ISSUE #2073: Proposal: Use Symbol for trait impls
### Motivation

Currently, `#[wasm_bindgen]` cannot be applied to trait impls, I guess this is to avoid method name collision? If that is the case, then JavaScript [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) is a perfect solution for this.

### Proposed Solution

Convert trait methods to JavaScript methods with symbols for names, and trait names can be converted to a namespace object of symbols. For example:

_This code:_

```rust
// src/lib.rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct Foo {
    value: i32,
}

#[wasm_bindgen]
impl Foo {
    #[wasm_bindgen(constructor)]
    pub fn new(value: i32) -> Self {
        Foo { value }
    }

    #[wasm_bindgen]
    pub fn read(&self) -> i32 {
        self.value
    }
}

#[wasm_bingen]
trait Read {
    fn read(&self) -> i32;
}

#[wasm_bindgen]
impl Read for Foo {
    #[wasm_bindgen]
    fn read(&self) -> i32 {
        self.value
    }
}
```

_can be converted to this TypeScript declaration:_

```typescript
export declare class Foo {
  public constructor(value: number);
  public read(this: Foo): number;

  // from Read::read in the Rust source
  public [Read.read](this: this): number;
}

// from Read trait in the Rust source
// namespaces in TypeScript are just objects in JavaScript
export declare namespace Read {
  // from Read::read in the Rust source
  const read: unique symbol;
}
```

ImUrX:
Tries to fix #2073 
I honestly don't know what I'm doing but I really want this and it sure will be a nice addition to the ecosystem.

- [ ] Correctly implement information struct of traits on the AST
- [ ] Extract information of trait with the proc-macro
- [ ] Generate typescript code for traits so you don't have a bad time developing
- [ ] Generate JS glue code correctly
- [ ] Make test unit

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -21,6 +21,8 @@ pub struct Program {
     pub enums: Vec<Enum>,
     /// rust structs
     pub structs: Vec<Struct>,
+    /// rust traits
+    pub traits: Vec<Trait>,
     /// custom typescript sections to be included in the definition file
     pub typescript_custom_sections: Vec<String>,
     /// Inline JS snippets
@@ -36,6 +38,7 @@ impl Program {
             && self.structs.is_empty()
             && self.typescript_custom_sections.is_empty()
             && self.inline_js.is_empty()
+            && self.traits.is_empty()
     }
 }
 
@@ -377,6 +380,42 @@ pub struct Variant {
     pub value: u32,
     /// The doc comments on this variant, if any
     pub comments: Vec<String>,
+}
+
+/// Information about a Trait being exported
+#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
+#[derive(Clone)]
+pub struct Trait {
+    /// The name of the trait in Rust code
+    pub rust_name: Ident,
+    /// The name of the trait in JS code
+    pub js_name: String,
+    /// All the methods of this trait to export
+    pub methods: Vec<TraitMethod>,
+    /// The doc comments on this trait, if provided
+    pub comments: Vec<String>,
+    /// Whether this trait is inspectable (provides toJSON/toString properties to JS)
+    pub is_inspectable: bool,
+    /// Whether to generate a typescript definition for this trait
+    pub generate_typescript: bool,
+}
+
+/// The method signature of a trait
+#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
+#[derive(Clone)]
+pub struct TraitMethod {
+    /// Comments extracted from the rust source.
+    pub comments: Vec<String>,
+    /// The rust method
+    pub function: Function,
+    /// The kind (static, named, regular)
+    pub method_kind: MethodKind,
+    /// The type of `self` (either `self`, `&self`, or `&mut self`)
+    pub method_self: Option<MethodSelf>,
+    /// The trait name, in Rust, this is attached to
+    pub rust_trait: Ident,
+    /// The name of the rust method on the rust side.
+    pub rust_name: Ident,
 }
 
 /// Unused, the type of an argument to / return from a function


ImUrX:
@alexcrichton sorry to bother but is there a way to generate JSDocs and ts signatures with the adapter types. I think I need to read a little more to understand how adapter type works

or maybe im going the wrong way

ImUrX:
Kinda related btw microsoft/TypeScript#10570

ImUrX:
> @alexcrichton sorry to bother but is there a way to generate JSDocs and ts signatures with the adapter types. I think I need to read a little more to understand how adapter type works
> 
> or maybe im going the wrong way

yeah this seems more complicated than I thought, is there no way to generate ts documentation without having code on the ``wit`` module? It really seems like the only way which means I need to modify the library more than I would like for generating the interface method signatures documentation...
@alexcrichton sorry for still bothering you, but what are your thoughts on how I should follow on this?

alexcrichton:
Sorry this is a feature which off the top of my head I don't really know how it would work or how it would be integrated. I unfortunately don't have the time to help out right now with deisgn as well.

ImUrX:
yeah understandable, this is currently a mess but it was quite interesting to do.
I'm gonna finish the implementation and then clean some code and make all needed tests and see what needs fixing

crates/backend/src/ast.rs
```diff
@@ -379,6 +382,44 @@ pub struct Variant {
...
+    /// The name of the trait in Rust code
+    pub rust_name: Ident,
+    /// The name of the trait in JS code
+    pub js_name: String,
+    /// All the methods of this trait to export
+    pub methods: Vec<TraitMethod>,
+    /// The doc comments on this trait, if provided
+    pub comments: Vec<String>,
+    /// Whether to generate a typescript definition for this trait
+    pub generate_typescript: bool,
```
stan-irl, line 398: is there any point to have this field? traits/interfaces only exist in TS

stan-irl:
this would honestly be awesome to have!!

Theres no way to guarantee API stability between TS/JS and rust right now because you cant have build time failure for JS type incompatibility. Its also really difficult to cover with tests because you cant force rust behaviour from TS/JS where the tests will be. Its very difficult to ensure that these codepaths get hit.

I cant believe this isn't a higher priority - how are teams working with this? Do they just deploy and wait for it to break?

felipellrocha:
@alexcrichton Any changes on this? I could really use the feature on our production system. Currently we have a huge, nasty workaround that I would love to get rid of.

stan-irl:
this PR would solve the following issues:
- https://github.com/rustwasm/wasm-bindgen/issues/2714
- https://github.com/rustwasm/wasm-bindgen/issues/1591
- https://github.com/rustwasm/wasm-bindgen/issues/1197

blockiosaurus:
Well my attempts to merge in 0.2.81 to this branch were unsuccessful, because I also don't know what I'm doing. Getting failures in the CLI on decode.

davidedellagiustina:
Quick question. Let's say I have a piece of code like the following:
```rust
#[wasm_bindgen]
pub struct A {}

#[wasm_bindgen]
impl A {
    pub fn a(&self) {
        // Do something
    }
}

pub trait Trait {
    fn b(&self);
}

impl Trait for A {
    fn b(&self) {
        // Do something else
    }
}
```

As far as I understand, this will only export a class `A` and its method `a()` in JavaScript, but I'd also like to call `A::b()` &ndash; is there a way to do this as of today?

ImUrX:
provide a wrapper method for calling it :P

davidedellagiustina:
@ImUrX ok, so that's the only solution for now. Thank you!

daxpedda:
I would be happy to review the Rust and JS part, but the TS part needs a different maintainer that has the time to do the design work and eventually review it.

ImUrX:
tbh the typescript part is easy, i just havent had the time and will to finish the PR, aside from the rust part of the typescript code being a lot of pain

crates/backend/src/ast.rs
ImUrX: Sorry i never answered this. You can use traits without typescript definitions. The proposed way of doing this was making the interface contain ``Symbol``s for each method. So when you want to run the function of a class implementing such trait you do
```js
struct[Trait.someMethodName]()
```

This is still the same in Typescript, because when you add a method of the same name on a class on Typescript, you override the method definition of the interface. Which is not normal behavior on rust

vdagonneau:
Hi, I'm interested in this feature, would you be ok if I took a stab at it using your PR as a base? I hope I'll be able to fix the last couple things.

ImUrX:
yeah, be free

pr1metine:
Hey, I'm also interested in this feature. Do you, @vdagonneau and everyone else, wanna team up and work on this PR?

haraldreingruber-dedalus:
This might also be related to this discussion: https://github.com/rustwasm/wasm-bindgen/discussions/3254

PR #2872, opened 2022-04-25
Added library to wasm-bindgen-cli

abrisco:
This may come off as an odd PR but would be something very beneficial to [bazelbuild/rules_rust::wasm_bindgen](https://github.com/bazelbuild/rules_rust/tree/0.3.0/wasm_bindgen) which are tools which use `wasm-bindgen` in the [Bazel](https://bazel.build/) build system. The functionality currently tries to recompile of the `wasm-bindgen-cli` crate but in order to achieve this, a ton of complexity was added to solve for issues where binary-only crates do not show up in `cargo metadata` outputs (https://github.com/google/cargo-raze/issues/218 has some good details on this issue). To avoid needing to maintain this complicated behavior, it would be very helpful to have `wasm-bindgen-cli` also contain a library target, which would allow it to appear in `cargo metadata` output. My hope is that this change has no noticeable impact to existing Cargo users but would be very helpful to Bazel users.

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen.rs
+++ b/crates/cli/src/bin/wasm-bindgen.rs
@@ -46,7 +46,7 @@ Options:
 ";
 
 #[derive(Debug, Deserialize)]
-struct Args {
+pub struct Args {
     flag_nodejs: bool,
     flag_browser: bool,
     flag_web: bool,
@@ -71,7 +71,7 @@ struct Args {
     arg_input: Option<PathBuf>,
 }
 
-fn main() {
+pub fn main() {
     env_logger::init();
     let args: Args = Docopt::new(USAGE)
         .and_then(|d| d.deserialize())
@@ -89,7 +89,7 @@ fn main() {
     process::exit(1);
 }
 
-fn rmain(args: &Args) -> Result<(), Error> {
+pub fn rmain(args: &Args) -> Result<(), Error> {
     let input = match args.arg_input {
         Some(ref s) => s,
         None => bail!("input file expected"),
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR, but I don't really understand the crate layout here so while this may work for Bazel it's a bit weird enough that at least for me I'd prefer to not maintain it. I don't understand the limitations of Bazel related to Cargo or the various other bits of tooling involved.

abrisco:
> Thanks for the PR, but I don't really understand the crate layout here so while this may work for Bazel it's a bit weird enough that at least for me I'd prefer to not maintain it. I don't understand the limitations of Bazel related to Cargo or the various other bits of tooling involved.

The only meaningful difference is that Bazel cannot consume binary-only crates. The functionality added to `lib.rs` doesn't need to be added but it'd help a lot if there was a library associated with the `cli` crate to make it accessible. Perhaps an empty `lib.rs` file could be added at the very least?

alexcrichton:
Sorry but "cannot consume binary-only crates" isn't really a description which would lead me naturally to a PR like this. I don't understand the consequences of Bazel's decision here and what that means for this crate. I'd rather not have a crate-specific workaround for what seems like a deficiency in Bazel's Rust integration.

hamza1311:
Also worth noting that including this crate as binary could lead to all sorts of confusing version conflicts. It'll be very hard to keep the versions of `wasm-bindgen` and `wasm-bindgen-cli` in sync

abrisco:
Thanks for taking a look!

PR #2875, opened 2022-04-28
Use conventional meta-item syntax for attributes

### LINKED ISSUE #2874: wasm_bindgen attribute shouldn't accept typescript_type without using it
### Describe the Bug

### Steps to Reproduce

```rust
#[wasm_bindgen]
pub struct MyStruct {
    hello: String,
}

#[wasm_bindgen]
impl MyStruct {
        todo!()
    }
}
```

### Expected Behavior

### Actual Behavior

TedDriggs:
# Disclaimers
1. I apologize for the diff size. The meaningful changes are in `crates/macro-support` - the other files are changed to avoid any commits that won't build.
2. This introduces a potentially-divisive change, using quotation marks for most key-value meta items. This has some advantages (such as working with rustfmt), but I know there are those in the community who really dislike the extra quotation marks. If wasm-bindgen has already decided not to use them, I'll close out this PR.
3. This is a major breaking change; almost all key-value meta items need updates. If people are receptive to this direction but concerned about a huge breaking change, I am happy to explore ways to temporarily support both formats.
4. I'm the author of `darling`, the attribute-parsing library this uses. There's probably a way to do this without `darling`, but it's surprisingly difficult to maintain the level of error quality without it.

# Commit message

Using the conventional form for attributes, which is compatible with rustfmt and syn's meta-item parsing. This change is breaking and substantial, but the necessary edits are largely mechanical.

With meta-items, it becomes possible to offload attribute parsing and validation to `darling`. This offload fixes #2874, as `darling` will automatically generate errors for unexpected meta-item names. This also makes it easier for people who are new to the codebase to understand how to add new options.


crates/backend/src/error.rs
```diff
@@ -30,6 +30,7 @@ enum Repr {
         text: String,
         span: Option<(Span, Span)>,
     },
+    Darling(TokenStream),
```
TedDriggs, line 33: This is not the right way to do this, but `darling` will likely need a change of some sort to fix this.

* There isn't an `impl ToTokens for darling::Error`, so we can't keep the error as-is in the `Repr` enum
* There's no way to read the span of a `darling::Error`, so we can't construct a `Single` from it (and therefore can't flatten them into a `Multi`)

crates/macro-support/src/d.rs
```diff
@@ -0,0 +1,383 @@
+//! Structs to receive attribute options for each supported syntax type.
```
TedDriggs, line 1: This module needs a different name.

crates/macro-support/src/d.rs
```diff
@@ -0,0 +1,383 @@
...
+        ($name:ident, $return:ty, $field:ident) => {
+            pub trait $name {
+                fn $field(&self) -> $return;
+            }
+        };
+    }
+
+    field_trait!(JsName, Option<&syn::LitStr>, js_name);
+    field_trait!(SkipTypescript, Flag, skip_typescript);
+}
```
TedDriggs, line 75: I tried - and failed - to find a good way to make these driven by a single macro. It'd be very doable with a proc-macro, but having a `proc-macro` crate that supports `macro-support` ...which is a proc-macro... felt too recursive.

crates/macro-support/src/d.rs
```diff
@@ -0,0 +1,383 @@
...
+        if opts.indexing_setter.is_present() {
+            operation_kind = ast::OperationKind::IndexingSetter;
+        }
+        if opts.indexing_deleter.is_present() {
+            operation_kind = ast::OperationKind::IndexingDeleter;
+        }
+
+        operation_kind
+    }
+}
```
TedDriggs, line 383: These shouldn't be copy-paste duplicated. A macro or intermediate trait would be better.

crates/macro/src/lib.rs
```diff
@@ -7,7 +7,8 @@ use quote::quote;
 
 #[proc_macro_attribute]
 pub fn wasm_bindgen(attr: TokenStream, input: TokenStream) -> TokenStream {
-    match wasm_bindgen_macro_support::expand(attr.into(), input.into()) {
+    let attr_args = syn::parse_macro_input!(attr as syn::AttributeArgs);
+    match wasm_bindgen_macro_support::expand(attr_args, input.into()) {
```
TedDriggs, line 11: There may be a way to construct an `AttributeArgs` from a `proc_macro2::TokenStream` but I couldn't get it to work. If that does exist, the `syn` dependency can be removed from this crate.

crates/web-sys/src/features/gen_AbortController.rs
```diff
@@ -3,7 +3,11 @@ use super::*;
 use wasm_bindgen::prelude::*;
 #[wasm_bindgen]
 extern "C" {
-    # [wasm_bindgen (extends = :: js_sys :: Object , js_name = AbortController , typescript_type = "AbortController")]
+    #[wasm_bindgen(
+        extends = "::js_sys::Object",
+        js_name = "AbortController",
+        typescript_type = "AbortController"
+    )]
```
TedDriggs, line 10: For anyone wondering how this diff is +55k, here's the answer; `rustfmt` now moves these to new lines.

crates/backend/src/error.rs
TedDriggs: The necessary change is now up as TedDriggs/darling#184.

TedDriggs: This has been updated, and no longer needs to change `backend::Error`'s internals.

alexcrichton:
Thanks for the PR here, but unfortunately this project isn't really in a place to manage or merge large breaking changes. I don't think that there's going to be a viable way to land this, even if the repo is fully updated, any time soon.

TedDriggs:
Makes sense: Getting that feedback was part of why I posted after 5 hours of work so I wouldn't be too deeply invested. A couple follow-up questions:

1. Is the change to having dedicated options structs per input-item-type interesting, if it can be done without changing the outward attribute API?
2. Is a non-breaking version of the API change - adding support for the quoted format _alongside_ the current format interesting, or would you prefer to stick to the generally-unquoted syntax long-term?

alexcrichton:
Sorry this is a big change I don't have the time to review, but overall making the proc-macro more readable or easier to understand/modify seems reasonable to me. Whether that could be achieved with the struct-per-input-item-type I'm not sure.

Personally I would prefer just one syntax. I'm not wed to with or without quotes, but I'd rather not have both.

TedDriggs:
> Sorry this is a big change I don't have the time to review

If I take out the diff parts caused by the breaking change, the new size is +554/-578; is that more tractable?

> Whether that could be achieved with the struct-per-input-item-type I'm not sure.

I have two ideas for this. Both of them start by adding `macro_support::Meta`, which is equivalent to `syn::Meta` except that it doesn't insist on the things to the right of `=` being valid literals.

# Option 1

1. Do low-level parsing using `syn::Parse` to turn the `attr` `TokenStream` into a `macro_support::Meta`.
2. (Somehow) convert each `macro_support::Meta` into a `syn::Meta` by wrapping values right of `=` in `LitStr`, preserving spans while doing so.
3. Pass these newly-created `syn::Meta` items to the derived `FromMeta` impls.

# Option 2
Create a proc-macro crate `macro-support-gen`, which exposes a derive-macro that's analogous to `darling::FromMeta` but operates on the `macro_support::Meta` type above. Then recreate some amount of `darling`'s meta-parsing loop in the `macro-support-gen` crate.

---

Option 1 is much easier and more maintainable, at the expense of traversing the attribute twice and sometimes doing more allocations.

TedDriggs:
Option 1 actually works! I'm going to update the review shortly.

TedDriggs:
I'm closing this in favor of #2880.

PR #2888, opened 2022-05-07
Fix some broken tests in `js-sys`

Liamolucko:
The first one created a date in UTC and then assumed that would be the same in the local time zone (as returned by `Date.prototype.toDateString`), which isn't always the case.

The second was calling `WebAssembly.instantiateStreaming` with a `Promise<Uint8Array>`, when it expects a `Response` or `Promise<Response>`; my guess as to why this wasn't failing in CI is that browsers only started enforcing it recently, and the one being used by CI doesn't.


alexcrichton:
👍

PR #2891, opened 2022-05-08
Make `Closure::new` work on stable

Liamolucko:
I've removed the nightly requirement by replacing `Unsize<T>` with a custom `IntoWasmClosure<T>` trait, implemented for `Fn` and `FnMut`.

I don't think this is a breaking change, since I don't think any stability guarantees are provided for nightly.

One thing I'm not sure about is whether `Closure::wrap` should be deprecated. It could still be kinda useful if you've already got a `Box<dyn Fn[Mut]>` and don't want to double-box it, but that seems like a rare case.


hamza1311 (approved):
This is a welcome improvement

alexcrichton:
Awesome, thanks for this!

Issue #2231, opened 2020-07-09
Downcast a JsValue to an exported struct


io12:
### Summary

In the code below, how can I cast the `JsValue` to `Foo`?

```rust
#[wasm_bindgen]
pub struct Foo(u8); // Assume this can't be serialized or deserialized

#[wasm_bindgen]
pub fn foo_downcast(foo: JsValue) {
    let foo: Foo = todo!();
    // ...
}
```

### Additional Details

`From<Foo> for JsValue` is implemented automatically, so I'd expect something like `TryInto<Foo> for JsValue` to be implemented too, but I couldn't find anything.

#### Reasons why you'd want to do this

1. Passing a vector or slice of objects to an exported function (currently `Vec<Foo>` and `&[Foo]` are unsupported, but `Vec<JsValue>` and `&[JsValue]` are supported, I think)
2. Having automatic casting rules for parameters of exported functions (automatically parsing strings if the parameter is a string, etc)

alexcrichton:
Ah yeah currently this isn't implemented. It would require an intrinsic of one form or another since the JS class lives in the JS file to test inheritance. This would be a nice feature to have though!

aweinstock314:
We've developed a workaround that discovers the runtime class name through `.__proto__.constructor.name`:

```
use wasm_bindgen::convert::FromWasmAbi;
pub fn generic_of_jsval<T: FromWasmAbi<Abi=u32>>(js: JsValue, classname: &str) -> Result<T, JsValue> {
    use js_sys::{Object, Reflect};
    let ctor_name = Object::get_prototype_of(&js).constructor().name();
    if ctor_name == classname {
        let ptr = Reflect::get(&js, &JsValue::from_str("ptr"))?;
        let ptr_u32: u32 = ptr.as_f64().ok_or(JsValue::NULL)? as u32;
        let foo = unsafe { T::from_abi(ptr_u32) };
        Ok(foo)
    } else {
        Err(JsValue::NULL)
    }
}
#[wasm_bindgen]
pub fn foo_of_jsval(js: JsValue) -> Option<Foo> {
    generic_of_jsval(js, "Foo").unwrap_or(None)
}
```

[...]

PR #2898, opened 2022-05-17
feat(jsvalue) Add `JsValue.ptr` to retrieve internal `idx` value

Hywan:
I'm trying to solve https://github.com/rustwasm/wasm-bindgen/issues/2231, and the first immediate step is to make `JsValue.idx` visible to the world. There is nothing particularly dangerous to allow that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,6 +130,11 @@ impl JsValue {
         }
     }
 
+    /// Returns the internal pointer value of this `JsValue`.
+    pub fn ptr(&self) -> u32 {
+        self.idx
+    }
+
     /// Creates a new JS value which is a string.
     ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
Personally I have always wanted to avoid this in wasm-bindgen if possible. This is a low-level implementation detail which is extremely hard to get right.

That being said it's been a long time since I originally felt that. I don't really mind that much any more, although I continue to feel that using this is highly likely to get yourself shot in the foot. 

If this is added, though, then I think it should be done "officially". For example if there's a getter there should also be a constructor. There should ideally be more than a sentence of documentation about what the index is and how to use it. Additionally I think it would be good to mention this in the book about perhaps the design rationale behind the indices and how things work today.

Hywan:
Closing in favor of #3088.

PR #2899, opened 2022-05-18
Fix memory getter checks when using shared memory

### LINKED ISSUE #2893: Broken memory handling in multithreaded Wasm
### Describe the Bug

Parallel Raytracing example seems broken.

### Steps to Reproduce

1. Go to [the hosted Raytracer example](https://wasm-bindgen.netlify.app/exbuild/raytrace-parallel/) or build it locally.
2. Click on 'Render'
3. See error in the DevTools console hinting at deserialization failing in `Scene::new`.

Liamolucko:
Fixes #2893

I opted to keep the current implementation when not using shared memory, and switch to the slightly more expensive equality check when it's enabled.

I've also added a CI job to run the browser tests with multithreading enabled, to hopefully catch stuff like this in future.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,6 +67,18 @@ jobs:
         WASM_BINDGEN_EXTERNREF: 1
         NODE_ARGS: --experimental-wasm-reftypes
 
+  test_threads:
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for this!

PR #2900, opened 2022-05-19
Cross platform make script using cargo-make for raytrace-parallel example

arnabanimesh:
Remove python dependency. Python server was working very slowly in local as compared to miniserve, hence changed it.


alexcrichton:
Thanks for the PR, but I would prefer to keep this example relatively simple and while Python isn't the fastest thing in the world it should suffice for most. This isn't necessarily "production ready", it's just meant to get the idea across.

PR #2903, opened 2022-05-19
Generate bindings for the OVR_multiview2 webgl extension

expenses:
IDL pulled from https://www.khronos.org/registry/webgl/extensions/OVR_multiview2/. Also see https://developer.mozilla.org/en-US/docs/Web/API/OVR_multiview2.


alexcrichton:
To confirm, is this a supported web api in browsers? Or is it an experimental feature that's not enabled everywhere yet? (if the latter this'll need to go into the `unstable` directory)

expenses:
> To confirm, is this a supported web api in browsers?

Yep! To quote https://github.com/gfx-rs/naga/pull/1933:

> The WebGL `OVR_multiview2` extension is implemented in:
>
> * Chrome on Windows, Linux and Android
> * Firefox on Android
>
> but NOT on Firefox on Linux afaik.

As reported by https://webglreport.com/?v=2

alexcrichton:
Ok sounds good, thanks!

PR #2905, opened 2022-05-20
Add runner headless support for msedge.

crlf0710:


hamza1311 (approved):
Thanks for the PR! Looks good to me

I'm wondering if there's any need to add the edge driver? Since edge uses chromium, using the chrome driver should produce the same results, right?

crlf0710:
Frankly i don't really know yet. I just followed MSEdge doc instructions and installed the msedgedriver, then discovered this runner doesn't support that natively. So i came up with this patch.

While msedge and chrome is both chromium, the version numbers become out of sync at times (Currently MSEdge stable is at 101, Chrome is at 102) So i figured installing msedgedriver v101 and chromedriver v102 might be less of a mess than installing chromedriver v101 and v102 at the same time to support the two installed browsers, even if that might work in theory :)

Would you like me to investigate more on whether i can get chromedriver working with my msedge installation?

PR #2906, opened 2022-05-20
Update the WebGPU IDL and regenerate web-sys

RSSchermer:
This is generated using an updated version of `wasm-bindgen-webidl` for which I'll submit a separate PR right after this (I want to be able to reference some of the generated code as an example, hence this ordering).

PR #2907, opened 2022-05-20
Namespace const

RSSchermer:
See https://github.com/rustwasm/weedle/pull/52

This comes with a companion PR https://github.com/rustwasm/wasm-bindgen/pull/2906 which contains new WebGPU bindings generated with these changes, which I figured might be useful for reference. See e.g. [gen_gpu_map_mode.rs](https://github.com/rustwasm/wasm-bindgen/pull/2906/files#diff-fb1d0139afd75f22c2519b2ebd529437297d775d7e33c6a91348282f12b2dacf) for an example of the code that is generated for namespaced constants.

This renames `InterfaceConst` and `InterfaceConstValue` to just `Const` and `ConstValue` and reuses those for namespace constants. It now also tracks API stability for namespaces, as the use-case at hand is WebGPU which is an unstable API. It currently adds the `#[cfg(web_sys_unstable_apis)]` attribute to both the namespace itself and the constants inside the namespace; I'm not sure if that's appropriate or redundant.

### DIFF
--- a/crates/webidl/Cargo.toml
+++ b/crates/webidl/Cargo.toml
@@ -21,7 +21,7 @@ proc-macro2 = "1.0"
 quote = '1.0'
 syn = { version = '1.0', features = ['full'] }
 wasm-bindgen-backend = { version = "=0.2.80", path = "../backend" }
-weedle = "0.12"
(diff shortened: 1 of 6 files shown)


alexcrichton:
Looks good, thanks!

Issue #2721, opened 2021-11-12
When using #[wasm_bindgen(readonly)], The type output to .d.ts file is void


madonoharu:
Hello.
Thank you for this great library!

### Describe the Bug

When `#[wasm_bindgen(readonly)]` is specified, the return value in d.ts file is `void`.

```rust
#[wasm_bindgen]
pub struct Foo {
    pub first: u32,

    #[wasm_bindgen(readonly)]
    pub second: u32,
}
```

output
```ts
export class Foo {
  free(): void;
/**
*/
  first: number;
/**
*/
  readonly second: void;
}
```

### Expected Behavior

```ts
export class Foo {
  free(): void;
/**
*/
  first: number;
/**
*/
  readonly second: number;
}
```

### Steps to Reproduce

https://github.com/MadonoHaru/wasm-bindgen-type-issue

anicholson:
This occurs for me too:

Versions:

Rust: 1.58.0
wasm-bindgen: "0.2.63"

Current workaround is to do the following:

1. add `#[wasm_bindgen(skip)]` before the field, and
2. implement each getter manually via `#[wasm_bindgen(getter)]`

Hywan:
The problem doesn't seem to be related to `readonly` but, when generating code for struct attributes, (i) a getter is generated, and (ii) a setter. In the case of `readonly`, only a getter is generated.

The bug seems to be in the getter parameter. The `ret_ty` is set to `Some("void")` everytime:

https://github.com/rustwasm/wasm-bindgen/blob/63872ee3514fa656f167e2d0a716501ecc299218/crates/cli-support/src/js/mod.rs#L2545-L2563

With your example, we have:

```
in AuxExportKind::Getter branch:
    field = "first"
    ret_ty = Some("void")

in AuxExportKind::Setter branch:
    field = "first"
    ret_ty = Some("number")

in AuxExportKind::Getter branch:
    field = "second"
    ret_ty = Some("void")
```

Unfortunately, all this code ends up in `push_accessor_ts`:

https://github.com/rustwasm/wasm-bindgen/blob/63872ee3514fa656f167e2d0a716501ecc299218/crates/cli-support/src/js/mod.rs#L3815-L3834

which overrides the `ty` variable (first item of a tuple stored in `typescript_fields`). So when we have a getter + a setter, the setter has the correct `ret_ty` and “fixes” the previous bug with the getter. That's why it appears only with `readonly`.

[...]

PR #2909, opened 2022-05-23
fix(typescript) Getters correctly define the inner results.

Hywan:
This patch fixes
https://github.com/rustwasm/wasm-bindgen/issues/2721. Let's consider
the following example:

```rust
pub struct Foo {
    /// Hello `first`.
    pub first: u32,

    /// Hello `second`.
    #[wasm_bindgen(readonly)]
    pub second: u32,
}
```

It outputs the following `.d.ts` file:

```typescript
export class Foo {
  free(): void;
/**
* Hello `first`.
*/
  first: number;
/**
* Hello `second`.
*/
  readonly second: void;
}
```

What's wrong here is that `second` has the type `void`. It should be
`number`.

What's happening? For `Foo.first`, a getter and a setter are
generated. The getter never has a return type (spoiler: that's the
bug), but the setter has a correct return type somehow (it's infererd
from its arguments, combined with a unit descriptor, see
`wasm_bindgen_cli_support::wit::Context::struct_`). When the getter
and the setter are processed, they both end up calling
`wasm_bindgen_cli_support::js::ExportedClass::push_accessor_ts`. This
function overwrites the return type for the same field everytime it is
called. So when called for the getter, the return type is missing, and
when called for the setter with a type, the bug is fixed.

For `Foo.second`, it's different. There is no setter generated because
the field is marked as `readonly`. So the bug appears.

To fix that, I've updated the getter `Function` descriptor, so that it
has an `inner_ret` value. This is passed to an `Adapter.inner_results`
at some point in the code, which is finally passed by
`JsFunction::process` to `JsFunction::typescript_signature` to
generate the typescript signature. And suddently, it's fixed!

However, we got:

```typescript
export class Foo {
  free(): void;
/**
* Hello `first`.
* @returns {number}
*/
  first: number;
/**
* Hello `second`.
* @returns {number}
*/
  readonly second: number;
}
```

We're making progress!

Nonetheless, the documentation is wrong now: The fields don't return
anything. This problem comes from the way `js_docs` is computed
globally for all export kinds in
`wasm_bindgen_cli_support::js::Context::generate_adapter`. This patch
also updates that to compute `js_docs` lately, for each branch. It is
specialized in the `AuxExportKind::Getter` and `AuxExportKind::Setter`
branches.

I hope this is the correct approach. I'm not sure about the difference
between `Adapter.results` and `Adapter.inner_results` as there is no
documentation unfortunately (kudos for every other super-well
documented place though).

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -229,6 +229,7 @@ impl<'a, 'b> Builder<'a, 'b> {
             asyncness,
         );
         let js_doc = self.js_doc_comments(&function_args, &arg_tys, &ts_ret_ty);
+
         Ok(JsFunction {
             code,
             ts_sig,
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2504,9 +2504,10 @@ impl<'a> Context<'a> {
                     false => None,
                 };
 
-                let docs = format_doc_comments(&export.comments, Some(js_doc));
                 match &export.kind {
                     AuxExportKind::Function(name) => {
+                        let docs = format_doc_comments(&export.comments, Some(js_doc));
+
                         if let Some(ts_sig) = ts_sig {
                             self.typescript.push_str(&docs);
                             self.typescript.push_str("export function ");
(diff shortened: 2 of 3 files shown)


alexcrichton:

Hywan:
I guess it's a good idea ;-). Sorry for forgetting!

Hywan:
I've added a test in `crates/typescript-tests/`. I first wanted to add a test in `crates/cli/tests/reference/` but the existing tests are broken on my machine, and as far as I can see, they seem to not be run in the CI at all.

Hywan:
I will as soon as I find free time!

PR #2910, opened 2022-05-23
feat(jsvalue) Implement `From<()>` for `JsValue`

Hywan:
This could be controversial, but I reckon it makes sense to map `()`
in Rust to `undefined` in JavaScript. When a Rust function returns
“nothing”, it's idiomatic to return `()`. In JavaScript, it is the
equivalent of returning `undefined`.

Supporting this pattern is useful for automatic Rust to JavaScript
conversions.

Edit: Turns out, `()` was already mapped to `JsValue` with the help of the `IntoJsValue` trait. Implementing `From<()>` for `JsValue` directly help removing 2 implementations of `IntoJsValue`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -806,6 +806,13 @@ where
     }
 }
 
+impl From<()> for JsValue {
+    #[inline]
+    fn from(_: ()) -> JsValue {
+        JsValue::UNDEFINED
+    }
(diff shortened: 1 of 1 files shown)


Hywan:
Closing due to failing tests. Will check on my computer later.

PR #2913, opened 2022-05-25
feat(cli-support): expose more data on `Output`

dsherret:
This is to be able to use the output in memory more than what's already available.

(I'm using the cli-support crate in Wasm so that we can build everything via a Deno script without needing to install wasm-bindgen globally)

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -611,9 +611,38 @@ fn unexported_unused_lld_things(module: &mut Module) {
 
 impl Output {
     pub fn js(&self) -> &str {
+        &self.gen().js
+    }
+
+    pub fn ts(&self) -> Option<&str> {
+        let gen = self.gen();
+        if gen.typescript {
+            Some(&gen.ts)
+        } else {
+            None
+        }
+    }
+
+    pub fn start(&self) -> Option<&String> {
+        self.gen().start.as_ref()
+    }
+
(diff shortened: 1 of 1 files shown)


hamza1311 (approved):
The changes look good to me. I like to provide the ability to ability to avoid a global binary install.

> (I'm using the cli-support crate in Wasm so that we can build everything via a Deno script without needing to install wasm-bindgen globally)

I'm concerned about recommending what you're trying to do though. That sounds like a disaster incoming because of version mismatch between the cli-support crate and wasm-bindgen crate used in the binary. If that is irrelevant (it's totally possible I'm totally misunderstanding your use-case), it would be nice if you could clarify what it is that you're trying achieve.

dsherret:
> I'm concerned about recommending what you're trying to do though. That sounds like a disaster incoming because of version mismatch between the cli-support crate and wasm-bindgen crate used in the binary. If that is irrelevant (it's totally possible I'm totally misunderstanding your use-case), it would be nice if you could clarify what it is that you're trying achieve.

Are you talking about the mismatch between the wasm-bindgen crate used in the .wasm file? I'm planning on analyzing the cargo manifest and only supporting the specific version of wasm-bindgen that the cli-support crate is for. I already found the mismatch between the global wasm-bindgen-cli and wasm-bindgen to sometimes give very cryptic errors (though maybe that's been fixed now) so at least this way we can ensure everyone is on the same version of wasm-bindgen-cli-support because it will be built into the script.

PR #2918, opened 2022-05-30
test(ci) Remove duplicate actions

Hywan:
I was wondering if running `cargo test --target wasm32-unknown-unknown`
_twice_ is necessary. If not, please discard this commit.


alexcrichton:
This probably had different configuration historically and when that extra configuration was removed I suspect this was forgotten to be removed.

Issue #2917, opened 2022-05-29
configurable wasm ImportObject


arnemileswinter:
### Motivation

Currently importing functions from JS into rust via extern ties to your Javascript code base heavily. 

### Proposed Solution

A library's generated wrapping `init` function should not create the import Object inline by assigning to {}  but also receive an optional object with predefined fields for wiring what rust expects

### Alternatives

Currently you'd have to write a js module that has an exported` let imports` where the modules wiring functions proxy method calls to that object. 

### Additional Context

I find it pretty confusing that you have to tie your rust code to a hard-coded js file, it would be easier to supply an object with functions during init that serves as the "extern"

PR #2920, opened 2022-05-30
user supplied imports on init

arnemileswinter:
see #2917

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -814,7 +814,7 @@ impl<'a> Context<'a> {
                     }}
                 }}
 
-                async function init(input{init_memory_arg}, userImports = {}) {{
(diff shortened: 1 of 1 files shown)


alexcrichton:
This is a pretty low-level feature that's quite easy to get wrong, mess up the internal state of wasm-bindgen, and inadvertently cause issues. Can you explain in more detail what this is going to be used for? 

One thing I'm worried about is that wasm-bindgen hasn't been designed for this use case. That means that with no tests here there's no guarantee that this will continue to work in the future. Furthermore I don't have at the forefront of my mind what the possible gotchas are within the tool today.

arnemileswinter:
Hold up i misunderstood docs
Closing now

Was mislead by this, with the top drop down stating rust, only just noticed it's web assembly Script not rust...

https://wasmbyexample.dev/examples/importing-javascript-functions-into-webassembly/importing-javascript-functions-into-webassembly.assemblyscript.en-us.html

PR #2924, opened 2022-06-02
Expose helpers to allow synchronous initialization

### LINKED ISSUE #2698: API to load the WASM synchronously
For a project I am working on it would be ideal to have an API that allows me to instantiate the WASM synchronously. So if you already have the bytes of the wasm file at hand you can call `load` or `loadSync` to just create a WebAssembly instance with all its imports.

Right now all functions are async functions and return Promises.

Is this something you'd be down to add? I am happy to jump in and submit a PR for this.

d3lm:
This PR exposes helpers that will allow to synchronously initialize the WASM module. The helpers are only exported when compiling for `web`. But we could also export them for `no-modules`.

What I have done is to extract some of the functionality from `init()` into helper functions which are now exported.

This fixes #2698.

### Todos

- [x] Add documentation
- [x] Do we only want to expose the helpers or maybe export a single function `initSync()`?

### Background
At [StackBlitz](https://stackblitz.com/) we are building [WebContainer](https://blog.stackblitz.com/posts/introducing-webcontainers/), a Node.js runtime for the browser, and we could benefit from this addition a lot because it would allow us to lazy load WASM modules on demand. Some features are split into separate WASM modules to improve load performance and those features are often not frequently used. By exposing these helpers we could synchronously instantiate modules in a worker like this then:
```js
import { getImports, initMemory, finalizeInit } from 'my_wasm_module.js';

const imports = getImports();

initMemory(imports[, maybe_memory]);

const bytes = "<sync_fetch_bytes>";

const module = new WebAssembly.Module(bytes);
const instance = new WebAssembly.Instance(module, imports);

const wasm = finalizeInit(instance, module);
```
ℹ️ The above only really works in a Web Worker but that's ok IMO as the main thread disallows sync APIs for good reasons.

What do you think @alexcrichton? Is that what you were thinking of in #2698?


alexcrichton:
This seems reasonable enough to me, thanks for the PR! Can you add some documentation for this as well?

d3lm:
Yep, will add some documentation.

alexcrichton:
Honestly this is so far outside my expertise I don't really have an opinion. It's not "obviously wrong" but I know very little about all the myriad embeddings that JS can be run in, so I have no idea how to gauge which is better to provide. I'm basically leaving such a judgement call to you.

crates/cli-support/src/js/mod.rs
```diff
@@ -469,7 +469,8 @@ impl<'a> Context<'a> {
             OutputMode::Web => {
                 self.imports_post.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, Some(&mut imports))?;
-                footer.push_str("export default init;\n");
+                footer.push_str("export { initSync }\n");
+                footer.push_str("export default init;");
```
d3lm, line 473: ℹ️ I have noticed there was two newlines at the end of the generated JS so I removed one newline.

crates/cli-support/src/js/mod.rs
```diff
@@ -733,11 +753,11 @@ impl<'a> Context<'a> {
...
         let mut imports_init = String::new();
-        if self.wasm_import_definitions.len() > 0 {
-            imports_init.push_str("imports.");
-            imports_init.push_str(module_name);
-            imports_init.push_str(" = {};\n");
-        }
+
+        imports_init.push_str("imports.");
+        imports_init.push_str(module_name);
+        imports_init.push_str(" = {};\n");
```
d3lm, line 759: ℹ️ I have noticed that if a module has no imports and we don't initialize the `imports` with the `module_name` (`wbg`) that it will break when it tries to initialize the memory, because it gets assigned to `imports.wbg.memory`.

crates/cli-support/src/js/mod.rs
```diff
@@ -814,26 +834,49 @@ impl<'a> Context<'a> {
...
+                function finalizeInit(instance, module) {{
                     wasm = instance.exports;
                     init.__wbindgen_wasm_module = module;
-
                     {post_instantiate}
                     {start}
                     return wasm;
                 }}
+
+                function initSync(bytes{init_memory_arg}) {{
```
d3lm, line 855: ℹ️ This function gets exported to synchronously initialize the module.

d3lm:
@alexcrichton This PR is ready for review.

alexcrichton:
Thanks!

d3lm:
@alexcrichton Nice! When do you think an we release this as a new version? So that the wasm-bindgen book is also updated with the example etc.

alexcrichton:
I'll look into doing a release next week

alexcrichton:
Ok I'm doing a publish at https://github.com/rustwasm/wasm-bindgen/pull/2935

d3lm:
Thanks a lot!

PR #2926, opened 2022-06-03
Add mediasession api

rozbb:
I also added some helpful docs for generating new interfaces from WebIDLs


rozbb:
MDN docs are [here](https://developer.mozilla.org/en-US/docs/Web/API/MediaSession) and spec is [here](https://w3c.github.io/mediasession)

PR #2935, opened 2022-06-14
Bump to 0.2.81

alexcrichton:


d3lm:
I noticed that the example I added in #2924 does not show up in the [book](https://rustwasm.github.io/docs/wasm-bindgen/). Do we need to publish this separately?

alexcrichton:
I believe that website is maintained by workflows in https://github.com/rustwasm/rustwasm.github.io which haven't been maintained in a long time to the point that they probably no longer work.

PR #2949, opened 2022-06-17
Replace `cfg=web_sys_unstable_apis` with `unstable` feature

rozbb:
This resolves a long-standing issue I've had with `web-sys`.

## Problem

I'm building a webapp that relies on the [MediaSession API](https://developer.mozilla.org/en-US/docs/Web/API/MediaSession). **There is no way for my crate to use an unstable API without user intervention** (whether that be by setting `RUSTCFLAGS` or otherwise). Currently, my build instructions say
>  Set the following in ~/.cargo/config:
> ```
> [build]
> rustflags = ["--cfg=web_sys_unstable_apis"]
> rustdocflags = ["--cfg=web_sys_unstable_apis"]
>```
This is an odd requirement, and a barrier to usage.

## Solution

Just gate all the unstable APIs behind a new `unstable` feature flag (in addition to their existing feature flags). This satisfies the requirement that users must explicitly opt into the unstable APIs, while streamlining their use for crates that really do want them.

I understand this change is of a structural nature, so I'd love to hear others' thoughts on this.

## What this PR has

* A simple switch in `web-sys` to using `#[cfg(feature = "unstable")]` instead of `#[cfg(web_sys_unstable_apis)]`
* An updated guide page on using the `unstable` feature

## What this PR does not have

For succinctness, I have _not_ regenerated the `web-sys` files. **Please let me know when I can regen `web-sys`** before this is merged

### DIFF
--- a/crates/web-sys/Cargo.toml
+++ b/crates/web-sys/Cargo.toml
@@ -14,7 +14,6 @@ edition = "2018"
 
 [package.metadata.docs.rs]
 all-features = true
-rustdoc-args = ["--cfg=web_sys_unstable_apis"]
 
 [lib]
 doctest = false
--- a/crates/webidl/src/generator.rs
+++ b/crates/webidl/src/generator.rs
@@ -41,7 +41,7 @@ fn comment(mut comment: String, features: &Option<String>) -> TokenStream {
 fn maybe_unstable_attr(unstable: bool) -> Option<proc_macro2::TokenStream> {
     if unstable {
         Some(quote! {
-            #[cfg(web_sys_unstable_apis)]
+            #[cfg(feature = "unstable")]
         })
     } else {
         None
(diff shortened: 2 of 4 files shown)


rozbb:
(I believe the failing tests are both due to the fact that I didn't push the rebuilt `web-sys`; again, let me know when I should do that)

alexcrichton:
Thanks for the PR, but this is currently an intentional design decision of web-sys. The goal is to have web-sys follow Cargo's definition of semver and if an `unstable` Cargo feature is used for these breaking APIs then that's unfortunately not being followed. It's understood that using the unstable APIs is not easy otherwise, but that was an intentional tradeoff to reflect that it's an unstable browser API where the browser API is just as likely to change as well.

rozbb:
Unfortunate, but I understand. I guess I'll just add the flags to my build scripts

Issue #694, opened 2018-08-13
Support exporting Rust functions as variadic JS functions


derekdreery:
Support the following example

```rust
#[wasm_bindgen(apply_ending_slice_bikeshedded)]
pub fn add(first: u32, second: u32, rest: &[u32]) -> u32 {
    rest.iter().sum() + first + second
}
```

callable from javascript as

```js
import { add } from './wasm';

console.log(add(1, 2, 3, 4, 5)); // prints 15
```

You could implement the transformation in javascript as

```js
function add(first, second, ..rest) {
    return add_from_rust(first, second, rest);
}
```

This is ~~a prerequisite for implementing #503~~ the reverse of #503.

alexcrichton:
Seems plausible to me! I think though that this is the opposite of #503, right? (this is about defining something in Rust whereas that's about importing something defined in JS)

derekdreery:
Yeah I don't know why I said prerequisite :)

joaofreires:
@alexcrichton maybe we can close this issue as #2954 has been merged.

PR #2954, opened 2022-06-19
Add exporting Rust functions as variadic JS functions

joaofreires:
This is an attempt to solve #694.

I kept the `variadic` `bindgen` attribute and assumed the last argument will be received as a JS Array object with the variadic values. Added a `variadic` flag inside the `Function`s structures, and changed the export signature of functions flagged as variadic, then only the JS signature changed. 

I found a limitation in this implementation around the argument typing export. If we export a `variadic` function with a type different from `JsValue`, we can generate the TS exports with wrong types.

For example:
```rust
#[wasm_bindgen(variadic)]
pub fn example(arr: &Array) {}
```
Will generate:
```js
export function example(...arr: Array<any>): void;
```

The workaround I found is create a new type and export the `typescript_type` (I tested extending the Array structure and it worked as "expected").

Suggestions are welcome.

### TODO:
 - create more unit tests covering more variadic function use posibilities
 - ~~add the variadic "dots" in the JSDoc generation code.~~

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -859,6 +859,7 @@ fn function_from_decl(
             rust_vis: vis,
             r#async: sig.asyncness.is_some(),
             generate_typescript: opts.skip_typescript().is_none(),
+            variadic: opts.variadic().is_some(),
         },
         method_self,
     ))
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -411,7 +411,6 @@ impl<'a> ConvertToAst<BindgenAttrs> for &'a mut syn::ItemStruct {
             };
 
             let attrs = BindgenAttrs::find(&mut field.attrs)?;
-            // assert_not_variadic(&attrs)?;
             if attrs.skip().is_some() {
                 attrs.check_used()?;
                 continue;
@@ -627,7 +626,6 @@ impl ConvertToAst<BindgenAttrs> for syn::ForeignItemType {
     type Target = ast::ImportKind;
 
     fn convert(self, attrs: BindgenAttrs) -> Result<Self::Target, Diagnostic> {
-        // assert_not_variadic(&attrs)?;
         let js_name = attrs
             .js_name()
             .map(|s| s.0)
@@ -678,7 +676,7 @@ impl<'a> ConvertToAst<(BindgenAttrs, &'a ast::ImportModule)> for syn::ForeignIte
         if self.mutability.is_some() {
             bail_span!(self.mutability, "cannot import mutable globals yet")
         }
-        // assert_not_variadic(&opts)?;
+
         let default_name = self.ident.to_string();
         let js_name = opts
             .js_name()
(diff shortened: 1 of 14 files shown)


crates/macro-support/src/parser.rs
```diff
@@ -1559,14 +1560,14 @@ fn assert_no_lifetimes(sig: &syn::Signature) -> Result<(), Diagnostic> {
...
-    Ok(())
-}
+// fn assert_not_variadic(attrs: &BindgenAttrs) -> Result<(), Diagnostic> {
+//     if let Some(span) = attrs.variadic() {
+//         let msg = "the `variadic` attribute can only be applied to imported \
+//                    (`extern`) functions";
+//         return Err(Diagnostic::span_error(*span, msg));
+//     }
+//     Ok(())
+// }
```
alexcrichton, line 1570: I think this can be removed now?

crates/macro-support/src/parser.rs
```diff
@@ -411,7 +411,7 @@ impl<'a> ConvertToAst<BindgenAttrs> for &'a mut syn::ItemStruct {
             };
 
             let attrs = BindgenAttrs::find(&mut field.attrs)?;
-            assert_not_variadic(&attrs)?;
+            // assert_not_variadic(&attrs)?;
```
hamza1311, line 414: Why have dead code in there? Can you please add a comment explaining why this is commented out?

crates/macro-support/src/parser.rs
hamza1311: :+1:

We shouldn't have dead code in there without a reason

joaofreires:
I created tests for simple variadic functions inside `crates/typescript-tests/src/simple_fn.[rs | ts]`. I'm not sure if we have more places with TS tests.
I also made a small update to the docs, but I'm sure it can be improved.

crates/macro-support/src/parser.rs
joaofreires: Removed, no longer needed.

alexcrichton:
Thanks!

ishitatsuyuki:
I think this regressed some part of processing somehow. I still haven't reduced a reproduction from my private codebase, but meanwhile here's the panic stacktrace I get:

```
thread 'main' panicked at 'assertion failed: mid <= self.len()', crates/cli-support/src/decode.rs:43:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/28b891916d4c85cd10fb2e9cfa8bc836a2c459f3/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/28b891916d4c85cd10fb2e9cfa8bc836a2c459f3/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/28b891916d4c85cd10fb2e9cfa8bc836a2c459f3/library/core/src/panicking.rs:48:5
   3: <&str as wasm_bindgen_cli_support::decode::Decode>::decode
   4: <alloc::vec::Vec<T> as wasm_bindgen_cli_support::decode::Decode>::decode
   5: <wasm_bindgen_cli_support::decode::Program as wasm_bindgen_cli_support::decode::Decode>::decode
   6: wasm_bindgen_cli_support::wit::process
   7: wasm_bindgen_cli_support::Bindgen::generate_output
   8: wasm_bindgen_cli_support::Bindgen::generate
   9: wasm_bindgen::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Reverting the merge commit makes things working again.

ishitatsuyuki:
Nevermind, looks like it's just a schema breaking change and I need to align my project's dependency to the git version too.

Issue #2947, opened 2022-06-16
Getters and setters are missing jsdoc types in 0.2.81 (not 0.2.80)


dsherret:
### Describe the Bug

In the latest 0.2.81 the .js file is missing jsdoc types on getters and setters. This is because in #2909 they were set to `None`: https://github.com/rustwasm/wasm-bindgen/pull/2909/files#diff-072222e984a549efeebf98f26d0c59a69d88abe246db7e0958654fc32471313eR2531

### Steps to Reproduce

Sorry, I do not have time right now to offer a simple reproduction, but I have pointed out the cause in the code above. Also a fix can be seen here, which is a revert of the above: https://github.com/rustwasm/wasm-bindgen/commit/387e821c8074b18e251fa5991ccc531302223961 -- but it is probably not a real fix since this was done for some reason in #2909

### Expected Behavior

Getters and setters should still have jsdoc comment types when not using TypeScript.

### Actual Behavior

The types are missing in 0.2.81

cc @Hywan

Hywan:
Hello @dsherret,
Can you confirm that https://github.com/rustwasm/wasm-bindgen/pull/2956 solves your problem? We should restore some disabled tests to ensure this behavior is totally fixed, I'll probably work on this soon.

dsherret:
@Hywan yup, it's fixed. Thanks so much for looking into it!

PR #2956, opened 2022-06-20
fix(cli-support) Differentiate JS docs and TS docs when generating adapters

Hywan:
A bug has been raised in
https://github.com/rustwasm/wasm-bindgen/issues/2947, which is the
consequence of a bug introduced in
https://github.com/rustwasm/wasm-bindgen/pull/2909.

This patch fixes it by differentiating JS docs and TS docs
clearly. The `Context::push_setter` and `Context::push_getter` methods
now receive a `js_docs` and a `ts_docs`, which prevent any confusions.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2564,31 +2564,33 @@ impl<'a> Context<'a> {
                     false => None,
                 };
 
+                let js_docs = format_doc_comments(&export.comments, Some(js_doc));
+
                 match &export.kind {
                     AuxExportKind::Function(name) => {
-                        let docs = format_doc_comments(&export.comments, Some(js_doc));
-
                         if let Some(ts_sig) = ts_sig {
-                            self.typescript.push_str(&docs);
+                            self.typescript.push_str(&js_docs);
                             self.typescript.push_str("export function ");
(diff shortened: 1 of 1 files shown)


Hywan:
Fail to build on rustc nightly because:

> error: package `time v0.3.10` cannot be built because it requires rustc 1.57.0 or newer, while the currently active rustc version is 1.56.0-nightly

because there is:

https://github.com/rustwasm/wasm-bindgen/blob/af2a7b9e762a2334ceb24e5f262a2be59c70dd48/.github/workflows/main.yml#L74

in some places.

@alexcrichton Which nightly versions are you willing to update to?

joaofreires:
> error: package time v0.3.10 cannot be built because it requires rustc 1.57.0 or newer, while the currently active rustc version is 1.56.0-nightly

Looks like this error is because the crate `time` launched a new release, tests prior the new version (using the crate `time v0.3.9`) were working fine.

https://crates.io/crates/time/versions

alexcrichton:
Updating that nightly is fine to do at any time, it's just pinned to avoid breaking when rust releases a new nightly.

joaofreires:
@alexcrichton I tested with `nightly-2022-05-19` and all tests passed.

PR #2962, opened 2022-06-23
Replace lazy_static with once_cell

jplatte:
… and another drive-by improvement.

Nicer to read because it isn't a macro, and most of the Rust ecosystem has switched over already.


hamza1311 (approved):
Looks good :+1:. Thanks for the PR

PR #2964, opened 2022-06-24
doc: fix typo in hello-world

rubenrua:
`hello-world` example does use Webpack.

### DIFF
--- a/guide/src/examples/hello-world.md
+++ b/guide/src/examples/hello-world.md
@@ -38,7 +38,7 @@ Our JS entry point is quite small!
 
 ## Webpack-specific files
 
-> **Note**: Webpack is not required for this example, and if you're interested
+> **Note**: Webpack is required for this example, and if you're interested
 > in options that don't use a JS bundler [see other examples][wab].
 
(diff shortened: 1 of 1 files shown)


guide/src/examples/hello-world.md
```diff
@@ -38,7 +38,7 @@ Our JS entry point is quite small!
 
 ## Webpack-specific files
 
-> **Note**: Webpack is not required for this example, and if you're interested
+> **Note**: Webpack is required for this example, and if you're interested
```
hamza1311, line 41: Can you please remove the entire note, since it doesn't make sense to keep it here?

hamza1311 (changes requested):

guide/src/examples/hello-world.md
rubenrua: Why? I found it interesting

hamza1311 (approved):
Thanks for the PR

(and sorry for unneeded previous comment of removing the note entirely)

rubenrua:
You are welcome.

PR #2978, opened 2022-07-05
Properly implement `From<usize/isize>` and `PartialEq<usize/isize>` for `JsValue`

### LINKED ISSUE #2977: `usize` and `isize` are converted into `BigInt` in some situations
https://github.com/rustwasm/wasm-bindgen/blob/d7072b17a70d181627904ce4e49293965ffeb912/src/lib.rs#L870

They are the same size as `u32` and `i32` on `wasm32` targets, so they should be represented as regular `number` values, right?

This makes the following function return a `BigInt` on the JS side:

```rs
#[wasm_bindgen]
pub async fn random_usize() -> usize { 4 }
```

While the following returns a normal number:

```rs
#[wasm_bindgen]
pub fn random_usize_sync() -> usize { 4 }
```

However, I believe the generated Typescript definitions report the function as returning a number in both cases, which is problematic in the `async` case.

Liamolucko:
Fixes #2977

Previously, `usize` and `isize` were converted to bigints when converting to JS values, when they should be converted to regular JS numbers due to being 32-bit. `PartialEq` was then implemented on top of that, checking whether `self == JsValue::from(n)`.

It could be argued that they should be bigints to try and be forwards-compatible with 64-bit WebAssembly; however, they're already represented as regular numbers everywhere else.

One major issue here is that this is a breaking change. You can't just use a number where a bigint is expected (or vice versa), so this will break any code that used `From` to convert from a `usize`/`isize` to `JsValue`. That happens most commonly indirectly, through the return type of async functions, which are internally converted using `Into<JsValue>`.

### DIFF
--- a/crates/typescript-tests/src/lib.rs
+++ b/crates/typescript-tests/src/lib.rs
@@ -7,4 +7,5 @@ pub mod simple_async_fn;
 pub mod simple_fn;
 pub mod simple_struct;
 pub mod typescript_type;
+pub mod usize;
 pub mod web_sys;
--- a/crates/typescript-tests/src/usize.rs
+++ b/crates/typescript-tests/src/usize.rs
@@ -1,0 +1,13 @@
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub fn usize_identity(x: usize) -> usize { x }
+
+#[wasm_bindgen]
+pub fn isize_identity(x: isize) -> isize { x }
+
+#[wasm_bindgen]
+pub async fn async_usize_identity(x: usize) -> usize { x }
+
+#[wasm_bindgen]
+pub async fn async_isize_identity(x: isize) -> isize { x }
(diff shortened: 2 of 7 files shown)


hamza1311 (approved):
Thanks for the fix!

Liamolucko:
Just in case I wasn't clear enough (it was sorta buried at the bottom of the PR description), **this will almost certainly break anything which relies on the previous behaviour**. I doubt that there are many things that do, and this isn't actually a breaking change to the API (so the code will still _compile_), but I want to double-check that you're okay with that.

hamza1311:
It should be fine. I assume the equality works as it's supposed to, right?

```rust
let one: JsValue = JsValue::from(10usize);
assert!(one == 10usize)
```

How the comparison happens was (and still is) an implementation detail so I don't see how this breaking change effects much.

RReverser:
This definitely broke serde-wasm-bindgen tests and took a bit to narrow down, but, luckily, it's only the tests that got broken, because Serde itself always serializes usize/isize as u64/i64, regardless of platform size, so our runtime behaviour didn't change.

PR #2986, opened 2022-07-09
Add support for exporting static getters and setters

### LINKED ISSUE #2895: Unable to bind a method that is both static and a getter
### Describe the Bug

I'm trying to create a [custom element](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements), which requires to expose observedAttributes method which is a `static get`.

```javascript
static get observedAttributes() { return ['c', 'l']; }
```

```rust
#[wasm_bindgen]
pub struct MyCustomElement;

#[wasm_bindgen]
impl MyCustomElement {
    #[wasm_bindgen(getter)]
    pub fn observed_attributes() {
        log::info!("a static get method");
    }
}
```
Compiling it will result in an error:
```log
[INFO]: Compiling to Wasm...
    Finished release [optimized] target(s) in 0.06s
[INFO]: Installing wasm-bindgen...
  left: `1`,
 right: `0`', crates/cli-support/src/js/binding.rs:166:9
Caused by: failed to execute `wasm-bindgen`: exited with exit status: 101
  full command: "<home>/.cache/.wasm-pack/wasm-bindgen-96f8477c88f36826/wasm-bindgen" "<project>/target/wasm32-unknown-unknown/release/datetime_widget.wasm" "--out-dir" "<project>/pkg" "--typescript" "--target" "web"
```
The error points at this code, which is just an assertion error.
```rust
        assert_eq!(js.stack.len(), adapter.results.len());
```

### Steps to Reproduce

1. Create a cargo library project
```sh
cargo new --lib myproj
```
2. Change `Cargo.toml` to have 
```toml
[lib]
crate-type = ["cdylib"]
```
3. Put the above rust code into `lib.rs`
4. Compile using `wasm-pack build --release --target=web`
5. The compilation error will show up.

If applicable, add a link to a test case (as a zip file or link to a repository we can clone).

### Expected Behavior

- static method that is also a getter should be possible
- Also tried using `static_method_of` which has the same compilation error
```rust
    #[wasm_bindgen(static_method_of = MyCustomElement, getter)]
    pub fn observed_attributes() {
        log::info!("a static get method");
    }
```

Liamolucko:
Fixes #2895

Exported methods with `#[wasm_bindgen(getter/setter)]` and no `self` parameter are now interpreted as getters/setters for static fields. Previously, attempting to do that would cause `wasm-bindgen` to crash.

This can already be done for imports with `static_method_of` + `getter`.

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -64,6 +64,9 @@ pub struct JsFunction {
     pub js_doc: String,
     pub ts_arg_tys: Vec<String>,
     pub ts_ret_ty: Option<String>,
+    /// Whether this function has a single optional argument.
+    ///
+    /// If the function is a setter, that means that the field it sets is optional.
     pub might_be_optional_field: bool,
     pub catch: bool,
     pub log_error: bool,
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1,7 +1,7 @@
 use crate::descriptor::VectorKind;
 use crate::intrinsic::Intrinsic;
 use crate::wit::{
-    Adapter, AdapterId, AdapterJsImportKind, AuxExportMethodKind, AuxReceiverKind, AuxValue,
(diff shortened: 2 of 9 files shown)


alexcrichton:

PR #2987, opened 2022-07-09
Don't copy doc comments on imported functions to generated `extern` blocks

Liamolucko:
This was causing lots of 'unused doc comment' warnings while compiling `web-sys`, since doc comments have no effect on `extern` blocks.

I've implemented this by removing any doc comments from the list of attributes we're copying and then adding them back to the public function.

I'm not sure how good of an idea it is to forward all attributes to the extern block in the first place, but I'm playing it safe and leaving it as is.


alexcrichton:

PR #2990, opened 2022-07-10
chore: Remove extra `dbg` calls

TheDoctor314:
These were probably left by mistake.


hamza1311 (approved):
> These were probably left by mistake.

Correct :+1: 

Thanks for removing them

PR #2991, opened 2022-07-13
Use `Closure::new` in examples

Liamolucko:
There isn't much of a reason to use `Closure::wrap` over `Closure::new` anymore, so this changes `wasm-bindgen`'s examples to use `Closure::new` as the recommended method of creating closures.

I also updated a couple of spots in `wasm-bindgen-futures` and fixed a typo in the docs of `Closure` which accidentally referred to `Closure::new` as `Closure::wrap`.


alexcrichton:
👍

PR #2995, opened 2022-07-15
Added `#[track_caller]` and Improved Debug message for `unwrap_throw()`

Billy-Sheppard:
Tested on firefox and edge (on fedora) with my own trait on top locally, I could not get it to run off a path dep due to 
```
thread 'main' panicked at 'assertion failed: mid <= self.len()', /home/billy/.cargo/registry/src/github.com-1ecc6299db9ec823/wasm-bindgen-cli-support-0.2.80/src/decode.rs:43:27
```
Help would be appreciated.

- Added `#[track_caller]`
- Using `core::panic::location` to retreive called file, line, col locations
- matches existing api

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,7 +8,6 @@
 #![no_std]
 #![allow(coherence_leak_check)]
 #![doc(html_root_url = "https://docs.rs/wasm-bindgen/0.2")]
-use std::format;
 use core::convert::TryFrom;
 use core::fmt;
 use core::marker;
@@ -17,6 +16,7 @@ use core::ops::{
     Add, BitAnd, BitOr, BitXor, Deref, DerefMut, Div, Mul, Neg, Not, Rem, Shl, Shr, Sub,
 };
 use core::u32;
+use std::format;
 
 use crate::convert::{FromWasmAbi, WasmSlice};
 
@@ -1318,7 +1318,7 @@ pub trait UnwrapThrowExt<T>: Sized {
             loc.line(),
             loc.column()
         );
-        self.expect_throw(&msg)
+        self.expect_throw(msg.as_str())
     }
 
     /// Unwrap this container's `T` value, or throw an error to JS with the
@@ -1358,7 +1358,6 @@ where
         }
     }
 }
-
 
 /// Returns a handle to this wasm instance's `WebAssembly.Module`
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1309,21 +1309,22 @@ pub fn anyref_heap_live_count() -> u32 {
 pub trait UnwrapThrowExt<T>: Sized {
     /// Unwrap this `Option` or `Result`, but instead of panicking on failure,
     /// throw an exception to JavaScript.
+    #[cfg(feature = "std")]
     #[track_caller]
     fn unwrap_throw(self) -> T {
         let loc = core::panic::Location::caller();
-        let mut msg = String::new();
-        core::fmt::write(
-            &mut msg,
-            core::format_args!(
-                "`unwrap_throw` failed (at {}:{}:{})",
-                loc.file(),
-                loc.line(),
-                loc.column()
-            ),
-        )
-        .expect("Error occurred while trying to write in String");
-        self.expect_throw(msg.as_str())
+        let msg = std::format!(
+            "`unwrap_throw` failed ({}:{}:{})",
+            loc.file(),
+            loc.line(),
+            loc.column()
+        );
+        self.expect_throw(&msg)
+    }
+
+    #[cfg(not(feature = "std"))]
+    fn unwrap_throw(self) -> T {
+        self.expect_throw("`unwrap_throw` failed")
     }
 
     /// Unwrap this container's `T` value, or throw an error to JS with the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1309,33 +1309,31 @@ pub fn anyref_heap_live_count() -> u32 {
 pub trait UnwrapThrowExt<T>: Sized {
     /// Unwrap this `Option` or `Result`, but instead of panicking on failure,
     /// throw an exception to JavaScript.
-    #[cfg(all(feature = "std", debug_assertions))]
-    #[track_caller]
-    fn unwrap_throw(self) -> T {
-        let loc = core::panic::Location::caller();
-        let msg = std::format!(
-            "`unwrap_throw` failed ({}:{}:{})",
-            loc.file(),
-            loc.line(),
-            loc.column()
-        );
-        self.expect_throw(&msg)
-    }
 
-    #[cfg(not(all(feature = "std", debug_assertions)))]
     fn unwrap_throw(self) -> T {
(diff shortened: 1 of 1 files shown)


Liamolucko (commented):
Great idea! `expect_throw` should probably show the call site as well.

Billy-Sheppard:
> Great idea! `expect_throw` should probably show the call site as well.

Should it? Happy to include it, but my initial thinking was custom messages probably already indicate where in the code it was called from.

Will get on to those other fixes ASAP. Tks for the review

src/lib.rs
```diff
@@ -17,6 +16,7 @@ use core::ops::{
     Add, BitAnd, BitOr, BitXor, Deref, DerefMut, Div, Mul, Neg, Not, Rem, Shl, Shr, Sub,
 };
 use core::u32;
+use std::format;
```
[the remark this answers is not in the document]
Billy-Sheppard: Seems to throw a CI error without. Agree its mildly strange. Could switch to write! if needed.

src/lib.rs
```diff
@@ -8,7 +8,6 @@
 #![no_std]
 #![allow(coherence_leak_check)]
 #![doc(html_root_url = "https://docs.rs/wasm-bindgen/0.2")]
-
```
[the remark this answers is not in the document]
Billy-Sheppard: Yes

Billy-Sheppard:
> The purpose of `unwrap_throw` was originally to be very small on codegen, so specifically not having features like this (or the usage of `format!`). I think for this you'd instead want to use `.unwrap()` which has all the features baked in.

For me (on Edge and Firefox Fedora/Silverblue), `.unwrap()` doesn't provide line references. Outputs: `Uncaught (in promise) RuntimeError: unreachable executed`

Perhaps this can be extracted to an optional feature or a separate function? I personally think it would be highly useful, but I understand why it potentially shouldn't be baked in. I have also updated the function to use `core::fmt::write` in place of `format!`;

Liamolucko:
> For me (on Edge and Firefox Fedora/Silverblue), `.unwrap()` doesn't provide line references. Outputs: `Uncaught (in promise) RuntimeError: unreachable executed`. Is there some other component to getting unwrap to work?

Yes - you need [`console_error_panic_hook`](https://docs.rs/console_error_panic_hook/latest/console_error_panic_hook/) to log panic messages to the console, which include the call site.

Billy-Sheppard:
Thanks for that Liam, that does work for me, not sure if that's documented somewhere and I missed it somehow.

`unwrap()` does cause slightly different behavior to `unwrap_throw()` still however. For instance the console reports two errors, one with the rust stack, and one with `Uncaught (in promise) RuntimeError: unreachable`. 

Perhaps my altered `unwrap_throw` could have `#[cfg(all(feature = "std", debug_assertions)]` on it?

Pauan:
For the record, I think it would be really useful if `unwrap_throw` displayed filename and line numbers in debug mode (but not release mode).

Billy-Sheppard:
After some productive discussion in [this PR](https://github.com/Pauan/rust-dominator/pull/70), it seems sensible to me to at least get some further explanation why this change couldn't be pushed through.

I've added debug assertions as I agree that's a sensible change.

Billy-Sheppard:
Sorry to ping, @alexcrichton, do you have any further thoughts on having the filename tracking only when compiling with `std` and `debug`?

Also just following up on the discussion around [console_error_panic_hook] (https://docs.rs/console_error_panic_hook/latest/console_error_panic_hook/), is this documented somewhere?

I'm not sure about the majority of WASM crates but without PR'ing to the ones that do use `unwrap_throw` in place of `unwrap`, it can be difficult to debug some more sticky issues. Obviously that's more of an issue with those crates but perhaps the documentation/messaging can be improved?

src/lib.rs
```diff
@@ -1196,17 +1196,33 @@ pub fn anyref_heap_live_count() -> u32 {
...
+
+    #[cfg(not(all(feature = "std", debug_assertions)))]
     fn unwrap_throw(self) -> T {
         self.expect_throw("`unwrap_throw` failed")
     }
 
     /// Unwrap this container's `T` value, or throw an error to JS with the
     /// given message if the `T` value is unavailable (e.g. an `Option<T>` is
     /// `None`).
+    #[track_caller]
```
alexcrichton, line 1220: Could this (and the others below) be `#[cfg_attr(debug_assertions, track_caller)]`?

src/lib.rs
```diff
@@ -1196,17 +1196,33 @@ pub fn anyref_heap_live_count() -> u32 {
 pub trait UnwrapThrowExt<T>: Sized {
     /// Unwrap this `Option` or `Result`, but instead of panicking on failure,
     /// throw an exception to JavaScript.
+    #[cfg(all(feature = "std", debug_assertions))]
+    #[track_caller]
+    fn unwrap_throw(self) -> T {
```
alexcrichton, line 1201: Instead of a duplicate method could this be inlined into the same function with something like:

```rust
if cfg!(debug_assertions) {
    // this method
} else {
    // existing method
}
```

alexcrichton (commented):
I think gating this behind `debug_assertions` is reasonable, yeah.

Billy-Sheppard:
I've made those changes you've suggested there @alexcrichton, how does it look? 

Is `cfg!(all(debug_assertions, track_caller, std))` the correct `if` guard?

> Also just following up on the discussion around [console_error_panic_hook](https://docs.rs/console_error_panic_hook/latest/console_error_panic_hook/), is this documented somewhere?

This doesn't seem to be mentioned in the guide, perhaps I should draft another PR and add in some documentation on this for the guide?

Liamolucko:
> Is `cfg!(all(debug_assertions, track_caller, std))` the correct `if` guard?

No, it should be `cfg!(all(debug_assertions, feature = "std"))`.

src/lib.rs
```diff
@@ -1196,17 +1196,31 @@ pub fn anyref_heap_live_count() -> u32 {
 pub trait UnwrapThrowExt<T>: Sized {
     /// Unwrap this `Option` or `Result`, but instead of panicking on failure,
     /// throw an exception to JavaScript.
+
```
Liamolucko, line 1199: I think you accidentally removed the `#[track_caller]` here.

Billy-Sheppard:
Do the track caller attrs have to remain then as before (`#[track_caller]`)? To cause rust to get the correct file/line/col reference?

Or is the full guard `cfg!(all(debug_assertions, track_caller, feature = "std"))`?

Won't that mean there will have to be two functions like before?

Liamolucko:
> Do the track caller attrs have to remain then as before (`#[track_caller]`)? To cause rust to get the correct file/line/col reference?

No, they should be `#[cfg_attr(debug_assertions, track_caller]`, which just means 'apply the `#[track_caller]` attribute if `cfg(debug_assertions)` is true'.

Billy-Sheppard:
Perfect, thanks Liam, I had a bit of a brain fart and conflated those two changes Alex suggested.  Should be working now.

Billy-Sheppard:
Any further work required on this?

PR #3002, opened 2022-07-22
Bump to 0.2.82

GalvinGao:
Me personally is in great need of #2965 because I'm trying to make vite to work and, as mentioned in https://github.com/Menci/vite-plugin-wasm/issues/5 it appears that such issue has already been fixed. It would be awesome to be able to use vite and typescript as I follow along the tutorial!


litaiquan:
Me too

Liamolucko:
You should also update the `wasm-bindgen-shared` schema version, since the schema was changed in #2954: https://github.com/rustwasm/wasm-bindgen/blob/a8e934df8244846b4a099f9b392fa7b01e25a536/crates/shared/src/lib.rs#L9

Apart from that, this looks good to me - cc @alexcrichton

GalvinGao:
huh, schema version check failed at https://github.com/rustwasm/wasm-bindgen/runs/7479700475?check_suite_focus=true. I noticed that the hash has also been changed in #2954, but it seems to me that specify version in `crates/shared/src/lib.rs` in sync with `crates/shared/Cargo.toml` makes sense to me. Anything I have done wrong?

Liamolucko:
That hash is the hash of `lib.rs`, so you have to update it as well (to `"18229224525184165582"`, as per the CI failure).

The point of it (I think) is to remind people to update the schema version by having that test fail when the schema in `lib.rs` changes.

alexcrichton:
Seems reasonable to me to publish, thanks! With CI fixed I can merge and publish/tag.

GalvinGao:
CI passed! Consider squash when merging plz :D

Issue #3009, opened 2022-07-28
BigUint64Array/BigInt64Array should be constructed from u64 arrays, not BigInt arrays


ishitatsuyuki:
In https://docs.rs/js-sys/latest/js_sys/struct.BigUint64Array.html, various methods like `pub fn copy_from(&self, src: &[BigInt])` takes a slice of `BigInt`s. But `BigInt` is an object, and this just doesn't look right. When indexed from JS code they should indeed return a BigInt object, but from the view of wasm their representation should be a u64.

Liamolucko:
You're right, but unfortunately it's a breaking change to fix it.

ishitatsuyuki:
Well, does the old signature even work? It sounds like it will read garbage since the type signature doesn't make much sense?

PR #3011, opened 2022-07-29
Convert Big[Ui,I]nt64Array bindings to use [u/i]64

ishitatsuyuki:
Using BigInt as the macro parameters leads to absurd signatures like
slices of BigInt objects (can't exist). Instead, just use the numeric
types. bindgen already converts them to bigint when appropriate.

Closes: https://github.com/rustwasm/wasm-bindgen/issues/3009
Fixes: d6d056cd ("Add math-related intrinsics/functions for `JsValue`s (#2629)")

**Breaking change**: This is technically a breaking change, but I'm not sure if anyone has managed to use it in the old state.

I added some brief tests but I haven't tested all the functions yet. Will do some more testing later in my own project.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -5854,9 +5854,9 @@ arrays! {
 
     /// `BigInt64Array()`
     /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for this! I agree that the previous signatures were too broken to work so fixing them should be fine, despite the breakage.

Issue #2963, opened 2022-06-24
web-sys examples broken since 0.2.81


jhinch:
### Describe the Bug

It appears as though all of the live web-sys examples in the wasm-bindgen guide are not working and confirmed they were not working when running locally as well. Examples tested which are not working:

* https://rustwasm.github.io/wasm-bindgen/examples/dom.html
* https://rustwasm.github.io/wasm-bindgen/examples/paint.html
* https://rustwasm.github.io/wasm-bindgen/examples/todomvc.html

It seems like other live examples are still working such as the hello_world example

* https://rustwasm.github.io/wasm-bindgen/exbuild/hello_world/

### Steps to Reproduce

1. Checkout main of wasm-bindgen
2. `cd examples/dom`
3. `npm install`
4. `npm run serve`
5. Open http://localhost:8080 in your browser
6. See that 'Hello rust!' is never appended to the page

### Expected Behavior

examples/dom is expected to append a <p> tag with the text `hello rust!`

### Actual Behavior

Nothing happens

### Additional Context

I ran a git bisect between 0.2.80 and the tip of main and it found 062aa5f70a1e8a170d0ae90123ebdeb3242ac782 as the first bad commit

Liamolucko:
I think this was caused by #2961, so it should be fixed now.

gengjiawen:
@Liamolucko Is there any plan to draft a new release ?

Liamolucko:
> @Liamolucko Is there any plan to draft a new release ?

0.2.82 was just released (in #3002)

PR #3015, opened 2022-08-02
Test examples in CI

Liamolucko:
This PR adds some testing of examples to CI. It doesn't actually test any of the functionality of the examples, it just runs them and checks that no errors get logged to the console. Hopefully, this should stop issues like #2963 from happening again.

I've also made CI build a couple extra examples that it didn't build before - previously only the examples that used webpack were built, but I've now made it build the rest as well by unifying on a `build.sh` for everything other than webpack.

Since WebDriver doesn't support getting at logs yet, I've implemented this using the experimental new [WebDriver BiDi](https://w3c.github.io/webdriver-bidi/) protocol which does. Currently I've only implemented support for testing with Firefox which has it built in, but there's also a [Chromium implementation](https://github.com/GoogleChromeLabs/chromium-bidi) if you think it's important to support that too.


.github/workflows/main.yml
```diff
@@ -112,7 +112,8 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: '16'
-    - run: cargo test
+    # Run everything except the example tests; we run them later in `test_examples`.
+    - run: cargo test --test must_use --test non_wasm --test std-crate-no-std-dep --test unwrap_throw
```
alexcrichton, line 116: Could testing of examples have something like `required-features = ['...']` to avoid the need to allow-list tests here? Otherwise this makes it unfortunately easy to forget to add a line here if a new test is added.

.github/workflows/main.yml
```diff
@@ -223,16 +224,23 @@ jobs:
...
+    - run: |
+        curl -L https://github.com/WebAssembly/binaryen/releases/download/version_109/binaryen-version_109-x86_64-linux.tar.gz -sSf > binaryen-version_109-x86_64-linux.tar.gz
+        tar -xz -f binaryen-version_109-x86_64-linux.tar.gz
     - run: |
         cargo build -p wasm-bindgen-cli
         ln -snf `pwd`/target/debug/wasm-bindgen $(dirname `which cargo`)/wasm-bindgen
     - run: mv _package.json package.json && npm install && rm package.json
     - run: |
-        for dir in `ls examples | grep -v README | grep -v asm.js | grep -v raytrace | grep -v without-a-bundler | grep -v wasm-in-web-worker | grep -v websockets | grep -v webxr | grep -v deno | grep -v synchronous-instantiation`; do
+        export PATH="$PWD/binaryen-version_109/bin:$PATH"
```
alexcrichton, line 235: Could this `PATH` append be done above where binaryen is downloaded? (e.g. `echo the-path >> $GITHUB_PATH` during the step)

alexcrichton (approved):
Sounds reasonble to me, thanks for this!

.github/workflows/main.yml
```diff
@@ -112,7 +112,8 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: '16'
-    - run: cargo test
+    # Run everything except the example tests; we run them later in `test_examples`.
+    - run: RUSTFLAGS="--cfg=disable_example_tests" cargo test
```
alexcrichton, line 116: Given how heavyweight these tests are in system dependencies could this be an opt-in feature rather than an opt-out feature? (e.g. via `required-features = ['...']`)

Liamolucko: I've moved it to a separate `example-tests` crate instead to avoid adding a debug-only feature to the root `wasm-bindgen` crate.

alexcrichton (approved):

PR #3016, opened 2022-08-02
Handle precompiled modules in initSync on target web

lukaslihotzki:
While the default init can handle a variety of possible input types, initSync currently needs a BufferSource. With this change, initSync can also accept a WebAssembly.Module, so it is more consistent to default init.

This may be useful for workers and worklets when the compiled module is reused and blocking the thread is acceptable.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -614,16 +614,19 @@ impl<'a> Context<'a> {
         } else {
             declare_or_export = "export";
 
-            sync_init_function.push_str(&format!("\
+            sync_init_function.push_str(&format!(
+                "\
+                {declare_or_export} type SyncInitInput = BufferSource | WebAssembly.Module;\n\
                 /**\n\
-                * Synchronously compiles the given `bytes` and instantiates the WebAssembly module.\n\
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -614,16 +614,19 @@ impl<'a> Context<'a> {
...
             declare_or_export = "export";
 
-            sync_init_function.push_str(&format!("\
+            sync_init_function.push_str(&format!(
+                "\
+                {declare_or_export} type SyncInitInput = BufferSource | WebAssembly.Module;\n\
                 /**\n\
-                * Synchronously compiles the given `bytes` and instantiates the WebAssembly module.\n\
+                * Uses a given `module` directly or compiles it from a BufferSource and then\n\
+                * instantiates the WebAssembly module.\n\
```
Liamolucko, line 622: This is a bit unclear; it makes it sound like the module doesn't get instantiated if it's precompiled. 

I think ```Instantiates the given `module`, which can either be bytes or a precompiled `WebAssembly.Module`.``` is a bit clearer.

Liamolucko (commented):
Looks good apart from one nitpick about the documentation.

PR #3026, opened 2022-08-09
Update docs of raytrace-parallel

lukaslihotzki:
* Consistently use the netlify mirror introduced in #2259
* Consider the possibility of running wasm threads with `--target web`


Liamolucko (approved):
Thanks!

PR #3031, opened 2022-08-16
Deprecate `JsValue::from_serde` and `JsValue::into_serde`

### LINKED ISSUE #1258: JsValue serde support shouldn't go through serde_json
Going through `serde_json` ends up including a bunch of `f64` formatting code, which is quite code size heavy.

I'm not super familiar with implementing serde serializers and deserializers, but *if* we can express "we only support things that can `AsRef<JsValue>`" as a trait bound for serialization, then I think we can go directly into and from `JsValue` without stringifying. This could be a perf win as well.

### LINKED ISSUE #2017: Allow Serde to bind Uint8Array to Vec<u8>
Not sure if this should be considered as a feature or a bug...

### Motivation

Currently, `wasm-bindgen` bind `Vec<u8>` return type to `Uint8Array` on the javascript type. However, when using `into_serde()` to pass the data to the Rust side, the method returns an error when parsing a `Uint8Array` into a `Vec<u8>`.

```rust
#[wasm_bindgen]
pub fn test_serde(data: JsValue)-> Vec<u8> {
    let test: Vec<u8> = data.into_serde().expect("This will crash if the input is a Uint8Array");
    test
}
```

Although the solution in a simple case like this is to simply take a `Vec<u8>` directly as an input, this doesn't work for more complex type. A hack around it wouild be to convert the Uint8Array into an Array on the rust side:

```rust
// This works with either an Array or a Uint8Array
#[wasm_bindgen]
pub fn test_serde(data: JsValue)-> Vec<u8> {
    let data = if JsCast::is_instance_of::<Uint8Array>(&data) {
        JsValue::from(Array::from(&data))
    }
    else {
        data
    };

    let test: Vec<u8> = data.into_serde().unwrap();
    test
}
```

My specific type context was to receive a `Vec<Vec<u8>>`, which would be of the type `Array<Uint8Array>` from the Javascript side. I'll be dumping the code snippet of the hack here in case someone else need it:

```rust
fn test(values: Array) {
    let values = JsValue::from(values.map(&mut |v, _, _| {
        if JsCast::is_instance_of::<Uint8Array>(&v) {
            JsValue::from(Array::from(&v))
        }
        else {
            v
        }
    }));
    //...
}
```

This does work, but is pretty ugly. From a user perspective, it makes sense that deserializing a `Uint8Array` into a `Vec<u8>` should work without this boilerplate.

### Alternatives

For the moment, I'm using the hack described above.

### LINKED ISSUE #2539: JsValue::from_serde is very slow compare to pass string from WASM to js and then using JSON.parse to parse data
### Summary

JsValue::from_serde is very slow compare to pass string from WASM to js and then using JSON.parse to parse data.

### Additional Details

I tried to use 'wasm-bindgen' to test performance of a general senario: request data from server at WASM and then pass data to JS, and find that the `JsValue::from_serde` is very slow.
And then I change to just pass response text to JS instead, and then using `JSON.parse` to parse text to json data, which compare to 'from_serde' is way better performance.

So I wanna ask if there's anything I can do to improve `from_serde` perfroamce?

The perfomrance profile (need to remove extra extension 'txt')

[Profile-from_serde_vs_json_parse.json.txt](https://github.com/rustwasm/wasm-bindgen/files/6389520/Profile-from_serde_vs_json_parse.json.txt)

The performance screenshot
image

And the demo repo to reproduce this issue: https://github.com/maple-leaf/rust-wasm-playgroud

### LINKED ISSUE #2770: Split up to `wasm-bindgen` and to a new `wasm-bindgen-core`
### Motivation

Break the `ahash` dependency cycle:

```
error: cyclic package dependency: package ahash v0.7.4 depends on itself. Cycle:
package ahash v0.7.4
... which is depended on by hashbrown v0.11.2
... which is depended on by indexmap v1.7.0
... which is depended on by serde_json v1.0.64
... which is depended on by wasm-bindgen v0.2.74
... which is depended on by js-sys v0.3.51
... which is depended on by getrandom v0.2.3
... which is depended on by ahash v0.7.4
```

[Currently `getrandom` uses `wasm-bindgen` just to be able to use: ](https://github.com/rust-random/getrandom/pull/220/files)

>    [The contents of:](https://github.com/tkaitchuck/aHash/issues/95#issuecomment-881242012)
>  * [wasm_bindgen::prelude::*](https://github.com/tkaitchuck/aHash/issues/95#issuecomment-881242012)
>  * [wasm_bindgen::JsCast](https://github.com/tkaitchuck/aHash/issues/95#issuecomment-881242012)
>  * [wasm_bindgen::memory()](https://github.com/tkaitchuck/aHash/issues/95#issuecomment-881242012)

Due to feature unification, several crates end up enabling features inside `wasm-bindgen` and its downstream dependencies that causes a dependency cycle that affects various parts of the rust community:

https://github.com/tkaitchuck/aHash/issues/95

### Proposed Solution
This: https://github.com/tkaitchuck/aHash/issues/95#issuecomment-881242012

> js-sys would then only depend on wasm-bindgen-core, the full wasm-bindgen would just reexport wasm-bindgen-core, and getrandom would just depend on wasm-bindgen-core and js-sys.

### Alternatives

Break the `ahash` dependency cycle somewhere else.

But creating a minimal wasm crate that can be used by low-level dependencies like `getrandom` without the risk of pulling in too many dependencies due to feature unification seems to be the most elegant solution.

### Additional Context

Crates like `wasm-bindgen`, `getrandom` are seemingly cornerstones to this ecosystem, it would be great if fundamental crates would compile.

Liamolucko:
I've listed `serde-wasm-bindgen` as the replacement, and changed the section of the guide that talks about Serde to talk about `serde-wasm-bindgen` instead of the deprecated methods.

I didn't remove it entirely because I can imagine someone remembering it and trying to look it back up, only to find that it no longer exists, which would quite frustrating. I also added a footnote about the deprecated methods in case someone remembers the old way and wants to know what happened.

There were several examples using `from_serde`/`into_serde`, which I updated to use `serde-wasm-bindgen` or not use `serde` altogether.

The `fetch` example was a bit weird, in that it took a JS value, parsed it into a Rust value, only to serialize it back into a JS value. I removed that entirely in favour of just passing the original JS value directly. I suppose it behaves slightly differently in that it loses the extra validation, but a panic isn't all that much better than a JS runtime error.

Fixes #2770
Fixes tkaitchuck/aHash#95

This also indirectly 'fixes' issues about the deprecated functions:

Resolves #1258 
Resolves #2017
Resolves #2539

### DIFF
--- a/guide/src/reference/arbitrary-data-with-serde.md
+++ b/guide/src/reference/arbitrary-data-with-serde.md
@@ -140,12 +140,13 @@ pub fn receive_example_from_js(val: JsValue) {
 
 [`gloo_utils::format::JsValueSerdeExt`]: https://docs.rs/gloo-utils/latest/gloo_utils/format/trait.JsValueSerdeExt.html
 
-## `JsValue::from_serde` / `JsValue::into_serde`
+## History
 
-In previous versions of `wasm-bindgen`, `JsValue::from_serde` and
-`JsValue::into_serde` were the recommended way of using Serde to convert
-to/from JS values. These were implemented using the JSON approach with
-`serde_json`. This caused problems when certain features of `serde_json` and
-other crates were enabled that caused it to depend on `wasm-bindgen`, creating
-a circular dependency, which is illegal in Rust and caused people's code to
-fail to compile. So, they were deprecated.
+In previous versions of `wasm-bindgen`, `gloo-utils`'s JSON-based Serde support
+(`JsValue::from_serde` and `JsValue::into_serde`) was built into `wasm-bindgen`
+itself. However, this required a dependency on `serde_json`, which had a
+problem: with certain features of `serde_json` and other crates enabled,
+`serde_json` would end up with a circular dependency on `wasm-bindgen`, which
+is illegal in Rust and caused people's code to fail to compile. So, these
+methods were extracted out into `gloo-utils` with an extension trait and the
+originals were deprecated.
--- a/guide/src/reference/arbitrary-data-with-serde.md
+++ b/guide/src/reference/arbitrary-data-with-serde.md
@@ -1,7 +1,7 @@
 # Serializing and Deserializing Arbitrary Data Into and From `JsValue` with Serde
 
 It's possible to pass arbitrary data from Rust to JavaScript by serializing it
-with [Serde](https://github.com/serde-rs/serde). This is done through the
+with [Serde](https://github.com/serde-rs/serde). This can be done through the
 [`serde-wasm-bindgen`](https://docs.rs/serde-wasm-bindgen) crate.
 
 ## Add dependencies
@@ -75,9 +75,9 @@ pub fn receive_example_from_js(val: JsValue) {
 
 ## JavaScript Usage
 
-In the `JsValue` that JavaScript gets, `field1` will be an `Object` (not a
-JavaScript `Map`), `field2` will be a JavaScript `Array` whose members are
-`Array`s of numbers, and `field3` will be an `Array` of numbers.
+In the `JsValue` that JavaScript gets, `field1` will be a `Map`, `field2` will
+be a JavaScript `Array` whose members are `Array`s of numbers, and `field3`
+will be an `Array` of numbers.
 
 ```js
 import { send_example_to_js, receive_example_from_js } from "example";
@@ -92,13 +92,53 @@ example.field2.push([5, 6]);
 receive_example_from_js(example);
 ```
 
+## An alternative approach - using JSON
+
+`serde-wasm-bindgen` works by directly manipulating JavaScript values. This
+requires a lot of calls back and forth between Rust and JavaScript, which can
+sometimes be slow. An alternative way of doing this is to serialize values to
+JSON, and then parse them on the other end. Browsers' JSON implementations are usually quite fast, and so this
+approach can outstrip `serde-wasm-bindgen`'s performance in some cases.
+
+That's not to say that using JSON is always faster, though - the JSON approach
+can be anywhere from 2x to 0.2x the speed of `serde-wasm-bindgen`, depending on
+the JS runtime and the values being passed. It also leads to larger code size
+than `serde-wasm-bindgen`. So, make sure to profile each for your own use
+cases.
+
+Here's the equivalent of the above examples using JSON:
+
+```rust
+#[wasm_bindgen]
+pub fn send_example_to_js() -> JsValue {
+    let mut field1 = HashMap::new();
+    field1.insert(0, String::from("ex"));
+    let example = Example {
+        field1,
+        field2: vec![vec![1., 2.], vec![3., 4.]],
+        field3: [1., 2., 3., 4.]
+    };
+
+    let json = serde_json::to_string(&example).unwrap();
+    js_sys::JSON::parse(&json).unwrap()
+}
+
+#[wasm_bindgen]
+pub fn receive_example_from_js(val: JsValue) {
+    // `JSON::stringify` returns a JavaScript string,
+    // which we then have to convert into a Rust string.
+    let json = String::from(js_sys::JSON::stringify(&val).unwrap());
+    let example: Example = serde_json::from_str(&json).unwrap();
+    ...
+}
+```
+
 ## `JsValue::from_serde` / `JsValue::into_serde`
 
 In previous versions of `wasm-bindgen`, `JsValue::from_serde` and
 `JsValue::into_serde` were the recommended way of using Serde to convert
-to/from JS values. These functions use `serde_json` to serialize types to JSON
-and parse them on the other end. However, this caused problems when certain
-features of `serde_json` and other crates were enabled that caused it to depend
-on `wasm-bindgen`, creating a circular dependency, which is illegal in Rust and
-caused people's code to fail to compile. So, they were deprecated in favor of
-`serde-wasm-bindgen`.
+to/from JS values. These were implemented using the JSON approach with
+`serde_json`. This caused problems when certain features of `serde_json` and
+other crates were enabled that caused it to depend on `wasm-bindgen`, creating
+a circular dependency, which is illegal in Rust and caused people's code to
+fail to compile. So, they were deprecated.
--- a/guide/src/reference/arbitrary-data-with-serde.md
+++ b/guide/src/reference/arbitrary-data-with-serde.md
@@ -1,21 +1,19 @@
 # Serializing and Deserializing Arbitrary Data Into and From `JsValue` with Serde
 
 It's possible to pass arbitrary data from Rust to JavaScript by serializing it
-to JSON with [Serde](https://github.com/serde-rs/serde). `wasm-bindgen` includes
-the `JsValue` type, which streamlines serializing and deserializing.
+with [Serde](https://github.com/serde-rs/serde). This is done through the
+[`serde-wasm-bindgen`](https://docs.rs/serde-wasm-bindgen) crate.
 
-## Enable the `"serde-serialize"` Feature
+## Add dependencies
 
-To enable the `"serde-serialize"` feature, do two things in `Cargo.toml`:
-
-1. Add the `serde` and `serde_derive` crates to `[dependencies]`.
-2. Add `features = ["serde-serialize"]` to the existing `wasm-bindgen`
-   dependency.
+To use `serde-wasm-bindgen`, you first have to add it as a dependency in your
+`Cargo.toml`. You also need the `serde` crate, with the `derive` feature
+enabled, to allow your types to be serialized and deserialized with Serde.
 
 ```toml
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
-wasm-bindgen = { version = "0.2", features = ["serde-serialize"] }
+serde-wasm-bindgen = "0.4"
 ```
 
 ## Derive the `Serialize` and `Deserialize` Traits
@@ -42,7 +40,7 @@ pub struct Example {
 }
 ```
 
-## Send it to JavaScript with `JsValue::from_serde`
+## Send it to JavaScript with `serde_wasm_bindgen::to_value`
 
 Here's a function that will pass an `Example` to JavaScript by serializing it to
 `JsValue`:
@@ -58,19 +56,19 @@ pub fn send_example_to_js() -> JsValue {
         field3: [1., 2., 3., 4.]
     };
 
-    JsValue::from_serde(&example).unwrap()
+    serde_wasm_bindgen::to_value(&example).unwrap()
 }
 ```
 
-## Receive it from JavaScript with `JsValue::into_serde`
+## Receive it from JavaScript with `serde_wasm_bindgen::from_value`
 
 Here's a function that will receive a `JsValue` parameter from JavaScript and
 then deserialize an `Example` from it:
 
 ```rust
 #[wasm_bindgen]
-pub fn receive_example_from_js(val: &JsValue) {
-    let example: Example = val.into_serde().unwrap();
+pub fn receive_example_from_js(val: JsValue) {
+    let example: Example = serde_wasm_bindgen::from_value(val).unwrap();
     ...
 }
 ```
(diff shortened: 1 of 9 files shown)


gthb:
😬😬😬 ... we went the other way (for a parser module compiled to WebAssembly), dropping `serde-wasm-bindgen` in favor of `JsValue::from_serde` ... effectively simply this change:
```diff
 pub fn parse(formula: &str) -> JsValue {
-    let serializer = serde_wasm_bindgen::Serializer::new().serialize_maps_as_objects(true);
     match native_parse(formula) {
         Ok(result) => {
-            result.serialize(&serializer).unwrap_or_else(|err| {
+            JsValue::from_serde(&result).unwrap_or_else(|err| {
```
and this _significantly_ reduced per-parse call overhead; we saw our measured parse times on real-world data change from:

* average: 92.6 μs,
* 99.9th percentile : 1.44 ms

to:

* average 38.6 μs
* 99.9th percentile : 390 μs

Plus `serde-wasm-bindgen` has the rather terrible issue https://github.com/cloudflare/serde-wasm-bindgen/issues/28 ... still unaddressed, and completely breaks module loading on Safari 14 and earlier.

So, you know, not thrilled about this idea. :-)

alexcrichton (commented):
I would agree that `serde-wasm-bindgen` is a bit of a different use case than the methods deprecated here so I don't think it's a direct replacement. Instead though could the `__wbindgen_json_{parse,serialize}` methods be exposed as first-class methods perhaps? That would give convenient access to `JSON` and additionally be able to lift the serde dependency up in to callers instead of baking it in `wasm-bindgen` itself.

andoriyu:
@alexcrichton I've started this PR https://github.com/rustwasm/gloo/pull/242 to extract this functionality. If `__wbindgen_json_{parse,serialize}` were publicly exposed, then behavior would match exactly. I've documented the difference in my PR.

hamza1311:
As far as I understand it, `__wbindgen_json_{parse,serialize}` is equivalent to `js_sys::JSON::{parse,stringify}` so those are already exposed. Or maybe I'm misunderstanding something?

alexcrichton:
Indeed! In that case I think it would be good to migrate the examples to that instead and to mention that as an alternative in addition to `serde-wasm-bindgen`

andoriyu:
@hamza1311 I think they slightly different: in my PR for gloo, when I passed `undefined` to `js_sys::JSON::stringify` it panicked while `__wbindgen_json_serialize` didn't.

Liamolucko:
> @hamza1311 I think they slightly different: in my PR for gloo, when I passed `undefined` to `js_sys::JSON::stringify` it panicked while `__wbindgen_json_serialize` didn't.

Looks like the intrinsic has a special case for `undefined`:

https://github.com/rustwasm/wasm-bindgen/blob/f197b0e104e904c9a10c1a5e22bea559f9e809e2/crates/cli-support/src/js/mod.rs#L3446-L3453

That should be able to be replicated in gloo pretty easily.

andoriyu:
Oh, it's using a replacer function. I used that initially, but changed it
because I thought it be too slow. Yeah, it's easy to replicate.

andoriyu:
@hamza1311  ~~Oh, it's using a replaces function~~ It's just replacing the entire object with null. That's similar to what I did - instead of calling `JSON.stringify` I return `String::new()` if the object is `undefinied`

alexcrichton:
@Liamolucko would you be up for rewording to indicate that `js_sys::JSON` is an alternative and leaving the examples as-is but updating them to use `js_sys::JSON`?

Liamolucko:
Okay, I've added a section to the guide page which talks about using JSON as an alternative to `serde-wasm-bindgen`. Currently it uses `js-sys` + `serde_json` directly, but it should probably be updated to use `gloo-utils` when rustwasm/gloo#242 is merged.

In my opinion, `serde-wasm-bindgen` is the better default, since the performance can swing either way and `serde-wasm-bindgen` has smaller code size. I agree that JSON should be mentioned though, since it can be much faster in some cases like @gthb's.

Also, I ran `serde-wasm-bindgen`'s benchmark suite on a couple different engines and uploaded the results [here](https://docs.google.com/spreadsheets/d/1hSz6oAJow--LCoev9TBXJCkKYMOpT_lv-qOGaCiUpHM/edit?usp=sharing), if you're curious exactly how it stacks up against JSON.

guide/src/reference/arbitrary-data-with-serde.md
```diff
@@ -94,23 +92,60 @@ example.field2.push([5, 6]);
...
+
+## `JsValue::from_serde` / `JsValue::into_serde`
+
+In previous versions of `wasm-bindgen`, `JsValue::from_serde` and
+`JsValue::into_serde` were the recommended way of using Serde to convert
+to/from JS values. These were implemented using the JSON approach with
+`serde_json`. This caused problems when certain features of `serde_json` and
+other crates were enabled that caused it to depend on `wasm-bindgen`, creating
+a circular dependency, which is illegal in Rust and caused people's code to
+fail to compile. So, they were deprecated.
```
gthb, line 151: It didn't end up being deprecation (it's not saying “don't use this anymore, we may remove support for it in some upcoming version”), so maybe
```suggestion
fail to compile. So, they were moved out of `wasm-bindgen` into `gloo_utils` to
resolve the circular dependency and replaced by `serde-wasm-bindgen` as the
default approach to use.
```
... or something like that.

src/lib.rs
```diff
@@ -214,6 +221,7 @@ impl JsValue {
     ///
     /// Returns any error encountered when serializing `T` into JSON.
     #[cfg(feature = "serde-serialize")]
+    #[deprecated = "causes dependency cycles, use `serde-wasm-bindgen` instead"]
```
gthb, line 224:
```suggestion
    #[deprecated = "causes dependency cycles, use `serde-wasm-bindgen` instead, or `gloo_utils::format::JsValueSerdeExt`"]
```

src/lib.rs
```diff
@@ -201,6 +201,13 @@ impl JsValue {
     /// Creates a new `JsValue` from the JSON serialization of the object `t`
     /// provided.
     ///
+    /// **This function is deprecated**, due to [creating a dependency cycle in
+    /// some circumstances][dep-cycle-issue]. Use [`serde-wasm-bindgen`]
+    /// instead, or manually call `serde_json::to_string` + `JSON.parse`.
```
gthb, line 206:
```suggestion
    /// instead, or `gloo_utils::format::JsValueSerdeExt`, or manually call
    /// `serde_json::to_string` + `JSON.parse`.
```

src/lib.rs
```diff
@@ -225,6 +233,13 @@ impl JsValue {
     /// Invokes `JSON.stringify` on this value and then parses the resulting
     /// JSON into an arbitrary Rust value.
     ///
+    /// **This function is deprecated**, due to [creating a dependency cycle in
+    /// some circumstances][dep-cycle-issue]. Use [`serde-wasm-bindgen`]
+    /// instead, or manually call `JSON.stringify` + `serde_json::from_str`.
```
gthb, line 238:
```suggestion
    /// instead, or `gloo_utils::format::JsValueSerdeExt`, or manually call
    /// `JSON.stringify` + `serde_json::from_str`.
```

src/lib.rs
```diff
@@ -236,6 +251,7 @@ impl JsValue {
     ///
     /// Returns any error encountered when parsing the JSON into a `T`.
     #[cfg(feature = "serde-serialize")]
+    #[deprecated = "causes dependency cycles, use `serde-wasm-bindgen` instead"]
```
gthb, line 254:
```suggestion
    #[deprecated = "causes dependency cycles, use `serde-wasm-bindgen` instead, or `gloo_utils::format::JsValueSerdeExt`"]
```

RReverser:
> Also, I ran `serde-wasm-bindgen`'s benchmark suite on a couple different engines and uploaded the results [here](https://docs.google.com/spreadsheets/d/1hSz6oAJow--LCoev9TBXJCkKYMOpT_lv-qOGaCiUpHM/edit?usp=sharing), if you're curious exactly how it stacks up against JSON.

Just to add some clarification not reflected in graph titles - after various added optimisations over time, serde-wasm-bindgen should be faster than serde-json based approach for almost all cases, _except_ when your data has a lot of strings.

Not much more we can do there, as having all strings serialized into one large string (JSON) on one side, passing that string once, and parsing it back on another side will be always faster than passing tons of small strings one-by-one (same overhead x N).

For other data types we can do a lot better than JSON though, and that's why twitter benchmark (lots of strings) looks so different than others.

guide/src/reference/arbitrary-data-with-serde.md
```diff
@@ -94,23 +92,61 @@ example.field2.push([5, 6]);
...
-performs more calls back and forth between JS and Wasm, which has not been fully
-optimized in all engines, meaning it can sometimes be a speed
-regression. However, in other cases, it is a speed up over the JSON-based
-stringification, so &mdash; as always &mdash; make sure to profile your own use
-cases as necessary.
+## An alternative approach - using JSON
+
+`serde-wasm-bindgen` works by directly manipulating JavaScript values. This
+requires a lot of calls back and forth between Rust and JavaScript, which can
+sometimes be slow. An alternative way of doing this is to serialize values to
```
RReverser, line 99: I feel like it's important to still call out the limitations in terms of type support in JSON-based approach here too (like the document did in the original version).

The new version of this section makes it sound like both approaches are equivalent and the only difference is performance. However, the support for various JavaScript types that are not representable by JSON is an even more important advantage of serde-wasm-bindgen than just runtime perf or code size IMO.

guide/src/reference/arbitrary-data-with-serde.md
```diff
@@ -94,23 +92,61 @@ example.field2.push([5, 6]);
...
+JSON, and then parse them on the other end. Browsers' JSON implementations are
+usually quite fast, and so this approach can outstrip `serde-wasm-bindgen`'s
+performance in some cases.
```
gthb, line 102:
```suggestion
performance in some cases. But this approach supports only types that can be
serialized as JSON, leaving out some important types that `serde-wasm-bindgen`
supports such as `Map`, `Set`, and array buffers.
```
... addressing @RReverser's [point about type support](https://github.com/rustwasm/wasm-bindgen/pull/3031/files#r956780944)

alexcrichton (approved):
This looks great to me, thanks again! I think further iteration is safe to happen in-tree

guide/src/reference/arbitrary-data-with-serde.md
RReverser: Oh, so this edit didn't get in? Can you make a separate PR please?

gthb: Voilà! https://github.com/rustwasm/wasm-bindgen/pull/3047

tomasro27:
I'm still seeing this issue after updating wasm-bindgen to latest 0.2.84.
I also upgraded getrandom on a private branch to use wasm-bindgen 0.2.84.
wasm-bindgen has deprecated JsValue::from_serde and JsValue::into_serde, but it still has serde-json as optional dependency. Is this the issue if I have other crates depending on serde_json?
Any help is appreciated.

On the error below I can see wasm-bindgen on latest version 0.2.84. 
```
error: cyclic package dependency: package `ahash v0.7.6` depends on itself. Cycle:
package `ahash v0.7.6`
    ... which satisfies dependency `ahash = "^0.7.0"` of package `hashbrown v0.12.3`
    ... which satisfies dependency `hashbrown = "^0.12"` of package `indexmap v1.9.2`
    ... which satisfies dependency `indexmap = "^1.5.2"` of package `serde_json v1.0.93`
    ... which satisfies dependency `serde_json = "^1.0"` of package `wasm-bindgen v0.2.84`
    ... which satisfies dependency `wasm-bindgen = "^0.2.84"` of package `js-sys v0.3.61`
    ... which satisfies dependency `js-sys = "^0.3"` of package `getrandom v0.2.81 (https://github.com/tomasro27/getrandom?rev=ddcf36ef7ee401af102debb9c758703b1e574461#ddcf36ef)`
    ... which satisfies dependency `getrandom = "^0.2.3"` of package `ahash v0.7.6`
```

RReverser:
Did you remove the `serde-serialize` feature from `wasm-bindgen` in your `Cargo.toml` after the update?

tomasro27:
@RReverser I did not have the `serde-serialize` feature, but from reading more on the ahash thread, the issue is probably that one of my dependencies directly or indirectly defines `wasm-bindgen` with the `serde-serialize` feature enabled.
Is that a fair assumption? meaning that I would need to track down any dependencies that use `wasm-bindgen` with `serde-serialize` in my dependency tree so that my project does not enable that feature indirectly. 

Is there a way to force indirect dependencies not using `serde-serialize` feature on `wasm-bindgen`? or is the only fix to upgrade those dependencies?

Thanks!

tomasro27:
For anyone that might find this helpful, you can run `cargo metadata > metadata.json` to generate a file with all the crates/dependencies on your project. Then find all usages of `wasm-bindgen` with the feature flag `serde-serialize` (make sure it is not a dev dependency, as that might be okay). After finding which crates depend on `wasm-bindgen` with the feature flag enabled, you will need to update those packages to the latest version that does not use the `serde-serialize` feature in their dependencies.

RReverser:
@tomasro27 FWIW for slightly simpler way you could also do `cargo tree -p wasm-bindgen` which will show you an inverted tree of where wasm-bindgen is used, along with any features.

Issue #3019, opened 2022-08-03
Dependent module support


lukaslihotzki:
### Motivation

Rust crates should be able to abstract worker and worklet handling without requiring extra effort by the application developer. For worker and worklet creation, a URL pointing to the worker module is needed. This worker module typically contains custom code supplied by the Rust crate that creates the corresponding worker. If the worker needs to execute the Rust code of the application, the worker module needs to `import` the main module generated by wasm_bindgen, so the WASM imports are available. Dependent module support would allow Rust crates to put their worker modules into the final package during the build process.

### Proposed Solution

The macro `wasm_bindgen::dependent_module!` takes a local file name to a JS module as input. Additionally, the macro generates an import statement to the main module currently generated by wasm_bindgen code. This import statement is then prepended to the module code. The resulting file is put in the output directory during the build process, similarly to snippets. In contrast to snippets, these modules are not imported by the main module. Instead, the macro resolves to code that returns the URL string to the corresponding file during run time. The macro caller can then use this URL string to create workers or worklets.

### Alternatives

Dependent module URLs can also be built at run time by querying the main module URL (for example, with import.meta.url), and then creating a blob or data URL with the import statement and constant code concatenated. The proposed wasm-audio-worklet example contains [a concise implementation of this approach](https://github.com/rustwasm/wasm-bindgen/pull/3017/files#diff-07cf5528c3f9b4e22f2489de3616bd3ec5adbfeea002b3cad8d7f9d9cd5c68b0) that only works with `--target web`. The wasm-mt and wasm_thread crates use similar, but more complex techniques that also work with `--target no-modules` by manually fetching the source code of the main module, so no import statements are needed.

Generating dependent modules at run time is the inferior solution, because it is opaque to wasm_bindgen. For example, Rust crates using `wasm_bindgen::dependent_module!` don't need to care about the different targets of wasm_bindgen. Also, build time bundlers can optimize or pack dependent modules. Additionally, run-time JS code generation introduces potential security vulnerabilities (like eval), so it should be avoided. In contrast, build-time dependent modules can even be used with restrictive CSP headers.

### Additional Context

[...]

PR #3032, opened 2022-08-16
Add wasm_bindgen::script_url

lukaslihotzki:
Rust code needs the current script URL to locate files relative to the script, and to run Rust code with working imports in workers and worklets. An intrinsic is the clean solution to provide the script URL.

The existing `#[wasm_bindgen(js_namespace = import, js_name = meta)]` approach currently used in wasm-audio-worklet only works because wasm-bindgen translates this to `import.meta`. `let o = import; let m = o.meta;` would break. Additionally, this approach works on ES module targets only. `document.currentScript` is not accessible at all over wasm_bindgen. That's why other projects use [artificially generated stack trace](https://github.com/chemicstry/wasm_thread/blob/main/src/script_path.js) to extract the script URL. That's unnecessary effort, especially because wasm_bindgen already needs the script URL to locate the WASM module on most targets. Therefore, wasm_bindgen just needs to pass the path to user code.

Only the bundler target has no obviously correct implementation. I chose `__webpack_public_path__`, which is webpack-specific. However, it should be more useful than `bail!` or the `file://`-URL `import.meta.url` is replaced with during compilation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -421,7 +421,7 @@ jobs:
         tar xf artifacts/doc_api/docs.tar.gz
         mv target/doc gh-pages/api
         mv artifacts/examples1 gh-pages/exbuild
-        mv artifacts/examples2 gh-pages/exbuild/raytrace-parallel
+        mv artifacts/examples2/* gh-pages/exbuild
         mv artifacts/benchmarks gh-pages/benchmarks
         tar czf gh-pages.tar.gz gh-pages
     - uses: actions/upload-artifact@v2
--- a/crates/cli-support/src/intrinsic.rs
+++ b/crates/cli-support/src/intrinsic.rs
@@ -229,6 +229,9 @@ intrinsics! {
         #[symbol = "__wbindgen_module"]
         #[signature = fn() -> Externref]
         Module,
+        #[symbol = "__wbindgen_script_url"]
+        #[signature = fn() -> String]
+        ScriptUrl,
         #[symbol = "__wbindgen_function_table"]
         #[signature = fn() -> Externref]
         FunctionTable,
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -379,6 +379,13 @@ impl<'a> Context<'a> {
             OutputMode::NoModules { global } => {
                 js.push_str("const __exports = {};\n");
                 js.push_str("let script_src;\n");
+                js.push_str("\
+                    if (typeof document === 'undefined') {
+                        script_src = location.href;
+                    } else {
+                        script_src = document.currentScript.src;
+                    }"
+                );
                 js.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, None)?;
                 footer.push_str(&format!("{} = Object.assign(init, __exports);\n", global));
@@ -701,11 +708,6 @@ impl<'a> Context<'a> {
                     stem = self.config.stem()?
                 ),
                 OutputMode::NoModules { .. } => "\
-                    if (typeof document === 'undefined') {
-                        script_src = location.href;
-                    } else {
-                        script_src = document.currentScript.src;
-                    }
                     if (typeof input === 'undefined') {
                         input = script_src.replace(/\\.js$/, '_bg.wasm');
                     }"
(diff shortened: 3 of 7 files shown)


daxpedda (changes requested):
When `omit_default_module_path` (not exactly sure what this is) is `true`, the `else` arm is missing setting `script_src`: https://github.com/lukaslihotzki/wasm-bindgen/blob/27df7a9edd2da5135a4edf8cc17f17c7d533dd14/crates/cli-support/src/js/mod.rs#L716.

I fixed it like this:
```rust
if let OutputMode::NoModules { .. } = self.config.mode {
    "\
            script_src = location.href;
        } else {
            script_src = document.currentScript.src;
        }"
    .to_string()
} else {
    String::from("")
}
```

Liamolucko (commented):
I don't really think that this should be a public API, since any use of it will be tied to a particular target. I think it's better off rolled up as an implementation detail of #3034. I've left some comments anyway, though, since I think this code will probably get reused there.

daxpedda:
> I don't really think that this should be a public API, since any use of it will be tied to a particular target. I think it's better off rolled up as an implementation detail of #3034. I've left some comments anyway, though, since I think this code will probably get reused there.

I tried to properly understand dependent module support (#3019), but I think I lack too much context.

Just to make sure that our use-case is covered: we are simply trying to create shared memory workers using the same WASM file, to do that we need the path to the JS shim created by `wasm-bindgen`, all alternative solutions are terrible workarounds or come with their own limitations. @Liamolucko Is this use-case actually covered by #3019? I can definitely not see it in #3034.

lukaslihotzki:
The use case of shared memory workers is already covered by the existing code in #3034. To create shared memory workers, you cannot just use the script_url, because you need a custom JS module run by the worker to receive the WASM module and WASM memory (SharedArrayBuffer) in onmessage. Then, this custom module uses the wasm_bindgen main module to call init (with module and memory as arguments) and the worker entry point function.

How to create a worker and handle onmessage is shown in the wasm-bindgen-futures change. How to receive the module and the memory to init a WASM worker is shown in the wasm-audio-worklet example.

If #3034 has a chance to get upstreamed, I can also add a complete example with #3034 WASM workers. Previously, I did not get any feedback about #3034.

daxpedda:
> The use case of shared memory workers is already covered by the existing code in #3034.

I'm aware of all the details necessary you pointed out. The part that I don't get is how does #3034 solve receiving the JS module, not the one that handles receiving the WASM module and memory, but the JS shim created by `wasm-bindgen`. That's the part that I can't seem to find.

lukaslihotzki:
The commit "Handle $wbg_main imports" adds a replacement of imports from `$wbg_main` to imports from the wasm-bindgen JS shim. You can see its use in https://github.com/rustwasm/wasm-bindgen/pull/3034/files#diff-1b7e3352a103f85c5662fd0cbba6127ab3464b67ee47e64a3b0d4d9aa34bd5e4R4.

daxpedda:
Ah, I see. It's a bit magic to me how this works but that's off-topic here.
Thanks!

daxpedda:
@Liamolucko could we reconsider adding this API, as #3034 is a big change and probably very far away and the current workarounds are pretty bad?

#3169 makes this PR even smaller, @lukaslihotzki if you could rebase?

hamza1311:
I think instead of increasing wasm-bindgen public API and adding another intrinsic, this can be implemented using already exposed API, so perhaps this should be part of gloo? inline_js snippet that exposes a function which returns the script path sounds like a good solution to me. I'd happy to accept a PR adding it to gloo-utils

daxpedda:
Unfortunately this isn't possible, as far as I know.
`import.meta.url` can be used externally, but `location.href`/`document.currentScript.src` cannot, because it is only evaluated during processing of the script file.

This is the same problem to why https://github.com/rustwasm/wasm-bindgen/pull/3169 was necessary.

Liamolucko:
> @Liamolucko could we reconsider adding this API, as #3034 is a big change and probably very far away and the current workarounds are pretty bad?

Yeah, I've warmed to this now. In combination with #3168, it should be possible to create target-agnostic things that use this, which was my main problem here.

However, access to the script URL is practically useless outside of the `web` and `no-modules` targets right now, because the rest of the targets don't expose an `init` function and can't be used for multithreading. So, I think that this should probably just support those two for now, returning an error if it's called on one of the other targets. Proper shims can be added for other targets in later PRs.

Also, I think `shim_url` might be a better name for this.

daxpedda:
Closing in favor of #3247.

Issue #2822, opened 2022-02-28
Implement WASM ABI conversion for `u128` and `i128`


bhgomes:
### Motivation

To cover all the integers in the Rust standard library, we should have an implementation for these two missing types `u128` and `i128`. We want to implement at least the `IntoWasmAbi` and `FromWasmAbi` traits.

### Proposed Solution

Since `u64` already implements this with the `Wasm64` ABI type, we can use something like

```rust
#[repr(C)]
pub struct Wasm128 {
    pub low: Wasm64,
    pub high: Wasm64,
}
```

and implement the `FromWasmAbi` and `IntoWasmAbi` traits as follows:

```rust
impl WasmAbi for Wasm128 {}

impl IntoWasmAbi for u128 {
    type Abi = Wasm128;

    #[inline]
    fn into_abi(self) -> Wasm128 {
        Wasm128 {
            low: (self as u64).into_abi(),
            high: ((self >> 64) as u64).into_abi(),
        }
    }
}

impl FromWasmAbi for u128 {
    type Abi = Wasm128;

    #[inline]
    unsafe fn from_abi(js: Wasm128) -> u128 {
        u128::from(u64::from_abi(js.low)) | (u128::from(u64::from_abi(js.high)) << 64)
    }
}
```

We can use a macro like the existing implementation to cover both the `u128` and `i128` cases.

### Alternatives

An alternative is to define a wrapper type around `u128` which I can use in my own crate which I then attach the `wasm_bindgen` macro to. This is undesirable if we can just put this directly in this library.

PR #3037, opened 2022-08-18
Pass 64-bit integers directly across the wasm ABI boundary

Liamolucko:
Previously, they were passed as a pair of `i32`s; this changes them to be passed as a single `i64`, which is directly converted to a bigint on the JS end.

I think they were previously passed as a pair because `wasm-bindgen`'s support for 64-bit integers predated the WebAssembly bigint integration that translates `i64`s to bigints. However, that feature has been around for quite a while now, so I think it should be fine to use.

I also bumped the schema version - although the schema itself hasn't changed, the ABI has, so using an older version of the CLI with this version of the `wasm-bindgen` crate or vice versa won't work and I want the version-mismatch warning to show up.

Fixes cloudflare/serde-wasm-bindgen#28. This doesn't rely on a global `BigInt64Array` like the previous `i32`-pair version, which means that code using 64-bit integers will no longer fail at load time in browsers without bigint support.

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -8,7 +8,7 @@ use crate::js::Context;
 use crate::wit::InstructionData;
 use crate::wit::{Adapter, AdapterId, AdapterKind, AdapterType, Instruction};
 use anyhow::{anyhow, bail, Error};
-use walrus::Module;
+use walrus::{Module, ValType};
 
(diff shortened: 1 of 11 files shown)


alexcrichton:
Oh how the times change!

Liamolucko:
> I wonder if basically the revert of this PR could be used for https://github.com/rustwasm/wasm-bindgen/issues/2822. Other than replacing 64 -> 128 and 32 -> 64, things should be pretty similar, right?

Should be, yeah.

PR #3039, opened 2022-08-20
Refactor: None is not an ImportModule

lukaslihotzki:
`ImportModule::None` is neither a module nor an import. (What would its contents or its `from` part be?) Therefore, the correct name for ImportModule would be ImportModuleOption, or `Option<ImportModule>` in idiomatic Rust. Also, this change allows to enforce non-`None` modules by the type system, which is handy, but not strictly needed, for #3019.

### DIFF
--- a/crates/backend/src/encode.rs
+++ b/crates/backend/src/encode.rs
@@ -239,18 +239,22 @@ fn shared_variant<'a>(v: &'a ast::Variant, intern: &'a Interner) -> EnumVariant<
 
 fn shared_import<'a>(i: &'a ast::Import, intern: &'a Interner) -> Result<Import<'a>, Diagnostic> {
     Ok(Import {
-        module: match &i.module {
-            Some(ast::ImportModule::Named(m, span)) => {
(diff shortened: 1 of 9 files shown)


crates/backend/src/encode.rs
```diff
@@ -239,19 +239,25 @@ fn shared_variant<'a>(v: &'a ast::Variant, intern: &'a Interner) -> EnumVariant<
...
     Ok(Import {
-        module: match &i.module {
-            ast::ImportModule::Named(m, span) => {
-                ImportModule::Named(intern.resolve_import_module(m, *span)?)
-            }
-            ast::ImportModule::RawNamed(m, _span) => ImportModule::RawNamed(intern.intern_str(m)),
-            ast::ImportModule::Inline(idx, _) => ImportModule::Inline(*idx as u32),
-            ast::ImportModule::None => ImportModule::None,
-        },
+        module: Option::transpose(i.module.as_ref().map(|m| shared_module(m, intern)))?,
```
Liamolucko, line 242:
```suggestion
        module: i.module.as_ref().map(|m| shared_module(m, intern)).transpose()?,
```

PR #2478, opened 2021-03-03
Add complete WebIDL for ReadableStream

DesmondWillowbrook:
Earlier: #2373 

To solve: #2252 
Found this [source](https://streams.spec.whatwg.org/#rs-class-definition), which is the official Streams spec.

PR #3046, opened 2022-08-29
web-sys: Added full Streams WebIDL

rozbb:
## Why this is needed

The [Streams living standard](https://streams.spec.whatwg.org) is currently broken up into `enabled/ReadableStream.webidl` and `unstable/{ReadableStreamBYOBReader.webidl, ReadableStreamDefaultReader.webidl, ReadableStreamGenericReader.webidl}`. There's two issues with this.

1. This does not encompass the entire standard. For example, `ReadableStreamDefaultController` does not appear here, and it happens to be the only way to support `fetch()` progress in Safari (MDN currently lists it as unsupported but I've verified that it works on [this site](https://fetch-progress.anthum.com)).
2. I don't think it accurately represents the stability of any component. The Streams spec is a Living Standard, and it makes no claim about `ReadableStream`'s API being more solidified than any other interface's.

So my thought is to take all of the Streams standard and put it in `enabled/`. If you'd rather it all be in `unstable/` that'd be okay too.

## Blocking problem

This currently fails to regenerate the Rust bindings with the following error:
```
Error: compiling WebIDL into wasm-bindgen bindings in file "webidls/enabled/Streams.webidl", line 12 column 1

Caused by:
    failed to parse webidl at byte position 75182
```
I think it's because the `ReadableStream` interface is decorated with `[Exposed=*, ...`, which the parser does not know how to parse. I don't know the best solution here. There is

1. Just modify the IDL to expand `*` into whatever namespaces it expands into. This is probably the simplest.
5. Update `weedle` to accept the new syntax
6. ~~Update to `weedle2`, though I haven't checked if it accepts this syntax~~ It [doesn't looks like](https://github.com/mozilla/uniffi-rs/blob/e913610353d4811c3f51a40be4f89d3b0101c5d8/weedle2/src/attribute.rs) `weedle2` supports wildcards in extended attribute lists.


alexcrichton:
Sorry I don't know how best to update weedle for the webidl syntax here. Otherwise are all of these APIs supported in major browsers at this time? If so this seems fine to land with the webidl bits updated but otherwise I don't think this is the right time to land it.

rozbb:
Just made a PR to weedle. I tried it on my local machine and it fixes the build.

Regarding stability: this API is supported in all major browsers except for Safari (see, e.g., [this support matrix](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader)). Would you rather it all be in `unstable/`? Your call

alexcrichton:
Ok that seems reasonable enough to me to support 👍 

I'd be happy to merge with weedle updated here

rozbb:
It's actually pinned to weedle master, so it should work already. Could you rerun the CI?

alexcrichton:
I did that and it looks like it picked up the changes, but it appears that web-sys still needs regenerating.

rozbb:
Ope, yup forgot about that. Pushed.

alexcrichton:
It looks like CI may not be happy with the new crate?

rozbb:
Didn't notice there were more files from the same API lying around. Fixed.

alexcrichton (approved):

rozbb:
Thank you for the quick turnaround! And also ofc thank you for maintaining this crate ❤️

Liamolucko:
There's a problem here - the WebIDL you pulled in doesn't include any of the `[Throws]` attributes that `web-sys` relies on to mark functions as throwing (because `[Throws]` is non-standard). That means that none of these functions return `Result` when some of them should. (You can see which in [Mozilla's WebIDL](https://hg.mozilla.org/mozilla-central/file/tip/dom/webidl/), which does include `[Throws]`.)

This also happened when `ReadableStream` was first added (in #2478), which is already stable, so I'm not sure what to do about that.

PR #3047, opened 2022-08-30
doc: mention inferior type support of JSON approach

gthb:
Reinstate the mention of inferior type support of the JSON approach (compared to the now-standard `serde-wasm-bindgen` approach) that was lost in https://github.com/rustwasm/wasm-bindgen/pull/3031

(From suggestion in https://github.com/rustwasm/wasm-bindgen/pull/3031#discussion_r957280652 that didn't make it into the merge.)


Liamolucko (approved):
LGTM. I'll merge this once CI finishes.

PR #3049, opened 2022-08-30
Better BigInt conversions

RReverser:
Add some `[iu](64|128)` intrinsics to convert large numbers to `JsValue` directly, without stringifying and parsing them back.

Also adding a few more tests as previously those conversions didn't have any test coverage.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -3323,9 +3323,19 @@ impl<'a> Context<'a> {
                 args[0].clone()
             }
 
-            Intrinsic::BigIntNew => {
+            Intrinsic::BigIntFromStr => {
                 assert_eq!(args.len(), 1);
                 format!("BigInt({})", args[0])
+            }
+
+            Intrinsic::BigIntFromI64 | Intrinsic::BigIntFromU64 => {
+                assert_eq!(args.len(), 1);
+                args[0].clone()
+            }
+
+            Intrinsic::BigIntFromI128 | Intrinsic::BigIntFromU128 => {
+                assert_eq!(args.len(), 2);
+                format!("{} << BigInt(64) | {}", args[0], args[1])
             }
 
             Intrinsic::StringNew => {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,7 +154,7 @@ impl JsValue {
     /// allocated large integer) and returns a handle to the JS version of it.
     #[inline]
     pub fn bigint_from_str(s: &str) -> JsValue {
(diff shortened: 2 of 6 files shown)


Liamolucko (commented):
Looks good apart from a minor nitpick.

src/lib.rs
```diff
@@ -926,7 +932,11 @@ externs! {
 
         fn __wbindgen_string_new(ptr: *const u8, len: usize) -> u32;
         fn __wbindgen_number_new(f: f64) -> u32;
-        fn __wbindgen_bigint_new(ptr: *const u8, len: usize) -> u32;
+        fn __wbindgen_bigint_from_str(ptr: *const u8, len: usize) -> u32;
+        fn __wbindgen_bigint_from_i64(n: i64) -> u32;
+        fn __wbindgen_bigint_from_u64(n: u64) -> u32;
+        fn __wbindgen_bigint_from_i128(hi: i64, lo: u64) -> u32;
+        fn __wbindgen_bigint_from_u128(hi: u64, lo: u64) -> u32;
```
hamza1311, line 939: Why not pass i128/u128 directly instead of hi/lo and computing the value later?

hamza1311 (approved):
Looks good to me!

src/lib.rs
RReverser: wasm-bindgen can't pass them natively yet, see #2822.

RReverser: Also, it would have to generate similar code even if it could, because Wasm JS BigInt integration currently allows only passing 64-bit values.

crates/cli-support/src/js/mod.rs
```diff
@@ -3323,11 +3323,21 @@ impl<'a> Context<'a> {
...
             }
 
+            Intrinsic::BigIntFromI64 | Intrinsic::BigIntFromU64 => {
+                assert_eq!(args.len(), 1);
+                args[0].clone()
+            }
+
+            Intrinsic::BigIntFromI128 | Intrinsic::BigIntFromU128 => {
+                assert_eq!(args.len(), 2);
+                format!("{} << BigInt(64) | {}", args[0], args[1])
```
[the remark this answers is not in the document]
Liamolucko: Ah, I didn't think of that.

Liamolucko (approved):

crates/cli-support/src/js/mod.rs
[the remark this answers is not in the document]
RReverser: (upd: I now see that wasm-bindgen already uses `n` syntax in some other places, at least after #3037, so maybe it's not such a big concern)

PR #3051, opened 2022-08-31
Allow returning Result from tests

### LINKED ISSUE #2565: Support Result return types for wasm_bindgen_test macro.
### Motivation

It would be nice to be able to use `?` in a `wasm_bindgen_test` similar to [non-wasm testing](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/question-mark-in-main-and-tests.html).

Currently to do this you do something like
```rust
#[wasm_bindgen_test]
fn error_test() {
    let foo : Result<(), String> = Ok(());
    assert!(foo.is_ok());
    let foo = foo.unwrap();
    // Do something with foo.
}
```

### Proposed Solution

Similar to how rust #[test] can return `Result<T, U>`:
```rust
#[test]
fn error_test() -> Result<(), String> {
    let foo = some_func_which_returns_result()?
    // Do something with foo.
}
```
We could have:
```rust
#[wasm_bindgen_test]
fn error_test() -> Result<(), String> {
    let foo = some_func_which_returns_result()?
    // Do something with foo.
}
```

I'll admit I've not looked into the difficulty of this task but I'm open to implementing it if there is interest in accepting a PR.

RReverser:
Makes it possible to use the newer `-> Result<(), E>` way of reporting test failures instead of just panicking in `#[wasm_bindgen_test]`, similarly to how you can do that in regular `#[test]`.

Resolves #2565.

### DIFF
--- a/crates/test/sample/tests/common/mod.rs
+++ b/crates/test/sample/tests/common/mod.rs
@@ -10,9 +10,9 @@ fn pass() {
 #[wasm_bindgen_test]
 async fn pass_after_2s() {
     console_log!("immediate log");
-    Timeout::new(Duration::new(1, 0)).await;
+    Timeout::new(Duration::from_secs(1)).await;
     console_log!("log after 1s");
-    Timeout::new(Duration::new(1, 0)).await;
+    Timeout::new(Duration::from_secs(1)).await;
     console_log!("log at end");
 }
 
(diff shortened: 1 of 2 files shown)


alexcrichton (approved):

cataggar:
Thank you @RReverser !!

PR #3052, opened 2022-09-01
web-sys: Added [Throws] attributes to Streams API

rozbb:
Following the comment in https://github.com/rustwasm/wasm-bindgen/pull/3046#issuecomment-1233461732, I went through `Streams.webidl` and added a `[Throws]` attribute on every method and member that [Mozilla](https://hg.mozilla.org/mozilla-central/file/tip/dom/webidl/) places a `[Throws]` attribute on.

cc @Liamolucko

### DIFF
--- a/crates/web-sys/webidls/enabled/Streams.webidl
+++ b/crates/web-sys/webidls/enabled/Streams.webidl
@@ -6,19 +6,21 @@
  *
  * The origin of this IDL file is
  * https://streams.spec.whatwg.org/#idl-index
+ * `[Throws]` attributes copied from Mozilla's webidl tree as of Aug. 31, 2022
+ * https://hg.mozilla.org/mozilla-central/file/tip/dom/webidl/
  */
 
 [Exposed=*, Transferable]
 interface ReadableStream {
-  constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
+  [Throws] constructor(optional object underlyingSource, optional QueuingStrategy strategy = {});
 
   readonly attribute boolean locked;
 
   Promise<undefined> cancel(optional any reason);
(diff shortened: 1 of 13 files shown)


crates/web-sys/webidls/enabled/Streams.webidl
```diff
@@ -6,19 +6,21 @@
...
   readonly attribute boolean locked;
 
   Promise<undefined> cancel(optional any reason);
-  ReadableStreamReader getReader(optional ReadableStreamGetReaderOptions options = {});
-  ReadableStream pipeThrough(ReadableWritablePair transform, optional StreamPipeOptions options = {});
+  [Throws] ReadableStreamReader getReader(optional ReadableStreamGetReaderOptions options = {});
+  [Throws] ReadableStream pipeThrough(ReadableWritablePair transform, optional StreamPipeOptions options = {});
   Promise<undefined> pipeTo(WritableStream destination, optional StreamPipeOptions options = {});
-  sequence<ReadableStream> tee();
+  [Throws] sequence<ReadableStream> tee();
```
Liamolucko, line 23: Although this is correct, it's a breaking change, since `ReadableStream` is already stable with methods which don't return `Result`. So, I think we'll have to leave these as being incorrectly infallible.

This isn't a problem for the other types because they were all previously unstable.

rozbb:
Sounds good! Will do

rozbb:
Done! Latest two PRs should resolve this.

hamza1311:
@rozbb can you also update this PR? There are merge conflicts

> Done! Latest two PRs should resolve this.

I assume this PR is still needed in addition to those, right?

Liamolucko:
> I assume this PR is still needed in addition to those, right? 

No, it isn't - #3065 included all the changes from this PR.

PR #3053, opened 2022-09-01
Return `false` from instanceof if the type being checked against doesn't exist

### LINKED ISSUE #2272: `web_sys::window` throws a `ReferenceError` in Web Workers, while it's likely meant to return `None`
### Describe the Bug

When `web_sys::window` is called inside a Web Worker, it throws a `ReferenceError: Window is not defined`, and so never returns. However, this function returns an `Option`, suggesting that it should return `None` when running in environments where `window` is not available.

### Steps to Reproduce

1. Create a new project with `wasm-pack new`
2. Add these dependencies to `Cargo.toml`:

   ```toml
   [dependencies]
   wasm-bindgen = "0.2"
   web-sys = { version = "0.3", features = ["Window"] }
   ```

3. Compile this function:

   ```rust
   use wasm_prelude::*;

   #[wasm_bindgen]
   pub fn test_window() -> bool {
     web_sys::window().is_some()
   }
   ```
4. Call this function inside a Web Worker

(Note: I actually tested this in context of my a bit odd, but open-source project. If needed, I will give instructions how to reproduce this bug there, as it makes the last step much easier)

### Expected Behavior

`test_window` returns `false` when run inside a Web Worker.

### Actual Behavior

`ReferenceError: Window is not defined` is printed to the console, and the function never returns.

### Additional Context

The error comes from `dyn_into::<Window>`. It calls this JS shim:

```js
export const __wbg_instanceof_Window_e8f84259147dce74 = function(arg0) {
    var ret = getObject(arg0) instanceof Window;
    return ret;
};
```

If the global object is not named `window`, then there's likely no `Window` class too. My suggestion is either that the `instanceof` check should be done in a `try/catch` block, or that `web_sys::window` should check if `Window` actually exists before downcasting.

Liamolucko:
Fixes #2272

Previously, checking if a value was an instance of a nonexistent type would result in an uncaught exception. This changes it to return `false` instead, avoiding having to do an extra feature-test of whether the type exists before calling `dyn_into`/`dyn_ref`/etc.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2956,7 +2956,20 @@ impl<'a> Context<'a> {
                 assert!(!variadic);
                 assert_eq!(args.len(), 1);
                 let js = self.import_name(js)?;
-                Ok(format!("{} instanceof {}", args[0], js))
+                write!(
+                    prelude,
+                    "\
+                    let result;
+                    try {{
+                        result = {} instanceof {};
+                    }} catch {{
+                        result = false;
+                    }}
+                    ",
+                    args[0], js,
+                )
+                .unwrap();
+                Ok("result".to_owned())
             }
 
             AuxImport::Static(js) => {


cschramm:
I just noticed that this code seems to raise the target of the generated code from supposedly ES2015 to ES2019. From what I can tell, it's nowhere documented what level is targeted (well, obviously at least ES2015), so it might be acceptable, but at least it seems unnecessary.

daxpedda:
Yeah, unfortunately there is no policy in place for a specific ES target and no checks are made in that regard.

I'm not familiar with the differences of these targets, but we can take a look at a PR that reduces the required target. In any case, as far as I can see this PR mitigated/fixed a certain bug, so this definitely needs to be preserved.

PR #3058, opened 2022-09-02
Add `TryFrom<JsValue>` for `[iu](64|128)`

### LINKED ISSUE #2350: Better support for u64 through JSValue::as_u64() and u64::into() -> JsValue

RReverser:
Adds a way to convert `JsValue` or a `BigInt` to `i64`/`u64`/`i128`/`u128` with type and range checks, returning the original `JsValue` otherwise.

This could be optimised a little bit further via more intrinsics, but it's good enough for the initial implementation, so leaving any optimisations for the future.

Fixes #2350.

### DIFF
--- a/crates/cli-support/src/intrinsic.rs
+++ b/crates/cli-support/src/intrinsic.rs
@@ -80,6 +80,10 @@ fn opt_f64() -> Descriptor {
     Descriptor::Option(Box::new(Descriptor::F64))
 }
 
+fn opt_i64() -> Descriptor {
+    Descriptor::Option(Box::new(Descriptor::I64))
+}
+
 intrinsics! {
     pub enum Intrinsic {
         #[symbol = "__wbindgen_jsval_eq"]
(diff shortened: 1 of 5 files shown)


RReverser:
Rebased.

Liamolucko (approved):

RReverser:
Any chance for a release sometime soon? I want to release new version of serde-wasm-bindgen that would depend on this PR + #3049.

Liamolucko:
v0.2.83 already includes those changes. (It doesn't seem to have been added as a proper GitHub release.)

RReverser:
> (It doesn't seem to have been added as a proper GitHub release.)

Ahh yeah that's what I was tracking / checking regularly, so missed a release.

cdecompilador:
Why not also adding `[iu](8|16|32)` to the pack?, I just needed for my application those implementations

RReverser:
> Why not also adding `[iu](8|16|32)` to the pack?, I just needed for my application those implementations

They are more complicated / would make a bit less sense because they're all represented by a generic `number` (Rust's `f64`) on the JS side, while bigints are all exact integers with separate type.

You can always convert from `f64` to specific integer type yourself though.

PR #3062, opened 2022-09-03
Add ability to instantiate WASM module without calling the start function

### LINKED ISSUE #2295: How to avoid extra calls to main when creating web worker from binary?
### Summary

For context: I've been experimenting with creating a library that spawns web workers that share memory with each other and the main thread.

I've been basing my experiments off of the Raytrace Parallel example: https://github.com/rustwasm/wasm-bindgen/tree/master/examples/raytrace-parallel

An issue I've run into is that `__wbindgen_start` is called whenever `init` is called. So the binary's `main` is invoked extra times whenever the library tries to spawn a worker and call `init` for it.

Is there some way to disable the automatic call to `__wbindgen_start` so that it can be invoked only for the main thread? Or should a special `init` be introduced for web workers? 

### Additional Details

Work arounds: 
* Require the binary using the library to insert some sort of check into main to detect that they're on a worker thread.
* Compile the binary as a `lib` and manually invoke `start`

daxpedda:
Currently it's impossible to properly implement WASM multi-threading as a library without the user having to use some workaround to prevent `main` to be called.

I was thinking of just adding a third parameter to the `init` function, not sure if that's a breaking change or not (I don't believe so). Happy to try other approaches to solve this issue.

Fixes #2295.

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -1459,7 +1459,7 @@ impl<'a> Context<'a> {
         let mut to_remove = Vec::new();
         for export in self.module.exports.iter() {
             match export.name.as_str() {
-                n if n.starts_with("__wbindgen") => {
+                n if n.starts_with("__wbindgen") && n != "__wbindgen_main" => {
                     to_remove.push(export.id());
                 }
                 _ => {}
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -43,12 +43,18 @@ struct InstructionBuilder<'a, 'b> {
     return_position: bool,
 }
 
+pub struct ProcessResult {
+    pub adapters: NonstandardWitSectionId,
+    pub aux: WasmBindgenAuxId,
+    pub start_found: bool,
+}
+
 pub fn process(
     module: &mut Module,
     externref_enabled: bool,
     wasm_interface_types: bool,
     support_start: bool,
-) -> Result<(NonstandardWitSectionId, WasmBindgenAuxId), Error> {
+) -> Result<ProcessResult, Error> {
     let mut storage = Vec::new();
     let programs = extract_programs(module, &mut storage)?;
 
@@ -88,7 +94,11 @@ pub fn process(
 
     let adapters = cx.module.customs.add(cx.adapters);
     let aux = cx.module.customs.add(cx.aux);
-    Ok((adapters, aux))
+    Ok(ProcessResult {
+        adapters,
+        aux,
+        start_found: cx.start_found,
+    })
 }
 
 impl<'a> Context<'a> {
@@ -476,22 +486,7 @@ impl<'a> Context<'a> {
             return Ok(());
         }
 
-        let prev_start = match self.module.start {
-            Some(f) => f,
-            None => {
-                self.module.start = Some(id);
-                return Ok(());
-            }
-        };
-
-        // Note that we call the previous start function, if any, first. This is
-        // because the start function currently only shows up when it's injected
-        // through thread/externref transforms. These injected start functions
-        // need to happen before user code, so we always schedule them first.
-        let mut builder = walrus::FunctionBuilder::new(&mut self.module.types, &[], &[]);
-        builder.func_body().call(prev_start).call(id);
-        let new_start = builder.finish(Vec::new(), &mut self.module.funcs);
-        self.module.start = Some(new_start);
+        self.module.exports.add("__wbindgen_main", id);
         Ok(())
     }
 
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -97,7 +97,7 @@ pub fn process(
     Ok(ProcessResult {
         adapters,
         aux,
-        start_found: cx.start_found,
+        start_found: cx.start_found && cx.support_start,
     })
 }
 
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -471,7 +471,7 @@ impl<'a> Context<'a> {
             OutputMode::Web => {
                 self.imports_post.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, Some(&mut imports))?;
-                footer.push_str("export { initSync }\n");
+                footer.push_str("export { initSync, initWithoutStart }\n");
                 footer.push_str("export default init;");
             }
         }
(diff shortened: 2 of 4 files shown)


Liamolucko (changes requested):
`__wbindgen_start` doesn't just call `main`; it also does stuff like initializing the stack and thread-local storage, that need to be called on every thread. So, you'll have to split those runtime-initialization things that need to be called by every thread out into a different function.

This should be the relevant part to change: https://github.com/rustwasm/wasm-bindgen/blob/f75a3f8986c56d752b687cebe47b53232f403ba7/crates/cli-support/src/wit/mod.rs#L468-L496

It's called for both `main` and `#[wasm_bindgen(start)]` functions.

daxpedda:
I didn't realize this, which caused me to encounter a lot of bugs that I couldn't really explain. Thanks for pointing that out!

I hope I got this right:
I changed `add_start_function()` to just add a new exported function called `__wbindgen_main` instead of modifying the start section. This leaves the initialization code untouched to be moved later to `__wbindgen_start` by `unstart_start_function()`.

Tested and works great!

crates/cli-support/src/wit/mod.rs
```diff
@@ -476,22 +486,7 @@ impl<'a> Context<'a> {
...
-
-        // Note that we call the previous start function, if any, first. This is
-        // because the start function currently only shows up when it's injected
-        // through thread/externref transforms. These injected start functions
-        // need to happen before user code, so we always schedule them first.
-        let mut builder = walrus::FunctionBuilder::new(&mut self.module.types, &[], &[]);
-        builder.func_body().call(prev_start).call(id);
-        let new_start = builder.finish(Vec::new(), &mut self.module.funcs);
-        self.module.start = Some(new_start);
+        self.module.exports.add("__wbindgen_main", id);
```
daxpedda, line 489: I didn't actually test if the normal `init` function works now, for some reason `__wbindgen_main` doesn't end up in the final WASM file. Is it being cleaned up somehow? Or is `module.exports.add()` not sufficient?

daxpedda: Indeed it was being cleaned up in `unexport_intrinsics`, I hope the way I fixed this is appropriate.

daxpedda:
It took me quiet some time to figure out what's going on, but I don't think I'm introducing a deep change by any stretch of the imagination.
Of course there could still be a bug that affects the major use-case of `wasm-bindgen`. So obviously a thorough review is necessary.

I would also like to mention that I'm working on a library like [`wasm_thread`](https://crates.io/search?q=wasm-thread) that should make things very usable, I'm using it right now and it's working completely abstracted without me having to write a ton of `cfg`s. I believe this PR and #3032 are the last missing pieces.

---

I think everybody in the Rust Community knows and appreciates the amount of work and all the responsibilities that you have. So I totally understand if you can't find time for this :heart:.

Is there somebody else that can be asked to review this?

Liamolucko:
A potential way for this to work without adding an `initWithoutStart` function would be for `wasm-bindgen` to only call `main` on the first thread that calls `init`. That would boil down to setting a byte to 1 (atomically) when the first thread starts, which all the other threads will see and skip running `main`.

That does have the potential to break any code that relies on the current behaviour of calling `main` on every call to `init`, though.

daxpedda:
If the desire is to not add another function, I think a solution that would have much less breakage is to add a third default argument to `init`, I assume that fewer people use weird syntax to pass arguments into `init` then people relying on `init` to always call `main`. Not that any of these solutions would alleviate any issues alexcrichton was having.

@Liamolucko I didn't realize that you have merge rights, is this PR mergeable? Can you review this? Is there some other solution to alleviate alexcrichton's concerns?

hamza1311:
I agree with Alex here.  
There's no true support for multi threading in WASM, and we should be applying bandaids to fill the hole.

As long *a* workaround is available, I wouldn't be in favor of merging this. Also inclusion of this function increases the JS bundle size, which is something worth considering

daxpedda:
> There's no true support for multi threading in WASM, and we should be applying bandaids to fill the hole.

I'm assuming you meant "we should **not**"?

I will attempt to make a case for this:

> As long _a_ workaround is available, I wouldn't be in favor of merging this.

I am currently building a library very much like [`wasm_thread`](https://crates.io/crates/wasm_thread) and currently there is no workaround that I can apply in the library itself. I have to tell users not to use a start function and to manually call their main function in their `index.html` to start their application. This is a very bad footgun.

I myself have hit that footgun when trying to use `wasm_thread` and I didn't understand why I'm getting some really bad infinite loops until I discovered this issue. It also prevents users from using any kind of tool that automatically starts up their application, like `trunk`, `wasm-pack` or `wasm-server-runner`.

So I don't really consider there to be a viable workaround, the only available one doesn't just introduce a footgun, it also locks you out from a lot of convenient tools that you can't use anymore.
> Also inclusion of this function increases the JS bundle size, which is something worth considering

Is the JS bundle the JS shim? I just measured it and you are right, it's noticeable. In a completely minimal "Hello, World!" example it's 5% both not-minimized and minimized. I must say though that in any reasonably sized project it's much less then 1%. Also `initSync` should have introduced an even bigger overhead.

---

From [alexcrichton](https://github.com/alexcrichton):
> AFAIK the current support implemented by wasm-bindgen is a sort of "local maximum" which while it's pretty unusable that's somewhat reflective of multithreading on the web at large.

I do know that currently multithreading in the browser through WASM has a lot of limitations and issues. But I have to point out, that it is currently underused because of two issues:
- It requires nightly, because of `-Zbuild-std`. See https://github.com/rust-lang/rust/issues/77839.
- There is no usable library to easily use threads from Rust code.

I can't really change anything about the nightly issue, but I can contribute to making it usable for now. [`wasm_thread`](https://crates.io/crates/wasm_thread) is a good example of what's currently possible, but it simply has too many footguns to be usable for the average user. I was able to remove all "major" footguns except the one addressed by this PR and I do consider it the last thing really necessary to make things as seamless to the user as possible.

Liamolucko (commented):
The implementation of this looks good, I think it's more a question of the API.

I agree that `initWithoutStart` feels like a bandaid. The way I'd prefer is the way I suggested where it's handled automatically, provided that that doesn't create much breakage. Failing that, I think that adding a third boolean parameter is slightly less of a bandaid, but it's still not great.

daxpedda:
> The way I'd prefer is the way I suggested where it's handled automatically, provided that that doesn't create much breakage.

I will look into it, I really don't mind this solution, it fits my needs very well.

daxpedda:
Replaced by #3092.

PR #3065, opened 2022-09-03
web-sys: Added Streams API

rozbb:
Builds on revert https://github.com/rustwasm/wasm-bindgen/pull/3064. This adds the Streams API again, but without the API-breaking `[Throws]` attributes


rozbb:
Note that the current diff view is wrt `main` as opposed to https://github.com/rustwasm/wasm-bindgen/pull/3064

rozbb:
Also, it'd be nice if someone could cut a new `web-sys` version some time soon :)
For a Yew website, the build process when switching to `HEAD` requires adding 4 patch lines and reinstalling `wasm-bindgen-cli` to match the new verison.

crates/web-sys/webidls/enabled/Streams.webidl
```diff
@@ -1,11 +1,15 @@
...
 /* Copyright © WHATWG (Apple, Google, Mozilla, Microsoft). This work is licensed under a Creative
  * Commons Attribution 4.0 International License. To the extent portions of it are incorporated into
  * source code, such portions in the source code are licensed under the BSD 3-Clause License
  * instead.
  *
  * The origin of this IDL file is
  * https://streams.spec.whatwg.org/#idl-index
+ * `[Throws]` attributes copied from Mozilla's webidl tree as of Aug. 31, 2022
+ * https://hg.mozilla.org/mozilla-central/file/tip/dom/webidl/
+ * To avoid API breakage, `[Throws]` attributes are ommitted from `ReadableStream` methods
```
Liamolucko, line 12:
```suggestion
 * To avoid API breakage, `[Throws]` attributes are omitted from `ReadableStream` methods
```

crates/web-sys/webidls/enabled/Streams.webidl
```diff
@@ -71,10 +75,10 @@ interface mixin ReadableStreamGenericReader {
 
 [Exposed=*]
 interface ReadableStreamDefaultReader {
-  constructor(ReadableStream stream);
+  [Throws] constructor(ReadableStream stream);
 
   Promise<ReadableStreamReadResult> read();
-  undefined releaseLock();
+  [Throws] undefined releaseLock();
```
Liamolucko, line 81: I don't see any way in the spec that this can fail... looking at Mozilla's implementation, though, it can throw an internal error, which is probably why this is here.

rozbb:
Thanks for being so thorough! I've gone through and removed all the unnecessary `[Throws]`. Just wanted to double check two instances:

1. The definition of `WriteableStream.abort` (and other `abort`s) is
```
If ! IsWritableStreamLocked(this) is true, return a promise rejected with a TypeError exception.
```
Does this count as a Throw? I don't think it does.

2. The member `readonly attribute Function size` is defined as
```
Return this's relevant global object's count queuing strategy size function.
```
This is also not a throwing function, right?

crates/web-sys/webidls/enabled/Streams.webidl
rozbb: Addressed in new commits

Liamolucko:
>     1. The definition of `WriteableStream.abort` (and other `abort`s) is
> 
> 
> ```
> If ! IsWritableStreamLocked(this) is true, return a promise rejected with a TypeError exception.
> ```
> 
> Does this count as a Throw? I don't think it does.

No, it doesn't, since marking it `[Throws]` would be interpreted as a synchronous error.

>     2. The member `readonly attribute Function size` is defined as
> 
> 
> ```
> Return this's relevant global object's count queuing strategy size function.
> ```
> 
> This is also not a throwing function, right?

Yes. The returned function could throw, but getting it can't.

crates/web-sys/webidls/enabled/Streams.webidl
```diff
@@ -0,0 +1,232 @@
...
+callback TransformerFlushCallback = Promise<undefined> (TransformStreamDefaultController controller);
+callback TransformerTransformCallback = Promise<undefined> (any chunk, TransformStreamDefaultController controller);
+
+[Exposed=*]
+interface TransformStreamDefaultController {
+  readonly attribute unrestricted double? desiredSize;
+
+  [Throws] undefined enqueue(optional any chunk);
+  undefined error(optional any reason);
+  [Throws] undefined terminate();
```
Liamolucko, line 199: I don't think this one can fail.

Liamolucko (commented):
I've just noticed one more thing that I don't think can throw, but after that I think this should be good to merge.

I didn't go through all of them, I just glanced at the ones which weren't duplicated across types (so i.e. not `close`, `enqueue` etc.)

crates/web-sys/webidls/enabled/Streams.webidl
rozbb: I also did a double take but yes it can: https://streams.spec.whatwg.org/#ts-default-controller-terminate

rozbb: Oh wait you're right, it can't

rozbb:
Thanks again, good catch. Should be fixed now.

Liamolucko (approved):
Thanks for all your work on this! And sorry about the whole fuss with having to revert.

PR #3068, opened 2022-09-05
Refactor: normalized encoding of ImportModule

lukaslihotzki:
With uninterpreted paths (not starting with `./`, `../`, or `/`), `ImportModule::Named` and `ImportModule::RawNamed` are equivalent (because the attributes are too). However, while the attributes should be designed for user convenience, the binary encoding should be optimized for simple processing. Attribute handling should be done purely in the `macro-support` crate. Therefore, it makes sense to normalize uninterpreted paths to `ImportModule::RawNamed` in the binary encoding.

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -880,13 +880,6 @@ impl<'a> Context<'a> {
     }
 
     fn determine_import(&self, import: &decode::Import<'_>, item: &str) -> Result<JsImport, Error> {
-        let is_local_snippet = match import.module {
-            Some(decode::ImportModule::Named(s)) => self.aux.local_modules.contains_key(s),
-            Some(decode::ImportModule::RawNamed(_)) => false,
-            Some(decode::ImportModule::Inline(_)) => true,
-            None => false,
-        };
-
         // Similar to `--target no-modules`, only allow vendor prefixes
         // basically for web apis, shouldn't be necessary for things like npm
         // packages or other imported items.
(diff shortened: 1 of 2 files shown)


lukaslihotzki:
This change helps with correctly supporting all kinds of paths in #3034.

crates/cli-support/src/wit/mod.rs
```diff
@@ -880,21 +880,14 @@ impl<'a> Context<'a> {
...
-
         // Similar to `--target no-modules`, only allow vendor prefixes
         // basically for web apis, shouldn't be necessary for things like npm
         // packages or other imported items.
         let vendor_prefixes = self.vendor_prefixes.get(item);
         if let Some(vendor_prefixes) = vendor_prefixes {
             assert!(vendor_prefixes.len() > 0);
 
-            if is_local_snippet {
+            if let Some(decode::ImportModule::Inline(_)) = &import.module {
```
Liamolucko, line 890: Shouldn't this be
```suggestion
            if let Some(decode::ImportModule::Inline(_) | decode::ImportModule::Named(_)) = &import.module {
```
since `Named` is now always a local snippet?

Liamolucko: Also, with that change, the below check for `ImportModule::Named` will never fire; I think it should be replaced with a check for `ImportModule::RawNamed`.

I think it's a mistake that it doesn't currently include `ImportModule::RawNamed`: that block of code predates the addition of `ImportModule::RawNamed` in #1353, and only fires when the module behaves the same as `ImportModule::RawNamed`.

Liamolucko (approved):

PR #3069, opened 2022-09-05
Add support for linked modules

### LINKED ISSUE #2335: Lazy loading of JS snippets

lukaslihotzki:
Like imported modules, linked modules can be added by any crate with the `wasm_bindgen::link_to!` macro and the existing `module`, `raw_module` or `inline_js` attributes. In contrast to imported modules, no imports are added to the JS shim generated by wasm-bindgen. Instead, the macro expands to an expression of type `String` that contains a link. This link may be relative and is suitable for creating workers or worklets, or dynamic imports.
Internally, the macro invocation `wasm_bindgen::link_to!(module = "/src/task/worker.js")` expands to code like (simplified):
```
{
	pub static _GENERATED: [u8; …] = …; // Program in custom section
	extern "C" {
		fn __wbindgen_link_9fdd114fc6bc2308() -> String; // link function
	}
	__wbindgen_link_9fdd114fc6bc2308()
}
```
This link function is implemented in the JS shim:
```
    imports.wbg.__wbindgen_link_9fdd114fc6bc2308 = function(arg0) {
        const ret = new URL('snippets/wasm-bindgen-futures-b12f15e5e844bdeb/src/task/worker.js', import.meta.url).toString();
        …
    };
```
The main reason for this approach is that it works both without bundlers and with bundlers (like webpack >= 5) that detect the `new URL(…, import.meta.url)` syntax, bundle and minify the linked file, and change the link. Depending on the target, the correct alternative for `import.meta.url` is used.

This is a useful part of #3034. In contrast to #3034, importing the JS shim is not supported, so this PR can only be used for pure JS workers or worklets (not considering workarounds).

Fixes #2335 (Just create a script element or dynamically import your linked module before using JS classes from Rust)


alexcrichton:
Apologies but I personally do not have the time to adequately review new major features like this. Others may be able to but I will be unable to provide review or issue support for this if it lands.

hamza1311:
Requesting my own review so I don't forget. This is something I've experimented with and can be very useful when it comes to lazy loading modules

lukaslihotzki:
I needed to rebase because of #3073. Because the `link_to!` macro is new, strict attribute checking can be enabled without breaking change.

hamza1311 (commented):
The implementation looks fine to me. It's worth noting that I'm not too familiar with this part of the code base so I could've missed something important.

A couple of things are missing here. Can you add tests and documentation for this feature?
It's not really possible to update the book[^1] so adding rustdoc comments (including an example) to the macro would be fine.

[^1]: I tried reaching out to @fitzgen for push access to the repo(s) but couldn't get a hold of them. If anyone else seeing this can grant me access, that would be great.

crates/backend/src/codegen.rs
```diff
@@ -130,6 +130,29 @@ impl TryToTokens for ast::Program {
     }
 }
 
+impl TryToTokens for ast::LinkToModule {
+    fn try_to_tokens(&self, tokens: &mut TokenStream) -> Result<(), Diagnostic> {
+        let mut program = TokenStream::new();
+        self.0.try_to_tokens(&mut program)?;
+        let link_function_name = self.0.linked_modules[0].link_function_name();
+        let name = Ident::new(&link_function_name, Span::call_site());
+        let abi_ret = quote! { <String as wasm_bindgen::convert::FromWasmAbi>::Abi };
```
Liamolucko, line 139:
```suggestion
        let abi_ret = quote! { <std::string::String as wasm_bindgen::convert::FromWasmAbi>::Abi };
```

crates/backend/src/codegen.rs
```diff
@@ -130,6 +130,29 @@ impl TryToTokens for ast::Program {
...
+        let extern_fn = extern_fn(&name, &[], &[], &[], abi_ret);
+        (quote! {
+            {
+                #program
+                #extern_fn
+
+                unsafe {
+                    <String as wasm_bindgen::convert::FromWasmAbi>::from_abi(#name())
```
Liamolucko, line 147:
```suggestion
                    <std::string::String as wasm_bindgen::convert::FromWasmAbi>::from_abi(#name())
```

crates/macro/src/lib.rs
```diff
@@ -18,6 +18,26 @@ pub fn wasm_bindgen(attr: TokenStream, input: TokenStream) -> TokenStream {
     }
 }
 
+/// This macro adds a linked module by module, raw_module or inline_js attribute.
+/// It expands to a String containing a link to that module. This link may be relative and
```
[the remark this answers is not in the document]
lukaslihotzki: I think I will make the output URLs always absolute. Relative URLs seemed nice because they usually can be used as paths in node. On the other hand, ensuring correctness even with `chdir` or `history.replace` seems difficult. More importantly, in the case when you `chdir` to another drive on Windows, you cannot return a string that is both a path and a relative URL referring to a linked module. Therefore, `link_to!` should just return absolute URLs, and users on node can simply call `url.pathToFileURL` if needed.

lukaslihotzki:
I switched to always absolute URLs so `chdir` or `history.replaceState` don't break the links. I added documentation, tests, all specific suggestions, and a fix for inline_js (covered by the tests).

Liamolucko:
I don't really think that `link_to!` should inherit the same way of specifying modules as `#[wasm_bindgen]`, because some of the options don't really make sense - e.g., I don't think `wasm_bindgen::link_to!(raw_module = "foo")` makes much sense.

As currently implemented, that would return the URL of a file named `foo` in the directory of the wasm-bindgen glue code, which doesn't seem like something you'd ever want to do.

An alternate way for it to work would be for it to return `"foo"` untouched, so that importing the URL would result in `import("foo")`, but that isn't really an actual URL and wouldn't work when passed to a `Worker`, which is the main purpose of this API. Also, you may as well skip `link_to!` and write `"foo"` directly.

So, I think there should be two ways of using the API:

- `wasm_bindgen::link_to!(inline_js = "...")` (same as current version)
- `wasm_bindgen::link_to!("./foo.js"), which expands to `wasm_bindgen::link_to!(inline_js = include_str!("./foo.js"))`

Note that you might need to use a `macro_rules!` wrapper over a private proc macro to accomplish the second one, since `include_str!` resolves relative to the current source file and I don't believe there's any way to do that from within a proc macro.

lukaslihotzki:
> I don't think `wasm_bindgen::link_to!(raw_module = "foo")` makes much sense. As currently implemented, that would return the URL of a file named foo in the directory of the wasm-bindgen glue code, which doesn't seem like something you'd ever want to do.

Relative paths (`./`, `../`) and absolute paths in both `module` and `raw_module` attributes are consistently handled like imported modules, regardless of using webpack or not. For module paths (not starting with `./`, `../`, `/`) it is slightly more complex.

Webpack implements module resolution on `new URL('…', import.meta.url)` as a fallback. So it would return a link to the `.`-export of package `foo` if the file `foo` does not exist. Browsers would also pick up the file if it exists and fail otherwise (failure is the correct outcome if you forgot to run the bundler). I think this ambiguity is acceptable in contrast to not supporting module paths at all. Most likely, file extensions are used to disambiguate between modules (and their exports) and files.

However, this functionality (for imported and for linked modules) is also be usable with just the `module` attribute. I think the `raw_module` attribute was introduced precisely to allow paths relative to the output glue code, so users can supply their own scripts. I can't see another use case for the existing `raw_module` attribute. If this is supported in imported modules, it should also be supported in linked modules for consistency.

> So, I think there should be two ways of using the API:

I would not like a syntax that sometimes uses attribute names (for inline_js) and sometimes does not (for file names). Also, I can't see how your proposal with a `macro_rules!` wrapper would work. The proc macro would receive a token stream literally containing `include_str!("…")`, which it cannot process without [eager expansion](https://rustc-dev-guide.rust-lang.org/macro-expansion.html#eager-expansion), which proc macros cannot do. Also, I like that file names of linked JS modules are also used for the output snippets (useful for debugging). File names would be lost when expanding to `inline_js`.

lukaslihotzki:
It would be okay to disallow some cases by `compile_error!` in this PR first, so they can be discussed and maybe allowed later. However, all existing options are reasonable (at least as reasonable as in `#[wasm_bindgen]`) and do work, so they should be expressible in the `link_to!` API. Therefore, we shouldn't treat "foo" as "./foo" for example.

lukaslihotzki:
It can be argued that module paths in `link_to!` may lead to surprising behavior because they only do module resolution as a fallback if the file does not exist. Therefore, I disallowed module paths to get this PR merged quickly. A future PR can allow them without breaking change (or even interpret them as relative paths instead).

lukaslihotzki:
@Liamolucko, as this PR seems to be stalled again, would you merge it if only the `inline_js` attribute was allowed? (There, we proposed exactly the same syntax.)

Liamolucko:
> @Liamolucko, as this PR seems to be stalled again, would you merge it if only the `inline_js` attribute was allowed? (There, we proposed exactly the same syntax.)

Oh, sorry, this had just completely slipped my mind; that wasn't the problem.

I didn't realise that Webpack supported resolving packages with `new URL(...)`; I thought it just behaved the same as it would un-bundled and used it as a purely as a hint to include the referenced files. Knowing that, module paths make more sense, although I think it still makes sense to leave them to a follow-up to work out the specifics (e.g. should resolving packages work in Node too?)

So, in light of that, I think the design of the macro is fine as-is. I'll just take another look at the code before merging this.

crates/macro/src/lib.rs
```diff
@@ -18,6 +18,26 @@ pub fn wasm_bindgen(attr: TokenStream, input: TokenStream) -> TokenStream {
...
+#[proc_macro]
+pub fn link_to(input: TokenStream) -> TokenStream {
+    match wasm_bindgen_macro_support::expand_link_to(input.into()) {
+        Ok(tokens) => {
+            if cfg!(feature = "xxx_debug_only_print_generated_code") {
+                println!("{}", tokens);
+            }
+            tokens.into()
+        }
+        Err(diagnostic) => (quote! { String::clone(#diagnostic) }).into(),
```
Liamolucko, line 37: Why is this `String::clone` here?

crates/macro/src/lib.rs
```diff
@@ -18,6 +18,26 @@ pub fn wasm_bindgen(attr: TokenStream, input: TokenStream) -> TokenStream {
     }
 }
 
+/// This macro adds a linked module by module, raw_module or inline_js attribute.
+/// It expands to a String containing a URL to that module. This URL can be used
+/// to create workers or worklets, for example:
```
Liamolucko, line 23: I don't feel like this does a great job of explaining what the macro actually does.

Maybe something like:

```suggestion
/// This macro takes a JS module as input and returns a URL that can be used to
/// access it at runtime.
///
/// The module can be specified in a few ways:
/// - You can use `inline_js = "..."` to create an inline JS file.
/// - You can use `module = "/foo/bar"` to reference a file relative to the
/// root of the crate the macro is invoked in.
///
/// The returned URL can be used for things like creating workers/worklets:
///
```

crates/macro/src/lib.rs
lukaslihotzki: The invalid expression should have the type `String` so the compiler or the analyzer can still typecheck the remaining program, even if the arguments are invalid.

lukaslihotzki:
@Liamolucko, do you have any other concerns?

Liamolucko:
> @Liamolucko, do you have any other concerns?

Yeah - I'm a bit concerned that this won't work with any bundlers other than Webpack. I don't want people who are using esbuild, rollup, etc. to have their code suddenly stop working in browsers other than Chrome because `wasm-bindgen-futures` switched to using this for its `Atomics.waitAsync` polyfill. There isn't any way that I'm aware of to get this to work universally, though.

I suppose a compromise could be to disable it in `wasm-bindgen-futures` by default, and then make it clear in the docs for `link_to!` that you might need to manually configure your bundler to include the linked modules, and then people who specifically choose to use it can do that.

lukaslihotzki:
Other bundlers are a valid concern. Luckily, it can be solved at bindgen time, because the generated JS import code can choose to return a data URI. It took me some time though to figure out embedding the content in `AuxImport::LinkTo` is a lot nicer than trying to pass `WasmBindgenAux` to the import code generator. Compared to compile time approaches, this prevents a dependency to set the feature unexpectedly. Also, this guarantees a consistent link-or-embed behavior and allows to display explanatory text if links are needed. Alternatively, links could be allowed automatically if embedding is impossible (maybe displaying a hint).

@Liamolucko, I'm sorry for the delay. Do you accept this approach? If so, I'll have to add tests with embedding probably. Is there anything else missing (I don't think so)?

crates/cli-support/src/js/mod.rs
```diff
@@ -3143,6 +3143,38 @@ impl<'a> Context<'a> {
...
+                } else {
+                    if let Some(content) = content {
+                        Ok(format!(
+                            "\"data:application/javascript,\" + encodeURIComponent(`{}`)",
+                            content.replace('`', "\\`")
+                        ))
+                    } else {
+                        Err(anyhow!("wasm-bindgen needs to be invoked with `--allow-links`, because \"{}\" cannot be embedded.\n\
+                             `--allow-links` is safe with webpack 5 or no bundler at all.\n\
+                             For other bundlers, ensure they support the `new URL('…', import.meta.url)` syntax.", path))
```
Liamolucko, line 3174: I don't know if 'ensure they support' is the right phrasing here. Often the solution would instead be to install a plugin which adds support, or to somehow tell the bundler to copy all the linked modules into the output directory where they're expected to be relative to the shim.

Liamolucko (commented):
Good idea to put it behind a flag! I think that works out well; by default, everything 'just works', and then if you want the stricter CSP and/or lazy loading, you can manually configure your bundler to make things work and turn on the flag.

crates/cli-support/src/wit/mod.rs
```diff
@@ -340,6 +341,47 @@ impl<'a> Context<'a> {
...
+                    self.unique_crate_identifier,
+                    *idx as usize + offset
+                ),
+                Some(inline_js[*idx as usize]),
+            ),
+        };
+        self.aux.import_map.insert(
+            id,
+            // content is embedded as data URI, so it should not be too long.
+            AuxImport::LinkTo(path, content.filter(|x| x.len() <= 512).map(str::to_string)),
```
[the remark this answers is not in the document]
lukaslihotzki: Thinking about this again, I agree. Huge data URLs feel somehow inefficient and wrong, but a hard limit is unnecessary. Data URLs are easy, and the clean, efficient solution can be enabled if needed.

(If huge data URLs were shown to cause problems, bindgen could still switch to blob URLs transparently. Let's keep it simple until then.)

Liamolucko (commented):
Looks good. I just have some nitpicks about the phrasing of the docs.

crates/macro/src/lib.rs
Liamolucko: You should probably add a comment about it, though:

```suggestion
        // This `String::clone` is here so that IDEs know this is supposed to be a
        // `String` and can keep type-checking the rest of the program even if the macro
        // fails.
        Err(diagnostic) => (quote! { String::clone(#diagnostic) }).into(),
```

lukaslihotzki:
@Liamolucko, I have overlooked the other suggestion. It is committed now.

Liamolucko (approved):
Thank you for all of your work on this!

Issue #3038, opened 2022-08-20
`strict-macro` checking should be enabled by default


lukaslihotzki:
### Describe the Bug

Unapplicable attributes are accepted inside wasm_bindgen `impl` and `extern` blocks.

### Steps to Reproduce

```
#[wasm_bindgen]
pub struct Foo;

#[wasm_bindgen]
impl Foo {
    #[wasm_bindgen(inline_js = "hello world", raw_module = "something.js")]
    pub fn foo() {
    }
}

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(inline_js = "hello world", raw_module = "something.js")]
    pub fn bar();
}
```

### Expected Behavior

Compile errors in `Foo::foo` and `bar` because the attributes `inline_js` and `raw_module` are not applicable for these items.

### Actual Behavior

These attributes are silently ignored.

Liamolucko:
`wasm-bindgen` already supports checking for this when the `strict-macro` feature is enabled:

```
error: unused #[wasm_bindgen] attribute
 --> src/lib.rs:8:20
  |
8 |     #[wasm_bindgen(inline_js = "hello world", raw_module = "something.js")]
  |                    ^^^^^^^^^

error: unused #[wasm_bindgen] attribute
 --> src/lib.rs:8:47
  |
8 |     #[wasm_bindgen(inline_js = "hello world", raw_module = "something.js")]
  |                                               ^^^^^^^^^^

error: unused #[wasm_bindgen] attribute
  --> src/lib.rs:14:20
   |
14 |     #[wasm_bindgen(inline_js = "hello world", raw_module = "something.js")]
   |                    ^^^^^^^^^

error: unused #[wasm_bindgen] attribute
  --> src/lib.rs:14:47
   |
14 |     #[wasm_bindgen(inline_js = "hello world", raw_module = "something.js")]
   |                                               ^^^^^^^^^^

error: could not compile `wasm-playground` due to 4 previous errors
```

So, perhaps that should be enabled by default? It would be a breaking change to do that, but it could also be argued that any code that does that is mistaken.

To do it non-breakingly, it could be a warning by default instead.

lukaslihotzki:
Can proc_macros warn in stable Rust?

lukaslihotzki:
[...]

PR #3070, opened 2022-09-05
Enforce strict attribute checking

lukaslihotzki:
Enforce strict attribute checking as I proposed in #3038. Enabling features should not break code, because the crate enabling the feature and the crate whose code breaks may be unrelated. This change is also allowed without major update by the [SemVer Compatibility chapter](https://doc.rust-lang.org/cargo/reference/semver.html) of the Cargo book:

> Some changes are marked as "minor", even though they carry the potential risk of breaking a build. This is for situations where the potential is extremely low, and the potentially breaking code is unlikely to be written in idiomatic Rust, or is specifically discouraged from use.

I don't know whether the precedence is stronger in the "or" or in the "and". In any case, potentially breaking code is unlikely to be written in idiomatic Rust because the idiomatic tool for specifying information with no effect are comments, not attributes. And such code is specifically discouraged from use, because the `strict-macro` feature is a form of discourgement.

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -94,7 +94,6 @@ macro_rules! methods {
     ($(($name:ident, $variant:ident($($contents:tt)*)),)*) => {
         $(methods!(@method $name, $variant($($contents)*));)*
 
-        #[cfg(feature = "strict-macro")]
         fn check_used(self) -> Result<(), Diagnostic> {
             // Account for the fact this method was called
             ATTRS.with(|state| state.checks.set(state.checks.get() + 1));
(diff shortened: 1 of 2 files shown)


alexcrichton:
Personally I'm not confident in doing this. Focusing on the impact of a change like this, I don't think there's any way to really gauge how much breakage is going to happen since it's not easy to enumerate the list of users of `wasm-bindgen`. I don't really want to lawyer our way into making a change like this, instead I'd prefer to focus on the practical impacts.

I completely agree that the existence of this feature is not how Cargo features are supposed to work. It is indeed not additive. There's nothing that can really be done about that without a new major version which I'm reluctant to do.

This seems like the lesser of two evils for now, personally. If/when an 0.3 or new major version bump ever happens this feature should of course be removed and enabled by default, I don't think anyone is disputing that.

RReverser:
Here's an alternative approach, sort of a compromise: https://github.com/rustwasm/wasm-bindgen/pull/3073. @lukaslihotzki @alexcrichton would love your thoughts.

lukaslihotzki:
Ok, let's merge #3073 instead.

PR #3072, opened 2022-09-06
Put FontFace constructor to Immutable slice whitelist

namse:
FontFace's constructor `new_with_u8_array` takes mutable slice which could be immutable slice. So I added it to whitelist.

It's my first time so I followed ImageData cases.


alexcrichton (approved):

namse:
Hurray! A small contribution to the great footsteps.

PR #3075, opened 2022-09-09
Export `initSync` for no-module builds

daxpedda:
Only the web target currently had `initSync` exported. This exports the function for no-module builds too.

Follow-up on #3016.


hamza1311 (approved):
Looks good :+1:

PR #3067, opened 2022-09-05
Update WebGPU's webidl

nical:
Regenerated the webidl from the current state of the spec.

PR #3079, opened 2022-09-12
Bumped crate versions

rozbb:
Following https://github.com/rustwasm/wasm-bindgen/pull/3067#issuecomment-1238338577 I just ran `./publish bump`. I did have to modify `publish.rs` though because `example-tests` was a new crate and did not fall into the allowlist or blocklist.


alexcrichton (approved):
Thanks!

PR #3084, opened 2022-09-15
Add no-modules support to wasm-bindgen-test-runner

daxpedda:
I need `--no-module` support for `wasm-bindgen-test-runner` and found no obvious way to achieve it otherwise.
`WASM_BINDGEN_USE_NO_MODULE` is modeled after `WASM_BINDGEN_USE_DENO`.

Potentially we may want to add a check to make sure that you can't set `WASM_BINDGEN_USE_NO_MODULE` without correctly setting up `wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);`?

Alternatively this could be moved to `wasm-bindgen` itself or to `wasm_bindgen_test_configure!()`?

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/index-headless.html
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/index-headless.html
@@ -36,6 +36,6 @@
 
      window.__wbg_test_invoke = f => f();
     </script>
-    <script src='run.js' type=module></script>
+    <!-- {IMPORT_SCRIPTS} -->
   </body>
 </html>
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/index.html
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/index.html
@@ -24,6 +24,6 @@
 
      window.__wbg_test_invoke = f => f();
     </script>
-    <script src='run.js' type=module></script>
+    <!-- {IMPORT_SCRIPTS} -->
   </body>
 </html>
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -33,6 +33,7 @@ enum TestMode {
     Node,
     Deno,
     Browser,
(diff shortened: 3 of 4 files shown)


daxpedda:
Found a bug, working on it ...

daxpedda:
Confirmed that it actually works correctly now.

hamza1311 (commented):
The implementation looks good to me. Not sure about the check for ensuring that wasm-bindgen-test is configured properly. What happens if you run this without the `wasm_bindgen_test_configure!(run_in_browser)`?

Also, I'm curious, what's the use-case where you need `--no-module` for tests?

daxpedda:
> Not sure about the check for ensuring that wasm-bindgen-test is configured properly. What happens if you run this without the `wasm_bindgen_test_configure!(run_in_browser)`?

It will ignore `WASM_BINDGEN_USE_NO_MODULE` and fall through to use Node or if you used `WASM_BINDGEN_USE_DENO` it will use Deno.
I also just noticed that if you use `wasm_bindgen_test_configure!(run_in_browser)` and `WASM_BINDGEN_USE_DENO`, it will also simply ignore `WASM_BINDGEN_USE_DENO`.

> Also, I'm curious, what's the use-case where you need `--no-module` for tests?

I'm building a library very similar to [`wasm_thread`](https://crates.io/crates/wasm_thread) for spawning web workers from within Rust, without requiring the user to add any additional JS code. Currently Firefox does not support ES modules in workers ([bugzilla issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687)), so to load the JS shim into the web worker, I have to generate it with `--no-module`.

Alternatively of course, there are some JS scripts out there that are able to convert an ES module into a regular JS script, but I decided to avoid that.

daxpedda:
@hamza1311 friendly ping.

hamza1311 (approved):
Looks good to me. Thanks for the ping (and clarification)

I'm going to go ahead and merge this. It's been sitting for a while and is quite a useful feature 

CC: @futursolo, any thoughts on this? Can this be used for automated testing with gloo-worker

PR #3087, opened 2022-09-18
add JsValue::is_array() + tests

### LINKED ISSUE #3086: Add `is_array()` method to `JsValue`
### Motivation

In my library tests, I want to assert that a `JsValue` is an `Array`.

### Proposed Solution

Add `JsValue::is_array()` method.

### Alternatives

Do nothing.

### Additional Context

Other such methods are available for ~primitive JS values e.g. `bool`, `object`, even `function`, etc

Would look like:

```rust
fn array_as_js_value() -> wasm_bindgen::JsValue {
    js_sys::Array::new().into()
}

fn test() {
    let val = array_as_js_value();
    assert!(val.is_array());
}
```

mfreeborn:
Closes #3086

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -3143,7 +3143,7 @@ impl<'a> Context<'a> {
 
             Intrinsic::IsArray => {
                 assert_eq!(args.len(), 1);
-                format!("{} instanceof Array", args[0])
+                format!("Array.isArray({})", args[0])
             }
 
             Intrinsic::IsUndefined => {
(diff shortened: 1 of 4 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -3141,6 +3141,11 @@ impl<'a> Context<'a> {
                 format!("typeof({}) === 'function'", args[0])
             }
 
+            Intrinsic::IsArray => {
+                assert_eq!(args.len(), 1);
+                format!("{} instanceof Array", args[0])
```
Liamolucko, line 3146: This should use [`Array.isArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) instead, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs_isarray.

Liamolucko (approved):

PR #2984, opened 2022-07-07
removes unnecessary drop()

mBornand:
The drop function is unnecessary as it will be droped anyway.
But if we leave it and the `struct` implements `Copy` clippy complains
about dropping the copy an not the original.

Without this, this code

```
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
#[derive(Copy, Clone)]
pub struct Test {
    pub a: u32
}
```
produces:
```
error: calls to `std::mem::drop` with a value that implements `Copy`. Dropping a copy leaves the original intact
 --> src/lib.rs:3:1
  |
3 | #[wasm_bindgen]
  | ^^^^^^^^^^^^^^^
  |
  = note: `#[deny(clippy::drop_copy)]` on by default
note: argument has type `Test`
 --> src/lib.rs:3:1
  |
3 | #[wasm_bindgen]
  | ^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_copy
  = note: this error originates in the attribute macro `wasm_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)
```
It is also a solution to the already closed #2858

Issue #2858, opened 2022-04-08
Clippy warns about `drop` on `struct` with `#[wasm_bindgen]` attribute that also derives `Copy`


bes:
### Describe the Bug

I want to embed a struct inside another struct and both structs have the `#[wasm_bindgen]` attribute:

If I just do it the naive way:

```rust
#[wasm_bindgen(js_name = Outer)]
pub struct Outer {
    #[wasm_bindgen]
    pub inner: Option<Inner>,
}

#[wasm_bindgen(js_name = Inner)]
pub struct Inner {
    #[wasm_bindgen]
    pub v: f64,
}
```

I get a compile error:

```shell
error[E0277]: the trait bound `inner::Inner: std::marker::Copy` is not satisfied
  --> src/outer.rs:11:16
   |
11 |     pub inner: Option<Inner>,
   |                ^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `inner::Inner`
   |
   = note: required because of the requirements on the impl of `std::marker::Copy` for `std::option::Option<inner::Inner>`
note: required by a bound in `outer::__wbg_get_outer_inner__const::__wbg_get_outer_inner::assert_copy`
  --> src/outer.rs:8:1
   |
8  | #[wasm_bindgen(js_name = Outer)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `outer::__wbg_get_outer_inner__const::__wbg_get_outer_inner::assert_copy`
   = note: this error originates in the attribute macro `wasm_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)
```

So I do what the compiler tells me, I make `Inner` derive `Clone` and `Copy`:

```rust
#[wasm_bindgen(js_name = Inner)]
#[derive(Clone, Copy)]
pub struct Inner {
    #[wasm_bindgen]
    pub v: f64,
}
```

Now the compiler is happy 🤗 but clippy is sad 😢:

```shell
error: calls to `std::mem::drop` with a value that implements `Copy`. Dropping a copy leaves the original intact
 --> src/inner.rs:7:1
  |
7 | #[wasm_bindgen(js_name = Inner)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[deny(clippy::drop_copy)]` on by default
note: argument has type inner::Inner
 --> src/inner.rs:7:1
  |
[...]

PR #3089, opened 2022-09-21
codegen: Ignore `clippy::drop_non_drop` in generated free fn

ijc:
Depending on the underlying types clippy can warn:

> call to `std::mem::drop` with a value that does not implement `Drop`. Dropping such a type only extends its contained lifetimes

However such a `drop` is ultimately harmless and in generated code it's easier to ignore it than to conditionally generate code with/without the drop.

This is an alternative to #2984 and is related to #2858.


Liamolucko:
Closing as #2984 was merged instead.

PR #3092, opened 2022-09-22
Call main only once

daxpedda:
As discussed in #3062, we use an atomic to only call the main function once.

For context, this change adjusts the `init` function to only call `main` or the function exported by `# [wasm_bindgen(start)]` only once even if `init` is called multiple times. This is important when using it in a context of spawning workers, otherwise each worker would call the `main` function again, which is probably unintended.

This might be considered a breaking change, because somebody out there might be relying on this behavior. But considering this is a nightly only feature (I believe), it should be fine?

The way I did this here is to re-use the thread counter. I had to somehow return it from `wasm_bindgen_threads_xform::Config::run()`, where it is created, so I created a `ThreadCounterAddr` that also has a method that is able to wrap the main function behind the atomic check.

I'm not really familiar with the code-base, but I did try my best to implement this as cleanly as possible, I'm open to suggestions of course.

Replaces #3062.

### DIFF
--- a/crates/threads-xform/src/lib.rs
+++ b/crates/threads-xform/src/lib.rs
@@ -1,6 +1,8 @@
 use anyhow::{anyhow, bail, Error};
 use std::cmp;
 use std::env;
+use walrus::ir::BinaryOp;
+use walrus::ir::LoadKind;
 use walrus::ir::Value;
 use walrus::{
     ir::MemArg, ExportItem, FunctionId, GlobalId, GlobalKind, InitExpr, InstrSeqBuilder, MemoryId,
@@ -22,6 +24,9 @@ pub struct Config {
     thread_stack_size: u32,
     enabled: bool,
 }
+
+#[derive(Clone, Copy)]
+pub struct ThreadCounterAddr(i32);
 
 impl Config {
     /// Create a new configuration with default settings.
@@ -103,9 +108,9 @@ impl Config {
     /// * Some stack space is prepared for each thread after the first one.
     ///
     /// More and/or less may happen here over time, stay tuned!
-    pub fn run(&self, module: &mut Module) -> Result<(), Error> {
+    pub fn run(&self, module: &mut Module) -> Result<Option<ThreadCounterAddr>, Error> {
         if !self.is_enabled(module) {
-            return Ok(());
+            return Ok(None);
         }
 
         let memory = wasm_conventions::get_memory(module)?;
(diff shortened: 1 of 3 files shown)


daxpedda:
@Liamolucko this is ready now.

crates/threads-xform/src/lib.rs
```diff
@@ -165,7 +170,25 @@ impl Config {
...
+        Ok(Some(ThreadCounterAddr(thread_counter_addr)))
+    }
+}
+
+impl ThreadCounterAddr {
+    pub fn wrap_start(self, memory: MemoryId, mut body: InstrSeqBuilder, id: FunctionId) {
+        // This happens after the thread counter was already increased, so "1"
+        // means we are in the first thread.
+        body.i32_const(self.0)
+            .load(memory, LoadKind::I32 { atomic: true }, ATOMIC_MEM_ARG)
```
Liamolucko, line 182: I think this could lead to a race condition - if a second thread starts between when the counter is incremented from 0 to 1 and when we load it here, this could be read as 2 instead of 1 despite this being the first thread, and then the start function won't get run at all.

This'll have to somehow get access to the result of the initial fetch-add done in `inject_start` to avoid the race condition.

Liamolucko (commented):
Sorry, I'm not sure why I took so long to get to this.

I think I've found a potential race condition, but if that can be fixed this looks good to me.

crates/threads-xform/src/lib.rs
daxpedda: So the reason why I thought this can not become a race condition is because this is called as part of the `init` function and you can't start a second thread because you don't have access to the WASM instantiation until the `init` function is finished.
I don't consider this brittle either because changing this behavior would be a really big change.

Still, I consider your suggestions an improvement, I will look into it now.

daxpedda: So I found two ways to potentially solve this:
1. The injected thread shim returns the thread counter, which we can take later and work with. The problem is that if we change the order of when shims are injected or if we insert a new shim, this will break, and I'm not sure how to fix it then except if make things more complicated and pass the thread counter around.
2. Save the thread counter into a thread local variable. Unfortunately I don't know how to make a thread local variable.

I consider the first solution worse then the current one, so my preference is to keep the current solution as I'm fairly certain it is correct.

daxpedda: I went through my assumptions again, I'm certain that this can't become a race condition.
Simply put, to make this a race condition the `init` function has to return the instantiated module without calling the start (`__wbindgen_start`) function.

Liamolucko: `init` isn't the only way to get a compiled module, though - you can use `WebAssembly.compile` to compile the module in advance, create a `SharedArrayBuffer`, and pass them to two threads which call `init` at the same time.

(I'm not sure what you mean by an 'instantiated module', but `init` takes a compiled `WebAssembly.Module` when initialising another thread.)

daxpedda: > `init` isn't the only way to get a compiled module, though - you can use `WebAssembly.compile` to compile the module in advance, create a `SharedArrayBuffer`, and pass them to two threads which call `init` at the same time.

I thought that this isn't supported? For example I took it that your comment here states that: https://github.com/rustwasm/wasm-bindgen/issues/3224#issuecomment-1372920396.

I guess if this is supported, then I have to implement the solution lined out above.
> (I'm not sure what you mean by an 'instantiated module', but `init` takes a compiled `WebAssembly.Module` when initialising another thread.)

Yes, that is what I mean. You can't get the `WebAssembly.Module` without calling the `init` function, unless, like you pointed out, you don't use the `init` function and compile it yourself.

daxpedda: See #3234, which addresses this issue.

daxpedda:
Replaced by #3234.

PR #3093, opened 2022-09-26
Add shell_test for wasm-audio-worklet example (build only)

lukaslihotzki:
Was part of #3069, but isn't helpful to actually test linked modules, because the shell tests are build-only.


Liamolucko (approved):
The shell tests aren't build only; they also load the example and check that no errors are logged to the console.

But either way, this looks good to me 👍

PR #3115, opened 2022-10-13
Allow non-literal constants in `typescript_custom_section`

### LINKED ISSUE #2828: typescript_custom_section doesn't accept the result of include_str macro
### Motivation

Instead of manually writing custom typings without syntax validation or highlighting like this
```rust
const GAME_ACTION: &'static str = r#"
export type GameAction =
  | { type: "ToggleRotation"; value: boolean }
  | { type: "ResizeWindow"; width: number; height: number };
"#;
```

### Proposed Solution

I'd prefer to extract the custom typings into a separate TS file and embed it using the following:
```rust
const GAME_ACTION: &'static str = include_str!("./custom-typings.ts");
```
But this is currently not supported and causes the following error message: `Expected a string literal to be used with #[wasm_bindgen(typescript_custom_section)].`

71:
Also:
- Bump version to 0.2.84, since the internal schema is changing.
- Bump MSRV for UI tests to 1.57 to allow `assert!` in `const fn`s.
- Update UI test outputs.

This will, among other things, allow `include_str!` in `#[wasm_bindgen(typescript_custom_section)]` (and fix #2828). This also opens the door for non-literal constants in `inline_js` (thus fixing #2787) with minimal changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -213,7 +213,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - run: rustup update --no-self-update 1.56.0 && rustup default 1.56.0
+    - run: rustup update --no-self-update 1.57.0 && rustup default 1.57.0
     - run: cargo test -p wasm-bindgen-macro
 
   build_examples:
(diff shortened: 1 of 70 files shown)


71:
Huh, _renaming_ a branch from GitHub considers that the branch is _deleted_ in open PRs. I made the same PR again in #3308.

PR #3117, opened 2022-10-17
Call __destroy_into_raw everywhere a value is consumed

bouk:
When using weak references the object needs to be unregistered from the finalization registry when we pass it to a Rust method that takes in the object by value. This wasn't happening, which meant that the finalizer was called despite the object already being freed. For me this manifested as a 'recursive use of an object detected which would lead to unsafe aliasing in rust' exception.

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -716,8 +716,7 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
             js.assert_class(&val, &class);
             js.assert_not_moved(&val);
             let i = js.tmp();
-            js.prelude(&format!("var ptr{} = {}.ptr;", i, val));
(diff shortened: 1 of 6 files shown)


alexcrichton:
Thanks! Would it be possible to add tests for this?

bouk:
@alexcrichton I've added a test which creates a bunch of garbage and then calls the GC, but unfortunately this exposed another problem: the classes test does a double borrow which leaves the `WasmRefCell` in a poisoned state, since the `RefMut` isn't dropped. I've added a `.free()` to the test to make it explicit, if you remove the free then the test I added fails instead.

Let me know what's the right way to proceed...

bouk:
I modified it to catch the `b.free()` exception so the tests pass at least

alexcrichton:
While not a great hack around this it seems reasonable enough to me for now, thanks!

Issue #2550, opened 2021-05-09
Expand the list of targets for pre-built binaries


OliverBrotchie:
### Motivation

A few tools such as WasmEdge (formerly SSVM) require pre-built wasm-bindgen-cli binaries to function properly.

### Proposed Solution

Adding build targets such as 'aarch64-apple-darwin' etc would alleviate this problem.

alexcrichton:
A PR for this would be most welcome!

PR #3129, opened 2022-11-01
Add aarch64-apple-darwin binary to releases

printfn:
I tested the aarch64 artifact at https://github.com/rustwasm/wasm-bindgen/actions/runs/3366801582 and it worked on my machine.

See #2550


alexcrichton:
Thanks!

PR #3136, opened 2022-11-06
unused that must be used

eli-front:
- Silence the following warning in the wasm audio worklet example

unused `Promise` that must be used
`#[warn(unused_must_use)]` on by default


alexcrichton:
Can this perhaps use `drop` or something else more idiomatic than an attribute?

daxpedda:
Thank you, I addressed the issue by [alexcrichton](https://github.com/alexcrichton) and merged it.

PR #3138, opened 2022-11-06
rust analyzer snake_case fix #2882

### LINKED ISSUE #2882: Generated constants should be UPPER_SNAKE_CASE
### Describe the Bug

Generated constants are not upper snake case and this triggers warnings from the `rust-analyzer`.

### Steps to Reproduce

Save this code to a `.rs` file:
```
#[wasm_bindgen]
extern "C" {
    pub type Something;
}
```

Import the required mods and run the `rust-analyzer`.

### Expected Behavior

No warnings should appear.

### Actual Behavior

The following error appears:
```
Diagnostic { code: DiagnosticCode("incorrect-ident-case"), message: "Constant `__wbg_generated_const_Something` should have UPPER_SNAKE_CASE name, e.g. `__WBG_GENERATED_CONST_SOMETHING`", range: 732..747, severity: WeakWarning, unused: false, experimental: false, fixes: Some([Assist { id: AssistId("change_case", QuickFix), label: "Rename to __WBG_GENERATED_CONST_SOMETHING", group: None, target: 732..747, source_change: None }]) }
```
image

### Additional Context

I've tried to disable the warning by using `#[allow(non_snake_case,non_upper_case_globals)]` but it does not do the trick. I've also tried to tweak the `rust-analyzer` to ignore this but I also failed.

eli-front:
Fixes #2882

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -1348,10 +1348,7 @@ impl<'a, T: ToTokens> ToTokens for Descriptor<'a, T> {
             return;
         }
 
-        let name = Ident::new(
-            &format!("__WBINDGEN_DESCRIBE_{}", ident.to_string().to_uppercase()),
-            ident.span(),
-        );
+        let name = Ident::new(&format!("__wbindgen_describe_{}", ident), ident.span());
         let const_name = Ident::new(
             &format!(
                 "__WBINDGEN_CONST_DESCRIBE_{}",
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -118,7 +118,7 @@ impl TryToTokens for ast::Program {
             const _: () = {
                 static _INCLUDED_FILES: &[&str] = &[#(#file_dependencies),*];
 
-                #[link_section = "__WASM_BINDGEN_UNSTABLE"]
+                #[link_section = "__wasm_bindgen_unstable"]
                 pub static _GENERATED: [u8; #generated_static_length] =
                     *#generated_static_value;
             };
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -138,7 +138,7 @@ impl ToTokens for ast::Struct {
         let new_fn = Ident::new(&shared::new_function(&name_str), Span::call_site());
         let free_fn = Ident::new(&shared::free_function(&name_str), Span::call_site());
         let free_fn_const = Ident::new(
-            &format!("{}__CONST", free_fn.to_string().to_uppercase()),
+            &format!("{}_CONST", free_fn.to_string().to_uppercase()),
             free_fn.span(),
         );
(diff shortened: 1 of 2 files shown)


Liamolucko (changes requested):

crates/backend/src/codegen.rs
```diff
@@ -1104,7 +1118,7 @@ impl TryToTokens for ast::ImportFunction {
             quote! {
                 #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
                 #(#attrs)*
-                #[link(wasm_import_module = "__wbindgen_placeholder__")]
+                #[link(wasm_import_module = "__WBINDGEN_PLACEHOLDER__")]
```
Liamolucko, line 1121: This also needs to remain lowercase.

crates/backend/src/codegen.rs
```diff
@@ -1334,9 +1348,15 @@ impl<'a, T: ToTokens> ToTokens for Descriptor<'a, T> {
             return;
         }
 
-        let name = Ident::new(&format!("__wbindgen_describe_{}", ident), ident.span());
+        let name = Ident::new(
+            &format!("__WBINDGEN_DESCRIBE_{}", ident.to_string().to_uppercase()),
+            ident.span(),
+        );
```
Liamolucko, line 1354: The CLI also relies on this being lowercase.

```suggestion
        let name = Ident::new(&format!("__wbindgen_describe_{}", ident), ident.span());
```

crates/test/src/rt/browser.rs
```diff
@@ -17,7 +17,7 @@ pub struct Browser {
 #[wasm_bindgen]
 extern "C" {
     type HTMLDocument;
-    static document: HTMLDocument;
+    static DOCUMENT: HTMLDocument;
```
Liamolucko, line 20: This still needs to be lowercase on the JS side:

```suggestion
    #[wasm_bindgen(js_name = document)]
    static DOCUMENT: HTMLDocument;
```

eli-front:
Thanks to @Liamolucko We have passed all CI tests!

FrankReh:
Any maintainers can review this?

Liamolucko:
Sorry, I completely let this slip. In the meantime, though, I found out that it's actually possible to not give consts names at all, by naming them `_`. I think that's a better solution, since it also makes it impossible for the consts to cause name collisions, so I opened #3198 which does that instead.

eli-front:
> Sorry, I completely let this slip. In the meantime, though, I found out that it's actually possible to not give consts names at all, by naming them `_`. I think that's a better solution, since it also makes it impossible for the consts to cause name collisions, so I opened #3198 which does that instead.

What was the original value for naming the constants?

Liamolucko:
> What was the original value for naming the constants?

I don't think there was any, I think it was just that nobody knew you didn't have to give them names.

PR #3142, opened 2022-11-12
Adding IDL for Animatable mixin

whizsid:
https://w3c.github.io/csswg-drafts/web-animations-1/#the-animatable-interface-mixin


whizsid:
At first I've added the webidl file to `unstable` folder as mentioned in the README file. But the previous CI failed because of that. Cuz Animatable IDL was linked to Element IDL here.
https://github.com/rustwasm/wasm-bindgen/blob/4f865308afbe8d2463968457711ad356bae63b71/crates/web-sys/webidls/enabled/Element.webidl#L252
So I added it to the `enabled` folder to make CI happy.

daxpedda:
This should be fine as an unstable API now.

PR #3150, opened 2022-11-16
Add WebXrHandInput Module

### LINKED ISSUE #3148: Add XrInputSource::hand
### Motivation

`XRInputSource` has a `hand` property which is not exposed by `web-sys`

### Proposed Solution

Add `XrInputSource::hand`

### Alternatives

Perhaps there is some context I am unaware of for why it isn't/shouldn't be added.

### Additional Context

[MDN Docs for the property](https://developer.mozilla.org/en-US/docs/Web/API/XRInputSource/hand)

rodrigocam:
As the Oculos Quest already support this API it would be nice to have this exposed.

resolves #3148


crates/web-sys/webidls/unstable/WebXRHandInputModule.webidl
```diff
@@ -0,0 +1,61 @@
+partial interface XRInputSource {
```
expenses, line 1: Maybe add a note like:

```suggestion
/* -*- Mode: IDL; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*- */
/*
 * WebXR Device API
 * W3C Working Draft, 19 April 2022
 * The origin of this IDL file is:
 * https://www.w3.org/TR/2022/WD-webxr-hand-input-1-20220419/
 */

partial interface XRInputSource {
```

This will allow someone to see if the webidl needs updating

Liamolucko (approved):
Thanks!

PR #3151, opened 2022-11-16
Undeprecate Safari WebIDL

nasso:
[Four years ago](https://github.com/rustwasm/wasm-bindgen/commit/bb7271aa4381120bc45cbe74500e9f72f7162da3), we manually deprecated `BaseAudioContext` and `AudioScheduledSourceNode` because Safari did not support them. The stable version of Safari at the time was 12.0 (September 2018). The current stable version of Safari is 16.1 (October 2022). Safari has introduced support for [`AudioScheduledSourceNode` in version 14.0](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode#browser_compatibility) (September 2020), and [`BaseAudioContext` in version 14.1](https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext#browser_compatibility) (April 2021).

### DIFF
--- a/crates/web-sys/src/features/gen_AudioScheduledSourceNode.rs
+++ b/crates/web-sys/src/features/gen_AudioScheduledSourceNode.rs
@@ -10,9 +10,7 @@ extern "C" {
     #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode)"]
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `AudioScheduledSourceNode`*"]
-    #[deprecated(note = "doesn't exist in Safari, use parent class methods instead")]
     pub type AudioScheduledSourceNode;
-    #[deprecated(note = "doesn't exist in Safari, use parent class methods instead")]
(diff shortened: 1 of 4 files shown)


nasso:
It looks like the generator got rid of most of the interface on `AudioContext` and `OfflineAudioContext` (since they can now be cast into `BaseAudioContext`, which now covers the interface).

Is this what we want? This is obviously pretty bad for backwards compatibility.

Should we still "mixin" the `BaseAudioContext` interface?

Liamolucko:
It's not as bad as it seems, because `AudioContext` implements `Deref<Target = BaseAudioContext>`, and so writing something like `ctx.destination()` would still work because the `AudioContext` gets auto-deref'd to `BaseAudioContext`. Cases like `AudioContext::destination(&ctx)` will break, though.

Liamolucko:
I reckon we should keep the mixin. Even though the potential breakage is pretty small, there's also not much downside to keeping the mixin and avoiding it.

We can still un-deprecate the superclasses though.

nasso:
Yeah, it would require new semver major release so I suspect we can just delay that to when we've build up enough changes?

I'll modify this PR to keep the mixin.

Liamolucko (approved):

PR #3161, opened 2022-11-21
Generate typescript type for imported JavaScript types 

bouk:
This replaces `: any` types when you have an imported JS type that you accept as a function argument e.g. this:

```
use wasm_bindgen::prelude::*;

#[wasm_bindgen(raw_module = "somepackage")]
extern "C" {
    pub type SomeType;
}

#[wasm_bindgen]
pub fn something(a: &SomeType) {
    drop(a);
}
```

Would generate `export function something(a: any): void;`, now it generates `export function something(a: import("somepackage").SomeType): void;` instead so it typechecks.


daxpedda:
@bouk happy to review this after a rebase!

crates/macro-support/src/parser.rs
```diff
@@ -687,17 +687,24 @@ impl<'a> ConvertToAst<(&ast::Program, BindgenAttrs, &'a Option<ast::ImportModule
...
-        (program, attrs): (&ast::Program, BindgenAttrs),
+        (program, attrs, module): (&ast::Program, BindgenAttrs, &'a Option<ast::ImportModule>),
     ) -> Result<Self::Target, Diagnostic> {
         let js_name = attrs
             .js_name()
             .map(|s| s.0)
             .map_or_else(|| self.ident.to_string(), |s| s.to_string());
+        let path_name = match module {
+            Some(ast::ImportModule::RawNamed(path, _)) => Some(path.clone()),
+            Some(ast::ImportModule::Named(path, _)) if !path.starts_with("/") => Some(path.clone()),
```
daxpedda, line 705:
```suggestion
            Some(ast::ImportModule::Named(path, _)) if !path.starts_with('/') => Some(path.clone()),
```
I'm not sure why checking for `starts_with('/')` isn't required elsewhere in the code.

daxpedda (commented):
Unfortunately, after looking through this, I don't think I can actually review/approve this.

I'm not sure how this will actually interact with how `module` works until now as I never really use it.
Also I'm not entirely sure what happens if the module is not a TypeScript module, not sure if that matters.

In any case, I will coordinate with the other maintainers and we will sort something out.

CI still needs to be fixed, in any case.

PR #3162, opened 2022-11-22
Increase interpreter buffer memory size

### LINKED ISSUE #2846: Long typescript_type leads to compile error in debug mode
### Describe the Bug

When compiling in debug mode, having a long typescript_type annotation for a custom type leads to a compiler error. Release mode works fine.

### Steps to Reproduce

Example code:

```rust
use wasm_bindgen::{prelude::*, JsCast};

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(typescript_type = "{                                                         }")]
    pub type Foo;
}

#[wasm_bindgen]
pub fn foo() -> Foo {
    JsValue::from_f64(0.0).unchecked_into()
}
```

Build with:
```
wasm-pack build --debug --target web
```

### Expected Behavior

```
Finished dev [unoptimized + debuginfo] target(s) in 0.34s
...
[INFO]: :-) Your wasm pkg is ready to publish at ...
```

### Actual Behavior

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.38s
[WARN]: :-) origin crate has no README
thread 'main' panicked at 'index out of bounds: the len is 256 but the index is 1073741823', [...]/wasm-bindgen-wasm-interpreter-0.2.79/src/lib.rs:307:17
```

### Additional Context

If you remove a single space from the typescript type definition, it works. There seems to be a maximum length allowed there, which seems to be around 58 characters (in my own project, the limit seems to be 54 characters for some reason). The content doesn't appear to matter.

Release builds work fine, this only fails in debug builds.

mvolfik:
Long typescript types don't fit into the buffer, causing stack pointer to underflow and wrap to numbers around `usize::MAX`

Fixes #2846

### DIFF
--- a/crates/wasm-interpreter/src/lib.rs
+++ b/crates/wasm-interpreter/src/lib.rs
@@ -72,7 +72,7 @@ impl Interpreter {
         // (the LLVM call stack, now the wasm stack). To handle that let's give
         // our selves a little bit of memory and set the stack pointer (global
         // 0) to the top.
-        ret.mem = vec![0; 0x100];
+        ret.mem = vec![0; 0x400];
         ret.sp = ret.mem.len() as i32;
 
         // Figure out where the `__wbindgen_describe` imported function is, if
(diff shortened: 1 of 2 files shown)


Liamolucko (approved):
This seems reasonable enough to me. Thanks!

PR #3169, opened 2022-11-26
Fix automatic .wasm file detection with no-modules output

### LINKED ISSUE #2502: Target no-modules failed to find its .wasm file automatically

daxpedda:
Currently with no-modules output the line `document.currentScript.src` will always return `null`, because it only works during processing of the script. This fixes the issue by caching `document.currentScript.src` beforehand.

Fixes #2502.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -378,6 +378,15 @@ impl<'a> Context<'a> {
             // function.
             OutputMode::NoModules { global } => {
                 js.push_str("const __exports = {};\n");
+                js.push_str("let script_src;\n");
+                js.push_str(
+                    "\
+                    if (typeof document === 'undefined') {
+                        script_src = location.href;
+                    } else {
+                        script_src = document.currentScript.src;
+                    }\n",
+                );
                 js.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, None)?;
                 footer.push_str(&format!(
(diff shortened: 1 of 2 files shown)


Liamolucko (approved):
Thanks!

Issue #3170, opened 2022-11-28
DataView / ArrayBuffer ergonomics are not well discoverable


chrysn:
### Motivation

Working with Web Bluetooth, web-sys APIs give me `js_sys::DataView` objects. Conceptually, these are JavaScript's version of (typed) slices. Accessing them as slices is hard, though.

### Proposed Solution

The documentation of DataView and Array should point users to the family of types around [js_sys](https://docs.rs/js-sys/latest/js_sys/index.html)::[Uint8Array](https://docs.rs/js-sys/latest/js_sys/struct.Uint8Array.html#), and how to turn one that was obtained into a such an array (my naïve `        let ble_response: js_sys::Uint8Array = characteristic.read_value().js2rs().await?.into();` did not do the trick).

### Alternatives

Up until in-depth documentation review with some guessing jumps, I turned my data views into vectors going through `my_vec.extend((0..ble_response.byte_length()).map(|i| ble_response.get_uint8(i)))` on them. This feels extremely unergonomic and inefficient.

PR #3171, opened 2022-11-28
Documentation extensions for ArrayBuffer and DataView

chrysn:
Closes: https://github.com/rustwasm/wasm-bindgen/issues/3170

---

* "the typed objects" previously could not be resolved by a user unfamiliar with the library. Giving one example should enable a human user to generalize; if there could be some shared category (module maybe) for the typed views, that could be linked instead.
* DataView had no pointer at all to these alternatives. It is worth adding one because a DataView is often the type browsers use (eg. the `read_characteristic` future resolves into one), so users will wind up here rather than at the buffer entry page (where they would take a conscious choice for build-time or call-time typing).
* The Uint8Array not being creatable from a `&DataView` but only from its `.buffer()` may be a bug or a feature, I really can't tell. The current text might read as if it's a bug, but if you could tell me how this is a feature, I'd be happy to rephrase.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -510,7 +510,7 @@ extern "C" {
     /// The `ArrayBuffer` object is used to represent a generic,
     /// fixed-length raw binary data buffer. You cannot directly
     /// manipulate the contents of an `ArrayBuffer`; instead, you
-    /// create one of the typed array objects or a `DataView` object
+    /// create one of the typed array objects ([Uint8Array] etc.) or a `DataView` object
     /// which represents the buffer in a specific format, and use that
     /// to read and write the contents of the buffer.
     ///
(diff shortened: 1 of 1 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -818,6 +818,12 @@ impl fmt::Debug for Boolean {
 // DataView
 #[wasm_bindgen]
 extern "C" {
+    /// An accessor to binary data stored inside an [ArrayBuffer]
+    ///
+    /// For statically typed views on the same buffer (which also offer tools for interacting with
+    /// Rust slices), see [Uint8Array], [Int64Array], [Float32Array] and so on. Beware that
+    /// `Uint8Array::new(&my_data_view)` *not* contain the full memory area; using the underlying
+    /// ArrayBuffer (`Uint8Array::new(&my_data_view.buffer())`) does.
```
Liamolucko, line 826: I'm not quite sure what this last sentence is supposed to be saying. `Uint8Array::new(&my_data_view)` doesn't work *at all*; it completely ignores the `DataView` and creates an empty `Uint8Array`. I'd phrase it something more along the lines of: 'A typed array cannot be created directly from a `DataView`; instead, it has to be created from its underlying `ArrayBuffer`: `Uint8Array::new(&my_data_view.buffer())`. Note that the resulting typed array will then be a view of the *entire* `ArrayBuffer`, even if the `DataView` was only a view of part of the `ArrayBuffer`.'

PR #3174, opened 2022-12-01
Improve error messages for invalidly invoked closures

Liamolucko:
The error message for an `Fn` closure being invoked after being dropped was confusing, because it was the same as for `FnMut`, 'closure invoked recursively or destroyed already'. The bit about being invoked recursively is misleading, since it's perfectly valid for an `Fn` closure to be invoked recursively. So, I removed that bit from the error message for `Fn` closures. I also rephrased 'destroyed already' to 'after being dropped', since I think that makes it a bit clearer that it's caused by the closure being dropped on the Rust side, rather than the more abstract wording of 'destroyed'.


alexcrichton (approved):
Seems reasonable to me! I think some tests may need an update though.

Issue #3176, opened 2022-12-02
Allow tearing down thread-allocated space from another thread


jrvidal:
### Motivation

:wave: Hello [again](https://github.com/rustwasm/wasm-bindgen/issues/2764). It turns out that, in our use case, we don't fully control the workers so we don't always have the chance to cleanly tear down thread-specific space with `__wbindgen_thread_destroy`.

### Proposed Solution

The bare minimum would be to expose `__stack_alloc` together with `__tls_base` and accept explicit arguments for those parameters in `__wbindgen_thread_destroy`. That way, a "leader" thread could perform the tear down (acting like a sort of scheduler).

### Alternatives

Do nothing. This goes deeper and deeper into accommodating thread-related use cases without guarantees of stability and no clear prospects on how Rust+Wasm+threads should/could look like.

jrvidal:
This time we went ahead and explored a potential solution: https://github.com/rustwasm/wasm-bindgen/pull/3177

PR #3177, opened 2022-12-02
allow thread destruction to run in other thread

jrvidal:
This is the bare minimum that it's needed to allow running thread destructors from another thread.

Fixes https://github.com/rustwasm/wasm-bindgen/issues/3176

### DIFF
--- a/crates/threads-xform/tests/basic.wat
+++ b/crates/threads-xform/tests/basic.wat
@@ -76,41 +76,56 @@
     global.set 1
     global.get 1
     call $__wasm_init_tls)
-  (func $__wbindgen_thread_destroy (type 0)
-    global.get 1
-    i32.const 128
-    call $__wbindgen_free
-    i32.const -2147483648
-    global.set 1
-    i32.const 393216
-    global.set 2
-    loop  ;; label = @1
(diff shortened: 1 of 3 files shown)


crates/threads-xform/tests/basic.wat
```diff
@@ -76,41 +76,56 @@
...
-    i32.const 393216
-    global.set 2
-    loop  ;; label = @1
-      i32.const 327684
-      i32.const 0
-      i32.const 1
-      i32.atomic.rmw.cmpxchg
-      if  ;; label = @2
+  (func $__wbindgen_thread_destroy (type 3) (param i32 i32)
+    (local i32 i32)
```
jrvidal, line 80: :warning: These are a bit annoying. I couldn't find a way in walrus to generate references to function parameters without creating spurious locals.

alexcrichton:
Sorry for the delay, but seems reasonable to me, thanks!

daxpedda:
@jrvidal I have questions about how this is supposed to be used.
- How do you get the `__stack_alloc` and `__tls_base` value for a specific thread?
- Can you call this function from inside Rust?

It should probably also be documented somewhere publicly like in https://github.com/rustwasm/wasm-bindgen/pull/2769.

PR #3182, opened 2022-12-07
[js-sys bindings] Intl.RelativeTimeFormat binding implementation

SoulSharer:
Bindings to [Intl.RelativeTimeFormat api](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) were missing, so I took the liberty of implementing them.


crates/js-sys/src/lib.rs
```diff
@@ -5528,6 +5528,65 @@ pub mod Intl {
...
+        pub type RelativeTimeFormat;
+
+        /// The `Intl.RelativeTimeFormat` object is a constructor for objects
+        /// that enable language-sensitive relative time formatting.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)
+        #[wasm_bindgen(constructor, js_namespace = Intl)]
+        pub fn new(locales: &Array, options: &Object) -> RelativeTimeFormat;
+
+        /// The `Intl.RelativeTimeFormat.prototype.format` property returns a getter function that
```
Liamolucko, line 5550:
```suggestion
        /// The `Intl.RelativeTimeFormat.prototype.format` property
```

The formatted string is returned directly; no getter function is involved.

SoulSharer: Good catch! Forgot to change this after removing getter. And its no longer a property ether. I'll change it.

SoulSharer: Corrected in [43d041e](https://github.com/rustwasm/wasm-bindgen/pull/3182/commits/43d041e7303f8602429d1c2aaa39495d8767502c).

Liamolucko:
I've got no idea why, but your change has somehow reduced the indentation of `rustc`'s error messages in one of the ui tests, which is why CI is failing. I've pushed a commit to update the expected output, which should fix it.

Liamolucko (approved):
Thanks!

SoulSharer:
> I've got no idea why, but your change has somehow reduced the indentation of `rustc`'s error messages in one of the ui tests, which is why CI is failing. I've pushed a commit to update the expected output, which should fix it.

Yeah, no idea, I thought something came from an upstream main, but it seems like an auto-merge issue. 🤔

PR #3193, opened 2022-12-10
Add `spawn_boxed_local`

Aandreba:
Adds the function `spawn_boxed_local` as another way to spawn futures.
It's body is the following:

```rust
#[inline]
pub fn spawn_boxed_local (future: Pin<Box<dyn 'static + Future<Output = ()>>>) {
    task::Task::spawn(future);
}
```

This method is useful to spawn already-boxed futures without requiring an extra allocation.

### DIFF
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -88,7 +88,7 @@ where
 ///
 /// The `future` will always be run on the next microtask tick even if it
 /// immediately returns `Poll::Ready`.
-/// 
+///
 /// This is a convenience function to spawn a `future` that is already boxed,
 /// and avoid doing an extra allocation to spawn it.
 ///
@@ -96,7 +96,7 @@ where
 ///
 /// This function has the same panic behavior as `future_to_promise`.
 #[inline]
-pub fn spawn_boxed_local (future: Pin<Box<dyn 'static + Future<Output = ()>>>) {
+pub fn spawn_boxed_local(future: Pin<Box<dyn 'static + Future<Output = ()>>>) {
     task::Task::spawn(future);
 }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Do async runtimes like tokio have precedent in the way of naming or existence for APIs like this?

Aandreba:
> Do async runtimes like tokio have precedent in the way of naming or existence for APIs like this?

I've just checked tokio and async_std, and they don't seem to offer a similar API.

alexcrichton:
Ah ok in that case I'd be a little hesitant to forge ahead with this ahead of those runtimes. Is the performance an issue for your application or was this more of a drive-by "nice to have"?

Aandreba:
> Is the performance an issue for your application or was this more of a drive-by "nice to have"?

It's not a performance issue right now, but I don't see any downside to adding it. The fact that it isn't implemented in tokio and async_std looks to me more like nobody bothered more than anything.

Aandreba:
Is the fact that there isn't an equivalent API on other runtimes a big problem? I don't see how one affects the other. Am I missing something?

Aandreba:
> If this isn't otherwise motivated already in larger communities with tighter performance requirements I'm not sure why it would otherwise be specifically motivated for here

Spawn on tokio has the following implementation.

```rust
pub fn spawn<T>(future: T) -> JoinHandle<T::Output>
    where
        T: Future + Send + 'static,
        T::Output: Send + 'static,
    {
        // preventing stack overflows on debug mode, by quickly sending the
        // task to the heap.
        if cfg!(debug_assertions) && std::mem::size_of::<T>() > 2048 {
            spawn_inner(Box::pin(future), None)
        } else {
            spawn_inner(future, None)
        }
    }
```

It appears that (when not on thebug and the future is not too big) tokio does some kind of optimization with the futures before boxing them. Since this is not the case here, I don't see why we shouldn't allow it.

PR #3197, opened 2022-12-13
fix: Intl.Numberformat.format takes a number and returns a string

MaartenStaa:


Liamolucko:
This is a duplicate of #2506, which unfortunately can't be merged right now because it's a breaking change.

MaartenStaa:
Ah thanks, my bad. I searched for `NumberFormat`, but didn't see the other PR. I'll keep an eye on that one.

Issue #3204, opened 2022-12-18
Tests are broken


uberFoo:
I was going to start hacking on enum support this morning. Before I broke stuff I figured I should run the tests: `warm-pack test --node`, and they failed. I bisected the repo and found:
```
╭─uberfoo@moon.local ~/projects/github/wasm-bindgen ‹3.1.3› ‹ebe65873› 
╰─➤  git bisect good      
df1b591a1cb88f0255844d26b13f5364fdb0950e is the first bad commit
commit df1b591a1cb88f0255844d26b13f5364fdb0950e
Author: João Freires <rmjoaovictor@gmail.com>
Date:   Wed Jun 22 17:44:52 2022 -0300
```
It was introduced in #2954. I haven't really looked into it much, but I have a bad feeling that fixing this is going to involve a lot of changes to the tests themselves. I'm going to look into it more.

TBH, I'm a bit surprised that nobody has even noticed.

Liamolucko:
That's not how the tests are meant to be run. See https://rustwasm.github.io/wasm-bindgen/contributing/testing.html; the main set of tests are run with `cargo test --target wasm32-unknown-unknown`.

As for why they're failing like that, it's because `wasm-pack` is attempting to use the latest published version of the CLI with the current git version of the macro, which is incompatible. Using `cargo test` builds the current git version of the CLI from source instead.

uberFoo:
It seemed likely I was doing something wrong. I found an old README in the repo that led me to believe that the two ways of running tests were similar. I went the wrong way because of some other problem running the tests with cargo. I'll have to go back and look into that. 

Thanks!

PR #3205, opened 2022-12-19
Validate the schema version before processing descriptors

Liamolucko:
While looking into #3204 I noticed that the error given by `wasm-pack test` was a panic while decoding descriptors, rather than the 'version mismatch' error that using an incompatible CLI version should give. It turns out that that was happening because descriptors were getting parsed before the custom section, and hence before the version check.

So, I reordered things so that the custom section is parsed (but not processed) before the descriptors, which includes checking the schema version. The custom section still gets processed afterwards, though, since that seems to depend on the results of processing descriptors.


alexcrichton:
👍

PR #3210, opened 2022-12-20
Docfix: Remove dangling word

expede:
# Problem

I noticed the [top-level of the generated docs](https://docs.rs/wasm-bindgen/0.2.83/wasm_bindgen/) has a very _mysterious_ dangling word:

Screenshot 2022-12-19 at 23 47 32

(Imported from... _from where?!_)

# Solution

I figured that it made more sense in the source, and it does! It just reads oddly on the front page of the docs. I just removed the one word, which is probably sufficient for both the HTML docs and those reading the source. Feel free to rephrase differently!

## Updated Preview

Screenshot 2022-12-19 at 23 54 02

It's small, but I hope it's helpful! This is an incredible project; thanks so much for the hard work!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,7 +44,7 @@ macro_rules! externs {
     )
 }
 
-/// A module which is typically glob imported from:
+/// A module which is typically glob imported
 ///
 /// ```
 /// use wasm_bindgen::prelude::*;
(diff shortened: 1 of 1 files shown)


Liamolucko (approved):
Thanks!

PR #3214, opened 2022-12-24
webidl: track stability of individual interface & namespace members

### LINKED ISSUE #2560: Keep track of stability for InterfaceMethod
### Describe the Bug

While trying to add the File System Access API WEBIDL in unstable, I noticed that while `AttributeInterfaceData` contains a `stability` member, `OperationData` does not. This means that new methods added by this unstable WEBIDL on global interfaces, like `Window::show_directory_picker`, aren't tagged with `#[cfg(web_sys_unstable_apis)]`.

### Steps to Reproduce

1. Clone https://github.com/pxlr-app/wasm-bindgen/tree/filesystemaccess
2. Generate the bindings `cargo run --release --package wasm-bindgen-webidl -- webidls src/features`

### Expected Behavior

The unstable WEBIDL adds new methods on `Window` that aren't tagged with `#[cfg(web_sys_unstable_apis)]` since the stability of methods inherits theirs from the interface.

### Actual Behavior

Function like `Window::show_directory_picker` is exposed to non-unstable feature.

Liamolucko:
Fixes #2560

It's possible for some members of an interface (or namespace) to be stable and others not, if a `partial interface` is defined in an unstable webidl file which adds to an interface defined in a stable file. This wasn't properly kept track of for all interface members before, most notably for operations. This PR fixes that.

This did unfortunately root out some WebXR-related bits of WebGL that were marked stable when they shouldn't have been. So, currently, this is a slight breaking change. I'm not sure if I should special-case them to be left as stable or accept the slight risk of breakage, since it seems unlikely that people would be using these isolated bits of WebXR without enabling `--cfg=web_sys_unstable_apis` for the rest of WebXR.

### DIFF
--- a/crates/web-sys/src/features/gen_WebGl2RenderingContext.rs
+++ b/crates/web-sys/src/features/gen_WebGl2RenderingContext.rs
@@ -5980,12 +5980,16 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `WebGl2RenderingContext`, `WebGlProgram`*"]
     pub fn link_program(this: &WebGl2RenderingContext, program: &WebGlProgram);
+    #[cfg(web_sys_unstable_apis)]
     # [wasm_bindgen (method , structural , js_class = "WebGL2RenderingContext" , js_name = makeXRCompatible)]
     #[doc = "The `makeXRCompatible()` method."]
     #[doc = ""]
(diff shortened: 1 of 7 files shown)


alexcrichton:
Thanks for this! I agree it's probably best to land this and deal with breakage as a follow-up, if any.

PR #3220, opened 2022-12-29
wasm-bindgen-webidl: only invoke `rustfmt` once

Liamolucko:
This significantly reduces how long it takes to run, from ~42s down to ~3s on my computer.


alexcrichton:
Good catch!

PR #3221, opened 2022-12-29
Add `JsCast` to `wasm_bindgen::prelude`

Liamolucko:
`JsCast` is a very commonly used trait, but for some reason has never been added to `wasm_bindgen::prelude`, leading to the constant annoyance of having to manually import it.


alexcrichton:
👍

PR #3225, opened 2023-01-06
Add VideoFrame and OffscreenCanvas to CanvasImageSource.

jneem:
According to [this](https://html.spec.whatwg.org/#2dcontext), `VideoFrame` and `OffscreenCanvas` are variants of `CanvasImageSource`, so I've added them. Since `CanvasImageSource` is also one of the variants of `ImageBitmapSource` [here](https://html.spec.whatwg.org/#images-2), I added them there too.


Liamolucko (approved):
Thanks!

PR #3231, opened 2023-01-12
BREAKING(web-sys): Remove support for `CanvasRenderingContext2D` and `BufferSource` from `createImageBitmap`

### LINKED ISSUE #3227: Window::create_image_bitmap_with_canvas_rendering_context_2d throws error
### Describe the Bug

I wanted to conveniently create an `ImageBitmap` from `CanvasRenderingContext2d`. When I found `Window::create_image_bitmap_with_canvas_rendering_context_2d`, I expected that it's a convenience method for doing so.

However, when I used it, I received the following error:

```text
TypeError: Failed to execute 'createImageBitmap' on 'Window': The provided value is not of type '(Blob or HTMLCanvasElement or HTMLImageElement or HTMLVideoElement or ImageBitmap or ImageData or OffscreenCanvas or SVGImageElement or VideoFrame)'.
```

### Steps to Reproduce

```rust
async fn update_buf_dims_gracefully(
    canvas: &HtmlCanvasElement,
    ctx: &CanvasRenderingContext2d,
    window: &Window,
) {
    let client_width= canvas.client_width();
    let client_height = canvas.client_height();

    let image_bitmap: JsFuture = window.
        create_image_bitmap_with_canvas_rendering_context_2d(ctx)
        .unwrap()
        .into();

    let image_bitmap = image_bitmap.await;
    match image_bitmap {
        Ok(image_bitmap) => { log(&image_bitmap) },
        Err(err) => { log(&err) }
    }

    canvas.set_width(client_width as u32);
    canvas.set_height(client_height as u32);
}
```

The code above is an incomplete implementation of a function relying on `Window::create_image_bitmap_with_canvas_rendering_context_2d`.

### Expected Behavior

`Window::create_image_bitmap_with_canvas_rendering_context_2d` should either be removed or be functional

### Actual Behavior

Results in an error.

### Additional Context

No additional context.

Liamolucko:
Resolves #3227

Both APIs are broken, to some extent, which is why I think it makes sense to remove them despite the potential breakage.

Calling `createImageBitmap` with a `BufferSource` will instantly throw an error, and calling it with a `CanvasRenderingContext2D` will also fail except in Firefox, where it gives a deprecation warning.

`BufferSource` was there as part of the abandoned video capture worker API (https://w3c.github.io/mediacapture-worker). It looks like we inherited it from Firefox, which supported it behind a flag at the time we copied our WebIDL from it but later removed it. I also removed the rest of that API while I was at it.

`CanvasRenderingContext2D` is there because, for whatever reason, Firefox non-standardly accepts it in `createImageBitmap`, and we took our WebIDL from them. It's deprecated, and https://bugzilla.mozilla.org/show_bug.cgi?id=1500782 talks about removing it, but it hasn't been done yet.


alexcrichton:
Sounds reasonable to me, thanks!

PR #3236, opened 2023-01-13
Call main only once take 4

### LINKED ISSUE #2295: How to avoid extra calls to main when creating web worker from binary?

daxpedda:
As discussed in #3092, this addresses the possible race condition if one doesn't use `init` when compiling the WASM module.

To make this possible I changed the way the start function was modified in various places.
Currently the start function was always taken from the module and then wrapped into a new function, constructing a pretty nested start function.

With this PR the start function is constantly being modified instead of wrapping it over and over again, this allows to use local variables, eg. I stored the current thread counter in a local variable that I can use later to determine if we are in the first thread or not.

I hope it's alright that I put some shared functionality into `wasm-bindgen-wasm-conventions`.

Replaces #3062, #3092 and #3234.
Fixes #2295.

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -338,7 +338,8 @@ impl Bindgen {
                 .context("failed getting Wasm module")?,
         };
 
-        self.threads
+        let thread_count = self
+            .threads
             .run(&mut module)
             .with_context(|| "failed to prepare module for threading")?;
 
@@ -383,6 +384,7 @@ impl Bindgen {
             programs,
             self.externref,
             self.wasm_interface_types,
+            thread_count,
             self.emit_start,
         )?;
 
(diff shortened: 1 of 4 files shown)


daxpedda:
This is ready now @Liamolucko.

Liamolucko (approved):
Looks good to me!

PR #3242, opened 2023-01-17
add bundler target to wasm-bindgen-test-runner

robertkiel:
tbd

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/bundler.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/bundler.rs
@@ -1,0 +1,132 @@
+use std::env;
+use std::ffi::OsString;
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+
+use anyhow::{Context, Error};
+
+// depends on the variable 'wasm' and initializes te WasmBindgenTestContext cx
+pub const SHARED_SETUP: &str = r#"
+const handlers = {};
+
+const wrap = method => {
+    const og = console[method];
+    const on_method = `on_console_${method}`;
+    console[method] = function (...args) {
+        og.apply(this, args);
+        if (handlers[on_method]) {
+            handlers[on_method](args);
+        }
+    };
+};
+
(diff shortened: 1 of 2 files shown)


Liamolucko:
This won't work. The bundler target relies on the [WebAssembly esm-integration proposal](https://github.com/WebAssembly/esm-integration) to import WebAssembly files, which Node (and all other runtimes that I'm aware of) doesn't support yet, unless you pass the `--experimental-wasm-modules` flag.

robertkiel:
> This won't work. The bundler target relies on the [WebAssembly esm-integration proposal](https://github.com/WebAssembly/esm-integration) to import WebAssembly files, which Node (and all other runtimes that I'm aware of) doesn't support yet, unless you pass the `--experimental-wasm-modules` flag.

So why don't run `node` with the flag? Within the client that we're creating this works quite well.

Liamolucko:
> > So why don't run `node` with the flag?
> 
> I guess we don't really want to support all the experimental stuff out there. I know little about Node.js, WDYT @Liamolucko.

I mean, we could, but I think it’d be better to avoid relying on experimental features, particularly when it’s being run inside wasm-bindgen-test-runner as opposed to being explicitly specified by the user.

@robertkiel, is there a particular reason why this needs to use esm-integration to load the wasm file rather than instantiating it the normal way? If not, this could avoid the experimental flag by instead using the web target, which is still ESM, and manually loading the wasm from the file system and passing it to `init` to sidestep the web target's usual fetch-based loading. Obviously the web target isn’t designed to be used with Node, but neither is the bundler target.

There’s also the issue of telling Node to interpret the shim as an ES module. This PR already names the `wasm-bindgen-test` runtime script `run.mjs` to make that get interpreted as an ES module, but it doesn't do the same for the shim, which means that Node interprets it as CommonJS if `”type”: “module”` isn’t set. It shouldn’t be too hard to fix that, by renaming the files from `.js` to `.mjs` after `wasm-bindgen-cli-support` writes them.

PR #3243, opened 2023-01-17
`#[wasm_bindgen(start)]` shouldn't require `pub`

daxpedda:
I noticed that every time I used `#[wasm_bindgen(start)]` it required me to add the `pub` visibility modifier, which I thought could be improved.


hamza1311 (approved):
Makes sense for `start`, considering that `fn main` is also not pub

PR #3245, opened 2023-01-19
fix: convert `None` to `null` instead of `undefined`

amrbashir:
Converting `None` to `undefined` in JS land, means that `Enum::Variant(None)` will be converted to `{ "Variant": undefined }` . This is mostly fine in JS but when trying to serialize the resulting object using `JSON.stringify`, it will skip `undefined` properties and I think that is counter-intutive so `None` should be mapped to `null` instead.

ref: tauri-apps/tauri#5993

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -983,10 +983,7 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
         Instruction::OptionRustFromI32 { class } => {
             js.cx.require_class_wrap(class);
             let val = js.pop();
-            js.push(format!(
-                "{0} === 0 ? undefined : {1}.__wrap({0})",
-                val, class,
-            ))
+            js.push(format!("{0} === 0 ? null : {1}.__wrap({0})", val, class,))
         }
 
         Instruction::CachedStringLoad {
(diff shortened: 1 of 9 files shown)


amrbashir:
hmmm seems like my change did nothing, not sure where I should be looking now tbh.

Liamolucko:
This is the code you'd need to change to do this:

https://github.com/rustwasm/wasm-bindgen/blob/cef9f4ac80800788f0738700432f705cfbace9ba/crates/cli-support/src/js/binding.rs#L1167-L1170

But, although I agree that mapping `None` to `null` makes more sense, I think it'll break people's code that relies on the current behaviour of it getting mapped to `undefined`. So, I'm not going to merge this, but I'll leave it open so that it can get merged if/when `wasm-bindgen` makes a breaking release.

amrbashir:
Thanks for the guidance, I went through the file and update some `Option` to JS mappings to `null` where it made sense and updated the tests accordingly, however there is still one test case that always fails
https://github.com/rustwasm/wasm-bindgen/blob/762582342f24078b309e620f5844fddca77e950b/tests/wasm/option.js#L10 

> So, I'm not going to merge this, but I'll leave it open so that it can get merged if/when `wasm-bindgen`makes a breaking release.

Understandable.

amrbashir:
Seems like there is some JS functions like `String.prototype.toLocaleUpperCase` that take optional arguments and, in this case, `None` mut be mapped to `undefined`

PR #3252, opened 2023-01-20
Implement `#[wasm_bindgen(start_async)]`

### LINKED ISSUE #3076: `async main()` support
### Motivation

I think it would be great to support something like this:
```rust
#[wasm_bindgen(start)]
async fn main() { }
```

It would add a bit of quality-of-life to prevent users from having to use weird workarounds. It would also help a bit with cross-platform codebases.

### Proposed Solution

The solution is probably the same as the ones used by [`tokio::main`](https://docs.rs/tokio/1.21.0/tokio/attr.main.html) and [`async_std::main`](https://docs.rs/async-std/1.12.0/async_std/attr.main.html), they usually remove the `async` keyword and insert their code inside the function. I believe the only current limitation of `#[wasm_bindgen(start)]` is that it doesn't remove the `async` keyword.

### Alternatives

- Stick with what we have now, which works quiet fine.
- Introduce a separate attribute? E.g.: `#[wasm_bindgen(start_async)]`.

### Additional Context

Original issue: https://github.com/rustwasm/wasm-bindgen/issues/1904
PR: https://github.com/rustwasm/wasm-bindgen/pull/1905

I'm happy to do a PR, but wanted to ask if this feature is desired first.
Also some pointers would be great, I briefly looked into it and the architecture around proc-macro handling in `wasm-bindgen` is quite big.

daxpedda:
Adds a new attribute called `start_async`, it has the exact same behavior as `start` but rewrites the function to a non-async version that contains the `spawn_local` inside it.

The purpose is to allow `async fn main()` to be used in Cargo binaries. It's similar to what `#[tokio::main]` and others do.

This could have also been done by adjusting `start`, but the issue is that it might break some behavior, for example one could rely on the function actually being `async`, which it wouldn't be after the rewrite. This could be mitigated by adjusting the behavior only for functions called `main`, but it's still a breaking change I would argue.

Fixes #3076.

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -10,7 +10,7 @@ use proc_macro2::{Delimiter, Ident, Span, TokenStream, TokenTree};
 use quote::ToTokens;
 use syn::parse::{Parse, ParseStream, Result as SynResult};
 use syn::spanned::Spanned;
-use syn::Lit;
+use syn::{Lit, ReturnType};
 
 thread_local!(static ATTRS: AttributeParseState = Default::default());
 
@@ -81,6 +81,7 @@ macro_rules! attrgen {
             (typescript_custom_section, TypescriptCustomSection(Span)),
             (skip_typescript, SkipTypescript(Span)),
             (start, Start(Span)),
+            (start_async, StartAsync(Span)),
             (skip, Skip(Span)),
             (typescript_type, TypeScriptType(Span, String, Span)),
             (getter_with_clone, GetterWithClone(Span)),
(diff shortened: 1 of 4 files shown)


daxpedda:
Alternatively we could introduce a new attribute called `main`, which differs from start in the sense that it is supposed to be used for the `fn main()` function.
No changes in functionality to `start_async`, but more descriptive.

daxpedda:
After some more experimentation I think there should definitely be a new attribute `main` that is different from `start`, supposed to only be used with `fn main()`.

The code generation could be much simpler and there are some limitations around exporting main multiple times that could be alleviated this way.

Issue #2716, opened 2021-11-04
U64 does not work in Safari


Firaenix:
### Describe the Bug

Safari 14 does not support BigUint64Array as mentioned in https://github.com/rustwasm/wasm-bindgen/pull/2441#issuecomment-806737455

### Steps to Reproduce
Run any wasm code/test that exposes u64 in safari.

https://github.com/Firaenix/wasm-bindgen-BigUint64Array-repro

### Expected Behavior
Should run in Safari (iOS and macOS) as in any other environment.

### Actual Behavior

Console error that causes the code not to run.

memoryruins:
Safari 15, released in September, [added support](https://developer.apple.com/documentation/safari-release-notes/safari-15-release-notes) for `BigInt64Array` and `BigUint64Array` on macOS and iOS. It might take a few more months before a majority of users have upgraded.

Firaenix:
I don't think that's a satisfying solution, especially when the documentation claims that it should work on Safari.
If we are already providing what is essentially a polyfill for u64, it shouldnt just be good enough to say that we should just wait for users to upgrade.

I'm working on a PR that should hopefully solve this issue for all BigInt supporting browsers

memoryruins:
To be clear, I wasn't offering a solution previously, only describing the current state of Safari support. An improved polyfill that handles Safari 14 would be great.

Firaenix:
All good, I misinterpreted in that case

Firaenix:
I have a branch on my fork which works in Safari and all other browsers that support BigInt:
https://github.com/bsvwasm/wasm-bindgen/tree/universal-ui64-patch

Firaenix:
https://github.com/rustwasm/wasm-bindgen/pull/2727

dovahcrow:
does #3037 fixed this issue?

Liamolucko:
> does #3037 fixed this issue?

Yes.

senolem:
Hello, https://github.com/rustwasm/wasm-bindgen/commit/bf109de9b2a46de0d8e7ffea198823e5a6744a1b was supposed to fix this issue, but release 0.2.83 still fails upon using BigInt64Array on Safari iOS 14.8. This has been tested through Ruffle and the issue started when they updated to 0.2.83 (https://github.com/ruffle-rs/ruffle/commit/1954f6162f6b53546f216efc0de3d029ea54cc5e). Any ideas?

Liamolucko:
[...]

PR #2886, opened 2022-05-04
Optimise memory view getters

Liamolucko:
While working on cloudflare/serde-wasm-bindgen#34, I noticed that a significant amount of time was spent getting memory views. Most of that time was spent in the getter for `WebAssembly.Memory.buffer`, so I've changed the memory view getters to check for invalidation without comparing to the wasm memory, by checking if the length is 0 (which indicates the buffer being detached).

`wasm-bindgen` benchmarks:

| Benchmark                                             | Before      | After       | Change  |
| ----------------------------------------------------- | ------------| ----------- | ------- |
| Access Node.nodeType with final                       | 364/s       | 383/s       | +5.22%  |
| Access Node.nodeType with structural                  | 1,166/s     | 1,189/s     | +1.97%  |
| Call a custom JS class Foo.bar method with final      | 6,248/s     | 6,189/s     | -0.94%  |
| Call a custom JS class Foo.bar method with structural | 6,642/s     | 6,667/s     | +0.38%  |
| Pass ascii_small to/from wasm-bindgen                 | 1,281,297/s | 1,823,742/s | +43.04% |
| Pass ascii_medium to/from wasm-bindgen                | 1,231,065/s | 1,723,498/s | +40.00% |
| Pass ascii_number to/from wasm-bindgen                | 1,161,427/s | 1,591,248/s | +37.01% |
| Pass ascii_date to/from wasm-bindgen                  | 918,950/s   | 1,138,144/s | +23.85% |
| Pass ascii_url to/from wasm-bindgen                   | 899,162/s   | 1,111,610/s | +23.63% |
| Pass ascii_link to/from wasm-bindgen                  | 864,511/s   | 1,105,580/s | +27.89% |
| Pass unicode to/from wasm-bindgen                     | 351,071/s   | 390,481/s   | +11.23% |

`serde-wasm-bindgen` benchmarks (not including cloudflare/serde-wasm-bindgen#34):

| Benchmark              | Before        | After         | Change  |
| ---------------------- | ------------- | ------------- | ------- |
| Parse canada           | 44.01 ms/iter | 34.52 ms/iter | -21.56% |
| Parse citm_catalog     | 10.32 ms/iter | 8.78 ms/iter  | -14.92% |
| Parse twitter          | 8.99 ms/iter  | 8.23 ms/iter  | -8.45%  |
| Serialize canada       | 11.74 ms/iter | 11.59 ms/iter | -1.28%  |
[...]

PR #3253, opened 2023-01-21
Lazily initialize memory views

Liamolucko:
This stops load-time errors in versions of Safari before 15 if the `BigInt64Array` or `BigUint64Array` memory views are included but never used.

Pointed out in https://github.com/rustwasm/wasm-bindgen/issues/2716#issuecomment-1397621198.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -787,10 +787,10 @@ impl<'a> Context<'a> {
             for kind in views {
                 writeln!(
                     init_memviews,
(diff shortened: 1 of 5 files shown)


alexcrichton:
Could this perhaps just be done for u64? I think that this was how it once was long ago and it was switched to the current pattern as it was slightly more performant.

Liamolucko:
I think you're thinking of #2886, which did include that change, but it wasn't where the performance wins came from; it was just another thing I happened to change.

alexcrichton:
Ah sorry, indeed!

PR #3263, opened 2023-01-24
Support `Result<(), impl Debug>` in `#[wasm_bindgen(start)]`

### LINKED ISSUE #3262: Support more types in `Result<(), E>` for `#[wasm_bindgen(start)]`
### Motivation

Reducing code duplication in cross-platform code-bases by getting as close as possible to what `fn main()` supports.
I would argue that #2994 and #2970 are also cross-platform issues, because error libraries are usually not targeted towards `wasm-bindgen` but all sorts of targets.

See #3076 and #3252 for more `fn main()` compatibility.

### Proposed Solution

Make `#[wasm_bindgen(start)]` as compatible as possible with [`Termination`](https://doc.rust-lang.org/std/process/trait.Termination.html), which is what is expected in `fn main()` by Rust.

Currently the return error type has to be a `JsValue` or `()`, but very commonly error types like `anyhow::Error` are used.
We should still continue supporting `JsValue` at the same time.

### Alternatives

I don't know of any alternative for `wasm-bindgen` to apply here. This can always be solved on the user side with more `#[cfg(...)]`s.

daxpedda:
This adds support to `#[wasm_bindgen(start)]` to support more types closer to what Rusts `fn main()` supports while not changing the current behavior of throwing `JsValue`s when available.

Looking at [`Termination`](https://doc.rust-lang.org/std/process/trait.Termination.html), which is what `fn main()` uses, I added support for [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) and `Result<(), impl Debug>`.
I intentionally left `ExitCode` unaddressed because I don't know exactly what to throw in that case.

I contemplated introducing a new trait, like `JsTermination` to mirror exactly what Rust does, but that's not really compatible with the current way `#[wasm_bindgen(start)]` works. We could however introduce it with a new `#[wasm_bindgen(main)]` if that is desired? See #3252 for a similar proposal.

The way this problem was solved is by using Autoref-based specialization, see [this case study by dtolnay](https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md), which is sometimes used to solve this kind of problem in macros.

Fixes #3262.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -489,7 +489,10 @@ impl TryToTokens for ast::Export {
                     quote! { () },
                     quote! { () },
                     quote! {
-                        <#syn_ret as wasm_bindgen::__rt::Start>::start(#ret.await)
+                        {
+                            use wasm_bindgen::__rt::Start;
+                            wasm_bindgen::__rt::StartWrapper(#ret.await).__wasm_bindgen_start()
+                        }
                     },
                 )
             } else {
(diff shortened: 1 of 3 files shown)


daxpedda:
See https://github.com/rustwasm/wasm-bindgen/pull/3252#issuecomment-1426911097.

I will introduce a new attribute that is supposed to only be used with `fn main()`.

PR #3268, opened 2023-01-29
Add support for setlike interfaces

### LINKED ISSUE #3060: `setlike` implementation
### Motivation

[`GpuSupportedFeatures`](https://docs.rs/web-sys/0.3.59/web_sys/struct.GpuSupportedFeatures.html) doesn't currently have a `has` method despite being `setlike`. This is needed for https://github.com/gfx-rs/wgpu/pull/2986. IDL:

```webidl
[Exposed=(Window, DedicatedWorker), SecureContext]
interface GPUSupportedFeatures {
    readonly setlike<DOMString>;
};
```

### Proposed Solution

Add a `has` method to interfaces with `setlike`.

gents83:
Generate missing operations on setlike interfaces such as **GPUSupportedFeatures**

Fixes #3060 

Taking as a reference maplike, except that it's not a map but just a set, we have an add() instead of set() and we do not have a get()

PS: It's my first request in _wasm-bindgen_ so better to double-check it :)


Liamolucko (approved):
Thanks!

PR #3269, opened 2023-01-30
Implement `ImageBitmapOptions`

daxpedda:
See the [WHATWG Spec](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmapoptions).

Adding an optional at the end was a problem because the current IDL generator will change the function name because of an overlapping second argument between functions. This cannot be fixed because it would affect all other functions.
So I opted to just implement a new table that does post-fixes, I assume this will be useful for other additions in the future.

WebIDL references:
https://html.spec.whatwg.org/multipage/webappapis.html#windoworworkerglobalscope-mixin:dom-createimagebitmap
https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#imagebitmapoptions


Liamolucko (approved):
Thanks!

PR #3248, opened 2023-01-19
Adjust examples and documentation to thread-aware `start`

daxpedda:
We didn't adjust the examples and documentation after #3236.
This addresses that.

I honestly wasn't able to test the `wasm-in-web-worker` example because I couldn't get wasm-pack working.

PR #3278, opened 2023-02-02
Revert #3248's changes to the `wasm-in-web-worker` example

Liamolucko:
#3248 tried to change the example to take advantage of #3236 and use a regular `#[wasm_bindgen(start)]` function. However, the start function still gets called on every thread in that example because it doesn't actually use wasm multithreading; each thread's instance has it's own separate memory pool and is completely unaware of the other thread. So, attempting to switch it to use a regular start function caused the worker thread to spawn another one, which does likewise, quickly spiralling out of control.

It also ended up panicking because, thinking it was the main thread, the code was trying to access `window` inside of a worker.

### DIFF
--- a/examples/wasm-in-web-worker/index.js
+++ b/examples/wasm-in-web-worker/index.js
@@ -1,9 +1,18 @@
-async function run_wasm() {
-    console.log('index.js loaded');
+// We only need `startup` here which is the main entry point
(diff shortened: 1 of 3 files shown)


daxpedda:
Thank you for figuring this out!
I didn't actually understand that this is how the example works.

PR #3279, opened 2023-02-02
Remove faulty `location.href` fallback and disable `--split-linked-modules` by default

### LINKED ISSUE #3277: Broken linked modules in no-modules target if shim is not parsed in document
### Describe the Bug

If the no-modules target is used and the shim is not parsed in a document, like for example in any kind of worker, any linked module will break.

This problem partially stems from https://github.com/rustwasm/wasm-bindgen/pull/3069, which now uses `new URL('path/to/linked/module', script_src)`.

When using the no-modules target, `script_src` can not possibly be correct when not used in a document, which was an oversight from this PR: https://github.com/rustwasm/wasm-bindgen/pull/3169.

The problem is that if you are in a worker and using `importScripts()`, `location.href` can never return the path to the imported script, the shim, which is what we want, instead `location.href` will return the path to the script that starts the worker, which is not useful in this context and in fact will break any path created for linked modules.

### Steps to Reproduce

I can't really get the examples working because wasm-pack doesn't work for me, but the simplest example is using the `Atomics.waitAsync` polyfill in Firefox inside a worker, which currently uses linked modules and will try to create a nested worker, which will fail because `new URL()` will report an invalid base, for example if the base is a blob.

### Solution

I think the solution is two-fold here:
- Remove the fallback to `location.href` if `document` is undefined. It is simply impossible to get the path to the shim with the no-modules target, at least as far as I have research.
- ~~Disable `--split-linked-modules` by default when using the no-modules target.~~
  `--split-linked-modules` is already disabled by default, but not when using `wasm-bindgen-cli-support`, e.g. `wasm-bindgen-test-runner`.

daxpedda:
This removes the faulty fallback to `location.href` on the no-modules target.
> The problem is that if you are in a worker and using `importScripts()`, `location.href` can never return the path to the imported script, the shim, which is what we want, instead `location.href` will return the path to the script that starts the worker, which is not useful in this context and in fact will break any path created for linked modules.

Additionally it disables `--split-linked-modules` by default when using `wasm-bindgen-cli-support`, which was probably an oversight. In return I added a new environment variable `WASM_BINDGEN_SPLIT_LINKED_MODULES` for `wasm-bindgen-test-runner` to enable it for whomever requires it.

Fixes #3277.

### DIFF
--- a/guide/src/reference/cli.md
+++ b/guide/src/reference/cli.md
@@ -122,3 +122,7 @@ using a plugin. Alternatively, you can leave the syntax as is and instead
 manually configure the bundler to copy all files in `snippets/` to the output
 directory, preserving their paths relative to whichever bundled file ends up
 containing the JS shim.
+
+On the no-modules target, `link_to!` won't work if used outside of a document,
+e.g. inside a worker. This is because it's impossible to figure out what the
+URL of the linked module is without a reference point like `import.meta.url`.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,6 +41,8 @@ jobs:
   test_wasm_bindgen:
     name: "Run wasm-bindgen crate tests (unix)"
     runs-on: ubuntu-latest
+    env:
+      WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
     steps:
     - uses: actions/checkout@v2
     - run: rustup update --no-self-update stable && rustup default stable
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -381,9 +381,7 @@ impl<'a> Context<'a> {
                 js.push_str("let script_src;\n");
                 js.push_str(
                     "\
-                    if (typeof document === 'undefined') {
-                        script_src = location.href;
-                    } else {
+                    if (typeof document !== 'undefined') {
                         script_src = new URL(document.currentScript.src, location.href).toString();
                     }\n",
                 );
@@ -720,7 +718,7 @@ impl<'a> Context<'a> {
                     stem = self.config.stem()?
                 ),
                 OutputMode::NoModules { .. } => "\
-                    if (typeof input === 'undefined') {
+                    if (typeof input === 'undefined' && script_src !== 'undefined') {
                         input = script_src.replace(/\\.js$/, '_bg.wasm');
                     }"
                 .to_string(),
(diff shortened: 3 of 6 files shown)


guide/src/reference/cli.md
```diff
@@ -122,3 +122,6 @@ using a plugin. Alternatively, you can leave the syntax as is and instead
 manually configure the bundler to copy all files in `snippets/` to the output
 directory, preserving their paths relative to whichever bundled file ends up
 containing the JS shim.
+
+The no-modules target will not be able to construct a `URL` if not used inside
+a document, e.g. a worker.
```
Liamolucko, line 127: I think this phrasing's a bit cleaner. Aside from that, looks good.

```suggestion
On the no-modules target, `link_to!` won't work if used outside of a document,
e.g. inside a worker. This is because it's impossible to figure out what the
URL of the linked module is without a reference point like `import.meta.url`.
```

lukaslihotzki:
I'm sorry for the trouble my PR caused.

Can `script_src` simply be set to a relative path instead? Like `./filename.js` (with the filename wasm-bindgen uses). Then, `link_to!` would also just return relative paths. (Relative paths break web pages using the history API to change the current URL, but this isn't applicable to workers.)

Liamolucko:
> Can `script_src` simply be set to a relative path instead? Like `./filename.js` (with the filename wasm-bindgen uses). Then, `link_to!` would also just return relative paths. (Relative paths break web pages using the history API to change the current URL, but this isn't applicable to workers.)

No, because relative paths get resolved by `fetch`, `new Worker` etc. relative to `location.href`, not to the path of the current file. So, it runs into the the same problems as setting `script_src` to `location.href`.

Liamolucko (approved):

lukaslihotzki:
Maybe I should not have hidden this in an edit, but have you seen my suggestion to use https://github.com/chemicstry/wasm_thread/blob/main/src/script_path.js?

daxpedda:
> Edit: This may not work if relative paths are still resolved to the original URL and not the `importScripts`-URL. However, can we use this: https://github.com/chemicstry/wasm_thread/blob/main/src/script_path.js

Just tried it and it does work indeed, but I don't think wasm-bindgen should use this, it isn't exactly stable or standardized, to say the least, it's a pretty hacky workaround.

lukaslihotzki:
I looked it up and Error.stack is indeed undefined by the ES standard.

PR #3296, opened 2023-02-11
Expose `WebAssembly.Instance.exports`

### LINKED ISSUE #2513: Add function to access instance exports or the instance itself.
### Motivation

I'm trying to write a library that can optionally work-alongside `wasm-bindgen` or stand-alone without `wasm-bindgen` based on a feature flag.

My library has some Javascript that requires access to the Wasm instance's exports (to communicate independent of `wasm-bindgen`), but right now it's quite tricky to expose the instance's exports to my library.

The approach I've found is to stash the value retuned by `wasm-bindgen`'s `init` on a global variable to access later:
```rust
const wasm = await init();
self.kwasm_exports = wasm;
```

A problem with this approach is that `main` is called by `init` so if my library is called by `main` then it won't yet have access the Wasm exports. A hack is to add a fake entry point to my library and then later call the entry point manually but that's a pretty invasive approach.

Ideally a library user should not have to change any of their Javascript code, they just need to flip a feature flag in my library. 

### Proposed Solution

If `wasm-bindgen` exposes a Rust function to access the instance's exports (or the instance itself) instead then that could be used instead to pass the information 

This feels similar in purpose to functions like `wasm-bindgen`'s [memory](https://docs.rs/wasm-bindgen/0.2.73/wasm_bindgen/fn.memory.html) function.

Such a function would be a useful escape-hatch for scenarios that are a little bit outside the norm, like my own.

daxpedda:
I was specifically attempting to use https://github.com/rustwasm/wasm-bindgen/pull/3177 but noticed it was impossible to call `__wbindgen_thread_destroy()` or get `__tls_base` and `__stack_alloc` from inside Rust.

~~Alternatively we could expose `WebAssembly.Instance.prototype.exports` directly instead.~~
EDIT: I decided to go ahead and expose the exports directly. Mainly because there were some targets that were much harder for me to implement exposing the instance for.

Or if we don't want a general-purpose solution like this we could expose  `__wbindgen_thread_destroy()`, `__tls_base` and `__stack_alloc` through more intrinsics.

I'm not aware of any downsides.
EDIT: I was thinking that maybe exposing exports directly is worse then exposing the instance, but it seems that you can't do anything else with the instance anyway: https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Instance.

Fixes #2513.

### DIFF
--- a/crates/cli-support/src/intrinsic.rs
+++ b/crates/cli-support/src/intrinsic.rs
@@ -252,6 +252,9 @@ intrinsics! {
         #[symbol = "__wbindgen_memory"]
         #[signature = fn() -> Externref]
         Memory,
+        #[symbol = "__wbindgen_exports"]
+        #[signature = fn() -> Externref]
+        Exports,
         #[symbol = "__wbindgen_module"]
         #[signature = fn() -> Externref]
         Module,
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -3498,6 +3498,11 @@ impl<'a> Context<'a> {
                 format!("init.__wbindgen_wasm_module")
             }
 
+            Intrinsic::Exports => {
+                assert_eq!(args.len(), 0);
+                "wasm".to_string()
+            }
+
             Intrinsic::Memory => {
                 assert_eq!(args.len(), 0);
                 let mut memories = self.module.memories.iter();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1082,6 +1082,7 @@ externs! {
 
         fn __wbindgen_not(idx: u32) -> u32;
 
+        fn __wbindgen_exports() -> u32;
         fn __wbindgen_memory() -> u32;
         fn __wbindgen_module() -> u32;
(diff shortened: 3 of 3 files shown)


daxpedda:
I noticed that this doesn't work without `externref`.

daxpedda:
@alexcrichton would you mind getting this in before the next release (#3400)?

alexcrichton:
Sure!

allsey87:
Quick question: is there any reason why:
```
wasm_bindgen::exports().js_typeof().as_string().unwrap()
```
is "object", while:
```
let exports = wasm_bindgen::exports().dyn_into::<js_sys::Object>().unwrap();
```
panics? From my testing, the `js_sys::Object` returned by `unchecked_into` seems to be correct/to work...

daxpedda:
This is very strange.

It has nothing to do with `wasm-bindgen` though, I tried it out and apparently `Instance.exports` returns false on `exports instanceof Object`, I don't know why.
But `typeof exports` returns `"Object"` and the developer console says that it's an `Object` as well.

No clue!

PR #3300, opened 2023-02-12
Clippy: Fixes and CI

lukaslihotzki:
This is the next big step after #3292, inspired by #3271. The remaining RUSTFLAGS exceptions and the examples code are left as future work.


.github/workflows/main.yml
```diff
@@ -38,6 +38,50 @@ jobs:
...
+        -A non_upper_case_globals
+        -A unused_doc_comments
+        -A clippy::disallowed_names
+        -A clippy::enum_variant_names
+        -A clippy::large_enum_variant
+        -A clippy::missing_safety_doc
+        -A clippy::new_without_default
+        -A clippy::overly_complex_bool_expr
+        -A clippy::too_many_arguments
+        -A clippy::type_complexity
```
daxpedda, line 58: I think it would be really important to put these lints into the code somehow, otherwise contributors won't notice an issue until the make a PR and it fails on CI.

We could add them to a global `.cargo/config.toml`?

lukaslihotzki: I agree. The problem with `.cargo/config.toml` is that the file shouldn't be checked in to git, like [the comment](https://github.com/rustwasm/wasm-bindgen/blob/0.2.84/.cargo/config#L1-L2) suggests. I'll try to move these exceptions into the crates that need it (shouldn't be that much duplication).

daxpedda: I'm not sure why it was decided that this shouldn't be checked into git, it's common solution, the Cargo documentation also doesn't indicate that one shouldn't check this into git either.

lukaslihotzki: Ok, I have added .config/cargo with allow rustflags. Warnings should not be generally denied, so the CI still needs to set RUSTFLAGS manually, which overrides .cargo/config, so the allows are duplicated once. I added comments in both directions, so it should be acceptable.

lukaslihotzki:
@Liamolucko, can you please review this PR? It contains trivial changes. Is there anything I should change?

lukaslihotzki:
@Liamolucko ping. This PR gets a lot of merge conflicts, can you tell when you can review it, so I can rebase it before?

daxpedda:
@lukaslihotzki I can review and merge this, so a rebase would be welcome!

crates/cli-support/src/wit/mod.rs
```diff
@@ -1525,15 +1519,15 @@ fn verify_schema_matches<'a>(data: &'a [u8]) -> Result<Option<&'a str>, Error> {
...
+    if !data.starts_with('{') || !data.ends_with('}') {
         bad!()
     }
     let needle = "\"schema_version\":\"";
     let rest = match data.find(needle) {
         Some(i) => &data[i + needle.len()..],
         None => bad!(),
     };
-    let their_schema_version = match rest.find("\"") {
+    let their_schema_version = match rest.find('\"') {
```
daxpedda, line 1530:
```suggestion
    let their_schema_version = match rest.find('"') {
```

crates/cli-support/src/wit/mod.rs
```diff
@@ -1545,15 +1539,15 @@ fn verify_schema_matches<'a>(data: &'a [u8]) -> Result<Option<&'a str>, Error> {
         Some(i) => &data[i + needle.len()..],
         None => bad!(),
     };
-    let their_version = match rest.find("\"") {
+    let their_version = match rest.find('\"') {
```
daxpedda, line 1542:
```suggestion
    let their_version = match rest.find('"') {
```

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -231,12 +227,15 @@ integration test.\
...
+                println!();
                 println!("Note that interactive mode is enabled because `NO_HEADLESS`");
                 println!("is specified in the environment of this process. Once you're");
                 println!("done with testing you'll need to kill this server with");
                 println!("Ctrl-C.");
-                return Ok(srv.run());
+                return {
+                    srv.run();
+                    Ok(())
+                };
```
daxpedda, line 238:
```suggestion
                srv.run();
                return Ok(());
```

crates/cli/tests/interface-types.rs
daxpedda, line 1: Oversight during rebasing I assume.

crates/externref-xform/src/lib.rs
```diff
@@ -18,7 +18,7 @@
 use anyhow::{anyhow, bail, Error};
 use std::cmp;
 use std::collections::{BTreeMap, HashMap, HashSet};
-use std::mem;
+
```
daxpedda, line 21: Let's keep it and use `mem::take` below.

crates/webidl-tests/enums.rs
```diff
@@ -28,10 +28,12 @@ fn invalid_enum_return() {
...
     assert!(!actually_a_triangle.is_square());
-    match actually_a_triangle.get_shape() {
-        ShapeType::Circle | ShapeType::Square => assert!(false),
-        _ => {} // Success
-    };
+    assert!(
+        !(matches!(
+            actually_a_triangle.get_shape(),
+            ShapeType::Circle | ShapeType::Square
+        ))
```
daxpedda, line 35:
```suggestion
        !matches!(
            actually_a_triangle.get_shape(),
            ShapeType::Circle | ShapeType::Square
        )
```

Cargo.toml
```diff
@@ -12,6 +12,7 @@ description = """
 Easy support for interacting between JS and Rust.
 """
 edition = "2018"
+rust-version = "1.56"
```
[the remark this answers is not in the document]
daxpedda: In any case, we can do this after this PR, let's not rebase this again.

crates/backend/src/ast.rs
```diff
@@ -349,7 +351,7 @@ pub struct Function {
 }
 
 /// Information about a Struct being exported
-#[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
```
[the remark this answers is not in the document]
lukaslihotzki: This is intentional, but I moved the fix to #3434 to stress its impact.

.github/workflows/main.yml
```diff
@@ -38,6 +38,49 @@ jobs:
...
+        -D warnings
+        -A dead_code
+        -A non_upper_case_globals
+        -A unused_doc_comments
+        -A clippy::large_enum_variant
+        -A clippy::missing_safety_doc
+        -A clippy::new_without_default
+        -A clippy::overly_complex_bool_expr
+        -A clippy::too_many_arguments
+        -A clippy::type_complexity
```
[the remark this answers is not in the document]
lukaslihotzki: No, the `RUSTFLAGS` environment variable replaces the `rustflags` from `.cargo/config`. You can try it out with:
```
cargo clippy --no-deps --all-features -p wasm-bindgen-threads-xform
export RUSTFLAGS="-D warnings"
cargo clippy --no-deps --all-features -p wasm-bindgen-threads-xform
```

The first run prints no warning, the second complains about `clippy::new-without-default` again.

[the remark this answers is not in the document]
lukaslihotzki: Ah, you meant using `--` instead of setting `RUSTFLAGS` in the environment.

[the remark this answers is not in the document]
lukaslihotzki: Excellent, with your approach, I can re-add the `webidl-tests`.

Cargo.toml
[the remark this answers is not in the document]
lukaslihotzki: Setting `rust-version = "1.56"` is safe (not introducing any breakage), because wasm-bindgen-macro-support (an integral crate) depends on syn 2.0 which depends on edition 2021 which is provided since Rust 1.56. Therefore, we don't have to announce any MSRV constraint.

Setting the `rust-version` too low is safe and prevents lints that require more recent Rust versions (as you said). wasm-bindgen doesn't have a guaranteed MSRV, and changing this is orthogonal to this MR. If a too low rust version was confusing, we could also add a comment that wasm-bindgen isn't guaranteed to work on such old Rust versions.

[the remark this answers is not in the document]
daxpedda: Yeah, you are right.

daxpedda (approved):
Thank you, amazing work!

PR #3305, opened 2023-02-13
Add name of the type that was failed to borrow

bouk:
This aids debugging a lot


alexcrichton (commented):
One of the intentions for these functions was to have a small code-size impact, so could the message here use `cfg!(debug_assertions)` for a proxy about whether to factor in the string or not?

bouk:
Yea I just realized this can bloat the binary quite a bit—I can make that change.

PR #3308, opened 2023-02-15
Allow non-literal constants in `typescript_custom_section`

### LINKED ISSUE #2828: typescript_custom_section doesn't accept the result of include_str macro

71:
Also:
- Bump version to 0.2.84, since the internal schema is changing.
- Bump MSRV for UI tests to 1.57 to allow `assert!` in `const fn`s.
- Update UI test outputs.

This will, among other things, allow `include_str!` in `#[wasm_bindgen(typescript_custom_section)]` (and fix #2828). This also opens the door for non-literal constants in `inline_js` (thus fixing #2787) with minimal changes.

This is a clone of #3115, which was closed by GitHub when I renamed the `main` branch.

### DIFF
--- a/crates/cli-support/src/decode.rs
+++ b/crates/cli-support/src/decode.rs
@@ -5,7 +5,7 @@ pub trait Decode<'src>: Sized {
 
     fn decode_all(mut data: &'src [u8]) -> Self {
         let ret = Self::decode(&mut data);
-        assert!(data.len() == 0);
+        assert_eq!(data.len(), 0);
         return ret;
     }
 }
@@ -152,6 +152,13 @@ macro_rules! decode_api {
         decode_enum!($name () $($variants)*);
         decode_api!($($rest)*);
     );
+}
+
+mod syn {
+    /// Alias `syn::Expr` to `str` when decoding; an expression is evaluated as
+    /// a Rust constant during compilation, so when we decode the buffer we can
+    /// expect a string instead.
+    pub type Expr = str;
 }
 
 wasm_bindgen_shared::shared_api!(decode_api);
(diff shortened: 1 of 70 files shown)


crates/cli-support/src/decode.rs
```diff
@@ -154,4 +154,11 @@ macro_rules! decode_api {
     );
 }
 
+mod syn {
+    /// Alias `syn::Expr` to `str` when decoding; an expression is evaluated as
+    /// a Rust constant during compilation, so when we decode the buffer we can
+    /// expect a string instead.
+    pub type Expr = str;
+}
```
daxpedda, line 162: Can we pass this into the macro instead?

daxpedda (commented):
Unfortunately I can't really review the TS part (which is just a test, I know), so I will ping another maintainer when we figure things out internally.

Needs a rebase.

71:
Sorry, I do plan to come back to this PR to respond to comments / rebase / apply suggested changes, but I've been pretty busy and haven't gotten around to doing it.

PR #3310, opened 2023-02-15
Fix large pointers from WASM

### LINKED ISSUE #2388: Can't work with > 2GiB memory due to ptr being interpreted as i32
I get this error when allocating > 2GiB and then calling into JS:

```
Failed to execute 'decode' on 'TextDecoder': The encoded data was not valid.
    at getStringFromWasm0 (fu0min2hevxxsl6tfui2k8dfe:44)
    at imports.wbg.__wbg_setnodeValue_4a75b94edda71829 (fu0min2hevxxsl6tfui2k8dfe:360)
```

It's due to the `ptr` passed to `getStringFromWasm0` being negative. If I add 2^32 to it (or do `ptr >>> 0`), then it is correct. I assume at some point it's being mistakenly interpreted as `i32`.

Here's the test case I used to narrow this down if it's useful:
```rust
use web_sys::{HtmlBodyElement, HtmlDocument};

#[wasm_bindgen(start)]
pub fn main() -> Result<(), JsValue> {
    let window = web_sys::window().unwrap();
    let document: HtmlDocument = window.document().unwrap().dyn_into().unwrap();
    let body: HtmlBodyElement = document.body().unwrap().dyn_into().unwrap();
    let x = document.create_text_node("abc");
    body.append_child(&x).unwrap();
    for i in 1_000_000..2_000_000 {
        let y = (0..(512 * 1024 * 1024))
            .map(|y| y as u8)
            .collect::<Vec<u8>>();
        std::mem::forget(y);
        let s = format!(
            "{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}{}",
            i, i, i, i, i, i, i, i, i, i, i, i, i, i, i, i, i, i, i, i
        );
        x.set_node_value(Some(&s));
        std::mem::forget(s);
    }
    Ok(())
}

```

### LINKED ISSUE #2957: Reading wrong address when memory usage exceeds 2GB
### Describe the Bug

When memory usage exceeds 2GB, the glue code generated by wasm-pack reads the wrong address.

### Steps to Reproduce

The following program uses 3GB of memory.

```rust
#[wasm_bindgen]
pub struct HugeData {
    _data1: Vec<u32>,
    data2: Vec<u32>,
}

#[wasm_bindgen]
impl HugeData {
    pub fn new() -> Self {
        let mut _data1 = vec![0; 3 << 27]; // 1.5 GB
        let mut data2 = vec![0; 3 << 27]; // 1.5 GB
        for i in 0..(3 << 27) {
            _data1[i as usize] = i;
            data2[i as usize] = i + (3 << 27);
        }
        Self { _data1, data2 }
    }

    pub fn get_last_element_as_boxed_slice(&self) -> Box<[u32]> {
        self.data2[(3 << 27) - 1..].to_vec().into_boxed_slice()
    }
}
```

### Expected Behavior

`get_last_element_as_boxed_slice()` should return `Uint32Array [ 805306367 ]`, where `805306367` is equal to `(3 << 28) - 1`.

### Actual Behavior

In my environment, `Uint32Array [ 537182218 ]` was returned.

The glue code generated by wasm-pack is as follows:

```javascript
/**
* @returns {Uint32Array}
*/
get_last_element_as_boxed_slice() {
    try {
        const retptr = wasm.__wbindgen_add_to_stack_pointer(-16);
        wasm.hugedata_get_last_element_as_boxed_slice(retptr, this.ptr);
        var r0 = getInt32Memory0()[retptr / 4 + 0];
        var r1 = getInt32Memory0()[retptr / 4 + 1];
        var v0 = getArrayU32FromWasm0(r0, r1).slice();
        wasm.__wbindgen_free(r0, r1 * 4);
        return v0;
    } finally {
        wasm.__wbindgen_add_to_stack_pointer(16);
    }
}
```

Here, `var r0` is a **signed** 32-bit integer, and it is negative when memory usage exceeds 2GB. Therefore, `getArrayU32FromWasm0(r0, r1)` reads memory of negative address, which causes this bug.

kristoff3r:
Pointers being passed from WASM to JS are interpreted as signed, so large pointers ended up negative. This prevented WASM programs from allocating more than 2GB.

To fix it we make pointer use unsigned (via >>> 0) for all malloc/realloc calls, and for all pointer arguments while generating the shim functions. Ideally there would be an abstraction that guarantees we do it in all cases, but that would require a major refactor.

To test it we use gg-alloc as the global allocator while running tests. This allocator was made specifically for testing this issue, but was never upstreamed. It only allocates memory above 2GiB.

Fixes #2388
Fixes #2957


kristoff3r:
Hmm, looking at the changes to the reference files it might be simpler to remove `pop_ptr` and just fix it inside `getStringFromWasm` instead.

alexcrichton:
Thanks! And yeah I think it's fine to fix without tests for now, allocating 2G in the browser tests is likely going to cause flakiness as well.

kristoff3r:
I decided to only change stuff in `cli-support/src/js/mod.rs`, as it seems cleaner (no pointers being fixed multiple times, avoids a half-baked abstraction). 

I also made gg-alloc an optional feature for wasm-bindgen-test, and with it the tests now pass, and they fail if I remove the fix.

emilk:
Thanks for fixing this - we run into this issue problem frequently in our memory intensive application.

Any plans for a new release with this critical fix? 🙏

Keavon:
@kristoff3r @alexcrichton quick question: does this fix double the available memory from 2GB to 4GB, or does it increase the cap beyond 4GB (to what?)? Thanks.

daxpedda:
Only to 4GB. Beyond 4GB requires wasm64 support.

PR #3312, opened 2023-02-16
Update expected results for UI tests

Liamolucko:
Apparently, #3293 didn't quite update them properly, causing CI to fail. I'm not sure why CI passed when checking the PR but failed on `main`, but this should fix the CI failure.


daxpedda (approved):
Didn't notice the PR, was making #3313 instead.

PR #3318, opened 2023-02-17
Replace CanvasCaptureMediaStream with CanvasCaptureMediaStreamTrack

allsey87:
`web-sys` currently has a binding for `CanvasCaptureMediaStream`, however there are no mentions of this interface on MDN or in W3C editors draft. The only mentions of this interface are floating around on [Stack Overflow from 2019](https://stackoverflow.com/search?q=CanvasCaptureMediaStream) and on [this page](https://reference.codeproject.com/dom/CanvasCaptureMediaStream) where it seems that perhaps Firefox once implemented it.

It seems that the [`HTMLCanvasElement.captureStream`](https://w3c.github.io/mediacapture-fromelement/#dom-htmlcanvaselement-capturestream) method now returns a [`MediaStream`](https://www.w3.org/TR/mediacapture-streams/#dom-mediastream) instance instead of a `CanvasCaptureMediaStream`. This `MediaStream` instance now contains a [`CanvasCaptureMediaStreamTrack`](https://w3c.github.io/mediacapture-fromelement/#the-canvascapturemediastreamtrack), which provides the [`requestFrame`](https://w3c.github.io/mediacapture-fromelement/#dom-canvascapturemediastreamtrack-requestframe) method and [`canvas`](https://w3c.github.io/mediacapture-fromelement/#dom-canvascapturemediastreamtrack-canvas) attribute that used to exist on `CanvasCaptureMediaStream`.

This PR should bring our bindings for canvas capture up to date. If someone could confirm that the `Pref` attribute in the WebIDL is correct that would be great. I cannot find any documentation on this and the content of these strings seems a bit random to me.


allsey87:
See https://github.com/w3c/mediacapture-fromelement/issues/76 for more details

Liamolucko:
This is a breaking change, which I'd rather avoid. I don't think there'd be much harm in just leaving the old API there in addition to the new API, so perhaps we could do that instead?

As for `Pref`, it's a non-standard extension used by Mozilla that's a holdover from when all of our WebIDL was imported from Firefox. It's completely ignored by the WebIDL bindings generator, so feel free to get rid of it. (Firefox uses it to keep track of which preference in about:config controls whether the API is available.)

allsey87:
Ok, I will remove `Pref` and leave the old API in place. I was surprised today to see that the latest version of Firefox at the time of writing still returns a `CanvasCaptureMediaStream` from `captureStream`.

allsey87:
@Liamolucko actually, I do not think this is a breaking change after all since I am proposing to add `captureStream` in #3317 and without this method, `CanvasCaptureMediaStream` is inaccessible via web-sys, right?

Liamolucko:
No, it can accessed via. `JsCast` (`dyn_into` and friends). The fact that it can *only* be accessed that way does make it a more minor breaking change, which I would be willing to make if necessary, but I don't really see the need when it can be left alone (or deprecated).

It's also not *entirely* obsolete, as Firefox hasn't yet updated its implementation of `captureStream` which still uses `CanvasCaptureMediaStream`.

allsey87:
Ok, so let's leave the old `CanvasCaptureMediaStream` in place and additionally introduce the newer, standardised `CanvasCaptureMediaStreamTrack`. `captureStream` will be added as per the specification and return an instance of `MediaStream`. I will close this PR and #3317 in favour of a new PR that includes both changes and is based of the current main branch.

allsey87:
Closing in favor of #3325

PR #3320, opened 2023-02-17
fix #3307: do nothing when calling init multiple times

### LINKED ISSUE #3307: Prevent accidental re-initialization
### Motivation

Calling `init`, or other initialization functions twice accidentally can cause subtle bugs. It should check if it ever got initialized before and just return if it did.

### Proposed Solution

In `finalizeInit`, perhaps a check like `if (wasm != null) return wasm;` as the first operation.

### Alternatives

Well, personally I'd love to use wasm-bindgen generated modules similar to how they've been designed in the browser APIs: During initialization, you provide the imports and the initialization call gives you all exports. Imports and exports are tied to a specific wasm module instance.
However, I realize two things:
* The current design is so that you can import wasm-bindgen generated projects in nodejs as if they were a normal library, without having to initialize anything.
* It would be quite a breaking/effortful change to provide a different API, *especially* if [eventually browsers just support ELM-style .wasm imports directly](https://github.com/WebAssembly/esm-integration/blob/main/proposals/esm-integration/README.md).

### Additional Context

When we called `init` twice, the issue we got was actually *really* subtle and annoying to debug:
We used the weak refs feature, i.e. `FinalizationRegistry`s were generated for us.

So in the main wasm-bindgen-generated `project.js` file (we used `wasm-pack build --target web`) we have following code generated:
```js
let wasm;

// ...

const PublicNodeFinalization = new FinalizationRegistry(ptr => wasm.__wbg_publicnode_free(ptr));

// ...

function finalizeInit(instance, module) {
    wasm = instance.exports;
    cachedInt32Memory0 = null;
    cachedUint32Memory0 = null;
    cachedUint8Memory0 = null;


    return wasm;
}
```

Now, imagine you cause `finalizeInit` to run once, then the `wasm` variable will be set to some module.
In our code, we now continued running operations on our `PublicNode`. This causes them to be allocated and returned to JS and registered in their respective `FinalizationRegistry`.
Please note that **the FinalizationRegistry closes over the wasm variable**.
So now our code accidentally called `finalizeInit` again, which **overwrites the last wasm value**, and created a buch more `PublicNode`s.

When eventually enough `PublicNode`s were created, GC kicked in and deleted some older ones. These older ones then had their finalizers run and that **caused the finalizers to be run on the "wrong" wasm memory**. The `ptr` parameter wouldn't make sense in the new `wasm` instance.

Unfortunately that was really hard to debug, since finalizers only run when you've accumulated a bunch of garbage.
There's also other oddities like the error causing a "memory access out of bounds" (likely due to the `ptr` value being too big for the new wasm instance), but straight up causing a SIGBUS in nodejs on M1 Macs.

erwanvivien:


Liamolucko (commented):
Thanks for this! I think it may be a better idea to throw an error when `init` is called twice rather than silently returning, though - that's probably a mistake someone would want to know about, and I think it's a bit weird for the arguments that get passed to `init` to just be silently ignored.

erwanvivien:
I think you're right, if a maintainer wants the change, I'll do it!

erwanvivien:
In fact, I think it would break in React like website, as React launches startup function twice in dev mode, meaning `init` would get called twice, and it would break

However, having an error log might be great

Liamolucko:
I don't really know anything about React. Is it not possible to just call `init` at the top level of the module, or keep track of whether you've called `init` yet and only call it if you haven't already?

erwanvivien:
You can keep track of course, and that's what apps use. However internally (and only in dev mode) they launch twice every component initialisation and destruction to assert you're cleaning the used items

I might be not clear, but it'll call init twice, I'm not 100% sure it would break, but it needs to be taken into account I guess?

daxpedda:
Can we extend the same check to `initSync`?
> I think you're right, if a maintainer wants the change, I'll do it!

Liamolucko has merge rights, currently he is probably your best bet to get anything merged in this repo.

erwanvivien:
Thanks for the heads up @daxpedda

erwanvivien:
> In fact, I think it would break in React like website

More precision on this: It would not break React website. But I think it would break the dev workflow, see [React Strict Mode](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects).

erwanvivien:
It's alright, thanks 👌

Issue #1841, opened 2019-10-29
Split crate into "standalone" and "JS" components


alexcrichton:
Over time `wasm-bindgen` has picked up more features to follow the WebAssembly proposals, namely the [interface types](https://github.com/webassembly/interface-types) proposal. Additionally `wasm-bindgen` has been used for new embeddings such as WASI! One of the major goals of interface types is to also help bring wasm code to many more places outside of the web browser.

Currently for this "outside the web browser" use case `wasm-bindgen` is acting as Rust's implementation of interface types (to follow the standard), and I think that's likely to continue being the case into the future. Unfortunately though this causes a bit of an impedance mismatch!

* From the start `wasm-bindgen` was built for the web and assumes JS. Lots of functionality isn't available in standalone webassembly (almost none of the APIs on `JsValue`, classes, etc)
* Currently debug builds, by default, fail for the interface types target because they don't optimize out calls to `__wbindgen_throw`. There are no intrinsics with a standalone wasm file!
* Restrictions on the web are not actually restrictions everywhere else. We support u64/i64 on the web by passing two i32 components, but https://github.com/rustwasm/wasm-bindgen/issues/1837 points out this is natively supported by all other runtimes.

Overall I think that to continue to faithfully pursue the standalone outside-the-web browser use case we'll want to do something about the current organization. I don't think this should be isolated to just the `wasm32-wasi` target, but it should include it!

I would expect an organization where you still use `#[wasm_bindgen]`-the-macro but if you're in a standalone environment you'd have a type like `Anyref` instead of `JsValue`. Additionally you would have no access to intrinsics in a standalone environment, unlike the JS environment. I think we'll want to tweak the features available in `#[wasm_bindgen]` as well, where if you're in a standalone environment we require a `module` for all imports and don't support any features like `js_namespace` and such.

I sort of envision this happening by doing something like the following:

* Spilt out a `wasm-bindgen-core` crate. This crate defines the `Anyref` type, uses zero intrinsics, and defines all the basic conversion traits.
* Add a dependency from `wasm-bindgen-core` to `wasm-bindgen-macro`, and reexport it
* The current `wasm-bindgen` crate would then depend on `wasm-bindgen-core`, reexporting most of its functionality and layering `JsValue` on top of `Anyref`.
[...]

PR #3324, opened 2023-02-19
Implemented compatibility mode with wasm32-wasi

Rob2309:
This PR adds the new --wasi-abi flag to wasm-bindgen. With this flag, wasm-bindgen generates bindings that are compatible with the ABI used by the rust target wasm32-wasi.
The changes should not affect the behavior of wasm-bindgen when the flag is not specified.

For implementing the behavior of wasm32-wasi, I introduced a few new Instructions to the "Standard".
They are `PackSlice`, `UnpackSlice`, `PackOption` and `UnpackOption`. These instructions translate between the passed-by-pointer structs used in wasm32-wasi and the format expected by the rest of wasm-bindgen.

I am very open to suggestions on this.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,3 +2,6 @@
 # that.
 [target.wasm32-unknown-unknown]
 runner = 'cargo run -p wasm-bindgen-cli --bin wasm-bindgen-test-runner --'
+
+[target.wasm32-wasi]
+runner = 'cargo run -p wasm-bindgen-cli --bin wasm-bindgen-test-runner --'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,50 @@ jobs:
         WASM_BINDGEN_EXTERNREF: 1
         NODE_ARGS: --experimental-wasm-reftypes
 
+  test_wasi_abi:
+    name: "Run wasm-bindgen crate tests with --wasi-abi (unix)"
+    runs-on: ubuntu-latest
+    env:
+      WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
+      TEST_WASI_ABI: 1
+    steps:
+    - uses: actions/checkout@v3
+    - run: rustup update --no-self-update stable && rustup default stable
+    - run: rustup target add wasm32-wasi
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16'
+    - uses: ./.github/actions/setup-geckodriver
+    - run: cargo test --target wasm32-wasi
+      env:
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: ./tests/wasi/wasi_browser.js
+    - run: cargo test --target wasm32-wasi --features serde-serialize
+      env:
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: ./tests/wasi/wasi_browser.js
+    - run: cargo test --target wasm32-wasi --features enable-interning
+      env:
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: ./tests/wasi/wasi_browser.js
+    - run: cargo test --target wasm32-wasi -p no-std
+    # - run: cargo test --target wasm32-wasi -p wasm-bindgen-futures
+    - run: cargo test --target wasm32-wasi --test wasm
+      env:
+        WASM_BINDGEN_WEAKREF: 1
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: tests/wasi/wasi_node.js
+    - run: cargo test --target wasm32-wasi --test wasm
+      env:
+        WASM_BINDGEN_WEAKREF: 1
+        WASM_BINDGEN_NO_DEBUG: 1
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: tests/wasi/wasi_node.js
+    - run: cargo test --target wasm32-wasi --test wasm --features serde-serialize
+      env:
+        WASM_BINDGEN_WEAKREF: 1
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: tests/wasi/wasi_node.js
+    - run: cargo test --target wasm32-wasi
+      env:
+        WASM_BINDGEN_EXTERNREF: 1
+        NODE_ARGS: --experimental-wasm-reftypes
+        REROUTE_WASI_SNAPSHOT_PREVIEW1: tests/wasi/wasi_node.js
+
   test_threads:
     name: "Run wasm-bindgen crate tests with multithreading enabled"
     runs-on: ubuntu-latest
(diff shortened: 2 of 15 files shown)


Rob2309:
It seems like my local setup is not similar enough to the CI, as for me all tests with exactly the same environment variables run without error...

Rob2309:
I am thinking about implementing the WASI systemcalls as optional intrinsics, so that a wasi environment is not needed when running the generated module.

Rob2309:
I just added the environment variable flag "WASM_BINDGEN_EMULATE_WASI" that introduces many WASI systemcalls as intrinsics. While not all syscalls are implemented yet, the tests run fine on wasm32-wasi without any external wasi shims.

daxpedda:
What is the exact use-case of `wasm-bindgen` for WASI? As far as I understood https://github.com/rustwasm/wasm-bindgen/issues/1841, it was for things like using the interface proposal, but that never came into fruition.

So currently I'm not sure what the purpose of `wasm-bindgen` is without access to JS. Right now I'm more in favor of https://github.com/rustwasm/wasm-bindgen/pull/3233, which seeks to disable the functionality of `wasm-bindgen` when targeting WASI, like it works for non-Wasm targets right now.

I'm wholly ignorant about WASI btw, so I'm probably missing the big-picture here.

Liamolucko:
> Apologies for any cluelessness here: would it be possible to compile Rust with the usual `wasm32-unknown-unknown` target while linking to a C library compiled with `wasm32-wasi`?

It *should* be possible, but currently isn't, mainly because Rust's `"C"` ABI on `wasm32-unknown-unknown` doesn't match the one used by clang (which is defined in https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md). There are a lot of different issues opened about this, but rustwasm/team#291 seems to be the main one.

The difference between them is that Rust flattens structs out into one parameter per field, whereas clang passes them via. a pointer. The `wasm32-wasi` target does match, which is why it works for linking to C code.

There's also then the issue of somehow providing the WASI imports needed by the C library, which would probably involve having to add a way to pass wasm imports to `init`. I'm not sure how you could get that to work on targets without `init`, though.

daxpedda:
> There's also then the issue of somehow providing the WASI imports needed by the C library, which would probably involve having to add a way to pass wasm imports to `init`. I'm not sure how you could get that to work on targets without `init`, though.

Maybe that's the part we could provide here in `wasm-bindgen`, but that's useless right now if the problem is still in Rust.

hamza1311:
As I understand it, this will allow compiling Rust code to `wasm32-wasi` and executing it _in browser_, right? I'm in favor of this as long as the compilation is for `wasm32-unknown-unknown` target but the generated code is compatible with wasm32-wasi ABI. That should allow C code.

That combined with https://github.com/rustwasm/wasm-bindgen/pull/3233, means that wasm32-wasi is _explicitly incompatible_ with wasm-bindgen but there is an option to use an ABI, that is the same as clang, thus allowing linking with C code

Question: assuming my understanding is correct (please correct me if it isn't), why not make this the default?

daxpedda:
> Question: assuming my understanding is correct (please correct me if it isn't), why not make this the default?

What do you mean with "default"?

hamza1311:
> > Question: assuming my understanding is correct (please correct me if it isn't), why not make this the default?
> 
> What do you mean with "default"?

The wasi-abi flag this PR introduces. Why not make it the default?

daxpedda:
Yeah, that sounds reasonable to me too.

Rob2309:
> The wasi-abi flag this PR introduces. Why not make it the default?

The problem I see with this is that this ABI is currently completely incompatible with any wasm target except wasm32-wasi. I don't think there is a good way to make this the default without breaking most stuff.

The main usecase I had for this PR is allowing to run a rust program linked to C code in a browser.

temeddix:
1+ for making `wasm32-wasi` the defaut. It will allow Rust wasm modules to use many things that wasn't possible before. Things like `std::thread` or `std::time::Instant`

daxpedda:
> 1+ for making `wasm32-wasi` the defaut. It will allow Rust wasm modules to use many things that wasn't possible before. Things like `std::thread` or `std::time::Instant`

`wasm-bindgen` doesn't control how you compile your Rust code. We were discussing the default for adding the shim, not which target is chosen.

The whole point here is to build on what Hamza said:
> I'm in favor of this as long as the compilation is for `wasm32-unknown-unknown` target but the generated code is compatible with wasm32-wasi ABI. That should allow C code.

Though I guess making `wasm-bindgen` compatible with Rust code that is compiled to `wasm32-wasi` is possible as well, it is a different discussion.

temeddix:
Thanks for the correction :)

I think it's time for reviews, as this PR basically fixes some limitations that arive from `wasm32-unknown-unknown`. `wasm32-wasi` is actively being developed and popular crates like `tokio` are adding support for it. Would this change make into the master? @Rob2309 mentioned that he's open to suggestions.

daxpedda:
As mentioned above, I agree mostly with Hamza that I would prefer a solution that adds WASI compatibility by adding a shim to make linked WASI libraries work. This means Rust should be compiled with `wasm32-unknown-unknown` but linked C/C++ libraries *can* be compiled with `wasm32-wasi`.

This isn't a solution that can be applied today because of #3454, so I would argue that fixing #3454 is the way forward right now.

Supporting `wasm32-wasi` directly can be discussed here: https://github.com/rustwasm/wasm-bindgen/issues/3421. Though I would generally prefer not to do that for reasons I wrote up there, but I'm happy to hear more arguments and views.

hamza1311:
To add to what I said previously, I would rather have wasm32-unknown-unknown be compatible with C ABI than wasm32-wasi be supported. wasm32-wasi has functions like file system access which obviously can't be provided in a browser environment so I'm a little hesitant to add wasm32-wasi compatibility mode. We would effectively be shipping a shim like [`browser_wasi_shim`](https://github.com/bjorn3/browser_wasi_shim) if we add this support

daxpedda:
> We would effectively be shipping a shim like [`browser_wasi_shim`](https://github.com/bjorn3/browser_wasi_shim) if we add this support

That's the idea.
But if this can be done without the help of `wasm-bindgen`, or make adjustments to enable that, even better.

temeddix:
`wasmer-js` claimed that it supports polyfills for system I/Os and threads polyfill(using webworkers) are also coming soon.

- https://github.com/wasmerio/wasmer-js/issues/332
- https://github.com/wasmerio/wasmer-js/pull/328

IMHO, maybe `wasmer-js` can be the standard WASI shim for `wasm-bindgen`

image

Rob2309:
I haven't really been paying attention on the wasm progress in rust. Is there still a realistic usecase for this PR?

daxpedda:
Now that https://github.com/rust-lang/rust/pull/117919 was merged, we could start thinking about the proposal I outlined in https://github.com/rustwasm/wasm-bindgen/pull/3324#issuecomment-1779753420.

PR #3325, opened 2023-02-19
Update Web IDLs for capturing from a HtmlCanvasElement

allsey87:
This PR adds a missing partial interface to the `HtmlCanvasElement` that includes the `captureStream` method. A related but missing IDL for `CanvasCaptureMediaStreamTrack` has also been added.

I have marked the non-standard `CanvasCaptureMediaStream` as deprecated, however this hasn't been marked as such in the generated Rust code. The binding for `CanvasCaptureMediaStream` has been left in place since it is still implemented in Firefox that way and developers who want to target Firefox can still make use of it as follows:
```rust
fn(canvas: HtmlCanvasElement) -> CanvasCaptureMediaStream  {
  canvas
    .capture_stream()
    .unwrap()
    .dyn_into::<CanvasCaptureMediaStream>()
    .unwrap()
}
```


crates/web-sys/webidls/enabled/CanvasCaptureMediaStream.webidl
```diff
@@ -10,7 +10,7 @@
  * W3C liability, trademark and document use rules apply.
  */
 
-[Pref="canvas.capturestream.enabled"]
+[Deprecated]
```
Liamolucko, line 13: If you want this to actually be marked deprecated on the Rust side, you can use `[RustDeprecated = "..."]`.

`[Deprecated]` is another Firefox holdover that the WebIDL generator ignores.

Liamolucko (approved):
Thanks! I just left one comment about how you could get `CanvasCaptureMediaStream` to actually be deprecated on the Rust side.

crates/web-sys/webidls/enabled/CanvasCaptureMediaStream.webidl
allsey87: @Liamolucko do you think it makes sense to mark a non-standard interface like this as deprecated on the Rust side? Is this consistent with other non-standard interfaces that we support?

allsey87:
@Liamolucko let me know if anything else needs to be done before this PR is merged :)

crates/web-sys/webidls/enabled/CanvasCaptureMediaStream.webidl
Liamolucko: I don't think there are any other non-standard interfaces in `web-sys` that we 'support'. There might some there by mistake, which wouldn't be deprecated (nothing is currently deprecated in `web-sys`), but that's not by any conscious choice so I wouldn't really take it as a precedent.

It feels a bit weird not to have any indication that this API is non-standard, but deprecating it would just be an inconvenience to anyone who actually needed to use it. So, I'm leaning towards not deprecating it.

allsey87: I agree. On one hand, I think it would be nice to indicate to developers using `web-sys` that they are using an interface that is not (or should not be) supported by browsers, but on the other hand, I can also see this as annoyance. In the long run, perhaps a general policy or approach could be adopted for such interfaces but that is certainly beyond the scope of this PR.

PR #3329, opened 2023-02-24
Don't force `TextDe/Encoder`

### LINKED ISSUE #2367: Unblock AudioWorklets: Find an alternative to TextEncoder / TextDecoder
[AudioWorklet's](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet) are a great fit for WebAssembly and therefor they're a great fit for Rust. `wasm-bindgen` is *almost* ready to be used in AudioWorklets, but not quite.

As discussed in a previous issue (https://github.com/rustwasm/wasm-pack/issues/689) the major blocker is that `TextEncoder` and `TextDecoder` are not available within AudioWorklets. `wasm-bindgen` would need to provide an alternative or find a work around.

----

In the above linked issue @CConnection made a prototype using a polyfill for TextEncoder / TextDecoder: https://github.com/anonyco/FastestSmallestTextEncoderDecoder

`wasm-bindgen` could potentially incorporate such a polyfill and then have some sort of "AudioWorklet" mode where it uses such a polyfill.

Or perhaps there are other ideas?

daxpedda:
Currently the generated JS shim calls `TextDe/Encoder`, even if it's not required by the user.
This forces users to provide a polyfill even when not needed.
Even if just a stub is provided as a polyfill, like done in the `wasm-audio-worklet` example, it would still require the class `TextDe/Encoder` to be defined, which prevents support detection by downstream users.

This adjust calls to `TextDe/Encoder` to do nothing if it's not available.
I don't think this has any use-case outside audio worklets.
This doesn't prevent users from using an actual polyfill if desired.

Cc @lukaslihotzki.
Fixes #2367.


Liamolucko (approved):
Thanks!

BlinkyStitt:
So I agree that TextDe/Encoder isn't needed outside of audio worklets, but your diff here removed the code from audio worklets where I do need it. Was that intentional? What is the best way to get this polyfilled inside of an Audio worklet?

https://github.com/rustwasm/wasm-bindgen/commit/78421dc53ccbe9dc4be437cd8b09961cb8159ddc#diff-07cf5528c3f9b4e22f2489de3616bd3ec5adbfeea002b3cad8d7f9d9cd5c68b0

daxpedda:
It was intentional indeed, as its not *required* anymore to get audio worklets to run.

Using the usual [`Worklet.addModule()`](https://developer.mozilla.org/en-US/docs/Web/API/Worklet/addModule) beforehand should perfectly suffice to inject any polyfill.

In general though I would recommend against using it in audio worklets, as any kind of memory allocation inside the audio worklet can cause stuttering among other issues, which is why the API was removed there to begin with. See https://github.com/WebAudio/web-audio-api/issues/2499#issuecomment-2002470074 for more information.

PR #3332, opened 2023-02-26
Rename some instances of WASM to Wasm

sgasse:
When I first contributed an example, I did not know that WebAssembly is abbreviated `Wasm` and not `WASM`. This commit fixes the mentionings in several places.

PR #3333, opened 2023-02-26
Update wasm-in-web-worker example

sgasse:
- Add note that Firefox will soon support ES modules.
- Fix a couple of `clippy` lints.
- Add link to compiled example.

Note: This is based on #3332 and should either wait for a potential merge of that PR or be rebased.


daxpedda:
FYI:
Firefox does not intend to ship module workers just yet, see https://bugzilla.mozilla.org/show_bug.cgi?id=1812591.
The [implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687) was done, but it was put behind a flag, see https://bugzilla.mozilla.org/show_bug.cgi?id=1812628.

Though it is active on nightly by default, it won't be in the beta or stable, as far as I understand it at least.

guide/src/examples/wasm-in-web-worker.md
```diff
@@ -1,11 +1,12 @@
-# WASM in Web Worker
+# Wasm in Web Worker
 
-[View full source code][code]
+[View full source code][code] or [view compiled example online][online]
 
 [code]: https://github.com/rustwasm/wasm-bindgen/tree/master/examples/wasm-in-web-worker
+[online]: https://sgasse.github.io/wasm_worker_interaction/wasm_worker_interaction/www/index.html
```
daxpedda, line 6: The link needs to be updated to point to Rustwasm instead.

guide/src/examples/wasm-in-web-worker.md
```diff
@@ -15,6 +16,15 @@ Firefox does not. To have compatibility across browsers, the whole example is
...
 with `--target no-modules`. The full command can be found in `build.sh`.
 
+However note that this may no longer be true when you read this. The
+[related bug][ff_worker_module_bug] has been closed. ES modules in web workers
+already work in Firefox 111.a nightly build. If you want to check if ES
+modules work as target for web workers in your current browser, check out if
+[this example][wasm_worker_with_modules] behaves as expected.
+
+[ff_worker_module_bug]: https://bugzilla.mozilla.org/show_bug.cgi?id=1247687
+[wasm_worker_with_modules]: https://sgasse.github.io/wasm_worker_interaction/wasm_module_js_worker/www/index.html
```
daxpedda, line 26: I don't think we need this information here. But we can update the note above with Firefox 114, which is the version expected to ship worker modules right now.

daxpedda (changes requested):
I'm happy to merge it like this.

We don't really have a style guide or formatter for our JS files, but this looks more consistent with the rest then before.

PR #3335, opened 2023-02-28
Use string replace in import instead of dynamic module creation

lukaslihotzki:
The last missing piece for a clean (static, CSP-compatible) solution for Wasm workers and worklets is the link back to the shim module in linked modules. This draft demonstrates how it can be hacked with a single string replace command.


daxpedda:
@lukaslihotzki I think this is an interesting demonstration, but I'm not sure if we want to add it to the example like this. Was your intention to actually merge this here?

Issue #1847, opened 2019-10-30
Support descriptions for JS/TS param/return comments


codehearts:
### Motivation

The generated JS and TS doc comments provide no descriptions for the parameters or return values. Some developers may want this information for additional context in editor tooltips

### Proposed Solution

I'm not sure what the ideal solution would be, but I could see a specially formatted line in the Rust comment being treated as the description for an items:

```rust
/// param variable_name: The description starts here
```

Or possible some sort of annotation

```rust
#[wasm_bindgen(param = variable_name, description = The description starts here)]
```

### Alternatives

The current alternative is to document the variables in the doc body, but editor tooltips will not show a description for the variables or return value in their tooltips and some developers may not like this

### Additional Context

I saw the discussion in #57 when generated comments were initially added, but there was no followup on adding this feature since Rust has no commenting convention for this

My motivation for opening this issue stems from Node.js developers at my company wanting more fully-fleshed out docs available to them when using our bound Node.js modules

alexcrichton:
Thanks for the report! I think this'd be a pretty reasonable feature to add!

codehearts:
I'd love to take a crack at this, is there a way to get feedback on which approach to take towards implementing this?

alexcrichton:
Sure yeah! It's probably good to comment here about a possible concrete design before implementing it, and that way anyone interested can likely help chime in and give feedback.

One question I'd have, we should copy documentation from the Rust source to the JS source, so I think this could be "solved" by writing the Rust documentation in the way that JS expects, right? Is that something you'd prefer to not do though?

codehearts:
That's a thought, maybe if a JS-doc formatted line exists for a variable it can be copied verbatim into the generated output. Otherwise, it generates the documentation as it currently does

Maybe leaving the type field blank causes bindgen to fill it out for you, I do really like how that ensures your docs match your Rust types

Rust | JS
---- | --
`///` | `* @param my_var {string}`
`/// @param my_var {} My string var` | `* @param my_var {string} My string var`
`/// @param my_var {string} My string var` | `* @param my_var {string} My string var`

This lines up with what I do currently for documenting throws, I end my Rust comment with `/// @throws {TypeError} When a non-string is passed` so the generated JS comment looks right

alexcrichton:
[...]

PR #3338, opened 2023-03-02
Add a skip_jsdoc attribute.

jneem:
This adds a `skip_jsdoc` attribute to suppress the generation of jsdoc `@param` and `@returns` comments.

This was motivated by #1847, but I ended up going a different route because I found it awkward to put long descriptions in attributes. The idea is that if you can skip the auto-generated jsdoc, you can comment the parameters and return values yourself:

```rust
/// @param {string} world - The world.
#[wasm_bindgen(skip_jsdoc)]
fn hello(world: &str) {}
```

It's a little sad that this approach doesn't keep the type information in sync, but I still think it's nicer than an alternative like

```rust
#[wasm_bindgen(param(world = "The world."))]
fn hello(world: &str) {}
```
 (especially if the description is longer)


jneem:
I'm not sure I did the schema-bumping correctly, but I tried to do what's in line with past practice: I bumped the schema version but not the crate versions, which seems to usually happen right when releasing.

Liamolucko (commented):
Thanks for this! The code looks good, but could you add a page about this to the guide?

The source Markdown is at `guide/src`, and this should specifically go in `guide/src/reference/attributes/on-rust-exports`.

You also have to add an entry to `guide/src/SUMMARY.md`, otherwise the page won't show up in the table of contents.

Also, yes, you bumped the schema version properly.

jneem:
Thanks for the pointers! I've added a guide entry.

Liamolucko (approved):
Thanks!

qtnx:
Looks like it's not working on the latest version 0.2.84
```
error: unknown attribute
   --> wasm/src/lib.rs:198:20
    |
198 |     #[wasm_bindgen(skip_jsdoc)]
    |                    ^^^^^^^^^^
```

jneem:
Yes: 0.2.84 was release in February and this was only merged in March.

qtnx:
> Yes: 0.2.84 was release in February and this was only merged in March.

Hmm, so it shouldn't be in the official documentation. Hopefully, a new version will be released soon, it's already April.

PR #3341, opened 2023-03-07
Use `convertToBlob` for OffscreenCanvas

namse:
OffscreenCavnas doesn't have method `toBlob` unlike Canvas. Instead of that, OffscreenCanvas has `convertToBlob` method. 
https://html.spec.whatwg.org/multipage/canvas.html#dom-offscreencanvas-converttoblob


daxpedda:
You need to re-generate the interface.
See https://rustwasm.github.io/docs/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html.

daxpedda:
I don't know how the maintainers see it, but this is a breaking change, I wouldn't consider it as such though, because calling this function before would fail anyway.

But while we are at it @namse, could we also add [`ImageEncodeOptions`](https://html.spec.whatwg.org/multipage/canvas.html#imageencodeoptions) to the WebIDL?

namse:
I didn't `ImageEncodeOptions` because the previous code didn't added `ImageEncodeOptions`.
Give me little time to add it, I think I can add it.

namse:
Wow, it's super easy to put new dictionary type, just paste webidl...!! Good experience to contribute. Thank you for suggestion.
I checked there are no name `ImageEncodeOptions` in other webidl so I think there is no conflict.

daxpedda:
The `web-sys` `[features]` sections has to be updated though, the generator should generate a `features` file somewhere that you can copy.

(we should probably document this in https://rustwasm.github.io/docs/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html)

namse:
Should I `ImageEncodeOptions = ["OffscreenCanvas"]` or `ImageEncodeOptions = []`? I think the second is right because it's not hierarchy relationship between `ImageEncodeOptions` and `OffscreenCanvas`.

daxpedda:
You should probably do whatever the generator spits out, otherwise the next person will overwrite it anyway.
I honestly don't know what the generator does in this case.

namse:
Generator didn't touch Cargo.toml. did I something wrong?

daxpedda:
Unfortunately the generator doesn't do it for you, it just generates a file called `features` (no file extension) in `crates/web-sys`, you should copy it's contents into the `Cargo.toml` `[features]` section of `web-sys`.

This file is gitignored, so you won't see it in your diff.

namse:
Found feature file and overwrite, no git diff. Cool! I was right :) Thank you for explain. Maybe we need a command to put features automatically to Cargo.toml.

daxpedda:
Funny, yeah, this part should probably be added to the wasm-bindgen book.

daxpedda (approved):
See https://github.com/rustwasm/wasm-bindgen/pull/3341#issuecomment-1458045890 for my thoughts on the breaking change.

Worst case scenario we can keep the old broken API and just deprecate it.

daxpedda:
No, I'm not a maintainer, just a contributor, we have to wait for a maintainer to approve and merge.

On that note, I'm seeing the CI will fail, it got stuck on something unrelated to this PR, probably best to restart it. The only way you can do it as a non-maintainer is just to close and re-open the PR, or force-push.

codeart1st:
I'm also lookig into this get fixed.

namse:
Any maintainer who can review this?

alexcrichton:
Sorry for the delay, but thanks!

PR #3344, opened 2023-03-09
Add WebTransport webidl to unstable

darioalessandro:
Added WebTransport webidl definitions to the unstable features.


darioalessandro:
@Liamolucko let me know the next steps 😄

Liamolucko (approved):
Looks good! I just left one comment about the URL of the WebIDL file being wrong.

I would have asked you to add `[Throws]` to any functions that can throw, which tells the code generator the function should return a `Result`, but it turns out that nothing in WebTransport can actually throw anyway, since everything's async and results in a promise rejection or similar instead.

PR #3347, opened 2023-03-10
fix audioworklet example local server error

chaosprint:


Liamolucko:
What error are you talking about? Do you mean the error your browser will give you trying to navigate to `http://[::]:8000`? If so, I'm in favour of fixing that, you just haven't made yourself very clear.

Also, the `raytrace-parallel` example has an identical Python server, so you should do the same for that one too.

Issue #3340, opened 2023-03-07
Run tests on web workers


whizsid:
### Motivation

In some code segments I have to depend on Web Workers APIs which not available in browser scope. So I need an option to run all the tests on a web worker.

### Proposed Solution

Run all the tests on a web worker and sending console outputs by using `postMessage` to main thread.

### Alternatives

Run each test on a separate web worker. This may break the existing behaviours for other targets. Example:- NodeJS and browser tests have shared memory access. But when using separate workers, we can not use a shared memory.

### Additional Context

I took some steps to implement it. But I got an error from `getObject` function inside the worker.

daxpedda:
Would you mind elaborating on what Web Worker APIs you are depending and why?

whizsid:
@daxpedda I wrote my crate to use `Window` on browser scope and `WorkerGlobalScope` on worker scopes. So I need to ensure whether my crate is working for all the scopes.

PR #3348, opened 2023-03-11
Run tests on worker

whizsid:
Fixing #3340 .


crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
```diff
@@ -45,36 +46,129 @@ pub fn spawn(
...
+                            method == "debug"
+                        ) {{
+                            console[method].apply(undefined, args[0]);
+                        }} else if (method == "start") {{
+                            document.getElementById("output").textContent = "";
+                        }} else if (method == "output") {{
+                            document.getElementById("output").textContent += args[0] + "\n";
+                        }}
+                    }}
+                }});
```
Liamolucko, line 121: This is more complicated than it needs to be. You don't really need to have a wasm module in both the main thread and the worker.

Instead, you should be able to have only the wasm module in the worker, have it keep track of what the contents of the output `<pre>` should be, and then post a message back to the main thread with the updated contents every time it changes.

Liamolucko (commented):
This looks like a good feature!

There are two reasons it's not working:

* You got the check for whether the worker should be an ES module or not backwards, and so errors are happening because you're trying to run an ES module as a script and vice versa.
* You're always using `importScripts` to import the JS shim, even when it's a module and that doesn't work.

I left some more specific comments about those and a couple other things.

Also, could you run `cargo fmt`?

whizsid:
I accidentally pulled from the wrong remote and polluted the current PR. :-( I will re-open a clean PR.

PR #3349, opened 2023-03-12
Fixes #3259: use verbose names for load / init / getImports

### LINKED ISSUE #3259: Fails with ` SyntaxError: Identifier 'init' has already been declared`
### Describe the Bug

Not sure if this is a wasm-bindgen issue o from rollup-plugin-rust, trying to include a function named  `init` from an external module named fails with:

```rust
#[wasm_bindgen(module = "@amplitude/analytics-browser")]
extern "C" {
    #[wasm_bindgen]
    pub async fn init(api_key: &str);
}
```

` SyntaxError: Identifier 'init' has already been declared (1376:15)`

The conflict seems to be with `https://github.com/rustwasm/wasm-bindgen/blob/979b335e4ef1fdd59bfb520d3868a39e61cd703d/crates/cli-support/src/js/mod.rs#L870`

### Steps to Reproduce

add an import to a function named `init`
### Expected Behavior

It should import the `init` function
### Actual Behavior

Fails with ` SyntaxError: Identifier 'init' has already been declared (1376:15)`

### Additional Context

Maybe there is an option to do that I haven't seen in the documentation?

``` import { init as init_wrapper } ...```

### PR DESCRIPTION
**Pull request by threepointone**
`init` / `load` / `getImports` are common names for functions, and clash when generated with `--target web`. This adds verbose prefixes to them so there's a smaller chance of clashing with already defined functions.

Fixes https://github.com/rustwasm/wasm-bindgen/issues/3259.

--- 

Hi! This is my first pull request here, and I don't really know how to write rust, but I think I got this right. Happy to iterate if you'd like.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -388,7 +388,7 @@ impl<'a> Context<'a> {
                 js.push_str("let wasm = undefined;\n");
                 init = self.gen_init(needs_manual_start, None)?;
                 footer.push_str(&format!(
-                    "{} = Object.assign(init, {{ initSync }}, __exports);\n",
+                    "{} = Object.assign(__wbg_init, {{ initSync }}, __exports);\n",
                     global
                 ));
             }
@@ -484,7 +484,7 @@ impl<'a> Context<'a> {
                 self.imports_post.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, Some(&mut imports))?;
                 footer.push_str("export { initSync }\n");
-                footer.push_str("export default init;");
+                footer.push_str("export default __wbg_init;");
             }
         }
 
@@ -650,7 +650,7 @@ impl<'a> Context<'a> {
                 memory_param = memory_param
             ));
 
-            setup_function_declaration = "export default function init";
+            setup_function_declaration = "export default function __wbg_init";
         }
         Ok(format!(
             "\n\
@@ -798,7 +798,7 @@ impl<'a> Context<'a> {
 
         let js = format!(
             "\
-                async function load(module, imports) {{
+                async function __wbg_load(module, imports) {{
                     if (typeof Response === 'function' && module instanceof Response) {{
                         if (typeof WebAssembly.instantiateStreaming === 'function') {{
                             try {{
@@ -833,19 +833,19 @@ impl<'a> Context<'a> {
                     }}
                 }}
 
-                function getImports() {{
+                function __wbg_get_imports() {{
                     const imports = {{}};
                     {imports_init}
                     return imports;
                 }}
 
-                function initMemory(imports, maybe_memory) {{
+                function __wbg_init_memory(imports, maybe_memory) {{
                     {init_memory}
                 }}
 
-                function finalizeInit(instance, module) {{
+                function __wbg_finalize_init(instance, module) {{
                     wasm = instance.exports;
-                    init.__wbindgen_wasm_module = module;
+                    __wbg_init.__wbindgen_wasm_module = module;
                     {init_memviews}
                     {start}
                     return wasm;
@@ -854,9 +854,9 @@ impl<'a> Context<'a> {
                 function initSync(module{init_memory_arg}) {{
                     if (wasm !== undefined) return wasm;
 
-                    const imports = getImports();
+                    const imports = __wbg_get_imports();
 
-                    initMemory(imports{init_memory_arg});
+                    __wbg_init_memory(imports{init_memory_arg});
 
                     if (!(module instanceof WebAssembly.Module)) {{
                         module = new WebAssembly.Module(module);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by daxpedda** [comment]
I believe you meant `--target no-modules`?
This will probably not be merged as it is a breaking change.

**Comment by threepointone** [comment]
The problem exists with `--target web`, discovered in this PR (https://github.com/automerge/automerge/pull/554)
The named export is initSync which I haven't renamed, so it shouldn't be a breaking change?

**Comment by daxpedda** [comment]
Ah, apologies, I didn't look properly into your changes!
But you did also modify `init` in the no-modules target, which is a breaking change.

May I ask though what the exact problem is? How does an exported name clash? Can't you name it whatever you like during export?
Apologies for my naivety, my web front end experience is pretty low.

**Comment by threepointone** [comment]
It shouldn't be a breaking change in no-modules either, since the name `init` isn't used directly (The `Object.assign` call adds properties directly on to the function init which is then assigned to the `wasm_bindgen`. 
(I think so, unless I'm missing some more context)

The original problem: When we're also exporting a function called `load` (or `init`, or `getImports`), it gets declared as a function in the generated js file. wasm-bindgen also generates a function with the same name, which clashes with it. This leads to issues like https://github.com/rustwasm/wasm-bindgen/issues/3259. The PR should fix it.

**Comment by daxpedda** [comment]
> It shouldn't be a breaking change in no-modules either, since the name `init` isn't used directly (The `Object.assign` call adds properties directly on to the function init which is then assigned to the `wasm_bindgen`.

Right, completely forgot how it works!

> The original problem: When we're also exporting a function called `load` (or `init`, or `getImports`), it gets declared as a function in the generated js file. wasm-bindgen also generates a function with the same name, which clashes with it. This leads to issues like #3259. The PR should fix it.

Interesting, I just tried it, it specifically happens when you use module imports because it just inserts `import { init } from 'wherever'` in the shim. Ideally we should rename on import I think, but it doesn't really matter, this PR is an "improvement" nonetheless.

Could you add "Fixes #3259." in the OP?

**Comment by threepointone** [comment]
Done! Thanks for the weekend look!

File: crates/cli-support/src/js/mod.rs [hunk 1]
```diff
@@ -388,7 +388,7 @@ impl<'a> Context<'a> {
                 js.push_str("let wasm = undefined;\n");
                 init = self.gen_init(needs_manual_start, None)?;
                 footer.push_str(&format!(
-                    "{} = Object.assign(init, {{ initSync }}, __exports);\n",
+                    "{} = Object.assign(wasm_bindgen_init_input, {{ initSync }}, __exports);\n",
```
Liamolucko, line 391: I'm not sure what `init_input` is supposed to mean. Could it just be `wasm_bindgen_init` instead?

Also, all the prefixed names like this that we already have in `wasm-bindgen` are prefixed with `__wbg_`, not `wasm_bindgen_`, so it might be better to call it `__wbg_init` for consistency.

crates/cli-support/src/js/mod.rs
```diff
@@ -833,19 +833,19 @@ impl<'a> Context<'a> {
                     }}
                 }}
 
-                function getImports() {{
+                function wasm_bindgen_getImports() {{
```
Liamolucko, line 836: Could this be called `wasm_bindgen_get_imports` (or, in line with the above comment, `__wbg_get_imports`) instead? Mixed casing conventions like this look weird.

Same for `wasm_bindgen_initMemory`.

crates/cli-support/src/js/mod.rs
```diff
@@ -798,7 +798,7 @@ impl<'a> Context<'a> {
 
         let js = format!(
             "\
-                async function load(module, imports) {{
+                async function wasm_bindgen_load_module_imports(module, imports) {{
```
Liamolucko, line 801: Again, why is this called `wasm_bindgen_load_module_imports` instead of just `wasm_bindgen_load` (or `__wbg_load`)? I think I see what you're doing now, you're adding the argument names onto the end of the function name, but I don't understand why. It just makes it harder to read in my opinion.

crates/cli-support/src/js/mod.rs
threepointone: agreed, done!

crates/cli-support/src/js/mod.rs
threepointone: agreed, done!

crates/cli-support/src/js/mod.rs
threepointone: agreed, done!

Liamolucko (approved):
Thanks!

shreevatsa:
Should this also have updated the example at https://github.com/rustwasm/wasm-bindgen/blob/a210654ac3f89b9ef11f1d3eb0ee2c9eaa814021/examples/without-a-bundler/index.html#L15 (and what is seen at https://rustwasm.github.io/wasm-bindgen/examples/without-a-bundler.html, however that is generated)?

daxpedda:
The relevant function names were only changed internally, their exported names remained the same.

shreevatsa:
I think the exported name changed too. At least, something broke in some code of mine between a few months ago (I think `wasm-bindgen` 0.2.79) and now, that I had to fix by [changing](https://github.com/shreevatsa/pdf-explorer/commit/464622a5bbe916d73186c0cb0fa02ab5bf17c240) the `init` to `__wbg_init`. But maybe I'm doing something incorrectly.

threepointone:
It’s a default export, so it doesn’t have a name (behind what you’d call it when you import it). 

From your link, it looks like you were manually rewriting the code to export the function (with its name) instead of importing the default. Could you explain why you had to do that instead of importing the default?

shreevatsa:
Ah I see, that's why it's `import init, { add } from`… rather than `import { init, add } from`…

With the correct form, it indeed works continues to work fine now; sorry for the noise then! I guess what happened is that I had at some point copied the example incorrectly and written something like `import {init`… and then when it didn't work I just hacked the code to export the `init`. I did suspect I was doing something wrong and things are clear now; thanks for your patience :-)

threepointone:
Glad you’re unblocked!

PR #3350, opened 2023-03-13
Add support for `performance.measureUserAgentSpecificMemory`

### LINKED ISSUE #3159: performance.measureUserAgentSpecificMemory
### Motivation

[`window.performance.memory`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory) in an API for accessing the memory use of JavaScript (and by extension, wasm).

Example JS: `window.performance.memory.usedJSHeapSize`

### Proposed Solution

Add a `Memory` feature flag to expose `window.performance.memory` with three fields:

* `js_heap_size_limit`
* `total_js_heap_size`
* `used_js_heap_size`

### Alternatives

WASM apps can install a custom `#[global_allocator]` that does its own bookkeeping of amount of RAM used.

### Additional Context

Knowing the memory used by a WASM application is useful for it to know whether or not to try free up some memory.

Liamolucko:
Fixes #3159

This is a Chrome-only API for measuring the memory usage of JS/wasm. So, I've marked it as unstable.


alexcrichton:
Thanks!

PR #3351, opened 2023-03-13
Worker tests

### LINKED ISSUE #3340: Run tests on web workers

whizsid:
Fixed #3340. Origin #3348


Liamolucko (commented):
Looks good! Could you put some examples of tests that run in a worker into `tests/worker`? That way, they'll get run in CI and make sure this feature stays tested.

whizsid:
@Liamolucko  In last month when I trying to implement tests, firefox did not supported ECMAScript module workers. So I could not run the tests without `WASM_BINDGEN_USE_NO_MODULE` in firefox. Github actions also configured to run on firefox. So I gave up.
.
But I saw Firefox is now supporting ECMAScript module workers behind a special preference. However I had to remove the `webdriver.json` from the `.gitignore` list.

crates/test/src/rt/worker.rs
```diff
@@ -0,0 +1,66 @@
...
+        self.writeln(&format!("test {} ... {}", name, s));
+    }
+
+    fn stringify_error(&self, err: &JsValue) -> String {
+        // TODO: this should be a checked cast to `Error`
+        let err = Error::from(err.clone());
+        let name = String::from(err.name());
+        let message = String::from(err.message());
+        let err = WorkerError::from(JsValue::from(err));
+        let stack = err.stack();
```
hamza1311, line 48: Nit: you can avoid creating a JsValue when you already have one
```suggestion
        let error = Error::from(err.clone());
        let name = String::from(error.name());
        let message = String::from(error.message());
        let err = WorkerError::from(err);
        let stack = err.stack();
```

tests/worker/memory.rs
```diff
@@ -0,0 +1,15 @@
+use std::sync::atomic::{AtomicU16, Ordering};
+use wasm_bindgen_test::{wasm_bindgen_test, wasm_bindgen_test_configure};
+
+static COUNTER: AtomicU16 = AtomicU16::new(12);
+
+#[wasm_bindgen_test]
+pub fn test_atomic_var_after_changed() {
+    assert_eq!(COUNTER.load(Ordering::SeqCst), 15);
+}
```
hamza1311, line 9: How do we guarantee that this is always executed after the other test?

hamza1311 (commented):
Implementation looks good to me. Can you add documentation for it? It can go in [wasm-bindgen-test section in the guide](https://github.com/rustwasm/wasm-bindgen/tree/main/guide/src/wasm-bindgen-test).

Thanks for taking the time to make this PR! Really appreciate it

crates/test/src/rt/worker.rs
whizsid: `err` is behind a shared reference. `WorkerError` can not constructed from `&JsValue`. So we have to clone it twice.

tests/worker/memory.rs
whizsid: Yes I also thought it. The unit tests should be isolated from other. But I had to test the atomics in workers. Because someone can implement the same functionalities using separate workers for each test. Then the atomics will not work like in `nodejs` and `browser`. However I removed it due to bad use of unit tests.

PR #3352, opened 2023-03-14
Remove outdated browser support caveats

lukaslihotzki:
- Although usage dropped to [0.49%](https://caniuse.com/usage-table), IE 11 is supported until [January 13, 2032](https://en.wikipedia.org/wiki/Internet_Explorer#End_of_life) on obscure Windows versions, so leave the caveat in there.
- Edge Legacy is end of life since March 9, 2021.
- The Wasm spec now allows passing 64-bit integers across the ABI (https://webassembly.github.io/spec/js-api/index.html). This is supported in all relevant browsers (https://webassembly.org/roadmap/) and used by wasm-bindgen (https://github.com/rustwasm/wasm-bindgen/pull/3049).

### DIFF
--- a/guide/src/reference/browser-support.md
+++ b/guide/src/reference/browser-support.md
@@ -2,9 +2,9 @@
 
 The output of `wasm-bindgen` includes a JS file, and as a result it's good to
 know what browsers that file is expected to be used in! By default the output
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #3360, opened 2023-03-17
Add attribute for custom path to `wasm_bindgen`

daxpedda:
This adds two new attributes: `wasm_bindgen` and `wasm_bindgen_futures`, which allow to specify the path to both crates if they were renamed or re-exported from a different crate.

This is analogous to [Tokio](https://docs.rs/tokio/1.26.0/tokio/attr.main.html#rename-package), [Serde](https://serde.rs/container-attrs.html#crate) or [Pollster](https://docs.rs/pollster/0.3.0/pollster/#macro).

### DIFF
--- a/crates/backend/src/ast.rs
+++ b/crates/backend/src/ast.rs
@@ -5,12 +5,13 @@
 use crate::{util::ShortHash, Diagnostic};
 use proc_macro2::{Ident, Span};
 use std::hash::{Hash, Hasher};
+use syn::Path;
 use wasm_bindgen_shared as shared;
 
 /// An abstract syntax tree representing a rust program. Contains
 /// extra information for joining up this rust code with javascript.
 #[cfg_attr(feature = "extra-traits", derive(Debug))]
-#[derive(Default, Clone)]
+#[derive(Clone)]
 pub struct Program {
     /// rust -> js interfaces
     pub exports: Vec<Export>,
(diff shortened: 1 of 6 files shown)


crates/backend/src/ast.rs
```diff
@@ -26,6 +27,26 @@ pub struct Program {
...
+        Self {
+            exports: Default::default(),
+            imports: Default::default(),
+            linked_modules: Default::default(),
+            enums: Default::default(),
+            structs: Default::default(),
+            typescript_custom_sections: Default::default(),
+            inline_js: Default::default(),
+            wasm_bindgen: syn::parse_quote! { wasm_bindgen },
+            wasm_bindgen_futures: syn::parse_quote! { wasm_bindgen_futures },
```
daxpedda, line 47: This should have been `::wasm_bindgen` and `::wasm_bindgen_futures` for added proc-macro hygiene. Unfortunately that would be a breaking change.

crates/macro-support/src/lib.rs
```diff
@@ -105,13 +105,60 @@ pub fn expand_class_marker(
...
+                break;
+            }
+        }
+
+        Ok(ClassMarker {
+            class,
+            js_class,
+            wasm_bindgen: wasm_bindgen.unwrap_or_else(|| syn::parse_quote! { wasm_bindgen }),
+            wasm_bindgen_futures: wasm_bindgen_futures
+                .unwrap_or_else(|| syn::parse_quote! { wasm_bindgen_futures }),
```
daxpedda, line 161: Same as above.

daxpedda:
The latest `wasm-pack` v0.11 release broke our CI somehow.

crates/backend/src/ast.rs
hamza1311: Mentioning it here for reference: we can make that change alongside https://github.com/rustwasm/wasm-bindgen/pull/3372.

crates/macro/ui-tests/wasm-bindgen.rs
```diff
@@ -0,0 +1,43 @@
+extern crate wasm_bindgen as extern_test;
```
hamza1311, line 1: Ideally we should have a separate crate which renames wasm-bindgen in Cargo.toml but this should also suffice

hamza1311 (approved):
Looks good to me :+1:

PR #3361, opened 2023-03-18
Update `syn` to v2

daxpedda:
This updates `syn` to v2.0.

There should be any functional changes, just hopefully better error messages.


Liamolucko (approved):
Thanks!

PR #3363, opened 2023-03-20
web-sys: exposing selection direction in `setRangeText()`

xnuk:
Related: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setRangeText


xnuk:
`build_examples` test error:
```
[INFO]: found wasm-opt at "/home/runner/work/wasm-bindgen/wasm-bindgen/binaryen-version_109/bin/wasm-opt"
Error: /home/runner/work/wasm-bindgen/wasm-bindgen/binaryen-version_109/bin/bin/wasm-opt binary does not exist
```

Seems this is not my problem I guess?

Liamolucko:
You changed `setRangeText`, not `setSelectionRange`. I'm not sure which one you were actually intending to fix.

`setSelectionRange` is already correct. If you were meaning to fix `setRangeText`, the proper way to do that is to define the enum it uses, not just replace it with `DOMString`:

```webidl
enum SelectionMode {
  "select",
  "start",
  "end",
  "preserve" // default
};
```

xnuk:
> You changed `setRangeText`, not `setSelectionRange`. I'm not sure which one you were actually intending to fix.

Yes it is `setRangeText`, my mistake. I'll fix the commit message later.

> the proper way to do that is to define the enum it uses, not just replace it with `DOMString`

`setSelectionRange` already uses `DOMString` for this (`"forward" | "backward" | "none"`):
```
undefined setSelectionRange(unsigned long start, unsigned long end, optional DOMString direction);
```
So If I change `setRangeText` to use enum, then `setSelectionRange` also should. Because I'm unfamiliar with the webidl language, it's ok somebody else do this.

Liamolucko:
The reason that `setRangeText` should use an enum for `selectionMode` is so that we're in sync with the [upstream WebIDL](https://html.spec.whatwg.org/multipage/input.html#htmlinputelement), which has it as an enum:

```webidl
  undefined setRangeText(DOMString replacement);
  undefined setRangeText(DOMString replacement, unsigned long start, unsigned long end, optional SelectionMode selectionMode = "preserve");
  undefined setSelectionRange(unsigned long start, unsigned long end, optional DOMString direction);
```

I think the reason that the upstream WebIDL has `setRangeText` using an enum and `setSelectionRange` not is because `setRangeText` actually enforces that you only provide one of those values, whereas `setSelectionRange` will still work if you give it something else.

For instance, `setSelectionRange(1, 2, "foobar")` will behave the same as `setSelectionRange(1, 2, "none")`, but `setRangeText("hello", 1, 2, "foobar")` will throw an error.

Issue #3356, opened 2023-03-16
future incompat dependencies


Nanai10a:
I did `cargo install wasm-bindgen-cli`, and seen:

```
warning: the following packages contain code that will be rejected by a future version of Rust: buf_redux v0.8.4, multipart v0.18.0, wasmparser v0.59.0, wasmparser v0.77.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

with using:

```
rustup 1.25.2 (2023-02-04)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.70.0-nightly (171693274 2023-03-14)`
```

here is a dependency tree:

- wasm-bindgen-cli v0.2.84
  - rouille v3.6.1
    - **multipart v0.18.0**
      - **buf_redux v0.8.4**
  - walrus v0.19.0
    - **wasmparser v0.77.0**
  - wasm-bindgen-cli-support v0.2.84
    - wit-validator v0.2.1
      - **wasmparser v0.59.0**

and report:

```
To solve this problem, you can try the following approaches:


- Some affected dependencies have newer versions available.
You may want to consider updating them to a newer version to see if the issue has been fixed.

wasmparser v0.59.0 has the following newer versions available: 0.60.0, 0.60.1, 0.60.2, 0.61.0, 0.62.0, 0.63.0, 0.63.1, 0.64.0, 0.65.0, 0.66.0, 0.67.0, 0.68.0, 0.69.0, 0.69.1, 0.69.2, 0.70.0, 0.71.0, 0.72.0, 0.73.0, 0.73.1, 0.74.0, 0.75.0, 0.76.0, 0.77.0, 0.78.0, 0.78.1, 0.78.2, 0.79.0, 0.80.0, 0.80.1, 0.80.2, 0.81.0, 0.82.0, 0.83.0, 0.84.0, 0.85.0, 0.86.0, 0.87.0, 0.88.0, 0.89.0, 0.89.1, 0.90.0, 0.91.0, 0.92.0, 0.93.0, 0.94.0, 0.95.0, 0.96.0, 0.97.0, 0.98.1, 0.99.0, 0.100.0, 0.101.0, 0.101.1, 0.102.0
[...]

PR #3372, opened 2023-03-27
Remove experimental interface types support

Liamolucko:
This is work-in-progress, I'm just opening this as a draft to make sure CI is passing.

This PR removes `wasm-bindgen`'s experimental support for interface types. This is because it relies heavily on [the archived `wit-*` family of crates](https://github.com/bytecodealliance/wasm-interface-types), which are now triggering future incompatibility warnings, as reported in #3356.

The support is far out of date with the upstream component model, so it wouldn't have been particularly useful in its current state anyway.


Liamolucko:
While working on this I noticed that wasm-bindgen throws an error if you try to use the multi-value transform without interface types enabled. I couldn't see any reason for that, so I've gotten rid of that check.

hamza1311:
Now that this is merged, I assume the next release will be semver breaking (0.3.0)? We should create a tracking issue/check list for that

alexcrichton:
I don't think this is a breaking change, so this should be good to keep up with the 0.2 track.

PR #3374, opened 2023-03-28
Add WebIDLs for Local Font Access

fenjalien:
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Local_Font_Access_API), [IDL](https://wicg.github.io/local-font-access/)

I'm not sure if I've done this correctly as running the `webidl` command doesn't output any features so I've had to add some manually.


crates/web-sys/webidls/unstable/LocalFontAccess.webidl
```diff
@@ -0,0 +1,20 @@
+[SecureContext]
+partial interface Window {
+  [Throws]
+  Promise<sequence<FontData>> queryLocalFonts(optional QueryOptions options = {});
```
daxpedda, line 4: (just self-documenting)
Throwing behavior is documented in the specification.
Implementation: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/font_access/window_font_access.idl.

PR #3375, opened 2023-03-30
Adds RegistrationOptions.type_

### LINKED ISSUE #3342: registration option "type" is missing for ServiceWorker registration
### Motivation

Hello, I am trying to register a service worker entirely in rust and it seems that I need to import as a module.
To manage this type of import, there is a parameter "type" that can be passed in the registration option.
And it seems that it is not present in web_sys registration options.
Note : https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register

### Proposed Solution

To add the "type" param in  [RegistrationOptions](https://docs.rs/web-sys/latest/web_sys/struct.RegistrationOptions.html) and manage it in the function [register_with_options](https://docs.rs/web-sys/latest/web_sys/struct.ServiceWorkerContainer.html#method.register_with_options).

### Alternatives

I am a real newbie in rust and wasm to think another option :)

mtsr:
Fixes #3342

### DIFF
--- a/crates/web-sys/src/features/gen_RegistrationOptions.rs
+++ b/crates/web-sys/src/features/gen_RegistrationOptions.rs
@@ -32,6 +32,19 @@ impl RegistrationOptions {
         let _ = r;
         self
     }
+    #[doc = "Change the `type` field of this object."]
+    #[doc = ""]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks and sorry for the delay!

PR #3386, opened 2023-04-13
Update ParentNode

namse:
Add three funcitons(replaceChildren, querySelector, querySelectorAll) for ParentNode


daxpedda:
I think the problem here is that the definitions you added apply to `Document` and others, that already have those definitions. So you have to remove them there to avoid these duplicates.

namse:
Actually my only interest is `replaceChildren`. Is it Ok to update `replaceChildren` only?

daxpedda:
I'm not a maintainer here, but seems fine to me.

PR #3395, opened 2023-04-22
Mark inform function as unsafe

kitcatier:
https://github.com/rustwasm/wasm-bindgen/blob/153a6aa9c7a989c1865df7f93b2ddbca1113a175/src/describe.rs#L52-L54
Hello, in the above test case, the function has a 'function not implemented on non-wasm32 targets'.
```rust
let _ = inform(10);
``` 
`thread 'main' panicked at 'function not implemented on non-wasm32 targets'`
This function should be marked as unsafe, if a function's entire body is unsafe, the function is itself unsafe and should be marked appropriately

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -180,8 +180,8 @@ impl ToTokens for ast::Struct {
                         }
                     }
                     use #wasm_bindgen::describe::*;
-                    inform(RUST_STRUCT);
-                    inform(#name_len);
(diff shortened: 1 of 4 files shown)


daxpedda:
This isn't what `unsafe` is intended for in Rust. Panicking isn't considered `unsafe`.
Most of the crate functions will panic on non-Wasm targets.

See https://doc.rust-lang.org/reference/unsafety.html.

Issue #3398, opened 2023-04-24
Generated bindings with no-modules target causes error in web extensions


holly-hacker:
### Describe the Bug

I am trying to build a web extension using [Mubelotix/wasm-extension-template](https://github.com/Mubelotix/wasm-extension-template) as a template, but I get a javascript error inside the generated bindings when loading the generated extension in a browser:
![Uncaught TypeError: Cannot read properties of null (reading 'src')](https://user-images.githubusercontent.com/13605369/233978151-982ecd17-e17e-432e-a777-17bab3fbf9e0.png)

The linked template uses wasm-pack with the cli argument `--target=no-modules` which generates binding code that starts like so:
```js
let wasm_bindgen;
(function() {
    const __exports = {};
    let script_src;
    if (typeof document === 'undefined') {
        script_src = location.href;
    } else {
        script_src = new URL(document.currentScript.src, location.href).toString();
    }
    let wasm;
    // ...
```

In the context of a web extension, `document` is defined but `document.currentScript` is null. This causes a TypeError when `document.currentScript.src` is executed.

The current `main` branch (commit 153a6aa9c7a989c1865df7f93b2ddbca1113a175) seems to output different code, as seen [here](https://github.com/rustwasm/wasm-bindgen/blob/153a6aa9c7a989c1865df7f93b2ddbca1113a175/crates/cli-support/src/js/mod.rs#L385):
```js
if (typeof document !== 'undefined') {
    script_src = new URL(document.currentScript.src, location.href).toString();
}
```

I haven't managed to build the `main` branch of wasm-pack yet to try this out, but I assume it will produce the same error.

### Steps to Reproduce

1. Clone [Mubelotix/wasm-extension-template](https://github.com/Mubelotix/wasm-extension-template)
2. Build the extension with `sh build.sh`
3. Inside a chromium-based browser, load the extension as an unpacked extension by following [Chrome's documentation](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#load-unpacked).
4. Browse to https://example.com, open devtools and observe the error.

### Expected Behavior

Wasm-bindgen should account for cases where `document.currentScript` is `null` or `undefined` and have a fallback.

### Actual Behavior

Wasm-bindgen assumes `document.currentScript` is not null when `document` is not undefined.

### Additional Context

- wasm-pack version: 0.11.0
- wasm-bindgen version: 0.2.83 (shipped with wasm-pack, I presume)

[...]

PR #3402, opened 2023-04-25
Fixed document.currentScript is null #3398

rimutaka:
This is a fix for https://github.com/rustwasm/wasm-bindgen/issues/3398 where `document.currentScript` can be null in a browser extension. It is a rough solution, probably not good enough to be merged, but I hope it saves time for maintainers coming up with a proper fix. 

### Changes: 

Check if document.currentScript or document.currentScript.src is null and replace it with an empty string so that the URL is set from `location.href`.

I do not know under what conditions `.src` property can be null and if it's right to replace it with an empty string. Maybe it should be a hard fail in that case. 

### Testing

* manually fixed the code in the output to see if it works for an extension
* ran `cargo test`
* __no tests with wasm-pack were done__

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -377,13 +377,14 @@ impl<'a> Context<'a> {
             // In `--target no-modules` mode we need to both expose a name on
             // the global object as well as generate our own custom start
             // function.
+            // `document.currentScript` property can be null in browser extensions
             OutputMode::NoModules { global } => {
                 js.push_str("const __exports = {};\n");
                 js.push_str("let script_src;\n");
                 js.push_str(
                     "\
                     if (typeof document !== 'undefined') {
-                        script_src = new URL(document.currentScript.src, location.href).toString();
+                        script_src = new URL((document.currentScript && document.currentScript.src) ? document.currentScript.src : '', location.href).toString();
                     }\n",
                 );
                 js.push_str("let wasm = undefined;\n");
--- a/crates/cli/tests/wasm-bindgen/main.rs
+++ b/crates/cli/tests/wasm-bindgen/main.rs
@@ -356,7 +356,7 @@ fn default_module_path_target_no_modules() {
     assert!(contents.contains(
         "\
     if (typeof document !== 'undefined') {
-        script_src = new URL(document.currentScript.src, location.href).toString();
+        script_src = new URL((document.currentScript && document.currentScript.src) ? document.currentScript.src : '', location.href).toString();
     }",
     ));
     assert!(contents.contains(


crates/cli-support/src/js/mod.rs
```diff
@@ -377,13 +377,14 @@ impl<'a> Context<'a> {
...
             // function.
+            // `document.currentScript` property can be null in browser extensions
             OutputMode::NoModules { global } => {
                 js.push_str("const __exports = {};\n");
                 js.push_str("let script_src;\n");
                 js.push_str(
                     "\
                     if (typeof document !== 'undefined') {
-                        script_src = new URL(document.currentScript.src, location.href).toString();
+                        script_src = new URL((document.currentScript && document.currentScript.src) ? document.currentScript.src : '', location.href).toString();
```
holly-hacker, line 387: `document.currentScript?.src ?? ''` may be cleaner here

daxpedda, line 387: We should probably just leave `script_src` as `undefined` in this case, so the check could just be moved into the `if` condition.

rimutaka: 1. tested leaving it `undefined` manually
2. changed the if-statement as requested

crates/cli-support/src/js/mod.rs
```diff
@@ -377,12 +377,13 @@ impl<'a> Context<'a> {
...
             // the global object as well as generate our own custom start
             // function.
+            // `document.currentScript` property can be null in browser extensions
             OutputMode::NoModules { global } => {
                 js.push_str("const __exports = {};\n");
                 js.push_str("let script_src;\n");
                 js.push_str(
                     "\
-                    if (typeof document !== 'undefined') {
+                    if (typeof document !== 'undefined' && document.currentScript && document.currentScript.src) {
```
daxpedda, line 386:
```suggestion
                    if (typeof document !== 'undefined' && typeof document.currentScript !== 'null') {
```
Seems the spec says it can only be `HTMLScriptElement` or `null`.
https://html.spec.whatwg.org/multipage/dom.html#dom-document-currentscript-dev

rimutaka: @daxpedda , changed as suggested, but I think `src` property can still be null. There is nothing in the spec about it being required and in at least one place it is implied as optional. We probably don't have to worry about those edge use cases.

daxpedda: I guess we can deal with it if we actually **know** that it can happen, otherwise we would do checks like this literally everywhere.

holly-hacker:
There is 1 failing test, but it seems unrelated to this PR. Is anything else blocking this merge?

daxpedda:
> There is 1 failing test, but it seems unrelated to this PR. Is anything else blocking this merge?

LGTM.
Unfortunately I'm not a maintainer here.

alexcrichton:
Thanks!

PR #3408, opened 2023-04-28
add verbose name for internal `ptr` property

### LINKED ISSUE #3388: Using "ptr" as struct field name causes infinite loop
### Describe the Bug

If a struct has `ptr` as a field name, instantiating the object will cause the application to enter an infinite loop and eventually error with "Error: too much recursion".

### Steps to Reproduce

```rust
#[wasm_bindgen] 
pub struct UhOh {
    pub ptr: *const u8
}

#[wasm_bindgen]
pub fn test() -> UhOh {
    UhOh { ptr: 0 as *const u8 }
} 
```

1. Run the above `test` function in JavaScript.
2. Notice the error about too much recursion in the console.

### Expected Behavior

The UhOh object should be returned from the test function.

### Actual Behavior

The following error appears in the console:
```
Uncaught (in promise) Error: Error: too much recursion
```

### Additional Context

I believe it is caused by the following part of the generated JS:
```javascript
export class UhOh {

    static __wrap(ptr) {
        const obj = Object.create(UhOh.prototype);
        obj.ptr = ptr;    // <---- entering infinite loop here

        return obj;
    }
```

snOm3ad:
This PR changes the name of the `ptr` property to `__wbg_ptr` during the JS generation step this prevents clashes with public struct fields in Rust client code.

Fixes #3388

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -503,7 +503,7 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
         }
         self.prelude(&format!(
             "\
-                if ({0}.ptr === 0) {{
+                if ({0}.__wbg_ptr === 0) {{
                     throw new Error('Attempt to use a moved value');
                 }}
             ",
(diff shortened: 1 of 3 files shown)


daxpedda:
Some tests fail, you can ignore the "UI compile-fail tests", but the "wasm-bindgen crate tests" are legit failures.

alexcrichton:
Thanks!

Issue #3419, opened 2023-05-11
`wasm-bindgen-test-runner` concurrency problem if two test are run at the same time in the same workspace


lumasepa:
### `wasm-bindgen-test-runner` concurrency problem if two test are run at the same time in the same workspace

If you run configure `wasm-bindgen-test-runner` as a test runner in a workspace and you ran concurrently two or more `cargo test` in different crates of the same workspace there is a possibility that the files served to the browser get mixed between the tests because the directory used by `wasm-bindgen-test-runner` is shared for all the crates.

Some context: 

I'm testing some software that uses libp2p in wasm, as a peer-to-peer network in my test I launch several `cargo tests` that represent different nodes in the network and they are launched concurrently.

### Steps to Reproduce

1. Create a workspace with two crates with wasm-bindgen-test
2. Launch both tests in parallel
3. It may fail if not repeat step 2 till it fails

### Expected Behavior

The directory used to serve files to the browser is not shared between crates. There should be a directory per crate or run.

### Actual Behavior

The directory used to serve files to the browser is shared between crates.

PR #3420, opened 2023-05-11
`wasm-bindgen-test-runner` fix run in parallel in same workspace

lumasepa:
Fix for #3419

Maybe we could delete the temporary directory after finishing the test so the target directory doesn't grow infinitely. If requested I will implement it.


crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -44,6 +44,11 @@ fn main() -> anyhow::Result<()> {
         None => bail!("must have a file to test as first argument"),
     };
 
+    let wasm_file_to_test_name = wasm_file_to_test
+        .file_name()
+        .and_then(|s| s.to_str())
+        .ok_or_else(|| anyhow!("file to test is not a valid file, can't extract file name"))?;
```
daxpedda, line 50:
```suggestion
        .context("file to test is not a valid file, can't extract file name")?;
```
That should work, see https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context.

daxpedda (changes requested):
This looks good to me, but yes, folders should be deleted after finishing.

daxpedda:
Actually I just confirmed that Cargo itself doesn't delete tests between builds or tests, but I still think we should delete the folders because `wasm-bindgen` doesn't do any fingerprinting and will regenerate the file anyway.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -32,6 +32,16 @@ enum TestMode {
     Worker { no_modules: bool },
 }
 
+struct TmpDirDeleteGuard {
+    path: PathBuf,
+}
+
+impl Drop for TmpDirDeleteGuard {
+    fn drop(&mut self) {
+        let _ = fs::remove_dir_all(&self.path);
```
daxpedda, line 41: Can we log the error here instead of discarding it?

daxpedda (changes requested):

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
lumasepa: `eprintln!` is fine?

lumasepa: I saw that you are using env_logger, will use log then.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -32,6 +33,18 @@ enum TestMode {
     Worker { no_modules: bool },
 }
 
+struct TmpDirDeleteGuard {
+    path: PathBuf,
+}
```
daxpedda, line 38: Small nit, let's make this a tuple struct.

daxpedda (changes requested):
LGTM, small nit.

daxpedda (approved):
LGTM, waiting for CI.

PR #3428, opened 2023-05-15
Improved version mismatch error message

### LINKED ISSUE #2619: version mismatch error message doesn't propose a working fix
### Describe the Bug
Via using the [trunk](https://github.com/thedodd/trunk) tool on my project, I encountered this warning:

```
Jul 13 11:58:31.495  INFO calling wasm-bindgen
error: 

version of wasm-bindgen that uses a different bindgen format than this binary:

  rust wasm file schema version: 0.2.69
     this binary schema version: 0.2.74

Currently the bindgen format is unstable enough that these two schema versions
must exactly match. You can accomplish this by either updating the wasm-bindgen
dependency or this binary.

You should be able to update the wasm-bindgen dependency with:

    cargo update -p wasm-bindgen

or you can update the binary with

    cargo install -f wasm-bindgen-cli

if this warning fails to go away though and you're not sure what to do feel free
to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!

Jul 13 11:58:31.516 ERROR ❌ error
error from HTML pipeline

Caused by:
    0: failed to spawn assets finalization
    1: wasm-bindgen call returned a bad status
```

Neither of the suggested fixes did anything, since a project dependency required exactly version 0.2.69.

What did work was `cargo install -f wasm-bindgen-cli --version 0.2.69`. 
Perhaps `wasm-bindgen` should suggest this as well.

snOm3ad:
Improved suggested commands (and wording) for the error that shows up whenever there is a version mismatch between the `wasm-bindgen` version that is used by the CLI tool and a compiled wasm file.

Fixes #2619

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -1479,16 +1479,17 @@ version of wasm-bindgen that uses a different bindgen format than this binary:
      this binary schema version: {}
 
 Currently the bindgen format is unstable enough that these two schema versions
-must exactly match. You can accomplish this by either updating the wasm-bindgen
-dependency or this binary.
+must exactly match. You can accomplish this by either updating this binary or 
+the wasm-bindgen dependency in the Rust project.
 
 You should be able to update the wasm-bindgen dependency with:
 
-    cargo update -p wasm-bindgen
+    cargo update -p wasm-bindgen --precise {my_version}
 
-or you can update the binary with
+don't forget to recompile your wasm file! Alternatively, you can update the 
+binary with:
 
-    cargo install -f wasm-bindgen-cli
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/wit/mod.rs
```diff
@@ -1479,16 +1479,17 @@ version of wasm-bindgen that uses a different bindgen format than this binary:
      this binary schema version: {}
```
daxpedda, line 1479:
```suggestion
     this binary schema version: {my_version}
```
Let's use the new syntax and remove the parameters.

daxpedda (changes requested):
LGTM, small nit.

daxpedda (approved):
Thank you!

PR #3431, opened 2023-05-15
Mark some `JsValue` functions as `const`

relrelb:


daxpedda (approved):
LGTM, but I would like @alexcrichton to take a look at this and see if this is sane.

alexcrichton (approved):

PR #3433, opened 2023-05-16
Add WebIDLs for experimental AudioContext sinkId, setSinkId

tylerbarker:
Hey, I'm working on some audio stuff and found these features I need were missing - thought it might be useful to contribute? Let me know if I'm missing something or have done this incorrectly. Cheers ✌️


tylerbarker:
Okay what I can gather from the failing CI is there's an issue because I mistakenly hard-coded the experimental feature decorators in `gen_AudioContext.rs`.

I can see now they need to be included in the `unstable/` directory instead, how would I go about that given AudioContext is already stable? Is there a way I can write a partial IDL in `unstable/` which I can include in the main IDL?

daxpedda:
Yes, you should be able to use [`partial interface`](https://webidl.spec.whatwg.org/#dfn-partial-interface), it's used quite often in `wasm-bindgen` already.

Let me know if you encounter any problems.

crates/web-sys/webidls/unstable/AudioContext.webidl
```diff
@@ -0,0 +1,6 @@
+partial interface AudioContext {
+  readonly        attribute DOMString sinkId;
+
+  [Throws]
```
daxpedda, line 4: Where did you get that this throws?
The spec doesn't mention it: https://webaudio.github.io/web-audio-api/#dom-audiocontext-setsinkid-domstring-or-audiosinkoptions-sinkid.
Chromium doesn't do it either: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/interfaces/webaudio.idl;l=83?q=setsinkid&ss=chromium%2Fchromium%2Fsrc.

crates/web-sys/webidls/unstable/AudioContext.webidl
```diff
@@ -0,0 +1,6 @@
...
+  Promise<undefined> setSinkId(DOMString sinkId);
```
daxpedda, line 5:
```suggestion
  [SecureContext] Promise<undefined> setSinkId ((DOMString or AudioSinkOptions) sinkId);
```
We should follow the specification.

crates/web-sys/webidls/unstable/AudioContext.webidl
```diff
@@ -0,0 +1,6 @@
+partial interface AudioContext {
+  readonly        attribute DOMString sinkId;
```
daxpedda, line 2:
```suggestion
  [SecureContext] readonly attribute (DOMString or AudioSinkInfo) sinkId;
```

daxpedda (changes requested):
We should implement the rest of the interface too:
- `AudioSinkInfo`
- `AudioSinkOptions`
- `AudioSinkType`
- `BaseAudioContext.onsinkchange`
- `AudioContextOptions.sinkId`

crates/web-sys/webidls/unstable/AudioContext.webidl
tylerbarker: Ah, yeah I've not properly understood the mechanism then, I was under the impression promises have the possibility of throwing and I needed it here like I saw else - will remove 👍

daxpedda: Just to clarify, `Promise`s throwing or not is handled by the `Promise` type itself, no need to label it in the WebIDL at all.

tylerbarker: FWIW this was what I was looking at: [MDN setSinkId Exceptions](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/setSinkId#exceptions)

tylerbarker: > Promises throwing or not is handled by the Promise type itself, no need to label it in the WebIDL at all.
That's helpful, thanks!

daxpedda: > FWIW this was what I was looking at: [MDN setSinkId Exceptions](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/setSinkId#exceptions)

I just tried it myself, I can confirm that the exceptions listed here are rejected promises and aren't thrown on the function itself.

daxpedda:
To generate the WebIDL correctly you have to run the following:
```
cd crates/web-sys
cargo run --release --package wasm-bindgen-webidl -- webidls src/features ./Cargo.toml
```

daxpedda (changes requested):
LGTM!
Missing an entry in the changelog.

daxpedda:
Thanks @tylerbarker!

I just rebased and fixed the changelog.

PR #3434, opened 2023-05-16
Fix the `extra-traits` feature in the backend crate

lukaslihotzki:
The backend crate does not compile with the `extra-traits` feature enabled. I noticed it while working on #3300, but this fix probably belongs to its own PR.


daxpedda (approved):
Fun. Alright, thank you!

PR #3435, opened 2023-05-17
Removing [Throws] from FileSystemWritableFileStream functions which pass errors via the returned promise rather than throwing directly

djozis:
Based on https://github.com/rustwasm/wasm-bindgen/pull/3327#discussion_r1195730766

It seems these functions don't throw directly.
In the above linked comment thread @daxpedda mentioned:

> According to the spec all three can throw: https://streams.spec.whatwg.org/#writablestream-get-a-writer.

But when I tested it, the condition on which they would throw while getting a writer seems to result in the promise having an error, not the function throwing.

Here's a test one can run in developer tools to verify this:
```
const storage = await window.navigator.storage.getDirectory();
const fileHandle = await storage.getFileHandle('test', { create: true });
const writable = await fileHandle.createWritable();
const writer = writable.getWriter(); // lock the stream
try {
    console.log(writable.write('x')); // will fail to get a writer internally
} catch(e) {
    // this won't run, instead the above promise will fail
    console.log("CAUGHT");
}
try {
    console.log(writable.seek(0)); // will fail to get a writer internally
} catch(e) {
    // this won't run, instead the above promise will fail
    console.log("CAUGHT");
}
try {
    console.log(writable.truncate(0)); // will fail to get a writer internally
} catch(e) {
    // this won't run, instead the above promise will fail
    console.log("CAUGHT");
}
```
I ran it on Chrome and Edge, which never printed `CAUGHT` and logged failed promises.
Weirdly, Firefox has `Throws` only on `write`: https://searchfox.org/mozilla-central/source/dom/webidl/FileSystemWritableFileStream.webidl#23-25
I'm not familiar with which condition would throw directly rather than result in a promise with an error based on any documentation I could find.

### DIFF
--- a/crates/web-sys/src/features/gen_FileSystemWritableFileStream.rs
+++ b/crates/web-sys/src/features/gen_FileSystemWritableFileStream.rs
@@ -16,7 +16,7 @@ extern "C" {
     #[doc = "[described in the `wasm-bindgen` guide](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html)*"]
     pub type FileSystemWritableFileStream;
     #[cfg(web_sys_unstable_apis)]
-    # [wasm_bindgen (catch , method , structural , js_class = "FileSystemWritableFileStream" , js_name = seek)]
+    # [wasm_bindgen (method , structural , js_class = "FileSystemWritableFileStream" , js_name = seek)]
     #[doc = "The `seek()` method."]
     #[doc = ""]
     #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream/seek)"]
@@ -25,12 +25,9 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API is unstable and requires `--cfg=web_sys_unstable_apis` to be activated, as"]
     #[doc = "[described in the `wasm-bindgen` guide](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html)*"]
-    pub fn seek_with_u32(
-        this: &FileSystemWritableFileStream,
-        position: u32,
-    ) -> Result<::js_sys::Promise, JsValue>;
+    pub fn seek_with_u32(this: &FileSystemWritableFileStream, position: u32) -> ::js_sys::Promise;
     #[cfg(web_sys_unstable_apis)]
-    # [wasm_bindgen (catch , method , structural , js_class = "FileSystemWritableFileStream" , js_name = seek)]
+    # [wasm_bindgen (method , structural , js_class = "FileSystemWritableFileStream" , js_name = seek)]
     #[doc = "The `seek()` method."]
     #[doc = ""]
     #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemWritableFileStream/seek)"]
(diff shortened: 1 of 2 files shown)


daxpedda:
There are some more philosophical problems here.

The interfaces we use are written according to the spec, not how browsers implement it. The spec says these functions can throw, but if implementations are able to make it not throw, I guess that's good for them, but we will should continue to follow the spec.

This probably has the following reasons:
- Able to support other browsers and platforms that have different implementations.
- Future-proofing, because browsers or platforms can change their implementation as long as it doesn't violate the spec.

That said, sometimes browsers will violate the spec consistently, which we might want to accommodate. E.g. a function that can't throw according to the spec, but browsers still throw, then we should mark it as throwing. But if we mark a function that should throw as non-throwing because browsers don't implement it this way, then we wouldn't be able to fix it without a breaking change.

So all in all, I would say the current interface is correct. In any case, we can't change it anymore, because this would be a breaking change.

Thank you for taking your time to delve into this btw, it is appreciated to have somebody else think about these problems and provide some input for us!

djozis:
Thanks for your clarifications, @daxpedda! I see what you're saying about needing to follow the spec rather than implementations, and avoiding breaking changes.

I'd like to understand one last thing, if I may. Looking at the "write" function (which is async) as an example: https://fs.spec.whatwg.org/#api-filesystemwritablefilestream-write
> The write(data) method steps are:
> Let writer be the result of [getting a writer](https://streams.spec.whatwg.org/#writablestream-get-a-writer) for [this](https://webidl.spec.whatwg.org/#this).
> ...

I understand that you see the first step, and on the linked documentation:
> To get a writer for a [WritableStream](https://streams.spec.whatwg.org/#writablestream) stream, return ? [AcquireWritableStreamDefaultWriter](https://streams.spec.whatwg.org/#acquire-writable-stream-default-writer)(stream). The result will be a [WritableStreamDefaultWriter](https://streams.spec.whatwg.org/#writablestreamdefaultwriter).
> This will throw an exception if stream is already locked.

My understanding is that you see this async `write` function where the first step states in its own documentation that it can throw an exception, and interpret that as that the async `write` function could synchronously throw an exception rather than asynchronously returning an error through the promise. This seemed ambiguous to me, which is why I checked the Chrome/Edge implementations to get an idea of how their developers interpreted this.

Do you have any thoughts or experiences to share on how you interpreted this? For example, is it a case of always erring on the side of caution, or do you maybe see other examples where similar documentation does result in async functions which synchronously throw exceptions, or maybe even standards somewhere else (either for web-sys/wasm-bindgen or for the documentation itself) which clarify how this sort of situation should be interpreted? I'm asking because I'm interested in refining how I interpret nuances in these web specs in the future, so I'm curious for any insights of someone more experienced.

Thanks again for taking the time to share!

daxpedda:
Throwing exceptions is not the same as rejecting a `Promise`, that's why it has this notation:
image
... compare that to:
image
... which just puts it to text.

If you follow down the references, you also land at [`SetUpWritableStreamDefaultWriter()`](https://streams.spec.whatwg.org/#set-up-writable-stream-default-writer), where you can more clearly see that they differentiate between throwing an exception and rejecting with one:
image
... which is the exception described in ["get a writer"](https://streams.spec.whatwg.org/#writablestream-get-a-writer).
Compare that to:
image
... which as before, talks about "rejecting" a `Promise`.

---

Standards aren't exactly easy to read and you have to patiently examine every word, it requires practice.
Unfortunately, the WHATWG documents have many mistakes and inconsistencies, so sometimes it might be very unclear what is meant.

At the end of the day, if you are really unsure about something, you can ask at that documents GitHub repo: https://github.com/whatwg/streams.

I hope this was helpful!

PR #3442, opened 2023-05-20
Clippy: Fixes and CI for examples and tests

lukaslihotzki:
@daxpedda

This also sets rust-version to 1.56 for most examples to prevent unnecessary changes. Where let else chains come handy, I upgraded to 1.65.

### DIFF
--- a/benchmarks/src/lib.rs
+++ b/benchmarks/src/lib.rs
@@ -59,7 +59,7 @@ pub fn fibonacci(n: i32) -> i32 {
     unsafe {
         FIB_HIGH = (a >> 32) as i32;
     }
-    return a as i32;
+    a as i32
 }
 
 #[wasm_bindgen]
--- a/benchmarks/src/lib.rs
+++ b/benchmarks/src/lib.rs
@@ -68,16 +68,16 @@ pub fn fibonacci_high() -> i32 {
 }
 
 #[wasm_bindgen]
-pub fn call_foo_bar_final_n_times(n: usize, foo: &Foo) {
+pub fn call_foo_bar_final_n_times(n: usize, js_foo: &Foo) {
     for _ in 0..n {
-        foo.bar_final();
+        js_foo.bar_final();
     }
 }
 
 #[wasm_bindgen]
-pub fn call_foo_bar_structural_n_times(n: usize, foo: &Foo) {
+pub fn call_foo_bar_structural_n_times(n: usize, js_foo: &Foo) {
     for _ in 0..n {
-        foo.bar_structural();
+        js_foo.bar_structural();
     }
 }
 
--- a/examples/todomvc/src/scheduler.rs
+++ b/examples/todomvc/src/scheduler.rs
@@ -13,6 +13,10 @@ pub struct Scheduler {
     running: RefCell<bool>,
 }
 
+fn deadlock() -> ! {
+    exit("This might be a deadlock");
+}
+
 impl Scheduler {
     /// Construct a new `Scheduler`
     pub fn new() -> Scheduler {
@@ -25,19 +29,13 @@ impl Scheduler {
     }
 
     pub fn set_controller(&self, controller: Controller) {
-        if let Ok(mut controller_data) = self.controller.try_borrow_mut() {
-            *controller_data = Some(controller);
-        } else {
-            exit("This might be a deadlock");
-        }
+        let Ok(mut controller_data) = self.controller.try_borrow_mut() else { deadlock() };
+        *controller_data = Some(controller);
     }
 
     pub fn set_view(&self, view: View) {
-        if let Ok(mut view_data) = self.view.try_borrow_mut() {
-            *view_data = Some(view);
-        } else {
-            exit("This might be a deadlock");
-        }
+        let Ok(mut view_data) = self.view.try_borrow_mut() else { deadlock() };
+        *view_data = Some(view);
     }
 
     /// Add a new message onto the event stack
@@ -45,19 +43,12 @@ impl Scheduler {
     /// Triggers running the event loop if it's not already running
     pub fn add_message(&self, message: Message) {
         let running = {
-            if let Ok(running) = self.running.try_borrow() {
-                *running
-            } else {
-                exit("This might be a deadlock");
-                false
-            }
+            let Ok(running) = self.running.try_borrow() else { deadlock() };
+            *running
         };
         {
-            if let Ok(mut events) = self.events.try_borrow_mut() {
-                events.push(message);
-            } else {
-                exit("This might be a deadlock");
-            }
+            let Ok(mut events) = self.events.try_borrow_mut() else { deadlock() };
+            events.push(message);
         }
         if !running {
             self.run();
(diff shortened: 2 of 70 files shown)


tests/wasm/no_shims.rs
```diff
@@ -22,20 +22,20 @@ use wasm_bindgen_test::*;
     module.exports.incoming_u32 = function () { return 4294967295; };
     module.exports.incoming_i32 = function () { return 0; };
     module.exports.incoming_f32 = function () { return 1.5; };
-    module.exports.incoming_f64 = function () { return 13.37; };
+    module.exports.incoming_f64 = function () { return Math.PI; };
```
daxpedda, line 25: π is not 13.37, but I guess this is a nice change.

tests/wasm/no_shims.rs
```diff
@@ -45,8 +45,8 @@ use wasm_bindgen_test::*;
     };
 
     module.exports.MyNamespace = {};
-    module.exports.MyNamespace.incoming_namespaced = function () { return 3.14; };
-    module.exports.MyNamespace.outgoing_namespaced = function (pi) { assert_eq(3.14, pi); };
+    module.exports.MyNamespace.incoming_namespaced = function () { return 13.37; };
+    module.exports.MyNamespace.outgoing_namespaced = function (pi) { assert_eq(13.37, pi); };
```
daxpedda, line 49: So why the change here?

tests/wasm/no_shims.rs
```diff
@@ -133,14 +133,14 @@ fn no_shims() {
...
-    assert_eq!(z, 13.37);
+    assert_eq!(z, std::f64::consts::PI);
     outgoing_f64(z);
 
     let w = many(x, y, z);
     assert_eq!(w, 42);
 
     let pi = incoming_namespaced();
-    assert_eq!(pi, 3.14);
+    assert_eq!(pi, 13.37);
```
daxpedda, line 143: Now I am just confused :laughing:.

daxpedda (changes requested):
Thank you so much for the amazing work!
I would prefer if we keep everything at the same MSRV: 1.56.

Maybe we should consider increasing it to something higher, I'm all for it, but in a separate PR.

benchmarks/src/lib.rs
```diff
@@ -68,16 +68,16 @@ pub fn fibonacci_high() -> i32 {
 }
 
 #[wasm_bindgen]
-pub fn call_foo_bar_final_n_times(n: usize, foo: &Foo) {
+pub fn call_foo_bar_final_n_times(n: usize, js_foo: &Foo) {
```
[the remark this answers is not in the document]
lukaslihotzki: `foo` is "use of a disallowed/placeholder name". A more descriptive name would highlight that this parameter is an object on the JS side (that's what makes it interesting for the benchmark). Therefore, `js_foo` (or `js_obj`?).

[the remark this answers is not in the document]
daxpedda: I honestly think `foo` is appropriate here, personally I would have `allow`ed the lint here.
But please follow your instincts here, this is just some subjective opinion from my side.

tests/wasm/js_objects.rs
```diff
@@ -172,12 +173,12 @@ fn serde() {
         .unwrap(),
     );
 
-    let foo = ret.into_serde::<SerdeFoo>().unwrap();
-    assert_eq!(foo.a, 2);
-    assert_eq!(foo.b, "bar");
-    assert!(foo.c.is_some());
-    assert_eq!(foo.c.as_ref().unwrap().a, 3);
-    assert_eq!(foo.d.a, 4);
+    let result = ret.into_serde::<SerdeFoo>().unwrap();
```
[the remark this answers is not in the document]
lukaslihotzki: `foo` is "use of a disallowed/placeholder name". Alternatively, maybe `roundtrip`? (because from_serde, then back into_serde).

tests/wasm/no_shims.rs
lukaslihotzki: `3.14` was used as an arbitrary value, and the lint suggested to use the correct PI constant. I thought that was a great idea (as a precision test), but not specifically for the namespaced test. Therefore, I swapped the PI constant to the regular incoming/outgoing test and used 13.37 as a simpler value for the namespaced test.

tests/wasm/js_objects.rs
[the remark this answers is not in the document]
daxpedda: As above, do what you think is right here.

tests/wasm/no_shims.rs
daxpedda: I see.
Using `Math.PI` was definitely a nice change.
But `outgoing_namespaced` specifically used the parameter name `pi`, which you changed to 13.37, so that weird.

I think it's perfectly fine like that, just change the parameter name of `outgoing_namespaced` to something else then `pi`.

daxpedda: Nice!

tests/wasm/classes.rs
```diff
@@ -599,12 +599,12 @@ impl OverriddenInspectable {
     }
 
     #[wasm_bindgen(js_name = toJSON)]
-    pub fn to_json(&self) -> String {
+    pub fn js_to_json(&self) -> String {
```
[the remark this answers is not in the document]
lukaslihotzki: For `to_string`, the lint `inherent_to_string` suggests to implement the `Display` trait instead of adding an inherent method. That is unsupported with wasm_bindgen. Therefore, I changed the practically irrelevent Rust-side name to `js_to_string` to highlight that this is the JS-side `toString` method, not designed for Rust callers. Maybe `to_string_js` would be better? Then, I did the same for `to_json` for consistency.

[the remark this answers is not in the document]
daxpedda: That's fine.

examples/todomvc/src/scheduler.rs
```diff
@@ -29,26 +25,39 @@ impl Scheduler {
...
     /// Triggers running the event loop if it's not already running
     pub fn add_message(&self, message: Message) {
         let running = {
-            let Ok(running) = self.running.try_borrow() else { deadlock() };
-            *running
+            if let Ok(running) = self.running.try_borrow() {
+                *running
+            } else {
+                exit("This might be a deadlock");
+                false
```
daxpedda:
```suggestion
                exit("This might be a deadlock");
```

examples/todomvc/src/scheduler.rs
```diff
@@ -29,26 +25,39 @@ impl Scheduler {
...
         };
         {
-            let Ok(mut events) = self.events.try_borrow_mut() else { deadlock() };
-            events.push(message);
+            if let Ok(mut events) = self.events.try_borrow_mut() {
+                events.push(message);
+            } else {
+                exit("This might be a deadlock");
+            }
         }
```
daxpedda: I think the extra `{ .. }` can be removed here.

examples/todomvc/src/scheduler.rs
```diff
@@ -57,46 +66,67 @@ impl Scheduler {
...
-                let Ok(mut running) = self.running.try_borrow_mut() else { deadlock() };
-                *running = true;
+                if let Ok(mut running) = self.running.try_borrow_mut() {
+                    *running = true;
+                } else {
+                    exit("This might be a deadlock");
+                }
             }
             self.next_message();
         }
```
daxpedda: Same as above.

examples/todomvc/src/scheduler.rs
```diff
@@ -57,46 +66,67 @@ impl Scheduler {
...
         let event = {
-            let Ok(mut events) = self.events.try_borrow_mut() else { deadlock() };
-            events.pop()
+            if let Ok(mut events) = self.events.try_borrow_mut() {
+                Some(events.pop())
+            } else {
+                exit("This might be a deadlock");
+                None
+            }
         };
```
daxpedda: As above.

examples/todomvc/src/scheduler.rs
```diff
@@ -57,46 +66,67 @@ impl Scheduler {
...
-                    if let Some(ref mut ag) = *view {
-                        ag.call(e);
+                    if let Ok(mut view) = self.view.try_borrow_mut() {
+                        if let Some(ref mut ag) = *view {
+                            ag.call(e);
+                        }
+                    } else {
+                        exit("This might be a deadlock");
                     }
                 }
```
daxpedda: As above.

daxpedda (changes requested):

examples/todomvc/src/scheduler.rs
lukaslihotzki: I don't see any `{ … }` that can be removed here.

daxpedda: Me neither, my mistake.

daxpedda (approved):
Thanks!

PR #3446, opened 2023-05-22
Use `table.fill` in externref-xform (`TODO`)

lukaslihotzki:


daxpedda (changes requested):
First off: this is amazing, thank you!

As far as I know the reference-type proposal is not a requirements for the atomic proposal, so either we will have to adjust the documentation or gate it.

I would prefer to gate it, which should be possible.

Will pull in [alexcrichton](https://github.com/alexcrichton) when we are ready to review this, I don't feel comfortable enough with Wasm's instruction set to approve this.

lukaslihotzki:
I don't understand where gating makes sense. The `table.set` and `table.fill` instructions are both newly defined in the reference-types proposal, so I don't see any additional requirements. If `externref_stack == 0`, neither of those instructions is used anyway.

daxpedda:
Ah, apologies! For some reason I thought we are in `threads-xform` :rofl:.
@alexcrichton if you wouldn't mind taking a look at this.

alexcrichton:
Yes I believe that this is correct but I'm also not sure whether it's worth doing. For small sizes it may be better to have the currently unrolled loop perf-wise and it may only be advantageous for larger sizes to use `table.fill`. I'm not entirely sure though in that I haven't benchmarked any wasm engine to see the difference.

daxpedda:
I would have argued that loop unrolling is a compilers job when it's applying optimizations, on the other hand this is not something I really know much about.

@lukaslihotzki would you mind research this a bit:
- Benchmark it.
- Find out if the LLVM or `wasm-opt` unrolls `table.fill`.
- See how it would make sense to proceed here.

Thank you!

daxpedda:
I'm going to go ahead and merge this.
I would argue that the current implementation is premature optimization as long as we don't have any data and the change proposed here is the straightforward way to go about it.

daxpedda (approved):

PR #3458, opened 2023-06-04
Feat: Fixed size array support

stegaBOB:


daxpedda:
Wow, that looks surprisingly less complicated then I imagined! Amazing work!

Keep in mind that I'm not entirely sure how this will conflict with https://github.com/rustwasm/wasm-bindgen/issues/3454 in the future. But definitely willing to merge this, https://github.com/rustwasm/wasm-bindgen/issues/3454 might be a long way off.

daxpedda:
I wouldn't feel comfortable enough reviewing something not FFI safe, so that's definitely preferable.

PR #3463, opened 2023-06-05
Take alignment into consideration during `malloc`

### LINKED ISSUE #3461: Exporting bigint slices panics
As of Rust v1.70, exporting a function like this:
```rust
#[wasm_bindgen]
pub fn test(_: &[u64]) {}
```
... will make it panic when called from JS:
```js
test([]);
```
```
panicked at 'misaligned pointer dereference: address must be a multiple of 0x8 but is 0x4'
```

Here is a minimal reproducible example:
https://github.com/daxpedda/wasm-bindgen-bigint-slice-issue

I confirmed the following scenarios:
- Only `u64` and `i64` integer types cause this issue.
- Rust v1.69 does not have this issue.
- The recent Rust changelog doesn't note any significant changes other then the upgrade to LLVM 16, which added default target features to Wasm:
  - Compiling with `-Ctarget-feature=-sign-ext,-mutable-globals` and `-Zbuild-std` on nightly-2023-06-03 does not fix the issue.
  - Compiling with `-Ctarget-feature=+sign-ext,+mutable-globals` on Rust v1.69 does not cause this issue.

This issue was originally discovered here: https://github.com/rustwasm/wasm-bindgen/actions/runs/5176047134/jobs/9324382988#step:7:409.

Update: apparently the bug is coming from https://github.com/rust-lang/rust/pull/98112.
It's possible to circumvent the check by just disabling `debug_assert`s.

daxpedda:
Unfortunately I had to hardcode an alignment of 4 because we didn't implement something to detect the target architecture when using `wasm-bindgen-cli`.

~~This is still missing adjustments on `__wbindgen_realloc` and `__wbindgen_free`.~~

This is fairly hacky, unfortunately I simply don't know enough about the `wasm-bindgen` codebase. Feel free to suggest how it could be done right.

Fixes #3461.


crates/threads-xform/src/lib.rs
```diff
@@ -344,9 +344,10 @@ fn inject_start(
             // we give ourselves a stack and we update our stack
             // pointer as the default stack pointer is surely wrong for us.
             |body| {
-                // local = malloc(stack.size) [aka base]
+                // local = malloc(stack.size, align) [aka base]
                 with_temp_stack(body, memory, stack, |body| {
                     body.i32_const(stack.size as i32)
+                        .i32_const(4)
```
Liamolucko, line 350:
```suggestion
                        .i32_const(16)
```

I think the stack's actually supposed to be aligned to 16 bytes (source: https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md#the-linear-stack).

crates/threads-xform/src/lib.rs
```diff
@@ -425,12 +427,14 @@ fn inject_destroy(
             // we're destroying somebody else's stack, so we can use our own
             body.local_get(stack_alloc)
                 .i32_const(stack.size as i32)
+                .i32_const(4)
```
Liamolucko, line 430:
```suggestion
                .i32_const(16)
```

crates/threads-xform/src/lib.rs
```diff
@@ -425,12 +427,14 @@ fn inject_destroy(
...
                 .call(free);
         },
         |body| {
             with_temp_stack(body, memory, stack, |body| {
                 body.global_get(stack.alloc)
                     .i32_const(stack.size as i32)
+                    .i32_const(4)
```
Liamolucko, line 437:
```suggestion
                    .i32_const(16)
```

crates/cli-support/src/js/binding.rs
```diff
@@ -813,12 +817,14 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
             let func = js.cx.pass_to_wasm_function(kind.clone(), *mem)?;
             let malloc = js.cx.export_name_of(*malloc);
             let i = js.tmp();
+            let align = std::cmp::max(kind.size(), 4);
```
Liamolucko, line 820: I'm not sure why this is needed; I think it should be fine for the alignment to be less than 4.

crates/cli-support/src/js/binding.rs
```diff
@@ -922,7 +928,7 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
             let malloc = js.cx.export_name_of(*malloc);
             let val = js.pop();
             js.prelude(&format!(
-                "var ptr{i} = isLikeNone({0}) ? 0 : {f}({0}, wasm.{malloc});",
+                "var ptr{i} = isLikeNone({0}) ? 0 : {f}({0}, wasm.{malloc}, 4);",
```
Liamolucko, line 931: Why is this hardcoded? Shouldn't it be `kind.size()`?

crates/cli-support/src/js/binding.rs
```diff
@@ -940,7 +946,7 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
             let malloc = js.cx.export_name_of(*malloc);
             let i = js.tmp();
             js.prelude(&format!(
-                "var ptr{i} = {f}({val}, wasm.{malloc});",
+                "var ptr{i} = {f}({val}, wasm.{malloc}, 4);",
```
Liamolucko, line 949: This also shouldn't be hardcoded to 4.

crates/cli-support/src/js/mod.rs
```diff
@@ -1415,8 +1415,8 @@ impl<'a> Context<'a> {
         self.expose_wasm_vector_len();
         self.global(&format!(
             "
-            function {}(arg, malloc) {{
-                const ptr = malloc(arg.length * {size}) >>> 0;
+            function {}(arg, malloc, align) {{
+                const ptr = malloc(arg.length * {size}, align) >>> 0;
```
Liamolucko, line 1419: This doesn't really need to be an argument on the JS side - we already have access to the size of each element, which is the same as their alignment for all the types we support, so you can statically set it to `{size}` instead.

crates/cli-support/src/js/mod.rs
```diff
@@ -1259,14 +1259,14 @@ impl<'a> Context<'a> {
             "\
                 if (realloc === undefined) {{
                     const buf = cachedTextEncoder.encode(arg);
-                    const ptr = malloc(buf.length) >>> 0;
+                    const ptr = malloc(buf.length, 4) >>> 0;
```
Liamolucko, line 1262: This could be reduced to an alignment of 1.

crates/cli-support/src/js/binding.rs
```diff
@@ -550,11 +550,13 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
...
         | Instruction::DeferCallCore(_) => {
             let invoc = Invocation::from(instr, js.cx.module)?;
-            let (params, results) = invoc.params_results(js.cx);
+            let (mut params, results) = invoc.params_results(js.cx);
 
             let mut args = Vec::new();
             let tmp = js.tmp();
             if invoc.defer() {
+                // substract alignment
+                params -= 1;
```
Liamolucko, line 559: It's not ideal to have to hardcode this to only work with `free`. I might push a commit of my own to work around this.

Liamolucko (commented):
Looks good for the most part!

crates/cli-support/src/js/binding.rs
Liamolucko: My idea didn't pan out, so I guess we can go with this. `DeferCallCore` should probably be renamed to `DeferFree` or something now that it only works with `free`, though.

crates/cli-support/src/js/binding.rs
```diff
@@ -564,6 +566,8 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
                     writeln!(js.prelude, "{name} = {arg};").unwrap();
                     args.push(name);
                 }
+                // add alignment
+                args.push(String::from("4"));
```
Liamolucko, line 570: This should probably be added as a field of the instruction instead of hardcoded.

crates/cli-support/src/js/binding.rs
daxpedda: For some reason I was assuming that the minimum alignment was 4, which is obviously nonsense.

crates/cli-support/src/js/binding.rs
daxpedda: Done.

crates/cli-support/src/js/mod.rs

crates/cli-support/src/wit/outgoing.rs
```diff
@@ -105,7 +105,7 @@ impl InstructionBuilder<'_, '_> {
                 // ... then defer a call to `free` to happen later
                 let free = self.cx.free()?;
                 self.instructions.push(InstructionData {
-                    instr: Instruction::DeferCallCore(free),
```
daxpedda, line 108: I believe an alignment of 1 here is correct?

crates/cli-support/src/js/binding.rs
daxpedda: See https://github.com/rustwasm/wasm-bindgen/pull/3463/files#r1219300930.

crates/cli-support/src/wit/outgoing.rs
```diff
@@ -105,7 +105,7 @@ impl InstructionBuilder<'_, '_> {
...
+                    instr: Instruction::DeferFree { free, align: 1 },
```
daxpedda, line 108: I believe an alignment of 1 here is correct?

crates/cli-support/src/js/binding.rs
```diff
@@ -548,53 +548,32 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
...
-                        js.push("ret".to_string());
-                    } else {
-                        for i in 0..n {
-                            js.push(format!("ret[{}]", i));
-                        }
-                    }
-                }
-            }
+            js.finally(&format!("{};", call));
         }
```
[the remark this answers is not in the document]
daxpedda: Well ... scratch that ...

crates/cli-support/src/js/binding.rs
```diff
@@ -548,13 +548,16 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
...
-        | Instruction::DeferCallCore(_) => {
+        | Instruction::DeferFree { .. } => {
             let invoc = Invocation::from(instr, js.cx.module)?;
-            let (params, results) = invoc.params_results(js.cx);
-
+            let (mut params, results) = invoc.params_results(js.cx);
             let mut args = Vec::new();
             let tmp = js.tmp();
             if invoc.defer() {
+                if let Instruction::DeferFree { .. } = instr {
```
daxpedda, line 557: Right now the only instruction that has `defer` is `DeferFree`, but I added this check nonetheless, I hope that makes sense.

crates/cli-support/src/wit/outgoing.rs
daxpedda: Actually I don't think it is, because even though this is a String, it's taking out the bytes from a `Int32Array` memory.
Maybe it shouldn't be a `Int32Array` in the first place? But that's an issue for another time.

Liamolucko: > Actually I don't think it is, because even though this is a String, it's taking out the bytes from a `Int32Array` memory.

Are you sure? I think you might be looking at the code that fetches the ptr/len of the string from memory.

crates/cli-support/src/wit/standard.rs
```diff
@@ -95,7 +95,10 @@ pub enum Instruction {
     CallCore(walrus::FunctionId),
     /// Schedules a function to be called after the whole lift/lower cycle is
     /// finished, e.g. to deallocate a string or something.
```
Liamolucko, line 97: This comment's now a bit outdated.

guide/src/contributing/design/exporting-rust.md
```diff
@@ -56,7 +56,7 @@ export function greet(arg0) {
     wasm.__wbindgen_boxed_str_free(ret);
     return realRet;
   } finally {
-    wasm.__wbindgen_free(ptr0, len0);
+    wasm.__wbindgen_free(ptr0, len0, 4);
```
Liamolucko, line 59:
```suggestion
    wasm.__wbindgen_free(ptr0, len0, 1);
```

Liamolucko (approved):
This is pretty much good to merge, I just noticed that the bit of the guide you updated is still using an alignment of 4 for strings.

PR #3466, opened 2023-06-06
Export JsValue from js-sys

kajacx:
This way, you can name the `JsValue` type without adding a dependency on `wasm-bindgen`.

On the other hand `js-sys` already depends on `wasm-bindgen`, so it's not that bad. Still, returning a type (or requiring it as an argument) that your crate doesn't provide simply feels wrong.

### DIFF
--- a/crates/js-sys/CHANGELOG.md
+++ b/crates/js-sys/CHANGELOG.md
@@ -9,6 +9,7 @@ Released YYYY-MM-DD.
 ### Added
 
 * Added a re-export of `wasm-bindgen`.
+  [#3466](https://github.com/rustwasm/wasm-bindgen/pull/3466)
 
 ### Changed
 
--- a/crates/js-sys/CHANGELOG.md
+++ b/crates/js-sys/CHANGELOG.md
@@ -8,7 +8,10 @@ Released YYYY-MM-DD.
 
 ### Added
 
-* Added a re-export of `wasm-bindgen`.
+* Re-export `wasm-bindgen` from `js-sys` and `web-sys`.
+  [#3466](https://github.com/rustwasm/wasm-bindgen/pull/3466)
+
+* Re-export `js-sys` from `web-sys`.
   [#3466](https://github.com/rustwasm/wasm-bindgen/pull/3466)
 
 ### Changed
--- a/crates/js-sys/CHANGELOG.md
+++ b/crates/js-sys/CHANGELOG.md
@@ -8,7 +8,7 @@ Released YYYY-MM-DD.
 
 ### Added
 
-* TODO (or remove section if none)
+* Added a re-export of `wasm-bindgen`.
 
 ### Changed
 
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -50,6 +50,8 @@ use wasm_bindgen::prelude::*;
 // * Arguments that are `JsValue`s or imported JavaScript types should be taken
 //   by reference.
 
+pub use wasm_bindgen::prelude::JsValue;
+
 macro_rules! forward_deref_unop {
     (impl $imp:ident, $method:ident for $t:ty) => {
         impl $imp for $t {
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -28,6 +28,7 @@ use std::mem;
 use std::str;
 use std::str::FromStr;
 
+pub use wasm_bindgen;
 use wasm_bindgen::prelude::*;
 
 // When adding new imports:
@@ -49,8 +50,6 @@ use wasm_bindgen::prelude::*;
 //
 // * Arguments that are `JsValue`s or imported JavaScript types should be taken
 //   by reference.
-
-pub use wasm_bindgen::prelude::JsValue;
 
 macro_rules! forward_deref_unop {
     (impl $imp:ident, $method:ident for $t:ty) => {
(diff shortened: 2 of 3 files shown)


daxpedda:
I don't think I would mind that.
It would probably make sense to re-export `wasm-bindgen` itself as well.

Could you add this to the changelog as well?

crates/js-sys/CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ Released YYYY-MM-DD.
 
 ### Added
 
-* TODO (or remove section if none)
+* Added a re-export of `wasm-bindgen`.
```
daxpedda, line 11: If you rebase onto `main` you will find a good template to do this.
The text is fine, just missing the link to the PR.

crates/js-sys/src/lib.rs
```diff
@@ -28,6 +28,7 @@ use std::mem;
 use std::str;
 use std::str::FromStr;
 
+pub use wasm_bindgen;
 use wasm_bindgen::prelude::*;
```
daxpedda, line 32:
```suggestion
pub use wasm_bindgen::prelude::*;
```
I actually meant that we should re-export `wasm-bindgen` **in addition** to `JsValue`, but the prelude is even better!

crates/js-sys/src/lib.rs
```diff
@@ -28,6 +28,7 @@ use std::mem;
 use std::str;
 use std::str::FromStr;
 
+pub use wasm_bindgen;
```
daxpedda, line 31:
```suggestion
pub extern crate wasm_bindgen;
```
This is how reexports are usually done.

daxpedda (changes requested):
> The downside is that unrelated parts of `wasm-bindgen` are exported as well (for example, you can annotate your methods with a `#[js_sys::wasm_bindgen::wasm_bindgen]` macro, which makes little sence, but going with a fine comb to reexport only the types used seems like a lot of work.

Reexports like this are just to make it easier to use dependencies without having to pull in sub-dependencies. So it's fine like this, it's a known pattern in Rust.

This is an opinionated change anyway, so we will wait for another maintainer to chime in.

crates/js-sys/src/lib.rs
kajacx: The only re-export I have worked was in [`wai-bindgen-wasmtime`](https://github.com/wasmerio/wai/blob/main/crates/wasmtime/src/lib.rs#L8) and [`wai-bindgen-wasmer`](https://github.com/wasmerio/wai/blob/main/crates/wasmer/src/lib.rs#L8). Also [this link](https://www.lurklurk.org/effective-rust/re-export.html) suggested using the `pub use dependency_crate` syntax.

But if you want `pub extern crate wasm_bindgen;`, go for it, it's your repo. I checked the "allow maintainers to make changes", so that means you can put the change into my fork of the repo? I'm not super familiar with this on GitHub.

hamza1311:
If this is done, I don't think it should be done in just js-sys. Why not web-sys as well?

crates/js-sys/src/lib.rs
daxpedda: I recently found out that `pub extern` and `pub use` basically emit the same things on docs.rs.

daxpedda:
I agree. WDYT overall though?

hamza1311:
> WDYT overall though?

I'm not a fan of it. All it saves is one line in Cargo.toml. But I also can't think of a reason not to add it

daxpedda:
It's generally considered good behavior to re-export crates that you use types of in your public API, which is the case for us in `js-sys` and `web-sys`.

But I'm happy to proceed now @kajacx if you finish this up.

crates/js-sys/src/lib.rs
daxpedda: I think we should leave it like this.

crates/js-sys/CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ Released YYYY-MM-DD.
...
```
daxpedda, line 11:
```suggestion
* Re-export `wasm-bindgen` from `js-sys` and `web-sys`.
  [#3466](https://github.com/rustwasm/wasm-bindgen/pull/3466)

* Re-export `js-sys` from `web-sys`.
```

daxpedda (changes requested):
Let's also re-export `wasm-bindgen` and `js-sys` from `web-sys`.

kajacx:
> All it saves is one line in Cargo.toml

Yea, but tracking down that line is non-trivial. You have to go to the source of `js-sys` or find it on crates.io to get the version of `wasm-bindgen` it is using. And then you have to keep it updated as `js-sys` and it's `wasm-bindgen` dependency gets updated.

If you could use `js_sys::wasm_bindgen` instead, then you would have a 100% guarantee that you are using the same version of `wasm_bindgen` as `js_sys` is using, without the need to manually update anything.

On the other hand, if someone uses my crate, and *they* want to use `JsValue` as well, then they will you have to write `my_crate::js_sys::wasm_bindgen::JsValue`, which is getting ridiculous.

In the end, this is a preference decision, so I think it's best to keep it to the authors of the crate to make the final call. For example, I will not be exporting `js-sys` from my crate in this way, since that would defeat it's purpose.

daxpedda:
> Let's also re-export `wasm-bindgen` and `js-sys` from `web-sys`.

@kajacx this is still missing, otherwise this is good to go.

kajacx:
> > Let's also re-export `wasm-bindgen` and `js-sys` from `web-sys`.
> 
> @kajacx this is still missing, otherwise this is good to go.

Done.

daxpedda:
Thanks!

PR #3475, opened 2023-06-09
Bump to 0.2.87

daxpedda:
Specifically because of https://github.com/rustwasm/wasm-bindgen/issues/3461.
Which caused a regression in `wasm-bindgen` with Rust v1.70.


alexcrichton:
Thanks!

PR #3478, opened 2023-06-12
Don't throw a TypeError if module.headers is undefined.

danielhjacobs:
See https://github.com/ruffle-rs/ruffle/issues/11504 and https://github.com/ruffle-rs/ruffle/pull/11505#issuecomment-1587946906.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -800,7 +800,7 @@ impl<'a> Context<'a> {
                                 return await WebAssembly.instantiateStreaming(module, imports);
 
                             }} catch (e) {{
-                                if (!module.headers || module.headers.get('Content-Type') !== 'application/wasm') {{
+                                if (typeof module.headers === 'undefined' || module.headers.get('Content-Type') !== 'application/wasm') {{
                                     console.warn(\"`WebAssembly.instantiateStreaming` failed \
                                                     because your server does not serve wasm with \
                                                     `application/wasm` MIME type. Falling back to \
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -800,7 +800,7 @@ impl<'a> Context<'a> {
                                 return await WebAssembly.instantiateStreaming(module, imports);
 
                             }} catch (e) {{
-                                if (module.headers.get('Content-Type') != 'application/wasm') {{
+                                if (!module.headers || module.headers.get('Content-Type') !== 'application/wasm') {{
```
daxpedda, line 803:
```suggestion
                                if (typeof module.headers === 'undefined' || module.headers.get('Content-Type') !== 'application/wasm') {{
```
I would prefer to be more explicit here.
Otherwise LGTM!

daxpedda (changes requested):

daxpedda (approved):
LGTM!

Do you know why this actually happens? I did some research and found that apparently headers are CORS controlled, so that's why they might be undefined. Did you dig into this?

danielhjacobs:
It looks like https://consent.cookiebot.com/Framework/IAB/consent-sdk-2.0.js, used on that site, overrides `Response` (https://www.cookiebot.com/en/). When I create a page with that as a script src, `new Response()` says `Object { cmpId: undefined, cmpVersion: undefined, gdprApplies: undefined, tcfPolicyVersion: undefined }`. I'm guessing that has something to do with this.

So I understand if wasm-bindgen doesn't want to account for every possible overridden value like this and therefore feels this is cookiebot's responsibility.

PR #3484, opened 2023-06-16
Add bindings for WebAssembly.Tag and WebAssembly.Exception

allsey87:
Add bindings for WebAssembly.Tag and WebAssembly.Exception. These new objects are documented on MDN and are available in both Firefox and Chrome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 * Extend `AudioContext` with unstable features supporting audio sink configuration.
   [#3433](https://github.com/rustwasm/wasm-bindgen/pull/3433)
+* Add bindings for `WebAssembly.Tag` and `WebAssembly.Exception`.
+  [#3484](https://github.com/rustwasm/wasm-bindgen/pull/3484)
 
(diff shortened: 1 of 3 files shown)


allsey87:
I am a little bit confused about why the "CI / Run UI compile-fail tests" are failing. It seems to me that this is unrelated to this PR.

daxpedda:
The CI failure is because you introduces new `#[wasm_bindgen]` types, so there are more types now that can be converted from `JsValue`.
To fix it you have to increase [this line](https://github.com/rustwasm/wasm-bindgen/blob/e84fd83f317ea5577f7d6eff283e1774e7d967ec/crates/macro/ui-tests/async-errors.stderr#L38) to 74.

daxpedda (changes requested):
`WebAssembly.Tag` only exists in the editors draft and isn't in the working draft yet. But it is implemented on all browsers, so I think it's fine to add this.

Could you add a changelog entry as well?

daxpedda (approved):
Thanks!

PR #3489, opened 2023-06-19
Add support for conditional ui detection and ff ctap2 detection

Firstyear:
This adds two important API's for Webauthn to web_sys.

The first is `isConditionalMediationAvailable` which allows detection of this feature for resident key username autocompletion. 

The second is `isExternalCTAP2SecurityKeySupported` which is needed for interoperating with firefox. Sadly firefox in some cases does not support CTAP2 and introduced an undocumented api per https://bugzilla.mozilla.org/show_bug.cgi?id=1526023 and this was never added to MDN. It was discussed by w3c/webauthn and seen as temporary per https://github.com/w3c/webauthn/issues/1173

However like all good temporary things, it still exists and it's required to detect if CTAP2 support exists in firefox - which it still does not for many users meaning it needs to be possible to adjust work flows in this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 
 ### Added
 
+* Add bindings for `PublicKeyCredential.isConditionalMediationAvailable`.
+  [#3489](https://github.com/rustwasm/wasm-bindgen/pull/3489)
+
 * Extend `AudioContext` with unstable features supporting audio sink configuration.
   [#3433](https://github.com/rustwasm/wasm-bindgen/pull/3433)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Added
 
-* Add bindings for `PublicKeyCredential.isConditionalMediationAvailable`.
+* Add bindings for `PublicKeyCredential.isConditionalMediationAvailable()`.
   [#3489](https://github.com/rustwasm/wasm-bindgen/pull/3489)
 
 * Extend `AudioContext` with unstable features supporting audio sink configuration.
--- a/crates/web-sys/src/features/gen_PublicKeyCredential.rs
+++ b/crates/web-sys/src/features/gen_PublicKeyCredential.rs
@@ -37,6 +37,17 @@ extern "C" {
     pub fn get_client_extension_results(
         this: &PublicKeyCredential,
     ) -> AuthenticationExtensionsClientOutputs;
+    #[cfg(web_sys_unstable_apis)]
+    # [wasm_bindgen (static_method_of = PublicKeyCredential , js_class = "PublicKeyCredential" , js_name = isConditionalMediationAvailable)]
+    #[doc = "The `isConditionalMediationAvailable()` method."]
+    #[doc = ""]
+    #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/isConditionalMediationAvailable)"]
+    #[doc = ""]
+    #[doc = "*This API requires the following crate features to be activated: `PublicKeyCredential`*"]
+    #[doc = ""]
+    #[doc = "*This API is unstable and requires `--cfg=web_sys_unstable_apis` to be activated, as"]
+    #[doc = "[described in the `wasm-bindgen` guide](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html)*"]
+    pub fn is_conditional_mediation_available() -> ::js_sys::Promise;
     # [wasm_bindgen (static_method_of = PublicKeyCredential , js_class = "PublicKeyCredential" , js_name = isUserVerifyingPlatformAuthenticatorAvailable)]
     #[doc = "The `isUserVerifyingPlatformAuthenticatorAvailable()` method."]
     #[doc = ""]
--- a/crates/web-sys/src/features/gen_PublicKeyCredential.rs
+++ b/crates/web-sys/src/features/gen_PublicKeyCredential.rs
@@ -44,4 +44,16 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `PublicKeyCredential`*"]
     pub fn is_user_verifying_platform_authenticator_available() -> ::js_sys::Promise;
+    # [wasm_bindgen (static_method_of = PublicKeyCredential , js_class = "PublicKeyCredential" , js_name = isConditionalMediationAvailable)]
+    #[doc = "The `isConditionalMediationAvailable()` method."]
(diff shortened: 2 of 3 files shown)


daxpedda:
Please take a look [here](https://github.com/rustwasm/wasm-bindgen/blob/main/guide/src/contributing/web-sys/supporting-more-web-apis.md) to see how to add new APIs.
Generally all APIs are generated through WebIDL and not added manually like this.

I'm happy to add `PublicKeyCredential.isConditionalMediationAvailable()` as an unstable API, but generally speaking we don't add APIs that have been removed like `PublicKeyCredential.isExternalCTAP2SecurityKeySupported()`.
I would advice you to add it manually wherever you need it, it is common to do so for legacy support and feature detection.

Firstyear:
> Please take a look [here](https://github.com/rustwasm/wasm-bindgen/blob/main/guide/src/contributing/web-sys/supporting-more-web-apis.md) to see how to add new APIs. Generally all APIs are generated through WebIDL and not added manually like this.

Would it be better to have this on the README on the front page because I honestly did not find this ... 

> 
> I'm happy to add `PublicKeyCredential.isConditionalMediationAvailable()` as an unstable API, but generally speaking we don't add APIs that have been removed like `PublicKeyCredential.isExternalCTAP2SecurityKeySupported()`. I would advice you to add it manually wherever you need it, it is common to do so for legacy support and feature detection.

The problem is that it's unclear from the wasm-bindgen book or web-sys how to extend an existing web-sys type for bindings.

Additionally, while this isn't "standard" it's not that it's removed either. It's actually a really important api for firefox and having a valid webauthn interaction. It's likely this API will be needed for many years to come too. It's a whole mess sadly :(

Firstyear:
Follow up  - manually attempting to define this for myself in a seperate crate fails. Is there a way to overload a type from web_sys:: so that these functions can be used? 

```
#[wasm_bindgen]
extern "C" {
    # [wasm_bindgen (js_class = "PublicKeyCredential" , js_name = isExternalCTAP2SecurityKeySupported, catch)]
    pub fn is_external_ctap2_securitykey_supported() -> Result<::js_sys::Promise, JsValue>;
}

// Also fails
#[wasm_bindgen]
extern "C" {
    # [wasm_bindgen ( static_method_of = web_sys::PublicKeyCredential, js_class = "PublicKeyCredential" , js_name = isExternalCTAP2SecurityKeySupported, catch)]
    pub fn is_external_ctap2_securitykey_supported() -> Result<::js_sys::Promise, JsValue>;
}
```

crates/web-sys/webidls/enabled/WebAuthentication.webidl
```diff
@@ -21,6 +21,11 @@ partial interface PublicKeyCredential {
     static Promise<boolean> isUserVerifyingPlatformAuthenticatorAvailable();
 };
 
+[SecureContext]
+partial interface PublicKeyCredential {
+    static Promise<boolean> isConditionalMediationAvailable();
+};
```
daxpedda, line 27: This has to go into the unstable folder.

daxpedda (changes requested):

Liamolucko:
> Maybe it would make sense to add it as a deprecated API. You can use `[Deprecated]` for that, search for it in the existing WebIDL and you will find plenty of examples.

`[Deprecated]` doesn't actually do anything with our WebIDL generator, those are all left over from when our WebIDL was imported from Gecko. We have `[RustDeprecated = "<deprecation message>"]` instead.

crates/web-sys/webidls/enabled/WebAuthentication.webidl
daxpedda: @Firstyear the way to move it to the unstable folder is to actually move the WebIDL to the [unstable folder](https://github.com/rustwasm/wasm-bindgen/tree/165ad00ccff4bc67705afdacecf894cb8250ee2e/crates/web-sys/webidls/unstable).

daxpedda:
Before I forget, could you also add a changelog entry?

Firstyear:
Changelog added, rebased to main, squashed commits.

crates/web-sys/src/features/gen_PublicKeyCredential.rs
```diff
@@ -37,11 +37,34 @@ extern "C" {
...
+    #[doc = ""]
+    #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/isConditionalMediationAvailable)"]
+    #[doc = ""]
+    #[doc = "*This API requires the following crate features to be activated: `PublicKeyCredential`*"]
+    pub fn is_conditional_mediation_available() -> ::js_sys::Promise;
+    # [wasm_bindgen (static_method_of = PublicKeyCredential , js_class = "PublicKeyCredential" , js_name = isExternalCTAP2SecurityKeySupported, catch)]
+    #[doc = "The `isExternalCTAP2SecurityKeySupported()` method."]
+    #[doc = ""]
+    #[doc = "*This API requires the following crate features to be activated: `PublicKeyCredential`*"]
+    pub fn is_external_ctap2_securitykey_supported() -> Result<::js_sys::Promise, JsValue>;
```
daxpedda, line 69: The point of the whole WebIDL generation is so you don't write it by hand. Adding to this interface manually is something we don't support.

Your WebIDL for `isConditionalMediationAvailable` works just fine and was added by the generator just above. If you want to add `isExternalCTAP2SecurityKeySupported`, you should add it via WebIDL as well.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 
 ### Added
 
+* Add bindings for `PublicKeyCredential.isConditionalMediationAvailable`.
```
```suggestion
* Add bindings for `PublicKeyCredential.isConditionalMediationAvailable()`.
```

daxpedda (changes requested):

crates/web-sys/src/features/gen_PublicKeyCredential.rs
Firstyear: I'm not going to add it? I thought I removed it from the PR.

daxpedda: You should follow [this guide](https://github.com/rustwasm/wasm-bindgen/blob/main/guide/src/contributing/web-sys/supporting-more-web-apis.md) to add the APIs. Currently you have modified the `crates/web-sys/src/features/gen_PublicKeyCredential.rs` file by hand; you shouldn't touch these files at all.

Just rerunning the WebIDL generator should fix the issue.

Firstyear: I don't understand what you mean here. I think this is getting too hard for me, and it's consuming a lot of my time to make little progress sorry.

Issue #3297, opened 2023-02-12
[BUG] only C-Style enums allowed with #[wasm_bindgen] / #[cfg] field not removed


erwanvivien:
### Describe the Bug

Cannot use a `#[cfg]` macro inside a `#[wasm_bindgen]` enum ?

### Steps to Reproduce

```bash
cargo new test-wasm-enum --lib
cd test-wasm-enum
```
replace lib.rs with:
```rust
#[cfg(target_arch = "wasm32")]
use wasm_bindgen::prelude::*;

#[cfg_attr(target_arch = "wasm32", wasm_bindgen, repr(C))]
#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
pub enum Operation {
    INCREMENT,
    DECREMENT,
    #[cfg(not(target_arch = "wasm32"))]
    CUSTOM(fn(i32) -> i32),
}

#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]
pub fn apply(op: Operation, x: i32) -> i32 {
    match op {
        Operation::INCREMENT => x + 1,
        Operation::DECREMENT => x - 1,
        #[cfg(not(target_arch = "wasm32"))]
        Operation::CUSTOM(f) => f(x),
    }
}
```
Then:
```bash
wasm-pack build
```

### Expected Behavior

No errors

### Actual Behavior

```
test-wasm $ wasm-pack build -t web --release
[INFO]: Checking for the Wasm target...
[INFO]: Compiling to Wasm...
   Compiling proc-macro2 v1.0.51
   Compiling quote v1.0.23
   Compiling syn v1.0.107
   Compiling wasm-bindgen-shared v0.2.84
   Compiling log v0.4.17
   Compiling wasm-bindgen-backend v0.2.84
   Compiling wasm-bindgen-macro-support v0.2.84
   Compiling wasm-bindgen-macro v0.2.84
   Compiling wasm-bindgen v0.2.84
   Compiling test-wasm v0.1.0 (/mnt/c/Users/Erwan/SynologyDrive/development/rust/test-wasm)
error: only C-Style enums allowed with #[wasm_bindgen]
  --> src/lib.rs:10:11
   |
10 |     CUSTOM(fn(i32) -> i32),
   |           ^^^^^^^^^^^^^^^^

error[E0433]: failed to resolve: use of undeclared type `Operation`
  --> src/lib.rs:16:9
   |
[...]

PR #3496, opened 2023-06-25
fix(#3297): Allow variant skip

erwanvivien:
Provides a way to avoid having `only C-Style enums allowed with #[wasm_bindgen]` by allowing user to omit an enum Variant, it's then at the user discretion to use #[cfg] directives to reflect this

Can be seen in the example:

```rust
#[wasm_bindgen]
pub enum Operation {
    INCREMENT,

    #[cfg(not(target_arch = "wasm32"))]
    #[wasm_bindgen(skip)]
    CUSTOM(fn(i32) -> i32),
}
```


erwanvivien:
As discussed with Daxpedda on Discord, this is not a good idea.

PR #2759, opened 2022-01-07
Remove unnecessary async import in hello world example

billhannah:
In Webpack 5, wasm-based imports are now supported, and can be imported directly

PR #3503, opened 2023-07-03
docs: add and hello_world switched to direct import

Stefan-Hanke:
If this PR is merged, it will update examples to use static import instead of dynamic ones (see #2759)

I just switched the **hello_world** and **add** examples, if ok I could just as well run through the other examples as well.

As a sidenote, npm is complaining about vulnerabilities due to the use of an older version of **webpack-dev-server** - bumping this to `4.15.1` fixes it. I'd just commit that also?

### DIFF
--- a/examples/add/index.js
+++ b/examples/add/index.js
@@ -1,6 +1,5 @@
 // For more comments about what's going on here, check out the `hello_world`
 // example
-const rust = import('./pkg');
-rust
-  .then(m => alert('1 + 2 = ' + m.add(1, 2)))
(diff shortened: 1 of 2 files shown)


daxpedda:
> As a sidenote, npm is complaining about vulnerabilities due to the use of an older version of **webpack-dev-server** - bumping this to `4.15.1` fixes it. I'd just commit that also?

Let's do that in a separate PR.

daxpedda (approved):
Thanks!

PR #3504, opened 2023-07-04
Add async variants of Atomics.wait

allsey87:
`Atomics.waitAsync` now has reasonable [support in most browsers](https://caniuse.com/mdn-javascript_builtins_atomics_waitasync) with Firefox being the notable exception, however they have a [bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1467846) for implementing it.

I used the binding for `Atomics.wait` as a reference and one thing that seems off to me is that the `typed_array` argument for the bigint varieties of these functions take an `Int32Array`, instead of a `BigInt64Array`/`BigUint64Array`. It seems to me that the most correct thing to do here is two make further variants of this function, e.g., add the `*_biguint` variants, which also take u64 as their value argument. If there is a consensus on this, I can add the variants and we can rename this PR to something like "Clean up Atomics.wait and Atomics.waitAsync".

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1209,7 +1209,7 @@ pub mod Atomics {
         /// You should use `wait_async_bigint` to operate on a `BigInt64Array` or a `BigUint64Array`.
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/waitAsync)
-        #[wasm_bindgen(js_namespace = Atomics, catch)]
+        #[wasm_bindgen(js_namespace = Atomics, catch, js_name = waitAsync)]
         pub fn wait_async(typed_array: &Int32Array, index: u32, value: i32) -> Result<Object, JsValue>;
 
         /// The static `Atomics.waitAsync()` method verifies that a given position in an
@@ -1224,7 +1224,7 @@ pub mod Atomics {
         /// You should use `wait_async_bigint` to operate on a `BigInt64Array` or a `BigUint64Array`.
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/waitAsync)
-        #[wasm_bindgen(js_namespace = Atomics, catch, js_name = wait)]
+        #[wasm_bindgen(js_namespace = Atomics, catch, js_name = waitAsync)]
         pub fn wait_async_bigint(
             typed_array: &Int32Array,
             index: u32,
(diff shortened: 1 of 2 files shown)


allsey87:
Actually, at least according to the docs that I can find, `BigUintArray` is not supported, only `BigIntArray`.

crates/js-sys/src/lib.rs
```diff
@@ -1197,6 +1197,70 @@ pub mod Atomics {
...
+        /// whether equal to either "not-equal" or "timed-out".
+        /// Note: This operation only works with a shared `Int32Array` and may be used
+        /// on the main thread.
+        ///
+        /// You should use `wait_async_bigint` to operate on a `BigInt64Array` or a `BigUint64Array`.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/waitAsync)
+        #[wasm_bindgen(js_namespace = Atomics, catch, js_name = waitAsync)]
+        pub fn wait_async_bigint(
+            typed_array: &Int32Array,
```
daxpedda, line 1233:
```suggestion
            typed_array: &BigInt64Array,
```

crates/js-sys/src/lib.rs
```diff
@@ -1197,6 +1197,70 @@ pub mod Atomics {
...
+        ) -> Result<Object, JsValue>;
+
+        /// Like `waitAsync()`, but with timeout
+        ///
+        /// This method is used to operate on a `BigInt64Array` or a `BigUint64Array`.
+        ///
+        /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/waitAsync)
+        #[wasm_bindgen(js_namespace = Atomics, catch, js_name = waitAsync)]
+        pub fn wait_async_with_timeout_bigint(
+            typed_array: &Int32Array,
```
daxpedda, line 1258:
```suggestion
            typed_array: &BigInt64Array,
```

daxpedda (changes requested):
I don't know what happened here, but I would argue that the current `wait_bigint` functions are just wrong, they should take a `BigInt64Array` instead of a `Int32Array`.

But I think we don't need to add to that any further.
Also we could definitely remove the documentation around `BigUint64Array`.

daxpedda (approved):
Thanks!

daxpedda:
I was thinking about using this here:
https://github.com/rustwasm/wasm-bindgen/blob/29205c51b0d617ec6fef17df859e7ac4429746bb/crates/futures/src/task/multithread.rs#L189-L190

But it wouldn't help much because we need the binding to detect support anyway. We also want to have the custom return type to avoid the overhead of `Reflect`, something we won't add in `js-sys` because we don't do this in general there.

allsey87:
> But it wouldn't help much because we need the binding to detect support anyway. We also want to have the custom return type to avoid the overhead of `Reflect`, something we won't add in `js-sys` because we don't do this in general there.

Yep, I was thinking that/stumbled across that when I was checking whether `waitAsync` was already implemented.

Regarding the non-async variants of `wait`, shall I put together a PR that fixes `type_array` for the `_bigint` variants? It would be a breaking change, although I guess no one is using this API...

hamza1311:
I'm fine with such breaking changes. It won't be the first time we made breaking changes to API that no one is/should be using ([as an example, removing interface types comes to mind](https://github.com/rustwasm/wasm-bindgen/pull/3372))

PR #3508, opened 2023-07-05
chore: run audit on examples

Stefan-Hanke:
If this PR is merged, it will update and cleanup the `package.json` files in the examples.

More specifically, I ran `for d in $(ls); do cd $d; npm i; npm audit fix --force; cd -; done` from the `examples` folder.

The command not only updated `webpack-dev-server` but also seemed to have reordered the entries (sorting them alphabetically) and did a minor reformatting (dropped a superfluous whitespace), but that looks ok IMO.

I did run `npm run build` followed by `npm run serve` on the `add` example to verify that it's still working as expected.

### DIFF
--- a/examples/add/package.json
+++ b/examples/add/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "build": "webpack",
-    "serve" : "webpack serve"
+    "serve": "webpack serve"
   },
   "devDependencies": {
     "@wasm-tool/wasm-pack-plugin": "1.5.0",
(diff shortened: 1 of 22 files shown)


Stefan-Hanke:
Fortunately the examples are built from the [`build_examples` job](https://github.com/rustwasm/wasm-bindgen/blob/main/.github/workflows/main.yml#L261).

~~Unfortunately those `grep -v` stuff seemed to not have been updated in lieu with changes of the folder names, e.g. there is also a `grep -v raytrace` but the directory is called `raytrace-parallel`~~ oh well it's a grep so it runs fine. It just looks a bit odd...

daxpedda (approved):
Thanks, it's appreciated!

PR #3511, opened 2023-07-07
Replace curl package into ureq

kerryeon:
When installing `wasm-bindgen-cli` in Alpine Linux, it's forced to install `openssl-sys` package.

But I found that the `openssl` dependency is only subject to `curl` package, which is used for headless operations.

So I think it's ok to adopt `rustls` instead, not installing the native openssl binaries by option.

And I replaced the old `curl` package into `ureq`, using `rustls` by default.

Therefore, this PR is a `rustls` support for `wasm-bindgen-cli`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,6 @@
   [#3480](https://github.com/rustwasm/wasm-bindgen/pull/3480)
 
 * Replaced `curl` with `ureq`. By default we now use Rustls instead of OpenSSL.
-  The `openssl` is now out of the dependencies in `wasm-bindgen-cli` crates, so that
-  the `vendored-openssl` feature is now out of supported features.
   [#3511](https://github.com/rustwasm/wasm-bindgen/pull/3511)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@
   because required arguments occurring after optional arguments are currently not supported by the generator.
   [#3480](https://github.com/rustwasm/wasm-bindgen/pull/3480)
 
+* Replaced `curl` package into `ureq`.
+  [#3511](https://github.com/rustwasm/wasm-bindgen/pull/3511)
+
 ### Fixed
 
 * Fixed bindings and comments for `Atomics.wait`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
   [#3480](https://github.com/rustwasm/wasm-bindgen/pull/3480)
 
 * Replaced `curl` with `ureq`. By default we now use Rustls instead of OpenSSL.
+  The `openssl` is now out of the dependencies in `wasm-bindgen-cli` crates, so that
+  the `vendored-openssl` feature is now out of supported features.
   [#3511](https://github.com/rustwasm/wasm-bindgen/pull/3511)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
   because required arguments occurring after optional arguments are currently not supported by the generator.
   [#3480](https://github.com/rustwasm/wasm-bindgen/pull/3480)
 
-* Replaced `curl` package into `ureq`.
+* Replaced `curl` with `ureq`. By default we now use Rustls instead of OpenSSL.
   [#3511](https://github.com/rustwasm/wasm-bindgen/pull/3511)
 
 ### Fixed
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -20,12 +20,15 @@ docopt = "1.0"
 env_logger = "0.8"
 anyhow = "1.0"
 log = "0.4"
-openssl = { version = '0.10.11', optional = true }
+native-tls = { version = "0.2", default-features = false, optional = true }
 rouille = { version = "3.0.0", default-features = false }
 serde = { version = "1.0", features = ['derive'] }
 serde_derive = "1.0"
 serde_json = "1.0"
-ureq = { version = "2.7", default-features = false, features = ["brotli", "gzip"] }
+ureq = { version = "2.7", default-features = false, features = [
+    "brotli",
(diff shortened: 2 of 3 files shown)


hamza1311 (approved):
This is a good solution in the meantime but, ultimately we should just replace curl with ureq.

I'm afraid that removing the features in the future could end being a breaking change

daxpedda:
@kerryeon would you be willing to rip out cURL entirely and just switch to [`ureq`](https://crates.io/crates/ureq)?
I would prefer that we fix this now instead of introducing a bunch of crate features we are going to deprecate anyway.

A note in the changelog would also be appreciated.

kerryeon:
Ok, let me dig in this migration job and notify it.

kerryeon:
@daxpedda Hello, please review when you have time :)

hamza1311 (approved):
Looks good to me. I'll merge this once the CI is finished

crates/cli/Cargo.toml
```diff
@@ -26,6 +25,7 @@ rouille = { version = "3.0.0", default-features = false }
 serde = { version = "1.0", features = ['derive'] }
 serde_derive = "1.0"
 serde_json = "1.0"
+ureq = { version = "2.7", default-features = false, features = ["gzip"] }
```
daxpedda, line 28:
```suggestion
ureq = { version = "2.7", default-features = false, features = ["brotli", "gzip"] }
```
Probably irrelevant, but why not while we are there.

crates/cli/Cargo.toml
```diff
@@ -44,4 +44,13 @@ name = "reference"
...
 
 [features]
-vendored-openssl = ['openssl/vendored']
+default = ["rustls-tls"]
+
+rustls-tls = ["ureq/tls"]
+
+# Enables native-tls specific functionality not available by default.
+native-tls = ["ureq/native-tls"]
+native-tls-vendored = ["native-tls", "openssl/vendored"]
```
daxpedda, line 53:
```suggestion
native-tls-vendored = ["native-tls", "native-tls/vendored"]
```
Unfortunately `ureq` doesn't re-expose this crate feature, but we should probably go as high as possible and not as low as possible. So let's just replace the `openssl` dependency with `native-tls`.

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@
   because required arguments occurring after optional arguments are currently not supported by the generator.
   [#3480](https://github.com/rustwasm/wasm-bindgen/pull/3480)
 
+* Replaced `curl` package into `ureq`.
```
daxpedda, line 33:
```suggestion
* Replaced `curl` with `ureq`. By default we now use Rustls instead of OpenSSL.
```

daxpedda (changes requested):
Thank you, a couple of nits here and there.

crates/cli/Cargo.toml
kerryeon: Agreed. When not turning on this feature, it is possible that we received an **un**decoded binary value as a HTTP response instead of an encoding failure error.

crates/cli/Cargo.toml
kerryeon: Great :)

daxpedda: In https://github.com/rustwasm/wasm-bindgen/pull/3511/commits/4cc4b2ee10fb2a0d02abbff1ce75d6cbec0617ec you removed the `vendored-openssl` crate feature, you can't really do that, that's a breaking change.

kerryeon: OK, I reverted the `vendored-openssl` feature.

kerryeon: oh wait a minute

kerryeon: Replaced two packages (features) like below.

1. `openssl` => `native-tls`
2. `vendored-openssl` => `openssl` + `native-tls/vendored`

crates/cli/Cargo.toml
```diff
@@ -44,4 +47,11 @@ name = "reference"
 harness = false
 
 [features]
-vendored-openssl = ['openssl/vendored']
+default = ["rustls-tls"]
+
+native-tls = ["dep:native-tls", "ureq/native-tls"]
+rustls-tls = ["ureq/tls"]                          # ureq uses rustls by default.
```
daxpedda, line 53:
```suggestion
rustls-tls = ["ureq/tls"]
```

crates/cli/Cargo.toml
```diff
@@ -44,4 +47,11 @@ name = "reference"
...
+
+# Legacy support
+openssl = ["native-tls"]
```
daxpedda, line 56:
```suggestion
openssl = ["dep:native-tls"]
```
We can have the `dep` here so we don't need it above, this is all legacy features anyway.

crates/cli/Cargo.toml
```diff
@@ -44,4 +47,11 @@ name = "reference"
 harness = false
 
 [features]
-vendored-openssl = ['openssl/vendored']
+default = ["rustls-tls"]
+
+native-tls = ["dep:native-tls", "ureq/native-tls"]
```
daxpedda, line 52:
```suggestion
native-tls = ["ureq/native-tls"]
```

CHANGELOG.md
```diff
@@ -30,6 +30,11 @@
   because required arguments occurring after optional arguments are currently not supported by the generator.
   [#3480](https://github.com/rustwasm/wasm-bindgen/pull/3480)
 
+* Replaced `curl` with `ureq`. By default we now use Rustls instead of OpenSSL.
+  The `openssl` is now out of the dependencies in `wasm-bindgen-cli` crates, so that
+  the `vendored-openssl` feature is now out of supported features.
```
daxpedda, line 35:
```suggestion
```
We don't need to communicate that, it's basically internal.

daxpedda (changes requested):
Small nits.

kerryeon:
Confirmed!

daxpedda (approved):
Thank you!

PR #3535, opened 2023-07-27
Add missing CSS loaders for todomvc app

### LINKED ISSUE #3443: CSS missing in compiled todomvc example
On both https://rustwasm.github.io/wasm-bindgen/exbuild/todomvc/ and https://wasm-bindgen.netlify.app/exbuild/todomvc/, the app logic works, but the style is broken, because the CSS is 404 Not Found.

snOm3ad:
Currently webpack does not bundle the `index.css` file located in the root directory. This PR adds the corresponding plugins and loaders to ensure the file is correctly bundled.

Fixes #3443

### DIFF
--- a/_package.json
+++ b/_package.json
@@ -5,7 +5,9 @@
   },
   "devDependencies": {
     "@wasm-tool/wasm-pack-plugin": "1.5.0",
+    "css-loader": "^6.8.1",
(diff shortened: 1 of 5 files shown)


Liamolucko (approved):
I don't know a whole lot about Webpack, but this all looks good to me. Thanks!

PR #3537, opened 2023-07-28
Changed `write` functions of FileSystemSyncAccessHandle to accept immutable buffers

### LINKED ISSUE #3536: Mutable slice passed for `FileSystemSyncAccessHandle` 's write functions in web_sys
### Describe the Bug

The type `FileSystemSyncAccessHandle` of the web_sys crate has two `write` functions which can take a `u8` slice to write to the handle. However, the slice has to be passed as mutable, which can be a problem if you want to integrate it with any kind of `Write` trait from std or other synchronous or asynchronous crates, which only provide an immutable reference to the slice of data to write.

### Steps to Reproduce

N/A

### Expected Behavior

I expect the write functions `write_with_u8_array` and `write_with_u8_array_and_options` to take an immutable reference for the slice of data to write.

### Actual Behavior

The write functions `write_with_u8_array` and `write_with_u8_array_and_options` only accept mutable slices.

### Additional Context

N/A

kipcode66:
Removed unnecessary `mut` for the write functions.

According to the MDN documentation, there should not be any modification applied to the buffer, so this should be completely fine.

Should resolve #3536

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@
 * Replaced `curl` with `ureq`. By default we now use Rustls instead of OpenSSL.
   [#3511](https://github.com/rustwasm/wasm-bindgen/pull/3511)
 
+* Changed mutability of the argument `buffer` in `write` functions to immutable for `FileSystemSyncAccessHandle` and `FileSystemWritableFileStream`.
+  It was also automatically changed for `IdbFileHandle`, which is deprecated.
+  [#3537](https://github.com/rustwasm/wasm-bindgen/pull/3537)
+
 ### Fixed
 
 * Fixed bindings and comments for `Atomics.wait`.
--- a/crates/web-sys/src/features/gen_FileSystemSyncAccessHandle.rs
+++ b/crates/web-sys/src/features/gen_FileSystemSyncAccessHandle.rs
@@ -163,7 +163,7 @@ extern "C" {
     #[doc = "[described in the `wasm-bindgen` guide](https://rustwasm.github.io/docs/wasm-bindgen/web-sys/unstable-apis.html)*"]
     pub fn write_with_u8_array(
         this: &FileSystemSyncAccessHandle,
-        buffer: &mut [u8],
+        buffer: &[u8],
     ) -> Result<f64, JsValue>;
     #[cfg(web_sys_unstable_apis)]
     #[cfg(feature = "FileSystemReadWriteOptions")]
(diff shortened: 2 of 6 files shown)


snOm3ad:
You might want to look at this [section](https://rustwasm.github.io/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html) of the `wasm-bindgen` guide. These files are generated by the crate `webidl` so it's not possible to modify them manually, you need to regenerate them.

The part that is of interest is here:

https://github.com/rustwasm/wasm-bindgen/blob/b76af3b6d0342231289cbebb6830a76e64be48a7/crates/webidl/src/util.rs#L511-L534

By default slice references that get passed to JS are mutable. In order to make them immutable you need to whitelist them and then regenerate the API.

snOm3ad:
LGTM! I'm not in the contributors list so I can't merge your PR, but maybe @daxpedda can.

kipcode66:
I noted that when I added this fix, it also modified the methods for IDBFileHandle, but when I go to MDN, the documentation for that class doesn't seem to exist anymore. Since it's not even available anyway, I don' think this should be an issue.

My belief is that it should probably be removed if it's not even listed in MDN.

hamza1311:
This is a breaking change ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ac901ed65a66ec084e0499180e5a2f99))

kipcode66:
It might be a breaking change, but it's in the unstable API (which from what I can understand, when an API is unstable, we should expect breaking changes if necessary).

In my specific use case, I am working with files that have sizes larger than I can load directly on the available RAM in a browser. In order to work on those files, I'm directly loading and saving into files after processing them in chunks. However, the chunks can still be quite large, and I already encountered files which would require chunks that, if I needed to allocate an intermediary buffer each time to change the mutability, would overflow the available memory.

I believe I could import a function that would use the right signature myself, but the cleaner solution would be to fix the API.

snOm3ad:
> This is a breaking change ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ac901ed65a66ec084e0499180e5a2f99))

This isn't the case though, from client side the reference is already required to be `mut`. There's no code out there that uses this API with an immutable reference.

```rust
fn original(b: &mut [u8]);
fn proposed(b: &[u8]);
```

Calling `proposed` like `proposed(&mut [1, 2, 3])` does not produce an error ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c4ed8773290fbb5690f196df8051ed7e)). What @kipcode66 is doing is moving the API from a stricter version into a less strict version.

daxpedda (changes requested):
Apologies for my absence.

I believe this is technically still a breaking change because it can mess up inference, but I consider this a very minor breaking change.
It is also true that this is an experimental API where a breaking change is not an issue, but unfortunately the whitelisting also affected `IdbFileHandle.write`, which isn't an experimental API, which is funny because this API is deprecated anyway.

All in all I'm happy to merge this.
Small nit: please add an entry to the changelog.

@hamza1311 I would appreciate your approval as well.

hamza1311 (approved):
I had missed that it's part of unstable API as part of my previous review, sorry about that. This looks good to me.

PR #3541, opened 2023-08-02
Handle raw identifiers in wasm_bindgen_test macro

### LINKED ISSUE #3288: wasm-bindgen-test macro panics when a function name uses raw ident
### Describe the Bug
`#[wasm_bindgen_test]` macro panics if a test function uses raw ident as the function name.

### Steps to Reproduce
Compile the below code.
```rust
#[wasm_bindgen_test]
fn r#if() {}
```

### Expected Behavior
Generates a valid identifier by trimming the raw identifier prefix.

### Actual Behavior
```
error: custom attribute panicked
   --> crates/parser2/tests/syntax_node.rs:106:1
    |
106 | #[wasm_bindgen_test]
    | ^^^^^^^^^^^^^^^^^^^^
    |
    = help: message: `"__wbgt_r#if_0"` is not a valid identifier
```

snOm3ad:
Currently the use of raw identifiers in test names produces an invalid function name. `Ident::new` panics whenever it encounters the `r#` prefix, instead `Ident::new_raw` should be used.

However, the `__wbgt` prefix we append forms a valid identifier already. So removing the raw ident prefix ("r#") is the only thing required to create a valid function name.

Fixes #3288

### DIFF
--- a/crates/test-macro/src/lib.rs
+++ b/crates/test-macro/src/lib.rs
@@ -89,7 +89,10 @@ pub fn wasm_bindgen_test(
     // We generate a `#[no_mangle]` with a known prefix so the test harness can
     // later slurp up all of these functions and pass them as arguments to the
     // main test harness. This is the entry point for all tests.
-    let name = format!("__wbgt_{}_{}", ident, CNT.fetch_add(1, Ordering::SeqCst));
+    let internal_id = ident.to_string()
+        .strip_prefix("r#")
+        .map_or(ident.clone(), |s| Ident::new(s, Span::call_site()));
+    let name = format!("__wbgt_{}_{}", internal_id, CNT.fetch_add(1, Ordering::SeqCst));
     let name = Ident::new(&name, Span::call_site());
     tokens.extend(
         (quote! {
--- a/crates/test-macro/src/lib.rs
+++ b/crates/test-macro/src/lib.rs
@@ -4,6 +4,7 @@
 extern crate proc_macro;
 
 use proc_macro2::*;
+use quote::format_ident;
 use quote::quote;
 use quote::quote_spanned;
 use std::sync::atomic::*;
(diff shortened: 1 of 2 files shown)


crates/test-macro/src/lib.rs
```diff
@@ -89,7 +89,15 @@ pub fn wasm_bindgen_test(
...
-    let name = format!("__wbgt_{}_{}", ident, CNT.fetch_add(1, Ordering::SeqCst));
+    let internal_id = ident
+        .to_string()
+        .strip_prefix("r#")
+        .map_or(ident.clone(), |s| Ident::new(s, Span::call_site()));
+    let name = format!(
+        "__wbgt_{}_{}",
+        internal_id,
+        CNT.fetch_add(1, Ordering::SeqCst)
+    );
```
daxpedda, line 100:
```suggestion
    let name = format!("__wbgt_{}_{}", ident.unraw(), CNT.fetch_add(1, Ordering::SeqCst));
```

See https://docs.rs/syn/2.0.28/syn/struct.Ident.html#method.unraw.

daxpedda (changes requested):

crates/test-macro/src/lib.rs
snOm3ad: Oh that's wonderful! Will update soon!

snOm3ad: @daxpedda unfortunately what we get back is a `proc_macro2::Ident` not a `syn::Ident`. There's no `unraw` method for the `proc_macro2` [version](https://docs.rs/proc-macro2/1.0.66/proc_macro2/struct.Ident.html), and `syn` is not a dependency for the `test-macro` crate so there's no way to convert between the two.

hamza1311: There's also [`format_ident!`](https://docs.rs/quote/latest/quote/macro.format_ident.html)

daxpedda: Hm, could you just pull in `syn`? It's in the dependency tree anyway.

daxpedda: > There's also [`format_ident!`](https://docs.rs/quote/latest/quote/macro.format_ident.html)

That's also a good point, we can just use that instead of the new `Ident` we create in the next line.

snOm3ad: Thank you both!

hamza1311 (approved):
Thanks!

daxpedda (approved):
Thanks!

Issue #3542, opened 2023-08-02
feat: add `id` to RtcDataChannel bindings


DougAnderson444:
https://github.com/rustwasm/wasm-bindgen/blob/b76af3b6d0342231289cbebb6830a76e64be48a7/crates/web-sys/src/features/gen_RtcDataChannel.rs#L21

I see `label` is here, could we also add the `id` property to `RtcDataChannel` ? 

https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/id

daxpedda:
Yes, this is quite straightforward to accomplish, see [the `wasm-bindgen` Book](https://github.com/rustwasm/wasm-bindgen/blob/main/guide/src/contributing/web-sys/supporting-more-web-apis.md) on how to do just that.
Let me know if you need any help.

DougAnderson444:
Excellent direction, thanks I'll do that. 
I take it since the `id` is between `0` and `65,534` would be an `unsigned long`?

```
readonly attribute unsigned long id;
```

I didn't see the definition for the difference between `unsigned long` and `unsigned short` anywhere

DougAnderson444:
I figured out that `unsigned short` compiles to `u16` which is the `65,534` we need

PR #3547, opened 2023-08-04
feat: add RTCDataChannel id

DougAnderson444:
Adds [`RTCDataChannel` readonly attribute `id`](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/id)

Set as type `unsigned short` compiles to `u16` which is the size of `id` (`0` to `65,534`)

fixes https://github.com/rustwasm/wasm-bindgen/issues/3542

### DIFF
--- a/crates/web-sys/src/features/gen_RtcDataChannel.rs
+++ b/crates/web-sys/src/features/gen_RtcDataChannel.rs
@@ -25,7 +25,7 @@ extern "C" {
     #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/id)"]
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `RtcDataChannel`*"]
-    pub fn id(this: &RtcDataChannel) -> u16;
+    pub fn id(this: &RtcDataChannel) -> Option<u16>;
     # [wasm_bindgen (structural , method , getter , js_class = "RTCDataChannel" , js_name = reliable)]
     #[doc = "Getter for the `reliable` field of this object."]
     #[doc = ""]
--- a/crates/web-sys/src/features/gen_RtcDataChannel.rs
+++ b/crates/web-sys/src/features/gen_RtcDataChannel.rs
@@ -19,6 +19,13 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `RtcDataChannel`*"]
     pub fn label(this: &RtcDataChannel) -> String;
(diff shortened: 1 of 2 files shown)


DougAnderson444:
I don't understand what I did wrong to upset the `git diff` step, all the generated code looks fine

DougAnderson444:
When I run the `cargo run --release --package wasm-bindgen-webidl -- webidls src/features ./Cargo.toml` command (Windows 10), I do get a rust fmt error:

```bash
Error: rustfmt failed

Caused by:
    The filename or extension is too long. (os error 206)
```

daxpedda (changes requested):
You should never modify the generated files yourself, only modify the WebIDL and then let the generator run.

I'm not sure what the issue is you are running into when running the WebIDL generator, it might be Windows specific, unfortunately I don't have access to a Windows machine right now. It would actually be nice if you can look into it and potentially fix it, but that's a matter for a separate PR.

Feel free to ping me if you don't get the generator to work and I can run it for you.

DougAnderson444:
> You should never modify the generated files yourself, only modify the WebIDL and then let the generator run.

I thought I did that :/ 

> I'm not sure what the issue is you are running into when running the WebIDL generator, it might be Windows specific, 

I'm not sure about this being a Windows issue anymore, because I just committed your suggestion, CI ran the generator, and `git diff --exit-code` still didn't pass. I see the generated code changed from `u16` to `Option<u16>` so that looks right.  So that shouldn't have anything to do with Windows, right?

> Feel free to ping me if you don't get the generator to work and I can run it for you.

Yes please do, I'm getting frustrated as I don't really know what to try next. I didn't think adding one line would be so involved! 😐

daxpedda:
> > I'm not sure what the issue is you are running into when running the WebIDL generator, it might be Windows specific,
> 
> I'm not sure about this being a Windows issue anymore, because I just committed your suggestion, CI ran the generator, and `git diff --exit-code` still didn't pass. I see the generated code changed from `u16` to `Option<u16>` so that looks right. So that shouldn't have anything to do with Windows, right?

Ah, apologies, I thought you did it by hand instead of running the generator.
I think the generator has a different output on Windows actually ... so looks like a bug to me.

> > Feel free to ping me if you don't get the generator to work and I can run it for you.
> 
> Yes please do, I'm getting frustrated as I don't really know what to try next. I didn't think adding one line would be so involved! neutral_face

No problem, it seems the generator runs correctly for me. Don't let this stop you from contributing though, I'm happy to fix this up in the future as well :wink:.

DougAnderson444:
Oh for sure I'm happy to help with open source, I love it 😁

Thanks for all your help

PR #3549, opened 2023-08-06
use fully qualified names in `wasm_bindgen_test` for items exported from std prelude

hlbarber:
## Motivation

Currently, the `wasm_bindgen_test` macro dooes not use fully qualified names for items exported from the prelude. This means that naming collisions (`struct Some`) or the removal of the std library prelude will cause compilation errors.

## Solution

- Identify and use fully qualified names in macros were we previously assumed the Rust std prelude. We use `::core` rather than `::std`.
- Add [`no_implicit_prelude`](https://doc.rust-lang.org/reference/names/preludes.html#the-no_implicit_prelude-attribute) to `crates/test-macro/ui-tests/should_panic.rs`.

### DIFF
--- a/crates/test-macro/src/lib.rs
+++ b/crates/test-macro/src/lib.rs
@@ -76,9 +76,11 @@ pub fn wasm_bindgen_test(
     let mut tokens = Vec::<TokenTree>::new();
 
     let should_panic = match should_panic {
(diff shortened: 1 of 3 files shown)


hamza1311 (approved):
Thank you for the PR! Great to see macro hygiene improved

PR #3554, opened 2023-08-10
Allow creating `Vec`s of and implement `TryFrom<JsValue>` for strings and exported Rust types

### LINKED ISSUE #111: Allow returning Vec<T>

### LINKED ISSUE #168: Ability to return a Vec<String>
Annotating a function which returns a `Vec<String>` with `#[wasm_bindgen]` gives the following error:

```
error[E0277]: the trait bound `std::boxed::Box<[std::string::String]>: wasm_bindgen::convert::IntoWasmAbi` is not satisfied
  --> src/lib.rs:12:1
   |
12 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::IntoWasmAbi` is not implemented for `std::boxed::Box<[std::string::String]>`
   |
   = help: the following implementations were found:
             <std::boxed::Box<[f64]> as wasm_bindgen::convert::IntoWasmAbi>
             <std::boxed::Box<[u32]> as wasm_bindgen::convert::IntoWasmAbi>
             <std::boxed::Box<[i16]> as wasm_bindgen::convert::IntoWasmAbi>
             <std::boxed::Box<[u16]> as wasm_bindgen::convert::IntoWasmAbi>
           and 5 others
   = note: required because of the requirements on the impl of `wasm_bindgen::convert::IntoWasmAbi` for `std::vec::Vec<std::string::String>`

```

### LINKED ISSUE #2231: Downcast a JsValue to an exported struct

### LINKED ISSUE #2452: Accept `Vec<T>` as function arguments
It would be great to accept vectors of objects (especially vectors of objects created by wasm bindgen itself). I believe that the workaround right now is to create a builder that has a method to add a new `T`, which one has to call for each entry in the vector. 

_Follow up for comment by @fjarri in https://github.com/rustwasm/wasm-bindgen/issues/111#issuecomment-733981170_

See https://github.com/nucypher/rust-umbral/issues/23 for an example.

Liamolucko:
This is a rebase of #2734 along with some fixes and tweaks. Sorry it took me so long to get to it!

This PR:
- Adds support for passing `Vec`s of Rust structs to/from JS.
- Adds support for passing `Vec<String>`s to/from JS.
- Implements `TryFrom<JsValue>` for Rust structs and strings.

The `Vec` impls are little inefficient, since they work by passing `Vec<JsValue>`s to/from JS like normal, and then translating them to the desired `Vec` with `TryFrom` / `Into<JsValue>`, which requires throwing away the first allocation and making a second one.

Fixes #111
Fixes #2452
Fixes #168
Fixes #2231

This also mostly supersedes #3088. That PR does support slightly more functionality (getting access to `&Struct` instead of forcing you to take ownership of it and invalidate JS's handle), but it also does so by exposing the internal `Anchor` type used by `RefFromWasmAbi`, which I'm not a huge fan of.

I haven't bumped the schema version, since the actual schema hasn't changed and so older versions of the CLI are mostly still compatible, although they do give this cryptic error message if you try and use the new `TryFrom<JsValue>` impl for Rust structs:

```
error: import of `__wbg_foo_unwrap` doesn't have an adapter listed
```

(where `foo` is the name of the struct.)

So maybe it's worth bumping it anyway, not sure.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -164,7 +164,7 @@ impl ToTokens for ast::Struct {
         let name = &self.rust_name;
         let name_str = self.js_name.to_string();
         let name_len = name_str.len() as u32;
-        let name_chars = name_str.chars().map(|c| c as u32);
+        let name_chars: Vec<u32> = name_str.chars().map(|c| c as u32).collect();
         let new_fn = Ident::new(&shared::new_function(&name_str), Span::call_site());
         let free_fn = Ident::new(&shared::free_function(&name_str), Span::call_site());
         let unwrap_fn = Ident::new(&shared::unwrap_function(&name_str), Span::call_site());
@@ -331,8 +331,11 @@ impl ToTokens for ast::Struct {
 
             impl #wasm_bindgen::describe::WasmDescribeVector for #name {
                 fn describe_vector() {
-                    use #wasm_bindgen::__rt::std::boxed::Box;
-                    <Box<[#wasm_bindgen::JsValue]> as #wasm_bindgen::describe::WasmDescribe>::describe();
+                    use #wasm_bindgen::describe::*;
+                    inform(VECTOR);
+                    inform(NAMED_EXTERNREF);
+                    inform(#name_len);
+                    #(inform(#name_chars);)*
                 }
             }
 
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -296,12 +296,12 @@ impl ToTokens for ast::Struct {
             }
 
             #[allow(clippy::all)]
-            impl wasm_bindgen::__rt::core::convert::TryFrom<wasm_bindgen::JsValue> for #name {
+            impl #wasm_bindgen::__rt::core::convert::TryFrom<#wasm_bindgen::JsValue> for #name {
                 type Error = ();
 
-                fn try_from(value: wasm_bindgen::JsValue)
-                    -> wasm_bindgen::__rt::std::result::Result<Self, Self::Error> {
-                    let js_ptr = wasm_bindgen::convert::IntoWasmAbi::into_abi(value);
+                fn try_from(value: #wasm_bindgen::JsValue)
+                    -> #wasm_bindgen::__rt::std::result::Result<Self, Self::Error> {
+                    let js_ptr = #wasm_bindgen::convert::IntoWasmAbi::into_abi(value);
 
                     #[link(wasm_import_module = "__wbindgen_placeholder__")]
                     #[cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))]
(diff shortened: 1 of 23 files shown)


daxpedda (commented):
I did a quick review and it LGTM.
Maybe I will find some time to make a proper review, but feel free to just merge if you are feeling confident about this.

Liamolucko:
> Does this allow returning `Vec<_>` or such from `extern`s? That's been a big limitation in my use of [ducktor](https://github.com/hamza1311/ducktor) for complicated types

If it's a `Vec` of `#[wasm_bindgen]`'d structs or `Strings, yes.

sinking-point:
Thank you @Liamolucko for sorting this out. It's so good to see my work being put to use.

cschramm:
I think this only partially fixes #111 and #168 (depending on their scopes).

Is it expected that it does not work for returning `Vec`s from *async* functions? For both `String` and `#[wasm_bindgen]`d structs I get an error that `From<Vec<_>>` is not implemented for `JsValue`.

Minimal example:
```rust
#[wasm_bindgen]
pub async fn f() -> Vec<String> {
    unimplemented!()
}
```

I can imagine that it might be an easy target, though.

Side node: *Passing* `Vec`s to `async fn`s works fine.

aspect:
As this came into `0.2.88` it just broke all our framework bindings - https://github.com/rustwasm/wasm-bindgen/issues/3685

TLDR; this implements a trait on types created by the developer who is using the wasm-bindgen framework, preventing the developer from creating his own `TryFrom<JsValue>` implementations, which can have many use cases.

PR #3557, opened 2023-08-11
web-sys: Add `HtmlElement` `inert` property

birktj:
This PR adds the `inert` property of `HtmlElement`. From what I can see it seems to be stable so I assume the fact that it is missing is simply an oversight.

See:
- https://html.spec.whatwg.org/multipage/interaction.html#inert
- https://html.spec.whatwg.org/multipage/dom.html#htmlelement
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert


hamza1311 (approved):
Thanks :+1:

PR #3561, opened 2023-08-17
Remove class wrap for constructors in Rust exports

### LINKED ISSUE #3213: Cannot inherit from structs created via `#[wasm_bindgen]` attribute as expected
### Describe the Bug
If I create a struct in rust and export the struct via `#[wasm_bindgen]` attribute, and if I then inherit this struct from javascript space, my Javascript derived class will only be recognized as the parent class. For example, if I override any methods of the parent class, only the parent class methods are called. Also, `obj instanceof DerivedClass == false` and `obj instance ParentClass == true`, where i would expect both `instanceof` conditions to be true. 

### Steps to Reproduce

1. Create a struct `A` in rust and export the struct via `#[wasm_bindgen]`

2. In javasript, create a class `B` that inherits the Rust struct `A`

3. Create an instance of your derived javascript class. `let obj = new B()`

4. test `expect(obj instanceof B).toBe(true)`

### Expected Behavior

Step 4 should pass

### Actual Behavior

Step 4 will fail (obj is not an instance of B)

### Additional Context

Why this matters is because when I override any methods of the Base class, the override methods are not being called, but instead only the base methods are being called.

snOm3ad:
After #1594 constructors of Rust exported structs started using class wrapping when generating JS shims. Wrapping erases prototype information from the object instance in JS and as a result it is not possible to override methods (via inheritance) of the generated class.

Additionally, some checks to ensure constructors always return an instance of `Self` were lost.

This PR fixes the above two issues by embedding the kind of export into the `CallExport` instruction and monitoring for constructor exports calls during the export adapter creation.

Once a constructor export call is detected and validated a new instruction `SelfFromI32` is pushed into the stack that avoids class wrapping.

Fixes #3213

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -515,7 +515,7 @@ impl<'a> Context<'a> {
             None => AuxExportKind::Function(export.function.name.to_string()),
         };
 
-        let id = self.export_adapter(export_id, descriptor)?;
+        let id = self.export_adapter(export_id, descriptor, &kind)?;
         self.aux.export_map.insert(
             id,
             AuxExport {
@@ -870,7 +870,13 @@ impl<'a> Context<'a> {
                 ret: descriptor.clone(),
                 inner_ret: Some(descriptor.clone()),
             };
-            let getter_id = self.export_adapter(getter_id, getter_descriptor)?;
+            let kind = AuxExportKind::Method {
+                class: struct_.name.to_string(),
+                name: field.name.to_string(),
+                receiver: AuxReceiverKind::Borrowed,
+                kind: AuxExportedMethodKind::Getter,
+            };
+            let getter_id = self.export_adapter(getter_id, getter_descriptor, &kind)?;
             self.aux.export_map.insert(
                 getter_id,
                 AuxExport {
@@ -878,12 +884,7 @@ impl<'a> Context<'a> {
                     arg_names: None,
                     asyncness: false,
                     comments: concatenate_comments(&field.comments),
-                    kind: AuxExportKind::Method {
-                        class: struct_.name.to_string(),
-                        name: field.name.to_string(),
-                        receiver: AuxReceiverKind::Borrowed,
-                        kind: AuxExportedMethodKind::Getter,
-                    },
+                    kind,
                     generate_typescript: field.generate_typescript,
                     generate_jsdoc: field.generate_jsdoc,
                     variadic: false,
@@ -902,7 +903,13 @@ impl<'a> Context<'a> {
                 ret: Descriptor::Unit,
                 inner_ret: None,
             };
-            let setter_id = self.export_adapter(setter_id, setter_descriptor)?;
+            let kind = AuxExportKind::Method {
+                class: struct_.name.to_string(),
+                name: field.name.to_string(),
+                receiver: AuxReceiverKind::Borrowed,
+                kind: AuxExportedMethodKind::Setter,
+            };
+            let setter_id = self.export_adapter(setter_id, setter_descriptor, &kind)?;
             self.aux.export_map.insert(
                 setter_id,
                 AuxExport {
@@ -910,12 +917,7 @@ impl<'a> Context<'a> {
                     arg_names: None,
                     asyncness: false,
                     comments: concatenate_comments(&field.comments),
-                    kind: AuxExportKind::Method {
-                        class: struct_.name.to_string(),
-                        name: field.name.to_string(),
-                        receiver: AuxReceiverKind::Borrowed,
-                        kind: AuxExportedMethodKind::Setter,
-                    },
+                    kind,
                     generate_typescript: field.generate_typescript,
                     generate_jsdoc: field.generate_jsdoc,
                     variadic: false,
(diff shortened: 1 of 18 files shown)


crates/cli-support/src/wit/mod.rs
```diff
@@ -1275,7 +1278,22 @@ impl<'a> Context<'a> {
         };
 
         let mut ret = args.cx.instruction_builder(true);
-        ret.outgoing(&signature.ret)?;
+        if let Instruction::CallExport(_, AuxExportKind::Constructor(ref class)) = call {
+            if let Descriptor::RustStruct(ref name) = signature.ret {
+                if class != name {
```
snOm3ad, line 1283: How useful is this check really? I mean all you have to do to bypass it is to wrap your foreign return type inside an `Option` or any other parameterized type and you've gone back to the old behaviour.

Liamolucko: It should reject `Option<T>` too - only the Rust struct that the constructor's for should be allowed.

guide/src/reference/attributes/on-rust-exports/constructor.md
```diff
@@ -32,3 +32,41 @@ import { Foo } from './my_module';
 const f = new Foo();
 console.log(f.get_contents());
 ```
+
+## Caveats
+
+Starting from v0.2.48 there is a bug in the `wasm-bindgen-cli-support` crate which breaks inheritance of exported Rust structs from JavaScript side (see [#3213](https://github.com/rustwasm/wasm-bindgen/issues/3213)). If you want to inherit a Rust struct such as:
```
Liamolucko, line 38: I don't think anyone cares about which internal crate caused the problem; you should probably just say `wasm-bindgen` or 'the `wasm-bindgen` CLI'.

Liamolucko (commented):
I had a bit of a different idea for how to implement this, which was to replace the generated `return {}` with `this.__wbg_ptr = {}` for constructors. `return` statements get inserted [here](https://github.com/rustwasm/wasm-bindgen/blob/ed43f546bcfb76e10c9615968faa3a1001b227ea/crates/cli-support/src/js/binding.rs#L177), and you can already check whether the function's a constructor using `self.constructor`, so it doesn't require piping extra state around. I should probably have mentioned this when you asked for pointers; sorry about that.

Anyway, this implementation is also fine so you can stick with it if you want, but getting rid of the `return` entirely is a bit cleaner IMO.

Liamolucko:
> I had a bit of a different idea for how to implement this, which was to replace the generated `return {}` with `this.__wbg_ptr = {}` for constructors. `return` statements get inserted [here](https://github.com/rustwasm/wasm-bindgen/blob/ed43f546bcfb76e10c9615968faa3a1001b227ea/crates/cli-support/src/js/binding.rs#L177), and you can already check whether the function's a constructor using `self.constructor`, so it doesn't require piping extra state around.

Oops, that wouldn't actually work, that'd try to assign an object to `this.__wbg_ptr`. You could work around that by manually getting rid of the final `RustFromI32` instruction if a constructor's being generated. (You could also check that it has the right return value at the same time: if the last instruction isn't a `RustFromI32`, it doesn't return a Rust struct, and then you can check that the contained `class` is correct.)

guide/src/reference/attributes/on-rust-exports/constructor.md
snOm3ad: I wasn't sure if it was worth even including this. But yeah good point, I'll make these changes.

snOm3ad:
> I had a bit of a different idea for how to implement this, which was to replace the generated `return {}` with `this.__wbg_ptr = {}` for constructors. `return` statements get inserted [here](https://github.com/rustwasm/wasm-bindgen/blob/ed43f546bcfb76e10c9615968faa3a1001b227ea/crates/cli-support/src/js/binding.rs#L177), and you can already check whether the function's a constructor using `self.constructor`, so it doesn't require piping extra state around. I should probably have mentioned this when you asked for pointers; sorry about that.
> 
> Anyway, this implementation is also fine so you can stick with it if you want, but getting rid of the `return` entirely is a bit cleaner IMO.

Hmm, I had not thought about this!

> You could work around that by manually getting rid of the final `RustFromI32` instruction if a constructor's being generated. 

I was thinking of instead passing the `constructor` field as a parameter to the `instruction` function. Since it's not enough to get rid of `RustFromI32` I'd also need to do the same for `OptionRustFromI32`. Something like this:

```rust
fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool, constructor: &Option<String>) -> Result<(), Error> {
    match instr {
        Instruction::RustFromI32 { class } => {
            if let Some(name) = constructor {
                if name != class {
                    bail!("constructor for `{}` cannot return `{}`", class, name);
                }
                let val = js.pop();
                js.push(format!("{} >>> 0;", val));
            } else {
                js.cx.require_class_wrap(class);
                let val = js.pop();
                js.push(format!("{}.__wrap({})", class, val));
            }
        }
    }
}
```
What do you think? I'd also have to modify the return like you mentioned above.  And, thank you for looking into this! It is indeed much cleaner. I'll probably make another PR based off from this one.

PR #3562, opened 2023-08-18
Remove class wrap for constructors in Rust exports

### LINKED ISSUE #3213: Cannot inherit from structs created via `#[wasm_bindgen]` attribute as expected

snOm3ad:
After #1594 constructors of Rust exported structs started using class wrapping when generating JS shims. Wrapping erases prototype information from the object instance in JS and as a result it is not possible to override methods (via inheritance) of the generated class.

This PR is rebased from #3561, it passes the constructor information from the `Builder` into the instruction translation function which ensures that constructors returning `Self` no longer rely on `__wrap`.

Fixes #3213

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -179,11 +179,7 @@ impl<'a, 'b> Builder<'a, 'b> {
             0 => {}
             1 => {
                 let val = js.pop();
-                if self.constructor.is_some() {
-                    js.prelude(&format!("this.__wbg_ptr = {}", val));
-                } else {
-                    js.prelude(&format!("return {};", val));
-                }
+                js.prelude(&format!("return {};", val));
             }
 
             // TODO: this should be pretty trivial to support (commented out
@@ -1002,31 +998,26 @@ fn instruction(
 
         Instruction::RustFromI32 { class } => {
             let val = js.pop();
-            if let Some(name) = constructor {
-                if name != class {
-                    bail!("constructor for `{}` cannot return `{}`", name, class);
+            match constructor {
+                Some(name) if name == class => {
+                    js.prelude(&format!("this.__wbg_ptr = {} >>> 0;", val));
+                    js.push(String::from("this"));
                 }
-                js.push(format!("{} >>> 0;", val));
-            } else {
-                js.cx.require_class_wrap(class);
-                js.push(format!("{}.__wrap({})", class, val));
+                Some(_) | None => {
+                    js.cx.require_class_wrap(class);
+                    js.push(format!("{}.__wrap({})", class, val));
+                }
             }
         }
 
         Instruction::OptionRustFromI32 { class } => {
+            assert!(constructor.is_none());
             let val = js.pop();
-            if let Some(name) = constructor {
-                if name != class {
-                    bail!("constructor for `{}` cannot return `{}`", name, class);
-                }
-                js.push(format!("{val} === 0 ? 0 : ({val} >>> 0);"));
-            } else {
-                js.cx.require_class_wrap(class);
-                js.push(format!(
-                    "{0} === 0 ? undefined : {1}.__wrap({0})",
-                    val, class,
-                ))
-            }
+            js.cx.require_class_wrap(class);
+            js.push(format!(
+                "{0} === 0 ? undefined : {1}.__wrap({0})",
+                val, class,
+            ));
         }
 
         Instruction::CachedStringLoad {
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -166,7 +166,12 @@ impl<'a, 'b> Builder<'a, 'b> {
         // more JIT-friendly. The generated code should be equivalent to the
         // wasm interface types stack machine, however.
         for instr in instructions {
-            instruction(&mut js, &instr.instr, &mut self.log_error)?;
+            instruction(
+                &mut js,
+                &instr.instr,
+                &mut self.log_error,
+                &self.constructor,
+            )?;
         }
 
         assert_eq!(js.stack.len(), adapter.results.len());
@@ -174,7 +179,11 @@ impl<'a, 'b> Builder<'a, 'b> {
             0 => {}
             1 => {
                 let val = js.pop();
-                js.prelude(&format!("return {};", val));
+                if self.constructor.is_some() {
+                    js.prelude(&format!("this.__wbg_ptr = {}", val));
+                } else {
+                    js.prelude(&format!("return {};", val));
+                }
             }
 
             // TODO: this should be pretty trivial to support (commented out
@@ -537,7 +546,12 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
     }
 }
 
-fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) -> Result<(), Error> {
+fn instruction(
+    js: &mut JsBuilder,
+    instr: &Instruction,
+    log_error: &mut bool,
+    constructor: &Option<String>,
+) -> Result<(), Error> {
     match instr {
         Instruction::ArgGet(n) => {
             let arg = js.arg(*n).to_string();
@@ -987,18 +1001,32 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
         }
 
         Instruction::RustFromI32 { class } => {
-            js.cx.require_class_wrap(class);
             let val = js.pop();
-            js.push(format!("{}.__wrap({})", class, val));
+            if let Some(name) = constructor {
+                if name != class {
+                    bail!("constructor for `{}` cannot return `{}`", name, class);
+                }
+                js.push(format!("{} >>> 0;", val));
+            } else {
+                js.cx.require_class_wrap(class);
+                js.push(format!("{}.__wrap({})", class, val));
+            }
         }
 
         Instruction::OptionRustFromI32 { class } => {
-            js.cx.require_class_wrap(class);
             let val = js.pop();
-            js.push(format!(
-                "{0} === 0 ? undefined : {1}.__wrap({0})",
-                val, class,
-            ))
+            if let Some(name) = constructor {
+                if name != class {
+                    bail!("constructor for `{}` cannot return `{}`", name, class);
+                }
+                js.push(format!("{val} === 0 ? 0 : ({val} >>> 0);"));
+            } else {
+                js.cx.require_class_wrap(class);
+                js.push(format!(
+                    "{0} === 0 ? undefined : {1}.__wrap({0})",
+                    val, class,
+                ))
+            }
         }
 
         Instruction::CachedStringLoad {
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -484,7 +484,29 @@ impl<'a> Context<'a> {
             Some(class) => {
                 let class = class.to_string();
                 match export.method_kind {
-                    decode::MethodKind::Constructor => AuxExportKind::Constructor(class),
+                    decode::MethodKind::Constructor => {
+                        let msg = "Trying to return JS primitive from constructor, return value will be ignored. Use a builder instead (remove `constructor` attribute).";
+                        match descriptor.ret {
+                            Descriptor::I8
+                            | Descriptor::U8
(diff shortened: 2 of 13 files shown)


crates/cli-support/src/js/binding.rs
```diff
@@ -987,18 +1001,32 @@ fn instruction(js: &mut JsBuilder, instr: &Instruction, log_error: &mut bool) ->
...
             let val = js.pop();
-            js.push(format!(
-                "{0} === 0 ? undefined : {1}.__wrap({0})",
-                val, class,
-            ))
+            if let Some(name) = constructor {
+                if name != class {
+                    bail!("constructor for `{}` cannot return `{}`", name, class);
+                }
+                js.push(format!("{val} === 0 ? 0 : ({val} >>> 0);"));
```
Liamolucko, line 1022: I mentioned in the broader review comment that I don't think `Option<T>` should be allowed anyway, but I'm also not too sure what this is trying to accomplish. Setting `__wbg_ptr` to 0 will simply cause any method calls on the type to fail, since that's what we set it too to mark that the struct's been freed, which doesn't seem super useful to me; there's no good way to check if it's `None`.

Liamolucko (commented):
I think I phrased myself poorly in the last PR - I wasn't trying to say that the constructor for `Foo` should accept either `Foo` or `Option<Foo>`, and disallow `Bar` or `Option<Bar>`, I was trying to say that `Option<Foo>` shouldn't be allowed either. `Foo` should be the *only* valid return type - `Option<Foo>`, `Bar`, and non-struct types such as `String` should all be disallowed.

As for choosing to pass `constructor` to `instruction`: sure, that's fine, I think it's about as clean as my suggestion. You'll need to insert a check that the function has the right return type somewhere else though, since you can't tell whether it's the last instruction or not.

Liamolucko:
Oops, I just noticed the CI failure. That's because of this test:

```rs
#[wasm_bindgen]
pub struct AsyncStruct;

#[wasm_bindgen]
impl AsyncStruct {
    #[wasm_bindgen(constructor)]
    pub async fn new() -> AsyncStruct {
        AsyncStruct
    }

    pub async fn method(&self) -> u32 {
        42
    }
}
```

It expects returning a `Future<Output = AsyncStruct>` to work. Apparently I added it in #3043, without thinking about it much.

Thinking back on it, I vaguely remember trying to fix this issue before, noticing that failing test and then bailing on it. (found it: https://github.com/rustwasm/wasm-bindgen/commit/052281f10a21eed7f243fe80d3c30fd7b80dbe87)

While that's probably the most sensible use of a constructor returning something other than the type it's a constructor for, I still don't think it should be allowed - just use a builder instead. I just added that test on a whim without really thinking about it. However, I'm now a bit worried that someone might already be relying on that, and this could actually cause breakage.

Ugh, I don't know - the breakage could be avoided by falling back on the old behaviour if a constructor returns the wrong type (which is basically what the last PR did, now that I think about it), but it feels kind of crappy to be explicitly supporting this weirdness. I guess we could print some kind of deprecation warning, but that'd only be able to show up when you ran the CLI, not when using the macro... or we could just risk the breakage. Do you have any opinion?

snOm3ad:
> I think I phrased myself poorly in the last PR - I wasn't trying to say that the constructor for `Foo` should accept either `Foo` or `Option<Foo>`, and disallow `Bar` or `Option<Bar>`, I was trying to say that `Option<Foo>` shouldn't be allowed either. `Foo` should be the _only_ valid return type - `Option<Foo>`, `Bar`, and non-struct types such as `String` should all be disallowed.

Initially, that's what I thought as well. Hence why in #3561 I inserted the check before generating the adapter. Because you can check the return type from the `signature` that is passed to `register_export_adapter`. 

>  You'll need to insert a check that the function has the right return type somewhere else though, since you can't tell whether it's the last instruction or not.

This is why inserting the check at this point is a bit more cumbersome because you now need to peek into the instructions that are being passed to generate this adapter. But this should not be that big of a deal, so long as we allow things like `Result<T = Self, E>`, `Option<T = Self>`, etc.

If we instead decide to remove support for these, then it becomes trickier to do the check here since you need to know the sequence of instructions these types generate. E.g. `Result<T, E>` is `Instruction::UnwrapResult` followed by `Instruction::RustFromI32`. Here, if you only check the last instruction you'd think that the constructor is returning `Self.

snOm3ad:
> Oops, I just noticed the CI failure. That's because of this test:
> 
> ```rust
> #[wasm_bindgen]
> pub struct AsyncStruct;
> 
> #[wasm_bindgen]
> impl AsyncStruct {
>     #[wasm_bindgen(constructor)]
>     pub async fn new() -> AsyncStruct {
>         AsyncStruct
>     }
> 
>     pub async fn method(&self) -> u32 {
>         42
>     }
> }
> ```
> 
> It expects returning a `Future<Output = AsyncStruct>` to work. Apparently I added it in #3043, without thinking about it much.

Interestingly enough the test passes if you return `this` from the constructor. I added this back from the original PR.

> While that's probably the most sensible use of a constructor returning something other than the type it's a constructor for, I still don't think it should be allowed - just use a builder instead. I just added that test on a whim without really thinking about it. However, I'm now a bit worried that someone might already be relying on that, and this could actually cause breakage.

I think there are already _a lot_ of users relying on this behaviour! The ray-tracing example in the repository returns `Result<T = Self, JsValue>` and a search of `#[wasm_bindgen(constructor)]` in sourcegraph shows multiple libraries using this pattern:

- https://sourcegraph.com/github.com/solana-labs/solana/-/blob/sdk/program/src/wasm/hash.rs
- https://sourcegraph.com/github.com/acteng/atip/-/blob/route_info/src/lib.rs
- https://sourcegraph.com/github.com/wnfs-wg/rs-wnfs/-/blob/wnfs-wasm/src/fs/public/file.rs

I think it might be possible to query `crates.io` to see how many crates are relying on a specific pattern. I remember seeing an article about this somewhere. If I can find it I'll post the data here just for future reference. If we stop sneak this in breakage is not a risk, it's a guarantee!

crates/cli-support/src/js/binding.rs
snOm3ad: Yeah, I wasn't sure what to do this about this either. In the current implementation sets the instance's `__wbg_ptr` is `undefined` since it never calls `__wrap`. I'm not sure if this is better tbh!

snOm3ad:
> Ugh, I don't know - the breakage could be avoided by falling back on the old behaviour if a constructor returns the wrong type (which is basically what the last PR did, now that I think about it), but it feels kind of crappy to be explicitly supporting this weirdness. I guess we could print some kind of deprecation warning, but that'd only be able to show up when you ran the CLI, not when using the macro

We could emit the warning from the macro by using something like [proc-macro-warning](https://crates.io/crates/proc-macro-warning).

Liamolucko:
> I think there are already _a lot_ of users relying on this behaviour! The ray-tracing example in the repository returns `Result<T = Self, JsValue>` and a search of `#[wasm_bindgen(constructor)]` in sourcegraph shows multiple libraries using this pattern:

Ah, I didn't think of `Result` - that should definitely be supported, having a constructor that can throw is a perfectly normal thing to do.

I also found an example of async constructors being used in the wild, so we probably need to support those too: https://github.com/RingsNetwork/rings-node/blob/master/node/src/browser/client.rs#L143-L153

So yeah, people are definitely already relying on this; let's not break it. I hadn't heard of sourcegraph before, thank you for pointing me to it!

We might still want to disallow some types though, because JavaScript handles constructor returns kind of weirdly:

```js
class Foo {
    constructor(x) {
        this.a = 2;
        return x;
    }
}

console.log(new Foo(null));      // Foo { a: 2 }
console.log(new Foo(undefined)); // Foo { a: 2 }
console.log(new Foo("hi"));      // Foo { a: 2 }
console.log(new Foo(12));        // Foo { a: 2 }
console.log(new Foo(() => {}));  // [Function]
console.log(new Foo({}));        // {}
```

Anything other than an object gets completely ignored, so I think we should disallow returning JS primitives (`String`, `i32`, etc.) as they won't work properly.

That also includes `Option<T>`, since it should be returning `null` or `undefined` in the `None` case, both of which get ignored. That's what's leading to the current implementation just returning an object with no `__wbg_ptr` set like you mentioned.

You'll still be able to trigger this by accident using `JsValue`, but that needs to be supported in order to not break async constructors.

The last thing to consider is whether to allow returning different Rust structs. I'm basically indifferent now, it's not any weirder than returning a `JsValue`, so probably just allow it as well.

You'll still need to check for it though - we can't just set `this.__wbg_ptr` to a pointer to the wrong type, so if it's a different class you have to go back to `__wrap`.

> Interestingly enough the test passes if you return `this` from the constructor. I added this back from the original PR:

That's working because `async fn`s get desugared down to regular synchronous functions which return `Promise`s. This means that they don't go through `RustFromI32` in the first place and have their code generated the same as a normal function.

Going back to returning `this` is probably also the way forwards here, since we need to generate a `return` statement whenever the constructor doesn't return `Self` (or `Result<Self, E>`), and it's easier to just do it unconditionally.

As for where to insert the type checks, you should be able to do it in the same place as I did in https://github.com/rustwasm/wasm-bindgen/commit/052281f10a21eed7f243fe80d3c30fd7b80dbe87: inside `Context::export`, since all of the information about the export is available there while it's being decoded.

crates/cli-support/src/wit/mod.rs
```diff
@@ -484,7 +484,30 @@ impl<'a> Context<'a> {
...
+                            | Descriptor::F32
+                            | Descriptor::F64
+                            | Descriptor::I64
+                            | Descriptor::U64
+                            | Descriptor::Boolean
+                            | Descriptor::Char
+                            | Descriptor::CachedString
+                            | Descriptor::String
+                            | Descriptor::Option(_)
+                            | Descriptor::Unit => bail!(msg),
```
Liamolucko, line 506: This should also include `Enum`, since (exported) enums are represented as integers in JS.

`Ref`s and `RefMut`s to any of these should probably also be included, since they'll still map to primitives while behind a reference, although it doesn't matter a whole lot since you can't return references right now anyway.

(Imported / string enums are handled purely at the macro level, so we don't need to worry about them here.)

crates/cli-support/src/wit/mod.rs
```diff
@@ -484,7 +484,30 @@ impl<'a> Context<'a> {
             Some(class) => {
                 let class = class.to_string();
                 match export.method_kind {
-                    decode::MethodKind::Constructor => AuxExportKind::Constructor(class),
+                    decode::MethodKind::Constructor => {
+                        let msg = "Trying to return JS primitive from constructor, return value will be ignored. Use a builder instead (remove `constructor` attribute).";
```
Liamolucko, line 488: This should include _which_ constructor is causing the error:

```suggestion
                        let msg = format!("The constructor for `{class}` tries to return a JS primitive, which would cause the return value to be ignored. Use a builder instead (remove the `constructor` attribute).");
```

(Don't actually use `format!` though, inline it into the `bail!`.)

Liamolucko (commented):
Looks good, I just have a couple small comments.

crates/cli-support/src/wit/mod.rs
snOm3ad: 👍  Done!

crates/cli-support/src/wit/mod.rs
snOm3ad: I moved this check to it's own function so that it is cleaner to check nested descriptors. Like you mentioned `Ref` and `RefMut` are not relevant for now, but it occurred to me that `Result` still is.

Liamolucko (approved):
Thanks!

PR #3566, opened 2023-08-24
Prioritized Task Scheduling API

daxpedda:
Implements two proposals:
- Prioritized Task Scheduling
- Early detection of input events

Currently only enabled by default in Chromium even though they are in a very early stage of the W3C standardization process.

See:
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API)
[Prioritized Task Scheduling draft](https://wicg.github.io/scheduling-apis) ([Firefox WebIDL](https://hg.mozilla.org/mozilla-central/file/014c9a0ccc445de93ad0fe0fb91472600aa5db53/testing/web-platform/tests/interfaces/scheduling-apis.idl))
[Early detection of input events draft](https://wicg.github.io/is-input-pending) ([Firefox WebIDL](https://hg.mozilla.org/mozilla-central/file/014c9a0ccc445de93ad0fe0fb91472600aa5db53/testing/web-platform/tests/interfaces/is-input-pending.idl))


crates/web-sys/webidls/unstable/scheduling-apis.webidl
```diff
@@ -0,0 +1,58 @@
...
+
+dictionary TaskControllerInit {
+  TaskPriority priority = "user-visible";
+};
+
+[Exposed=(Window,Worker)]
+interface TaskController : AbortController {
+  constructor(optional TaskControllerInit init = {});
+
+  undefined setPriority(TaskPriority priority);
```
Liamolucko, line 40: I probably should have mentioned this _before_ you merged, but I think this needs to be annotated with `[Throws]`. I'm not sure why it isn't in Firefox, but the spec says it's supposed to throw if it gets called reentrantly (i.e. inside a `prioritychange` handler).

daxpedda: My bad, should have waited longer, gonna make a fix momentarily.

daxpedda: Thanks for catching that!

PR #3577, opened 2023-08-28
fix: Optional JSDoc `@param`

### LINKED ISSUE #3576: JSDoc: Option `@param` syantax missing
### Steps to Reproduce

Input:
```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(typescript_type = "Config")]
    pub type Config;
}

#[wasm_bindgen]
pub fn format(input: &str, config: Option<Config>) -> Result<String, String> {
    todo!();
}
```

### Expected Behavior

```JavaScript
/**
* @param {string} input
* @param {Config | undefined} [config]
* @returns {string}
*/
export function format(input, config) {
// ...
```

Please note the square brackets `[` `]` around the `config`.

ref: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values

### Actual Behavior

```JavaScript
/**
* @param {string} input
* @param {Config | undefined} config
* @returns {string}
*/
export function format(input, config) {
// ...
```

### Additional Context

I found this issue when I ran tests with deno.
https://github.com/wasm-fmt/ruff_fmt/actions/runs/6001784974/job/16276722905

magic-akari:
- fix #3576

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@
 * Fixed bug allowing JS primitives to be returned from exported constructors.
   [#3562](https://github.com/rustwasm/wasm-bindgen/pull/3562)
 
-* Fixed Optional parameters in JSDoc
+* Fixed optional parameters in JSDoc.
   [#3577](https://github.com/rustwasm/wasm-bindgen/pull/3577)
 
 ## [0.2.87](https://github.com/rustwasm/wasm-bindgen/compare/0.2.86...0.2.87)


daxpedda (changes requested):
Otherwise looks great!

Liamolucko:
An issue I can see here is that this will currently produce optional parameters before required parameters; e.g. this function:

```rs
#[wasm_bindgen]
pub fn foo(a: Option<i32>, b: i32) {}
```

generates this JSDoc:
```js
/**
* @param {number | undefined} [a]
* @param {number} b
*/
```

`a` is marked as optional even though `b` is required, so it's impossible to actually leave it out.

I wasn't sure if this actually mattered, but it seems like it does because TypeScript complains about it:

```
test.js(5,17): error TS1016: A required parameter cannot follow an optional parameter.
```

magic-akari:
> An issue I can see here is that this will currently produce optional parameters before required parameters; e.g. this function:

@Liamolucko You are right. I think it should be handled by reversing the order like typescript_signature.

CHANGELOG.md
```diff
@@ -77,6 +77,9 @@
 * Fixed bug allowing JS primitives to be returned from exported constructors.
   [#3562](https://github.com/rustwasm/wasm-bindgen/pull/3562)
 
+* Fixed Optional parameters in JSDoc
```
daxpedda, line 80:
```suggestion
* Fixed optional parameters in JSDoc.
```

daxpedda (approved):
LGTM!
Thanks @Liamolucko for catching that.

PR #3580, opened 2023-08-29
Add bindings for `CanvasTransform.setTransform(DOMMatrix2DInit)`

### LINKED ISSUE #3578: web_sys: missing set_transform with matrix function
### Motivation

The function [setTransform](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setTransform) on Canvas 2D can accept a matrix on modern browsers. This is currently not exposed in `web_sys`. However, the sister function `getTransform` [is exposed](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.CanvasRenderingContext2d.html#method.get_transform).

### Proposed Solution

Expose a `set_transform_with_matrix` function, to match the `get_transform` ones.

daxpedda:
As in the title.

See:
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setTransform)
[WHATWG spec](https://html.spec.whatwg.org/multipage/canvas.html#2dcontext:dom-context-2d-settransform-matrix) ([Firefox WebIDL](https://hg.mozilla.org/mozilla-central/file/18ac70a5128d520641cced852fc0059536c52713/dom/webidl/CanvasRenderingContext2D.webidl#l171))
[Geometry Interfaces](https://drafts.fxtf.org/geometry/#dictdef-dommatrix2dinit)

Fixes #3578.


crates/web-sys/src/features/gen_CanvasPattern.rs
```diff
@@ -19,5 +19,5 @@ extern "C" {
     #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/CanvasPattern/setTransform)"]
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `CanvasPattern`, `SvgMatrix`*"]
-    pub fn set_transform(this: &CanvasPattern, matrix: &SvgMatrix);
+    pub fn set_transform_with_default_dom_matrix_2d_init(this: &CanvasPattern, matrix: &SvgMatrix);
```
Liamolucko, line 22: This is a breaking change (and not really a desirable one, either).

I can see what happened here: you used `FIXED_INTERFACES` to prevent `CanvasRenderingContext2d::set_transform` being changed to mean the `DomMatrix` overload with its default value, but that affected this method too since it doesn't distinguish between methods on different types.

I think `FIXED_INTERFACES` will have to be extended to allow specifying both the type name and method name to fix this.

PR #3581, opened 2023-08-29
Filter interface fixes by type name

daxpedda:
Follow-up to #3580.

Thanks @Liamolucko!

### DIFF
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -196,6 +196,7 @@ pub enum TypePosition {
 impl<'src> FirstPassRecord<'src> {
     pub fn create_imports(
         &self,
+        type_name: Option<&str>,
         container_attrs: Option<&ExtendedAttributeList<'src>>,
         id: &OperationId<'src>,
         data: &OperationData<'src>,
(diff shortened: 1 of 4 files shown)


crates/webidl/src/util.rs
```diff
@@ -500,8 +501,13 @@ impl<'src> FirstPassRecord<'src> {
         }
 
         for interface in &mut ret {
-            if let Some(fixed) = FIXED_INTERFACES.get(&interface.name.to_string().as_ref()) {
-                interface.name = rust_ident(fixed);
+            if let Some(type_name) = type_name {
+                if let Some(map) = FIXED_INTERFACES.get(type_name) {
+                    if let Some(fixed) = map.get(&interface.name.to_string().as_ref())
```
Liamolucko, line 506: I think it'd be nicer to use `Option::and_then` rather than having these nested `if let`s, but that's just a nitpick / matter of opinion. Feel free to disregard.

Liamolucko (approved):
Looks good!

crates/webidl/src/util.rs
daxpedda: Thanks!
Couldn't merge all three because `map` is a temporary variable.

PR #3582, opened 2023-08-29
Remove `ReadableStreamByobReader::read_with_u8_array()`

daxpedda:
See #3579.

This is technically a breaking change, but seeing as this would always fail it seems fine to me. We could always deprecate it instead.
@Liamolucko WDYT?


Liamolucko (commented):
Yeah, I'm fine with removing it.

Implementation-wise, I think `[RustNotWasmMemory]` should probably apply to `BufferSource` as well, although it doesn't actually matter right now since it's only used on one API.

PR #3593, opened 2023-09-04
Add attribute `crate` to proc-macro `wasm-bindgen-test`

### LINKED ISSUE #3588: Support `crate` attribute in proc-macro `wasm_bindgen_test`
### Motivation

We have multiple crates that we would like to test using `wasm-bindgen-test`.

For the ease-of-use on our side we create a proc-macro that wrap multiple other proc-macro (`rstest`, `tokio::test`, custom testbed) for testing.

To use that proc-macro the developer would have to import (add as a dependency) a single crate that provide the proc-macro + re-export `rstest`, `tokio::test` and testbed.

But to support `wasm-bindgen-test` we're required to alongside the test crate to also import `wasm-bindgen-test` and we would like to have it re-exported in the test crate instead.

Currently `wasm-bindgen-test` proc macro use a fixed path to the crate `wasm_bindgen_test` (cf the generated function by the proc-macro `wasm_bindgen_test`

```rust
#[no_mangle]
pub extern "C" fn __wbgt_case_1_invalid_type_deserialization_2(
    cx: &::wasm_bindgen_test::__rt::Context,
) {
    let test_name = "misc::serde_invalid_type_deserialization::case_1_invalid_type_deserialization";
    cx.execute_sync(test_name, case_1_invalid_type_deserialization, None);
}
```

### Proposed Solution

A solution would be to do something like [`tokio::test`](https://docs.rs/tokio/1.32.0/tokio/attr.test.html) handle [renaming the crate](https://docs.rs/tokio/1.32.0/tokio/attr.test.html#rename-package) with the `crate` attribute provided to the proc-macro.

A clear and concise description of the feature you would like to add, and how it solves the motivating problem.

### Alternatives

An alternative would have to add `wasm-bindgen-test` to each of our crates that will be tested with it.
For the easy-of-use we would have preferred to import a single crate that re-expose those _utility_ crates.

### Additional Context

This feature will required to:

- Handle a new token here:
  
  https://github.com/rustwasm/wasm-bindgen/blob/86fd961bfec72587a7f5eec7d5dd3e4c1b57ee0d/crates/test-macro/src/lib.rs#L23-L31

- Tweak the generated function at 
  
  https://github.com/rustwasm/wasm-bindgen/blob/86fd961bfec72587a7f5eec7d5dd3e4c1b57ee0d/crates/test-macro/src/lib.rs#L99

FirelightFlagboy:
Also rework how the attributes of the proc-macro are parsed using `syn` (it was simpler to handle an attribute that has a value).

Closes #3588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,8 @@
 * Add bindings for `CanvasTransform.setTransform(DOMMatrix2DInit)`.
   [#3580](https://github.com/rustwasm/wasm-bindgen/pull/3580)
 
-* Add attribute `crate` in the proc-macro `wasm_bindgen_test`.
+* Add a `crate` attribute to the `wasm_bindgen_test` proc-macro to specify a
+  non-default path to the `wasm-bindgen-test` crate.
   [#3593](https://github.com/rustwasm/wasm-bindgen/pull/3593)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@
 * Add bindings for `CanvasTransform.setTransform(DOMMatrix2DInit)`.
   [#3580](https://github.com/rustwasm/wasm-bindgen/pull/3580)
 
+* Add attribute `crate` in the proc-macro `wasm_bindgen_test`.
+  [#3593](https://github.com/rustwasm/wasm-bindgen/pull/3593)
+
 ### Changed
 
 * Updated the WebGPU WebIDL.
--- a/crates/test-macro/Cargo.toml
+++ b/crates/test-macro/Cargo.toml
@@ -14,6 +14,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0"
 quote = "1.0"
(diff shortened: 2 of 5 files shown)


daxpedda (commented):
I think the current failure is because `wasm-bindgen-test-macro` isn't defining the full feature set it requires from `syn`? But LGTM otherwise.

Gonna review it as soon as CI passes.

daxpedda (changes requested):
Apologies, could you also add an entry in the changelog?

LGTM otherwise.

CHANGELOG.md
```diff
@@ -42,6 +42,9 @@
 * Add bindings for `CanvasTransform.setTransform(DOMMatrix2DInit)`.
   [#3580](https://github.com/rustwasm/wasm-bindgen/pull/3580)
 
+* Add attribute `crate` in the proc-macro `wasm_bindgen_test`.
```
daxpedda, line 45:
```suggestion
* Add a `crate` attribute to the `wasm_bindgen_test` proc-macro to specify a
  non-default path to the `wasm-bindgen-test` crate.
```

daxpedda:
Thanks!

Issue #3552, opened 2023-08-10
Support for `multivalue`


CryZe:
### Motivation

Multivalue is now finally properly supported by LLVM 17, which made its way into recent Rust nightlies.

### Proposed Solution

It seems like a bunch of bindings that wasm-bindgen generates are using multivalue, but wasm-bindgen doesn't generate the JS side correctly to return multiple values.

### Alternatives

The bindings could be adjusted to just not use multivalue at all, then the JS side doesn't need to be changed.

### Additional Context

None so far

Liamolucko:
Huh, I didn't realise that enabling multivalue affected the C ABI. That... kind of seems like a bug to me? Even though it's currently wrong, the C ABI is still supposed to be a fixed ABI that matches C, not change depending on what target features are enabled.

> The bindings could be adjusted to just not use multivalue at all, then the JS side doesn't need to be changed.

That was already pretty much the plan anyway, we're trying to make it so that `wasm-bindgen` no longer relies on Rust's incorrect C ABI (#3454).

fornwall:
When it comes to potentially enabling `multivalue` in LLVM by default (trying to get it into LLVM 18 or 19 perhaps? see https://github.com/WebAssembly/tool-conventions/issues/158), is there any ballpark timeline estimate for `multivalue` support in `wasm-bindgen` (not something to be rushed and certainly not demanding anything, just trying to get a sense of what is realistic and desirable for LLVM:s defaults next year)?

Liamolucko:
The ideal solution would be to wait for `rustc` to switch to the standard C ABI (rust-lang/rust#71871), since that ABI never uses multivalue anyway and so `wasm-bindgen` doesn't have to worry about it. But I don't know how long that will take, since it needs a future-incompatibility lint to tell people to upgrade from old versions of `wasm-bindgen` that will break with the standard ABI, and then we'll need to wait for everyone to upgrade before we can actually make the change.

Now that I think about it though, it should actually be fine to switch only return types to match the standard C ABI right away. The two ABIs are identical in how they handle return types, as long as multivalue is disabled, and `wasm-bindgen` is already broken when multivalue is enabled anyway. That would be enough to fix this.

[...]

PR #3595, opened 2023-09-06
Make `wasm-bindgen` compatible with the standard C ABI

Liamolucko:
This is the first step of #3454: making `wasm-bindgen` compatible with [the standard(ish) Wasm C ABI](https://github.com/WebAssembly/tool-conventions/blob/main/BasicCABI.md), which `clang` uses.

This PR more or less implements @alexcrichton's suggestion in https://github.com/rustwasm/wasm-bindgen/issues/3454#issuecomment-1564482584, to allow `{From,Into}WasmAbi` impls to manually specify multiple types that `Self` should get splatted into when passed as an argument. I implemented it a bit differently though - instead of actually adding multiple associated types to `{From,Into}WasmAbi`, I changed the meaning of `WasmAbi`:

```rust
pub trait WasmAbi {
    type Prim1: WasmPrimitive;
    type Prim2: WasmPrimitive;
    type Prim3: WasmPrimitive;
    type Prim4: WasmPrimitive;

    fn split(self) -> (Self::Prim1, Self::Prim2, Self::Prim3, Self::Prim4);
    fn join(prim1: Self::Prim1, prim2: Self::Prim2, prim3: Self::Prim3, prim4: Self::Prim4)
        -> Self;
}
```

This means that `{From,Into}WasmAbi` impls can still return single, typed values like before, and then it's up to those types to implement `WasmAbi` and specify how they should be splatted. Also, `WasmAbi` types no longer actually have to be FFI-safe themselves - even when they're passed as return values, `split` is still called, and those values are then put into a `#[repr(C)]` struct (`WasmRet<T>`).

The new trait `WasmPrimitive` now has the old meaning of `WasmAbi`, minus allowing `#[repr(C)]` types - any Wasm primitive type.

This PR doesn't actually change `wasm-bindgen`'s ABI at all, just how it's implemented. This meant that I didn't have to touch the CLI.

---

One thing that this PR doesn't fix is #3552. `extern "C"` functions still return `#[repr(C)]` structs the same as before, since the retptr handling happens to be the same between the current and standard C ABI. That is, unless you turn on multivalue, in which case the return value will get splatted with the current ABI. (I'm referring to `rustc`'s `multivalue` target feature here, not `wasm-bindgen`'s own multivalue post processing step.)

It's not a blocker for migrating the ABI, since `wasm-bindgen`'s already broken when multivalue is enabled, but it's still be nice to fix it. 

I'm not sure how it could be fixed without always using a retptr though, which I'd rather avoid. I've experimented a little with type system shenanigans like this:

```rust
trait WasmRetCfg: WasmAbi {
    type Retptr: Retptr;
    type Ret: WasmPrimitive;

    fn ret(self, retptr: Self::Retptr) -> Self::Ret;
    fn recover(retptr: Self::Retptr, ret: Self::Ret) -> Self;
}
```

There'd be one impl where `Retptr` and `Ret` get set to `()` and `Self::Prim1` respectively, if `Self`'s last 3 primitives are all `()`, and another one where `Retptr` and `Ret` get set to `*mut WasmRet<Self>` and `()` respectively, if `Self::Prim2` (and potentially `Prim3`, `Prim4`) isn't `()`. Then a generated import `foo() -> T` would look something like this:

```rust
fn foo() -> T {
    type Abi = <T as FromWasmAbi>::Abi;
    extern "C" foo_extern(retptr: Abi::Retptr) -> Abi::Ret;

    // `Target` is `WasmRet<Abi>` if `Retptr` is `*mut WasmRet<Abi>`, otherwise `()`.
    let ret_space: MaybeUninit<<Abi::Retptr as Retptr>::Target> = MaybeUninit::uninit();
    let retptr = <Abi::Retptr as Retptr>::from_ptr(ret_space.as_mut_ptr());
    let ret = foo_extern(retptr);
    let real_ret = Abi::recover(retptr, ret);
    real_ret.join()
}
```

Unfortunately, I couldn't convince `rustc` that `Prim2` not being `()` meant that those two impls didn't conflict. Maybe someone who's more of a type system guru than me can figure it out, but it might be impossible without specialisation. It'll become unnecessary if/when the C ABI gets fixed anyway.


daxpedda:
I've given it a quick glance, really nice solution! I imagined this to be way messier, but it's pretty neat!
Definitely will try to review it over the weekend.

Even though this only touched unstable ABI would you mind adding an entry in the changelog?

Thinking of deployment here, we could expose an unstable `cfg` flag, like `web_sys_unstable_apis`, that switches all generated APIs by to this new ABI. But this is pointless right now, we have to fix Rust's side first, which I assume will need some nightly feature, then we can try it in `wasm-bindgen` as well.

Liamolucko:
> Even though this only touched unstable ABI would you mind adding an entry in the changelog?

Done. I've also just realised that I forgot to add a changelog entry in #3554, should I open a PR to add one?
 
> Thinking of deployment here, we could expose an unstable `cfg` flag, like `web_sys_unstable_apis`, that switches all generated APIs to this new ABI. But this is pointless right now, we have to fix Rust's side first, which I assume will need some nightly feature, then we can try it in `wasm-bindgen` as well.

I'm not sure what you mean. The new generated bindings still work fine with the existing C ABI, the whole point is that they now work with _both_ that and the standard one. So there's no need for this to be unstable.

I'm now realising that I've been using the word 'ABI' to mean two different things interchangeably: actual Rust ABIs like `extern "C"`, and the format of the bindings that `wasm-bindgen` generates. Sorry about that, I'll try and only use it for the first meaning from here on.

Liamolucko:
> I mean specifically to switch to the new ABI, that doesn't work until Rust fixes it's C ABI, right?

I've been assuming that Rust will just directly cut over to the fixed ABI, so that all of `wasm-bindgen`'s existing `extern "C"` declarations will automatically start referring to the new ABI without `wasm-bindgen` having to do anything.

Are you suggesting that there should be some kind of testing period, where there's a perma-nightly `"C-fixed"` ABI, and then `wasm-bindgen` gets an unstable feature to use it and make sure everything's working?

daxpedda:
> Are you suggesting that there should be some kind of testing period, where there's a perma-nightly `"C-fixed"` ABI, and then `wasm-bindgen` gets an unstable feature to use it and make sure everything's working?

I was assuming that this would be the case, but I'm not advocating for it. If we can skip it and move directly to the fixed C ABI I'm fine with that, but that's not up to us.

In any case, there's nothing for us to do here right now until we know what will happen on Rusts side.

src/convert/impls.rs
```diff
@@ -317,70 +342,52 @@ impl IntoWasmAbi for () {
...
-    type Abi = ResultAbi<T::Abi>;
+impl<T, E> ReturnWasmAbi for Result<T, E>
+where
+    T: IntoWasmAbi,
+    E: Into<JsValue>,
+    T::Abi: WasmAbi<Prim3 = (), Prim4 = ()>,
+{
+    type Abi = Result<T::Abi, u32>;
     #[inline]
     fn return_abi(self) -> Self::Abi {
```
daxpedda, line 385:
```suggestion
    type Abi = Result<T::Abi, u32>;

    #[inline]
    fn return_abi(self) -> Self::Abi {
```

CHANGELOG.md
```diff
@@ -70,6 +70,9 @@
   When exported constructors return `Self`.
   [#3562](https://github.com/rustwasm/wasm-bindgen/pull/3562)
 
+* Made `wasm-bindgen` forwards-compatible with the standard C ABI.
+  [#3595](https://github.com/rustwasm/wasm-bindgen/pull/3595)
```
daxpedda, line 74: I actually meant mentioning that we broke `WasmAbi` & co.
Anyway, we should mention both.

daxpedda (approved):
LGTM!

PR #3598, opened 2023-09-06
Add experimental `ViewTransition` interface

tigerros:
This adds the experimental `ViewTransition` interface. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition) and [W3C specs](https://drafts.csswg.org/css-view-transitions/#the-domtransition-interface). I hope experimental APIs are allowed.

I followed the contributing guidelines, but let me know if I did something wrong.


daxpedda (changes requested):
LGTM!
Could you add an entry to the changelog?

tigerros:
Added.

daxpedda (approved):
Thanks!
Just waiting for CI.

tigerros:
@daxpedda I wanted to ask an additional question, when can it be moved into the `webidls/enabled` directory? I looked at the contributing guide and there was no mention of this.

daxpedda:
We don't have an official guidelines for this. Basically when we *think* that the API isn't going to change anymore.

It's a candidate recommendation, so I'm assuming this isn't going to change anymore, but it's only implemented by Chromium and has [very low usage across the web](https://chromestatus.com/metrics/feature/timeline/popularity/4383).

Considering how close this is to recommendation I would prefer to wait unless there is a strong incentive otherwise.

tigerros:
Alright. If you're wondering, the reason why I added this is because someone requested this as a feature in [Dioxus](https://github.com/DioxusLabs/dioxus), which is an SPA and thus would benefit quite a bit from this.

PR #3603, opened 2023-09-10
Replace Bash with Python for Windows compatibility

jthemphill:
I wanted to get this example running on my Windows machine without fiddling with Powershell environment variables, so I just rewrote the shell scripts as Python scripts.

Since this example *already* requires you to have a Python3 installation, this PR makes the example work cross-platform without introducing a new dependency on Python3.


daxpedda:
I would be fine with it since the server is already in Python. Though I dislike that we are using shell scripts or Python at all here.

Considering all this stuff needs nightly anyway, maybe we could use [cargo scripts](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#script) instead?

Though this LGTM, I'm not familiar with Python, so another maintainer should have a look at this.

Liamolucko:
This also seems fine to me.

I think it'd make sense to do the same thing for the `wasm-audio-worklet` example as well, since it also uses multithreading and as a result needs a basically identical build script and Python server. The same goes for #3606 and #3607.

daxpedda (approved):
Thanks!

Liamolucko (commented):
`wasm-audio-worklet`'s README should also be updated.

examples/wasm-audio-worklet/README.md
```diff
@@ -9,7 +9,7 @@ online][compiled]
 You can build the example locally with:
 
 ```
-$ ./run.sh
+$ python3 run.py
 ```
 
 (or running the commands on Windows manually)
```
Liamolucko, line 15: Get rid of this line.

jthemphill:
Almost all of this was merged in when you merged #3603, because I had rebased #3603 onto this branch. The only things left to merge in this PR now are the README changes. Sorry for not communicating this better😥

examples/raytrace-parallel/README.md
```diff
@@ -12,4 +12,4 @@ You can build the example locally with:
 $ python3 run.py
 ```
 
-and then visiting http://localhost:8080 in a browser should run the example!
+and then visiting http://localhost:8000 in a browser should run the example!
```
jthemphill, line 15: `wasm-audio-worklet` uses port 8080, but `raytrace-parallel` uses port 8000:

https://github.com/rustwasm/wasm-bindgen/blob/792be2d1faf7072fd51358558eacfc58b6c48642/examples/raytrace-parallel/server.py#L12

daxpedda: That's fine, thank you!

daxpedda (approved):
Again, apologies for messing up in #3606.

PR #3605, opened 2023-09-12
add .sha256 checksums to binary distribution tarballs

cormacrelf:
Hiya. This adds a `*.tar.gz.sha256` file alongside each tarball pushed to GitHub Releases. Having a hash available does not magically add much in terms of supply chain security, it just enables tools like Bazel and Buck to pull binaries and ensure they're the same one as last time. The concrete need is to generate a JSON manifest of a GitHub release including sha256 hashes, without actually downloading the tarballs and hashing them at generation time. I have not tested the pipeline, but note that:

1. GitHub's ubuntu-latest builder includes coreutils, so the sha256sum command should be available.
2. I checked that the `*` at the end of the glob will match those .sha256 suffixes as well as the tarballs, against the glob package in use. `npm i -g glob; sha256sum CHANGELOG.md > CHANGELOG.md.sha256; glob '*.md*'`

So I think that will work. This is not a blocker because scripts for generating manifests can always fall back on downloading the tarballs to /tmp and hashing them there. So no need to push a release just for this.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -486,7 +486,7 @@ jobs:
             tmp/$name/
           chmod +x tmp/$name/wasm*
           tar czvf gh-release/$name.tar.gz -C tmp $name
-          sha256sum gh-release/$name.tar.gz > gh-release/$name.tar.gz.sha256
+          sha256sum gh-release/$name.tar.gz > gh-release/$name.tar.gz.sha256sum
         }
         mk x86_64-unknown-linux-musl dist_linux_x86_64_musl
         mk aarch64-unknown-linux-gnu dist_linux_aarch64_gnu
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -486,6 +486,7 @@ jobs:
             tmp/$name/
           chmod +x tmp/$name/wasm*
           tar czvf gh-release/$name.tar.gz -C tmp $name
+          sha256sum gh-release/$name.tar.gz > gh-release/$name.tar.gz.sha256
         }
         mk x86_64-unknown-linux-musl dist_linux_x86_64_musl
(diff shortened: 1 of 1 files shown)


cormacrelf:
Here you go:

- https://buck2.build/docs/api/rules/#http_archive
- https://bazel.build/rules/lib/repo/http
- A generated [release manifest](https://github.com/facebook/buck2/blob/main/shim/third-party/proto/releases.bzl)
- An example of [doing it the slow way](https://github.com/facebook/buck2/blob/4be9f29d34edf261ded864310303b0325c0628ac/shim/third-party/proto/update.py#L72-L74) in the python script that generated the above file

daxpedda:
I don't these contain what I was looking for.

Specifically I was trying to find some standard how to deploy these hashes in the first place. Your PR suggests to deploy them as part of the release artifacts. Is **that** documented somewhere? Does a tool actually make use of this path, e.g. just adding `.sha256` at the end of the downloaded artifact's path to get its hash?

.github/workflows/main.yml
```diff
@@ -486,6 +486,7 @@ jobs:
             tmp/$name/
           chmod +x tmp/$name/wasm*
           tar czvf gh-release/$name.tar.gz -C tmp $name
+          sha256sum gh-release/$name.tar.gz > gh-release/$name.tar.gz.sha256
```
daxpedda, line 489: Thank you for putting this all together, very nice!
So following what you found that is already used in the wild, let's go with `*.sha256sum`, what do you say?
```suggestion
          sha256sum gh-release/$name.tar.gz > gh-release/$name.tar.gz.sha256sum
```

daxpedda (changes requested):

.github/workflows/main.yml
cormacrelf: No worries. Very well!

PR #3606, opened 2023-09-12
Update parallel raytracing .gitignore

jthemphill:
Running the build script creates some `*.d.ts` files and a `snippets` folder. Adding these to the `.gitignore`.


examples/raytrace-parallel/.gitignore
```diff
@@ -1,2 +1,5 @@
 raytrace_parallel.js
+raytrace_parallel.d.ts
 raytrace_parallel_bg.wasm
+raytrace_parallel_bg.wasm.d.ts
+snippets
```
daxpedda, line 5:
```suggestion
snippets

```

daxpedda (changes requested):
Couldn't we use the `--no-typescript` flag instead?
`snippets` would still be required though.

Liamolucko:
I'm not really sure why `raytrace-parallel` (and `wasm-audio-worklet`) puts its output into the current directory in the first place. All the other (non-Webpack-based) examples put their output into `pkg` instead: changing `raytrace-parallel` to do to the same would get rid of the need to even have this `.gitignore`, since `pkg` is already ignored by `examples/.gitignore`.

jthemphill:
> All the other (non-Webpack-based) examples put their output into `pkg` instead: changing `raytrace-parallel` to do to the same would get rid of the need to even have this `.gitignore`, since `pkg` is already ignored by `examples/.gitignore`.

Sure! But since this now involves changing `build.sh` or the future `build.py`, I'm going to rebase this atop #3603 to avoid skew.

examples/raytrace-parallel/.gitignore
```diff
@@ -1,2 +1 @@
-raytrace_parallel.js
-raytrace_parallel_bg.wasm
+snippets
```
daxpedda, line 1: I thought now that everything is in `pkg` we don't need the `.gitignore` at all?

examples/raytrace-parallel/index.html
daxpedda, line 1: Please leave out the HTML formatting from this PR. Same for the other file.

daxpedda (changes requested):

examples/raytrace-parallel/.gitignore
jthemphill: We still generate a `snippets/worker.js` file which is not in `pkg`. I'm not sure why it's there or how to move it into `pkg`. Do you not see this when you run `build.sh` or `build.py`?

daxpedda: Nope, for me the snippets folder lands in the `pkg` directory, as it should. Could you double check?

jthemphill: Oh, yes it is in `pkg` now!

daxpedda (approved):
Thank you!

jthemphill:
> Sure! But since this now involves changing `build.sh` or the future `build.py`, I'm going to rebase this atop #3603 to avoid skew.

Oh, um... heads up that I had rebased this on #3603. This is because I thought #3603 would be merged first, and because the `.gitignore` depended on us storing everything in `pkg`, which happened in a change I made to `build.py`.

The end result of all this is that you have already merged most of #3603, probably without realizing it😬

daxpedda:
... I sincerely apologize, I've somehow missed that, this is most definitely not your fault.
I will revert this so we can properly continue in #3603.

daxpedda:
I included this change in #3603, so you don't have to re-open this.

PR #3607, opened 2023-09-12
Add `rust-toolchain.toml` to raytracing example

jthemphill:
The raytracing example requires [a nightly build of Rust, and the source code to the Rust std library](https://rustwasm.github.io/docs/wasm-bindgen/examples/raytrace.html).

Running `build.sh` while your stable toolchain is active will result in an error:

```
error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
```

I tried changing `build.sh` to run `cargo +nightly`, but this broke CI.

My inclination here is to describe these dependencies with [a toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file). It's one less build flag to forget.


examples/raytrace-parallel/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "nightly"
+components = ["rust-src"]
+targets = ["wasm32-unknown-unknown"]
```
daxpedda, line 4:
```suggestion
targets = ["wasm32-unknown-unknown"]

```

daxpedda (changes requested):
Thank you, this is great!

examples/raytrace-parallel/rust-toolchain.toml
daxpedda, line 4:
```suggestion
targets = ["wasm32-unknown-unknown"]
profile = "minimal"
```
Would this work as well?

daxpedda (changes requested):
CI is failing because of the toolchain file.
I just took a quick look at the documentation to see if there is some flag to ignore it but couldn't really find anything. Would you mind exploring what to do here?

A quick solution that comes to mind here is to just use `+stable` to override the toolchain file.

examples/raytrace-parallel/rust-toolchain.toml
jthemphill: I suppose we can give it a try...

The error we're getting is

```
error: 'cargo-clippy' is not installed for the toolchain 'nightly-x86_64-unknown-linux-gnu'
```

And `profile = "minimal"`  avoids installing clippy.

I don't see how not installing clippy will help us with a "clippy not installed" error. Could someone change the CI action to run `cargo +stable clippy`, or to install nightly clippy before checking the examples that require a nightly build?

jthemphill:
Okay, I do think the clippy CI run needs to either install a nightly version of clippy, or it needs to change the command from `cargo clippy` to `cargo +stable clippy` to specify that we use stable clippy even when a`rust-toolchain.toml` file calls for nightly.

examples/raytrace-parallel/rust-toolchain.toml
daxpedda: Ah no, these two problems are unrelated. I was just asking if `profile = "minimal"` is enough to run this example, that's all!

daxpedda:
> `cargo +stable clippy`

I'm in favor of that.

daxpedda (approved):
Thanks.

PR #3610, opened 2023-09-13
Cell<usize> -> AtomicUsize in WasmRefCell

jthemphill:
I've been playing around with threads in Webassembly, but kept running into RefCell runtime errors:

```
Error: recursive use of an object detected which would lead to unsafe aliasing in rust
```

This was odd, because my code wasn't calling any mutable methods. It should be safe to call non-mutable methods from multiple threads simultaneously.

I dug into the generated webassembly and found that these were the result of calling `borrow_fail()` because `$var2 == -1`:

```wasm
    block $label0
      local.get $var0
      if
        local.get $var0
        i32.load
        local.tee $var2
        i32.const -1
        i32.eq
        br_if $label0
        local.get $var0
        local.get $var2
        i32.const 1
        i32.add
        i32.store
    ...
    end $label0
    call $wasm_bindgen::__rt::borrow_fail::h6f0a38fe489cbe71
```

which corresponds to [these lines](https://github.com/rustwasm/wasm-bindgen/blob/51e89ebee5bedb585464b492c6456048093054f8/src/lib.rs#L1474C2-L1498C10) in the `wasm_bindgen` source code.

After staring at the memory location that we load into `$var2`, it became clear that the memory location was underflowing, and that this was the result of a data race. `WasmRefCell` is not thread-safe!

After replacing the unsynchronized `Cell<usize>` with an `AtomicUsize` and patching the new `wasm_bindgen` crate into my project, I get this generated webassembly:

```wasm
    block $label1
      block $label0
        local.get $var0
        i32.eqz
        br_if $label0
        local.get $var0
        i32.atomic.load
        local.set $var1
        loop $label2
          local.get $var1
          i32.const -1
          i32.eq
          br_if $label1
          local.get $var0
          local.get $var1
          local.get $var1
          i32.const 1
          i32.add
          i32.atomic.rmw.cmpxchg
    ...
    end $label1
    call $wasm_bindgen::__rt::borrow_fail::h47c30cbcd4dcfd1d
```

And my code runs like a dream!

### DIFF
--- a/examples/thread-object-sharing/index.js
+++ b/examples/thread-object-sharing/index.js
@@ -1,0 +1,78 @@
+/** @type {import('./pkg/thread_object_sharing')} */
+import initWasm, { OpaqueObject } from "./pkg/thread_object_sharing.js";
+
+// Create a few webworkers, leaving some breathing room so our computer doesn't die
+const numWorkers = navigator.hardwareConcurrency / 2;
+const workers = [];
+for (let i = 0; i < numWorkers; ++i) {
+    workers.push(new Worker(
+        new URL('./worker.js', import.meta.url),
+        { name: 'Call OpaqueObject.get() a bunch of times', type: 'module' },
+    ));
+}
+
+// Load the webassembly and send our memory to the workers
+const wasmInternals = await initWasm();
+for (const worker of workers) {
+    worker.postMessage({ type: 'load', memory: wasmInternals.memory });
+}
+
+const opaqueObject = new OpaqueObject(47);
+
+// True iff webworkers are running
+let running = false;
+
+// Visual representation of opaqueObject.get()
+const valueMeter = window.document.getElementById('value-meter');
+const valueDisplayText = window.document.getElementById('value-display-text');
+
+// Change the rate at which the webworkers increment opaqueObject.get()'s value
+const incrementInput = window.document.getElementById('increment-input');
+incrementInput.onchange = (event) => {
+    const incrementVal = event.target.value;
+    for (const worker of workers) {
+        worker.postMessage({ type: 'increment', incrementVal });
+    }
+}
+
+/**
+ * Start the webworkers
+ */
+function start() {
+    for (const worker of workers) {
+        worker.postMessage({ type: 'start', opaqueObjectPtr: opaqueObject.__wbg_ptr });
+    }
+    running = true;
+}
+
+/**
+ * Stop the webworkers
+ */
+function stop() {
+    for (const worker of workers) {
+        worker.postMessage({ type: 'stop' });
+    }
+    running = false;
+}
+
+// Clickable button to start and stop the webworkers
+const toggleButton = window.document.getElementById('toggle-button');
+toggleButton.onclick = () => {
+    if (running) {
+        stop();
+    } else {
+        start();
+    }
+    toggleButton.innerText = running ? 'stop' : 'start';
+};
+toggleButton.removeAttribute('disabled');
+toggleButton.innerText = running ? 'stop' : 'start';
+
+function render() {
+    const val = opaqueObject.get();
+    valueMeter.setAttribute('value', val);
+    valueDisplayText.innerText = val.toString();
+
+    window.requestAnimationFrame(render);
+}
+window.requestAnimationFrame(render);
--- a/examples/thread-object-sharing/worker.js
+++ b/examples/thread-object-sharing/worker.js
@@ -1,0 +1,50 @@
+/** @type {import('./pkg/thread_object_sharing')} */
+import init, { OpaqueObject } from "./pkg/thread_object_sharing.js";
+
+/**
+ * @typedef {{type: 'load', memory: WebAssembly.Memory}} WorkerRequestLoad
+ * @typedef {{type: 'start', opaqueObjectPtr: number}} WorkerRequestStart
+ * @typedef {{type: 'stop'}} WorkerRequestStop
+ * @typedef {{type: 'increment', 'incrementVal': number}} WorkerRequestIncrement
+ * @typedef {WorkerRequestLoad | WorkerRequestStart | WorkerRequestStop | WorkerRequestIncrement} WorkerRequest
+ */
+
+/** @type {number | undefined} */
+let intervalTimer = undefined;
+let incrementVal = 10;
+
+/** @type {OpaqueObject} */
+const opaqueObject = Object.create(OpaqueObject.prototype);
+
+function tick() {
+    opaqueObject.add(incrementVal);
+}
+
+self.onmessage = async (event) => {
+    /**
+     * @type {WorkerRequest}
+     */
+    const request = event.data;
+
+    switch (request.type) {
+        case 'load':
+            await init(undefined, request.memory);
+            break;
+        case 'start':
+            opaqueObject.__wbg_ptr = request.opaqueObjectPtr;
+            if (intervalTimer === undefined) {
+                intervalTimer = self.setInterval(tick, 15);
+                console.log('starting intervalTimer');
+            }
+            break;
+        case 'stop':
+            console.log('stopping intervalTimer');
+            self.clearInterval(intervalTimer);
+            intervalTimer = undefined;
+            break;
+        case 'increment':
+            incrementVal = request.incrementVal;
+            console.log(`New increment value: ${incrementVal}`);
+            break;
+    }
+};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1399,9 +1399,10 @@ pub fn function_table() -> JsValue {
 pub mod __rt {
     use crate::JsValue;
     use core::borrow::{Borrow, BorrowMut};
-    use core::cell::{Cell, UnsafeCell};
+    use core::cell::UnsafeCell;
     use core::convert::Infallible;
     use core::ops::{Deref, DerefMut};
+    use core::sync::atomic::{AtomicUsize, Ordering};
 
     pub extern crate core;
     #[cfg(feature = "std")]
@@ -1452,7 +1453,7 @@ pub mod __rt {
     /// to not panic in libstd. Instead when it "panics" it calls our `throw`
     /// function in this crate which raises an error in JS.
     pub struct WasmRefCell<T: ?Sized> {
-        borrow: Cell<usize>,
+        borrow: AtomicUsize,
         value: UnsafeCell<T>,
     }
 
@@ -1463,7 +1464,7 @@ pub mod __rt {
         {
             WasmRefCell {
                 value: UnsafeCell::new(value),
-                borrow: Cell::new(0),
+                borrow: AtomicUsize::new(0),
             }
         }
 
@@ -1472,28 +1473,35 @@ pub mod __rt {
         }
 
         pub fn borrow(&self) -> Ref<T> {
-            unsafe {
-                if self.borrow.get() == usize::max_value() {
-                    borrow_fail();
-                }
-                self.borrow.set(self.borrow.get() + 1);
-                Ref {
-                    value: &*self.value.get(),
-                    borrow: &self.borrow,
-                }
+            match self
+                .borrow
+                .fetch_update(Ordering::Acquire, Ordering::Relaxed, |borrow| {
+                    (borrow != usize::max_value()).then(|| borrow + 1)
+                }) {
+                Ok(_) => unsafe {
+                    Ref {
+                        value: &*self.value.get(),
+                        borrow: &self.borrow,
+                    }
+                },
+                Err(_) => borrow_fail(),
             }
         }
 
         pub fn borrow_mut(&self) -> RefMut<T> {
-            unsafe {
-                if self.borrow.get() != 0 {
-                    borrow_fail();
-                }
-                self.borrow.set(usize::max_value());
-                RefMut {
-                    value: &mut *self.value.get(),
-                    borrow: &self.borrow,
-                }
+            match self.borrow.compare_exchange(
+                0,
+                usize::max_value(),
+                Ordering::Acquire,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => unsafe {
+                    RefMut {
+                        value: &mut *self.value.get(),
+                        borrow: &self.borrow,
+                    }
+                },
+                Err(_) => borrow_fail(),
             }
         }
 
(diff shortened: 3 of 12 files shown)


src/lib.rs
```diff
@@ -1472,28 +1473,52 @@ pub mod __rt {
...
+                        return Ref {
+                            value: &*self.value.get(),
+                            borrow: &self.borrow,
+                        };
+                    },
+                    Err(new_borrow) => {
+                        borrow = new_borrow;
+                    }
+                };
             }
```
daxpedda, line 1497: I'm assuming the `loop` and `compare_exchange()` is necessary because you want to make sure that your overflow check is actually correct.

But I would think that using `fetch_add()` should be good enough, because it returns the previous value, which then can be checked instead.

src/lib.rs
```diff
@@ -1528,13 +1553,13 @@ pub mod __rt {
 
     impl<'b, T: ?Sized> Drop for Ref<'b, T> {
         fn drop(&mut self) {
-            self.borrow.set(self.borrow.get() - 1);
+            self.borrow.fetch_sub(1, Ordering::SeqCst);
```
daxpedda, line 1556: Is there a reason we need `SeqCst` instead of `Relaxed` everywhere?

src/lib.rs
```diff
@@ -1472,28 +1473,52 @@ pub mod __rt {
...
+                        return RefMut {
+                            value: &mut *self.value.get(),
+                            borrow: &self.borrow,
+                        };
+                    },
+                    Err(new_borrow) => {
+                        borrow = new_borrow;
+                    }
+                };
             }
```
daxpedda, line 1522: Using a `compare_exchange()` here should do the trick without a `loop` and `load()`.
Also we shouldn't use the `weak` version of `compare_exchange()` unless we are fine with spurious fails, which we aren't, or am I missing something here?

daxpedda (changes requested):
Would you mind putting together some minimal example so I can look at the output code as well?
It would probably be prudent to use a `cfg` if the output on non-atomic Wasm is worse.

A changelog entry would be appreciated as well.

src/lib.rs
jthemphill: SGTM! We can `self.borrow.compare_exchange(0, usize::max_value())` and `borrow_fail()` if the `compare_exchange()` failed.

src/lib.rs
jthemphill: > But I would think that using `fetch_add()` should be good enough, because it returns the previous value, which then can be checked instead.

What if someone else sets `self.borrow.store(usize::max_value())` before we call `self.borrow.fetch_add(1)`? I think it would overflow to `0` and unlock the value! Sure we could check the previous value, but untold amounts of damage could happen while we're doing so.

src/lib.rs
jthemphill: Digging a bit deeper, it looks like [today, all atomic accesses in WebAssembly are SeqCst](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#atomic-memory-accesses):

> Currently all atomic memory access instructions are [sequentially consistent](https://en.wikipedia.org/wiki/Sequential_consistency). Instructions with other memory orderings may be provided in the future.

But what's the "right" ordering that's maximally correct and efficient, and remains so as we add weaker orderings to the WebAssembly spec?

I'm pretty sure `Relaxed` is wrong. Compilers are allowed to reorder other memory accesses across a `Relaxed` store or load, so the compiler could generate code that accesses `self.value` before we've confirmed that it's safe to access.

[The Rustonomicon's guide to Atomics](https://doc.rust-lang.org/nomicon/atomics.html) says that you should use `Acquire` to mark the beginning of a critical section and `Release` to mark the end:

> Intuitively, an acquire access ensures that every access after it stays after it. However operations that occur *before* an acquire are free to be reordered to occur *after* it. Similarly, a release access ensures that every access before it stays before it. However operations that occur *after* a release are free to be reordered to occur *before* it.

> Basic use of release-acquire is therefore simple: you acquire a location of memory to begin the critical section, and then release that location to end it.

We can also look at the memory orderings used for Rust's [`std::sync::RwLock`](https://doc.rust-lang.org/src/std/sys/unix/locks/futex_rwlock.rs.html#77-156) implementation. We are essentially rewriting `RwLock::try_read()` and `RwLock::try_write()`, but calling `borrow_fail()` when those functions would return `false`.

```
pub fn try_read(&self) -> bool {
    self.state
        .fetch_update(Acquire, Relaxed, |s| is_read_lockable(s).then(|| s + READ_LOCKED))
        .is_ok()
}

pub fn try_write(&self) -> bool {
    self.state
        .fetch_update(Acquire, Relaxed, |s| is_unlocked(s).then(|| s + WRITE_LOCKED))
        .is_ok()
}

pub unsafe fn read_unlock(&self) {
    let state = self.state.fetch_sub(READ_LOCKED, Release) - READ_LOCKED;

    // It's impossible for a reader to be waiting on a read-locked RwLock,
    // except if there is also a writer waiting.
    debug_assert!(!has_readers_waiting(state) || has_writers_waiting(state));

    // Wake up a writer if we were the last reader and there's a writer waiting.
    if is_unlocked(state) && has_writers_waiting(state) {
        self.wake_writer_or_readers(state);
    }
}

pub unsafe fn write_unlock(&self) {
    let state = self.state.fetch_sub(WRITE_LOCKED, Release) - WRITE_LOCKED;

    debug_assert!(is_unlocked(state));

    if has_writers_waiting(state) || has_readers_waiting(state) {
        self.wake_writer_or_readers(state);
    }
}
```

So I'm going to go ahead and rewrite this to use `Acquire` and `Release` orderings when we want to acquire and release our semaphore, using `RwLock` as a guide. But unfortunately this is all kind of theoretical and the generated wasm is always the same.

jthemphill:
Added a minimal repro! Try it before and after my WasmRefCell change. Also, try changing `OpaqueObject::get()` and `OpaqueObject::add()` to use `Ordering::Relaxed and observe how that changes the generated webassembly code😉

jthemphill:
[CI / build_examples](https://github.com/rustwasm/wasm-bindgen/actions/runs/6180222802/job/16776349094?pr=3610) fails because my new example requires nightly. Note that the command calls `grep -v raytrace | grep -v wasm-audio-worklet`, because it can't compile the other threading examples either...

jthemphill:
When I compile `OpaqueObject::get()` with a non-atomic `WasmRefCell`, I get this webassembly:

```wasm
(func $opaqueobject_get (;31;) (export "opaqueobject_get") (param $var0 i32) (result i32)
  (local $var1 i32)
  block $label1
    block $label0
      ;; check for null pointer
      local.get $var0
      i32.eqz
      br_if $label0
      ;; check if self.borrow == usize::max_value(), fail otherwise
      local.get $var0
      i32.load
      local.tee $var1
      i32.const -1
      i32.eq
      br_if $label1
        ;; self.borrow += 1
        local.get $var0
        local.get $var1
        i32.const 1
        i32.add
        i32.store
        ;; My function's code
        local.get $var0
        i32.atomic.load offset=4
        local.set $var1
        local.get $var0
        local.get $var0
      ;; self.borrow -= 1
      i32.load
      i32.const -1
      i32.add
      i32.store
      local.get $var1
      return
    end $label0
    call $wasm_bindgen::__rt::throw_null::h6303defe0ac2760e
    unreachable
  end $label1
  call $wasm_bindgen::__rt::borrow_fail::hf8796a362aaab80b
  unreachable
)
```

And when I run my example, all the workers get the `borrow_fail()` error:

```
Error: recursive use of an object detected which would lead to unsafe aliasing in rust
```

As a fun note on memory orderings, try changing `OpaqueObject::get()` to fetch and store with `Ordering::Relaxed`. You won't get the same bug! The compiler will cache the old value of `self.borrow()` instead of subtracting 1:

```wasm
(func $opaqueobject_get (;33;) (export "opaqueobject_get") (param $var0 i32) (result i32)
    (local $var1 i32)
    (local $var2 i32)
    block $label1
      block $label0
        ;; check for null pointer
        local.get $var0
        i32.eqz
        br_if $label0
        ;; check if self.borrow == usize::max_value(), fail otherwise
        local.get $var0
        i32.load
        local.tee $var1
        i32.const -1
        i32.eq
        br_if $label1
        local.get $var0
        ;; self.borrow += 1
        local.get $var1
        i32.const 1
        i32.add
        i32.store
        local.get $var0
        i32.atomic.load offset=4
        local.set $var2
        local.get $var0
        ;; Uses the old value stored in `$var1` instead of emitting `(i32.const -1) (i32.add)` instructions!
        local.get $var1
        i32.store
        local.get $var2
        return
      end $label0
      call $wasm_bindgen::__rt::throw_null::h6303defe0ac2760e
      unreachable
    end $label1
    call $wasm_bindgen::__rt::borrow_fail::hf8796a362aaab80b
    unreachable
  )
```

examples/thread-object-sharing/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+// Load the webassembly and send our memory to the workers
+const wasmInternals = await initWasm();
+for (const worker of workers) {
+    worker.postMessage({ type: 'load', memory: wasmInternals.memory });
+}
+
+const opaqueObject = new OpaqueObject(47);
+
+// True iff webworkers are running
```
daxpedda, line 22:
```suggestion
// True if webworkers are running
```

examples/thread-object-sharing/worker.js
```diff
@@ -0,0 +1,50 @@
...
+     * @type {WorkerRequest}
+     */
+    const request = event.data;
+
+    switch (request.type) {
+        case 'load':
+            await init(undefined, request.memory);
+            break;
+        case 'start':
+            opaqueObject.__wbg_ptr = request.opaqueObjectPtr;
```
daxpedda, line 34: Is this really a supported use-case?
This is going to become the big question in this PR to begin with: can this be triggered with supported code in the first place?

src/lib.rs
daxpedda: > I'm pretty sure `Relaxed` is wrong. Compilers are allowed to reorder other memory accesses across a `Relaxed` store or load, so the compiler could generate code that accesses `self.value` before we've confirmed that it's safe to access.

I don't believe this is correct, we don't have a "critical section" per se, we have a check for the returned value, this can't be reordered, because the `if` **has** to happen before it's body is executed.

I don't think `RwLock` is comparable because they have to sync access between multiple threads over a critical section, e.g. [this part](https://doc.rust-lang.org/1.72.0/src/std/sys/unix/locks/futex_rwlock.rs.html#202-218). We don't have that, for us it doesn't really matter who comes first as we don't coordinate state access atomically, we just give value access in certain conditions.

daxpedda (changes requested):

src/lib.rs
```diff
@@ -1472,28 +1473,35 @@ pub mod __rt {
...
+                    (borrow != usize::max_value()).then(|| borrow + 1)
+                }) {
+                Ok(_) => unsafe {
+                    Ref {
+                        value: &*self.value.get(),
+                        borrow: &self.borrow,
+                    }
+                },
+                Err(_) => borrow_fail(),
             }
```
daxpedda, line 1488: Continuation from https://github.com/rustwasm/wasm-bindgen/pull/3610#discussion_r1325773792:
the body of match will never be executed unless `fetch_update()` succeeds, this can't be reordered. So the question here is does this need to be coordinated with another access? As far as I can see: no.

E.g. dropping `RefMut` can happen before, in which case this succeeds, or happen after, in which case it doesn't succeed. Either outcome is satisfactory to us and doesn't allow unsafe behavior.

jthemphill:
> This is going to become the big question in this PR to begin with: can the issue trying to be fixed here be triggered with supported code in the first place?

I've [crossposted to internals.rust-lang.org](https://internals.rust-lang.org/t/should-wasm-bindgen-wasmrefcell-be-threadsafe/19567/2). Hopefully someone knowledgeable enough to help with this question will come along🤞

src/lib.rs
Liamolucko: > So the question here is does this need to be coordinated with another access? As far as I can see: no.

It does: it needs to be coordinated with any potential writes that another thread did while holding a `RefMut`. At the very least, the destructor for `RefMut` needs to use `Release` ordering, since otherwise the compiler (and processor) is free to reorder writes to `self.value` after setting `self.borrow` back to 0. And then that needs to match up with an `Acquire` here.

`Ref`'s destructor also still needs `Release` ordering because the compiler's free to reorder reads from `self.value` to happen after `self.borrow` gets set back to 0, by which point another thread might have already started modifying the value.

> the body of this `match` will never be executed unless `fetch_update()` succeeds, this can't be reordered.

You're probably right that the compiler can't reorder it, but I think it's still possible for subsequent reads from `self.value` to read outdated values anyway, e.g. if that value's cache line is still in the CPU's [invalidation queue](https://en.wikipedia.org/wiki/MESI_protocol#Invalidate_Queues).

I wouldn't be surprised if CPUs do all kinds of other weird stuff that I don't know about: I only found out about invalidation queues by skimming https://marabos.nl/atomics because this seemed sketchy. So I think we should try and make this work according to the theoretical memory model, rather than honing in too much on exactly what the compiler's doing.

Liamolucko (commented):
Putting the sketchiness of manually setting `__wbg_ptr` aside, I don't know if it's a great idea to create a multithreaded `RefCell` in the first place. Unless you only ever use immutable references, it'd be very easy to create non-deterministic panics with this.

To be fair, in this case only immutable references are ever used anyway, but I still think it's a bad idea to create a footgun like this. There's a reason that `std` doesn't have a thread-safe `RefCell`.

src/lib.rs
daxpedda: > At the very least, the destructor for `RefMut` needs to use `Release` ordering, since otherwise the compiler (and processor) is free to reorder writes to `self.value` after setting `self.borrow` back to 0.

> `Ref`'s destructor also still needs `Release` ordering because the compiler's free to reorder reads from `self.value` to happen after `self.borrow` gets set back to 0, by which point another thread might have already started modifying the value.

Why would reading/writing to `self.value` be able to be reordered at all? What am I missing here?

> So I think we should try and make this work according to the theoretical memory model, rather than honing in too much on exactly what the compiler's doing.

I agree, but I definitely don't know enough about atomic operations, maybe I should read that whole book, thanks for sharing!

daxpedda: https://doc.rust-lang.org/1.72.0/std/cell/struct.UnsafeCell.html:
> At all times, you must avoid data races. If multiple threads have access to the same `UnsafeCell`, then any writes must have a proper happens-before relation to all other accesses (or use atomics).

Is what I was looking for I guess.

daxpedda:
I'm going to close this as we don't plan to make `WasmRefCell` be shared between workers in the first place.
@jthemphill I hope we can resolve your needs in [the forum](https://internals.rust-lang.org/t/should-wasm-bindgen-wasmrefcell-be-threadsafe/19567).

Issue #3203, opened 2022-12-18
Use queueMicrotask instead of Promise.then in wasm-bindgen-futures


hamza1311:
This allows us to avoid issues that may arise when using Promises to queue microtasks. This explained well in [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide#enqueueing_microtasks):

> By introducing queueMicrotask(), the quirks that arise when sneaking in using promises to create microtasks can be avoided. For instance, when using promises to create microtasks, exceptions thrown by the callback are reported as rejected promises rather than being reported as standard exceptions. Also, creating and destroying promises takes additional overhead both in terms of time and memory that a function which properly enqueues microtasks avoids.

Also see

https://github.com/rustwasm/wasm-bindgen/blob/eebb201ac2310e2ca0319fbed51a9243cc350a64/crates/futures/src/queue.rs#L51-L53

Also see: #2392

snOm3ad:
This has been fixed by #3611

PR #3611, opened 2023-09-13
`wasm-bindgen-futures`: use `queueMicrotask` for next tick runs (#3203)

### LINKED ISSUE #2392: Can't catch async Wasm traps in JS
I'm trying to catch Wasm traps in Javascript with `window.addEventListener('error', (event) => { ... })`. This works for some traps, but not for those in [`spawn_local`](https://docs.rs/wasm-bindgen-futures/0.4/wasm_bindgen_futures/fn.spawn_local.html)ed tasks. For those, `window.addEventListener('unhandledrejection', (event) => { ... })` works in Firefox and Safari, but not Chrome or Edge. I guess it's related to https://github.com/getsentry/sentry-javascript/issues/2518, though in my case the Wasm is on the same origin. I'm using the JS generated by wasm-bindgen to load the Wasm.

Unless there's something I'm overlooking this may well be a bug in Chrome/Edge, but a solution/workaround for wasm-bindgen might be to try `queueMicrotask` [here](https://github.com/rustwasm/wasm-bindgen/blob/d6d222c9006d4c834042a1ac8cf2e3ebe5861232/crates/futures/src/queue.rs#L55-L60) and fall back to `Promise.then`? Or catch the promise rejection and [rethrow it using `setTimeout`](https://stackoverflow.com/a/30741722/750875) (ugly!)?

flumm:
but use `catch` to wrap it in a Result to be able to fall back to the previous `Promsise.then` variant.

since this is my first pull request here, i hope the code is ok this way, but a few questions remain:

* i simply imported queueMicrotask in global scope, since this should work for both non workers (window) and workers AFAICT
* i tried to run the tests mentioned here: https://rustwasm.github.io/docs/wasm-bindgen/contributing/testing.html but 
   the first one failed (regardless of my changes) and `cargo test -p ui-tests` also did not work (the remaining tests worked, as well as the ones directly in `wasm-bindgen-futures`
 * i tested #2392 with this, and this now seems to work
 * is the way i fall back to the old implementation ok? i couldn't find a way to import bindings 'optionally' or test during runtime if a function exists?
 * also is embedding such functionality from javascript directly ok? orwould it better be to add that to web-sys/js-sys first and use that?

Fixes #2392.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,7 +97,6 @@
   This should avoid quirks, like exceptions thrown get now properly reported
   as normal exceptions rather than as rejected promises.
   [#3611](https://github.com/rustwasm/wasm-bindgen/pull/3611)
-  [#2392](https://github.com/rustwasm/wasm-bindgen/issues/2392)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,13 @@
   `#[repr(C)]` types.
   [#3595](https://github.com/rustwasm/wasm-bindgen/pull/3595)
 
+* Use `queueMicrotask` in `wasm-bindgen-futures` for scheduling tasks on the next tick.
+  If that is not available, use the previous `Promise.then` mechanism as a fallback.
+  This should avoid quirks, like exceptions thrown get now properly reported
+  as normal exceptions rather than as rejected promises.
+  [#3611](https://github.com/rustwasm/wasm-bindgen/pull/3611)
+  [#2392](https://github.com/rustwasm/wasm-bindgen/issues/2392)
+
 ### Fixed
 
 * Fixed bindings and comments for `Atomics.wait`.
--- a/crates/futures/src/queue.rs
+++ b/crates/futures/src/queue.rs
@@ -8,11 +8,9 @@ use wasm_bindgen::prelude::*;
 extern "C" {
     #[wasm_bindgen]
     fn queueMicrotask(closure: &Closure<dyn FnMut(JsValue)>);
-}
 
-#[wasm_bindgen]
-extern "C" {
     type Global;
+
     #[wasm_bindgen(method, getter, js_name = queueMicrotask)]
     fn hasQueueMicrotask(this: &Global) -> JsValue;
 }
--- a/crates/futures/src/queue.rs
+++ b/crates/futures/src/queue.rs
@@ -4,6 +4,19 @@ use std::collections::VecDeque;
 use std::rc::Rc;
 use wasm_bindgen::prelude::*;
 
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen]
+    fn queueMicrotask(closure: &Closure<dyn FnMut(JsValue)>);
+}
+
+#[wasm_bindgen]
+extern "C" {
+    type Global;
+    #[wasm_bindgen(method, getter, js_name = queueMicrotask)]
+    fn hasQueueMicrotask(this: &Global) -> JsValue;
+}
+
 struct QueueState {
     // The queue of Tasks which are to be run in order. In practice this is all the
     // synchronous work of futures, and each `Task` represents calling `poll` on
@@ -42,17 +55,21 @@ pub(crate) struct Queue {
     state: Rc<QueueState>,
     promise: Promise,
     closure: Closure<dyn FnMut(JsValue)>,
+    has_queue_microtask: bool,
 }
 
 impl Queue {
     // Schedule a task to run on the next tick
     pub(crate) fn schedule_task(&self, task: Rc<crate::task::Task>) {
         self.state.tasks.borrow_mut().push_back(task);
-        // Note that we currently use a promise and a closure to do this, but
-        // eventually we should probably use something like `queueMicrotask`:
-        // https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask
+        // Use queueMicrotask to execute as soon as possible. If it does not exist
+        // fall back to the promise resolution
         if !self.state.is_scheduled.replace(true) {
-            let _ = self.promise.then(&self.closure);
+            if self.has_queue_microtask {
+                queueMicrotask(&self.closure);
+            } else {
+                let _ = self.promise.then(&self.closure);
+            }
         }
     }
     // Append a task to the currently running queue, or schedule it
@@ -70,6 +87,11 @@ impl Queue {
             tasks: RefCell::new(VecDeque::new()),
         });
 
+        let has_queue_microtask = js_sys::global()
+            .unchecked_into::<Global>()
+            .hasQueueMicrotask()
+            .is_function();
+
         Self {
             promise: Promise::resolve(&JsValue::undefined()),
 
@@ -82,6 +104,7 @@ impl Queue {
             },
 
             state,
+            has_queue_microtask,
         }
     }
 }


daxpedda:
> * i simply imported queueMicrotask in global scope, since this should work for both non workers (window) and workers AFAICT

That's fine.

> * i tried to run the tests mentioned here: https://rustwasm.github.io/docs/wasm-bindgen/contributing/testing.html but
>  the first one failed (regardless of my changes) and `cargo test -p ui-tests` also did not work (the remaining tests worked, as well as the ones directly in `wasm-bindgen-futures`

The contributor experience is awful in this regard and should be improved/fixed.

> * i tested [Can't catch async Wasm traps in JS #2392](https://github.com/rustwasm/wasm-bindgen/issues/2392) with this, and this now seems to work

Thanks, I added it to the OP.

> * is the way i fall back to the old implementation ok? i couldn't find a way to import bindings 'optionally' or test during runtime if a function exists?

We could use [`OnceLock`](https://doc.rust-lang.org/1.72.0/std/sync/struct.OnceLock.html), but v1.70 seems high right now and we don't have a proper MSRV policy. Alternatively we could use [`once_cell`](https://crates.io/crates/once_cell), but that would be adding another dependency.

@Liamolucko what do you think of adding `once_cell` to the dependency tree of `wasm-bindgen-futures`, personally I'm fine with it.
> * also is embedding such functionality from javascript directly ok? orwould it better be to add that to web-sys/js-sys first and use that?

Probably, but our story for `WindowOrGlobalWorker` is pretty weak, so I think it's better to leave it like that.

daxpedda (changes requested):
Would like to figure out first how we could cache `queueMicrotask` availability first, but otherwise LGTM.

Would you mind adding a note to the changelog as well?

Liamolucko:
> > * i tried to run the tests mentioned here: https://rustwasm.github.io/docs/wasm-bindgen/contributing/testing.html but
> >   the first one failed (regardless of my changes) and `cargo test -p ui-tests` also did not work (the remaining tests worked, as well as the ones directly in `wasm-bindgen-futures`
> 
> The contributor experience is awful in this regard and should be improved/fixed.

The specific two problems there are that the main `wasm32-unknown-unknown` tests need you to to specify `WASM_BINDGEN_SPLIT_LINKED_MODULES=1`, as there's one test that relies on `--split-linked-modules` being enabled, and that the UI tests need to be run with a specific version of Rust since error messages change between versions (currently v1.69.0).

The `WASM_BINDGEN_SPLIT_LINKED_MODULES` one is actually quite annoying; it might be a good idea to add that to `.cargo/config` so that you don't have to specify it every time.

But yeah, those should both be either documented or fixed.

> @Liamolucko what do you think of adding `once_cell` to the dependency tree of `wasm-bindgen-futures`, personally I'm fine with it.

I'm fine with that too.

flumm:
> We could use [`OnceLock`](https://doc.rust-lang.org/1.72.0/std/sync/struct.OnceLock.html), but v1.70 seems high right now and we don't have a proper MSRV policy. Alternatively we could use [`once_cell`](https://crates.io/crates/once_cell), but that would be adding another dependency.
> 
> @Liamolucko what do you think of adding `once_cell` to the dependency tree of `wasm-bindgen-futures`, personally I'm fine with it.
> 

maybe i'm missing part of the puzzle, but wouldn't it be enough to try executing it once (with e.g. an empty closure) in `new()` and save the result of `.is_err()` of that into the queue as struct member?

then we could do:

```
if <...> && (self.needs_fallback || queueMicrotask(...)) {
```
?
then it's a simple boolean check without `OnceLock` or `once_cell` ?
or how did you mean to use them?


> Would like to figure out first how we could cache `queueMicrotask` availability first, but otherwise LGTM.
> 
> Would you mind adding a note to the changelog as well?

sure no problem, do you prefer i push another commit, or should i force push my branch?
(what is preferred in general? AFAIR github mangles reviews when force pushing, but adding commits can be cumbersome when a pull request has many changes? Sorry for the stupid questions, but I'm used to
mailing list workflow, and seldom do github pull requests)

> The specific two problems there are that the main `wasm32-unknown-unknown` tests need you to to specify `WASM_BINDGEN_SPLIT_LINKED_MODULES=1`, as there's one test that relies on `--split-linked-modules` being enabled, and that the UI tests need to be run with a specific version of Rust since error messages change between versions (currently v1.69.0).
> 
> The `WASM_BINDGEN_SPLIT_LINKED_MODULES` one is actually quite annoying; it might be a good idea to add that to `.cargo/config` so that you don't have to specify it every time.
> 
> But yeah, those should both be either documented or fixed.

thanks that worked for the first test, but running `cargo test -p ui-tests` results here (rustc 1.69.0) with
> error: package ID specification `ui-tests` did not match any packages

daxpedda:
> > We could use [`OnceLock`](https://doc.rust-lang.org/1.72.0/std/sync/struct.OnceLock.html), but v1.70 seems high right now and we don't have a proper MSRV policy. Alternatively we could use [`once_cell`](https://crates.io/crates/once_cell), but that would be adding another dependency.
> > @Liamolucko what do you think of adding `once_cell` to the dependency tree of `wasm-bindgen-futures`, personally I'm fine with it.
> 
> maybe i'm missing part of the puzzle, but wouldn't it be enough to try executing it once (with e.g. an empty closure) in `new()` and save the result of `.is_err()` of that into the queue as struct member?
> 
> then we could do:
> 
> ```
> if <...> && (self.needs_fallback || queueMicrotask(...)) {
> ```
> 
> ? then it's a simple boolean check without `OnceLock` or `once_cell` ?
> or how did you mean to use them?

That would be better indeed!

Instead of calling `queueMicrotask()` with an empty closure, you should probably be able to do:
```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(getter)]
    fn queueMicrotask() -> JsValue;
}
```
Which returns `undefined` if not present. Never tried it with global, but should work just fine. Then we should also be able to remove the `catch` from the actualy `queueMicrotask()` bindings, which AFAIK can't actually throw.

flumm:
> That would be better indeed!
> 
> Instead of calling `queueMicrotask()` with an empty closure, you should probably be able to do:
> 
> ```rust
> #[wasm_bindgen]
> extern "C" {
>     #[wasm_bindgen(getter)]
>     fn queueMicrotask() -> JsValue;
> }
> ```
> 
> Which returns `undefined` if not present. Never tried it with global, but should work just fine. Then we should also be able to remove the `catch` from the actualy `queueMicrotask()` bindings, which AFAIK can't actually throw.

it seems that does not work, AFAICS the 'getter' property in the macro is ignored for non struct members

possible alternatives i saw were using the [`Window::get`](https://docs.rs/web-sys/0.3.64/web_sys/struct.Window.html#method.get) but i guess that does not work for workers anymore (and i couldn't find an alternative for workers...)

or using something like:

```rust
#[wasm_bindgen(inline_js = "export function queueMicrotaskExists() { return typeof queueMicrotask === 'function' }")]
extern "C" {
    fn queueMicrotaskExists() -> bool;
}
```

but that seems also overly complicated...

(also you did not answer my question about the pull request update, should i force push or add commits?)

Thanks!

daxpedda:
> possible alternatives i saw were using the [`Window::get`](https://docs.rs/web-sys/0.3.64/web_sys/struct.Window.html#method.get) but i guess that does not work for workers anymore (and i couldn't find an alternative for workers...)

We would want to avoid that because of the overhead from the string.
The solution is probably something like this:
```rust
#[wasm_bindgen]
extern "C" {
	type Global;

	#[wasm_bindgen(method, getter, js_name = queueMicrotask)]
	fn hasQueueMicrotask(this: &Global) -> JsValue;
}

let global: Global = js_sys::global().unchecked_into();
let has_queue_microtask = global.hasQueueMicrotask();
```

> (also you did not answer my question about the pull request update, should i force push or add commits?)

Apologies!
The PR will be squashed before merging anyway, so you can do w/e you like. Personally I only need a "nice" commit history if it makes reviewing easier, which for such a small PR is irrelevant to me.

flumm:
ah thanks, the 'Global' part was what i was missing, big thanks :)

i'll update my branch shortly

CHANGELOG.md
```diff
@@ -92,6 +92,13 @@
   `#[repr(C)]` types.
   [#3595](https://github.com/rustwasm/wasm-bindgen/pull/3595)
 
+* Use `queueMicrotask` in `wasm-bindgen-futures` for scheduling tasks on the next tick.
+  If that is not available, use the previous `Promise.then` mechanism as a fallback.
+  This should avoid quirks, like exceptions thrown get now properly reported
+  as normal exceptions rather than as rejected promises.
+  [#3611](https://github.com/rustwasm/wasm-bindgen/pull/3611)
+  [#2392](https://github.com/rustwasm/wasm-bindgen/issues/2392)
```
daxpedda, line 100: We only post the PR, not related issues, in the changelog.

crates/futures/src/queue.rs
```diff
@@ -4,6 +4,19 @@ use std::collections::VecDeque;
...
+    #[wasm_bindgen]
+    fn queueMicrotask(closure: &Closure<dyn FnMut(JsValue)>);
+}
+
+#[wasm_bindgen]
+extern "C" {
+    type Global;
+    #[wasm_bindgen(method, getter, js_name = queueMicrotask)]
+    fn hasQueueMicrotask(this: &Global) -> JsValue;
+}
```
daxpedda, line 18: Let's merge those two into the same block.

daxpedda (changes requested):

PR #3614, opened 2023-09-14
Fix example tests

Liamolucko:
I went to enable testing for the `synchronous-instantiation` example now that Firefox supports module workers, but then found that the `wasm-audio-worklet` example was failing because the test server doesn't set the headers needed to enable `SharedArrayBuffer`.

It turns out that CI wasn't failing because it's been broken this whole whole time: it specifies the path to the built examples as simply `exbuild`, which doesn't work because the tests are run with their working directory set to `crates/example-tests`, not the root of the repo. This means that any requests that the examples try to make will
404. So this PR specifies it as an absolute path instead.

At the moment, Firefox doesn't directly indicate any kind of error when navigation fails, which meant that the tests would just silently fail without actually testing anything. According to the spec, `browsingContext.navigate` is supposed to wait for the navigation to complete, and result in an error if something goes wrong; but I think Firefox is behind, because it seems to instead immediately return.

To work around this, I've made it so that the tests manually wait for the `network.responseCompleted` event to check if fetching the page suceeded, and so this shouldn't happen again.

~~I've left the actual fix for the `wasm-audio-worklet` example commented out to make sure that CI actually catches the issue now; that's why this PR is a draft.~~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -323,7 +323,7 @@ jobs:
     - run: rustup update --no-self-update stable && rustup default stable
     - run: cargo test -p example-tests
       env:
-        EXBUILD: ${{ env.GITHUB_WORKSPACE }}/exbuild
+        EXBUILD: ${{ github.workspace }}/exbuild
 
   build_benchmarks:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


daxpedda (approved):
I'm unfamiliar with WebDriver, but this LGTM.

Would prefer if we don't make so many inline structs, but this is an issue that was already there.

PR #3618, opened 2023-09-15
Lower the ES target for instanceof code

cschramm:
Since #3053 the code makes use of the ES2019 optional catch binding syntax, while generated code in general seems to be compatible with ES2015.


daxpedda (approved):
LGTM! Thank you!

PR #3620, opened 2023-09-15
Replace`filter_map().next()` with `find_map`

snOm3ad:
`rust-analyzer` emits a weak warning for this which leads to unnecessary noise in some editors.


Liamolucko (approved):
Thanks!

PR #3621, opened 2023-09-17
fix: handle raw identifier in field

### LINKED ISSUE #3586: Raw identifiers cause error
### Describe the Bug

Trying to use a struct with a raw identifier in wasm causes an error to be thrown.

### Steps to Reproduce

```rust
use wasm_bindgen::prelude::*;
#[wasm_bindgen]
pub struct Column {
  pub r#type: u8,
}
```

### Expected Behavior

I expect valid rust to compile to valid wasm.

### Actual Behavior

`rustc: custom attribute panicked`

### Additional Context

[Docs about raw identifiers](https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html)

Solo-steven:
fix #3586, using `Ident.unraw` to handle raw identifier.


CHANGELOG.md
```diff
@@ -103,6 +103,9 @@
 
 ### Fixed
 
+* Fixed `wasm_bindgen_macro_support` macro to handle raw identifiers in field names.
```
daxpedda, line 106:
```suggestion
* Fixed `wasm_bindgen` macro to handle raw identifiers in field names.
```

daxpedda (commented):
Do you think you can add a test for this somewhere into the existing ones?

Solo-steven:
Hi @daxpedda, sure! I think I can add a test in `tests/wasm/getter_and_setter.rs`, It is correctly place to add test case ?

daxpedda:
Apologies for the delay!
> I think I can add a test in `tests/wasm/getter_and_setter.rs`, It is correctly place to add test case ?

That seems fine to me.

Solo-steven:
Hi @daxpedda, it is ok! and thanks for your reply. There is a question for running tests when I follow the guide and run 
```
cargo test --target wasm32-unknown-unknown
```
my computer throws some error and shows that process panic,
```
Ok
driver status: signal: 9 (SIGKILL)                
Error: http://127.0.0.1:52624/session: status code 500
error: test failed, to rerun pass `--test headless`

Caused by:
  process didn't exit successfully: `cargo run -p wasm-bindgen-cli --bin wasm-bindgen-test-runner -- /Users/litingwei/Desktop/projects/for-osp/wasm-bindgen/target/wasm32-unknown-unknown/debug/deps/headless-4e40f55ce9c3cd6d.wasm` (exit status: 1)
```

Did I miss something to configure, I haven't changed the `webdriver.json` yet, thanks for any help, and sorry for bothering you again.

daxpedda:
The error usually has nothing to do with the `webdriver.json`, you could try using `WASM_BINDGEN_SPLIT_LINKED_MODULES=1`.
Honestly though, the problem here is missing documentation on how to run the tests properly on a local machine. If this doesn't solve your problem I recommend you to not waste any time on this and just use the CI to run tests.

> [..], thanks for any help, and sorry for bothering you again.

No, thank you for your contribution, you are helping us here!

tests/wasm/getters_and_setters.rs
```diff
@@ -32,6 +34,22 @@ pub struct Rules {
     pub field: i32,
 }
 
+#[wasm_bindgen]
+pub struct RulesWithRawField {
+    pub r#field: i32,
```
daxpedda, line 39: While we are at it, would you mind using a keyword just to test an actual use-case?
E.g.:
```suggestion
    pub r#mod: i32,
```

daxpedda (changes requested):

tests/wasm/getters_and_setters.rs
Solo-steven: @daxpedda sure! I am not sure if the test I wrote is correct or not, it seems we only need to test if `r#mod` field expanded by `#[wasm_bindgen]` macro is correct or not, so the beblow section seems unnecessary.
```
impl RulesWithRawField {
    #[wasm_bindgen]
    pub fn get_field_value(&self) -> i32 {
        self.r#field
    }
    #[wasm_bindgen]
    pub fn set_field_value(&self, value: i32) {
        self.r#field = value;
    }
}
// ...

```

daxpedda: I think we can leave this here, just to double check if the generated code in the methods can also handle raw identifiers.

tests/wasm/getters_and_setters.rs
```diff
@@ -32,6 +34,23 @@ pub struct Rules {
     pub field: i32,
 }
 
+#[wasm_bindgen]
+pub struct RulesWithRawField {
+    pub r#mod: i32,
+}
+#[wasm_bindgen]
+#[allow(non_snake_case)]
+impl RulesWithRawField {
```
daxpedda, line 43: Just some nits:
```suggestion
}

#[wasm_bindgen]
impl RulesWithRawField {
```

daxpedda (changes requested):
CI says formatting still has to be fixed.
Other then the nit LGTM.

Solo-steven:
> CI says formatting still has to be fixed. Other then the nit LGTM.

Hi @daxpedda, thanks for your review, I fixed the fmt problem, thanks ~

daxpedda (approved):
Thank you!

PR #3625, opened 2023-09-19
Unwrap argument types in macro

### LINKED ISSUE #3078: Macro expansion code, export function parameter is &T
I have a structure `TestStruct`,  When I take its reference as a parameter of an export function, it's OK, but once I put this export function in the macro, I will get the following error:
```rust
error[E0277]: the trait bound `&TestStruct: FromWasmAbi` is not satisfied
  --> src\style.rs:49:9
   |
49 |         #[wasm_bindgen]
   |         ^^^^^^^^^^^^^^^ the trait `FromWasmAbi` is not implemented for `&TestStruct`
...
55 | my_out_export!(&TestStruct);
   | --------------------------- in this macro invocation
```
### The Code:
```rust
#[wasm_bindgen]
pub struct TestStruct(usize);

#[wasm_bindgen]
pub fn a_right_example(s: &TestStruct) { println!("a_right_example") }
macro_rules! my_out_export {
    ($x: ty) => {
        #[wasm_bindgen]
        pub fn a_error_example(s: $x) {
			println!("a_error_example")
		}
    };
}
my_out_export!(&TestStruct);
```

snOm3ad:
`syn` wraps argument types in a `syn::TypeGroup` for free functions/methods generated from `macro_rules!` this means that references end up generating wrong code during proc macro expansion.

This PR unwraps argument types in `syn::TypeGroup` and matches against the inner element for such cases.

Fixes #3078


crates/backend/src/codegen.rs
```diff
@@ -523,8 +523,12 @@ impl TryToTokens for ast::Export {
             argtys.push(&*arg.ty);
             let i = i + offset;
             let ident = Ident::new(&format!("arg{}", i), Span::call_site());
-            let ty = &arg.ty;
-            match &*arg.ty {
+            let ty = match &*arg.ty {
+                syn::Type::Group(syn::TypeGroup { ref elem, .. }) => elem,
+                _ => &arg.ty,
+            };
```
Liamolucko, line 529: Could you also handle `Type::Paren` while you're at it?

Also, it might be a good idea to use a `while let` here in case there are multiple nested `Type::Paren`s / `Type::Group`s.

daxpedda (changes requested):
Could you add a test for this?

crates/cli/tests/wasm-bindgen/main.rs
daxpedda, line 1: Does testing this really require a CLI test? I thought the error is triggered during compile-time?

snOm3ad: Yes, it is triggered during compile time. Would it be better to put them in `tests/wasm`?

daxpedda: Yes, that sounds fine to to me.

daxpedda (approved):
Thanks!

PR #3631, opened 2023-09-23
Replace all instances of `Self` in methods

### LINKED ISSUE #3105: Returning Option<Self> in #[wasm_bindgen] impl block gives error
### Summary

This may not be the best example, but for simplicity I am adapting the code from this [issue](https://github.com/rustwasm/wasm-bindgen/issues/137). The problem is as follows:

The following return types work for `new` with the `#[wasm_bindgen]` attribute applied to the impl: `Universe`, `Self`, `Option<Universe>`. However, this return type appears not to work: `Option<Self>`. Is there some deep reason for this, or is this potentially a bug?

```rust
extern crate wasm_bindgen;

use wasm_bindgen::prelude::*;

#[repr(u8)]
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum Cell {
    Dead = 0,
    Alive = 1,
}

#[wasm_bindgen]
pub struct Universe {
    width: u32,
    height: u32,
    cells: Vec<Cell>,
}

/// Public methods, exported to JavaScript.
#[wasm_bindgen]
impl Universe {
    // The tutorial has the signature:
    // pub fn new() -> Universe {
    // This signature works:
    // pub fn new() -> Option<Universe> {
    pub fn new() -> Option<Self> {
        let width = 128;
        let height = 128;

        let cells = (0..width * height)
            .map(|i| {
                if i % 2 == 0 || i % 7 == 0 {
                    Cell::Alive
                } else {
                    Cell::Dead
                }
            })
            .collect();

        Some(Self {
            width,
            height,
            cells,
        })
    }
}
```

This gives the error:

```
error[E0401]: can't use generic parameters from outer function
   --> src\lib.rs:177:28
    |
139 | impl Universe {
    | ---- `Self` type implicitly declared here, by this `impl`
...
177 |     pub fn new() -> Option<Self> {
    |                            ^^^^
    |                            |
    |                            use of generic parameter from outer function
    |                            use a type here instead
```

The same problem arises when attempting to use `Self` with `Result` and other generics.

snOm3ad:
Currently the `wasm_bindgen` macro only replaces top level occurrences of the generic parameter `Self` in exported methods. This leads to an error when returning parametrized types containing `Self` due to Rust inner items rules.

This PR fixes the issue by drilling down the return type `TokenStream` and replacing all instances of `Self` with the actual type name.

Fixes #3105


crates/macro-support/src/parser.rs
```diff
@@ -803,6 +803,57 @@ pub(crate) fn is_js_keyword(keyword: &str) -> bool {
...
+                            }
+                        }
+                        _ => {}
+                    }
+                }
+            }
+        }
+    }
+
+    let path_str = quote::quote! { #path }.to_string();
```
daxpedda, line 849:
```suggestion
    let path_str = path.to_token_stream().to_string();
```

crates/macro-support/src/parser.rs
```diff
@@ -803,6 +803,57 @@ pub(crate) fn is_js_keyword(keyword: &str) -> bool {
...
+    if path_str.contains("Self") {
```
daxpedda, line 850: Couldn't we just call `walk_path_change_self()` and let it just do it's thing anyway? I'm assuming `to_token_stream()` and `to_string()` have to go through the `Path` to build it anyway, so it wouldn't be more expensive.

crates/macro-support/src/parser.rs
```diff
@@ -803,6 +803,57 @@ pub(crate) fn is_js_keyword(keyword: &str) -> bool {
...
+fn replace_self(ty: syn::Type, self_ty: Option<&Ident>) -> syn::Type {
+    let self_ty = match self_ty {
+        Some(i) => i,
+        None => return ty,
+    };
+
+    let mut path = match get_ty(&ty) {
+        syn::Type::Path(syn::TypePath { qself: None, path }) => path.clone(),
+        other => return other.clone(),
+    };
```
Liamolucko, line 816: This should also handle types other than paths.

crates/macro-support/src/parser.rs
snOm3ad: Yeah now that you mentioned it, seems like there is no way around it. Will push an update later today!

crates/macro-support/src/parser.rs
```diff
@@ -803,6 +803,57 @@ pub(crate) fn is_js_keyword(keyword: &str) -> bool {
...
+                });
+                for ty in types {
+                    match ty {
+                        syn::Type::Path(syn::TypePath { ref mut path, .. }) => {
+                            if path.segments.len() == 1 && path.segments[0].ident == "Self" {
+                                path.segments[0].ident = self_ty.clone();
+                            } else {
+                                walk_path_change_self(path, self_ty)
+                            }
+                        }
```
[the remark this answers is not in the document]
snOm3ad: Good catch! Yeah I'll try to handle all possible types. AFAIK references are not allowed in return types, but I did noticed this function is also called for types in arguments which do allow references. So maybe I'll function args differently.

snOm3ad:
I was also thinking of turning the type into a `String` then use `replace("Self", self_ty)` then parse it back into a `syn::Type`, this would have the advantage of not having to change this function when new types are supported for return types or argument types. Do you have any thoughts on this?

daxpedda:
> I was also thinking of turning the type into a `String` then use `replace("Self", self_ty)` then parse it back into a `syn::Type`, this would have the advantage of not having to change this function when new types are supported for return types or argument types. Do you have any thoughts on this?

I guess this could work and would obviously simplify the code a lot. The only thing I can think of that could use the word "Self" in a type is a macro, which I don't think is supported by Rust atm, or a const generic, Rust only supports integers right now.

My impression is that the additional code required to cover the remaining bases shouldn't be too bad. I'm not too worried about what else Rust might add in the future, it would be most likely a breaking change in `syn` anyway, so it's not like we would miss it.

Liamolucko:
> I was also thinking of turning the type into a `String` then use `replace("Self", self_ty)` then parse it back into a `syn::Type`, this would have the advantage of not having to change this function when new types are supported for return types or argument types. Do you have any thoughts on this?

That would also catch types that contain `Self` in their name though, e.g. `SelfPortrait`.

snOm3ad:
> > I was also thinking of turning the type into a `String` then use `replace("Self", self_ty)` then parse it back into a `syn::Type`, this would have the advantage of not having to change this function when new types are supported for return types or argument types. Do you have any thoughts on this?
> 
> That would also catch types that contain `Self` in their name though, e.g. `SelfPortrait`.

That's so sad! Oh well, will try to have an update on this later today. Thank you both for your comments!

PR #3639, opened 2023-10-03
Fix some calls to `free()` missing alignment

### LINKED ISSUE #3638: Alignment parameter is not passed to __wbindgen_free in js bindings.
https://github.com/rustwasm/wasm-bindgen/blob/f117406532fb2fd813d998e2ea98ad56e4a2ae4c/crates/cli-support/src/js/binding.rs#L1059

I suppose it should pass in `4` by default.

Built with: `--target wasm32-unknown-unknown -Z build
-std=std,panic_abort`
Generated binding with `--web`

daxpedda:
Apparently I missed some calls to `free()` in #3463.
@Liamolucko would appreciate if you'd take a look as well.

Thanks @Polish-Civil.
Fixes #3638.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,9 @@
 * Fix bug with function arguments coming from `macro_rules!`.
   [#3625](https://github.com/rustwasm/wasm-bindgen/pull/3625)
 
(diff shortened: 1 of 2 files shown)


Liamolucko (approved):
LGTM!

PR #3642, opened 2023-10-05
Add `sendEncodings` in `RTCRtpTransceiverInit`

logist322:
This PR adds [spec compliant](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiverinit-sendencodings) `sendEncodings` field in `RTCRtpTransceiverInit`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


daxpedda (changes requested):
Otherwise LGTM, thank you!

Issue #3530, opened 2023-07-20
Request to release a new version of web-sys


stephanemagnenat:
### Motivation

I need [`WorkerGlobalScope.performance`](https://github.com/rustwasm/wasm-bindgen/commit/4aae6b38e6bed37d3b908f6e99611549840ab395) for adding timing support in web workers in [Rust CV](https://github.com/rust-cv/cv/pull/75).

### Proposed Solution

A new minor release.

daxpedda:
I would have liked to fix https://github.com/rustwasm/wasm-bindgen/pull/3483 first, but unfortunately we don't seem to have the time.

@stephanemagnenat if you replicate https://github.com/rustwasm/wasm-bindgen/pull/3475 it would help to get a release now instead of later.

stephanemagnenat:
@daxpedda thanks for the fast answer. Is there a script to bump minor versions, or is it a manual process so far?

daxpedda:
AFAIK a manual process. It should be straightforward with a couple of search-and-replace.

gabrielDevlog:
Hi,

A release would be great. Is there any ETA for this ?

Thanks

daxpedda:
There is none AFAIK.

You can get the process started by replicating #3475.

gabrielDevlog:
Ok i'll try. Should i follow this process: https://rustwasm.github.io/docs/wasm-bindgen/contributing/publishing.html

daxpedda:
If you just scroll up and look at the earlier discussion you will find all your answers.

> Ok i'll try. Should i follow this process ? https://rustwasm.github.io/docs/wasm-bindgen/contributing/publishing.html

I've done it manually so far because I didn't know about this tool, feel free to try it though.

hamza1311:
@daxpedda do you have permissions to make a release? I was thinking of setting up a GitHub actions workflow for it, but I don't have the permissions on crates.io and a secret needs be added before publishing can be done

daxpedda:
I do not, but the way it worked until now is to set everything up and ping Alex Crichton.
Making an action for it would be ideal though!

gabrielDevlog:
Well I'm trying manually but its not as easy as it seems. How do you get this value "APPROVED_SCHEMA_FILE_HASH" ?

https://github.com/rustwasm/wasm-bindgen/pull/3475/files#diff-515c29e8b5cfb75260c999069bc92df2bf301e4bbeddafd858b269cfac79d9f4R11

[...]

PR #3647, opened 2023-10-07
Add enum types for function parameters and return types

### LINKED ISSUE #2154: Typescript enum definitions are not used in function signatures

ChristianIvicevic:
The adjusted typescript-tests unfortunately aren't great to assert the correctness of the generated types because a generated `(a: number) => void` could still be assigned to a `(a: Foo) => void` as the enum values in `Foo` are merely `number`s. Therefore the tests currently only asserts that something structurally compatible is generated, but there is no check whether the expected enum is referenced. Classic TypeScript...

Fixes #2154

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -1384,7 +1384,7 @@ impl ToTokens for ast::Enum {
         let enum_name = &self.rust_name;
         let name_str = self.js_name.to_string();
         let name_len = name_str.len() as u32;
-        let name_chars = name_str.chars().map(|c| c as u32).collect::<Vec<_>>();
+        let name_chars = name_str.chars().map(|c| c as u32);
         let hole = &self.hole;
         let cast_clauses = self.variants.iter().map(|variant| {
             let variant_name = &variant.name;
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -1382,6 +1382,9 @@ impl<'a> ToTokens for DescribeImport<'a> {
 impl ToTokens for ast::Enum {
     fn to_tokens(&self, into: &mut TokenStream) {
         let enum_name = &self.rust_name;
+        let name_str = self.js_name.to_string();
+        let name_len = name_str.len() as u32;
+        let name_chars = name_str.chars().map(|c| c as u32).collect::<Vec<_>>();
         let hole = &self.hole;
         let cast_clauses = self.variants.iter().map(|variant| {
             let variant_name = &variant.name;
@@ -1433,6 +1436,8 @@ impl ToTokens for ast::Enum {
                 fn describe() {
                     use #wasm_bindgen::describe::*;
                     inform(ENUM);
+                    inform(#name_len);
+                    #(inform(#name_chars);)*
                     inform(#hole);
                 }
             }
--- a/crates/typescript-tests/src/enums.ts
+++ b/crates/typescript-tests/src/enums.ts
@@ -6,3 +6,8 @@ const a3: wbg.Foo.A = 1;
 const b1: wbg.Foo = wbg.Foo.B;
 const b2: wbg.Foo.B = wbg.Foo.B;
 const b3: wbg.Foo.B = 3;
+
+const fn_expects_enum: (_: wbg.Foo) => void = wbg.fn_expects_enum;
+const fn_returns_enum: () => wbg.Foo = wbg.fn_returns_enum;
+const fn_expects_option_enum: (_?: wbg.Foo | undefined) => void = wbg.fn_expects_option_enum;
+const fn_returns_option_enum: () => wbg.Foo | undefined = wbg.fn_returns_option_enum;
--- a/crates/typescript-tests/src/enums.ts
+++ b/crates/typescript-tests/src/enums.ts
@@ -9,5 +9,5 @@ const b3: wbg.Foo.B = 3;
 
 const fn_expects_enum: (_: wbg.Foo) => void = wbg.fn_expects_enum;
 const fn_returns_enum: () => wbg.Foo = wbg.fn_returns_enum;
-const fn_expects_option_enum: (_?: wbg.Foo | undefined) => void = wbg.fn_expects_option_enum;
+const fn_expects_option_enum: (_?: wbg.Foo) => void = wbg.fn_expects_option_enum;
 const fn_returns_option_enum: () => wbg.Foo | undefined = wbg.fn_returns_option_enum;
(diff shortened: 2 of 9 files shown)


crates/backend/src/codegen.rs
```diff
@@ -1382,6 +1382,9 @@ impl<'a> ToTokens for DescribeImport<'a> {
 impl ToTokens for ast::Enum {
     fn to_tokens(&self, into: &mut TokenStream) {
         let enum_name = &self.rust_name;
+        let name_str = self.js_name.to_string();
+        let name_len = name_str.len() as u32;
+        let name_chars = name_str.chars().map(|c| c as u32).collect::<Vec<_>>();
```
daxpedda, line 1387: I believe `quote!` can handle iterators? If it does there would be no need to use `collect()`.

crates/typescript-tests/src/enums.ts
```diff
@@ -6,3 +6,8 @@ const a3: wbg.Foo.A = 1;
 const b1: wbg.Foo = wbg.Foo.B;
 const b2: wbg.Foo.B = wbg.Foo.B;
 const b3: wbg.Foo.B = 3;
+
+const fn_expects_enum: (_: wbg.Foo) => void = wbg.fn_expects_enum;
+const fn_returns_enum: () => wbg.Foo = wbg.fn_returns_enum;
+const fn_expects_option_enum: (_?: wbg.Foo | undefined) => void = wbg.fn_expects_option_enum;
```
daxpedda, line 12:
```suggestion
const fn_expects_option_enum: (_?: wbg.Foo) => void = wbg.fn_expects_option_enum;
```

daxpedda (changes requested):
As far as I can tell our testing suite actually doesn't verify the TS output against something we write.

In any case, LGTM!

daxpedda (approved):

ChristianIvicevic:
@daxpedda Please allow me some questions regarding your release process and testing/using pre-release versions of the crates.

1. What is your typical release cadence and is there any plans on when to release next?
2. I wanted to give my new changes a try and replaced my local dependency in my project with a path to my fork
```toml
wasm-bindgen = { git = "https://github.com/ChristianIvicevic/wasm-bindgen.git", branch = "2154-enums-in-function-signatures" }
```
Running wasm-pack yielded an error
```
thread 'main' panicked at 'unknown descriptor: 66', /Users/christian.ivicevic/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasm-bindgen-cli-support-0.2.87/src/descriptor.rs:164:22
```
This made me realize that there might be some mismatch with my globally installed version of wasm-pack. Is this expected? How can I properly use my fork until a proper release is available?

daxpedda:
> 1. What is your typical release cadence and is there any plans on when to release next?

See https://github.com/rustwasm/wasm-bindgen/issues/3530.

> 2. I wanted to give my new changes a try and replaced my local dependency in my project with a path to my fork

You can't without also installing `wasm-bindgen-cli` from the same custom fork.
`cargo install wasm-bindgen-cli --git https://github.com/ChristianIvicevic/wasm-bindgen --branch 2154-enums-in-function-signatures -f` (not sure if that `-f` is needed)

I'm not sure exactly how `wasm-pack` functions, but if it directly depends on a `wasm-bindgen-*` crate, it might need to be forked as well.

ChristianIvicevic:
I'll give that manual override a try and if it doesn't work I'll wait until either the user in the linked issue has opened a release PR; otherwise, I'll look into it myself 😅 Thanks for pointing me to that issue.

ChristianIvicevic:
Alright, your proposal to override the globally installed CLI (even I didn't had it installed previously) has worked and the output is correct. I'll stick with that until an official release has been pushed.

Thanks again for the support.

Liamolucko:
> Running wasm-pack yielded an error
> 
> ```
> 
> thread 'main' panicked at 'unknown descriptor: 66', /Users/christian.ivicevic/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasm-bindgen-cli-support-0.2.87/src/descriptor.rs:164:22
> 
> ```

Oh - this PR should've bumped the schema version (in `wasm-bindgen-shared`), since adding the `name` field to `Descriptor::Enum` means that `wasm-bindgen`'s now incompatible with the previous version.

This is why you got this cryptic error message rather than a proper 'version mismatch' error.

daxpedda:
> Oh - this PR should've bumped the schema version (in `wasm-bindgen-shared`), [..]

I'm not totally familiar with how this works. I know that when we bump the version for a new release this will get a new schema version anyway.

As far as I understand it though, this PR didn't update the `shared/lib.rs` file, which is why the schema version was not bumped. Is there some other way to update the schema version or how is this supposed to be done?

Liamolucko:
> I'm not totally familiar with how this works. I know that when we bump the version for a new release this will get a new schema version anyway.

No, it's done manually. In principle, it's possible for the interface between the macro and the CLI to stay the same between two versions of `wasm-bindgen`, in which case the schema version shouldn't be bumped: it should be left as the previous version of `wasm-bindgen`.

Then you would be able to freely use, for example, version 0.2.87 of the CLI with version 0.2.88 of the macro. However since changes to the interface between the macro and CLI are very common, this hasn't happened in a long time - not since I've been involved, at the very least.

> As far as I understand it though, this PR didn't update the `shared/lib.rs` file, which is why the schema version was not bumped. Is there some other way to update the schema version or how is this supposed to be done?

Yeah, in principle the schema version is supposed to just keep track of changes to the schema in that file (hence the name). But since the version mismatch system is built on top of it, it's morphed into a general way of keeping track of changes in the interface between the macro and the CLI.

daxpedda:
Thanks, this makes sense to me now!

So how exactly to we bump the schema version if nothing changes in `shared/lib.rs`? Do we just bump this:
https://github.com/rustwasm/wasm-bindgen/blob/bfe644fe13ee84b190345c53d419a0919a4a946f/crates/shared/src/lib.rs#L9

It's a bit weird to me to bump the version there without bumping the version in `wasm-bindgen` in general.

Liamolucko:
> So how exactly to we bump the schema version if nothing changes in `shared/lib.rs`? Do we just bump this:
> 
> https://github.com/rustwasm/wasm-bindgen/blob/bfe644fe13ee84b190345c53d419a0919a4a946f/crates/shared/src/lib.rs#L9

Yes, and then update the hash.

PR #3648, opened 2023-10-09
rename main -> start on `wasm-bindgen(start)` example

kazuk:
I got error with this example.

```
error: entry symbol `main` declared multiple times
  |
  = help: did you use `#[no_mangle]` on `fn main`? Use `#[start]` instead
```

In this document, "main" is specified as the function name. However, compilation will fail due to the previously mentioned error if "main" is used. This is presumably because "main" is a function name reserved in Rust as the starting function for binaries.

rename to `start` then succeeded.

### DIFF
--- a/guide/src/reference/attributes/on-rust-exports/start.md
+++ b/guide/src/reference/attributes/on-rust-exports/start.md
@@ -6,7 +6,7 @@ soon as the wasm module is instantiated.
 
 ```rust
 #[wasm_bindgen(start)]
-fn main() {
+fn start() {
     // executed automatically ...
 }
 ```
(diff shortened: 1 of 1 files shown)


daxpedda (changes requested):
This is a bit surprising to me, but I believe it's better this way anyway to differentiate `#[wasm_bindgen(start)]` from `#[wasm_bindgen(main)]`.

There are bunch of mentions to the `main()` functions in this file, those would need to be updated to `start` as well.

kazuk:
I re-test this. I got build error on `wasm-pack test`.

So `main` symbol conflicting with rust's `cargo test` generated main.

daxpedda (approved):
Thank you!

PR #3655, opened 2023-10-13
fix ABI for raw pointers

jrvidal:
:wave: Hello from Stackblitz

We've noticed very buggy behavior when our Rust code requires a sizable amount of memory (above 2 GiB!). I've traced it back to `__wbindgen_malloc` returning negative pointers (!!), which screams "`i32` overflow" :grimacing: 

From what I can gather, changing the `WasmDescribe` impl. for raw pointers (so they follow `u32` ABI conventions instead of `i32`'s) should fix it. I've added a `references` test to `wasm-bindgen-cli`, not sure how appropriate/useful is.

### DIFF
--- a/crates/cli/tests/reference/pointers.d.ts
+++ b/crates/cli/tests/reference/pointers.d.ts
@@ -1,0 +1,12 @@
+/* tslint:disable */
+/* eslint-disable */
+/**
+* @param {number} input
+* @returns {number}
+*/
+export function const_pointer(input: number): number;
+/**
+* @param {number} input
+* @returns {number}
+*/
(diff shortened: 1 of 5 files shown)


daxpedda (approved):
Yeah, this definitely looks like a bug to me:
https://github.com/rustwasm/wasm-bindgen/blob/46f52e97f2b62d10dcaed248e6217b15742d27f1/src/convert/impls.rs#L208-L242

@Liamolucko would you mind confirming this?

Liamolucko (approved):
Yep, I agree this is a bug. Thanks!

d3lm:
Is there a chance we can get this released?

PR #3656, opened 2023-10-13
try_reserve_exact has been stabilized nearly two years ago

mvduin:
use Vec's try_reserve_exact method (stabilized in rust 1.57) instead of an ad-hoc reimplementation thereof


hamza1311 (approved):
Agreed, we should use the stdlib version

PR #3666, opened 2023-10-23
Update the node version in the GitHub workflows from 16 to 20

thomasetter:
This required removing the --experimental-wasm-reftypes flag since it was removed from v8 in
https://github.com/v8/v8/commit/1771e4aaa3f285b3829539c1f80dbde01500fc21 which is included in Node `>= 18`


daxpedda (approved):
Thank you!

PR #3667, opened 2023-10-24
Updated Notification APIs

maccesch:


maccesch:
I don't exactly understand why this on check didn't go through. I followed the contribution guidelines. What can I do to fix this?

daxpedda:
This is just missing the `Cargo.toml` adjustments, looking at the CI results you can easily see what needs to be done:
```diff
--- a/crates/web-sys/Cargo.toml
+++ b/crates/web-sys/Cargo.toml
@@ -880,6 +880,7 @@ NodeFilter = []
 NodeIterator = []
 NodeList = []
 Notification = ["EventTarget"]
+NotificationAction = []
 NotificationBehavior = []
 NotificationDirection = []
 NotificationEvent = ["Event", "ExtendableEvent"]
```
> I don't exactly understand why this on check didn't go through. I followed the contribution guidelines. What can I do to fix this?

You probably forgot to point the tool towards the `Cargo.toml` file. The issue is that the book isn't updated, but you find the updated version here: https://rustwasm.github.io/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html.
Note the last argument `./Cargo.toml`.

maccesch:
I replaced the whole idl file now with the latest spec from that link. I hope that's ok? It makes some breaking changes but the previous version of the notification API seems to have been very old.

crates/web-sys/webidls/enabled/Notification.webidl
```diff
@@ -1,60 +1,32 @@
...
-
-  [Pure]
-  readonly attribute DOMString? icon;
-
-  [Constant, Func="mozilla::dom::DOMPrefs::NotificationRIEnabled"]
+  readonly attribute DOMString lang;
+  readonly attribute DOMString body;
+  readonly attribute DOMString tag;
+  readonly attribute USVString image;
+  readonly attribute USVString icon;
```
daxpedda, line 21: Unfortunately we can't just introduce breaking changes like that.
```suggestion
  readonly attribute DOMString? lang;
  readonly attribute DOMString? body;
  readonly attribute DOMString? tag;
  readonly attribute USVString image;
  readonly attribute USVString? icon;
```

crates/web-sys/webidls/enabled/Notification.webidl
```diff
@@ -87,10 +53,16 @@ enum NotificationPermission {
   "granted"
 };
 
-callback NotificationPermissionCallback = undefined (NotificationPermission permission);
-
 enum NotificationDirection {
-  "auto",
-  "ltr",
+  "auto" ,
+  "ltr" ,
```
daxpedda, line 58:
```suggestion
  "auto",
  "ltr",
```

crates/web-sys/webidls/enabled/Notification.webidl
daxpedda, line 1: Please add a newline at the end.

daxpedda (changes requested):
Unfortunately we can't just correct the API like this because that would require a breaking change.

I think removing `GetNotificationOptions`, `NotificationBehavior` and `Notification.get()` is fine though because they don't actually exist anymore. @Liamolucko and @hamza1311 would you be fine with that as well?

Could you also add a note in the changelog? Especially mention the removal of these APIs separately.

daxpedda (changes requested):

Liamolucko:
> I think removing `GetNotificationOptions`, `NotificationBehavior` and `Notification.get()` is fine though because they don't actually exist anymore. @Liamolucko and @hamza1311 would you be fine with that as well?

Yeah, I'm fine with this.

hamza1311:
I'm fine with it. If it "breaks" any code, that code is already broken and now there is an error message for it at compile time.

PR #3669, opened 2023-10-25
Return an error if there are two enums with the same name

thomasetter:
This fixes the issue that two enums with the same name cause the "second" one to overwrite the "first" one.

Currently, looking at the output of `target/wasm32-unknown-unknown/wbg-tmp-wasm-<some-hex-characters>.wasm/wasm-bindgen-test.js` generated with 
```
WASM_BINDGEN_SPLIT_LINKED_MODULES=1 cargo test --target wasm32-unknown-unknown
```

the part with the enums (was at line 4849 for me) looks like this:

```js
**
*/
module.exports.Color = Object.freeze({ Green:0,"0":"Green",Yellow:1,"1":"Yellow",Red:2,"2":"Red", });
/**
*/
module.exports.JsRenamedEnum = Object.freeze({ A:10,"10":"A",B:20,"20":"B", });
/**
*/
module.exports.EnumArrayElement = Object.freeze({ Unit:0,"0":"Unit", });
/**
* annotated enum type
*/
module.exports.AnnotatedEnum = Object.freeze({
/**
* annotated enum variant 1
*/
Variant1:0,"0":"Variant1",
/**
* annotated enum variant 2
*/
Variant2:1,"1":"Variant2", });
/**
*/
module.exports.Color = Object.freeze({ Green:0,"0":"Green",Yellow:1,"1":"Yellow",Red:2,"2":"Red", });
/**
*/
module.exports.ColorWithCustomValues = Object.freeze({ Green:21,"21":"Green",Yellow:34,"34":"Yellow",Red:2,"2":"Red", });
/**
*/
module.exports.MyEnum = Object.freeze({ One:1,"1":"One",Two:2,"2":"Two", });
```

The second assignment `module.exports.Color =`  overwrites the first one, causing the first one to have no effect.

The reason this was not captured by the tests is that the two enums were identical in shape (same unit keys/values).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,9 @@
   enhancing type safety and compatibility.
   [#3647](https://github.com/rustwasm/wasm-bindgen/pull/3647)
 
+* Stop tolerating name collisions on enums, this would previously just emit one of them.
+  [#3669](https://github.com/rustwasm/wasm-bindgen/pull/3669)
+
 ### Fixed
 
 * Fixed `wasm_bindgen` macro to handle raw identifiers in field names.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,7 +112,7 @@
   enhancing type safety and compatibility.
   [#3647](https://github.com/rustwasm/wasm-bindgen/pull/3647)
 
-* Stop tolerating name collisions on enums, this would previously just emit one of them.
+* Throw an error on enum name collisions, previously only one enum would be emitted.
   [#3669](https://github.com/rustwasm/wasm-bindgen/pull/3669)
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


daxpedda (changes requested):
LGTM!
Could you add a note in the changelog as well?

thomasetter:
Thanks for the quick review, added a note in the changelog under `Changed`.

CHANGELOG.md
```diff
@@ -112,6 +112,9 @@
   enhancing type safety and compatibility.
   [#3647](https://github.com/rustwasm/wasm-bindgen/pull/3647)
 
+* Stop tolerating name collisions on enums, this would previously just emit one of them.
```
daxpedda, line 115:
```suggestion
* Throw an error on enum name collisions, previously only one enum would be emitted.
```

Issue #3090, opened 2022-09-21
Add the OpenSSF Scorecard GitHub Action


gabibguti:
### Motivation

Hello, I'm working on behalf of Google and the [Open Source Security Foundation][ossf] to help essential open-source projects improve their supply-chain security. Given wasm-bindgen importance to integrate Rust Wasm modules and JavaScript, we want to offer our support to improve the project's security posture.

### Proposed Solution

Would you consider adopting an OpenSSF tool called [Scorecards][sc]? Scorecards runs dozens of automated security checks to help maintainers better understand their project's supply-chain security posture. It is developed by the OpenSSF, [in partnership with GitHub][sc-blog].

I see wasm-bindgen already follows security best practices, such as release signing and CI testing. Adding Scorecards could help encourage other best practices, such as pinning dependencies and using fuzzing tools. It also checks for actions that increase risk, such as not setting least privilege permissions on tokens and using dangerous code patterns on GitHub Workflows. Overall, it could be a great tool to guide security decisions and show off your current hard work on security practices.

Would you be interested in a PR which adds this Action? Optionally, it can also publish your results to the OpenSSF REST API, which allows a [badge][badge] with the project's score to be added to its README.

### Additional Context

The [Scorecard GitHub Action](https://github.com/ossf/scorecard-action) is very lightweight and runs on each change to the repository's main branch. The results of its checks are available on to the project's [security dashboard](https://github.com/rustwasm/wasm-bindgen/security), and include suggestions on how to solve any issues (see examples below). The Action does not run or interact with any workflows, but merely parses them to identify possible vulnerabilities. Over 1600 projects have added the action already.

![Code scanning dashboard with multiple alerts, including Code-Review and Token-Permissions][img-security]

![Detail of a Token-Permissions alert, indicating the specific file and remediation steps][img-detail]

[badge]: https://openssf.org/blog/2022/09/08/show-off-your-security-score-announcing-scorecards-badges/
[ossf]: https://openssf.org/
[sc]: https://github.com/ossf/scorecard
[sc-blog]: https://github.blog/2022-01-19-reducing-security-risk-oss-actions-opensff-scorecards-v4/
[sc-gha]: https://github.com/ossf/scorecard-action
[img-security]: https://user-images.githubusercontent.com/15221358/190184391-84ca1844-259a-4b3b-9c86-74adadbea7f1.png
[...]

PR #3670, opened 2023-10-25
Add Scorecard

gabibguti:
Resolves https://github.com/rustwasm/wasm-bindgen/issues/3090

This PR adds [Scorecard](https://github.com/ossf/scorecard) security tool. Thanks for trying the tool out! And let me know if you have any feedback.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -63,7 +63,6 @@ jobs:
         with:
           name: SARIF file
           path: results.sarif
-          retention-days: 5
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,72 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '25 15 * * 3'
+  push:
+    branches: [ "main" ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@0864cf19026789058feabb7e87baa5f140aac736 # v2.3.1
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          # (Optional) "write" PAT token. Uncomment the `repo_token` line below if:
+          # - you want to enable the Branch-Protection check on a *public* repository, or
+          # - you are installing Scorecard on a *private* repository
+          # To create the PAT, follow the steps in https://github.com/ossf/scorecard-action#authentication-with-pat.
+          # repo_token: ${{ secrets.SCORECARD_TOKEN }}
+
+          # Public repositories:
+          #   - Publish results to OpenSSF REST API for easy access by consumers
+          #   - Allows the repository to include the Scorecard badge.
+          #   - See https://github.com/ossf/scorecard-action#publishing-results.
+          # For private repositories:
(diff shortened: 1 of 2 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
daxpedda, line 35: Is there a reason why we aren't just using the published versions?

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+          publish_results: true
+
+      # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
+      # format to the repository Actions tab.
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
```
daxpedda, line 66:
```suggestion
```
Don't see why we don't just leave it up.

daxpedda:
Can we also add the badge here or can we do this only as a follow-up?

.github/workflows/scorecard.yml
gabibguti: One of the security best practices Scorecard recommends is using commit SHAs to reference dependency versions instead of, in this case, branches like `main` or tags like `v4`. The reason is that branches and tags get automatically the latest updates and with that it can get a malicious commit without notice.

.github/workflows/scorecard.yml
gabibguti: Sure.

daxpedda:
I'm looking at the results here: https://securityscorecards.dev/viewer/?uri=github.com/rustwasm/wasm-bindgen.

Is there some way to add exceptions to binary artifacts? Currently it includes stuff from examples and benchmarks.
SAST is also an issue, we do use Clippy, but as far as I can see OSFF doesn't support detecting that.

.github/workflows/scorecard.yml
daxpedda: As far as I can see OSFF doesn't even support Cargo projects, is that correct?
In any case, seeing that this is a library pinning dependencies is imo a terrible idea.

I feel similarly about GitHub Actions dependencies, they should definitely not be pinned in a public project, where zero-day exploits can be used against us by anybody who can trigger the CI workflow, which is more or less everybody.

daxpedda (commented):
LGTM to me, would only like to see the GitHub Actions dependencies unpinned, this is a security practice I don't really agree is appropriate for `wasm-bindgen`, I elaborated on this above.

OSFF doesn't really support Cargo project, so I'm starting to be a bit skeptical of it's usefulness here. Overall, I'm still fine with this.

@alexcrichton I feel like you should give your stamp of approval here as well.

alexcrichton:
I'd personally be a bit wary of this given the amount of configuration required in this repository in addition to some of the results. Some of the aspects on the security card I don't understand (e.g. I haven't heard of SAST before) and some aren't really that applicable (fuzzing for this repo, pinned dependencies, wasms-interpreted-as-binary, signed releases, etc). 

All that being said I'm not too too involved in wasm-bindgen any more so I don't think I should count as a "final say" of sorts, I think it's fine to add this if you'd like to @daxpedda

daxpedda:
@hamza1311 @Liamolucko, you guys have any opinion on that?
I'm currently not against it but not for it as well, pretty much for the same reasons Alex Crichton pointed out: the result doesn't seem very representative to me for `wasm-bindgen`.

So unless there is any further approval I'm gonna close this.

.github/workflows/scorecard.yml
gabibguti: > As far as I can see OSFF doesn't even support Cargo projects, is that correct?

Yes, true. The Scorecard Pinned-Dependencies checks does not identify Cargo dependencies. I will report that for Scorcard and suggest the documentation is improved to inform which languages/tools are supported.

> In any case, seeing that this is a library pinning dependencies is imo a terrible idea.

I partially agree. I think there are pros and cons. My current view is that pinning build and release dependencies is still "safer" for these processes. But pinning the library dependencies and test dependencies is not something I agree with too. So, pinning GitHub actions dependencies in this workflow would still make sense for me but I understand it comes with a maintainance cost.

gabibguti:
In any case, thanks for the valuable feedback!

hamza1311:
> The Scorecard Pinned-Dependencies checks does not identify Cargo dependencies

This reduces the usefulness by _a lot_.

Also, from [the docs](https://github.com/ossf/scorecard/blob/49c0eed3a423f00c872b5c3c9f1bbca9e8aae799/docs/checks.md#dangerous-workflow)
> Risk: Critical (vulnerable to repository compromise)
>
> This check determines whether the project's GitHub Action workflows has dangerous code patterns.

image

> ### Dangerous-Workflow (critical risk)
> 
> no dangerous workflow patterns detected

This makes me question the effectiveness of the product. It reports dangerous workflow as critical risk but then says it didn't find any dangerous workflow patterns?

At this point:
- it can't detect supply chain vulnerabilities
- it can't filter results to actual library code
- it can't properly report the issues (see above)
- static code analysis does not use clippy -- what does it use then? if it's better than clippy, then effort should be put to making clippy better. We already run clippy so it's not too important for us

I'm with @daxpedda on closing the PR

daxpedda:
Alright, as their are no

daxpedda:
Alright, as there are no further approvals, I'm going to close this.

Thanks @gabibguti nonetheless.

Issue #3421, opened 2023-05-11
WASI support


daxpedda:
This issue will be used to collect discussions and issues about WASI support in `wasm-bindgen`.

Generally speaking `wasm-bindgen` doesn't currently support WASI, but we should figure out if we want to support it (https://github.com/rustwasm/wasm-bindgen/pull/3324) or disable it (https://github.com/rustwasm/wasm-bindgen/pull/3233).

Input from stakeholders would help a lot!

csnover:
A library I was working with compiled and linked to a C library using the `cc` crate, and I believe that wasm-bindgen is not usable in that case unless it supports WASI. (I say ‘believe’ because I was never able to find a channel that was still active to get confirmation from someone authoritative.)

daxpedda:
I guess this is one use-case: having WASI dependencies.

But I would argue that this is really a bad workaround, C libraries can target Wasm without WASI or Emscripten just fine. If any WASI function is used it wouldn't work in the browser anyway without a compatibility shim.

On the other hand I understand that the only Wasm target C/C++ libraries usually support is Emscripten or WASI if they are more modern. As hacky as I described this above, it's a workaround many actually require.

Though I've honestly no clue if you can use a WASI dependency successfully or not with `wasm-bindgen`, so more input would be appreciated. Thanks @csnover.

csnover:
Getting a WASI provider into the browser that covered the required system calls was trivial with https://github.com/bjorn3/browser_wasi_shim and required less than 10 LOC change to the bootstrap JS to make it send the required imports when instantiating the module, so this is no barrier.

The third-party C dependency has no knowledge of wasm and used system calls which are available in the WASI shim; I do not understand the comment about it being a ‘bad workaround’ so would be interested to understand more where you are coming from in this regard.

I am not paying attention to the wasm ecosystem to understand what is what, but as an outsider exploring the space for the first time, it looked like WASI being positioned as the standard ABI. Maybe that is just an wrong impression though?

csnover:
I understand what you were saying now, thank you for sharing.

For what it is worth I don’t have much skin in the game at all here; I was just told by someone recently that the emscripten target for Rust did not work, took that claim at face value, saw that WASI was being developed under the WebAssembly group, and assumed this meant the ecosystem had agreed to move toward WASI as the universal ABI for runtimes.

[...]

PR #3673, opened 2023-10-30
Conditionally enable bindings for `wasm32-wasi` with `--wasi` flag

temeddix:
With #3233(Jan 2023), generating bindings for WASI was disabled.

- #3233

However, with #3595(Sep 2023) `wasm-bindgen` gained compatibility with `wasm32-wasi`'s standard C ABI. It was stated in #3454 by @Liamolucko that `wasm-bindgen` already supports the standard C ABI that `wasm32-wasi` uses.

- #3595 
- https://github.com/rustwasm/wasm-bindgen/issues/3454#issuecomment-1782638916

This PR aims to enable bindings for `wasm32-wasi` under the new `--wasi` flag. This will allow ecosystem around `wasm-bindgen` such as `wasm-pack` to make use of `wasm32-wasi`.

I'm very open to feedback, so please tag me if there's room for improvements. I should perhaps add or modify some tests for now.

### DIFF
--- a/crates/backend/build.rs
+++ b/crates/backend/build.rs
@@ -1,6 +1,4 @@
-// This `build.rs` makes `[package] links = ...` work in `Cargo.toml`.
 fn main() {
-    println!("cargo:rerun-if-changed=build.rs");
     let enable_wasi_env_var = std::env::var("WASM_BINDGEN_ENABLE_WASI").unwrap_or_default();
     if enable_wasi_env_var == "1" {
         println!("cargo:rustc-cfg=enable_wasi");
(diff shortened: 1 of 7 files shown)


daxpedda:
Personally I have stated in https://github.com/rustwasm/wasm-bindgen/issues/3421 why I think enabling `wasm32-wasi` in `wasm-bindgen` **directly** is a mistake.

So I think the way forward here is to continue discussion in #3421 to see if we can find a good reason to actually support that.

daxpedda:
Also, please make sure it makes sense to move forward with a PR before you make one with us. It feels really bad closing PRs like that, I prefer not to waste anybodies time.

PR #3674, opened 2023-10-31
Add workflow to bump versions and create releases

hamza1311:
This PR adds a new workflow called bump. This workflow is to be triggered manually when a new release is to be published. The process of releases is as follows:

1. Run the bump flow from `master`
2. Review/merge the bump PR
3. Run the publish flow from `bump` branch

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -1,0 +1,34 @@
+name: Bump packages
+
+permissions:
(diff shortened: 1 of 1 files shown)


daxpedda:
Do we really need this two branch system? Can we not just bump and publish from master?
I'm assuming somebody still has to actually add the crates.io token?

PR #3680, opened 2023-11-02
Update release workflow

hamza1311:
Combine bump and release into a single step

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -5,20 +5,10 @@ permissions:
 
 on:
   workflow_dispatch:
-    inputs:
-      action:
-          description: 'bump or publish'
-          required: true
-          default: 'bump'
-          type: choice
-          options:
-            - bump
-            - publish
 
 jobs:
   bump:
     runs-on: ubuntu-latest
-    if: github.event.inputs.action == 'bump'
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/bump.yml
```diff
@@ -28,24 +18,13 @@ jobs:
...
-  publish:
-    runs-on: ubuntu-latest
-    if: github.event.inputs.action == 'publish'
-    steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - run: rustc publish.rs
-      - name: assert we're on bump branch
-        run: test "$(git git branch --show-current)" = "bump"
+      - run: git commit -S -F /tmp/msg
```
hamza1311, line 26: Will GitHub handle `-S` for signed commits properly? Not sure

daxpedda: No clue, worth the try.

daxpedda (approved):
Thank you for doing this!

.github/workflows/bump.yml
Systemcluster: See https://github.com/actions/runner/issues/667

As far as I understand you can't with the default git config, but there are some workarounds.

PR #3695, opened 2023-11-09
Allow SharedArrayBuffer to be passed to DataView constructor

### LINKED ISSUE #3694: Support Passing a SharedArrayBuffer to DataView
### Motivation

Modern browsers allow a regular `ArrayBuffer` as well as a `SharedArrayBuffer` to be passed to the constructor of `DataView`. Currently, the wasm-bindgen version `DataView` allow only a `ArrayBuffer` to be passed.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/DataView#buffer

### Proposed Solution

Update the implementation of `DataView` to accept a `&JsValue`  instead of a `&ArrayBuffer`.

### Alternatives

N/A

### Additional Context

N/A

kyledecot:
Fixes #3694 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/DataView#buffer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## [Unreleased](https://github.com/rustwasm/wasm-bindgen/compare/0.2.88...main)
+
+### Added
+
+* Allow SharedArrayBuffer to be passed to DataView constructor
+  [#3695](https://github.com/rustwasm/wasm-bindgen/pull/3695)
+
 ## [0.2.88](https://github.com/rustwasm/wasm-bindgen/compare/0.2.87...0.2.88)
 
 Released 2023-11-01
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Added
 
-* Allow SharedArrayBuffer to be passed to DataView constructor
+* Add additional constructor to `DataView` for `SharedArrayBuffer`.
   [#3695](https://github.com/rustwasm/wasm-bindgen/pull/3695)
 
 ## [0.2.88](https://github.com/rustwasm/wasm-bindgen/compare/0.2.87...0.2.88)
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -1596,7 +1596,7 @@ extern "C" {
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView)
     #[wasm_bindgen(constructor)]
-    pub fn new(buffer: &ArrayBuffer, byteOffset: usize, byteLength: usize) -> DataView;
+    pub fn new(buffer: &JsValue, byteOffset: usize, byteLength: usize) -> DataView;
```
daxpedda, line 1599: AFAIK this isn't [officially](https://doc.rust-lang.org/1.73.0/cargo/reference/semver.html) considered a breaking change, but it can still actually break code.

Though I believe I would be fine with that, I think it's better to just introduce another constructor function, like we do in other APIs.

daxpedda (changes requested):

crates/js-sys/src/lib.rs
kyledecot: No problem. As this is my first contribution to this project I'm unfamiliar w/ naming conventions, etc. Where should I look for a good example of where something similar was done so I can maintain consistency?

daxpedda (changes requested):
One last thing, apologies, I keep forgetting that:
could you add an entry in the changelog?

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## [Unreleased](https://github.com/rustwasm/wasm-bindgen/compare/0.2.88...main)
+
+### Added
+
+* Allow SharedArrayBuffer to be passed to DataView constructor
```
```suggestion
* Add additional constructor to `DataView` for `SharedArrayBuffer`.
```

daxpedda (approved):
Thank you!

PR #3703, opened 2023-11-12
`__wbindgen_thread_destroy` has optional params

AlexErrant:
This PR updates the generated TypeScript to reflect the fact that `__wbindgen_thread_destroy` has optional params.

Ref: https://github.com/rustwasm/wasm-bindgen/blob/9fb3bca16876c756266274f78fcd0214e0581eaa/crates/threads-xform/src/lib.rs#L394-L395

Before:

```ts
export interface InitOutput {
  readonly start: () => void;
  readonly run: () => void;
  readonly child_entry_point: (a: number) => void;
  readonly __wbindgen_malloc: (a: number, b: number) => number;
  readonly __wbindgen_realloc: (a: number, b: number, c: number, d: number) => number;
  readonly __wbindgen_exn_store: (a: number) => void;
  readonly __wbindgen_thread_destroy: (a: number, b: number) => void;
  readonly __wbindgen_start: () => void;
}
```

After:
```ts
export interface InitOutput {
  readonly start: () => void;
  readonly run: () => void;
  readonly child_entry_point: (a: number) => void;
  readonly __wbindgen_malloc: (a: number, b: number) => number;
  readonly __wbindgen_realloc: (a: number, b: number, c: number, d: number) => number;
  readonly __wbindgen_exn_store: (a: number) => void;
  readonly __wbindgen_thread_destroy: (a?: number, b?: number) => void;
  readonly __wbindgen_start: () => void;
}
```

Ideally I'd like better names too (`a` and `b`) but for the life of me I can't tell if it's in the wasm or not. Maybe just hardcode it since I'm already doing `__wbindgen_thread_destroy`? 🤷‍♂️ Oh well this is a strict improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,16 @@
 
 * Add additional constructor to `DataView` for `SharedArrayBuffer`.
   [#3695](https://github.com/rustwasm/wasm-bindgen/pull/3695)
-* Remove experimental status from `wasm_bindgen::module()` and update docs to say it works with `--target web`.
+
+* Stabilize `wasm_bindgen::module()`.
   [#3690](https://github.com/rustwasm/wasm-bindgen/pull/3690)
 
 ### Fixed
 
 * The DWARF section is now correctly modified instead of leaving it in a broken state.
   [#3483](https://github.com/rustwasm/wasm-bindgen/pull/3483)
-* Update the TypeScript signature of `__wbindgen_thread_destroy` to indicate that it has optional parameters.
+
+* Update the TypeScript signature of `__wbindgen_thread_destroy` to indicate that it's parameters are optional.
   [#3703](https://github.com/rustwasm/wasm-bindgen/pull/3703)
 
 ## [0.2.88](https://github.com/rustwasm/wasm-bindgen/compare/0.2.87...0.2.88)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,15 @@
 
 * Add additional constructor to `DataView` for `SharedArrayBuffer`.
   [#3695](https://github.com/rustwasm/wasm-bindgen/pull/3695)
+* Remove experimental status from `wasm_bindgen::module()` and update docs to say it works with `--target web`.
+  [#3690](https://github.com/rustwasm/wasm-bindgen/pull/3690)
 
 ### Fixed
 
 * The DWARF section is now correctly modified instead of leaving it in a broken state.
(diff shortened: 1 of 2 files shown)


daxpedda (changes requested):
This is unfortunately hacky :/

As far as I can tell we don't have more information available because we manually insert these functions into the module instead of exposing them with the `wasm_bindgen` macro. We could still actually add `wasm-bindgen` style descriptors manually to address this ... but it's questionable if that's worth it.

In any case, this is an improvement nevertheless, I don't see why not to take it.

Could you add a line to the changelog?

AlexErrant:
Also added a changelog for https://github.com/rustwasm/wasm-bindgen/pull/3690

CHANGELOG.md
```diff
@@ -7,11 +7,15 @@
...
 * Add additional constructor to `DataView` for `SharedArrayBuffer`.
   [#3695](https://github.com/rustwasm/wasm-bindgen/pull/3695)
+* Remove experimental status from `wasm_bindgen::module()` and update docs to say it works with `--target web`.
+  [#3690](https://github.com/rustwasm/wasm-bindgen/pull/3690)
 
 ### Fixed
 
 * The DWARF section is now correctly modified instead of leaving it in a broken state.
   [#3483](https://github.com/rustwasm/wasm-bindgen/pull/3483)
+* Update the TypeScript signature of `__wbindgen_thread_destroy` to indicate that it has optional parameters.
```
daxpedda, line 17:
```suggestion

* Update the TypeScript signature of `__wbindgen_thread_destroy` to indicate that it's parameters are optional.
```

CHANGELOG.md
```diff
@@ -7,11 +7,15 @@
 
 * Add additional constructor to `DataView` for `SharedArrayBuffer`.
   [#3695](https://github.com/rustwasm/wasm-bindgen/pull/3695)
+* Remove experimental status from `wasm_bindgen::module()` and update docs to say it works with `--target web`.
```
daxpedda, line 10:
```suggestion

* Stabilize `wasm_bindgen::module()`.
```
I don't think we ever added entries for documentation changes, so let's not start it here.

daxpedda (changes requested):

daxpedda:
Thank you!

linonetwo:
Hi, I'm searching for solution for #3818 

May I ask, is this renamed to `__wbindgen_free`?

daxpedda:
This function is made to destroy spawned threads, not the main instance.

PR #3704, opened 2023-11-13
Add a test runner to two of the typescript tests to actually run them

thomasetter:
This adds Jest as the test runner framework.
I have chosen this framework because it was the framework I managed to get running and because it seems reasonably popular:
https://2022.stateofjs.com/en-US/libraries/testing/

This uncovers that some of the tests are unrealistic code which would crash at runtime.

Before this PR, these tests were just compiled with tsc, but never run.

This PR also adds the new test script to the `crates/typescript-tests/run.sh` file which makes them run in the CI.


hamza1311 (commented):
Can you also update CI to run the tests

thomasetter:
Good point!

They are already running in the CI, as can be seen in the last few lines of `Run cd crates/typescript-tests && ./run.sh` in [this pipeline run](https://github.com/rustwasm/wasm-bindgen/actions/runs/6851664380/job/18628444982?pr=3704).

I clarified this in the PR description by mentioning the `run.sh`.

PR #3678, opened 2023-11-01
Fix ambiguous associated type error when enum has variant called Error

mlwilkerson:
Closes #3677

PR #3709, opened 2023-11-15
Replace TryFrom with custom TryFromJsValue trait

### LINKED ISSUE #3685: 0.2.88 implements TryFrom on all ABI-bound objects, breaking API bindings that have custom TryFrom implementations
`0.2.88` automatically derives `TryFrom` trait for any struct that uses `#[wasm_bindgen]`.

This prevents any developer from creating custom `TryFrom<JsValue>` conversions for the ABI-bound struct.

We have a rather large Rust/JS API that uses `TryFrom` to provide custom processing for JS objects. For example, JS code can pass a Rust struct, a buffer, or a hex string, which the `TryFrom` resolves. This is extremely useful because JS objects can be of different types. The object received from JS does not need to be an ABI-bound object in order to be converted to a native Rust struct.

This addition makes it no longer possible to create custom `TryFrom` handling between `JsValue` and a Rust struct that is also WASM ABI-bound.

I believe that such imposition should not be present in the attribute macro, or it should be possibly gated by a feature or a macro attribute. It does not seem quite right that an attribute macro creates functionality that is typically created by derive macros.  

This is semantically not different than (for example) `wasm_bindgen` implementing a custom `Clone` trait, preventing the developer from being able to implement custom `Clone` functionality for his objects...

Perhaps there should be a `FromJsValue` derive macro available?  Or such conversion should be moved into a custom `try_from_js_abi()` function implemented on each ABI-bound object? (implementing a function will create much less of a potential interference with underlying frameworks).

biryukovmaxim:
This commit replaces the usage of the standard library's TryFrom trait with a custom TryFromJsValue trait added for handling the conversion from JavaScript values in WASM. This trait has been implemented for different types across multiple modules to ensure consistent error handling on failed conversions.

fixes #3685

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@
 
 * The DWARF section is now correctly modified instead of leaving it in a broken state.
   [#3483](https://github.com/rustwasm/wasm-bindgen/pull/3483)
-* Custom TryFromJsValue Trait: Fixed an issue where #[wasm_bindgen] automatically derived the TryFrom trait for any struct, preventing custom TryFrom<JsValue> implementations. The wasm-bindgen derive macro has been updated to utilize the new TryFromJsValue trait.    [#3709](https://github.com/rustwasm/wasm-bindgen/pull/3709)
 
+* Fixed an issue where `#[wasm_bindgen]` automatically derived the `TryFrom` trait for any struct, preventing custom `TryFrom<JsValue>` implementations. It has been updated to utilize a new `TryFromJsValue` trait instead.
+  [#3709](https://github.com/rustwasm/wasm-bindgen/pull/3709)
 
 * Update the TypeScript signature of `__wbindgen_thread_destroy` to indicate that it's parameters are optional.
   [#3703](https://github.com/rustwasm/wasm-bindgen/pull/3703)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 * The DWARF section is now correctly modified instead of leaving it in a broken state.
   [#3483](https://github.com/rustwasm/wasm-bindgen/pull/3483)
+* Custom TryFromJsValue Trait: Fixed an issue where #[wasm_bindgen] automatically derived the TryFrom trait for any struct, preventing custom TryFrom<JsValue> implementations. The wasm-bindgen derive macro has been updated to utilize the new TryFromJsValue trait.    [#3709](https://github.com/rustwasm/wasm-bindgen/pull/3709)
+
 
 * Update the TypeScript signature of `__wbindgen_thread_destroy` to indicate that it's parameters are optional.
   [#3703](https://github.com/rustwasm/wasm-bindgen/pull/3703)
--- a/src/convert/traits.rs
+++ b/src/convert/traits.rs
@@ -2,6 +2,7 @@ use core::borrow::Borrow;
 use core::ops::{Deref, DerefMut};
 
 use crate::describe::*;
+use crate::JsValue;
 
 /// A trait for anything that can be converted into a type that can cross the
 /// wasm ABI directly, eg `u32` or `f64`.
@@ -248,4 +249,12 @@ impl<T: WasmAbi> WasmRet<T> {
     pub fn join(self) -> T {
         T::join(self.prim1, self.prim2, self.prim3, self.prim4)
     }
+}
+
+pub trait TryFromJsValue: Sized {
+    /// The type returned in the event of a conversion error.
+    type Error;
+
+    /// Performs the conversion.
+    fn try_from_js_value(value: JsValue) -> Result<Self, Self::Error>;
 }
--- a/src/convert/traits.rs
+++ b/src/convert/traits.rs
@@ -250,11 +250,3 @@ impl<T: WasmAbi> WasmRet<T> {
         T::join(self.prim1, self.prim2, self.prim3, self.prim4)
     }
 }
-
-pub trait TryFromJsValue: Sized {
-    /// The type returned in the event of a conversion error.
-    type Error;
-
-    /// Performs the conversion.
-    fn try_from_js_value(value: JsValue) -> Result<Self, Self::Error>;
-}
--- a/src/convert/traits.rs
+++ b/src/convert/traits.rs
@@ -2,7 +2,6 @@ use core::borrow::Borrow;
 use core::ops::{Deref, DerefMut};
 
 use crate::describe::*;
-use crate::JsValue;
 
 /// A trait for anything that can be converted into a type that can cross the
 /// wasm ABI directly, eg `u32` or `f64`.
--- a/src/convert/traits.rs
+++ b/src/convert/traits.rs
@@ -251,6 +251,19 @@ impl<T: WasmAbi> WasmRet<T> {
     }
 }
 
+/// `TryFromJsValue` is a trait for converting a JavaScript value (`JsValue`)
+/// into a Rust type. It is particularly useful when dealing with JavaScript
+/// objects in WebAssembly, allowing for custom conversion logic that might
+/// be needed when interacting with JavaScript APIs or passing data between
+/// Rust and JavaScript.
+///
+/// This trait is intended as a solution to the limitations imposed by the
+/// automatic derivation of `TryFrom<JsValue>` in `wasm-bindgen`, which
+/// restricts the ability to implement custom conversions for ABI-bound structs.
+///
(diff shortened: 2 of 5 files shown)


daxpedda (changes requested):
Could you add an entry to the changelog as well?
LGTM otherwise.

daxpedda:
@hamza1311 and @Liamolucko, the idea here is to yank the last version and a new one with this change.
Myself I'm not entirely fond of the whole idea, but I believe if somebody is affected and willing to provide a fix as well, we should take that as something that needs to be addressed. So I would vote in favor.

Let me know what you think.

Liamolucko:
> Myself I'm not entirely fond of the whole idea, but I believe if somebody is affected and willing to provide a fix as well, we should take that as something that needs to be addressed. So I would vote in favor.

I feel the same way.

This does un-fix #2231 though, since `wasm_bindgen::convert` is an internal module and so `TryFromJsValue` is effectively private.

The easy solution is just to move it to `wasm_bindgen::TryFromJsValue`, but it's a bit weird to make people use a custom trait rather than the standard `TryFrom<JsValue>`. Alternatively we could make a `TryFromJsValue` derive macro, but that's a lot more work.

biryukovmaxim:
> This does un-fix #2231 though, since `wasm_bindgen::convert` is an internal module and so `TryFromJsValue` is effectively private.

convert module is public, it already exports From/IntoWasmAbi, and others. But these two looks like custom `From` and `Into` that include additional constraints.

Liamolucko:
> convert module is public, it already exports From/IntoWasmAbi, and others. But these two looks like custom `From` and `Into` that include additional constraints.

Yes, it's _technically_ public, but see its documentation:

> This is mostly an internal module, no stability guarantees are provided. Use at your own risk.

It's only really public so that macros can access it, nobody be using anything from in there directly.

biryukovmaxim:
> > This is mostly an internal module, no stability guarantees are provided. Use at your own risk.
> 
> It's only really public so that macros can access it, nobody should be using anything from in there directly.

Oh. I see. Your suggestion is to move the trait to wasm_bindgen - root module?

biryukovmaxim:
> Could you add an entry to the changelog as well? LGTM otherwise.

done

daxpedda:
> Alternatively we could make a `TryFromJsValue` derive macro, but that's a lot more work.

Even if we want to do that one day, it would still require `TryFromJsValue` trait to be public.
I agree though this is weird, but I can't come up with a better solution.

src/lib.rs
```diff
@@ -1901,3 +1912,11 @@ impl From<JsError> for JsValue {
         error.value
     }
 }
+
+pub trait TryFromJsValue: Sized {
```
daxpedda, line 1916: This will require some documentation now.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 * The DWARF section is now correctly modified instead of leaving it in a broken state.
   [#3483](https://github.com/rustwasm/wasm-bindgen/pull/3483)
+* Custom TryFromJsValue Trait: Fixed an issue where #[wasm_bindgen] automatically derived the TryFrom trait for any struct, preventing custom TryFrom<JsValue> implementations. The wasm-bindgen derive macro has been updated to utilize the new TryFromJsValue trait.    [#3709](https://github.com/rustwasm/wasm-bindgen/pull/3709)
+
```
daxpedda, line 16:
```suggestion

* Fixed an issue where `#[wasm_bindgen]` automatically derived the `TryFrom` trait for any struct, preventing custom `TryFrom<JsValue>` implementations. It has been updated to utilize a new `TryFromJsValue` trait instead.
  [#3709](https://github.com/rustwasm/wasm-bindgen/pull/3709)
```

daxpedda (changes requested):

daxpedda:
I'm on the fence about moving `TryFromJsValue` just back to `convert` and just figure out what to do here another time.
Any input would be appreciated.

src/convert/traits.rs
```diff
@@ -249,3 +250,24 @@ impl<T: WasmAbi> WasmRet<T> {
...
+/// be needed when interacting with JavaScript APIs or passing data between
+/// Rust and JavaScript.
+///
+/// This trait is intended as a solution to the limitations imposed by the
+/// automatic derivation of `TryFrom<JsValue>` in `wasm-bindgen`, which
+/// restricts the ability to implement custom conversions for ABI-bound structs.
+///
+/// Types implementing this trait must specify their conversion logic from
+/// `JsValue` to the Rust type, handling any potential errors that may occur
+/// during the conversion process.
```
daxpedda, line 266:
```suggestion
/// [`TryFromJsValue`] is a trait for converting a JavaScript value ([`JsValue`])
/// into a Rust type. It is used by the [`wasm_bindgen`](wasm_bindgen_macro::wasm_bindgen)
/// proc-macro to allow conversion to user types.
///
/// Types implementing this trait must specify their conversion logic from
/// [`JsValue`] to the Rust type, handling any potential errors that may occur
/// during the conversion process.
```

I removed the historical part, which I would like to pretend just never happened instead of having to remember to remove it at some later point when it becomes only confusing instead of useful for new users.

Also adjusted some of the wording implying that this is anything else then internal stuff for now.

daxpedda (changes requested):
LGTM, just some nits about docs.

As mentioned before, this unfixes #2231, but we can figure out exactly what to do here at some later point.

@biryukovmaxim you should probably hold off on any major changes for now, we are still discussing how exactly to proceed, there's no need to move `TryFromJsValue` back and forth upon every whimsical comment I make! (don't move it back now :smile:)

dani-garcia:
I've hit a problem with the TryFrom changes that is unrelated to #3685, but that would probably also occur with this PR, related to the ambiguity of using `Self::Error` when `#[wasm_bindgen]` is applied to an enum containing an `Error` variant.

```
#[wasm_bindgen]
pub enum LogLevel {
    Trace,
    Debug,
    Info,
    Warn,
    Error,
}
```

The way to solve that would be to use the fully qualified syntax for `Self`:
```diff
fn try_from_js_value(value: #wasm_bindgen::JsValue)
-                   -> #wasm_bindgen::__rt::std::result::Result<Self, Self::Error> {
+                   -> #wasm_bindgen::__rt::std::result::Result<Self, <Self as TryFromJsValue>::Error> {
```

I can open a PR myself after this one is merged if you prefer to merge this one as-is though.

daxpedda:
Could you open a PR based targeting `main`? We can merge that right now and let this PR rebase.
Thanks for bring this up!

Liamolucko:
> I've hit a problem with the TryFrom changes that is unrelated to #3685, but that would probably also occur with this PR, related to the ambiguity of using `Self::Error` when `#[wasm_bindgen]` is applied to an enum containing an `Error` variant.
> 
> ```
> #[wasm_bindgen]
> pub enum LogLevel {
>     Trace,
>     Debug,
>     Info,
>     Warn,
>     Error,
> }
> ```
> 
> The way to solve that would be to use the fully qualified syntax for `Self`:
> 
> ```diff
> fn try_from_js_value(value: #wasm_bindgen::JsValue)
> -                   -> #wasm_bindgen::__rt::std::result::Result<Self, Self::Error> {
> +                   -> #wasm_bindgen::__rt::std::result::Result<Self, <Self as TryFromJsValue>::Error> {
> ```
> 
> I can open a PR myself after this one is merged if you prefer to merge this one as-is though.

I think this was already fixed in #3678, can you check if it still happens using a git version of the `wasm-bindgen` macro?

dani-garcia:
My bad, seems this was fixed on master a bit ago, and I was looking at the struct changes of this PR instead of the enum changes. I just tested both master and this PR and they both work as expected, sorry for the confusion!

Liamolucko (approved):
I don't like regressing on #2231, but I guess the breaking changes are a higher priority.

Liamolucko:
> > Alternatively we could make a `TryFromJsValue` derive macro, but that's a lot more work.
> 
> Even if we want to do that one day, it would still require `TryFromJsValue` trait to be public. I agree though this is weird, but I can't come up with a better solution.

By the way, what I meant by a `TryFromJsValue` derive macro was a derive macro named `TryFromJsValue` that implements `TryFrom<JsValue>`. I guess that is a bit confusing when there's also an internal trait with the same name, but the internal one could be renamed to `TryFromJsValueInternal` or something to avoid that (not that that needs to happen in this PR).

daxpedda:
I'm gonna go ahead and merge this now and get the new release going.

PR #3723, opened 2023-11-26
Remove more Gecko-internal enums/dicts from WebIDL

evilpie:


daxpedda (changes requested):
Thank you for doing this!
Could you add a note in the changelog as well?

daxpedda (approved):
Thanks!

PR #3725, opened 2023-11-28
Remove span in generated field getter

### LINKED ISSUE #3707: #[wasm_bindgen] attribute not work in macro_rules
### Describe the Bug

When putting struct attributed `#[wasm_bindgen]` in a custom macro_rules, it throws error `cannot find value 'js' in this scope`.

### Steps to Reproduce

code snippet:

```rust
// a macro that output exactly the same input struct
macro_rules! make_struct {
    {
        $( #[$attr:meta] )*
        $svis:vis struct $record_name:ident {
            $(
                $fvis:vis $fname:ident : $ftype:ty
            ),* $(,)?
        }
    } => {
        $( #[$attr] )*
        $svis struct $record_name {
            $(
                $fvis $fname : $ftype
            ),*
        }
    };
}

// testing
make_struct!{
    #[wasm_bindgen]
    #[derive(Debug, Clone, Copy)]
    pub struct Test {
        pub a: i32 // error: cannot find value `js` in this scope
    }
}
```

### Expected Behavior

In my imagination, macro rules should be expanded before the proc macros, thus the above code should work as normal.

Correct me if I'm wrong.

Liamolucko:
Fixes #3707

Previously, `(*js).borrow().#rust_name` was being given the span of `rust_name`, which seems like a fairly harmless thing to do at first glance. However, it turns out that the span of a token also affects its hygiene - turns out proc macros have hygiene too, not just declarative macros!

This caused a problem because the declaration of `js` had a span of `Span::call_site()`, but it was being accessed with `rust_name`'s span, which might be a different context where `js` doesn't exist.

Usually this is fine because `Span::call_site()` is in the same context as the struct definition: normal code. However, when you put `#[wasm_bindgen]` inside a `macro_rules!`, `Span::call_site()` is now in the context of that `macro_rules!`, while `rust_name`'s span is still in normal code which can't access variables from inside `macro_rules!`.

~~To fix that I've just removed the span from `(*js).borrow().#rust_name`, making it `Span::call_site()`. I don't think this should have any effect on diagnostics anyway, since I don't see how that code could ever cause a compile error.~~ Turns out it was used, so I just got rid of the span on `js` instead of the whole thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 * Add bindings for `UserActivation`.
   [#3719](https://github.com/rustwasm/wasm-bindgen/pull/3719)
 
+### Fixed
+
+* Fixed a compiler error when using `#[wasm_bindgen]` inside `macro_rules!`.
+  [#3725](https://github.com/rustwasm/wasm-bindgen/pull/3725)
+
 ### Removed
 
 * Removed Gecko-only `InstallTriggerData` and Gecko-internal `FlexLineGrowthState`, `GridDeclaration`, `GridTrackState`,
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -389,7 +389,14 @@ impl ToTokens for ast::StructField {
         };
         let maybe_assert_copy = respan(maybe_assert_copy, ty);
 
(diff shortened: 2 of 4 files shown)


daxpedda (approved):
Very funny business.

daxpedda (approved):

PR #3726, opened 2023-11-28
Remove (Gecko) WebIDL files that don't generate any bindings

evilpie:
I don't think this needs a changelog entry, because it's not observable outside the crate.


daxpedda (approved):
Indeed, thank you!

PR #3734, opened 2023-12-04
Implement User Timing Level 3

PaulWagener:


daxpedda:
Unfortunately this can't be merged until the next breaking change.

PaulWagener:
That's unfortunate, but understandable.
Hopefully a breaking change release will be added to the roadmap someday soon...

guybedford:
I've pushed a change here that guards the signature changes behind the `web_sys_unstable_apis` guard, while making the non-breaking changes without that guard. Required some generator rejigging based on the same approach for attributes.

This is therefore no longer a breaking change, and may land.

PR #3735, opened 2023-12-06
(Tiny PR) Show `JsValue` is not Send/Sync explicitly

fzyzcjy:
Hi thanks for the library! I was confused why JsValue is not Send/Sync, and after searching I see https://github.com/rustwasm/wasm-bindgen/pull/955. Therefore, I make this tiny PR to add doc telling users why this is the case.


fzyzcjy:
Hmm OK, Rust is not stablized yet for that grammar...

daxpedda:
Just for reference: this should already be visible, even if not that notable.
image

fzyzcjy:
Thank you!

PR #3740, opened 2023-12-09
 Add a typescript test (not just a compile test) for enums

thomasetter:
Also changed from opt-in to opt-out for running TypeScript tests.

This will make sure that new tests are run by default.

### DIFF
--- a/crates/typescript-tests/src/enums.ts
+++ b/crates/typescript-tests/src/enums.ts
@@ -1,13 +1,28 @@
-import * as wbg from '../pkg/typescript_tests';
+import * as wbg from "../pkg/typescript_tests";
+import { expect, jest, test } from "@jest/globals";
 
-const a1: wbg.Foo = wbg.Foo.A;
-const a2: wbg.Foo.A = wbg.Foo.A;
-const a3: wbg.Foo.A = 1;
-const b1: wbg.Foo = wbg.Foo.B;
-const b2: wbg.Foo.B = wbg.Foo.B;
-const b3: wbg.Foo.B = 3;
+test("construction", () => {
+  const a1: wbg.Foo = wbg.Foo.A;
+  const a2: wbg.Foo.A = wbg.Foo.A;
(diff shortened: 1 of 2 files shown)


thomasetter:
Friendly ping :)

daxpedda (commented):
LGTM!

@Liamolucko would you mind taking a look at this? Unfortunately I'm completely unfamiliar with TS, so I don't feel comfortable approving something like this.

crates/typescript-tests/src/enums.ts
```diff
@@ -1,13 +1,28 @@
...
+  expect(a1).toStrictEqual(a2);
+  const a3: wbg.Foo.A = 1;
+  expect(a1).toStrictEqual(a3);
+
+  const b1: wbg.Foo = wbg.Foo.B;
+  const b2: wbg.Foo.B = wbg.Foo.B;
+  expect(b1).toStrictEqual(b2);
+  const b3: wbg.Foo.B = 3;
+  expect(b1).toStrictEqual(b3);
+  expect(() => { return a1 as wbg.Foo != b1 }).toBeTruthy();
```
Liamolucko, line 16: This asserts that the function itself is truthy, not that `a1 as wbg.Foo != b1` is truthy.

```suggestion
  expect(a1).not.toStrictEqual(b1);
```

Liamolucko (commented):
Personally I don't think this is super necessary, I think a compile test is enough: making sure that `wasm-bindgen` works properly at runtime is already covered by the regular tests in `tests/wasm`.

Having extra tests doesn't hurt though, so I'm fine to merge this regardless.

thomasetter:
Thanks, just for the motivation why I am doing this:
The problem with compile-only tests is that they just assert properties of the binding (`.d.ts`), however the binding and the underlying JS can (and I believe in the past have, but do not remember the location/commits) drift apart, which will then only be discovered by users after a release.

Liamolucko (approved):
Thanks!

PR #3742, opened 2023-12-11
Update `RTCRtpCodec`

logist322:
This PR updates `RTCRtpCodec` accordingly to [spec](https://w3c.github.io/webrtc-pc/#dom-rtcrtpcodec).
Also added unstable fields for `RTCRtpEncodingParameters`.


crates/web-sys/webidls/unstable/RTCRtpEncodingParameters.webidl
daxpedda, line 1: We should split this into two different files.

crates/web-sys/webidls/enabled/RTCRtpSender.webidl
```diff
@@ -51,28 +51,26 @@ dictionary RTCRtcpParameters {
...
-  DOMString      sdpFmtpLine;
+dictionary RTCRtpCodec {
+  required DOMString     mimeType;
+  required unsigned long clockRate;
+  unsigned short         channels;
+  DOMString              sdpFmtpLine;
 };
 
+dictionary RTCRtpCodecParameters : RTCRtpCodec {
+  required octet payloadType;
```
daxpedda, line 62:
```suggestion
  required unsigned short payloadType;
```
This would be a breaking change otherwise.

daxpedda: I think `required` also has to be removed, because the `new()` function has changed now.

crates/web-sys/webidls/enabled/RTCRtpSender.webidl
```diff
@@ -51,28 +51,26 @@ dictionary RTCRtcpParameters {
...
 
-dictionary RTCRtpCodecParameters {
-  unsigned short payloadType;
-  DOMString      mimeType;
-  unsigned long  clockRate;
-  unsigned short channels = 1;
-  DOMString      sdpFmtpLine;
+dictionary RTCRtpCodec {
+  required DOMString     mimeType;
+  required unsigned long clockRate;
```
daxpedda, line 56:
```suggestion
  DOMString     mimeType;
  unsigned long clockRate;
```
I believe both of these can't be required, because it changes the `new()` function of `RTCRtpCodecParameters`, which is a breaking change.

daxpedda (changes requested):
We have to be careful not to introduce breaking changes.
Otherwise this is fine.

crates/web-sys/webidls/enabled/RTCRtpSender.webidl
```diff
@@ -51,28 +51,26 @@ dictionary RTCRtcpParameters {
   boolean   reducedSize;
 };
 
-dictionary RTCRtpCodecParameters {
-  unsigned short payloadType;
+dictionary RTCRtpCodec {
   DOMString      mimeType;
   unsigned long  clockRate;
```
daxpedda, line 56: Apologies, I think these two have to be `required` after all.
You should take a look yourself and see what causes a breaking change in the API or not.

crates/web-sys/webidls/unstable/RTCRtpEncodingParametersCodec.webidl
daxpedda, line 1: Could you add a newline at the end of this file and the other as well?

daxpedda (changes requested):
Can you add a changelog entry or modify the last one you made?

PR #3746, opened 2023-12-16
Point to component model instead of host bindings (fixes #3143)

### LINKED ISSUE #3143: Link in documentation intro
I was just going through the [documentation](https://rustwasm.github.io/docs/wasm-bindgen/):

> The wasm-bindgen tool is sort of half polyfill for features like the [host bindings proposal](https://github.com/WebAssembly/host-bindings) and half features for empowering high-level interactions between JS and wasm-compiled code (currently mostly from Rust).

The host-bindings proposal seems to be [inactive](https://github.com/WebAssembly/interface-types), and it links to the [component model repo](https://github.com/WebAssembly/component-model). 

Should the documentation be updated (at least to explain why the proposal is inactive)?

ydirson:
This is a first pass based on textual occurrences.  Please someone with more context double-check, some occurrences that
did apply to host-bindings could possibly not apply to component-model, especially when mentioned as an example.

### DIFF
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -614,7 +614,7 @@ pub fn is_structural(
     item_attrs: Option<&ExtendedAttributeList>,
     container_attrs: Option<&ExtendedAttributeList>,
 ) -> bool {
-    // Note that once component model is implemented we'll want to switch this
+    // Note that once host bindings is implemented we'll want to switch this
     // from `true` to `false`, and then we'll want to largely read information
     // from the WebIDL about whether to use structural bindings or not.
     true || has_named_attribute(item_attrs, "Unforgeable")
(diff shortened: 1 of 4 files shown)


crates/webidl/src/util.rs
```diff
@@ -614,7 +614,7 @@ pub fn is_structural(
     item_attrs: Option<&ExtendedAttributeList>,
     container_attrs: Option<&ExtendedAttributeList>,
 ) -> bool {
-    // Note that once host bindings is implemented we'll want to switch this
+    // Note that once component model is implemented we'll want to switch this
```
daxpedda, line 617:
```suggestion
    // Note that once host bindings is implemented we'll want to switch this
```
I honestly don't know what the intentions here were, but I wouldn't update it just to make it easier to use `git blame` (until we not only figure out what was meant here, but actually also figure out if this is still correct or not).

daxpedda (approved):
Thanks!

PR #3748, opened 2023-12-19
fix missing typeof operator in __wbg_init() template in no-modules

bbradshaw:
I noticed a TypeError originating in the generated javascript wrapper when using the `--no-modules` flag (I was messing around with web workers which can result in `script_src` actually being undefined). 

It appears the check is for `undefined` is against the value of `string_src`, rather than the type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


daxpedda (changes requested):
I think we have messed this up before, not sure what happened here.
Thank you!

bbradshaw:
Cheers 🍻

PR #3761, opened 2023-12-29
CI fix for Rust 1.75

thomasetter:
Fix some changed ordering in a wat output file.

Ran `cargo clippy --no-deps --all-features -p wasm-bindgen-macro-support --fix -- -D warnings`

Allow unused imports in web-sys, since I could not figure out how else this should be fixed.


.github/workflows/main.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     - run: cargo clippy --no-deps --all-features -p wasm-bindgen-wasm-interpreter -- -D warnings
     - run: cargo clippy --no-deps --all-features -p wasm-bindgen-webidl -- -D warnings
     - run: cargo clippy --no-deps --all-features -p webidl-tests -- -D warnings
-    - run: cargo clippy --no-deps --all-features --target wasm32-unknown-unknown -p web-sys -- -D warnings
+    - run: cargo clippy --no-deps --all-features --target wasm32-unknown-unknown -p web-sys -- -D warnings -A unused-imports
```
Liamolucko, line 68: Could you use an `#![allow(unused_imports)]` to do this inside the code instead? This fixes it in CI but it'll still give a warning when you run it locally.

Specifically, it should probably go at the top of `crates/web-sys/src/features/mod.rs`, which is generated here: https://github.com/rustwasm/wasm-bindgen/blob/c0ee195d7ebbb18b8517d115c4f3bfee4b797a3d/crates/webidl/src/lib.rs#L822-L830

That'll make all the `#![allow(unused_imports)]` attributes generated inside `crates/webidl/src/generator.rs` redundant though, so you might want to get rid of those too.

thomasetter: `#![allow(unused_imports)]` at the top of the file did not work when I tried it, maybe because the file defines different modules:

image

[GitHub action run for `Run wasm-bindgen-webidl crate tests`](https://github.com/rustwasm/wasm-bindgen/actions/runs/7366677081/job/20049157369)

Adding it to each use statement seems to work though.

Do you see a better solution?

Liamolucko: Weird. I think it's because this uses `include!` instead of declaring the module normally: https://github.com/rustwasm/wasm-bindgen/blob/c0ee195d7ebbb18b8517d115c4f3bfee4b797a3d/crates/webidl-tests/main.rs#L18-L20

I wouldn't have expected that to make a difference, but apparently it's intentional: https://github.com/rust-lang/rust/issues/94340. So I agree that just adding `#[allow(unused_imports)]` to each `use` is the best solution.

Liamolucko (approved):
Thanks!

Issue #1373, opened 2019-03-21
compile-time internal error on passing &Function to js


limira:
I encountered `internal error: entered unreachable code` with wasm-bindgen when I tried the git master version at merge of #1353 (and still there at the merge of #1372, not sure how far ago when the bug appear. But the bug is not in 0.2.39 release)

You can see the minimal code that produce the error [here](https://github.com/limira/wb-bug). In words, I have a library `wb-lib` that wraps around `Foo` in `foo.js`. `Foo` is a JS class that need a callback that the Rust code will pass a `&js_sys::Function` to it. An app `wb-app` depends on `wb-lib`, it initialize a `Foo` object.

Compile the provided repo (using a matched version of wasm-bindgen-cli) will cause the compile time error as shown in the `backtrace` file (also in repo). Note that I run `wasm-bindgen --web`.

No error if arguments not include a `js_sys::Function`. No error with v0.2.39.

alexcrichton:
Thanks for the report and for the reproduction! I think this should be "fixed" in https://github.com/rustwasm/wasm-bindgen/pull/1377 where "fixed" here means "produces a better error message". The real fix though for this is to add `[patch]` annotations for `web-sys` in addition to `wasm-bindgen` because the problem was that too many crates showed up in the crate graph, confusing the CLI!

Issue #549, opened 2018-07-25
Duplicate symbol linker errors when doing `cargo test -p js-sys --target wasm32-unknown-unknown`


fitzgen:
On latest rust nightly and master `wasm-bindgen`, if I try to run the new wasm test runner, I get this:

```
$ cargo test -p js-sys --target wasm32-unknown-unknown
   Compiling js-sys v0.2.14 (file:///home/fitzgen/wasm-bindgen/crates/js-sys)
   Compiling wasm-bindgen-test v0.2.14 (file:///home/fitzgen/wasm-bindgen/crates/test)
error: linking with `rust-lld` failed: exit code: 1
  |
[...]

PR #3767, opened 2024-01-08
Fix cargo wasm_bindgen = { path=../wasm-bindgen } local builds

brainstorm:
Fixes https://github.com/rustwasm/wasm-bindgen/discussions/3764#discussioncomment-8022392

The cargo `links` change was introduced 5 years ago in commit https://github.com/rustwasm/wasm-bindgen/commit/7b6731fd3b4f766b10fa5126d1ca87e96bb3aa9e by @alexcrichton, so most probably there was (still is?) a very good justification about the error messages. OTOH, in my (comparatively very limited) experience with `wasm-bindgen`, this change is the difference between **`getrandom` errors** and **compiling successfully** in a local setting.

/cc @mmalenic

### DIFF
--- a/crates/shared/Cargo.toml
+++ b/crates/shared/Cargo.toml
@@ -12,8 +12,3 @@ dependency.
 """
 edition = "2018"
 rust-version = "1.57"
-
-# Because only a single `wasm_bindgen` version can be used in a dependency
-# graph, pretend we link a native library so that `cargo` will provide better
-# error messages than the esoteric linker errors we would otherwise trigger.
-links = "wasm_bindgen"


daxpedda:
I don't fully understand why this was needed in the first place.

But before I start digging into this why don't we see how to solve your issue?
I use local `wasm-bindgen` patches all the time and don't seem to have any problems so far.

Could you set up a repo with a minimal example that shows your issue?
Or if it's small enough feel free to post it here.

brainstorm:
Fair points and easy to reproduce, create a crate with the following `Cargo.toml`:

```toml
[package]
name = "wasm-bindgen-cargo-links-reproducer"
version = "0.1.0"
edition = "2021"

[dependencies]
#getrandom = { version = "0.2.11" } 			# This works fine
getrandom = { version = "0.2.11", features = ["js"] }	# This goes boom if `links` is not removed from crates/shared/Cargo.toml
wasm-bindgen = { path = "../wasm-bindgen" }
```

daxpedda:
This should fix the issue:
```toml
[dependencies]
getrandom = { version = "0.2", features = ["js"] }
wasm-bindgen = "0.2"

[patch.crates-io]
wasm-bindgen = { path = "../wasm-bindgen" }
```

brainstorm:
I see. Is the links keyword still needed though? Do those "esoteric error messages" still persist nowadays?

I see value in not having to patch Cargo.toml as a much better developer experience, makes sense?

daxpedda:
> I see. Is the links keyword still needed though? Do those "esoteric error messages" still persist nowadays?

I don't know, I think it would be good if we could remove that if it's not needed.
If you want to dig into it please feel free to.

> I see value in not having to patch Cargo.toml as a much better developer experience, makes sense?

AFAIK using `patch` is the recommended way to do this in Caro.
So I consider both these issues unrelated at this point.

brainstorm:
>  I think it would be good if we could remove that if it's not needed. If you want to dig into it please feel free to.

What would be the criteria to get this PR accepted then? 

I don't see any "esoteric error messages" anywhere and there's no details on the situations that triggered them on the commit message or elsewhere so I would suggest to remove that `links` line as this PR intends?

daxpedda:
> > I think it would be good if we could remove that if it's not needed. If you want to dig into it please feel free to.
> 
> What would be the criteria to get this PR accepted then?

Probably understanding/reproducing what the original issue was and making sure it's not there anymore.
E.g. following the breadcrumbs: https://github.com/rustwasm/wasm-bindgen/issues/549, https://github.com/rustwasm/wasm-bindgen/pull/772, https://github.com/rustwasm/wasm-bindgen/issues/1373.

> I don't see any "esoteric error messages" anywhere and there's no details on the situations that triggered them on the commit message or elsewhere so I would suggest to remove that `links` line as this PR intends?

The only reason why I would be in favor of removing it is because it's unnecessary at this point, but for that we would have make sure that it actually is.
Considering that it doesn't actually cause any problems AFAICT, e.g. if you use Cargo as recommended and use a `patch` entry, I wouldn't want to remove something that is potentially still acting as a fix for something we don't understand.

daxpedda:
Closing due to inactivity.

PR #3773, opened 2024-01-10
Remove outdated js snippet caveat on imports

crajcan:
I'm not sure if `import` statements are universally implemented now but myself and others seem to be having success using them in snippets.

### DIFF
--- a/guide/src/reference/js-snippets.md
+++ b/guide/src/reference/js-snippets.md
@@ -59,11 +59,6 @@ make use of `inline_js` but instead to leverage `module` where possible.
 While quite useful local JS snippets currently suffer from a few caveats which
 are important to be aware of. Many of these are temporary though!
 
-* Currently `import` statements are not supported in the JS file. This is a
-  restriction we may lift in the future once we settle on a good way to support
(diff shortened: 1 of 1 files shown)


daxpedda:
I honestly don't know anything about this limitation.
Going through git blame didn't help either.

@Liamolucko do you know something about the historical context here? Testing it out locally seems to work fine for me.

Liamolucko:
> @Liamolucko do you know something about the historical context here? Testing it out locally seems to work fine for me.

No, but I found the original RFC for snippets and [there's a note about it there](https://github.com/alexcrichton/rfcs-1/blob/local-js-dependencies/text/006-local-js-dependencies.md#js-files-depending-on-other-js-files). It looks like what the caveat's supposed to mean is that imports don't work relative to the original path of the snippet like you'd expect, but there's nothing stopping you from writing them anyway and having them be resolved relative to the final path of the snippet. The fact that that works seems to be an unintended side effect of not processing the JS at all.

daxpedda:
Thank you!
@crajcan at best I would update the documentation then instead of removing it.

Issue #3774, opened 2024-01-11
[Proposal] Capture coverage information in wasm-pack test


aDogCalledSpot:
This follows up on my work in: https://github.com/rustwasm/wasm-bindgen/issues/2276

I have a very rough proof of concept and would like to discuss further implementation details before moving towards a PR.

## The rough idea

- The usual `profiler_builtins` aren't implemented for wasm, so we need to use [`minicov`](https://github.com/Amanieu/minicov/)
- We want to issue a call to `minicov::capture_coverage` after the run of every test
    - The data can be dumped across as many files as we like, so we can either dump into a file after every single test, or dump it all at after all tests have run - which would offer more performance
- Afterwards the generated .profraw files can be merged into a .profdata
- A .profdata can be used to generate an HTML coverage report

### Building the code

If `minicov` is linked, then `RUSTFLAGS="-Cinstrument-coverage -Zno-profiler-runtime --emit=llvm-ir" wasm-pack test` is sufficient for getting the standard compiled wasm from the called `cargo build`.

However, `wasm-interpreter` then runs over the file and tries to fetch all the `wbgt_describe`s. Profiling information is already included in these. I made some attempts at using `#[coverage(off)]` to avoid the generation of calls to the profiler in these functions but was unsuccessful. We can retry this approach if we have the rest working until then I propose the following workaround:

#### The workaround

All fields related to profiling use i64. Since the interpreter disallows anything other than i32, we can ignore the i64s if we want to capture coverage (add some flag). Otherwise panic as usual.

This means we add the workaround for the instructions i64 const, i64 store, i64 load and i64 add.

### Generating the profraw

Using `minicov::capture_coverage` works fine for getting the data for a single test. Printing the data to a file however doesn't work from a browser. Since the browser already connects to a small rouille server, I suggest creating two extra endpoints in that server: 

- ~~`/__coverage/append`~~
  - ~~expects the `Vec<u8>` generated by minicov in the request body~~
  - ~~appends it to a vec `coverage`~~
  - ~~should be called after every test~~
  - This just moves around an unecessarily large amount of data. We can simply dump once after all tests have run
- `/__coverage/dump`
  - expects the `Vec<u8>` generated by minicov in the request body
  - simply dumps the coverage to a file
  - should be called after all tests have run so we only generate a single profraw

The code is pretty straightforward here.

### Merging the profraws and generating the HTML

[...]

Issue #2276, opened 2020-08-08
Code Coverage in wasm-bindgen-test?


anp:
### Motivation

I'd like to be able to use `-Zprofile` when building my wasm-bindgen tests with wasm-pack and have those tests produce coverage files. I currently run my tests in a headless browser with wasm-pack and an ideal solution would have a clear integration story there with a good out of the box experience.

### Proposed Solution

The instrumentation emitted by default writes coverage profiling data to disk in a location relative to the binary being run. If there's a nice way to override those callbacks for a single binary/target then wasm-bindgen could presumably collect the coverage buffers in its JS test harness and forward them to the CLI for persisting to disk.

### Alternatives

I'm still looking at LLVM's coverage docs and I'm not sure if it's straightforward to override the instrumentation collection buffers like that. If it's not then I'd be willing to compile in a larger (wasi?) runtime into my tests to provide the needed filesystem I/O. I think the LLVM instrumentation at least offers environment variables to control the paths where data is written.

### Additional Context

I looked for related issues and couldn't find any. I think that "test artifact extraction" from wasm tests might also be useful for [supporting snapshot testing](https://github.com/mitsuhiko/insta/issues/120) but I'm not familiar enough with relevant internals to say for sure.

alexcrichton:
To some degree WebAssembly instrumentation and coverage should be easier in theory than native binaries because it's easy to change/inject instructions into a binary (e.g. at the top/end of all blocks). That would require work investing into some sort of wasm transform, however, which would require a side table (like dwarf info) about what instruction corresponds to what. 

Otherwise LLVM's infrastructure may work with tweaks, but AFAIK no work has been done towards that. I would suspect you'd probably want to start out with wasm32-wasi since it'll be easier as many things already "just work", but if something works there seems reasonable to try to extend it to wasm-bindgen!

anp:
Makes sense and thanks for the advice! I am skimming https://motley-coder.com/2019/03/31/webassembly-coverage/ and it looks like compiler-rt support for the target is the main thing.

Amanieu:
You could try using [minicov](https://github.com/Amanieu/minicov/) which embeds a stripped down version of the LLVM profiling runtime. It has no dependencies on libc or any external code so it should compile on WASM. You'll need to write out the profraw file yourself though (see the README).

bbarwik:
[...]

PR #3782, opened 2024-01-14
Implement coverage output

aDogCalledSpot:
This follows up from https://github.com/rustwasm/wasm-bindgen/issues/3774.

Fixes https://github.com/rustwasm/wasm-bindgen/issues/2276.
Fixes https://github.com/rustwasm/wasm-bindgen/issues/3774.

# Overview

wasm-pack can set `WASM_BINDGEN_TEST_COVERAGE` to allow generating coverage data. `WASM_BINDGEN_TEST_PROFRAW_OUT` takes a path to specify where the generated .profraw file should be saved.

Test runners dump their coverage output after all tests have run. For browser tests, the data is sent to the server which can then dump the data into a profraw. For node/deno tests, the data is dumped directly to a file.

Merging the profraw files into a profdata and generating human-readable output (usually HTML) from the profdata is left to the user.

# Changes

## Interpreter changes

Generating profiling information causes i64s to be placed in the describe blocks that the interpreter parses. I tried adding `#[coverage(off)]` to these blocks to stop the profiling code from being generated but it didn't seem to be sufficient. I've kept comments about where I added these annotations in the diff here in case anyone has a suggestion of where else I could try adding them. All my attempts are in `codegen.rs`.

`#[coverage(off)]` isn't stabilized but I'm willing to open a stabilization PR if we are sure that it fixes our issues here. I'll be sure to document what I found more precisely in a separate issue to ensure the efforts aren't lost.

Until then, I added a coverage flag which allows skipping the instructions which are causing issues. If the flag isn't set, behavior is the same as before.

## Dumping the coverage

The coverage is dumped through the `__wbgtest_cov_dump()` command which requires enabling the `coverage` feature of `wasm-bindgen-test`.

### Browsers
The clients call `__wbgtest_cov_dump` and send the resulting data to the server. The server has an additional endpoint which writes coverage to a file.

### Node/Deno
The output is gathered and is dumped to a file directly.

## Usage

It's easiest to test if you also use [my fork of `wasm-pack`](https://github.com/aDogCalledSpot/wasm-pack/tree/coverage) (I'll follow-up with a PR there once we're happy with how everything works here).

Enable the coverage feature of the `wasm-bindgen-test` dependency.

```
# Build for use with minicov
RUSTFLAGS="-Cinstrument-coverage -Zno-profiler-runtime --emit=llvm-ir" wasm-pack test --chrome --headless --coverage --profraw-out wtest.profraw
# Generate a {file_name}.o which contains the mapping of wasm instructions
# to source code lines.
clang target/wasm32-unknown-unknown/debug/deps/{file_name}.ll -Wno-override-module -c -o {file_name}.o
# Merge all profraws into a profdata
llvm-profdata merge --sparse wtest.profraw -o coverage.profdata
# Build HTML output from the profdata
llvm-cov show --instr-profile=coverage.profdata {file_name}.o --format=html --output-dir=coverage/ --sources .
```

### Possible mistakes you can make:

So there are three things the user has to get right:

- `RUSTFLAGS` need to be set correctly when building in order to generate profiling information (profiling)
- `WASM_BINDGEN_TEST_COVERAGE` environment variable needs to be set (flag)
- `coverage` feature in `wasm_bindgen_test` needs to enabled (feature)

Here's the output of all the ways you can mix these up:

**No profiling, no feature, no flag**:
Fine, just as before

**No profiling, no feature, flag**:
Prints error saying that coverage was supposed to be dumped but feature disabled

**No profiling, feature, no flag**:
Fine, just as before. Note that you might have issues if you have a crate depending on `wasm-bindgen-test` that also has `minicov` as a dependency since you'll have multiple definitions of functions. This is why we have the feature.

**No profiling, feature, flag**:
Prints error saying that empty coverage data was received. Says what RUSTFLAGS to use (I wanted to have this as a warning but warnings don't seem to be forwarded to the command line in headless mode)

**Profiling, no feature, no flag**:
`wasm-bindgen-interpreter` will panic. Prints a hint. (*)

**Profiling, no feature, flag**:
`TypeError: The specifier "env" was a bare specifier`
(The JS interpreter can't find the calls to the profiling information)(**)

**Profiling, feature, no flag**:
`wasm-bindgen-interpreter` will panic (somewhere else for some reason). Prints a hint.(***)

**Profiling, feature, flag**:
You get coverage! 

(*) If we don't need the workaround in the interpreter, I'm pretty sure this would just take longer to compile but run fine

(**) A better message here would be great but I don't know how to detect this since the problem occurs at link time but it is only printed at runtime.

(**\*) If we don't need the workaround in the interpreter, this would probably also be fine.

# TODOs

- [x] Write an issue documenting the issues with `#[coverage(off)]` and link to it where appropriate
- [x] Write an entry to the book documenting this usage and the possible mistakes you can make and link to it where appropriate

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -256,11 +256,10 @@ fn coverage_args(tmpdir: &Path) -> Option<PathBuf> {
     }
 
     // Profraw path is ignored if coverage isn't enabled
-    env::var_os("WASM_BINDGEN_TEST_COVERAGE")?;
-    log::warn!("Coverage support is still considered highly experimental!");
+    env::var_os("WASM_BINDGEN_UNSTABLE_TEST_COVERAGE")?;
     // TODO coverage link to wasm-bindgen book documenting correct usage
 
-    match env::var_os("WASM_BINDGEN_TEST_PROFRAW_OUT") {
+    match env::var_os("WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_OUT") {
         Some(s) => {
             let mut buf = PathBuf::from(s);
             if buf.is_dir() {
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -15,6 +15,7 @@ use anyhow::{anyhow, bail, Context};
 use log::error;
 use std::env;
 use std::fs;
+use std::path::Path;
 use std::path::PathBuf;
 use std::thread;
 use wasm_bindgen_cli_support::Bindgen;
@@ -187,6 +188,8 @@ fn main() -> anyhow::Result<()> {
         b.split_linked_modules(true);
     }
 
+    let coverage = coverage_args(&tmpdir);
+
     b.debug(debug)
         .input_module(module, wasm)
         .keep_debug(false)
@@ -198,8 +201,8 @@ fn main() -> anyhow::Result<()> {
     let args: Vec<_> = args.collect();
 
     match test_mode {
-        TestMode::Node => node::execute(module, &tmpdir, &args, &tests)?,
-        TestMode::Deno => deno::execute(module, &tmpdir, &args, &tests)?,
+        TestMode::Node => node::execute(module, &tmpdir, &args, &tests, coverage)?,
+        TestMode::Deno => deno::execute(module, &tmpdir, &args, &tests, coverage)?,
         TestMode::Browser { no_modules } | TestMode::Worker { no_modules } => {
             let srv = server::spawn(
                 &if headless {
@@ -216,6 +219,7 @@ fn main() -> anyhow::Result<()> {
                 &tests,
                 no_modules,
                 matches!(test_mode, TestMode::Worker { no_modules: _ }),
+                coverage,
             )
             .context("failed to spawn server")?;
             let addr = srv.server_addr();
@@ -241,4 +245,30 @@ fn main() -> anyhow::Result<()> {
         }
     }
     Ok(())
+}
+
+fn coverage_args(tmpdir: &Path) -> Option<PathBuf> {
+    fn generated(tmpdir: &Path) -> String {
+        format!(
+            "{}.profraw",
+            tmpdir.file_name().and_then(|s| s.to_str()).unwrap()
+        )
+    }
+
+    // Profraw path is ignored if coverage isn't enabled
+    env::var_os("WASM_BINDGEN_TEST_COVERAGE")?;
+    log::warn!("Coverage support is still considered highly experimental!");
+    // TODO coverage link to wasm-bindgen book documenting correct usage
+
+    match env::var_os("WASM_BINDGEN_TEST_PROFRAW_OUT") {
+        Some(s) => {
+            let mut buf = PathBuf::from(s);
+            if buf.is_dir() {
+                buf.push(generated(tmpdir));
+            }
+            buf
+        }
+        None => PathBuf::from(generated(tmpdir)),
+    }
+    .into()
 }
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -264,7 +264,6 @@ fn coverage_args(tmpdir: &Path) -> Option<PathBuf> {
         .map(|s| s.to_str().unwrap().to_string())
         .unwrap_or_default();
 
-
     match env::var_os("WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_OUT") {
         Some(s) => {
             let mut buf = PathBuf::from(s);
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -248,26 +248,32 @@ fn main() -> anyhow::Result<()> {
 }
 
 fn coverage_args(tmpdir: &Path) -> Option<PathBuf> {
-    fn generated(tmpdir: &Path) -> String {
-        format!(
-            "{}.profraw",
+    fn generated(tmpdir: &Path, prefix: &str) -> String {
+        let res = format!(
+            "{prefix}{}.profraw",
             tmpdir.file_name().and_then(|s| s.to_str()).unwrap()
-        )
+        );
+        res
     }
 
     // Profraw path is ignored if coverage isn't enabled
     env::var_os("WASM_BINDGEN_UNSTABLE_TEST_COVERAGE")?;
     // TODO coverage link to wasm-bindgen book documenting correct usage
 
+    let prefix = env::var_os("WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_PREFIX")
+        .map(|s| s.to_str().unwrap().to_string())
+        .unwrap_or_default();
+
+
     match env::var_os("WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_OUT") {
         Some(s) => {
             let mut buf = PathBuf::from(s);
             if buf.is_dir() {
-                buf.push(generated(tmpdir));
+                buf.push(generated(tmpdir, &prefix));
             }
             buf
         }
-        None => PathBuf::from(generated(tmpdir)),
+        None => PathBuf::from(generated(tmpdir, &prefix)),
     }
     .into()
 }
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
@@ -1,8 +1,9 @@
 use std::borrow::Cow;
-use std::ffi::OsString;
 use std::fs;
+use std::io::{Read, Write};
 use std::net::SocketAddr;
 use std::path::Path;
+use std::{ffi::OsString, path::PathBuf};
 
 use anyhow::{anyhow, Context, Error};
 use rouille::{Request, Response, Server};
@@ -16,8 +17,29 @@ pub fn spawn(
     tests: &[String],
     no_module: bool,
     worker: bool,
+    coverage: Option<PathBuf>,
 ) -> Result<Server<impl Fn(&Request) -> Response + Send + Sync>, Error> {
     let mut js_to_execute = String::new();
+
+    let (cov_import, cov_dump) = if coverage.is_some() {
+        (
+            if no_module {
+                "let __wbgtest_cov_dump = wasm_bindgen.__wbgtest_cov_dump;"
+            } else {
+                "__wbgtest_cov_dump,"
+            },
+            r#"
+            // Dump the coverage data collected during the tests
+            const coverage = __wbgtest_cov_dump();
+            await fetch("/__coverage/dump", {
+                method: "POST",
+                body: coverage
+            });
+            "#,
+        )
+    } else {
+        ("", "")
+    };
 
     let wbg_import_script = if no_module {
         String::from(
@@ -28,6 +50,7 @@ pub fn spawn(
             let __wbgtest_console_info = wasm_bindgen.__wbgtest_console_info;
             let __wbgtest_console_warn = wasm_bindgen.__wbgtest_console_warn;
             let __wbgtest_console_error = wasm_bindgen.__wbgtest_console_error;
+            {cov_import}
             let init = wasm_bindgen;
             "#,
         )
@@ -41,6 +64,7 @@ pub fn spawn(
                 __wbgtest_console_info,
                 __wbgtest_console_warn,
                 __wbgtest_console_error,
+                {cov_import}
                 default as init,
             }} from './{}';
             "#,
@@ -95,6 +119,7 @@ pub fn spawn(
 
                 cx.args({1:?});
                 await cx.run(tests.map(s => wasm[s]));
+                {cov_dump}
             }}
 
             onmessage = function(e) {{
@@ -175,6 +200,7 @@ pub fn spawn(
                 cx.args({1:?});
 
                 await cx.run(test.map(s => wasm[s]));
+                {cov_dump}
             }}
 
             const tests = [];
@@ -218,6 +244,25 @@ pub fn spawn(
                 )
             };
             return set_isolate_origin_headers(Response::from_data("text/html", s));
+        } else if request.url() == "/__coverage/dump" {
+            let profraw_path = coverage.as_ref().expect(
+                "Received coverage dump request but server wasn't set up to accept coverage",
+            );
+            // This is run after all tests are done and dumps the data received in the request
+            // into a single profraw file
+            let mut profraw =
+                std::fs::File::create(profraw_path).expect("Couldn't create .profraw for coverage");
+            let mut data = Vec::new();
+            request
+                .data()
+                .expect("Expected coverage data in body")
+                .read_to_end(&mut data)
+                .expect("Failed to read message body");
+
+            profraw
+                .write_all(&data)
+                .expect("Couldn't dump coverage data to profraw");
+            return Response::text("Coverage dumped");
         }
 
         // Otherwise we need to find the asset here. It may either be in our
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
@@ -245,24 +245,21 @@ pub fn spawn(
             };
             return set_isolate_origin_headers(Response::from_data("text/html", s));
         } else if request.url() == "/__coverage/dump" {
-            let profraw_path = coverage.as_ref().expect(
-                "Received coverage dump request but server wasn't set up to accept coverage",
-            );
-            // This is run after all tests are done and dumps the data received in the request
-            // into a single profraw file
-            let mut profraw =
-                std::fs::File::create(profraw_path).expect("Couldn't create .profraw for coverage");
-            let mut data = Vec::new();
-            request
-                .data()
-                .expect("Expected coverage data in body")
-                .read_to_end(&mut data)
-                .expect("Failed to read message body");
-
-            profraw
-                .write_all(&data)
-                .expect("Couldn't dump coverage data to profraw");
-            return Response::text("Coverage dumped");
+            fn internal_err(s: &str) -> Response {
+                log::error!("{s}");
+                let mut ret = Response::text(s);
+                ret.status_code = 500;
+                ret
+            }
+            let Some(profraw_path) = &coverage else {
+                return internal_err(
+                    "Received coverage dump request but server wasn't set up to accept coverage",
+                );
+            };
+            return match handle_coverage_dump(profraw_path, request) {
+                Ok(()) => Response::empty_204(),
+                Err(e) => internal_err(&format!("Failed to dump coverage: {e}")),
+            };
         }
 
         // Otherwise we need to find the asset here. It may either be in our
@@ -309,6 +306,21 @@ pub fn spawn(
         }
         response
     }
+}
+
+fn handle_coverage_dump(profraw_path: &Path, request: &Request) -> anyhow::Result<()> {
+    // This is run after all tests are done and dumps the data received in the request
+    // into a single profraw file
+    let mut profraw = std::fs::File::create(profraw_path)?;
+    let mut data = Vec::new();
+    if let Some(mut r_data) = request.data() {
+        r_data.read_to_end(&mut data)?;
+    }
+    // Warnings about empty data should have already been handled by
+    // the client
+
+    profraw.write_all(&data)?;
+    Ok(())
 }
 
 /*
--- a/crates/test/Cargo.toml
+++ b/crates/test/Cargo.toml
@@ -20,6 +20,8 @@ wasm-bindgen = { path = '../..', version = '0.2.90' }
 wasm-bindgen-futures = { path = '../futures', version = '0.4.40' }
 wasm-bindgen-test-macro = { path = '../test-macro', version = '=0.3.40' }
 gg-alloc = { version = "1.0", optional = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
 minicov = { version = "0.3", optional = true }
 
 [lib]
--- a/crates/test/Cargo.toml
+++ b/crates/test/Cargo.toml
@@ -8,6 +8,12 @@ repository = "https://github.com/rustwasm/wasm-bindgen"
 edition = "2018"
 rust-version = "1.57"
 
+[features]
+default = []
+experimental = ["coverage"]
+
+coverage = ["minicov"]
+
 [dependencies]
 console_error_panic_hook = '0.1'
 js-sys = { path = '../js-sys', version = '0.3.67' }
@@ -16,6 +22,7 @@ wasm-bindgen = { path = '../..', version = '0.2.90' }
 wasm-bindgen-futures = { path = '../futures', version = '0.4.40' }
 wasm-bindgen-test-macro = { path = '../test-macro', version = '=0.3.40' }
 gg-alloc = { version = "1.0", optional = true }
+minicov = { version = "0.3", optional = true }
 
 [lib]
 test = false
--- a/crates/test/Cargo.toml
+++ b/crates/test/Cargo.toml
@@ -10,9 +10,7 @@ rust-version = "1.57"
 
 [features]
 default = []
-experimental = ["coverage"]
-
-coverage = ["minicov"]
+unstable-coverage = ["minicov"]
 
 [dependencies]
 console_error_panic_hook = '0.1'
--- a/crates/test/src/coverage.rs
+++ b/crates/test/src/coverage.rs
@@ -1,0 +1,29 @@
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[cfg(feature = "coverage")]
+#[wasm_bindgen]
+pub fn __wbgtest_cov_dump() -> Vec<u8> {
+    let mut coverage = Vec::new();
+    unsafe {
+        minicov::capture_coverage(&mut coverage).unwrap();
+    }
+    if coverage.is_empty() {
+        console_error!(
+            "Empty coverage data received. Make sure you compile the tests with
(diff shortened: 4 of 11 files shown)


crates/backend/src/codegen.rs
```diff
@@ -173,6 +173,7 @@ impl ToTokens for ast::Struct {
         (quote! {
             #[automatically_derived]
             impl #wasm_bindgen::describe::WasmDescribe for #name {
+                // #[coverage(off)]
```
daxpedda, line 176: I think we should leave a more descriptive comment instead.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -198,8 +201,8 @@ fn main() -> anyhow::Result<()> {
     let args: Vec<_> = args.collect();
 
     match test_mode {
-        TestMode::Node => node::execute(module, &tmpdir, &args, &tests)?,
-        TestMode::Deno => deno::execute(module, &tmpdir, &args, &tests)?,
+        TestMode::Node => node::execute(module, &tmpdir, &args, &tests, coverage)?,
+        TestMode::Deno => deno::execute(module, &tmpdir, &args, &tests, coverage)?,
```
daxpedda, line 205: I can't really review Node or Deno stuff, so I will be pulling in another maintainer when we get there.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -242,3 +246,29 @@ fn main() -> anyhow::Result<()> {
...
+    fn generated(tmpdir: &Path) -> String {
+        format!(
+            "{}.profraw",
+            tmpdir.file_name().and_then(|s| s.to_str()).unwrap()
+        )
+    }
+
+    // Profraw path is ignored if coverage isn't enabled
+    env::var_os("WASM_BINDGEN_TEST_COVERAGE")?;
+    log::warn!("Coverage support is still considered highly experimental!");
```
daxpedda, line 260:
```suggestion
    env::var_os("WASM_BINDGEN_UNSTABLE_TEST_COVERAGE")?;
```

crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
```diff
@@ -218,6 +244,25 @@ pub fn spawn(
...
+            request
+                .data()
+                .expect("Expected coverage data in body")
+                .read_to_end(&mut data)
+                .expect("Failed to read message body");
+
+            profraw
+                .write_all(&data)
+                .expect("Couldn't dump coverage data to profraw");
+            return Response::text("Coverage dumped");
```
daxpedda, line 265: We probably want to log an error instead of panicking here.
Might be simpler to move this block into it's own function returning a `Result` and handling logging at the call site.

crates/test/Cargo.toml
```diff
@@ -8,6 +8,12 @@ repository = "https://github.com/rustwasm/wasm-bindgen"
 edition = "2018"
 rust-version = "1.57"
 
+[features]
+default = []
+experimental = ["coverage"]
+
+coverage = ["minicov"]
```
daxpedda, line 15:
```suggestion
unstable-coverage = ["minicov"]
```

crates/test/src/lib.rs
```diff
@@ -21,6 +21,24 @@ macro_rules! console_log {
...
 }
 
+/// Helper macro which acts like `println!` only routes to `console.warn`
+/// instead.
+#[macro_export]
+macro_rules! console_warn {
+    ($($arg:tt)*) => (
+        $crate::__rt::log_warn(&format_args!($($arg)*))
+    )
+}
```
daxpedda, line 31: Let's stick only with what we need here.
AFAICT this is unused.

daxpedda (commented):
Looking forward to see your report on what the hell is going on with `#[coverage(off)]`.

crates/backend/src/codegen.rs
daxpedda: After looking through the rest of your code I guess this is just your TODO.

aDogCalledSpot: I left it in there for now in case you have any suggestions where else it might be worth checking if it could be added. I'm pretty sure that the issue is that I'm missing another `#[coverage(off)]` somewhere but I don't know the codegen code well enough.

I'll write up a separate issue that goes into a bit more depth on the topic and probably just remove these comments. No point having a commented line with the caption "I think  this solves issue X but it doesn't".

daxpedda:
On that not, are you sure t hat `llvm-cov` can't get the debug data from the Wasm file?
Right now the `wasm-bindgen-test-runner` actually throws away all the debug info, so did you try keeping it and see how `llvm-cov` reacts then?

aDogCalledSpot:
> On that not, are you sure t hat `llvm-cov` can't get the debug data from the Wasm file? Right now the `wasm-bindgen-test-runner` actually throws away all the debug info, so did you try keeping it and see how `llvm-cov` reacts then?
> 
> EDIT:
> 
> https://github.com/rustwasm/wasm-bindgen/blob/8797e9320238ae80b0406721cb3d718d3aae4af5/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs#L192

Yes. I played around with that quite a bit. I discarded all the changes regarding "keep_debug"  since it wasn't working.

daxpedda:
> (Tested this on my project and got dependency error around `console_error_panic_hook` which still seems to depend on the main `wasm-bindgen` repo, see [#3774 (comment)](https://github.com/rustwasm/wasm-bindgen/issues/3774#issuecomment-1891985718))

@egasimus I recommend using the [`patch` section](https://doc.rust-lang.org/1.75.0/cargo/reference/overriding-dependencies.html#the-patch-section) instead of overwriting dependencies directly.
E.g. I used the following:
```toml
[patch.crates-io]
wasm-bindgen-test = { git = "https://github.com/aDogCalledSpot/wasm-bindgen", branch = "coverage" }
wasm-bindgen = { git = "https://github.com/aDogCalledSpot/wasm-bindgen", branch = "coverage" }
js-sys = { git = "https://github.com/aDogCalledSpot/wasm-bindgen", branch = "coverage" }
web-sys = { git = "https://github.com/aDogCalledSpot/wasm-bindgen", branch = "coverage" }
wasm-bindgen-futures = { git = "https://github.com/aDogCalledSpot/wasm-bindgen", branch = "coverage" }
```

aDogCalledSpot:
I've typed up a report on the `#[coverage(off)]` stuff. I'll turn it into an issue once this is ready to merge otherwise: https://gist.github.com/aDogCalledSpot/509309bb2aa4b251ce4d497bca9baf40

njelich:
Really glad that the [wasmcov method](https://github.com/hknio/wasmcov) is spreading :) Would be nice if we got an attribution in contributions using the workaround.

aDogCalledSpot:
> Really glad that the [wasmcov method](https://github.com/hknio/wasmcov) is spreading :) Would be nice if we got an attribution in contributions using the workaround.

I honestly wasn't aware of `wasmcov` until now. Looking at it now I also don't feel as if it addresses the main contributions of this PR:

 - Skipping over the issues in the interpreter (as unelegant as it may be)
 - Sending the coverage data to the server and writing it to a profraw from there
 
Using the `.ll` file to extract the debug info was taken from https://github.com/hknio/code-coverage-for-webassembly which appears to predate wasmcov by a few weeks.

daxpedda:
> Using the `.ll` file to extract the debug info was taken from https://github.com/hknio/code-coverage-for-webassembly which appears to predate wasmcov by a few weeks.

I think adding some attribution to https://github.com/hknio/code-coverage-for-webassembly would be nice!

aDogCalledSpot:
Looking more closely it seems as if the auhor of the page I linked is affiliated with wasmcov (all working for the same company).

There isn't any code specific to the workaround in this PR since this would be a manual step done after running the tests. I'll add an attribution to the documentation in the book.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -242,3 +246,28 @@ fn main() -> anyhow::Result<()> {
...
+            "{}.profraw",
+            tmpdir.file_name().and_then(|s| s.to_str()).unwrap()
+        )
+    }
+
+    // Profraw path is ignored if coverage isn't enabled
+    env::var_os("WASM_BINDGEN_UNSTABLE_TEST_COVERAGE")?;
+    // TODO coverage link to wasm-bindgen book documenting correct usage
+
+    match env::var_os("WASM_BINDGEN_TEST_PROFRAW_OUT") {
```
daxpedda, line 262:
```suggestion
    match env::var_os("WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_OUT") {
```

crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
```diff
@@ -218,6 +244,22 @@ pub fn spawn(
...
+                ret.status_code = 500;
+                ret
+            }
+            let Some(profraw_path) = &coverage else {
+                return internal_err(
+                    "Received coverage dump request but server wasn't set up to accept coverage",
+                );
+            };
+            return match handle_coverage_dump(profraw_path, request) {
+                Ok(()) => Response::text("Coverage dumped"),
```
daxpedda, line 260: I think an empty response is just as good.

crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
```diff
@@ -218,6 +244,22 @@ pub fn spawn(
...
             };
             return set_isolate_origin_headers(Response::from_data("text/html", s));
+        } else if request.url() == "/__coverage/dump" {
+            fn internal_err(s: &str) -> Response {
+                log::error!("{s}");
+                let mut ret = Response::text(s);
+                ret.status_code = 500;
+                ret
+            }
+            let Some(profraw_path) = &coverage else {
```
daxpedda, line 254:
```suggestion
            let Some(profraw_path) = coverage else {
```

crates/test/src/rt/mod.rs
```diff
@@ -233,6 +233,14 @@ extern "C" {
...
     pub fn js_console_log(s: &str);
 
+    #[wasm_bindgen(js_namespace = console, js_name = error)]
+    #[doc(hidden)]
+    pub fn js_console_error(s: &str);
+
+    #[wasm_bindgen(js_namespace = console, js_name = warn)]
+    #[doc(hidden)]
+    pub fn js_console_warn(s: &str);
+
```
daxpedda, line 243: This should be dead code by now.

crates/wasm-interpreter/src/lib.rs
```diff
@@ -67,6 +74,7 @@ impl Interpreter {
     /// the `module` passed to `interpret` below.
     pub fn new(module: &Module) -> Result<Interpreter, anyhow::Error> {
         let mut ret = Interpreter::default();
+        ret.coverage = std::env::var_os("WASM_BINDGEN_TEST_COVERAGE").is_some();
```
daxpedda, line 77:
```suggestion
        ret.coverage = std::env::var_os("WASM_BINDGEN_UNSTABLE_TEST_COVERAGE").is_some();
```

crates/test/src/coverage.rs
```diff
@@ -0,0 +1,29 @@
...
+    if coverage.is_empty() {
+        console_error!(
+            "Empty coverage data received. Make sure you compile the tests with
+        RUSTFLAGS=\"-Cinstrument-coverage -Zno-profile-runtime --emit=llvm-ir\"",
+        );
+    }
+    coverage
+}
+
+/// Called when setting WASM_BINDGEN_TEST_COVERAGE but coverage feature is disabled.
```
daxpedda, line 19:
```suggestion
/// Called when setting WASM_BINDGEN_UNSTABLE_TEST_COVERAGE but coverage feature is disabled.
```

crates/wasm-interpreter/src/lib.rs
```diff
@@ -219,12 +227,13 @@ impl Interpreter {
         log::debug!("arguments {:?}", args);
         let local = match &func.kind {
             walrus::FunctionKind::Local(l) => l,
-            _ => panic!("can only call locally defined functions"),
+            _ => panic!("can only call locally defined functions. If you're compiling with profiling information you might want to set WASM_BINDGEN_TEST_COVERAGE (--coverage flag in wasm-pack)"),
```
daxpedda, line 230:
```suggestion
            _ => panic!("can only call locally defined functions. If you're compiling with profiling information you might want to set WASM_BINDGEN_UNSTABLE_TEST_COVERAGE (--coverage flag in wasm-pack)"),
```

crates/wasm-interpreter/src/lib.rs
```diff
@@ -255,15 +264,16 @@ struct Frame<'a> {
...
         use walrus::ir::*;
 
         let stack = &mut self.interp.scratch;
 
         match instr {
             Instr::Const(c) => match c.value {
                 Value::I32(n) => stack.push(n),
-                _ => panic!("non-i32 constant"),
+                Value::I64(_) if coverage => stack.push(0),
+                _ => panic!("non-i32 constant. If you're compiling with profiling information you might want to set WASM_BINDGEN_TEST_COVERAGE (--coverage flag in wasm-pack)"),
```
daxpedda, line 276:
```suggestion
                _ => panic!("non-i32 constant. If you're compiling with profiling information you might want to set WASM_BINDGEN_UNSTABLE_TEST_COVERAGE (--coverage flag in wasm-pack)"),
```

daxpedda (changes requested):
Apologies, most of my comments are just cleanup.

I think it would be good to also wire up the strategy I talked about in https://gist.github.com/aDogCalledSpot/509309bb2aa4b251ce4d497bca9baf40?permalink_comment_id=4833636#gistcomment-4833636, even if we can't solve it before a merge, presumably getting `#[coverage(off)]` on some functions is still a plus.

crates/cli/src/bin/wasm-bindgen-test-runner/server.rs
aDogCalledSpot:
```
error[E0507]: cannot move out of `coverage.0`, as `coverage` is a captured variable in an `Fn` closure
   --> crates/cli/src/bin/wasm-bindgen-test-runner/server.rs:254:38
    |
20  |     coverage: Option<PathBuf>,
    |     -------- captured outer variable
...
221 |     let srv = Server::new(addr, move |request| {
    |                                 -------------- captured by this `Fn` closure
...
254 |             let Some(profraw_path) = coverage else {
    |                      ------------    ^^^^^^^^
    |                      |
    |                      data moved here
    |                      move occurs because `profraw_path` has type `PathBuf`, which does not implement the `Copy` trait
    |
help: consider borrowing the pattern binding
    |
254 |             let Some(ref profraw_path) = coverage else {
```

aDogCalledSpot:
I got it working with the features. All ready from my end then.

njelich:
@aDogCalledSpot The wasmcov method was first published under that repo. As you can see both repos are in the Hacken organization. We are cleaning up and will set a redirect with docs soon.

njelich:
@aDogCalledSpot Btw, for a lot of the stuff that needs to be done (merging prof raw files, processing, etc - I am exposing those functions in wasmcov - its meant to be an easy to use library, so maybe you will find it useful for wasm-bindgen?)

aDogCalledSpot:
I'm afraid I'm a bit lost with the feature. Even in a minimal example I can't seem to make it do what I would want to:

Do I understand correctly that you would want this to compile:

```rust
#[allow_inter(coverage_attribute)]
#[coverage(off)]
pub fn foo() -> u32 {
    5
}
```

because this still complains about coverage being an experimental feature.

aDogCalledSpot:
> @aDogCalledSpot Btw, for a lot of the stuff that needs to be done (merging prof raw files, processing, etc - I am exposing those functions in wasmcov - its meant to be an easy to use library, so maybe you will find it useful for wasm-bindgen?)

I think it's out of scope for wasm-bindgen. I don't think there should be any more happening here in terms of test than native cargo test can do. I'm looking to implement more of the handling in cargo-llvm-cov instead.

daxpedda:
`allow_internal_unstable` is for proc macros only.

So the problem we are trying to solve here is how to make proc-macros that emit code using unstable features without requiring the user to use `#![feature(coverage_attribute)]` on their end.

This is important because any library that is using the `#[wasm_bindgen]` macro would have to do that, which isn't really feasible.

The solution is to use `allow_internal_unstable` to allow proc-macros to emit code using unstable features without requiring enabling that feature in the library the code is emitted into (as long as we are compiling with nightly ofc).
```rust
#[proc_macro_attribute]
#[cfg_attr(feature = "unstable-coverage", allow_internal_unstable(coverage_attribute))]
pub fn wasm_bindgen(attr: TokenStream, input: TokenStream) -> TokenStream {
```

Using `#[coverage(off)]` in a `wasm-bindgen` library directly, not in proc-macro output, we just need to use the usual `#![feature(coverage_attribute)]`, which has to be guarded behind a cargo feature as well.

crates/backend/src/codegen.rs
```diff
@@ -173,6 +173,7 @@ impl ToTokens for ast::Struct {
         (quote! {
             #[automatically_derived]
             impl #wasm_bindgen::describe::WasmDescribe for #name {
+                #[cfg_attr(feature = "unstable-coverage", allow_internal_unstable(coverage_attribute))]
```
daxpedda, line 176:
```suggestion
                #[coverage(off)]
```
Is fine here, but you have to insert it conditionally depending on `feature = "unstable-coverage"`, so through a variable, e.g. `#coverage`.

On the proc-macro itself you put `allow_internal_unstable`:
```rust
#[proc_macro_attribute]
#[cfg_attr(feature = "unstable-coverage", allow_internal_unstable(coverage_attribute))]
pub fn wasm_bindgen(attr: TokenStream, input: TokenStream) -> TokenStream {
```
Now any library using `wasm-bindgen` with `feature = "unstable-coverage"` can do that as long as they are on nightly without themselves having to use `#![feature(coverage_attribute)]`.

aDogCalledSpot:
Thanks for elaborating :heart:

It works when using the nightly compiler now but the CI doesn't which is causing it to fail on the clippy tests. Should we make clippy use the nightly compiler in the CI?

aDogCalledSpot:
For the book I'll cover it quite briefly and link refer to the wasm-pack docs where I'll cover it in more depth since that's what seems to be done for the other arguments too.

daxpedda:
> It works when using the nightly compiler now but the CI uses stable which is causing it to fail on the clippy tests. Should we make clippy use the nightly compiler in the CI?


Actually this is a bit problematic now that I think about it ... we might not want to do it with a crate feature after all, because `--all-features` worked before but now it wouldn't.

I think we will have to switch to using `cfg(web_sys_unstable_test_coverage)`, similarly to how `cfg(web_sys_unstable_apis)` is used now.
> For the book I'll cover it quite briefly and refer to the wasm-pack docs where I'll cover it in more depth since that's what seems to be done for the other arguments too.

Unfortunately I (as a maintainer) have no relation to `wasm-pack` so I don't want to rely on it that much (maybe other maintainers can chime in on this), but would prefer if we have a *complete* documentation in `wasm-bindgen`, it doesn't have to be in depth.

crates/backend/src/lib.rs
```diff
@@ -27,6 +27,11 @@
 #![cfg_attr(feature = "extra-traits", deny(missing_debug_implementations))]
 #![deny(missing_docs)]
 #![doc(html_root_url = "https://docs.rs/wasm-bindgen-backend/0.2")]
+#![cfg_attr(
+    feature = "unstable-coverage",
+    feature(allow_internal_unstable),
+    allow(internal_features)
+)]
```
daxpedda, line 34: I think this is a leftover.

crates/test/src/lib.rs
```diff
@@ -59,3 +68,9 @@ macro_rules! wasm_bindgen_test_configure {
 
 #[path = "rt/mod.rs"]
 pub mod __rt;
+
+// Make this only available to wasm32 so that we don't
+// import minicov on other archs.
+// That way you can use normal cargo test without minicov
```
daxpedda, line 74: I agree this is a good point, but then we should move the dependency to `[target.'cfg(target_arch = "wasm32")'.dependencies]` as well.

aDogCalledSpot:
> I think we will have to switch to using `cfg(web_sys_unstable_test_coverage)`, similarly to how `cfg(web_sys_unstable_apis)` is used now.

I feel like this adds a really large hurdle for the very small benefit of not generating the coverage data for these functions which also doesn't solve the issues it's supposed to. By using the feature flag you can already opt into getting the profraw even if the macros generate coverage output.

I would prefer to drop this commit from the PR and instead add a reference to the commit in my gist/follow-up issue. That way anyone working on this issue can check it out and compile with nightly and then easily continue where I left off.

daxpedda:
> I would prefer to drop this commit from the PR and instead add a reference to the commit in my gist/follow-up issue.

Keep in mind that when we solve this we will still have to switch away from the crate feature breaking everybodies workflow.

Considering this is an unstable feature I'm fine with that anyway.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
aDogCalledSpot: I think this is the only thing missing.

guide/src/wasm-bindgen-test/coverage.md
daxpedda, line 1: To be complete this should include how to actually generate test coverage data.
This includes the workarounds mentioned in OP.

aDogCalledSpot: In line with the rest of the docs the example uses wasm-pack. How to do run the tests without wasm-pack is linked and arguments are documented.

guide/src/wasm-bindgen-test/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Arguments to the test runner
+
+If you are using `wasm-pack` to run the tests, refer to the `wasm-pack` documentation.
+
+Otherwise, you can use the following environment variables when [executing the test runner][1] to control the coverage output:
+
+[1]: usage.html#appendix-using-wasm-bindgen-test-without-wasm-pack
+
+- `WASM_BINDGEN_UNSTABLE_TEST_COVERAGE` to generate a single `.profraw` in your current working directory.
```
daxpedda, line 33: This is to enable it, right?

guide/src/wasm-bindgen-test/coverage.md
```diff
@@ -0,0 +1,45 @@
...
+- `WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_OUT` to control the file name of the profraw or the directory in which it is placed
+- `WASM_BINDGEN_UNSTABLE_TEST_PROFRAW_PREFIX` to add a custom prefix to the profraw files. This can be useful if you're running the tests automatically in succession.
+
+### Example
+
+```sh
+RUSTFLAGS="-Cinstrument-coverage -Zno-profiler-runtime --emit=llvm-ir" wasm-pack test --coverage --profraw-out cov_data/
+# Generate the debug info on the host
+clang target/wasm32-unknown-unknown/debug/deps/{your-dot-wasm-without-extension}.ll -Wno-override-module -c -o wasm.o
```
daxpedda, line 42: I assume `{your-dot-wasm-without-extension}` wants the test file, not the library file?
Would be nice of that could be clearer.

daxpedda:
So I managed to make it work with `#[coverage(off)]`: https://github.com/daxpedda/wasm-bindgen/commit/a81f944080c2f04beff660899671786b4917e822.
I probably went a bit overboard with tagging functions, but at least all `fn describe*()` functions have to be tagged.
The commit obviously can't be applied as-is because I didn't guard the `#![feature(coverage_attributes)]` nor the `#[coverage(off)]` attributes.

I also figured out what the issue is: https://rust.godbolt.org/z/YTdPKqPj3.
Apparently `#[coverage(off)]` doesn't account for inlining functions.
This might be a Wasm-specific issue, I didn't check.

Would you mind trying it and see if it works for you as well?
I'm happy to make a PR on your branch and fix it up properly.

aDogCalledSpot:
This does indeed work. I tried moving the feature to a `cfg` option like `web_sys_unstable_apis` but I'm unsure about the usage because passing the `cfg` option to `RUSTFLAGS` doesn't seem to pass it through when specifying a different target. I can't get it to work for `web_sys_unstable_apis` either. 

`RUSTFLAGS="--cfg=unstable_coverage" cargo build --target wasm32-unknown-unknown` doesn't work
`RUSTFLAGS="--cfg=unstable_coverage" cargo build` works
`CARGO_BUILD_TARGET="wasm32-unknown-unknown" RUSTFLAGS="--cfg=unstable_coverage" cargo build` doesn't work either.

daxpedda:
The problem is probably that `wasm-bindgen-test-runner` doesn't read `RUSTFLAGS`, so it has to be an environment variable.

We might want to use an environment variable for `wasm-bindgen-macro` as well?

If that doesn't work we are not gonna get around having an environment variable for `wasm-bindgen-test-runner` and a `cfg` for `wasm-bindgen-macro`.

I know all this is a bit unfortunate, but I can't come up with a better solution, I'm all ears if you have one.

aDogCalledSpot:
> The problem is probably that `wasm-bindgen-test-runner` doesn't read `RUSTFLAGS`, so it has to be an environment variable.

I don't think so, since `RUSTFLAGS="cfg=unstable_coverage" cargo build --target wasm32-unknown-unknown` shouldn't run `wasm-bindgen-test-runner`.

I noticed that the flag is being passed in correctly to `wasm-bindgen-test` which pulls in `wasm-bindgen` as a dependency (also sees the `cfg`) which, in  turn, pulls in `wasm-bindgen-macro` which doesn't see the `cfg`. The whole path down from there doesn't appear to recognize the flag.

aDogCalledSpot:
I had a brief look at why `llvm-cov` can't find the debug info. The issue isn't the DWARF info. `llvm-dwarfdump hello_world.wasm` works fine. The problem is that [LLVM looks for a section called `__llvm_prf_names` and doesn't find it in the wasm](https://github.com/llvm/llvm-project/blob/main/llvm/lib/ProfileData/Coverage/CoverageMappingReader.cpp#L1098).

So the issue arises at build time.

daxpedda:
I'm going to close this due to inactivity.
It is going to continue to be tracked in https://github.com/rustwasm/wasm-bindgen/issues/2276.

This is a highly desired feature and I would be happy to review it if somebody wants to continue the work done here!

Issue #3315, opened 2023-02-16
`wasm-bindgen-wasm-interpreter` can't handle extra block


Niedzwiedzw:
```
Finished release [optimized] target(s) in 0.08s
[INFO]: License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory
[INFO]: Installing wasm-bindgen...
thread 'main' panicked at 'unknown instruction Block(Block { seq: Id { idx: 1 } })', crates/wasm-interpreter/src/lib.rs:367:18
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: wasm_bindgen_wasm_interpreter::Interpreter::call
   3: wasm_bindgen_wasm_interpreter::Interpreter::interpret_descriptor
   4: wasm_bindgen_cli_support::descriptors::execute
   5: wasm_bindgen_cli_support::Bindgen::generate_output
   6: wasm_bindgen_cli_support::Bindgen::generate
   7: wasm_bindgen::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Error: Running the wasm-bindgen CLI
Caused by: failed to execute `wasm-bindgen`: exited with exit status: 101
  full command: "/home/niedzwiedz/.cache/.wasm-pack/wasm-bindgen-5f26acfc988649a3/wasm-bindgen" "/home/niedzwiedz/crionis/crio-app/target/wasm32-unknown-unknown/release/editor.wasm" "--out-dir" "/home/niedzwiedz/crionis/crio-app/pkg" "--typescript" "--target" "bundler" "--out-name" "index"
```

EDIT: https://github.com/rustwasm/wasm-pack/issues/1229 <- I've opened this issue here to, cause I have no idea what causes the bug

EDIT: https://github.com/rustwasm/wasm-bindgen/blob/629a623adfe27c67a21408c120d3466965575e45/crates/wasm-interpreter/src/lib.rs#L367 this is the line that crashes

csnover:
[...]

PR #3784, opened 2024-01-14
Apply only allowed attributes to generated functions

seancribbs:
This is a response to #3315. @daxpedda [suggested](https://github.com/rustwasm/wasm-bindgen/issues/3315#issuecomment-1879052173) that we pass down specific attributes and then allow the user to specify additional allowed attributes that could be attached to the bindgen and descriptor functions. 

This is an initial implementation of the first part of this idea, namely, that only the `cfg` and `cfg_attr` attributes are allowed to be passed through. There may be others that should be allowed by default, but since these are involved in conditional compilation, I felt they would be most relevant to apply to the generated functions.

I'm mostly looking for feedback on the direction here before I dive too much deeper. Thanks!

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -678,6 +678,14 @@ impl TryToTokens for ast::Export {
         };
         let nargs = self.function.arguments.len() as u32;
         let attrs = &self.function.rust_attrs;
+        let safe_attrs = attrs
+            .iter()
+            .filter(|a| {
+                let path = a.path();
+                path.is_ident("cfg") || path.is_ident("cfg_attr")
+            })
+            .cloned()
+            .collect::<Vec<_>>();
 
         let start_check = if self.start {
             quote! { const _ASSERT: fn() = || -> #projection::Abi { loop {} }; }
@@ -688,7 +696,7 @@ impl TryToTokens for ast::Export {
         (quote! {
             #[automatically_derived]
             const _: () = {
-                #(#attrs)*
+                #(#safe_attrs)*
                 #[cfg_attr(
                     all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))),
                     export_name = #export_name,
@@ -745,7 +753,7 @@ impl TryToTokens for ast::Export {
                 #describe_args
                 #describe_ret
             },
-            attrs: attrs.clone(),
+            attrs: safe_attrs,
             wasm_bindgen: &self.wasm_bindgen,
         }
         .to_tokens(into);


crates/backend/src/codegen.rs
```diff
@@ -745,7 +753,7 @@ impl TryToTokens for ast::Export {
                 #describe_args
                 #describe_ret
             },
-            attrs: attrs.clone(),
+            attrs: safe_attrs,
```
daxpedda, line 756: I believe this line could be removed entirely, right?

daxpedda (changes requested):
So by default we have to pass through all attributes, because anything else is a breaking change.

What we can do is to only pass `cfg` and `cfg_attr` by default for the functions generated by `wasm-bindgen`, e.g. `*_describe_*()`. This is theoretically still a breaking change but I think proc-macros that generate `cfg`s are pretty rare (which are the only ones we really need to support here).

I believe these changes along should suffice to solve #3315, adding an attribute to allow passing down more specific user-specified attributes is not necessary until requested.

I'm not entirely sure but I think you have already achieved what I outlined above with this PR.
It would be great if we can add a test.

@Liamolucko considering this is theoretically a breaking change, I would very much like your approval as well here.

crates/backend/src/codegen.rs
daxpedda: Oops, had to expand the code a bit to see what's going on here.
So yeah, ignore what I said above.

seancribbs:
> It would be great if we can add a test.

Yes, it wasn't clear where to add tests. It's not a compilation error/warning, so the ui-tests didn't seem to apply. Any advice?

daxpedda:
> Yes, it wasn't clear where to add tests. It's not a compilation error/warning, so the ui-tests didn't seem to apply. Any advice?

I would have imagined them to be added via compilation errors as UI tests. E.g. use `cfg` attributes and check if methods are still accessible. Serde could be used to test foreign attributes.

Of course ideally we would have proper proc-macro tests, if you feel like adding them that would be great. But I'm happy to accept UI tests instead of nothing.

daxpedda:
Apologies, after further thinking I believe you are right, this doesn't make sense in a UI test (and Serde doesn't have any method related attributes I believe).

I believe non-UI tests can be added in `/tests/wasm/file.rs` (you don't have to add a corresponding `.js` file).

seancribbs:
@daxpedda I was able to add a test to the existing test suite in `tests/wasm`. Please let me know if that is sufficient to demonstrate the efficacy of the change.

tests/wasm/attributes.rs
daxpedda, line 1: Can we add another test with custom proc-macros, to make sure they are skipped for the generated functions?
Feel free to add any reasonable dev-dependencies to make it happen, e.g. `tracing`.

daxpedda (changes requested):

Liamolucko:
> @Liamolucko considering this is theoretically a breaking change, I would very much like your approval as well here.

I think it should be fine, I'd be very surprised if anyone's relying on attributes getting copied to `wasm-bindgen`'s internal functions.

Issue #3766, opened 2024-01-08
Improve type assertion messages


rimutaka:
### Motivation

Type checking functions inform what type was expected, but not what type was encountered. E.g. 

```javascript
function _assertBoolean(n) {
    if (typeof(n) !== 'boolean') {
        throw new Error('expected a boolean argument');
    }
}
```
It makes debugging harder because the encountered type can be anything and it requires manual edits to figure out what's going on. Time wasted.

### Proposed Solution

Change the message to something like

```javascript
        throw new Error(`expected a boolean argument, encountered ${typeof(n)}`);
```

### Alternatives

1. Manually edit _wasm_mod.js_ to print the encountered type for debugging
2. Print the type before the param reaches _wasm_mod.js_

Both feel wasteful. I think printing it out as part of the error has zero cost and saves a lot of grief.

It would also be inline with the helpful ways of Rust compiler.

### Would maintainers accept a PR for this?

I'm happy to work on a PR if there is interest to add this feature.

daxpedda:
Yeah, I only see upsides.
Happy to review a PR!

rimutaka:
@daxpedda , I hit a snag with adding tests.
The docs explain some of it, but a lot remains to be discovered by the contributors.
I'm slowly getting there :)

Would you consider a PR explaining the testing framework in more detail and some how-to's?

daxpedda:
> Would you consider a PR explaining the testing framework in more detail and some how-to's?

I'm unsure exactly what you have in mind, but I'm happy to review any PRs improving documentation.
Just check up with me (or any other maintainer) before doing any significant amount of work.

daxpedda:
Fixed in #3787.

daxpedda:
Thanks @rimutaka, it's very appreciated!

PR #3787, opened 2024-01-16
 Added "type found" msg to expose_assert_* fns #3766

rimutaka:
This PR prints out the type of the argument encountered when the type assertion fails as discussed in #3766

- old msg: `expected a number argument`
- mew msg: `expected a number argument, found string`

#### The core change
* added `, found ${typeof(n)}` to int, str and bool assertions

#### Related changes

Simple tests did not have anything for boolean type assertion, so I added one because this PR changes the bool assertion and needs a test.  
Was it included elsewhere and mine is a duplicate?

Only negative assertions were tested. E.g. passing a string to a number arg.  
I added positive tests as well. E.g. passing a number to a number arg.  
Maybe there was a reason why they were omitted?

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1147,7 +1147,7 @@ impl<'a> Context<'a> {
         self.global(
             "
             function _assertNum(n) {
-                if (typeof(n) !== 'number') throw new Error('expected a number argument');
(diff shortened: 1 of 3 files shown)


daxpedda (approved):
I'm not sure if we had any tests in place somewhere else or why some tests where omitted. But my guess is that a lot of tests are just missing and I'm happy that you added those.

Thank you!

PR #3789, opened 2024-01-17
Introduce Array::from_slice for easy array building

Ekleog:
Hey! I've been looking for a way to easily build `Array`s, and eventually gave up. So I'm submitting this PR, that I think would be useful to easily build `Array`s of simple types at least :)

One thing that made me sad is, I apparently was unable to find a way to make slice-of-str typecheck without adding an `impl From<&&str> for JsValue`. I'd have loved to make it `impl<T> From<&T> for JsValue where JsValue: From<T>`, but 1. I'm not even sure whether that'd be allowed by orphan rules, and 2. I'm pretty sure it'd break backward-compatibility anyway.

Maybe I missed the one trait I should use, that'd support any of the ways to specify a JsValue? I can already say that `AsRef<JsValue>` is not it, because `["foo"].into_iter().collect::<Array>()` does not typecheck; and thus `JsCast` must not be it either.

For some context, I'm needing a way to easily build arrays for the users of [my IndexedDB bindings crate](https://crates.io/crates/indexed-db), as I just noticed that doing it inside the crate itself prevents some potential use cases (aka multi-entry indices).

What do you think about this? :)

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -748,6 +748,18 @@ impl Array {
 
         output
     }
+
+    /// Converts a slice to a JS array
+    pub fn from_slice<'a, T>(slice: &'a [T]) -> Array
+    where
+        JsValue: From<&'a T>,
+    {
+        let res = Array::new_with_length(u32::try_from(slice.len()).unwrap());
+        for (i, v) in slice.iter().enumerate() {
+            res.set(i as u32, JsValue::from(v));
+        }
+        res
+    }
 }
 
 impl std::iter::IntoIterator for Array {
--- a/crates/js-sys/tests/wasm/Array.rs
+++ b/crates/js-sys/tests/wasm/Array.rs
@@ -99,6 +99,14 @@ fn to_vec() {
 }
 
 #[wasm_bindgen_test]
+fn from_slice() {
+    assert_eq!(
+        Array::from_slice(&["a", "b", "c"]).to_vec(),
+        vec![JsValue::from("a"), JsValue::from("b"), JsValue::from("c")],
+    );
+}
+
+#[wasm_bindgen_test]
 fn iter() {
     let array = vec![JsValue::from("a"), JsValue::from("b"), JsValue::from("c")]
         .into_iter()
--- a/crates/macro/ui-tests/async-errors.stderr
+++ b/crates/macro/ui-tests/async-errors.stderr
@@ -27,6 +27,7 @@ error[E0277]: the trait bound `wasm_bindgen::JsValue: From<BadType>` is not sati
    | ^^^^^^^^^^^^^^^ the trait `From<BadType>` is not implemented for `wasm_bindgen::JsValue`
    |
    = help: the following other types implement trait `From<T>`:
+             <wasm_bindgen::JsValue as From<&'a &'b str>>
              <wasm_bindgen::JsValue as From<&'a String>>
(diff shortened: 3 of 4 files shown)


Ekleog:
Looks like there are quite a few places where wasm-bindgen-macro compile-fail message changes with the From<&&str>. I'll fix them all if you confirm you want this new impl for JsValue, as it doesn't make much sense if I'm going to change the trait or discard this PR altogether :)

daxpedda:
What exactly are you trying to do that you are trying to solve with this PR?
Could you explain that with code?

Ekleog:
My context is both these functions:
- https://docs.rs/indexed-db/latest/indexed_db/struct.ObjectStore.html#method.build_index
- https://docs.rs/indexed-db/latest/indexed_db/struct.Index.html#method.get_all_in

90% of the time, they will be taking slice arguments, so I wrote their API as taking slices in. However, I just discovered that some parts of IndexedDB, that I thought would work fine with Arrays, do require scalar values (aka multi-entry indexes).

My goal is to make it easy for the user, to call these functions both with a scalar value and with a temporary slice.

In other words:
```rust
// Instead of this:
let keys = Array::new();
keys.push(JsString::from("foo"));
keys.push(JsString::from("bar"));
store.build_index("baz", keys);

// I'd like the user to be able to write this:
store.build_index("baz", &["foo", "bar"]);

// But that'd break some parts of IndexedDB, so I'd like to fall back to this:
store.build_index("baz", JsValue::from(&["foo", "bar"]));
```

You can see examples of that code being used in practice here:
https://github.com/Ekleog/crdb/blob/e3987d5edfbb3c2af279e9e303eab6e8e13dc456/src/client/indexed_db.rs#L58-L90

Or here:
https://github.com/Ekleog/crdb/blob/e3987d5edfbb3c2af279e9e303eab6e8e13dc456/src/client/indexed_db.rs#L153-L157

To understand why it'd be a bit of a mess if I forced the user to build the Array manually each time.

That being said, if you think this is not good for wasm-bindgen itself, I'll just expose a macro to help users build the arrays, but I think such a function could be useful for other use cases too :)

daxpedda:
Does using `Array::from_iter()` solve your problem?
```rust
store.build_index("baz", Array::from_iter(["foo", "bar"].map(JsValue::from)));
```

I'm aware this isn't as nice as what you proposed, but considering it's not actually adding an efficient conversion, I would like to not add any more high-level stuff to `JsValue`.

Ekleog:
Considering the space it still takes (and the fact it's not very discoverable), I guess I'll still write a macro to make that easier for users, like `compound_key!("foo", "bar")`.

Actually that makes me think, maybe an `array![]` macro, similar to the `vec![]` macro, would be more reasonable to introduce into `js-sys`. WDYT about it?

daxpedda:
> Actually that makes me think, maybe an `array![]` macro, similar to the `vec![]` macro, would be more reasonable to introduce into `js-sys`. WDYT about it?

I would prefer not to. This hasn't really been a written down or enforced policy so far, but I would really like to keep `wasm-bindgen`/`js-sys`/`web-sys` as low-level as possible. Basically the only thing I aim to let live here is whats necessary for higher-level implementations to do their job. Everything else should go into crates like `gloo`.

Thank you for the conversation though, I really appreciate somebody thinking with me about this!

Ekleog:
I see, thank you for your answers and for taking the time to answer me! :)

PR #3797, opened 2024-01-19
use immutable buffers in SubtleCrypto methods

### LINKED ISSUE #3795: SubtleCrypto verify methods accepts mutable reference that should be immutable?
### Describe the Bug

Some of the verify methods on `SubtleCrypto` take mutable references to the `signature` bytes. It doesn't seem desirable that a method for verifying signatures would mutate the signature bytes, so I wonder if this is a mistake?

### Steps to Reproduce

1. Go to https://docs.rs/web-sys/0.3.67/web_sys/struct.SubtleCrypto.html
2. Note that some of the `verify____` methods accept `signature: &mut [u8]`

### Expected Behavior

I would expect that a signature verification would never need to mutate the signature.

### Actual Behavior

Parameter is explicitly set to be mutable; git history doesn't indicate a reason why.
https://github.com/rustwasm/wasm-bindgen/blob/55a7fb848a88c0e1200baf1bf670a051b14f5025/crates/web-sys/src/features/gen_SubtleCrypto.rs#L540

### Additional Context

The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/verify) don't suggest that the signature might be mutated. They don't promise it won't, so I'm just going on instinct here.

eric-seppanen:
This change makes all the methods in `SubtleCrypto` use `&[u8]` parameters instead of `&mut [u8]`.

Here are the affected parameters:

- `decrypt(data)`: the [WebCrypto API states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-encrypt):
  > Let data be the result of getting a copy of the bytes held by the data parameter passed to the decrypt method.
- `digest(data)`: the [WebCrypto API states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-digest): 
  > Let data be the result of getting a copy of the bytes held by the data parameter passed to the digest method.
- `encrypt(data)`: the [WebCrypto API states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-encrypt):
  > Let data be the result of getting a copy of the bytes held by the data parameter passed to the encrypt method.
- `sign(data)`: the [WebCrypto API states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-sign):
  > Let data be the result of getting a copy of the bytes held by the data parameter passed to the sign method.
- `unwrapKey(wrapped_key)`: the [WebCrypto API states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-unwrapKey):
  > Let wrappedKey be the result of getting a copy of the bytes held by the wrappedKey parameter passed to the unwrapKey method.
- `verify(signature, data)`: the [WebCrypto API states](https://www.w3.org/TR/WebCryptoAPI/#SubtleCrypto-method-verify):
  > Let signature be the result of getting a copy of the bytes held by the signature parameter passed to the verify method.
  > Let data be the result of getting a copy of the bytes held by the data parameter passed to the verify method.

I take the API documentation language "getting a copy of the bytes" to mean that the original buffer is left unmodified.

I can't find any evidence that these buffers would be modified. I'm not sure how strong an argument that is.

Fixes #3795.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Stabilize `ClipboardEvent`.
   [#3791](https://github.com/rustwasm/wasm-bindgen/pull/3791)
 
+* Use immutable buffers in `SubtleCrypto` methods.
+  [#3797](https://github.com/rustwasm/wasm-bindgen/pull/3797)
+
 ## [0.2.90](https://github.com/rustwasm/wasm-bindgen/compare/0.2.89...0.2.90)
 
 Released 2024-01-06
--- a/crates/web-sys/src/features/gen_SubtleCrypto.rs
+++ b/crates/web-sys/src/features/gen_SubtleCrypto.rs
@@ -49,7 +49,7 @@ extern "C" {
         this: &SubtleCrypto,
         algorithm: &::js_sys::Object,
         key: &CryptoKey,
-        data: &mut [u8],
+        data: &[u8],
     ) -> Result<::js_sys::Promise, JsValue>;
     #[cfg(feature = "CryptoKey")]
     # [wasm_bindgen (catch , method , structural , js_class = "SubtleCrypto" , js_name = decrypt)]
@@ -62,7 +62,7 @@ extern "C" {
         this: &SubtleCrypto,
         algorithm: &str,
         key: &CryptoKey,
-        data: &mut [u8],
+        data: &[u8],
     ) -> Result<::js_sys::Promise, JsValue>;
     #[cfg(feature = "CryptoKey")]
     # [wasm_bindgen (catch , method , structural , js_class = "SubtleCrypto" , js_name = deriveBits)]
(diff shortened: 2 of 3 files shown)


daxpedda (changes requested):
LGTM!
Can you add an entry in the changelog?

eric-seppanen:
Added a changelog entry.

Rebased due to merge conflicts in the changelog; I made no other changes from the previous version.

daxpedda (approved):
Thank you!

eric-seppanen:
I realize after the fact that this change made the 0.3.68 release not semver-compatible, because the signatures of some `SubtleCrypto` methods changed. Example: [0.3.67](https://docs.rs/web-sys/0.3.67/web_sys/struct.SubtleCrypto.html#method.verify_with_object_and_u8_array_and_buffer_source), [0.3.68](https://docs.rs/web-sys/0.3.68/web_sys/struct.SubtleCrypto.html#method.verify_with_object_and_u8_array_and_buffer_source).

Apologies for the breakage! Perhaps [cargo-semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) could have caught this?

daxpedda:
AFAIK changing a parameter from `&T` to `&mut T` should not be a breaking change.

Is that what you meant or did I miss something else?

eric-seppanen:
@daxpedda Oops, you're right. (Though it's `&mut T` -> `&T`).  I confused myself by copying some code from a workspace using 0.3.68 into another project using 0.3.67 and it didn't compile. But that's not semver breakage, it's just that downgrade doesn't work. Sorry for the noise.

PR #3800, opened 2024-01-20
Support no-capture behavior in wasm-bindgen-test

Ekleog:
Note: this is still a draft, I need to put that behind a config option. But before I complete it all, I wanted to get an idea what y'all think about this?

My context is, I want to fuzz a crate of mine that needs to run in a browser. I cannot fuzz for now, because fuzzers don't support wasm32. So I want to at least have a very long-running proptest that generates random data.

But then in order to be able to know what's happening, I now need to have a view of what the test actually displays, so that I can distinguish between "running fine" and "deadlocked/panicked".

So before I turn it into a proper configuration option, WDYT about this change?

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
@@ -145,10 +145,15 @@ pub fn run(server: &SocketAddr, shell: &Shell, timeout: u64) -> Result<(), Error
     shell.status("Waiting for test to finish...");
     let start = Instant::now();
     let max = Duration::new(timeout, 0);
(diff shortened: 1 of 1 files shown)


daxpedda:
I'm not entirely sure what exactly you are trying to show in the shell, but in general it would be good if we align more with `cargo test`.
- We should show test results in the shell as soon as possible, no reason to wait until everything is done.
- Using [`--nocapture`](https://doc.rust-lang.org/1.75.0/cargo/commands/cargo-test.html#display-options) should I guess show the console output in the shell if there is any.

Currently the output in the shell is a complete mess, so any improvements on that front are appreciated.

PR #3803, opened 2024-01-21
Implement `--skip` flag in test runner

daxpedda:
This implements the [`--skip` flag](https://doc.rust-lang.org/1.75.0/rustc/tests/index.html#--skip-filter) for `wasm-bindgen-test-runner`. I tried to stay as close as possible to the behavior of `cargo test`, including the output.

### DIFF
--- a/crates/test/src/rt/mod.rs
+++ b/crates/test/src/rt/mod.rs
@@ -130,8 +130,14 @@ struct State {
     /// Include ignored tests.
     include_ignored: Cell<bool>,
 
+    /// Tests to skip.
+    skip: RefCell<Vec<String>>,
+
     /// Counter of the number of tests that have succeeded.
     succeeded: Cell<usize>,
+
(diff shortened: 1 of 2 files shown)


crates/test/src/rt/mod.rs
```diff
@@ -289,20 +297,23 @@ impl Context {
...
             let arg = arg.as_string().unwrap();
             if arg == "--include-ignored" {
                 self.state.include_ignored.set(true);
+            } else if arg == "--skip" {
+                skip.push(
+                    args.next()
+                        .expect("Argument to option 'skip' missing")
+                        .as_string()
+                        .unwrap(),
+                );
```
Liamolucko, line 316: This doesn't quite match `cargo test`'s parsing, since `cargo test` supports `--skip=foo` as well as `--skip foo`.

daxpedda: Done.

PR #3808, opened 2024-01-22
Shrink JS-allocated strings down to the correct size before passing them to Rust

### LINKED ISSUE #3801: UB when copying strings to WASM.
### Describe the Bug

When converting a JS string into a Rust string via `JsString::into`, dropping the produced String generates a layout that violates the `dealloc` contract.

### Steps to Reproduce

1. Checkout https://github.com/khoover/base32768-rs/tree/talc-error
2. From `base32768-rs/examples/js-bench`, run
```
RUSTFLAGS="-C target-cpu=mvp" wasm-pack build --debug --target nodejs
node bench-runner.js
```
3. See error

### Expected Behavior

Some benchmark timings being emitted to console.

### Actual Behavior

A `RuntimeError` crashes node while the benchmarks are running.

```
khoover@DESKTOP-CUT25MA:~/base32768-rs/examples/js-bench$ node pkg/bench-runner.js 
Took 7 ms to load and init bench wasm
wasm://wasm/00189ade:1


RuntimeError: unreachable
    at __rust_start_panic (wasm://wasm/00189ade:wasm-function[1201]:0x449d6)
    at rust_panic (wasm://wasm/00189ade:wasm-function[1173]:0x44891)
    at std::panicking::rust_panic_with_hook::hc20eadded6bfe687 (wasm://wasm/00189ade:wasm-function[363]:0x32839)
    at std::panicking::begin_panic_handler::{{closure}}::h82415fe35b0e2001 (wasm://wasm/00189ade:wasm-function[443]:0x36681)
    at std::sys_common::backtrace::__rust_end_short_backtrace::h71f504d46a203d88 (wasm://wasm/00189ade:wasm-function[1196]:0x449b3)
    at rust_begin_unwind (wasm://wasm/00189ade:wasm-function[708]:0x3e2fa)
    at core::panicking::panic_fmt::h7a368385936888dc (wasm://wasm/00189ade:wasm-function[947]:0x425b9)
    at core::panicking::panic::h7bbea3773b752235 (wasm://wasm/00189ade:wasm-function[859]:0x41116)
    at talc::Talc<O>::free::h14aa50d45611ce00 (wasm://wasm/00189ade:wasm-function[81]:0x1a142)
    at <talc::talck::Talck<R,O> as core::alloc::global::GlobalAlloc>::dealloc::hcf08400e756d2221 (wasm://wasm/00189ade:wasm-function[496]:0x386e4)
```

### Additional Context

Initially reported as an allocator bug in https://github.com/SFBdragon/talc/issues/24. After [investigation](https://github.com/SFBdragon/talc/issues/24#issuecomment-1902467670), seems like the allocator is being passed a different layout than the most recent alloc for the given pointer; std doesn't catch because it [ignores layout information](https://github.com/alexcrichton/dlmalloc-rs/issues/32) for wasm32-unknown-unknown.

Also of note: the `Box<[u16]>` produced by the `encode` function can contain unpaired surrogates and other potentially invalid BMP codepoints. If we comment out the function that tries to do a round trip on these unpaired surrogate-containing sequences, the program stops crashing even with `Talck` as the allocator.

This triggers UB for when using, e.g., `let s: String = js_string.into();` on any JS string that contains non-ASCII codepoints, where at least one requires 2 or 4 UTF-8 bytes to encode.

Liamolucko:
Fixes #3801.

I opted to solve it this way rather than just pass the capacity to Rust as well because it means the allocation isn't up to 3x bigger than it needs to be anymore. I also removed a TODO about fixing that.


crates/cli-support/src/js/mod.rs
```diff
@@ -1314,6 +1309,7 @@ impl<'a> Context<'a> {
                     const ret = encodeString(arg, view);
                     {debug_end}
                     offset += ret.written;
+                    ptr = realloc(ptr, len, offset, 1) >>> 0;
```
khoover, line 1312: Could add a check for whether `len != offset` before reallocing again, to avoid spurious allocs.

Liamolucko: I considered that, but I didn't think it'd be worth it since in the vast majority of cases the string won't take up the maximum amount of space anyway.

I'd be fine to add it if you want but I don't think the difference would be significant either way.

daxpedda (commented):
Would you mind adding a changelog entry as well considering this is an actual user-facing fix?

PR #3811, opened 2024-01-25
Run `cargo publich` on deployment

daxpedda:
Unfortunately we didn't find the time to properly fix up the `bump` workflow, so in the meantime I have opted to replace it by just running `cargo publish` in the deployment step.


.github/workflows/main.yml
```diff
@@ -518,3 +518,8 @@ jobs:
       if: startsWith(github.ref, 'refs/tags/')
       with:
         files: "gh-release/*.tar.gz*"
+    - uses: dtolnay/rust-toolchain@stable
+    - run: rustc publish.rs
+    - run: ./publish publish
+      env:
+        CRATES_IO_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
```
Liamolucko, line 525: I think you need to copy the `if: startsWith(github.ref, 'refs/tags/')` from the above GitHub release step down to here. Right now I think this will trigger a publish on every push to `main`.

.github/workflows/main.yml
```diff
@@ -518,3 +518,8 @@ jobs:
       if: startsWith(github.ref, 'refs/tags/')
       with:
         files: "gh-release/*.tar.gz*"
+    - uses: dtolnay/rust-toolchain@stable
```
Liamolucko, line 521: Rust is already installed earlier within this CI job, so you don't need to do it again.

I'm not actually particularly sure why, it doesn't seem to be used and it's also nightly for some reason. But it shouldn't be installed twice, so one of them has to be removed.

Liamolucko (changes requested):

.github/workflows/main.yml
daxpedda: Done.

.github/workflows/main.yml
daxpedda: I removed the original one because I too can't see where it's used, if something is wrong it should fail at regular CI already I believe.

.github/workflows/main.yml
```diff
@@ -518,3 +517,11 @@ jobs:
...
+      if: startsWith(github.ref, 'refs/tags/')
+    - run: rustc publish.rs
+      if: startsWith(github.ref, 'refs/tags/')
+    - run: ./publish publish
+      if: startsWith(github.ref, 'refs/tags/')
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
```
daxpedda, line 527: This was `CRATES_IO_TOKEN` before in `bump.yml`, which I think was incorrect.
I can't actually check the name of the secret, but worst case scenario we will find out when we try and publish.

Liamolucko (approved):

PR #3813, opened 2024-01-25
Fix example code in guide

akiomik:
An error occurs because the function `cancelInterval` does not exist in js. The correct function is `clearInterval`.

See: https://developer.mozilla.org/en-US/docs/Web/API/clearInterval


daxpedda (approved):
Thank you!

Issue #1769, opened 2019-09-16
Example doesn't mention running a HTTP Server


matthewrobertbell:
https://github.com/rustwasm/wasm-bindgen/tree/master/examples/without-a-bundler

This works fine when running build.sh. Literally following the instructions (build then open index.html) doesn't work, because of CORS. Adding a mention to run "python3 -m http.server" should make this more friendly.

Thanks for your work!

fitzgen:
Good idea! Care to make a pull request?

daxpedda:
Resolved by #3819.

PR #3819, opened 2024-01-31
Add deployment note for first `wasm-pack build --target web` example

maartin0:
Added a note to the example for a minimal build without a bundler (see: #1769)


guide/src/examples/without-a-bundler.md
```diff
@@ -35,9 +35,18 @@ Otherwise the rest of the deployment magic happens in `index.html`:
 {{#include ../../../examples/without-a-bundler/index.html}}
 ```
 
+> **Note**: You cannot directly open `index.html` in your web browser due to [CORS][cors]
+> limitations. Instead, you can set up a quick development environment using
+> Python's built-in HTTP server:
+> ```sh
+> wasm-pack build --target web
+> python3 -m http.server 8080
```
stefnotch, line 43: Would it be possible to suggest some other kind of HTTP server? (One that's based on Node.js or Rust)

As for why I'm asking: Installing Python on Windows can be a bit finicky and does not always lead to a good user experience.

stefnotch: I don't see any changes, I'm guessing something got lost during a git command somewhere?

maartin0: Oops thanks for pointing that out; should be fixed now I think

stefnotch: Yay, nice

daxpedda (approved):

daxpedda:
Thank you!

maartin0:
No problem, do you want to close the issue as well?

PR #3825, opened 2024-02-05
readme: build badge for Github actions, not Azure pipeline

bergey:
The last build on the Azure pipeline was in 2021, so tracking its success doesn't seem useful.  The link here only shows builds on branch `main`, but unfortunately that includes branches named `main` on other Github forks.  I would not be surprised if the badge has the same issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
   </p>
 
   <p>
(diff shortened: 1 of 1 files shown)


daxpedda (approved):
Thank you!

PR #3826, opened 2024-02-05
readme: also link the guide built from the main branch

bergey:
I was confused by the two copies of the Guide on the website, and I hope this will prevent similar confusion for others.  I haven't actually tracked down how these are deployed, so please correct me if I have misunderstood the versions.  The changes from [this commit touching the Guide](https://github.com/rustwasm/wasm-bindgen/commit/fb518d3b190ca8885dbac5e149ce1bbef3d8a92b) are present [here](https://rustwasm.github.io/wasm-bindgen/reference/passing-rust-closures-to-js.html) but not [here](https://rustwasm.github.io/docs/wasm-bindgen/reference/passing-rust-closures-to-js.html).

The [pre-release version of the wasm-pack book](https://rustwasm.github.io/wasm-pack/book/) also has a banner at the top linking to the release version.  LMK if you want similar for this Guide, I will try to figure out how.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@
   </p>
 
   <h3>
-    <a href="https://rustwasm.github.io/docs/wasm-bindgen/">Guide</a>
+    <a href="https://rustwasm.github.io/docs/wasm-bindgen/">Guide (release)</a>
+    <span> | </span>
+    <a href="https://rustwasm.github.io/wasm-bindgen/">Guide (main branch)</a>
     <span> | </span>
     <a href="https://docs.rs/wasm-bindgen">API Docs</a>
     <span> | </span>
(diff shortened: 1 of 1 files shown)


daxpedda:
There is an issue currently with deploying the release version, the one currently up is completely outdated.
I would prefer not to link it at all until it is actually fixed.

bergey:
Makes sense.  LMK if that's something I can help with.

daxpedda:
> Makes sense. LMK if that's something I can help with.

I honestly have no clue about the deployment of the pages.
My guess is that its done somewhere in the CI, so feel free to dig around there and see whats going on.

daxpedda (approved):
Thank you!

PR #3840, opened 2024-02-08
allow tracks without streams

Niedzwiedzw:


Niedzwiedzw:
as per
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addTrack#adding_tracks_to_multiple_streams

media streams are actually optional

daxpedda (changes requested):
We only generate Web APIs, we don't add them manually. Please consult the [`wasm-bindgen` Book](https://rustwasm.github.io/wasm-bindgen/contributing/web-sys/supporting-more-web-apis.html) on how to add or change Web API generation.

In this case the WebIDL has changed as in the specification you have linked, which we have to apply here as well.

PR #3847, opened 2024-02-20
Introduce impl TryFrom for Number that succeeds iff the value is within the safe range

Ekleog:
As a counterpoint, raised by MaulingMonkey on discord: it could also make sense to round the number, so maybe TryFrom<u64> & co should not be implemented for Number?

OTOH, I personally think the added convenience is worth it, and by implementing TryFrom rather than From we make it explicit that the conversion is fallible, which should make it obvious that the behavior is "error out if the number is outside the safe range". Especially considering that `Number` in JavaScript is often used to describe integers in the safe range, probably much more often than it is actually used to describe floating point numbers.

A design question I had while writing this: should the `Error` type be just the passed-in value? It seems to be the way other `TryFrom` implementations in `js-sys` are currently done, so I went ahead with the same idea here.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -2793,6 +2793,24 @@ macro_rules! number_from {
 }
 number_from!(i8 u8 i16 u16 i32 u32 f32 f64);
 
+macro_rules! number_try_from {
+    ($($x:ident)*) => ($(
+        impl TryFrom<$x> for Number {
+            type Error = $x;
+
+            #[inline]
+            fn try_from(x: $x) -> Result<Number, Self::Error> {
+                if x as f64 >= Number::MIN_SAFE_INTEGER && x as f64 <= Number::MAX_SAFE_INTEGER {
+                    Ok(Number::unchecked_from_js(JsValue::from(x as f64)))
+                } else {
+                    Err(x)
+                }
+            }
+        }
+    )*)
+}
+number_try_from!(i64 u64 i128 u128);
+
 // TODO: add this on the next major version, when blanket impl is removed
 /*
 impl convert::TryFrom<JsValue> for Number {
--- a/crates/js-sys/tests/wasm/Number.rs
+++ b/crates/js-sys/tests/wasm/Number.rs
@@ -1,4 +1,7 @@
-use std::f64::{INFINITY, NAN};
+use std::{
+    convert::TryFrom,
+    f64::{INFINITY, NAN},
+};
 
 use js_sys::*;
 use wasm_bindgen::prelude::*;
@@ -69,6 +72,21 @@ fn new() {
     let v = JsValue::from(n);
     assert!(v.is_object());
     assert_eq!(Number::from(v).value_of(), 42.);
+}
+
+#[wasm_bindgen_test]
+fn try_from() {
+    assert_eq!(Number::try_from(42u128).unwrap(), 42.);
+    assert_eq!(
+        Number::try_from(Number::MAX_SAFE_INTEGER as u64).unwrap(),
+        Number::MAX_SAFE_INTEGER
+    );
+    assert_eq!(
+        Number::try_from(Number::MIN_SAFE_INTEGER as i128).unwrap(),
+        Number::MIN_SAFE_INTEGER
+    );
+    assert!(Number::try_from(Number::MAX_SAFE_INTEGER as u128 + 1).is_err());
+    assert!(Number::try_from(Number::MIN_SAFE_INTEGER as i64 - 1).is_err());
 }
 
 #[wasm_bindgen_test]


crates/js-sys/src/lib.rs
```diff
@@ -2793,6 +2793,24 @@ macro_rules! number_from {
...
 
+macro_rules! number_try_from {
+    ($($x:ident)*) => ($(
+        impl TryFrom<$x> for Number {
+            type Error = $x;
+
+            #[inline]
+            fn try_from(x: $x) -> Result<Number, Self::Error> {
+                if x as f64 >= Number::MIN_SAFE_INTEGER && x as f64 <= Number::MAX_SAFE_INTEGER {
+                    Ok(Number::unchecked_from_js(JsValue::from(x as f64)))
```
daxpedda, line 2804:
```suggestion
                let x_f64 = x as f64;

                if x_f64 >= Number::MIN_SAFE_INTEGER && x_f64 <= Number::MAX_SAFE_INTEGER {
                    Ok(Number::unchecked_from_js(JsValue::from(x_f64)))
```

crates/js-sys/tests/wasm/Number.rs
```diff
@@ -1,4 +1,7 @@
-use std::f64::{INFINITY, NAN};
+use std::{
+    convert::TryFrom,
+    f64::{INFINITY, NAN},
```
daxpedda, line 3:
```suggestion
```
I assume this was from an earlier iteration?

daxpedda (changes requested):
> A design question I had while writing this: should the Error type be just the passed-in value? It seems to be the way other TryFrom implementations in js-sys are currently done, so I went ahead with the same idea here.

Where did you see that in `js-sys`?

---

I think I'm also in favor of doing the conversion like this, but FWIW alternatively we could make it infallible by using -/+infinity when the number is not in range.

Please add a changelog entry as well.

Liamolucko:
> > A design question I had while writing this: should the Error type be just the passed-in value? It seems to be the way other TryFrom implementations in js-sys are currently done, so I went ahead with the same idea here.
> 
> Where did you see that in `js-sys`?

The `TryFrom<BigInt>` impls for `i64`, `u64`, `i128` and `u128` do that, as well as all the `TryFrom` implementations in `wasm-bindgen`.

The argument against doing the same here would be that all those impls convert from `JsValue` (or an imported JS type) to a standard Rust type; you might want to get the `JsValue` back on failure to avoid having to needlessly clone it. In this case, we're converting from an integer to a `JsValue`, so there isn't as much need to get the original value back on failure when the input's `Copy` anyway.

That would line up with what `std` does: looking at all of `std`'s implementations of `TryFrom<T>`, it seems like whenever `T` can't be cheaply cloned it gives back the `T` on failure, otherwise it uses a real error type. For example `impl TryFrom<Vec<T>> for [T; N]` has `type Error = Vec<T>`, whereas `impl TryFrom<&[T]> for [T; N]` has `type Error = FromSliceError`. I don't actually know if this is an official policy or anything but it seems to be the case.

The closest analogy to `JsValue` would be `impl TryFrom<Arc<[T]>> for Arc<[T; N]>`, since both are cheap but not free to clone: and it gives back the `Arc` on failure. So I guess using a proper error type would make more sense.

crates/js-sys/tests/wasm/Number.rs
Ekleog: I'm not sure what you mean here? These imports were already present before my changes, and do seem to be required

daxpedda: Apologies, I was sleep walking apparently :sweat_smile:!

Ekleog:
Thank you for your feedback! I have:
- hoisted `x_f64` as per the suggestion
- added a changelog entry
- switched the code to use a `TryFromIntError` type, copy-pasted from the libstd, the only change being that I didn't import `Error` globally. We apparently cannot construct the libstd's variant easily, and doing eg. `u32::try_from(u64::MAX)?` to return it would look like too much of a hack for me — let me know if you'd have preferred that approach

Hopefully this is good to go! :)

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 * Add bindings for `CanvasState.reset()`, affecting `CanvasRenderingContext2D` and `OffscreenCanvasRenderingContext2D`.
   [#3844](https://github.com/rustwasm/wasm-bindgen/pull/3844)
 
+* Add `TryFrom` implementations for `Number`, that allow losslessly converting from 64- and 128-bits numbers
```
daxpedda, line 14:
```suggestion
* Add `TryFrom` implementations for `Number`, that allow losslessly converting from 64- and 128-bits numbers.
  [#3847](https://github.com/rustwasm/wasm-bindgen/pull/3847)
```

crates/js-sys/src/lib.rs
```diff
@@ -2793,6 +2793,45 @@ macro_rules! number_from {
...
+macro_rules! number_try_from {
+    ($($x:ident)*) => ($(
+        impl TryFrom<$x> for Number {
+            type Error = $x;
+
+            #[inline]
+            fn try_from(x: $x) -> Result<Number, Self::Error> {
+                let x_f64 = x as f64;
+                if x_f64 >= Number::MIN_SAFE_INTEGER && x_f64 <= Number::MAX_SAFE_INTEGER {
+                    Ok(Number::unchecked_from_js(JsValue::from(x_f64)))
```
daxpedda, line 2825:
```suggestion
                    Ok(Number::from(x_f64))
```
Just figured this could work as well.

daxpedda (changes requested):
> * switched the code to use a `TryFromIntError` type, copy-pasted from the libstd, the only change being that I didn't import `Error` globally. We apparently cannot construct the libstd's variant easily, and doing eg. `u32::try_from(u64::MAX)?` to return it would look like too much of a hack for me — let me know if you'd have preferred that approach

Yeah, my main concern with using Stds `TryFromIntError` type is that we aren't really in control of what it displays exactly so it could become confusing in the future. Though I'm not happy with adding a dedicated error type for this I don't have a good alternative suggestion.

So except the couple of nits its LGTM.

crates/js-sys/src/lib.rs
```diff
@@ -2793,6 +2793,37 @@ macro_rules! number_from {
 }
 number_from!(i8 u8 i16 u16 i32 u32 f32 f64);
 
+/// The error type returned when a checked integral type conversion fails.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct TryFromIntError(pub(crate) ());
```
daxpedda, line 2798:
```suggestion
pub struct TryFromIntError(());
```

daxpedda (approved):
No, I will go ahead and fix the CI first.

Ekleog:
Awesome, thank you! I just handled your review comment and merged master in, hopefully this is now good to go :)

daxpedda:
@Liamolucko would appreciate your approval here as well.

crates/js-sys/src/lib.rs
```diff
@@ -2793,6 +2793,37 @@ macro_rules! number_from {
...
+        fmt.write_str("out of range integral type conversion attempted")
+    }
+}
+
+impl std::error::Error for TryFromIntError {}
+
+macro_rules! number_try_from {
+    ($($x:ident)*) => ($(
+        impl TryFrom<$x> for Number {
+            type Error = $x;
```
Liamolucko, line 2811: You never actually updated this to use `TryFromIntError`.

Ekleog: Oh. Now I feel stupid ._. This is now fixed, thank you!

PR #3848, opened 2024-02-20
`wasm-bindgen-futures`: Use single-threaded implementation when multi-threading is impossible

daxpedda:
I encountered this when using the atomics target feature while running in a shared worker in Firefox, where cross-origin isolation is impossible and therefor sharing a `SharedArrayBuffer` with a dedicated worker is not available, which was the workaround used for Firefox not implementing `Atomics.waitAsync()`.

The solution is to use the single-threaded implementation of `wasm-bindgen-futures` when we detect that multi-threading is impossible.

Note: I somehow managed to accidentally push this commit onto the main branch ... which I promptly reverted and added a branch protection so this doesn't happen again.

### DIFF
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -50,18 +50,40 @@ pub use js_sys;
 pub use wasm_bindgen;
 
 mod task {
+    use std::future::Future;
+    use std::pin::Pin;
+
     use cfg_if::cfg_if;
 
-    cfg_if! {
-        if #[cfg(target_feature = "atomics")] {
-            mod wait_async_polyfill;
-            mod multithread;
-            pub(crate) use multithread::*;
+    #[cfg(target_feature = "atomics")]
+    mod multithread;
+    mod singlethread;
+    #[cfg(target_feature = "atomics")]
+    mod wait_async_polyfill;
 
-        } else {
-            mod singlethread;
-            pub(crate) use singlethread::*;
-         }
+    pub(crate) fn spawn(future: Pin<Box<dyn Future<Output = ()> + 'static>>) {
+        cfg_if! {
+            if #[cfg(target_feature = "atomics")] {
+                #[wasm_bindgen::prelude::wasm_bindgen]
+                extern "C" {
+                    /// Returns [`crossOriginIsolated`](https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated) global property.
+                    #[wasm_bindgen(js_name = crossOriginIsolated)]
+                    static CROSS_ORIGIN_ISOLATED: bool;
+                }
+
+                if *CROSS_ORIGIN_ISOLATED {
+                    multithread::Task::spawn(future)
+                } else {
+                    singlethread::Task::spawn(future)
+                }
+            } else {
+                singlethread::Task::spawn(future)
+            }
+        }
+    }
+
+    pub(crate) trait Task {
+        fn run(&self);
     }
 }
 
@@ -81,7 +103,7 @@ pub fn spawn_local<F>(future: F)
 where
     F: Future<Output = ()> + 'static,
 {
-    task::Task::spawn(Box::pin(future));
+    task::spawn(Box::pin(future));
 }
 
 struct Inner {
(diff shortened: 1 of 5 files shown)


daxpedda:
> This causes a problem because the single-threaded runtime always runs tasks on the current thread, meaning that if the task started on another 'thread' it's now technically been sent to a different thread despite not necessarily being `Send`, which is unsound.

How could you sent it to another thread though? I mean its true that in this scenario its using the same `SharedArrayBuffer`, but how is access from another achieved in such a case?

> At first I thought this wasn't an issue in practice because the wrapper JS only lets you instantiate the wasm module once, except that you can create more than one copy of the JS wrapper using e.g. the `?i=1` trick and then instantiate both of them with the same `SharedArrayBuffer`.

That sounds terribly unsound to me in the first place ... because it means that the initialization of the module is run twice on the same piece of memory?

Liamolucko:
> > This causes a problem because the single-threaded runtime always runs tasks on the current thread, meaning that if the task started on another 'thread' it's now technically been sent to a different thread despite not necessarily being `Send`, which is unsound.
> 
> How could you send it to another" thread" though? I mean its true that in this scenario its using the same `SharedArrayBuffer`, but how is access from another "thread" achieved in such a case?

By sending the task's `Waker` to another thread the same way you would normally, then calling `Waker::wake`. When using the single-threaded implementation, `wake` always polls the task on the thread where it was called, so calling `wake` on a task from another thread accidentally moves it to that thread. For example:

```rust
use std::sync::Mutex;
use std::task::Waker;
use std::{future, thread};

use wasm_bindgen::prelude::*;

static WAKER: Mutex<Option<Waker>> = Mutex::new(None);

#[wasm_bindgen]
pub fn spawn_task() {
    wasm_bindgen_futures::spawn_local(future::poll_fn(|cx| {
        web_sys::console::log_1(&JsValue::from_str(&format!("{:?}", thread::current())));
        *WAKER.lock().unwrap() = Some(cx.waker().clone());
        std::task::Poll::Pending
    }))
}

#[wasm_bindgen]
pub fn wake_task() {
    if let Some(waker) = WAKER.lock().unwrap().take() {
        waker.wake();
    }
}
```

```js
import init1, * as mod1 from "./pkg/wasm_bindgen_playground.js?i=1";
import init2, * as mod2 from "./pkg/wasm_bindgen_playground.js?i=2";

// this should probably be using `wasm_bindgen::memory()` but it doesn't really matter here
const memory = new WebAssembly.Memory({ initial: 18, maximum: 16384, shared: true });
await init1("./pkg/wasm_bindgen_playground_bg.wasm", memory);
await init2("./pkg/wasm_bindgen_playground_bg.wasm", memory);

mod1.spawn_task();
// let the task get polled so that waking it doesn't do nothing
await Promise.resolve();
mod2.wake_task();
```

Without this PR, this prints:

```
Thread { id: ThreadId(1), name: None, .. }
Thread { id: ThreadId(1), name: None, .. }
```

polling the task on the same thread both times like it should.

With this PR, it prints:

```
Thread { id: ThreadId(1), name: None, .. }
Thread { id: ThreadId(2), name: None, .. }
```

polling it on a different thread the second time.

crates/futures/src/lib.rs
```diff
@@ -50,18 +50,40 @@ pub use js_sys;
...
+    mod wait_async_polyfill;
+
+    pub(crate) fn spawn(future: Pin<Box<dyn Future<Output = ()> + 'static>>) {
+        cfg_if! {
+            if #[cfg(target_feature = "atomics")] {
+                #[wasm_bindgen::prelude::wasm_bindgen]
+                extern "C" {
+                    /// Returns [`crossOriginIsolated`](https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated) global property.
+                    #[wasm_bindgen(js_name = crossOriginIsolated)]
+                    static CROSS_ORIGIN_ISOLATED: bool;
```
Liamolucko, line 71: This only exists inside browsers, so it'll break if you try to use it inside Node or Deno.

You should be able to check if it exists at runtime, and then if it doesn't we can conservatively assume that multithreading is possible.

daxpedda:
Thanks @Liamolucko, I didn't understand exactly what you meant with its "moved to the other thread", but after trying your example and studying the `wasm-bindgen-futures` implementation I understood the issue.

Unfortunately I didn't not find a way to solve it. Initially I thought I could make a `Promise` and call `resolveFunc` when waking it, which would be similar to the multi-threaded example. But the issue is that I can't actually store and retrieve `resolveFunc` (a `js_sys::Function`, because well, its another thread.

Let me know if you have another solution in mind!

PR #3852, opened 2024-02-23
Add support for `Option<*const T>`, `Option<*mut T>` and `NonNull<T>`

daxpedda:
I specifically didn't implement taking `NonNull<T>` as a parameter to avoid having to add a runtime check somewhere.

~~EDIT: Not sure whats going wrong here, but will leave it at draft until I figure it out.~~

### DIFF
--- a/examples/guide-supported-types-examples/non_null.js
+++ b/examples/guide-supported-types-examples/non_null.js
@@ -1,12 +1,12 @@
 import {
-    take_pointer_by_value,
-    return_pointer,
-  } from './guide_supported_types_examples';
-  import { memory } from './guide_supported_types_examples_bg';
-  
-  let ptr = return_pointer();
-  let buf = new Uint8Array(memory.buffer);
-  let value = buf[ptr];
-  console.log(`The byte at the ${ptr} address is ${value}`);
-  
-  take_pointer_by_value(ptr);
+  take_pointer_by_value,
+  return_pointer,
+} from './guide_supported_types_examples';
+import { memory } from './guide_supported_types_examples_bg';
+
+let ptr = return_pointer();
+let buf = new Uint8Array(memory.buffer);
+let value = buf[ptr];
+console.log(`The byte at the ${ptr} address is ${value}`);
+
+take_pointer_by_value(ptr);
--- a/guide/src/SUMMARY.md
+++ b/guide/src/SUMMARY.md
@@ -53,7 +53,7 @@
     - [Exported Rust Types](./reference/types/exported-rust-types.md)
     - [`JsValue`](./reference/types/jsvalue.md)
     - [`Box<[T]>` and `Vec<T>`](./reference/types/boxed-slices.md)
-    - [`*const T`, `*mut T`](./reference/types/pointers.md)
+    - [`*const T` and `*mut T`](./reference/types/pointers.md)
     - [`NonNull<T>`](./reference/types/non-null.md)
     - [Numbers](./reference/types/numbers.md)
     - [`bool`](./reference/types/bool.md)


src/convert/impls.rs
```diff
@@ -223,6 +224,20 @@ impl<T> FromWasmAbi for *const T {
...
+        0
+    }
+}
+
+impl<T> OptionFromWasmAbi for *const T {
+    #[inline]
+    fn is_none(js: &u32) -> bool {
+        *js == 0
+    }
+}
```
Liamolucko, line 239: This doesn't work: since `*const T` describes itself as a `u32`, as far as the CLI's concerned an `Option<*const T>` is the same as an `Option<u32>`, and `Option<u32>` doesn't use this representation. It directly implements `IntoWasmAbi`, without using `OptionIntoWasmAbi`, and works by adding an extra `u32` that represents 'is this some?'.

If you wanted to, you could add a new descriptor for pointers that does work this way, but I think doing it the same way as `Option<u32>` is better anyway since it's perfectly valid for a `*const T` to be null, and so that shouldn't be treated as some sentinel value.

On the other hand, null isn't a valid value for `NonNull<T>`, and so doing it this way for that would be fine (but it'd need a new descriptor).

examples/guide-supported-types-examples/src/non_null.rs
```diff
@@ -0,0 +1,13 @@
...
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub unsafe fn take_pointer_by_value(x: Option<NonNull<u8>>) {
+    Box::from_raw(x.unwrap().as_ptr());
+}
+
+#[wasm_bindgen]
+pub fn return_pointer() -> Option<NonNull<u8>> {
+    NonNull::from(Box::new(42).leak())
```
Liamolucko, line 12: I don't think this is valid: `Box::leak` can only be called as `Box::leak(x)`, not `x.leak()`, and this needs to either be wrapped in `Some` or not return an `Option`.

Also using `Box::leak` + `Box::from_raw` is a bit questionable, this should probably use `Box::into_raw` instead.

guide/src/reference/types/non-null.md
```diff
@@ -0,0 +1,17 @@
+# `*const T` and `*mut T`
```
Liamolucko, line 1:
```suggestion
# `NonNull<T>`
```

tests/wasm/simple.rs
```diff
@@ -62,6 +64,20 @@ pub unsafe fn simple_raw_pointers_work(a: *mut u32, b: *const u8) -> *const u32
...
+pub unsafe fn simple_option_raw_pointers_work(
+    a: Option<*mut u32>,
+    b: Option<*const u8>,
+) -> Option<*const u32> {
+    *a.unwrap() = *b.unwrap() as u32;
+    a.map(|ptr| ptr as *const _)
+}
+
+#[wasm_bindgen]
+pub unsafe fn simple_option_nonnull_work(a: Option<NonNull<u32>>) -> Option<NonNull<u32>> {
```
Liamolucko, line 77:
```suggestion
pub fn simple_option_nonnull_work(a: Option<NonNull<u32>>) -> Option<NonNull<u32>> {
```

Manipulating raw pointers isn't actually unsafe, just dereferencing them.

examples/guide-supported-types-examples/src/non_null.rs
daxpedda: Using [`Box::leak()` with `Box::from_raw()` is supported according to the documentation](https://doc.rust-lang.org/1.76.0/std/boxed/struct.Box.html#method.leak).

The reason why this is used is to get access to the infallible conversion to `NonNull`, compared to `NonNull::new()`. But in this case we return an `Option` anyway.

crates/cli-support/src/wit/incoming.rs
```diff
@@ -155,6 +155,8 @@ impl InstructionBuilder<'_, '_> {
 
             // Largely synthetic and can't show up
             Descriptor::ClampedU8 => unreachable!(),
+
+            Descriptor::NonNull => self.number(AdapterType::NonNull, WasmVT::I32),
```
Liamolucko, line 159: It might be a good idea to add some kind of comment mentioning that this is unused right now due to only `Option<NonNull<T>>` implementing `FromWasmAbi`.

crates/cli-support/src/wit/standard.rs
```diff
@@ -308,6 +309,8 @@ pub enum Instruction {
     OptionEnumFromI32 {
         hole: u32,
     },
+    InNonNullSentinel,
+    OutNonNullSentinel,
```
Liamolucko, line 313: I'm not particularly a fan of these names; all the other `Instruction`s tend to use `From` and `To`, not `In` and `Out`, so they don't match very well. `OptionNonNullFromI32` and `I32FromOptionNonNull` would be more in line with everything else.

tests/wasm/simple.rs
```diff
@@ -62,6 +66,35 @@ pub unsafe fn simple_raw_pointers_work(a: *mut u32, b: *const u8) -> *const u32
...
 }
 
+#[wasm_bindgen]
+pub unsafe fn simple_option_raw_pointers_work(
+    a: Option<*mut u32>,
+    b: Option<*const u8>,
+) -> Option<*const u32> {
+    *a.unwrap() = *b.unwrap() as u32;
+    a.map(|ptr| ptr as *const _)
+}
```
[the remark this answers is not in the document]
daxpedda: Yeah, this was just checking if it compiles.

In any case, I hooked this and the old raw pointer tests up as well now.

crates/cli-support/src/js/binding.rs
```diff
@@ -1217,6 +1217,18 @@ fn instruction(
...
+        Instruction::InNonNullSentinel => {
+            let val = js.pop();
+            js.cx.expose_is_like_none();
+            js.assert_optional_number(&val);
+            js.push(format!("isLikeNone({0}) ? 0 : {0}", val));
+        }
+
+        Instruction::OutNonNullSentinel => {
+            let val = js.pop();
+            js.push(format!("{0} === 0 ? undefined : {0}", val));
```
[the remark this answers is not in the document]
daxpedda: I resolved it inside the `Instruction::WasmToInt` directly instead.

crates/cli-support/src/wit/incoming.rs
daxpedda: I removed it and added an `unimplemented!()` instead.

crates/cli-support/src/js/binding.rs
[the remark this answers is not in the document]
Liamolucko: `WasmToInt` isn't used in the optional case though, so this is still needed.

crates/cli-support/src/js/binding.rs
```diff
@@ -1217,6 +1217,18 @@ fn instruction(
...
         }
+
+        Instruction::OptionNonNullFromI32 => {
+            let val = js.pop();
+            js.cx.expose_is_like_none();
+            js.assert_optional_number(&val);
+            js.push(format!("isLikeNone({0}) ? 0 : {0}", val));
+        }
+
+        Instruction::I32FromOptionNonNull => {
```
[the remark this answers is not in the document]
daxpedda: That's my bad!
I did double-check it but still mixed it up :sweat_smile: !

examples/guide-supported-types-examples/non_null.js
Liamolucko, line 1: This file's indentation is a bit messed up.

guide/src/SUMMARY.md
```diff
@@ -53,7 +53,8 @@
     - [Exported Rust Types](./reference/types/exported-rust-types.md)
     - [`JsValue`](./reference/types/jsvalue.md)
     - [`Box<[T]>` and `Vec<T>`](./reference/types/boxed-slices.md)
-    - [`*const T` and `*mut T`](./reference/types/pointers.md)
+    - [`*const T`, `*mut T`](./reference/types/pointers.md)
```
Liamolucko, line 56: Why the change from 'and' to ','?

daxpedda: Thank you for the very careful review!

I previously had this and `NonNull` in the same file, so it was "`*const T`, `*mut T` and `NonNull`", I then split it into separate files and remove "and `NonNull`".

Liamolucko (approved):
LGTM!

PR #3855, opened 2024-02-25
improvement on `IntoWasmAbi` and `FromWasmAbi` docs

alivarastepour:


daxpedda (commented):
The changes make it sound like the ABI is an actor doing something, compared to the relatively passive speech the original documentation used.

I'm not a native English speaker, but I find this way harder to read.

PR #3856, opened 2024-02-25
Allow using `'static` lifetimes in functions

### LINKED ISSUE #1187: Not sound to use lifetimes in function signatures
Seems like wasm-bindgen prohibits the use of lifetime specifiers in certain places where the compiler requires them. I have a test repo here: https://github.com/attente/wasm-bindgen-lifetime-issue

Building it, the error is:

```
[nix-shell:~/wasm-bindgen-lifetime-issue]$ cargo build --release --target=wasm32-unknown-unknown
   Compiling wasm-bindgen-lifetime-issue v0.1.0 (/home/william/wasm-bindgen-lifetime-issue)
error: it is currently not sound to use lifetimes in function signatures
  --> src/lib.rs:10:31
   |
10 |     pub fn f() -> Result<(), &'static str> {
   |                               ^^^^^^^

error: Could not compile `wasm-bindgen-lifetime-issue`.

To learn more, run the command again with --verbose.
```

Unfortunately if I remove the `'static` lifetime specifier, then the compiler will complain:

```
[nix-shell:~/wasm-bindgen-lifetime-issue]$ cargo build --release --target=wasm32-unknown-unknown
   Compiling wasm-bindgen-lifetime-issue v0.1.0 (/home/william/wasm-bindgen-lifetime-issue)
error[E0106]: missing lifetime specifier
  --> src/lib.rs:10:30
   |
10 |     pub fn f() -> Result<(), &str> {
   |                              ^ help: consider giving it a 'static lifetime: `&'static`
   |
   = help: this function's return type contains a borrowed value, but there is no value for it to be borrowed from

For more information about this error, try `rustc --explain E0106`.
error: Could not compile `wasm-bindgen-lifetime-issue`.

To learn more, run the command again with --verbose.
```

Is there any workaround for this issue?

daxpedda:
Currently `#[wasm_bindgen]` disallows any kind of lifetime on functions.
This change allows it to accept `'static`.

Fixes #1187.

### DIFF
--- a/crates/macro/ui-tests/invalid-imports.stderr
+++ b/crates/macro/ui-tests/invalid-imports.stderr
@@ -1,9 +1,3 @@
-error: it is currently not sound to use lifetimes in function signatures
- --> ui-tests/invalid-imports.rs:7:16
-  |
-7 |     fn f() -> &'static u32;
-  |                ^^^^^^^
-
 error: imported methods must have at least one argument
   --> ui-tests/invalid-imports.rs:10:5
    |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 
 ### Changed
 
-* Allow using `'static` lifetimes in functions marked with `#[wasm_bindgen]`.
+* Allow using `'static` lifetimes in functions marked with `#[wasm_bindgen]`. This does not allow references where they were not allowed before!
   [#3856](https://github.com/rustwasm/wasm-bindgen/pull/3856)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@
 * Add support for `Option<*const T>`, `Option<*mut T>` and `NonNull<T>`.
   [#3852](https://github.com/rustwasm/wasm-bindgen/pull/3852)
 
+### Changed
+
+* Allow using `'static` lifetimes in functions marked with `#[wasm_bindgen]`.
+  [#3856](https://github.com/rustwasm/wasm-bindgen/pull/3856)
+
 ### Fixed
 
 * Make .wasm output deterministic when using `--reference-types`.
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -1652,12 +1652,14 @@ fn assert_no_lifetimes(sig: &syn::Signature) -> Result<(), Diagnostic> {
     }
 
     impl<'ast> syn::visit::Visit<'ast> for Walk {
-        fn visit_lifetime(&mut self, i: &'ast syn::Lifetime) {
-            self.diagnostics.push(err_span!(
-                i,
-                "it is currently not sound to use lifetimes in function \
+        fn visit_lifetime(&mut self, lifetime: &'ast syn::Lifetime) {
+            if lifetime.ident != "static" {
+                self.diagnostics.push(err_span!(
+                    lifetime,
+                    "it is currently not sound to use lifetimes in function \
                  signatures"
-            ));
+                ));
+            }
         }
     }
     let mut walk = Walk {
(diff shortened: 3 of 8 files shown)


Liamolucko:
This applies to arguments too, where using `'static` isn't okay. It's still sound because the way references work internally is that `#[wasm_bindgen]` generates a local variable and takes a reference to it, so you just get a borrow checker error, but it's a worse error than you'd have gotten before:

```rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn invalid(_: &'static str) {}
```

```
error[E0597]: `*arg0` does not live long enough
 --> src/lib.rs:3:1
  |
3 | #[wasm_bindgen]
  | ^^^^^^^^^^^^^^-
  | |             |
  | |             `*arg0` dropped here while still borrowed
  | argument requires that `*arg0` is borrowed for `'static`
  |

For more information about this error, try `rustc --explain E0597`.
error: could not compile `wasm-bindgen-playground` (lib) due to 1 previous error
```

LGTM if you're okay with that though.

crates/macro/ui-tests/invalid-imports.stderr
```diff
@@ -1,9 +1,3 @@
-error: it is currently not sound to use lifetimes in function signatures
- --> ui-tests/invalid-imports.rs:7:16
-  |
-7 |     fn f() -> &'static u32;
-  |                ^^^^^^^
```
Liamolucko, line 5: It's a little annoying that this seems to be getting suppressed by the other errors, so the new error ('cannot return references in #[wasm_bindgen] imports yet') isn't being tested.

Liamolucko:
Also, this fixes #1187 (the thing it was originally opened for anyway, not the struct use case someone brought up later).

crates/macro/ui-tests/invalid-imports.stderr
daxpedda: Separated the tests.

daxpedda:
> LGTM if you're okay with that though.

I'm fine with that considering its the same error message you already get when doing:
```rust
#[wasm_bindgen]
pub fn invalid(_: &str) {}
``` 

What kind of references are actually allowed in exported functions? Otherwise I could go ahead and outright ban them.

Liamolucko:
> I'm fine with that considering its the same error message you already get when doing:
> 
> ```rust
> #[wasm_bindgen]
> pub fn invalid(_: &str) {}
> ```

What do you mean? That example compiles just fine.

> What kind of references are actually allowed in exported functions? Otherwise I could go ahead and outright ban them.

Up until now, just the kind above, and I assumed the whole point of this PR was to allow ones like this:

```rs
#[wasm_bindgen]
pub fn foo() -> &'static str {
    "hello world"
}
```

I didn't test it thoroughly enough and that actually still doesn't compile because of an extra error ('cannot return a borrowed ref with #[wasm_bindgen]'), but it now works if you wrap it in something:

```rs
#[wasm_bindgen]
pub fn foo() -> Option<&'static str> {
    Some("hello world")
}
```

That's also the use-case #1187 was asking for (although it should probably be reopened now that I've realised this doesn't fully fix it).

daxpedda:
> > I'm fine with that considering its the same error message you already get when doing:
> > ```rust
> > #[wasm_bindgen]
> > pub fn invalid(_: &str) {}
> > ```
> 
> What do you mean? That example compiles just fine.

Yep, my mistake.

> Up until now, just the kind above, and I assumed the whole point of this PR was to allow ones like this:
> 
> ```rust
> #[wasm_bindgen]
> pub fn foo() -> &'static str {
>     "hello world"
> }
> ```

Sorry, should have probably mentioned this in OP. My problem was just lifetimes in pointers as parameters, e.g. `par1: *const FnOnce(&'static u32)`. I was originally considering removing the whole lifetime check, but I wasn't really sure what other issues this was addressing, so just enabling `'static` seemed safe to me.

> That's also the use-case #1187 was asking for (although it should probably be reopened now that I've realised this doesn't fully fix it).

I've re-opened it.

---

My PR's have been pretty sloppy lately, so thank you for your diligent and careful reviews!

PR #3314, opened 2023-02-16
Add `WASM_BINDGEN_TEST_ADDRESS`

daxpedda:
This introduces a new environment variable for `wasm-bindgen-test`: `WASM_BINDGEN_TEST_ADDRESS`, which allows to customize the address and port used when using it in non-headless mode.

I needed this to test stuff on a different computer without having to setup some sort of port sharing.

PR #3861, opened 2024-02-29
Allow overriding headless test URL

lynn:
I have wasm code that uses web-sys to make cross-origin requests with cookies. To test this interactively, I navigate to `http://localhost.mydomain.org:8000` (which I've set up to resolve to 127.0.0.1) instead of `http://localhost:8000`.

This PR would allow me to do the same for headless tests. I imagine it's also useful in combination with #3314.

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
@@ -121,8 +121,8 @@ pub fn run(server: &SocketAddr, shell: &Shell, timeout: u64) -> Result<(), Error
 
     // Visit our local server to open up the page that runs tests, and then get
     // some handles to objects on the page which we'll be scraping output from.
-    let url = std::env::var("WASM_BINDGEN_TEST_ADDRESS")
-        .unwrap_or_else(|_| format!("http://{}", server));
+    let url =
+        std::env::var("WASM_BINDGEN_TEST_ADDRESS").unwrap_or_else(|_| format!("http://{}", server));
     shell.status(&format!("Visiting {}...", url));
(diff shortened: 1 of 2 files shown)


daxpedda (changes requested):
I don't think we need a new environment variable here, lets reuse `WASM_BINDGEN_TEST_ADDRESS`.

Please add an entry in the changelog as well.

lynn:
It feels a bit wrong to me to reuse that variable, because WASM_BINDGEN_TEST_ADDRESS is necessarily an address specifying where to _bind_ the test server, and this variable is a URL specifying _how_ to reach it. In Rust terms, one is a `SocketAddr` and the other is a `Url`. But I'll make the requested changes.

daxpedda (approved):
Thank you!

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -121,7 +121,8 @@ pub fn run(server: &SocketAddr, shell: &Shell, timeout: u64) -> Result<(), Error
 
     // Visit our local server to open up the page that runs tests, and then get
     // some handles to objects on the page which we'll be scraping output from.
-    let url = format!("http://{}", server);
+    let url =
+        std::env::var("WASM_BINDGEN_TEST_ADDRESS").unwrap_or_else(|_| format!("http://{}", server));
```
daxpedda, line 125:
```suggestion
    let url = format!("http://{}", std::env::var("WASM_BINDGEN_TEST_ADDRESS").unwrap_or(server));
```
I was briefly thinking that this might be a good idea, but maybe there is a use case out there for `https`?

PR #3865, opened 2024-03-02
Revert "Allow using `'static` lifetimes in functions (#3856)"

daxpedda:
This reverts #3856. After some careful thinking about this PR, I don't think it achieves a whole lot.
While it actually enables one more valid signature (`fn foo() -> Option<&'static str>`), it otherwise achieves little.

Eventually the whole lifetime check should be removed, because we already do "lifetime checks" in the generated code, the check just improves the error message, but prevents some specific uses. So all in all I don't think its worth the tradeoff.


Liamolucko (approved):
I was actually thinking of finishing the job and disabling the 'cannot return references' error for exported functions so that plain `&'static T` would work; but since you mentioned in #3851 you want to make a new release soon I think it's better to revert this now and fix it properly later than release things in an incomplete state.

andreamancuso:
Hi @daxpedda and @Liamolucko 

Is there an open issue I can track for these changes?

Many thanks!

daxpedda:
Cc https://github.com/rustwasm/wasm-bindgen/issues/1187.

PR #3866, opened 2024-03-02
Don't allow invalid Unicode scalar values in `char`

### LINKED ISSUE #2269: UB with `impl FromWasmAbi for char` when unpaired surrogate is given
### Describe the Bug

Wasm-bindgen assumes the "char" (which is a JavaScript `String`) to be converted is always valid and uses `char::from_u32_unchecked`. However the JS `String` may contain a single "char" of an unpaired surrogate (for example when the JS code uses `charAt`), in which using `char::from_u32_unchecked` with it creates an invalid `char` and causes UB.

### Steps to Reproduce

```rust
// Rust:
#[wasm_bindgen]
pub fn take_char_by_value(x: char) {}
```

```js
// JS:
take_char_by_value('😃'.charAt(0));
```

### Expected Behavior

I don't have an idea what behaviour would be the best. Probably use U+FFFD like how strings are handled? Perhaps if the type is `Option<char>` it should use `char::from_u32` instead, but then this might cause ambiguity between `null`-ish and invalid char.

### Actual Behavior

An invalid `char` is produced.

daxpedda:
This adds runtime checks for passing `char` from JS, which prevents invalid Unicode scalar values.

Fixes #2269.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2116,6 +2116,19 @@ impl<'a> Context<'a> {
         );
     }
 
+    fn expose_assert_char(&mut self) {
+        if !self.should_write_global("assert_char") {
+            return;
+        }
+        self.global(
+            "
+            function _assertChar(c) {
+                if (typeof(c) !== 'number' || c >= 0x110000 || (c >= 0xD800 && c < 0xE000)) throw new Error(`expected a number argument that is a valid Unicode scalar value, found ${c}`);
+            }
+            ",
+        );
+    }
+
     fn expose_make_mut_closure(&mut self) -> Result<(), Error> {
         if !self.should_write_global("make_mut_closure") {
             return Ok(());
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2123,7 +2123,7 @@ impl<'a> Context<'a> {
         self.global(
             "
             function _assertChar(c) {
-                if (typeof(c) !== 'number' || c >= 0x110000 || (c >= 0xD800 && c < 0xE000)) throw new Error(`expected a number argument that is a valid Unicode scalar value, found ${c}`);
+                if (typeof(c) === 'number' && (c >= 0x110000 || (c >= 0xD800 && c < 0xE000))) throw new Error(`expected a valid Unicode scalar value, found ${c}`);
             }
             ",
         );
--- a/tests/wasm/char.rs
+++ b/tests/wasm/char.rs
@@ -4,6 +4,7 @@ use wasm_bindgen_test::*;
 #[wasm_bindgen(module = "tests/wasm/char.js")]
 extern "C" {
     fn js_identity(c: char) -> char;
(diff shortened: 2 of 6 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -2116,6 +2116,19 @@ impl<'a> Context<'a> {
...
     }
 
+    fn expose_assert_char(&mut self) {
+        if !self.should_write_global("assert_char") {
+            return;
+        }
+        self.global(
+            "
+            function _assertChar(c) {
+                if (typeof(c) !== 'number' || c >= 0x110000 || (c >= 0xD800 && c < 0xE000)) throw new Error(`expected a number argument that is a valid Unicode scalar value, found ${c}`);
```
Liamolucko, line 2126: 'expected a number argument' isn't a great error message here because `c` isn't actually something we received from user code, it's the result of `codePointAt(0)`. Maybe just more explicitly say 'unpaired surrogate found'?

That isn't the right message for the case where it isn't a number though, which can happen if the string is empty. In that case `codePointAt` returns undefined, which gets coerced to `'\0'`. (That means it is technically a change in non-unsound behaviour, but I don't think it really matters.)

crates/cli-support/src/js/mod.rs
```diff
@@ -2116,6 +2116,19 @@ impl<'a> Context<'a> {
...
     }
 
+    fn expose_assert_char(&mut self) {
+        if !self.should_write_global("assert_char") {
+            return;
+        }
+        self.global(
+            "
+            function _assertChar(c) {
+                if (typeof(c) === 'number' && (c >= 0x110000 || (c >= 0xD800 && c < 0xE000))) throw new Error(`expected a valid Unicode scalar value, found ${c}`);
```
daxpedda, line 2126: I believe this should work, I also added a couple of tests for this as well.

crates/cli-support/src/js/mod.rs
daxpedda: > That isn't the right message for the case where it isn't a number though, which can happen if the string is empty. In that case `codePointAt` returns undefined, which gets coerced to `'\0'`. (That means it is technically a change in non-unsound behaviour, but I don't think it really matters.)

I addressed this in https://github.com/rustwasm/wasm-bindgen/pull/3866/files#r1509971507, there should be no change in behavior now.

tests/wasm/char.rs
```diff
@@ -12,11 +13,21 @@ pub fn rust_identity(c: char) -> char {
...
+
 #[wasm_bindgen]
 pub fn rust_js_identity(c: char) -> char {
     js_identity(c)
 }
 
+#[wasm_bindgen]
+pub fn rust_js_option_identity(c: Option<char>) -> Option<char> {
+    js_option_identity(c)
+}
```
Liamolucko, line 29: This doesn't seem to be used.

daxpedda:
CI is currently broken because of https://github.com/rust-lang/rust/issues/121932, which will be fixed tomorrow (hopefully) by https://github.com/rust-lang/rust/pull/121933.

Liamolucko (approved):

PR #3870, opened 2024-03-04
Account for failure in publishing process

daxpedda:
We just had some funny failures in the CI during publishing: https://github.com/rustwasm/wasm-bindgen/actions/runs/8138814872/job/22240636231#step:13:1.

This PR changes the code to actually fail the process when we encounter any error during publishing.

### DIFF
--- a/publish.rs
+++ b/publish.rs
@@ -216,6 +216,6 @@ fn publish(krate: &Crate) {
         .status()
         .expect("failed to run cargo");
     if !status.success() {
-        println!("FAIL: failed to publish `{}`: {}", krate.name, status);
+        panic!("FAIL: failed to publish `{}`: {}", krate.name, status);
     }
 }


publish.rs
```diff
@@ -216,6 +216,6 @@ fn publish(krate: &Crate) {
         .status()
         .expect("failed to run cargo");
     if !status.success() {
-        println!("FAIL: failed to publish `{}`: {}", krate.name, status);
+        panic!("FAIL: failed to publish `{}`: {}", krate.name, status);
```
daxpedda, line 219: This would probably be better by return a result all the way up to `main()`, but this should work fine for our purposes.

torokati44:
It's strange that while I see errors during publishing `wasm-bindgen-shared` and ` wasm-bindgen-backend` too, and yet, they _did_ get end up being updated on `crates.io`. :thinking:

daxpedda:
> It's strange that while I see errors during publishing `wasm-bindgen-shared` and ` wasm-bindgen-backend` too, and yet, they _did_ get end up being updated on `crates.io`. 🤔

Its because I've rerun the CI five times by now :sweat_smile:.
`wasm-bindgen` is the last crate that just refuses to be published, so I'm waiting on a request for crates.io permissions to figure it out locally.

torokati44:
Surely it will work for the 6th time... 😈😜

Liamolucko (approved):

daxpedda:
Here is a newer version of `publish.rs`: https://github.com/bytecodealliance/wasm-tools/blob/a178c9bea591bbcd7608cf0de6f9662eb50f4e40/ci/publish.rs.

Unfortunately it isn't really compatible with our workspace, but would be nice if one day we can update ours as well. The important feature would be to not try and publish crates that were already published, to avoid the rate-limiting issue.

PR #3873, opened 2024-03-05
Copy port from headless test server when using WASM_BINDGEN_TEST_ADDRESS

lynn:
When putting #3861 to practice, I realized that it's a bit hard to use, because the headless test server might open on a different port every time. It's better to copy the port from the server when it isn't specified. So for example when it happens to open on `127.0.0.1:23456` and WASM_BINDGEN_TEST_ADDRESS is set to `http://localhost.mydomain.org` we should connect to `http://localhost.mydomain.org:23456` in the test runner (not the default port 80).


crates/cli-support/src/js/mod.rs
```diff
@@ -1768,6 +1768,7 @@ impl<'a> Context<'a> {
         self.memview("Float64", memory)
     }
 
+    #[allow(dead_code)]
```
lynn, line 1771: I had to add some `#[allow(dead_code)]` annotations to compile without warnings. `memview_function` does seem genuinely unused.

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -121,8 +121,14 @@ pub fn run(server: &SocketAddr, shell: &Shell, timeout: u64) -> Result<(), Error
 
     // Visit our local server to open up the page that runs tests, and then get
     // some handles to objects on the page which we'll be scraping output from.
-    let url =
-        std::env::var("WASM_BINDGEN_TEST_ADDRESS").unwrap_or_else(|_| format!("http://{}", server));
+    let url = match std::env::var("WASM_BINDGEN_TEST_ADDRESS") {
+        Ok(u) => {
+            let mut url = Url::parse(&u)?;
+            let _ = url.set_port(url.port().or(Some(server.port())));
```
daxpedda, line 127: Would prefer a `if url.port.is_none()` here instead.

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -121,8 +121,14 @@ pub fn run(server: &SocketAddr, shell: &Shell, timeout: u64) -> Result<(), Error
...
+            url.as_str().to_owned()
```
daxpedda, line 128:
```suggestion
            url.to_string()
```
This should work I believe.

crates/cli-support/src/js/mod.rs
daxpedda: Would you mind going ahead and just deleting it?

daxpedda (changes requested):

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -121,8 +121,24 @@ pub fn run(server: &SocketAddr, shell: &Shell, timeout: u64) -> Result<(), Error
...
-    let url =
-        std::env::var("WASM_BINDGEN_TEST_ADDRESS").unwrap_or_else(|_| format!("http://{}", server));
+    //
+    // If WASM_BINDGEN_TEST_ADDRESS is set, use it as the local server URL,
+    // trying to inherit the port from the server if it isn't specified.
+    let url = match std::env::var("WASM_BINDGEN_TEST_ADDRESS") {
+        Ok(u) => {
+            let mut url = Url::parse(&u)?;
+            if url.port().is_none() {
+                if url.set_port(Some(server.port())).is_err() {
```
daxpedda, line 131: I don't believe we have to account for that error, because in that case the URL wouldn't really be usable by us, so using `unwrap()` should be fine here.

https://docs.rs/url/2.5.0/url/struct.Url.html#method.set_port

Issue #3698, opened 2023-11-10
Use the newly fixed debugging information


daxpedda:
Just dropping some random thoughts before I forget them.

- Using it in `wasm-bindgen-test` instead of `console_error_panic_hook` (don't replace it entirely because we still need the panic message).
- Add a new documentation page noting how to actually deploy it.
  - Generate sourcemaps, e.g. using [`wasm2map`](https://crates.io/crates/wasm2map).
  - Strip DWARF in case of sourcemaps.
  - Notes on browser support and how to serve source files if not bundled.
- Maybe think about adding an option to generate source maps or split debug info directly in `wasm-bindgen`, so I would argue this would be more the domain of `wasm-pack` and similar.

hamza1311:
A page in the documentation about this information would be very helpful

xxshady:
for some reason wasm2map generates empty sourcemap file when i call it on .wasm generated by `wasm-pack`
`{"version":3,"names":[],"sources":[""],"sourcesContent":[],"mappings":""}`

commands i use:
`wasm-pack build --dev --target no-modules`
`cargo wasm2map --bundle-sources -p rust_wasm_bg.wasm -b http://127.0.0.1:8080`

it generates correct sourcemap with .wasm from target/wasm32-unknown-unknown/debug though

daxpedda:
You have to use the [`--keep-debug` CLI option](https://rustwasm.github.io/docs/wasm-bindgen/reference/cli.html#--keep-debug) to not discard the DWARF information when post-processing with `wasm-bindgen`.

How to do that with `wasm-pack` is something you might want to ask there.

xxshady:
yeah, thank you, with `--keep-debug` wasm2map generates correct sourcemap, i did not found how to enable it in wasm-pack, but using wasm-bindgen CLI directly it worked:
1. `cargo build --target wasm32-unknown-unknown`
2. `wasm-bindgen --target <your target> --out-dir pkg --keep-debug target/wasm32-unknown-unknown/debug/<crate name>.wasm`
(if you don't have wasm-bindgen CLI installed: `cargo install wasm-bindgen-cli`)
3. `cargo wasm2map pkg/<crate name>_bg.wasm`
(you probably also need to provide other flags if your target is browser or nodejs, see wasm2map [example](https://github.com/mtolmacs/wasm2map/blob/55978542eddc08cd43895999fb3ad5586020d837/example/run.bat#L7))
[...]

PR #3483, opened 2023-06-15
Bump `walrus` to v0.20

daxpedda:
Just checking what CI says for now.

Fixes #2389.

PR #3876, opened 2024-03-08
Deprecate `--keep-debug` flag

daxpedda:
Debug information is already controlled by Rustc/Cargo, but the reason this flag was introduced is to control debug information coming from the pre-compiled Std libraries, see https://github.com/rustwasm/wasm-bindgen/pull/466.

This can now be done by the user instead in Cargo with the [`strip`](https://doc.rust-lang.org/1.76.0/cargo/reference/profiles.html#strip) option, which is available since v1.59 (or directly with Rustc [`-C strip`](https://doc.rust-lang.org/1.76.0/rustc/codegen-options/index.html#strip) since v1.58).

Additionally, this will require no user action at all when v1.77 releases, which will strip Std's debug information automatically when no debug information is requested (e.g. the release profile), see https://github.com/rust-lang/cargo/pull/13257.

This should significantly improve the stacktraces with the downside of significantly increased Wasm file sizes in release mode for users who don't strip debug information or are using Cargo versions below v1.77.

We might want to hold off merging this until Rust v1.77 release, which would be 21th March, but considering how close this is (13 days) I think its alright.

See https://github.com/rustwasm/wasm-bindgen/pull/3483, which made this finally possible.
See https://github.com/rustwasm/wasm-bindgen/issues/3698 for tracking other debug information related improvements.

Blocked on https://github.com/rustwasm/walrus/pull/284 and a new Walrus release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,9 @@
 * Improved error messages for `self` arguments in invalid positions.
   [#4276](https://github.com/rustwasm/wasm-bindgen/pull/4276)
 
+* Deprecate `--keep-debug`. Debug information is now retained by default.
+  [#3876](https://github.com/rustwasm/wasm-bindgen/pull/3876)
+
 ### Fixed
 
 * Fixed methods with `self: &Self` consuming the object.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,6 +135,7 @@ resolver = "2"
 
 [patch.crates-io]
 js-sys = { path = 'crates/js-sys' }
+walrus = { git = "https://github.com/daxpedda/walrus", branch = "empty-dwarf" }
 wasm-bindgen = { path = '.' }
 wasm-bindgen-futures = { path = 'crates/futures' }
 web-sys = { path = 'crates/web-sys' }
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -30,7 +30,6 @@ pub struct Bindgen {
     omit_imports: bool,
     demangle: bool,
     keep_lld_exports: bool,
-    keep_debug: bool,
     remove_name_section: bool,
     remove_producers_section: bool,
     omit_default_module_path: bool,
@@ -101,7 +100,6 @@ impl Bindgen {
             omit_imports: false,
             demangle: true,
             keep_lld_exports: false,
-            keep_debug: false,
             remove_name_section: false,
             remove_producers_section: false,
             emit_start: true,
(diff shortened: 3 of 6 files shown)


Liamolucko:
> Debug information is already controlled by Rustc/Cargo, but the reason this flag was introduced is to control debug information coming from the pre-compiled Std libraries, see #466.

Interesting, I didn't know that history.

> This should significantly improve the stacktraces with the downside of significantly increased Wasm file sizes in release mode for users who don't strip debug information or are using Cargo versions below v1.77.

I didn't think debug info affected stack traces. At least at a first glance, browsers only seem to use the `name` section for that, which is already included without `--keep-debug`.

It should still be useful for actually debugging, though.

> We might want to hold off merging this until Rust v1.77 release, which would be 21th March, but considering how close this is (13 days) I think its alright.

Yeah, this probably won't be getting included in a wasm-bindgen release until after that point anyway.

daxpedda:
> > This should significantly improve the stacktraces with the downside of significantly increased Wasm file sizes in release mode for users who don't strip debug information or are using Cargo versions below v1.77.
> 
> I didn't think debug info affected stack traces. At least at a first glance, browsers only seem to use the `name` section for that, which is already included without `--keep-debug`.

Strange, it does improve the stack trace for me significantly.

daxpedda:
So after some digging, I found that `wasm-opt` doesn't currently support support our DWARF output, for whatever reason. I opened https://github.com/WebAssembly/binaryen/issues/6391 for that.

The main worry here is that `wasm-pack` suddenly stops working. This shouldn't be a problem as soon as Rust v1.77 hits, as then debug information will be stripped by default anyway.

I will update the PR when that happens.

daxpedda:
> > > This should significantly improve the stacktraces with the downside of significantly increased Wasm file sizes in release mode for users who don't strip debug information or are using Cargo versions below v1.77.
> > 
> > 
> > I didn't think debug info affected stack traces. At least at a first glance, browsers only seem to use the `name` section for that, which is already included without `--keep-debug`.
> 
> Strange, it does improve the stack trace for me significantly.

You are right, I had the [C/C++ DevTools Support](https://chromewebstore.google.com/detail/cc++-devtools-support-dwa/pdcpmagijalfljmkmjngeonclgbbannb) extension enabled, which is why I was getting better stacktraces.

daxpedda:
Apparently `wasm-opt` still fails unless we use `-Cstrip=symbols`, I'm happy to still move forwards with this as it isn't really a `wasm-bindgen` issue.

WDYT @Liamolucko.

daxpedda:
I'm unsure why CI fails, everything seems to work locally for me ...

daxpedda:
The issue was actually quite simple: Walrus generated empty custom sections even if no debugging information was available. `wasm-opt` apparently can't handle that.

I made a fix in https://github.com/rustwasm/walrus/pull/284.

daxpedda:
I will go ahead and close this. The current plan is as follows:
- By default, we will emit a source map and the DWARF output will be removed from the Wasm module.
- Using `--keep-debug` will keep the DWARF output in the Wasm module but still emitting a source map.
- Maybe we will add a `--no-source-map` switch to disable the default source map output.

PR #3882, opened 2024-03-10
Stabilize Web Share API

### LINKED ISSUE #2514: Implement the canShare() and share() methods for Navigator
### Motivation

I would like to be able to invoke the native sharing interface when available in a project I'm working on, so that users can share the link to any platform they like, rather than only ones I've specifically added share links for.

### Proposed Solution

Implement [Navigator.canShare()](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/canShare) and [Navigator.share()](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share) in web_sys::Navigator.

### Alternatives

The only alternative I can think of would be interfacing with JavaScript directly in my code. The proposed solution is clearly superior because it makes things much more convenient for anyone who wants to use the share API.

Does anyone know how to do this?

burrscurr:
[Web Share API is a W3C Recommendation](https://www.w3.org/TR/web-share/) since 30 May 2023. The webidl file was already up to date, meaning changes are mainly cosmetic (doc strings/drop of `web_sys_unstable_apis` feature).

Fixes #2514.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 * Implement `From<NonNull<T>>` for `JsValue`.
   [#3877](https://github.com/rustwasm/wasm-bindgen/pull/3877)
 
+### Changed
+
+* Stabilize Web Share API.
+  [#3882](https://github.com/rustwasm/wasm-bindgen/pull/3882)
+
 ### Fixed
 
 * Copy port from headless test server when using `WASM_BINDGEN_TEST_ADDRESS`.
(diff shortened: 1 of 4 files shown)


burrscurr:
This should also close #2514.

daxpedda (changes requested):
Missing an entry in the changelog.
Otherwise LGTM!

daxpedda (approved):
Thank you!

PR #3887, opened 2024-03-12
Simplify audio worklet sample

rom1v:
Import the .js directly without a complex `dependent_module!` macro.

---

The macro `dependent_module!()` returns an url like `blob:http://localhost:8080/43856cc8-5014-409b-97bd-978a7a06ba3e`.

When trying to understand how this sample worked, I found that we can directly import the `wasm_audio_worklet.js` in the worklet Javascript file instead. I'm not sure if it does all the same, but it works in Firefox and Chromium, and we can get rid of the whole `dependent_macro.rs` mechanism.

What do you think?

_(I guess you won't like that the worklet imports the hardcoded path `./pkg/wasm_audio_worklet.js`?)_


daxpedda:
This only works for this example because of the way it is hosted.
E.g. this could not be wrapped into a library, but the current implementation supports this correctly.

It could maybe be done through the `package.json`, but unfortunately I'm not familiar enough with the webpack ecosystem to say more.

With that said I'm going to close this, but please feel free to continue the discussion if you have any more ideas or feedback!

PR #3888, opened 2024-03-12
Add methods for Typed Array and Array

swnb:
1. add [find_last_index](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex) and [find_last](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLast) for **Array**
2. add [copy_within](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/copyWithin) for **Typed Array**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Implement `From<NonNull<T>>` for `JsValue`.
   [#3877](https://github.com/rustwasm/wasm-bindgen/pull/3877)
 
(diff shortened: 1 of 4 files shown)


daxpedda (changes requested):
Needs an entry in the changelog.
Otherwise LGTM, thank you for adding tests as well!

daxpedda (approved):
Thank you!

PR #3890, opened 2024-03-13
add keys, values, entries methods for FileSystemDirectoryHandle

jmacmahon:
These methods were missing, and a directory handle is pretty useless without a way to list the contents of the directory!


jmacmahon:
Seems I've done this wrong and made the modifications in generated code.

Issue #3468, opened 2023-06-06
Interning internal enum representations and field names


grovesNL:
### Motivation

We use a lot of enums for WebGPU at high frequency, e.g. https://docs.rs/web-sys/0.3.63/web_sys/struct.GpuRenderPassDepthStencilAttachment.html#method.stencil_load_op

These enums are represented as strings in the WebIDL, e.g. https://github.com/rustwasm/wasm-bindgen/blob/5f10d6f0d512b2a57710973efa6ade6afd6ef39a/crates/web-sys/webidls/unstable/WebGPU.webidl#L1016-L1019

This is problematic because we end up going through [`JsValue::from_str` and creating a significant amount of temporary string allocations](https://docs.rs/web-sys/0.3.63/src/web_sys/features/gen_GpuRenderPassDepthStencilAttachment.rs.html#151-152) for both string enum values and field names.

### Proposed Solution

Values like `load` and `clear` in the example above could be internally interned (i.e., cached on the JS side) so we don't create temporary strings for these.

I don't know how we'd expose this option, but ideally we'd want to be able to control whether string enum values and field names are interned somehow. This might need to be fine-grained so that only high-frequency enums/field names are interned.

We could consider lazy-loading *all* string enum values and field names into a cache for interning but that might be excessive.

### Alternatives
We could use wasm-bindgen's existing `enable-interning` feature somehow but we can't use the `intern` function on enums or field names since they're generated internally in the WebIDL bindings.

We could also consider internally representing these as something other than a string (e.g., remapping to `u32` and mapping it back to a string on the JS side).

### Additional Context

This adds significant overhead in real use cases so it's not a micro-optimization. [I raised this to the WebGPU specification in the early days](https://github.com/gpuweb/gpuweb/pull/71#issuecomment-415038836) (hoping we could use `u32` instead to avoid the string overhead). We're seeing the performance problems show up in projects now since wasm host bindings aren't available yet.

I'd guess this affects a lot of other generated bindings that are called at high frequencies too (e.g., DOM access, audio) so this could improve performance for a lot of projects, especially if they're sensitive to GC hitching (like most WebGPU applications).

daxpedda:
> We could consider lazy-loading _all_ string enum values and field names into a cache for interning but that might be excessive.

That's the solution I came up with immediately as well. It doesn't sound too excessive to be honest, considering applications that use WebGPU are probably not small.

[...]

Issue #2139, opened 2020-05-15
implement OptionIntoWasmAbi for JsValue


ctaggart:
### Motivation

On several occasions, I've want to map a parameter as `Option<&JsValue>` or a return type as `Option<JsValue>`, just to indicate explicitly that it is optional. I realize JsValue supports null & undefined. 

### Proposed Solution

Implement the trait.

### Alternatives

I think of `JsValue` as the TypeScript `any` type. TypeScript also has the [non-primitive object type](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types) that does not support null or undefined, but everything else. May be it would be better to add that so it makes more sense to have a `Options<JsSome>` or something like that.

alexcrichton:
The reason we haven't implemented this is that there's no way to distinguish between `Some(null)` and `None` in JS, so there's no way to losslessly represent the value in JS. We've currently been requiring that Rust opts-in to `.unwrap_or(...)` or something like that.

ctaggart:
Okay, I may revisit this again in the future with more concrete examples, after I ship the generator.

PR #3898, opened 2024-03-19
Don't use reflect in webidl dictionary setters

Davidster:
Implements solution 1 from https://github.com/rustwasm/wasm-bindgen/issues/3468#issuecomment-1589958213.

Solution 3 will come after assuming this goes smoothly


Davidster:
woops, looks like the CI is failing. will check that tmrw.

Davidster:
I've run into the following error with the codegen:

it generates some shim functions now such as:

```rust
#![allow(unused_imports)]
#![allow(clippy::all)]
use super::*;
use wasm_bindgen::prelude::*;
#[wasm_bindgen]
extern "C" {
    # [wasm_bindgen (extends = :: js_sys :: Object , js_name = RequestInit)]
    #[derive(Debug, Clone, PartialEq, Eq)]
    #[doc = "The `RequestInit` dictionary."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub type RequestInit;
    # [wasm_bindgen (method , setter = body)]
    fn body_shim(this: &RequestInit, val: Option<&::wasm_bindgen::JsValue>);
    #[cfg(feature = "RequestCache")]
    # [wasm_bindgen (method , setter = cache)]
    fn cache_shim(this: &RequestInit, val: RequestCache);
    #[cfg(feature = "RequestCredentials")]
    # [wasm_bindgen (method , setter = credentials)]
    fn credentials_shim(this: &RequestInit, val: RequestCredentials);
    # [wasm_bindgen (method , setter = headers)]
    fn headers_shim(this: &RequestInit, val: &::wasm_bindgen::JsValue);
    # [wasm_bindgen (method , setter = integrity)]
    fn integrity_shim(this: &RequestInit, val: &str);
    # [wasm_bindgen (method , setter = method)]
    fn method_shim(this: &RequestInit, val: &str);
    #[cfg(feature = "RequestMode")]
    # [wasm_bindgen (method , setter = mode)]
    fn mode_shim(this: &RequestInit, val: RequestMode);
    #[cfg(feature = "ObserverCallback")]
    # [wasm_bindgen (method , setter = observe)]
    fn observe_shim(this: &RequestInit, val: &ObserverCallback);
    #[cfg(feature = "RequestRedirect")]
    # [wasm_bindgen (method , setter = redirect)]
    fn redirect_shim(this: &RequestInit, val: RequestRedirect);
    # [wasm_bindgen (method , setter = referrer)]
    fn referrer_shim(this: &RequestInit, val: &str);
    #[cfg(feature = "ReferrerPolicy")]
    # [wasm_bindgen (method , setter = referrerPolicy)]
    fn referrer_policy_shim(this: &RequestInit, val: ReferrerPolicy);
    #[cfg(feature = "AbortSignal")]
    # [wasm_bindgen (method , setter = signal)]
    fn signal_shim(this: &RequestInit, val: Option<&AbortSignal>);
}
impl RequestInit {
    #[doc = "Construct a new `RequestInit`."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub fn new() -> Self {
        #[allow(unused_mut)]
        let mut ret: Self = ::wasm_bindgen::JsCast::unchecked_into(::js_sys::Object::new());
        ret
    }
    #[doc = "Change the `body` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub fn body(&mut self, val: Option<&::wasm_bindgen::JsValue>) -> &mut Self {
        self.body_shim(val);
        self
    }
    #[cfg(feature = "RequestCache")]
    #[doc = "Change the `cache` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestCache`, `RequestInit`*"]
    pub fn cache(&mut self, val: RequestCache) -> &mut Self {
        self.cache_shim(val);
        self
    }
    #[cfg(feature = "RequestCredentials")]
    #[doc = "Change the `credentials` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestCredentials`, `RequestInit`*"]
    pub fn credentials(&mut self, val: RequestCredentials) -> &mut Self {
        self.credentials_shim(val);
        self
    }
    #[doc = "Change the `headers` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub fn headers(&mut self, val: &::wasm_bindgen::JsValue) -> &mut Self {
        self.headers_shim(val);
        self
    }
    #[doc = "Change the `integrity` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub fn integrity(&mut self, val: &str) -> &mut Self {
        self.integrity_shim(val);
        self
    }
    #[doc = "Change the `method` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub fn method(&mut self, val: &str) -> &mut Self {
        self.method_shim(val);
        self
    }
    #[cfg(feature = "RequestMode")]
    #[doc = "Change the `mode` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`, `RequestMode`*"]
    pub fn mode(&mut self, val: RequestMode) -> &mut Self {
        self.mode_shim(val);
        self
    }
    #[cfg(feature = "ObserverCallback")]
    #[doc = "Change the `observe` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `ObserverCallback`, `RequestInit`*"]
    pub fn observe(&mut self, val: &ObserverCallback) -> &mut Self {
        self.observe_shim(val);
        self
    }
    #[cfg(feature = "RequestRedirect")]
    #[doc = "Change the `redirect` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`, `RequestRedirect`*"]
    pub fn redirect(&mut self, val: RequestRedirect) -> &mut Self {
        self.redirect_shim(val);
        self
    }
    #[doc = "Change the `referrer` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
    pub fn referrer(&mut self, val: &str) -> &mut Self {
        self.referrer_shim(val);
        self
    }
    #[cfg(feature = "ReferrerPolicy")]
    #[doc = "Change the `referrerPolicy` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `ReferrerPolicy`, `RequestInit`*"]
    pub fn referrer_policy(&mut self, val: ReferrerPolicy) -> &mut Self {
        self.referrer_policy_shim(val);
        self
    }
    #[cfg(feature = "AbortSignal")]
    #[doc = "Change the `signal` field of this object."]
    #[doc = ""]
    #[doc = "*This API requires the following crate features to be activated: `AbortSignal`, `RequestInit`*"]
    pub fn signal(&mut self, val: Option<&AbortSignal>) -> &mut Self {
        self.signal_shim(val);
        self
    }
}
impl Default for RequestInit {
    fn default() -> Self {
        Self::new()
    }
}
```

Which seems to look ok except that it produces the following error:

```rust
5 | #[wasm_bindgen]
  | ^^^^^^^^^^^^^^^ the trait `OptionIntoWasmAbi` is not implemented for `&wasm_bindgen::JsValue`
```

Looks like this is a case of https://github.com/rustwasm/wasm-bindgen/issues/2139. I'm not quite sure how to solve it. Any ideas?

Davidster:
Found a way to workaround the issue with some dark magic... let me know what you think

Davidster:
I will fix the merge conflict once this is ready to be merged. I needed to work on an older version in order be able to build my engine against my wasm_bindgen fork

Davidster:
I noticed when compiling my engine against this branch that the generated js file got substantially larger since the new 'shim' functions added by the generator are getting added regardless of whether the function is actually used. Any idea about how to mitigate this?

Davidster:
Ran into another problem with this approach...

```rust
error: expected `,`
  --> crates/web-sys/src/features/gen_WidevineCdmManifest.rs:19:42
   |
19 |     # [wasm_bindgen (method , setter = x - cdm - codecs)]
   |                                          ^
error: expected `,`
  --> crates/web-sys/src/features/gen_WidevineCdmManifest.rs:21:42
   |
21 |     # [wasm_bindgen (method , setter = x - cdm - host - versions)]
   |                                          ^
error: expected `,`
  --> crates/web-sys/src/features/gen_WidevineCdmManifest.rs:23:42
   |
23 |     # [wasm_bindgen (method , setter = x - cdm - interface - versions)]
   |                                          ^
error: expected `,`
  --> crates/web-sys/src/features/gen_WidevineCdmManifest.rs:25:42
   |
25 |     # [wasm_bindgen (method , setter = x - cdm - module - versions)]
   |  
```

is there any way to escape the `-` characters here?

daxpedda:
> is there any way to escape the `-` characters here?

You will have to use quotation marks and parse it as a string instead of an ident.

---

Why are the `*_shim` functions necessary?

Davidster:
Wouldn't adding quotation marks be a breaking change though?

I didn't consider whether the shims were needed or not tbh. But they would be necessary for the conversion from Option to JsValue since `OptionIntoWasmAbi` is not implemented for JsValue, like for RequestInit::body()

daxpedda:
> Wouldn't adding quotation marks be a breaking change though?

You don't have to require it, just supporting it.
I didn't look deeper in whats going on though and I don't know why you would need `-` in the first place.

> I didn't consider whether the shims were needed or not tbh. But they would be necessary for the conversion from Option to JsValue since `OptionIntoWasmAbi` is not implemented for JsValue, like for RequestInit::body()

I'm not following.
But I remembered why this was necessary: to keep the API the same by returning `&mut Self`.
So all good there!

Davidster:
the `-` is required because the WidevineCDMManifest webidl type has some attributes with `-` in the name 😢  see https://github.com/rustwasm/wasm-bindgen/blob/88efe468e68cf4f0d8cee87ee3a01de5f5739656/crates/web-sys/webidls/enabled/WidevineCDMManifest.webidl#L11

---

For the Option thing, it's like I described in my comment: https://github.com/rustwasm/wasm-bindgen/pull/3898#issuecomment-2010887953.

I'm not 100% sure I understand the compile error completely, but my guess is that the wasm-bindgen macro for body shim:

```rust
#[wasm_bindgen(method, setter = body)]
fn body_shim(this: &RequestInit, val: Option<&::wasm_bindgen::JsValue>);
```

Is trying to use the OptionIntoWasmAbi trait on the val parameter in order to be able to pass it to JS. So the fix was to change the signature to this:

```rust
#[wasm_bindgen(method, setter = body)]
fn body_shim(this: &RequestInit, val: &::wasm_bindgen::JsValue);
```

And convert from Option -> JsValue before calling the shim, like this:

```rust
#[doc = "Change the `body` field of this object."]
#[doc = ""]
#[doc = "*This API requires the following crate features to be activated: `RequestInit`*"]
pub fn body(&mut self, val: Option<&::wasm_bindgen::JsValue>) -> &mut Self {
    self.body_shim(val.unwrap_or(&::wasm_bindgen::JsValue::NULL));
    self
}
```

This sadly requires a special check for Option<&::wasm_bindgen::JsValue> in the code:

https://github.com/rustwasm/wasm-bindgen/pull/3898/commits/c67717dc626e985767307594f1436733465bd7bc#diff-d8976bd33550e367803742179ae21879225e06a871ccfe1e29d8b89af9106d02R697

Davidster:
I've added support for the hyphens, this PR should be ready for review now :smile: thanks!

daxpedda (changes requested):
LGTM so far.

Are you having issues generating the WebIDL?
That's still the last step missing.

crates/webidl/Cargo.toml
```diff
@@ -20,7 +20,7 @@ heck = "0.3"
 log = "0.4.1"
 proc-macro2 = "1.0"
 quote = '1.0'
-syn = { version = '2.0', features = ['full'] }
+syn = { version = '2.0', features = ['full', 'extra-traits'] }
```
daxpedda, line 23: Note to self: figure out how to remove this requirement.

Davidster:
Nope I'm all good with the webidl, it's just a huge diff and I wasn't sure if those are generated by the CI. I'll commit them

crates/webidl/Cargo.toml
Davidster: I took a quick look at it, one approach could be to do the check on the witter type here: https://github.com/rustwasm/wasm-bindgen/blob/a9f06701c78c9623772fe3f1a3e2cf0d60241114/crates/webidl/src/lib.rs#L380 and then pass along a boolean `is_js_value_ref_option_type`.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 * Stabilize Web Share API.
   [#3882](https://github.com/rustwasm/wasm-bindgen/pull/3882)
 
+* Don't use reflect in WebIDL dictionary setters. Also, importing getters/setters from JS now supports names containing hyphens by optionally surrounding the name with quotes, e.g. `#[wasm_bindgen(method, setter = "x-cdm-codecs")]`.
+  [#3898](https://github.com/rustwasm/wasm-bindgen/pull/3898)
```
daxpedda, line 23: I think the point here is not supporting hyphens, but strings in general.
It would also be good to mention that Web API bindings are going to be larger now.

crates/webidl/Cargo.toml
daxpedda: Please go ahead and do that then.

daxpedda (changes requested):
Small nits, otherwise LGTM!

This is a rather large change, so I will pull in @Liamolucko for a second approval as well.

CHANGELOG.md
Davidster: Looking at the generated code for `#[wasm_bindgen(method, setter = "x-cdm-codecs")]` I'm realising I really don't understand how it works.. I was never able to find the generated JS code that calls the setter or assigns the value even when closely inspecting a dummy project that makes use of it. But it somehow magically works :shrug: 

Anyways, I updated the changelog as requested

crates/webidl/Cargo.toml
Davidster: I found a way to do it by matching the idl type and replicating the logic from IdlType::to_syn_type but I don't love it.. it feels much less robust. It's your call though.

Liamolucko (approved):
LGTM!

crates/webidl/src/generator.rs
```diff
@@ -642,39 +679,58 @@ impl DictionaryField {
...
+        let shim_name = self.shim_name();
+
+        let js_value_ref_type = shared_ref(
+            leading_colon_path_ty(vec![rust_ident("wasm_bindgen"), rust_ident("JsValue")]),
+            false,
+        );
+        let js_value_ref_option_type = option_ty(js_value_ref_type.clone());
+
+        let shim_args = if ty == &js_value_ref_option_type {
+            quote! { val.unwrap_or(#js_value_ref_type::NULL) }
```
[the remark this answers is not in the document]
Davidster: lol, nice catch, the fact that this code actually worked was a total accident

Liamolucko:
> I noticed when compiling my engine against this branch that the generated js file got substantially larger since the new 'shim' functions added by the generator are getting added regardless of whether the function is actually used. Any idea about how to mitigate this? 

Could you elaborate a bit more? I'm not sure if you mean that the shims are never even referenced in Rust (in which case no JS for them should be getting generated), or that they could theoretically be used but aren't in practice (in which case I don't really have a good solution).

Davidster:
Looking at it again I think I was just confused at the time. It makes sense that a bunch of shims would get added to my JS file since they are indeed referenced from rust. 

The part that confused me was that even the non-shim versions weren't showing up in my JS file before the changes from this PR. But this makes sense because those are compiled into the wasm, not the JS :smile:

crates/webidl/Cargo.toml
daxpedda: This is pretty ugly indeed, but I think it's very important for a library like `wasm-bindgen` to reduce its dependency footprint.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 * Stabilize Web Share API.
   [#3882](https://github.com/rustwasm/wasm-bindgen/pull/3882)
 
+* Stop using reflect in WebIDL dictionary setters, instead create a shim to allow the wasm_bindgen macro to implement the function in javascript which avoids passing the attribute name string over the WASM/JS boundary. This increases the size of the Web API bindings but should be more efficient. Also, importing getters/setters from JS now supports specifying the JS attribute name as a string, e.g. `#[wasm_bindgen(method, setter = "x-cdm-codecs")]`.
```
daxpedda, line 22:
```suggestion
* Generate JS bindings for WebIDL dictionary setters instead of using `Reflect`. This increases the size of the Web API bindings but should be more performant. Also, importing getters/setters from JS now supports specifying the JS attribute name as a string, e.g. `#[wasm_bindgen(method, setter = "x-cdm-codecs")]`.
```

daxpedda (approved):
Thank you!

PR #3906, opened 2024-03-29
Add js import enum benchmark

Davidster:
In preparation for solution 3 from https://github.com/rustwasm/wasm-bindgen/issues/3468#issuecomment-1589958213, I have added a benchmark for the case of passing imported enums to a js function.

Current benchmark results is sitting at 212/s ±3.51% (around half a microsecond per function call, which is extremely slow)

Spoiler: I have a (unfinished) branch containing solution 3 and the benchmark goes up to 12,183/s ±1.1% which brings it in line with the 'Call a custom JS class Foo.bar method with structural' benchmark. In my engine, this means I save 2ms per frame in scenes with lots of objects (where set_index_buffer is called often).


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
 * Added support for returning `Vec`s from async functions.
   [#3630](https://github.com/rustwasm/wasm-bindgen/pull/3630)
 
+* Add js import enum benchmark
+  [#3906](https://github.com/rustwasm/wasm-bindgen/pull/3906)
```
daxpedda, line 18: The changelog is only for public facing changes.

.gitignore
daxpedda, line 1: Missing a newline at the end.

benchmarks/globals.js
daxpedda, line 1: Missing a newline at the end.

daxpedda (changes requested):
Apart from the nits, LGTM!

daxpedda (approved):
Thank you!

PR #3915, opened 2024-04-05
Improve the performance of 'string enums'

### LINKED ISSUE #3468: Interning internal enum representations and field names

Davidster:
Closes #3468


crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,113 @@ impl ToTokens for ast::ImportType {
...
+            impl #wasm_bindgen::convert::FromWasmAbi for #enum_name {
+                type Abi = u32;
 
-                #[inline]
-                fn into_abi(self) -> Self::Abi {
-                    <#wasm_bindgen::JsValue as #wasm_bindgen::convert::IntoWasmAbi>::into_abi(self.into())
+                unsafe fn from_abi(val: u32) -> Self {
+                    match val {
+                        #(#variant_indices => #variant_paths_ref,)*
+                        _ => Self::__Invalid
```
Davidster, line 1088: Returning __Invalid here preserves the current behaviour and allows the invalid_enum_return test to pass. Is this the approach we want to take or would panicking be better?

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,113 @@ impl ToTokens for ast::ImportType {
...
+                inform(#length);
+                #(inform(#chars);)*
+            }
+        });
+
         (quote! {
             #(#attrs)*
-            #vis enum #name {
-                #(#variants = #variant_indexes_ref,)*
+            #[non_exhaustive]
```
Davidster, line 1062: Should we explore removing the non_exhaustive here?

Davidster:
For your convenience, this is what a fully-expanded ImportEnum looks like now:

<details>

```rust
#![feature(prelude_import)]
//! Raw API bindings for Web APIs
//!
//! This is a procedurally generated crate from browser WebIDL which provides a
//! binding to all APIs that browsers provide on the web.
//!
//! This crate by default contains very little when compiled as almost all of
//! its exposed APIs are gated by Cargo features. The exhaustive list of
//! features can be found in `crates/web-sys/Cargo.toml`, but the rule of thumb
//! for `web-sys` is that each type has its own cargo feature (named after the
//! type). Using an API requires enabling the features for all types used in the
//! API, and APIs should mention in the documentation what features they
//! require.
#![doc(html_root_url = "https://docs.rs/web-sys/0.3")]
#![allow(deprecated)]
#[prelude_import]
use std::prelude::rust_2018::*;
#[macro_use]
extern crate std;
mod features {
    #[cfg(feature = "RequestMode")]
    #[allow(non_snake_case)]
    mod gen_RequestMode {
        #![allow(unused_imports)]
        #![allow(clippy::all)]
        use wasm_bindgen::prelude::*;
        ///The `RequestMode` enum.
        ///
        ///*This API requires the following crate features to be activated: `RequestMode`*
        #[non_exhaustive]
        #[repr(u32)]
        pub enum RequestMode {
            SameOrigin = 0u32,
            NoCors = 1u32,
            Cors = 2u32,
            Navigate = 3u32,
            #[automatically_derived]
            #[doc(hidden)]
            __Invalid,
        }
        #[automatically_derived]
        impl ::core::fmt::Debug for RequestMode {
            #[inline]
            fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
                ::core::fmt::Formatter::write_str(
                    f,
                    match self {
                        RequestMode::SameOrigin => "SameOrigin",
                        RequestMode::NoCors => "NoCors",
                        RequestMode::Cors => "Cors",
                        RequestMode::Navigate => "Navigate",
                        RequestMode::__Invalid => "__Invalid",
                    },
                )
            }
        }
        #[automatically_derived]
        impl ::core::clone::Clone for RequestMode {
            #[inline]
            fn clone(&self) -> RequestMode {
                *self
            }
        }
        #[automatically_derived]
        impl ::core::marker::Copy for RequestMode {}
        #[automatically_derived]
        impl ::core::marker::StructuralPartialEq for RequestMode {}
        #[automatically_derived]
        impl ::core::cmp::PartialEq for RequestMode {
            #[inline]
            fn eq(&self, other: &RequestMode) -> bool {
                let __self_tag = ::core::intrinsics::discriminant_value(self);
                let __arg1_tag = ::core::intrinsics::discriminant_value(other);
                __self_tag == __arg1_tag
            }
        }
        #[automatically_derived]
        impl ::core::marker::StructuralEq for RequestMode {}
        #[automatically_derived]
        impl ::core::cmp::Eq for RequestMode {
            #[inline]
            #[doc(hidden)]
            #[coverage(off)]
            fn assert_receiver_is_total_eq(&self) -> () {}
        }
        #[automatically_derived]
        impl wasm_bindgen::convert::IntoWasmAbi for RequestMode {
            type Abi = u32;
            #[inline]
            fn into_abi(self) -> u32 {
                self as u32
            }
        }
        #[automatically_derived]
        impl wasm_bindgen::convert::FromWasmAbi for RequestMode {
            type Abi = u32;
            unsafe fn from_abi(val: u32) -> Self {
                match val {
                    0u32 => RequestMode::SameOrigin,
                    1u32 => RequestMode::NoCors,
                    2u32 => RequestMode::Cors,
                    3u32 => RequestMode::Navigate,
                    4u32 => RequestMode::__Invalid,
                    _ => {
                        ::core::panicking::unreachable_display(
                            &"The JS binding should only ever produce a valid value or the specific 'invalid' value",
                        );
                    }
                }
            }
        }
        #[automatically_derived]
        impl wasm_bindgen::convert::OptionFromWasmAbi for RequestMode {
            #[inline]
            fn is_none(val: &u32) -> bool {
                *val == 5u32
            }
        }
        #[automatically_derived]
        impl wasm_bindgen::convert::OptionIntoWasmAbi for RequestMode {
            #[inline]
            fn none() -> Self::Abi {
                5u32
            }
        }
        #[automatically_derived]
        impl wasm_bindgen::describe::WasmDescribe for RequestMode {
            fn describe() {
                use wasm_bindgen::describe::*;
                inform(IMPORT_ENUM);
                inform(11u32);
                inform(82u32);
                inform(101u32);
                inform(113u32);
                inform(117u32);
                inform(101u32);
                inform(115u32);
                inform(116u32);
                inform(77u32);
                inform(111u32);
                inform(100u32);
                inform(101u32);
                inform(4u32);
                inform(5u32);
                inform(4u32);
                inform(11u32);
                inform(115u32);
                inform(97u32);
                inform(109u32);
                inform(101u32);
                inform(45u32);
                inform(111u32);
                inform(114u32);
                inform(105u32);
                inform(103u32);
                inform(105u32);
                inform(110u32);
                inform(7u32);
                inform(110u32);
                inform(111u32);
                inform(45u32);
                inform(99u32);
                inform(111u32);
                inform(114u32);
                inform(115u32);
                inform(4u32);
                inform(99u32);
                inform(111u32);
                inform(114u32);
                inform(115u32);
                inform(8u32);
                inform(110u32);
                inform(97u32);
                inform(118u32);
                inform(105u32);
                inform(103u32);
                inform(97u32);
                inform(116u32);
                inform(101u32);
            }
        }
        #[automatically_derived]
        impl wasm_bindgen::__rt::core::convert::From<RequestMode>
        for wasm_bindgen::JsValue {
            fn from(val: RequestMode) -> Self {
                wasm_bindgen::JsValue::from_str(
                    match val {
                        RequestMode::SameOrigin => "same-origin",
                        RequestMode::NoCors => "no-cors",
                        RequestMode::Cors => "cors",
                        RequestMode::Navigate => "navigate",
                        RequestMode::__Invalid => {
                            wasm_bindgen::throw_str(
                                "Converting an invalid import enum back to a string is currently not supported",
                            )
                        }
                        _ => {
                            ::core::panicking::unreachable_display(
                                &"All possible variants should have been checked",
                            );
                        }
                    },
                )
            }
        }
    }
    #[cfg(feature = "RequestMode")]
    #[allow(unused_imports)]
    pub use gen_RequestMode::*;
}
pub use features::*;
pub use js_sys;
pub use wasm_bindgen;
```

</details>

Davidster:
I'm not sure what's wrong with the CI. The test passes locally for me. Any ideas?

WinterAlexander (approved):

daxpedda:
Apologies for the delay, I was on vacation, still catching up.
Planning to take a look tomorrow!

> I'm not sure what's going on with the CI. The `cargo test --manifest-path crates/web-sys/Cargo.toml --target wasm32-unknown-unknown --all-features` test passes locally for me. Any ideas?

Run with `RUSTFLAGS=--cfg=web_sys_unstable_apis`, which reproduced the bug for me locally as well.

Davidster:
No worries! Thanks I missed the flag

crates/wasm-interpreter/src/lib.rs
```diff
@@ -68,11 +68,9 @@ impl Interpreter {
...
         let mut ret = Interpreter::default();
 
-        // The descriptor functions shouldn't really use all that much memory
-        // (the LLVM call stack, now the wasm stack). To handle that let's give
-        // our selves a little bit of memory and set the stack pointer (global
-        // 0) to the top.
-        ret.mem = vec![0; 0x400];
+        // Give ourselves some memory and set the stack pointer
+        // (the LLVM call stack, now the wasm stack, global 0) to the top.
+        ret.mem = vec![0; 0x8000];
```
Davidster, line 73: So this was... fun to debug

The wasm interpreter/runtime that's initialized here holds onto a stack pointer that starts at the highest memory location mem.len(), which was 0x400 originally (1024). Then, as the program runs and local variables are pushed onto the stack the stack pointer is decremented for each new variable. The other changes in this PR increased the complexity of some of the describes by a large amount, and so the stack pointer eventually got decremented so many times in one of those functions that it fell below 0 which is illegal (a.k.a stack overflow). So I increased the program memory up to 0x2000 (8192) and then the crash stopped happening.

In particular it was the GpuTextureFormat enum that was causing the problem due to the number of enum variants it has (and therefore the number of instructions required to implement the describe function).

This post helped me understand this issue: https://hackmd.io/@pepyakin/SkmPKGhiq

MartensCedric (approved):
LGTM

adrientremblay (approved):
Good job

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,115 @@ impl ToTokens for ast::ImportType {
 impl ToTokens for ast::ImportEnum {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         let vis = &self.vis;
-        let name = &self.name;
-        let expect_string = format!("attempted to convert invalid {} into JSValue", name);
+        let enum_name = &self.name;
+        let name_str = enum_name.to_string();
+        let name_len = name_str.len() as u32;
+        let name_chars = name_str.chars().map(|c| c as u32);
```
daxpedda, line 1030:
```suggestion
        let name_chars = name_str.chars().map(u32::from);
```

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,115 @@ impl ToTokens for ast::ImportType {
...
             .collect();
 
         // Borrow variant_paths because we need to use it multiple times inside the quote! macro
         let variant_paths_ref = &variant_paths;
 
         let wasm_bindgen = &self.wasm_bindgen;
 
+        let describe_variants = self.variant_values.iter().map(|variant_value| {
+            let length = variant_value.len() as u32;
+            let chars = variant_value.chars().map(|c| c as u32);
```
daxpedda, line 1053:
```suggestion
            let chars = variant_value.chars().map(u32::from);
```

crates/backend/src/codegen.rs
daxpedda: What was your motivation to add it in the first place?

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,115 @@ impl ToTokens for ast::ImportType {
...
-                    match self {
-                        #(#variant_paths_ref => #variant_strings,)*
-                        #name::__Nonexhaustive => panic!(#expect_string),
-                    }
-                }
-
-                #vis fn from_js_value(obj: &#wasm_bindgen::JsValue) -> Option<#name> {
-                    obj.as_string().and_then(|obj_str| Self::from_str(obj_str.as_str()))
-                }
-            }
```
daxpedda, line 1086: As unfortunate as the existence of those are, we can't really remove them, as that would be a breaking change.

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,115 @@ impl ToTokens for ast::ImportType {
...
-                    #wasm_bindgen::JsValue::from(obj.to_str())
+            impl #wasm_bindgen::__rt::core::convert::From<#enum_name> for
+                #wasm_bindgen::JsValue
+            {
+                fn from(val: #enum_name) -> Self {
+                    #wasm_bindgen::JsValue::from_str(
+                        match val {
+                            #(#variant_paths_ref => #variant_values,)*
+                            #enum_name::__Invalid => #wasm_bindgen::throw_str("Converting an invalid import enum back to a string is currently not supported"),
+                            _ => unreachable!("All possible variants should have been checked")
```
daxpedda, line 1129: Why is this line necessary?

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,115 @@ impl ToTokens for ast::ImportType {
...
-                fn from(obj: #name) -> #wasm_bindgen::JsValue {
-                    #wasm_bindgen::JsValue::from(obj.to_str())
+            impl #wasm_bindgen::__rt::core::convert::From<#enum_name> for
+                #wasm_bindgen::JsValue
+            {
+                fn from(val: #enum_name) -> Self {
+                    #wasm_bindgen::JsValue::from_str(
+                        match val {
+                            #(#variant_paths_ref => #variant_values,)*
+                            #enum_name::__Invalid => #wasm_bindgen::throw_str("Converting an invalid import enum back to a string is currently not supported"),
```
daxpedda, line 1128: To keep the same behavior as before we should not throw here, but we can instantiate a `JsValue::UNDEFINED` instead.

crates/backend/src/codegen.rs
```diff
@@ -1024,114 +1024,115 @@ impl ToTokens for ast::ImportType {
...
-                #[inline]
-                fn is_none(abi: &Self::Abi) -> bool { <::js_sys::Object as #wasm_bindgen::convert::OptionFromWasmAbi>::is_none(abi) }
+            impl #wasm_bindgen::describe::WasmDescribe for #enum_name {
+                fn describe() {
+                    use #wasm_bindgen::describe::*;
+                    inform(IMPORT_ENUM);
+                    inform(#name_len);
+                    #(inform(#name_chars);)*
+                    inform(#invalid);
+                    inform(#hole);
```
daxpedda, line 1114: It seems to me like we can remove those from the description, as they can be derived from `variant_count`.

crates/cli-support/src/descriptor.rs
```diff
@@ -67,7 +68,16 @@ pub enum Descriptor {
...
     NamedExternref(String),
-    Enum { name: String, hole: u32 },
+    Enum {
+        name: String,
+        hole: u32,
+    },
+    ImportEnum {
+        name: String,
+        invalid: u32,
+        hole: u32,
```
daxpedda, line 78: Despite my suggestion to remove them from the description, I think it would be nice if we still keep those fields for ease-of-use.

crates/cli-support/src/descriptor.rs
daxpedda, line 1: The schema version needs to be bumped as well, see https://github.com/rustwasm/wasm-bindgen/pull/3651/files for an example.

daxpedda (changes requested):
Looks great so far, basically just a couple of nits.

Again, I apologize for the delay, I extended my vacation a bit spontaneously.

crates/backend/src/codegen.rs
Davidster: oops, missed the `#vis`

crates/backend/src/codegen.rs
Davidster: as per https://rust-lang.github.io/rfcs/2008-non-exhaustive.html, __Nonexhaustive was the old way of implementing non-exhaustive enums so I wanted to preserve the original behavior to limit the scope of this PR.

crates/backend/src/codegen.rs
Davidster: it's needed because the enum is [marked as exhaustive](https://github.com/rustwasm/wasm-bindgen/pull/3915/files?diff=split&w=0#diff-5b89f2e6e335fbc4662194fa0fe557251f21ef1c7a18d50db8e63a30851ca3bbR1062)

crates/backend/src/codegen.rs
Davidster: from what I can tell from the current impl, the current behavior in the equivalent case (__Nonexhaustive) will actually panic here. where do you see it giving an undefined?

crates/backend/src/codegen.rs
Davidster: done in 164c457b779941b9a211930b7976a3490dabb3ed

crates/backend/src/codegen.rs
Davidster: done in 164c457b779941b9a211930b7976a3490dabb3ed

crates/backend/src/codegen.rs
Davidster: done in 164c457b779941b9a211930b7976a3490dabb3ed

crates/backend/src/codegen.rs
Davidster: done in 164c457b779941b9a211930b7976a3490dabb3ed

crates/cli-support/src/descriptor.rs
Davidster: done in 164c457b779941b9a211930b7976a3490dabb3ed

crates/backend/src/codegen.rs
```diff
@@ -1078,114 +1078,128 @@ impl ToTokens for ast::ImportType {
...
                         _ => None,
                     }
                 }
 
-                fn to_str(&self) -> &'static str {
-                    match self {
-                        #(#variant_paths_ref => #variant_strings,)*
-                        #name::__Nonexhaustive => panic!(#expect_string),
-                    }
-                }
```
daxpedda, line 1135: This function is still missing.

crates/backend/src/codegen.rs
daxpedda: No, you are right, this did indeed panic, I missed that `to_str()` panics on the `__Nonexhaustive` variant.

(using `JsValue::UNDEFINED` was just an alternative suggestion, I thought that originally it just converts it to a `"__Nonexhaustive"` string)

crates/backend/src/codegen.rs
daxpedda: Non-exhaustive enums can be matched exhaustively inside the crate they are defined in.
So this should not be required, unless I'm missing something here.

crates/backend/src/codegen.rs
daxpedda: Don't get me wrong, I think using `#[non_exhaustive]` is the right thing to do, but we didn't remove the hidden variant, which we can't for backwards compatibility reasons, so it seems to be unrelated to this PR.

Or am I missing something here?
Either way I'm fine with leaving it there.

crates/backend/src/codegen.rs
daxpedda: We want to stay backwards compatible, so it is fine like this.

daxpedda (changes requested):

crates/backend/src/codegen.rs
Davidster: Isn't this one private?

crates/backend/src/codegen.rs
Davidster: Ah yeah, forgot about that detail

crates/backend/src/codegen.rs
daxpedda: Maybe I'm missing something obvious here, but AFAIU, this doesn't only affect WebIDL, where it being private would not expose it to the user, but it affects any `#[wasm_bindgen] enum ...`.

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@
 * Generate JS bindings for WebIDL dictionary setters instead of using `Reflect`. This increases the size of the Web API bindings but should be more performant. Also, importing getters/setters from JS now supports specifying the JS attribute name as a string, e.g. `#[wasm_bindgen(method, setter = "x-cdm-codecs")]`.
   [#3898](https://github.com/rustwasm/wasm-bindgen/pull/3898)
 
+* Greatly improve the performance of sending WebIDL 'import enums' across the JavaScript boundary by converting the enum variant string to/from an int.
```
daxpedda, line 34: This should probably point out that this affects any enum definitions, not just WebIDL.

crates/backend/src/codegen.rs
Davidster: I'm realizing that nothing is currently stopping users from matching the __Nonexhaustive variant. In the link I shared, it seems like the unstable attribute is what's needed to make it unmatchable. So apparently adding non_exhaustive here is a breaking change.

crates/backend/src/codegen.rs
Davidster: Ah right, I hadn't thought about that use case. Thanks

crates/backend/src/codegen.rs
daxpedda: We did mark it as `#[doc(hidden)]` and its name starts with `__`, this intents to clearly mark it as unstable. This is unfortunately what has been done in the past before `#[non_exhaustive]` existed, so I think we are in the clear here, even though this is not clearly specified anywhere.

Unfortunately we aren't able to use the `#[unstable]` attribute, so it is what it is.
I think ideally here we would neither need `#[non_exhaustive]` nor the extra variant and make it so it panics/throws instead of constructing this variant. Unfortunately this would not be backwards compatible, so we are stuck with this until the next breaking change.

CHANGELOG.md
Davidster: Ive been calling them ImportEnums but they should really be called StringEnums as per the docs: https://rustwasm.github.io/wasm-bindgen/reference/types/imported-js-types.html. I think that's a better name. 

Shall I rename them across the board in the code? "ImportEnum" is a misnomer.

crates/backend/src/codegen.rs
daxpedda: In the case of WebIDL, yes.
But in the case of user defined string enums it should be possible to have them be exhaustive.

Davidster: Ok yeah, so the webidl->rust generator would add the attribute instead

crates/backend/src/codegen.rs
```diff
@@ -1078,114 +1078,128 @@ impl ToTokens for ast::ImportType {
...
-                fn from(obj: #name) -> #wasm_bindgen::JsValue {
-                    #wasm_bindgen::JsValue::from(obj.to_str())
+            impl #wasm_bindgen::describe::WasmDescribe for #enum_name {
+                fn describe() {
+                    use #wasm_bindgen::describe::*;
+                    inform(IMPORT_ENUM);
+                    inform(#name_len);
+                    #(inform(#name_chars);)*
+                    inform(#invalid);
+                    inform(#variant_count);
```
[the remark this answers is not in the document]
Davidster: done in 130ba7c692abe1b40ea04d8737ce49b6af82a385

crates/backend/src/codegen.rs
Davidster: done in 130ba7c692abe1b40ea04d8737ce49b6af82a385

crates/backend/src/codegen.rs
Davidster: done in 130ba7c692abe1b40ea04d8737ce49b6af82a385

crates/shared/src/schema_hash_approval.rs
```diff
@@ -8,7 +8,7 @@
 // If the schema in this library has changed then:
 //  1. Bump the version in `crates/shared/Cargo.toml`
 //  2. Change the `SCHEMA_VERSION` in this library to this new Cargo.toml version
-const APPROVED_SCHEMA_FILE_HASH: &str = "10197913343580353876";
```
daxpedda, line 11: Needs another update.

daxpedda (approved):
LGTM!

@Liamolucko would appreciate your review/approval as well!

crates/shared/src/schema_hash_approval.rs
Davidster: done in 9efa4f264b9cde6cb6f604c69eba57dde8f9facc

Davidster:
thanks for the review!

Liamolucko (approved):
Looks great!

crates/backend/src/encode.rs
```diff
@@ -339,8 +339,8 @@ fn shared_import_type<'a>(i: &'a ast::ImportType, intern: &'a Interner) -> Impor
     }
 }
 
-fn shared_import_enum<'a>(_i: &'a ast::ImportEnum, _intern: &'a Interner) -> ImportEnum {
-    ImportEnum {}
+fn shared_import_enum<'a>(_i: &'a ast::StringEnum, _intern: &'a Interner) -> StringEnum {
```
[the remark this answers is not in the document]
Davidster: I wasn't sure for this one. Based on impl of shared_import_kind it feels like the name should stay the same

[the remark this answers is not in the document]
daxpedda: I'm also fine as-is.

daxpedda:
@Davidster thank you for your contribution!

PR #3920, opened 2024-04-10
Add Support for cargo nextest

spigaz:
This is still work in progress:
https://nexte.st/book/custom-test-harnesses.html

Progress
- [x] Added support for **-h** and **--help** w/tests
- [x] Added support for **-V** and **--version** w/tests
- [x] Added support for **--list --format terse** w/tests
- [x] Added support for **--list --format --ignored** w/tests
- [x] Add support for **\<test-name\> --nocapture --exact** w/tests
- [x] Reinstated support for **--include-ignored** w/tests
- [x] Reinstated support for **--skip name** w/tests
- [x] Reinstated support for **--skip=name** w/tests
- [x] Reinstated support for **filter** w/tests
- [x] Solved the temporary directory conflict caused by the concurrent invocation
- [x] Update **--include-ignored** to use clap
- [x] Update **--skip name** to use clap
- [x] Added tests to execute tests on nodejs
- [x] Fixed support and added tests to execute tests on deno
- [x] Add tests to execute tests on chrome
- [x] Added tests to execute tests on firefox
- [x] Add tests to execute tests on safari
- [x] Add tests to execute tests on edge
- [x] Add tests with cargo nextest invoking wasm-bindgen-test-runner
- [x] Use feature! macro to simplify testing
- [x] Use feature! test_mode parameter to execute the specification over all the test_modes

Updated to use clap
- docopt was used already in wasm-bindgen, but its unmaintained https://github.com/docopt/docopt.rs and recommends clap or structopt, but structopt's docs state "As clap v3 is now out, and the structopt features are integrated into (almost as-is), structopt is now in maintenance mode"

Updated the macro wasm_bindgen_test
- To allow listing handling as required by nextext

Testing
- I wasn't sure what was the best place to put the tests in, because of the custom test runner in the cli crate, so I placed them on main tests folder.

- I used a variation of BDD that I use for many years now.
-- Although it seems a bit more verbose, it makes writing tests a lot simpler and faster, anyone can understand them and add new ones.
-- When something breaks its very easy to understand why.
-- Although conter-intuitive, in practice I have found that they are a lot easier to update on refactors.

Overhead
- When used from cargo nextest the overhead is pretty real, as it has to load the runtime and the wasm into it, so it can be a lot, but it runs them in parallel, so it makes up for it on tests that take a long to execute.

Architecture:
- I would prefer to organize the code into:
-- a folder with the CLI and environment stuff
-- a folder with the wasm handling
-- a folder with the runtimes


daxpedda:
Apologies for the delay, I was on vacation, still catching up.
Planning to take a look tomorrow!

spigaz:
That's okay!
I still have one missing piece ahead to get it working with cargo nextest, so its fine.

spigaz:
I was waiting for your review, because the more I move forward, the harder it will get.

Even with simple things, I end up finding more bugs and things to refactor.

Now I was trying to add a test to run deno, its raising the error "document is not defined", apparently it seems deno is broken, I have tried with several option even directly in the console, but I can get it to accept the document.
https://docs.deno.com/runtime/manual/node/faqs#getting-type-errors-like-cannot-find--document--or--htmlelement

I got big further in the runtime, when I force it to use the Runtime::None, but not there yet...

spigaz:
@daxpedda I have been trying to create a macro to simplify the tests, this will be particular useful for tests that should be run over the different runtimes supported.

I'm still working on the syntax, as the macro by itself, is allowing different usage patterns.

But right now, this is the format already working:
```rust
feature! {
    given_there_is_an_assembly_with_one_failing_test();
    when_wasm_bindgen_test_runner_is_invoked_with_the_option("-V");

    "Outputs the version" {
        then_the_standard_output_should_have(
            &format!("wasm-bindgen-test-runner {}", env!("CARGO_PKG_VERSION")),
        );
    }

    "Returns success" {
        then_success_should_have_been_returned();
    }
}
```

It expands to two tests
```rust
#[test]
fn outputs_the_wasm_bindgen_test_runner_version_information_feature() {
    let mut context = Context::new();
    given_there_is_an_assembly_with_one_failing_test(&mut context);
    when_wasm_bindgen_test_runner_is_invoked_with_the_option(&mut context, "-V");
    then_the_standard_output_should_have(
        &context,
        &format!("wasm-bindgen-test-runner {}", env!("CARGO_PKG_VERSION")),
    );
}

#[test]
fn returns_success_feature() {
    let mut context = Context::new();
    given_there_is_an_assembly_without_anything(&mut context);
    when_wasm_bindgen_test_runner_is_invoked_with_the_option(&mut context, "-V");
    then_success_should_have_been_returned(&context);
}
```

If the target platform is wasm it uses [wasm_bindgen_test::wasm_bindgen_test] instead.

This allows for a more compact file, because there are sometimes 5-7 different outcomes for a single execution context

The idea is by default to respect the single outcome, allowing easy troubleshooting of regressions, but its possible to aggregate the executions on CI for faster execution times.

Cargo.toml
```diff
@@ -55,6 +55,15 @@ serde_derive = "1.0"
...
 
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+assert_cmd = "1.0"
+lazy_static = "1.4.0"
+predicates = "1.0.0"
+rand = "0.8.5"
+regex = "1.10.4"
+
+[dev-dependencies]
+
```
daxpedda, line 66:
```suggestion
```

crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
daxpedda, line 1: I would prefer if we move any changes to Deno to a separate PR to make it easier for me to review.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -152,6 +211,9 @@ fn main() -> anyhow::Result<()> {
         },
         Some(_) => bail!("invalid __wasm_bingen_test_unstable value"),
         None if std::env::var("WASM_BINDGEN_USE_DENO").is_ok() => TestMode::Deno,
+        None if std::env::var("WASM_BINDGEN_USE_BROWSER").is_ok() => TestMode::Browser {
+            no_modules: std::env::var("WASM_BINDGEN_USE_NO_MODULE").is_ok(),
+        },
```
daxpedda, line 216: Why this change?
I think we could at least discuss and do this in a separate PR.

crates/cli/Cargo.toml
```diff
@@ -23,6 +23,7 @@ bin-dir = "wasm-bindgen-{ version }-{ target }/{ bin }{ binary-ext }"
 docopt = "1.0"
 env_logger = "0.8"
 anyhow = "1.0"
+fs2 = "0.4.3"
```
daxpedda, line 26: I assume this was an accidental leftover?

crates/test-macro/src/lib.rs
```diff
@@ -80,6 +80,12 @@ pub fn wasm_bindgen_test(
         None => quote! { ::core::option::Option::None },
     };
 
+    let ignore_str = match ignore.clone() {
+        Some(Some(_)) => "$",
+        Some(None) => "$",
```
daxpedda, line 85:
```suggestion
        Some(_) => "$",
```

daxpedda (changes requested):
> Updated to use docopt
> 
> * It was used already in wasm-bindgen
> * Not sure if you want to use something more modern, its easy to change
> * Just checked its unmaintained https://github.com/docopt/docopt.rs and recommends clap or structopt, but structopt's docs state "As clap v3 is now out, and the structopt features are integrated into (almost as-is), structopt is now in maintenance mode"

I think updating to Clap would be nice, seeing that Cargo uses it as well.
That said, lets do this in a follow-up PR.

> Updated the macro wasm_bindgen_test
> 
> * To export and additional method to allow listing handling as required by nextext
> * Not sure if it might make sense to override current export

I think it would be better to update the current macro if possible.

> Testing
> 
> * I wasn't sure what was the best place to put the tests in, because of the custom test runner in the cli crate, so I placed them on main tests folder.
> * I used a variation of BDD that I use for many years now.
>       -- Although it seems a bit more verbose, it makes writing tests a lot simpler and faster, anyone can understand them and add new ones.
>       -- When something breaks its very easy to understand why.
>       -- Although conter-intuitive, in practice I have found that they are a lot easier to update on refactors.

I'm unsure what exactly is going on here and I didn't review the tests yet, but from a glance I would like to change the following:
- Shouldn't the tests be in `/crates/cli/tests` instead of in `/tests`?
- Not sure why some folders are prefixed with `_`.
- Tests include extremely repetitive code and use ambiguous terms like "assembly". While I don't mind having all these files, I would prefer if we can cut down on the repeated code entirely.
- Why did you choose Firefox to be its own target? We should use the name "browser" instead.

> Overhead
> 
> * When used from cargo nextest the overhead is pretty real, as it has to load the runtime and the wasm into it, so it can be a lot, but it runs them in parallel, so it makes up for it on tests that take a long to execute.

That was to be expected, in follow-up steps we should discuss how to cut that down. The first idea that comes to mind is to introduce a way to start and end the headless browser separately from all the tests. That way each test doesn't spawn its own headless browser, which should cut down on overhead significantly.

Additionally some tests could be run in parallel without requiring their own context, e.g. tests in workers. But that would require some significant refactoring.

> Architecture:
> 
> * I would prefer to organize the code into:
>     -- a folder with the CLI and environment stuff
>     -- a folder with the wasm handling
>     -- a folder with the runtimes

Considering that the changes aren't actually that big, I would prefer if we do any further refactoring in a separate PR.

> @daxpedda I have been trying to create a macro to simplify the tests, this will be particular useful for tests that should be run over the different runtimes supported.

I think that's a great idea, but lets not use the name `feature`, because its unclear exactly what that means. Simply `test` or `multi_target_test` (the term target is clear in this context) would be enough I believe.

---

Again, I apologize for the delay. I spontaneously extended my vacation and after coming back my backlog was so huge I had to prioritize other things until I reached this PR.

crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
spigaz: I have no problem with that, but you will have to guide me on the best way to do that.

It wasn't my intention, I though it was working, so I added a test for it, and then it needed two fixes without them it wasn't working for sure, as you can test.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
spigaz: I believe I added it to be able to execute the tests on the browser without having require the runtime in the code, that way you can check if your assembly runs in all runtimes without having to recompile it.

crates/cli/Cargo.toml
spigaz: I don't remember, I'm going to review that.

crates/test-macro/src/lib.rs
spigaz: Yes, this was a left over, that I didn't clean up, sorry.

spigaz:
@daxpedda That's okay, the only issue is that the details aren't as fresh as they were.

I already improved the tests a lot, and I hope I'm able to improve them a bit more still, my objective is to make them more clear and intuitive and easier to change.

I haven't pushed those changed yet, because there are some things remaining in the macro.

I have no issues with the naming, I just use a naming from BDD, but different perspectives always enrich the solution.

I'm going to finish those, once I push that, I'm going to tackle your requests one by one.

crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
daxpedda: This is a great feature, but still belongs in a separate PR.

spigaz: This is the same issue with the deno fixes, I didn't planned for them, both had the same underlying objectives:
- add tests to make sure the application remained working as close as possible to the original, because of the docopt and nextest implied changes
- enforce that the nextest api is implemented consistently across all the supported runtimes

That being said, I actually do agree with you, but please advise me the best way to do this.
If I open PRs for them, once they are approved and merged, should a rebase to the latest version work?

daxpedda: I see ... I guess the issue is that running the tests on all targets would be more complicated without this change.
I'm fine with leaving it in this PR in this case, but it needs some added documentation.

crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
daxpedda: Deno is different then the browser issue, the main reason why I don't want it in this PR is because I can't review it.
So I'm fine with proceeding here simply without testing Deno.

spigaz:
Just a quick update, I was able to move forward with the multi_target_test model, I'm still flushing out details in the library as I convert the existing tests I created.

But its pretty much working, right now it already runs on all the runtimes that we specify, the problem is that its flushing issues and inconsistencies specific as expected.

One of the problems was that the browser was adding extra tabs to the output (I used cargo test with a #[test] as reference), its fixed now.

The other issue I'm having is some instability in the Safari Driver, unfixed for now.

But aside from that, the model seems to work well to ensure there is a consistent behaviour across all runtimes, but I should know more as I move forward.

I'll try to finish this part ASAP, but the blockers I believe will be only the issues that it will uncover.

daxpedda:
> The other issue I'm having is some instability in the Safari Driver, unfixed for now.

FWIW, I've been having spurious failures using Safaridriver lately as well, so far the only error I encountered is a port binding issue. My guess here is that the new ARM MacOS runners have some issue with binding random ports in quick succession or an issue around that with the driver itself.

spigaz:
> > The other issue I'm having is some instability in the Safari Driver, unfixed for now.
> 
> FWIW, I've been having spurious failures using Safaridriver lately as well, so far the only error I encountered is a port binding issue. My guess here is that the new ARM MacOS runners have some issue with binding random ports in quick succession or an issue around that with the driver itself.

In my case, the problem seems to be related to the proper termination of safaridriver, after the execution the instances remain open.

If I don't terminate manually the browser instances from previous executions, I get a 500 when it tries to get the session id.

```rust
let mut x: Result<Response, Error> = self.post("/session", &request);
while x.is_err() {
    x = self.post("/session", &request);
}
let x = x?;
```

But if I change the code to this, it hangs till I close the instances from the previous execution.

As the tests I'm running, run several instances of wasm-bindgen-test-runner, they hang, but if I terminate manually the pending instances, all tests end properly.

So now I'm trying to understand how to properly terminate the driver.

spigaz:
> > The other issue I'm having is some instability in the Safari Driver, unfixed for now.
> 
> FWIW, I've been having spurious failures using Safaridriver lately as well, so far the only error I encountered is a port binding issue. My guess here is that the new ARM MacOS runners have some issue with binding random ports in quick succession or an issue around that with the driver itself.

I was able to trace down the issue I was having, it was caused by multiple parallel safari invocations, once I added a lock to only allow one at a time, the problem was gone.

That lock doesn't limit test execution in practice, as Safari by design only allows one session at a time
https://developer.apple.com/documentation/webkit/about_webdriver_for_safari#2957226

The issue seems to be triggered by each safaridriver triggering the execution of a Safari --automation, and for some reason when that happens, even with workarounds, they have to be manually terminated for the execution to continue.

This solution seems to be cleaner, but the Safari --automation remains in memory, as it did before.

I removed most experiments, but left one, I moved the BackgroundChild into inside the Client, I did it because I was having some issues with the drop order, but left it in, because the code was cleaner, not sure if those issues remain.

I'm going to add some more safari tests to stress it more, then I'll add a parallel PR like I did with deno.

spigaz:
> > > The other issue I'm having is some instability in the Safari Driver, unfixed for now.
> > 
> > 
> > FWIW, I've been having spurious failures using Safaridriver lately as well, so far the only error I encountered is a port binding issue. My guess here is that the new ARM MacOS runners have some issue with binding random ports in quick succession or an issue around that with the driver itself.
> 
> I'm going to add some more safari tests to stress it more, then I'll add a parallel PR like I did with deno.

I ended up still achieving some more instability, that led me to beef up the lock, it seems pretty robust now.

But sometimes Safari --automation still refuses to allow the creation of a new session, so in those situations, I updated it to kill it, and ask safaridriver again for a new session, forcing it to initiate it again, that seems to fix the remaining issues.

I tested terminating the Safari --automation for every wasm-bindgen-test-runner execution, but that imposes a severe delay in the tests execution overall.

Right now, as I converted more tests into the multi_target_test mode, I'm dealing with some inconsistencies of output between the targets, but the Safari instability seems at least a lot better.

daxpedda:
@spigaz I believe you are spending quite a significant amount of time on testing facilities that don't have to be this perfect compared to all the other untested features `wasm-bindgen` contains (which is unfortunate of course). While this is **really** appreciated, I would prefer to split efforts into multiple PRs instead of trying to do everything at once.

spigaz:
@daxpedda That wasn't my objective, but nextest implied changes in many different places, including in the arguments handling.

As a rule I try to add tests before introducing changes, to make sure I don't make it worse.

Because of that and the nextest tests I ended up with a lot of tests, that when I generalised found even more issues that I tried to fix. 

Right now I'm just trying to iron out some issues with safari to avoid breaking the CI, it seems okay already, perhaps some starvation in the safari instance access.

But my point is to make it like we did with the deno PR, using the test base as reference.

daxpedda:
The testing you are trying to implement here **greatly** appreciated, but it is of way higher quality then this repository is used to. E.g. there are a lot of features that aren't tested for every browser and every target.

What I'm trying to say is that bringing the PR to a mergable state could be way simpler if the tests are in a comparable scope to the rest of the repository. Extending and improving tests can be done in a separate PR.

Ofc I will leave the decision on how to proceed here to you, I'm certainly not against getting this well tested in the same PR!

spigaz:
@daxpedda I understand your reasoning, and to be honest, I'm trying to narrow things to make it easier for you to review.

The problem is that cargo nextest stresses wasm-bindgen-test a lot, in my repo, 771 times, that means that all the instability issues pop up:
- [x] Deno, Safari, Firefox - seem to be running just fine
- [ ] Chrome and Edge fail sometimes, about once in each run
- [ ] Node - I don't know, I have the ES module issues because of local node config

Anyway, I was finally able to remove the hacks I added to get cargo nextest working, because of the shared directory, using a ResourceCoordinator.

I just have some minor things, then I'm going to let you take the lead on this and I can create as much PRs as necessary.

ifiokjr:
This looks great! Has it stalled?

spigaz:
> This looks great! Has it stalled?

@ifiokjr No, its done. Although now I have some merge conflicts to solve.

I just have been working on a way to get the tests to be more intuitive, to see if it eases the merging.

I just didn't pushed it, because it didn't seem to be a priority for anyone and I would break the existing tests, but if its a priority for you, I can try to speed up things...

daxpedda:
Done in #4356.

PR #3930, opened 2024-04-19
Fixes #3929 Allow exporting functions named "default".

### LINKED ISSUE #3929: Exports with `#[wasm_bindgen(js_name = default)]` gets exported as `fn _default`
### Describe the Bug

I'm trying to create a plugin for the javascript project [EventCatalog ](https://www.eventcatalog.dev/) in Rust with wasm_bindgen.
Their plugin system is fairly simple and it works by:

1. Adding your plugin as a nodejs package.
2. Place the name of the added package in a config file.
3. When the EventCatalog gets build, it `require` your package/plugin that's defined in the config file and calls the default exported function, i.e. whats exported with `export default ...` in the main module of the added nodejs package. For the curios you can find the code for the plugin loading code [here](https://github.com/boyney123/eventcatalog/blob/a3607d8b41af2553a213c6bc7415d35787f63a09/packages/eventcatalog/scripts/generate.js#L4).

I tried to create my plugin with the following code:
```rust
pub fn init_my_plugin() {
    ...
}
```

Unfortunately the above function gets exported as `_default` instead of `default` and the generated JS code (with `--target nodejs`) looks like this:
```js
module.exports._default = function() {
    wasm._default();
};
```
which means that `EventCatalog` fails to call my plugin. This issue is very much related to #3012 and is just the other side of the coin. I think it's reasonable that if you can import `fn default`, you should also be able to export `fn default`.

### Expected Behavior

The following Rust function
```rust
pub fn init_my_plugin() {
    ...
}
```
should be exported as `default`and not `_default`and the generated javascript code should be:
```js
module.exports.default = function() {
    wasm.default();
};
```

### Additional Context

I'll create a PR soon (PR created #3930)

nicklaswj:
This PR fixes #3929.
This PR adds the keyword `default`to the argument `skip_keywords` for `fn function_from_decl` when exporting functions.

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -327,6 +327,13 @@ impl Bindgen {
                 .context("failed getting Wasm module")?,
         };
 
+        // Check that no exported symbol is called "default" if we target web.
+        if matches!(self.mode, OutputMode::Web)
+            && module.exports.iter().any(|export| export.name == "default")
+        {
+            bail!("exported symbol \"default\" not allowed for --target web")
+        }
+
         let thread_count = self
             .threads
             .run(&mut module)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 
 ### Added
 
+* Allow exporting functions named `default`. Throw error in wasm-bindgen-cli if --target web and
+  an exported symbol is named `default`.
+  [#3930](https://github.com/rustwasm/wasm-bindgen/pull/3930)
+
 * Added support for arbitrary expressions when using `#[wasm_bindgen(typescript_custom_section)]`.
   [#3901](https://github.com/rustwasm/wasm-bindgen/pull/3901)
 
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -799,7 +799,7 @@ impl ConvertToAst<BindgenAttrs> for syn::ItemFn {
             false,
             None,
             false,
-            None,
+            Some(&["default"]),
         )?;
         attrs.check_used();
         Ok(ret.0)


daxpedda (changes requested):
Missing a changelog entry.

I believe that this should be safe to add. @Liamolucko would you mind double-checking me on this?

Liamolucko:
> I believe that this should be safe to add. @Liamolucko would you mind double-checking me on this?

This doesn't work with `--target web`, since the default export is already used for `init`. But it works with all other targets.

I don't particularly like adding target-compatibility caveats like this, but I also can't really imagine a scenario where it's an issue - discovering that you aren't allowed to set the default export on `--target web` isn't really any weirder than discovering that it got silently renamed to `_default`.

If we do go ahead with this it'd be nice if the CLI printed a proper error when trying to create a default export on the web target, rather than delaying it until the JS is actually used and throws a 'duplicate default export' error.

nicklaswj:
> If we do go ahead with this it'd be nice if the CLI printed a proper error when trying to create a default export on the web target, rather than delaying it until the JS is actually used and throws a 'duplicate default export' error.

I think that sounds reasonable enough. I'll try to extend my PR with a check for `--target web`and `#[wasm_bindgen(js_name = default)]`

nicklaswj:
I've added a check that throws an error if wasm-bindgen-cli is called with --target web and there exist a exported symbol called "default". I'm notoriously bad at creating good error messages, so feedback is welcome 😆

nicklaswj:
@daxpedda @Liamolucko Not to be pushy, but is there a chance I can get a new review ?

daxpedda:
I agree with whats said in https://github.com/rustwasm/wasm-bindgen/pull/3930#issuecomment-2067918594, I didn't know that this can't work with the Web target and I think its not the best idea to introduce such target-incompatible changes.

In hindsight I believe that https://github.com/rustwasm/wasm-bindgen/pull/2855 should not have simply renamed functions, but instead thrown an error.

So I currently think the way forward is to instead of renaming exports where it would make a semantic difference, export them with the `export { _default as 'default' }` syntax (inspired by https://github.com/rustwasm/wasm-bindgen/pull/2855#discussion_r1191380905) or e.g. in the case of the no-modules target can export as-is. Where this is not possible it should return an error.

Let me know if this doesn't address your problem @nicklaswj.
Apologies for the delay, I was on vacation.

daxpedda:
The point is to make exporting `default` possible on all targets.

In this case the issue was that `default` is a keyword in ES modules, which prevented this PR from making a fix for both targets. My suggested solution is to use the `export { _default as 'default' }` syntax to work around that limitation for the Web target.

This work around could be applied to other keywords as well, therefore a fix should attempt to craft a general solution.

Let me know if anything else is unclear.

Liamolucko:
> In hindsight I believe that #2855 should not have simply renamed functions, but instead thrown an error.

I agree.

> In this case the issue was that `default` is a keyword in ES modules, which prevented this PR from making a fix for both targets. My suggested solution is to use the `export { _default as 'default' }` syntax to work around that limitation for the Web target.

The PR already ends up doing that though: more specifically `export { default1 as default }` (turns out it still works without the quotes). It seems to be because of some old code [here](https://github.com/rustwasm/wasm-bindgen/blob/9b37613bbab0cd71f55dcc782d59dd00c1d4d200/crates/cli-support/src/js/mod.rs#L3904-L3910) that special-cases `default` (but it seems like it was written to allow importing `default`, not exporting it).

The issue is that the default export is reserved for `init` on the web target, and so it can't be used to export anything else. It's also literally just an export named 'default' with a bit of syntax sugar around it, so there's no way to have both a default export and an export named 'default' - they're the same thing.

nicklaswj:
I do not really have an opinion on how we solve it, however it is blocking me. @Liamolucko, if I understand your latest comment correct, I feel like we haven't really reached a conclusion whether I should change anything in my PR or not - is that correctly understood?

Liamolucko:
> @Liamolucko, if I understand your latest comment correct, I feel like we haven't really reached a conclusion whether I should change anything in my PR or not - is that correctly understood?

Well, there are 2 options:

1. Only error if `default` is exported on the web target (current state of this PR)
2. Error regardless for uniformity

At this point I'm leaning towards option 1 (how often do people really switch targets anyway?), but I was waiting to see what @daxpedda's thoughts were.

> however it is blocking me.

In the meantime, how important is it that the default export comes directly out of wasm-bindgen? Is there anything stopping you from making a JS wrapper that reexports it as `default`?

daxpedda:
> The issue is that the default export is reserved for `init` on the web target, and so it can't be used to export anything else. It's also literally just an export named 'default' with a bit of syntax sugar around it, so there's no way to have both a default export and an export named 'default' - they're the same thing.

Well that's unfortunate, and here I thought I found a solution that could work.

> Well, there are 2 options:
> 1. Only error if `default` is exported on the web target (current state of this PR)
> 2. Error regardless for uniformity
> 
> At this point I'm leaning towards option 1 (how often do people really switch targets anyway?), ...

Yeah I think I'm fine with proceeding with the PR then. A solution that came to mind is to add an attribute to explicitly opt-in to the target incompatibility, but I think that's overkill.

crates/cli-support/src/lib.rs
```diff
@@ -327,6 +327,13 @@ impl Bindgen {
                 .context("failed getting Wasm module")?,
         };
 
+        // Check that no exported symbol is called "default" if we target web.
+        if matches!(self.mode, OutputMode::Web)
```
daxpedda, line 331: I'm unfamiliar with the other targets, but does this really only affect the `web` target?

nicklaswj: As far as I can see, then yes. I tried to compile a dummy crate for all targets and grep'ing for default in each of the generated JS files. I only found default to be present in `web`.

daxpedda: My question was if the Web target is the only target having issues with the identifier `default`.

If the other targets didn't contain `default`, it would mean they have an issue exporting it. Web shouldn't contain `default` because you made it error out.

So I'm a bit confused about whats going on.

nicklaswj: No you misunderstood me. What I did to sanity check this was to create a new Rust project that does not export any symbol called `default`, e.g.  lib.rs looks something like this:
```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn greet() {
    alert("Hello, wasm-target-test!");
}
```
then for each target `bundler, nodejs, web, no-modules` I checked that each generated JS file didn't contain or use the symbol `default`or the keyword `export default <foo>`. Which is true for every target except `web`

Liamolucko (approved):
Thanks!

PR #3932, opened 2024-04-24
20240424 webauthn idls

Firstyear:
Update Webauthn IDLs to Level 3


daxpedda:
Checking the [spec](https://www.w3.org/TR/webauthn-3/) I can see that its still in the draft stage. We usually make those APIs unstable.

We made exceptions in the past for APIs that are unlikely to change e.g. because they are so [widely used](https://chromestatus.com/metrics/feature/popularity) that they can't change without breaking the web, which I assume is not the case here.

Firstyear:
They are pretty unlikely to change though, since a lot of is it based on webauthn-l2. Regardless, we can just wait I guess?

daxpedda:
> They are pretty unlikely to change though, since a lot of is it based on webauthn-l2.

Just looking at the repo it seems to me that there are still a lot of small details up for change.

> Regardless, we can just wait I guess? 

Moving it behind unstable is what we usually do.

Firstyear:
What's the right way to mark this "unstable" then?

daxpedda:
Just move the WebIDL files from the `enabled` folder into the `unstable` one and run the generator again.

Let me know if you run into any problems.

daxpedda:
Replaced by #4000.

Issue #3921, opened 2024-04-10
Missing WebIDL processing features overview


MOZGIII:
I was doing a manual implementation for the WebTransport types, and here's case-study-based a list of missing features that makes WebIDL codegen really unusable:

1. Lack of getters support for dictionaries. A lot of dictionaries are used in return positions: `ReadableStream`, pretty much everywhere in the `getStats` for WebTransport, etc.
1. Dictionaries setters do not allow unsetting values (for the most part) - we might want to either make them all take `Option<T>` instead of `T`, or generate an additional accessor to allow `unset_...`.
1. Untyped async calls - the WebIDL has enough data to properly deduce the type, but that is not done, and an untype `js_sys::Promise` it returned instead. We should consider adding types support to the `js_sys::Promise`, and also generating the calls that return promises as real `async fn`s. This is, however, blocked by https://github.com/rustwasm/wasm-bindgen/pull/3919, as `async` fn generation is broken.
1. Well, https://github.com/rustwasm/wasm-bindgen/pull/3919. The `async` fn generation is broken.
1. The `catch` is not generated correctly. This is an issue with WebIDL definitions, as they lack the machine-readable data on whether a call can throw or not - but nonetheless I count this towards the list of issues of the WebIDL-generated rust code.
1. `ReadableStream`/`WriteableStream` should have a generic indicating what type they produce - they are capable of streaming not just the bytes (i.e. binary data, where they act like Read/Write), but objects too (acting more like a stream / async iterator in rust sense).
1. Inability to turn certain hot calls `final` at the codegen level.
1. I can't just install the latest `wasm_bindgen_webidl` as a dependency and use it in my crate - as the latest version is not published, and the last published is very old and conflicts with the rest of the crates. This is by far the easiest issue to solve.

Also, some nits I encountered while implementing things manually.
1. `missing_docs` lint is triggered for `wasm_bindgen` macro enums.
1. No explicit support for attributes as `wasm_bindgen` macro, as in syntax sugar for generating both getter and setter from the same attribute.
1. Declaring dictionaries manually is too verbose, they should really be a first-class item in the macro.

Not sure this is the right place for this, and those issue are most definitely reported before - but this is a case study organized in one place. Hopefully if all those are solved somehow we'd get a way nicer codegen.
For now, using WebIDL is much more painful than using custom type declarations.

pablosichert:
[...]

Issue #2921, opened 2022-06-01
web_sys: Dictionaries don't have getters


rozbb:
### Describe the Bug

`dictionary` types from WebIDL do not have getters. You can only mutate dictionaries, and never read them. This is weird, considering how they're often used as event types or return types.

### Steps to Reproduce

Example: [`AudioDecoderSupport`](https://docs.rs/web-sys/0.3.57/web_sys/struct.AudioDecoderSupport.html) has no getters, despite being [defined](https://w3c.github.io/webcodecs/#config-support-info) as a return type to indicate to the caller which formats are supported. defined here  has no getters 

### Expected Behavior

`config(&self) -> AudioDecoderConfig` should be a getter

`set_config(&mut self, val: &AudioDecoderConfig)` should be a setter

### Actual Behavior

`config(&mut self, val: &AudioDecoderConfig)` is the only function available regarding the underlying `config` field, and it is a setter.

### Additional Context

[This](https://github.com/rustwasm/wasm-bindgen/blob/cbcf3e78ac1e17884cceb855fced173ddde37020/crates/webidl/src/generator.rs#L658-L673=) is the generating code. I've dumped several hours into modifying it to generate a setter, but I've failed so far. My main issue is **how do you generically convert the a `JsValue` from `js_sys::Reflect::get` into the appropriate return type?** There are lots of special cases that work but it's unclear how to do this generically. E.g., how do I turn a `JsValue` into an `Option<i32>`?

Very stumped here, but willing to do work to make a PR. Please comment with pointers and I'll be on it.

daxpedda:
Duplicate of #1793.

> My main issue is **how do you generically convert the a `JsValue` from `js_sys::Reflect::get` into the appropriate return type?** There are lots of special cases that work but it's unclear how to do this generically. E.g., how do I turn a `JsValue` into an `Option<i32>`?

Probably through [`JsValue::as_f64()`](https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html#method.as_f64), but this shouldn't be a concern anymore since #3898.

PR #3933, opened 2024-04-24
Implement getters for WebIDL dictionaries

### LINKED ISSUE #1793: Add getters for WebIDL's dictionaries.

pablosichert:
Heyo,

this PR implements getters for WebIDL dictionaries. The diff is bit noisy due to the autogenerated files; you can look at the actual changes to the code generation in the last two files `crates/webidl/src/generator.rs` and `crates/webidl/src/lib.rs` and corresponding tests in `crates/webidl-tests/dictionary.rs`.

Currently, setters are implemented as `my_property(&mut self, val: ...)`. To avoid breaking backwards compatibility, I have generated a new trait `MyDictionaryTypeGetters` for every `MyDictionaryType`. It contains definitions in the form of `my_property(&self) -> ...`.

Personally, I would suggest to rename all setters to `set_my_property` and place the `my_property` getters back in the `impl MyDictionaryType` as conventionally used in Rust. Preferably in another PR since as already mentioned would break backwards compatibility.

For the dictionary getter return types I have applied the same logic that is currently used for Interface getters:

https://github.com/rustwasm/wasm-bindgen/blob/e78db23553e70a48767a9bafaac3a4ffb4cba08d/crates/webidl/src/lib.rs#L688-L692

https://github.com/rustwasm/wasm-bindgen/blob/e78db23553e70a48767a9bafaac3a4ffb4cba08d/crates/webidl/src/generator.rs#L298-L300

This should close #1793 and #2921, and address point 1 of #3921.


crates/webidl/src/generator.rs
```diff
@@ -815,6 +930,17 @@ impl Dictionary {
...
+            #unstable_attr
+            #getter_trait_doc_comment
+            pub trait #getter_trait_name {
+                #(#getter_trait_fields)*
+            }
+
+            #unstable_attr
+            impl #getter_trait_name for #name {
+                #(#getter_trait_impl_fields)*
+            }
```
pablosichert, line 942: That's the main addition of this PR...

crates/webidl/src/generator.rs
```diff
@@ -654,14 +656,97 @@ impl DictionaryField {
...
+        let Some(return_ty) = return_ty else {
+            return None;
+        };
+
+        quote! {
+            #unstable_attr
+            #cfg_features
+            #doc_comment
+            #unstable_docs
+            fn #name(&self) -> #return_ty;
```
pablosichert, line 713: ...with getter signatures that are defined here...

crates/webidl/src/generator.rs
```diff
@@ -654,14 +656,97 @@ impl DictionaryField {
...
+            return None;
+        };
+        let shim_name = self.shim_name();
+
+        quote! {
+            #unstable_attr
+            #cfg_features
+            fn #name(&self) -> #return_ty {
+                self.#shim_name()
+            }
```
pablosichert, line 744: ...and implemented here

MOZGIII:
> Personally, I would suggest to rename all setters to `set_my_property` and place the `my_property` getters back in the `impl MyDictionaryType` as conventionally used in Rust. Preferably in another PR since as already mentioned would break backwards compatibility.

I agree. We should bite the bullet, break the compatibility and do it (in another PR ofc).

MOZGIII:
I like the idea of having traits here. We could probably go even further - have a trait with a WebIDL-generated declaration of the whole interface of a given type - not just for getters. That way we can include getters, setters and methods (both static and not).
The possible benefit from this would be nicer composability of wasm types.

I think we could create an issue do discuss this further, if this sounds like a viable direction to explore.

pablosichert:
> The possible benefit from this would be nicer composability of wasm types.

I think you are referring to what I have an mind, but just to elaborate the thought: I'm not fully sure yet how hard it would be to make it possible, but if that would allow e.g. implementing `FooGetters` for `Bar` where `Bar: Foo`, that would be great. Currently you need to explicitly downcast the type.

crates/webidl-tests/dictionary.rs
```diff
@@ -56,3 +56,30 @@ fn required() {
...
+        use crate::generated::ManyTypesGetters;
+        assert_eq!(many_types.a(), "a");
+        assert_eq!(many_types.n1(), 1);
+        assert_eq!(many_types.n2(), 2);
+        assert_eq!(many_types.n3(), 3);
+        assert_eq!(many_types.n4(), 4);
+        assert_eq!(many_types.n5(), 5);
+        assert_eq!(many_types.n6(), 6);
+        assert_eq!(many_types.c(), other_dict);
+    }
```
pablosichert, line 84: Unfortunately, due to the signature name clash this is a bit awkward to use right now.

If you move the `use crate::generated::ManyTypesGetters;` to the parent block, this test breaks.

MOZGIII:
Let's take a [`WritableStream`](https://streams.spec.whatwg.org/#writablestream) for example.

It it was a trait, could just implement it for a `JsValue`, like so: `impl WritableStream for JsValue { ... }` (with some codegen of course).

Then, if needed, we could very elegantly extend `WritableStream` with [`WebTransportSendStream`](https://w3c.github.io/webtransport/#webtransportsendstream) with an explicit `trait WebTransportSendStream: WritableStream { ... }` - and also implement the `WebTransportSendStream` for `JsValue`.

---

<details><summary>WebIDL</summary>
<p>

```webidl
[Exposed=*, Transferable]
interface WritableStream {
  constructor(optional object underlyingSink, optional QueuingStrategy strategy = {});

  readonly attribute boolean locked;

  Promise<undefined> abort(optional any reason);
  Promise<undefined> close();
  WritableStreamDefaultWriter getWriter();
};
```

```webidl
[Exposed=(Window,Worker), SecureContext, Transferable]
interface WebTransportSendStream : WritableStream {
  attribute WebTransportSendGroup? sendGroup;
  attribute long long sendOrder;
  Promise<WebTransportSendStreamStats> getStats();
  WebTransportWriter getWriter();
};
```

</p>
</details> 

The traits would look somewhat like this:

<details><summary>Traits</summary>
<p>

```rust
trait WritableStream: Sized + JsAny {
    fn new() -> Self;
    fn new_with_underlying_sink(underlying_sink: &impl JsObject) -> Self;
    // ...

    fn locked(&self) -> bool;

    fn abort(&self) -> Promise<()>;
    fn abort_with_reason(&self, reason: &impl JsAny) -> Promise<()>;

    fn close(&self) -> Promise<()>;

    fn get_writer(&self) -> Result<impl WritableStreamDefaultWriter>;
}

trait WebTransportSendStream: WritableStream + JsAny {
    fn send_group(&self) -> Option<impl WebTransportSendGroup>;
    fn set_send_group(&self, value: &impl WebTransportSendGroup) -> Option<impl WebTransportSendGroup>;
    fn unset_send_group(&self);

    fn send_order(&self) -> i64;
    fn set_send_order(&self, value: i64);
    fn unset_send_order(&self);
    
    fn get_stats(&self) -> Promise<impl WebTransportSendStreamStats>;

    fn get_writer(&self) -> Result<impl WebTransportWriter>;
}
```

</p>
</details> 

and impls like this:


<details><summary>Impls</summary>
<p>

```rust
impl WritableStream for JsValue {
    fn new() -> Self {
        __shim_WritableStream_new()
    }

    fn new_with_underlying_sink(underlying_sink: &impl JsObject) -> Self {
        __shim_WritableStream_new_with_underlying_sink(underlying_sink)
    }

    // ... and so on.
}

impl WebTransportSendStream for JsValue {
    fn send_group(&self) -> Option<impl WebTransportSendGroup> {
        __shim_WebTransportSendStream_send_group()
    }
}
```

</p>
</details>

pablosichert:
Cool, tests pass now.

Just a note that this will break for people that use `web_sys::*` in their code, but I guess they should be aware that they are opening up to that danger by using a wildcard.

daxpedda:
I think that the easiest way forward would be to just implement getters as `get_my_property()`.

Obviously this isn't Rusty, but I believe its better then introducing a trait for each type while still remaining backwards compatible.

@Liamolucko would appreciate your input as well.

Liamolucko:
> I think that the easiest way forward would be to just implement getters as `get_my_property()`.

I agree; I think the messiness of adding the extra traits outweighs the benefit of the better names.

I'm also not really comfortable with the potential breakage of `use web_sys::*`.

pablosichert:
Do you keep track of changes that you'd want to include in a future breaking change? Since arguably this would be a great candidate, and the previous naming does not follow common Rust guidelines.

I can't imagine anyone complaining that their setters are now called `set_*` over of the current naming usually reserved to getters.

Even though I disagree with the naming, happy to make the necessary changes to `get_*` if this is what you'd prefer to merge in.

daxpedda:
> Do you keep track of changes that you'd want to include in a future breaking change?

We don't, mainly because we don't see a breaking change on the horizon anytime soon.

> Even though I disagree with the naming, happy to make the necessary changes to `get_*` if this is what you'd prefer to merge in.

Indeed that's how we would like to proceed here.

MOZGIII:
I was thinking about this, and there is a breaking change that I'd want to merge in with the async fs - that one didn't get any attention for a while not...

So, maybe it is time to fork? It would not be too bad to have an experimental fork, especially since most of the code is generated. I'd very much welcome the changes in here, and my use cases really really don't care about the breaking changes - as long as the underlying objects and be converted back and forth between the `JSValue` types of the fork and the upstream - that would be fine.

I can spend some resources setting up the repos and etc, but it would be great to collab with someone and formulate the needs for the fork upfront - like having more frequent breaking releases to allow advancement.

daxpedda:
Replaced with #3993.

PR #3938, opened 2024-04-26
Update dispatch_event to take Event or CustomEvent

SorenHolstHansen:
This updates the dispatch_event method on EventTarget to take either Event or CustomEvent


Liamolucko:
`CustomEvent` is a subclass of `Event`, so this shouldn't be necessary. You can convert a `CustomEvent` into an `Event` via. [its `Deref<Target = Event>` impl](https://docs.rs/web-sys/latest/web_sys/struct.CustomEvent.html#deref-methods-Event).

This would also be a breaking change.

SorenHolstHansen:
> `CustomEvent` is a subclass of `Event`, so this shouldn't be necessary. You can convert a `CustomEvent` into an `Event` via. [its `Deref<Target = Event>` impl](https://docs.rs/web-sys/latest/web_sys/struct.CustomEvent.html#deref-methods-Event).
> 
> This would also be a breaking change.

Thanks, didn't know!

PR #3940, opened 2024-04-28
Fix Rust values getting GC'd while still borrowed and not getting GC'd if created via. constructor

### LINKED ISSUE #3854: FinalizationRegistry.register() call missing in exported rust types that define a constructor
### Describe the Bug

`FinalizationRegistry.register()` call missing in exported rust types that define a constructor, preventing cleanup.

### Steps to Reproduce

1. Add this rust code to one of the wasm-bindgen tests, e.g. `test/api.rs`:

```rust
#[wasm_bindgen]
pub struct ObjectWithConstructor {}

#[wasm_bindgen]
impl ObjectWithConstructor {
    #[wasm_bindgen(constructor)]
    pub fn new() -> Self {
        Self {}
    }
}
```

2. Run tests to generate JS code

3. View generated JS code:

```javascript
const ObjectWithConstructorFinalization = (typeof FinalizationRegistry === 'undefined')
    : new FinalizationRegistry(ptr => wasm.__wbg_objectwithconstructor_free(ptr >>> 0));
/**
*/
class ObjectWithConstructor {

    __destroy_into_raw() {
        const ptr = this.__wbg_ptr;
        this.__wbg_ptr = 0;
        ObjectWithConstructorFinalization.unregister(this);
        return ptr;
    }

    free() {
        const ptr = this.__destroy_into_raw();
        wasm.__wbg_objectwithconstructor_free(ptr);
    }
    /**
    */
    constructor() {
        const ret = wasm.objectwithconstructor_new();
        this.__wbg_ptr = ret >>> 0;
        return this;
    }
}
module.exports.ObjectWithConstructor = ObjectWithConstructor;
```

### Expected Behavior

I expect there to be a line in the constructor to register, e.g. `ObjectWithConstructorFinalization.register(this, this.__wbg_ptr, this);`

### Actual Behavior

In the `ObjectWithConstructor` code, there is never a call to register the object destruction callback. 

If we compare to this object, which does not define a constructor:

```rust
#[wasm_bindgen]
pub struct ObjectNoConstructor {}

#[wasm_bindgen]
impl ObjectNoConstructor {
    pub fn new() -> Self {
        Self {}
    }
}
```

It has a `register` call in the `__wrap` function:

```javascript
const ObjectNoConstructorFinalization = (typeof FinalizationRegistry === 'undefined')
    : new FinalizationRegistry(ptr => wasm.__wbg_objectnoconstructor_free(ptr >>> 0));
/**
*/
class ObjectNoConstructor {

    constructor() {
        throw new Error('cannot invoke `new` directly');
    }

    static __wrap(ptr) {
        ptr = ptr >>> 0;
        const obj = Object.create(ObjectNoConstructor.prototype);
        obj.__wbg_ptr = ptr;
        ObjectNoConstructorFinalization.register(obj, obj.__wbg_ptr, obj);
        return obj;
    }

    __destroy_into_raw() {
        const ptr = this.__wbg_ptr;
        this.__wbg_ptr = 0;
        ObjectNoConstructorFinalization.unregister(this);
        return ptr;
    }

    free() {
        const ptr = this.__destroy_into_raw();
        wasm.__wbg_objectnoconstructor_free(ptr);
    }
    /**
    * @returns {ObjectNoConstructor}
    */
    static new() {
        const ret = wasm.objectnoconstructor_new();
        return ObjectNoConstructor.__wrap(ret);
    }
}
module.exports.ObjectNoConstructor = ObjectNoConstructor;
```

### Additional Context

Should the register call be generated [here](https://github.com/rustwasm/wasm-bindgen/blob/0c09e154cf37b7e7f2308787d5186056cf85b533/crates/cli-support/src/js/binding.rs#L1021)?

Liamolucko:
Fixes #3854.
Supersedes #3743.

This implements the alternative reference-counting based fix that I proposed in #3743.

I also added some tests for it. It turns out there was already an existing test around GC, which meant that Node's `--expose-gc` flag was already enabled for manually triggering garbage collection.

I ended up merging it together with my tests, and then running them sequentially since attempting to run multiple GCs in parallel seemed to make it flaky. It was still somewhat flaky after that point, though: the garbage collector just decided not to clean up the `OwnedValue`s sometimes. For whatever reason, 'warming it up' by just running a bunch of pointless GCs before the tests seems to fix it; I might try and investigate why at some point, but for now it works.


daxpedda (approved):
LGTM!

I couldn't try it out locally because I'm on vacation right now.

PR #3941, opened 2024-04-29
Added getCapabilities() to RtcRtpReceiver

erik-moqvist:


Liamolucko (commented):
Thanks!

erik-moqvist:
Updated with added changelog entry.

PR #3978, opened 2024-06-06
Update `AudioBufferSourceNode` to latest W3C draft

pablosichert:
With this PR, the `AudioBufferSourceNode` WebIDL file is updated to the latest version of the W3C draft at https://webaudio.github.io/web-audio-api/#AudioBufferSourceNode.

Namely, the `grain_` prefix from the `offset` and `duration` input parameters is removed.

The removed `attribute EventHandler onended;` can be accessed through the `AudioScheduledSourceNode` base class.

### DIFF
--- a/crates/web-sys/src/features/gen_AudioBufferSourceNode.rs
+++ b/crates/web-sys/src/features/gen_AudioBufferSourceNode.rs
@@ -86,20 +86,6 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `AudioBufferSourceNode`*"]
     pub fn set_loop_end(this: &AudioBufferSourceNode, value: f64);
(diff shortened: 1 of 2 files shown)


daxpedda (changes requested):
Unfortunately this would be a breaking change, so we have to be a bit more careful.

1. Its fine just updating parameter names in WebIDL, but that also changes method names in the generated Rust code, which would be a breaking change. You can use [this](https://github.com/rustwasm/wasm-bindgen/blob/ad826519586351e1cafbd203e74e8db5b46a89a0/crates/webidl/src/constants.rs#L107) to adjust the names in the generator.
2. The removal of methods is fine as long as they actually got removed from browsers. Could you check the usual browsers? (you can also check their source code if that's easier)

A changelog entry has to be added as well.

daxpedda:
Replaced by #4020.

PR #3981, opened 2024-06-11
Add bindings for `queueMicrotask`

daxpedda:
Add [`queueMicrotask`](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask), pretty straightforward.

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask)
[WHATWG](https://html.spec.whatwg.org/#dom-queuemicrotask)


Liamolucko (approved):
LGTM!

PR #3989, opened 2024-06-17
Added experimental User-Agent Client Hints

maccesch:
Spec: https://wicg.github.io/ua-client-hints/
MDN: https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API


daxpedda (approved):
Thanks!

PR #3990, opened 2024-06-17
Fixed deno support.

spigaz:
To be honest I just tried to minimize changes as much as possible to get it working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,9 @@
 * Fix MDN links to static interface methods.
   [#4010](https://github.com/rustwasm/wasm-bindgen/pull/4010)
 
+* Fixed Deno support.
+  [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
+
 --------------------------------------------------------------------------------
 
 ## [0.2.92](https://github.com/rustwasm/wasm-bindgen/compare/0.2.91...0.2.92)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,9 +94,6 @@
 * Fixed Deno support.
   [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
 
-* Fixed Deno support.
-  [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
-
 --------------------------------------------------------------------------------
 
 ## [0.2.92](https://github.com/rustwasm/wasm-bindgen/compare/0.2.91...0.2.92)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,9 @@
 
 * Fix MDN links to static interface methods.
   [#4010](https://github.com/rustwasm/wasm-bindgen/pull/4010)
+ 
+* Fixed Deno support.
+  [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
 
 * Fixed Deno support.
   [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -345,7 +345,8 @@ impl<'a> Context<'a> {
             }}
 
             const wasmInstance = (await WebAssembly.instantiate(wasmCode, imports)).instance;
-            const wasm = wasmInstance.exports;",
+            const wasm = wasmInstance.exports;
+            export const __wasm = wasm;",
             module_name = module_name
         )
     }
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
@@ -24,7 +24,7 @@ pub fn execute(
         // `wasm-bindgen-test-runner` which forwards them to deno which we
         // forward to the test harness. this is basically only used for test
         // filters for now.
-        cx.args(Deno.args.slice(1));
+        cx.args(Deno.args);
 
         const tests = [];
 "#,
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
@@ -27,7 +27,7 @@ pub fn execute(
         cx.args(Deno.args);
 
         const tests = [];
-"#,
+    "#,
         module,
         console_override = SHARED_SETUP,
     );
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
@@ -18,7 +18,7 @@ pub fn execute(
 
         {console_override}
 
-        // global.__wbg_test_invoke = f => f();
+        window.__wbg_test_invoke = f => f();
 
         // Forward runtime arguments. These arguments are also arguments to the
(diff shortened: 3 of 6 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -345,7 +345,8 @@ impl<'a> Context<'a> {
             }}
 
             const wasmInstance = (await WebAssembly.instantiate(wasmCode, imports)).instance;
-            const wasm = wasmInstance.exports;",
+            const wasm = wasmInstance.exports;
+            export const __wasm = wasm;",
```
daxpedda, line 349: Note for reviewer: this is probably undesirable and at the very least should be limited to Deno.

daxpedda:
@Liamolucko is this something you are able to review?

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
```diff
@@ -30,7 +30,7 @@ wrap("info");
 wrap("warn");
 wrap("error");
 
-cx = new wasm.WasmBindgenTestContext();
+const cx = new wasm.WasmBindgenTestContext();
```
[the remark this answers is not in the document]
spigaz: No, that is required for deno, because deno implementation uses a part of the node implementation.

crates/cli-support/src/js/mod.rs
spigaz: It didn't broke node or any of the implementations I tested so far, so I wasn't sure if deno was just being more strict than the others.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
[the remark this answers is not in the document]
daxpedda: I think it would be good if you explain what you did exactly and why you did it in the OP or comments.

(maybe its more obvious for the actual reviewer)

crates/cli-support/src/js/mod.rs
daxpedda: I wasn't worried about breaking, I'm more worried about exporting anything, even if prefixed by `__`, unless its absolutely necessary. We don't want anyone starting to rely on it unless we intend to and we also don't want to occupy a slot unless necessary.

Again, there is no real issue here, we should just double-check if this is necessary.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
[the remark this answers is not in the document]
spigaz: To be honest I don't remember all the details.

But by looking at the diffs, I don't think I made any decisions, I tried to follow the original intentions, I just added missing pieces and changed the order of some code to get it working.

I agree its hard to understand because the errors surfaced mainly in the generated run.js file, in this case, I was getting a variable missing error.

On another the variable was used before it was declared for instance...

crates/cli-support/src/js/mod.rs
spigaz: I understand your concerns, the generated code requires this variable both on node and on deno
```js
const ok = await cx.run(tests.map(n => wasm.__wasm[n]))
```
For some reason, without the export it says its not declared on deno, perhaps because deno uses an import:
```js
import * as wasm from "./{0}.js";
```
while node still uses a require:
```js
const wasm = require("./{0}")
```
In practice it was being "exported" for node already, intentionally or not.

spigaz:
Thanks to the multi target tests I was able to isolate an issue with the deno handling of the arguments, I already updated it to this PR also.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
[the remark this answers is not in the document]
Liamolucko: I think the old version didn't work in Deno because it has [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) enabled by default, which doesn't allow implicitly declaring global variables like that.

crates/cli-support/src/js/mod.rs
Liamolucko: It seems like this was added for Node only in #2012: https://github.com/rustwasm/wasm-bindgen/blob/ad826519586351e1cafbd203e74e8db5b46a89a0/crates/cli-support/src/js/mod.rs#L278

I don't like it very much, but this isn't making things any worse and removing it from Node is probably out of scope for this PR.

Liamolucko (approved):
LGTM, thanks!

I would've asked you to test this in CI, but it seems like that's already covered by #3920 anyway.

crates/cli-support/src/js/mod.rs
spigaz: I'm not sure about the best approach either here or in other places, but it seems clear to me that some refactors are in order.

I don't think its just me, but when I'm just trying to fix issues, I try to avoid refactoring as much as possible to make the PR easier to review, but that strategy adds up, and its pretty clear that some refactors are in order.

Although, in this case and some others, I'm not sure the best approach is, as my point of view was quite narrow, focused on specific issues.

But I think opening issues with those changes, so that someone can pick up and provide a PR might be a very good idea.

spigaz:
@daxpedda Is there anything missing for this one to be merged?

crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
```diff
@@ -18,19 +18,16 @@ pub fn execute(
...
         // filters for now.
-        cx.args(Deno.args.slice(1));
-
-        const ok = await cx.run(tests.map(n => wasm.__wasm[n]));
-        if (!ok) Deno.exit(1);
+        cx.args(Deno.args);
 
         const tests = [];
-    "#,
+"#,
```
daxpedda, line 30: Why the space removal there?

CHANGELOG.md
```diff
@@ -90,6 +90,12 @@
 
 * Fix MDN links to static interface methods.
   [#4010](https://github.com/rustwasm/wasm-bindgen/pull/4010)
+ 
+* Fixed Deno support.
+  [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
+
+* Fixed Deno support.
+  [#3990](https://github.com/rustwasm/wasm-bindgen/pull/3990)
```
daxpedda, line 98:
```suggestion
```
Changelog entry was accidentally duplicated.

daxpedda (commented):
As mentioned before, I can't really review this.

But seeing that Liamolucko has already approved it, I will go ahead and merge it after the nits are remedied.

CHANGELOG.md
spigaz: I have to rebase twice, because conflicts on it kept showing up, although I was careful, not sure how that passed. Thanks!

crates/cli/src/bin/wasm-bindgen-test-runner/deno.rs
spigaz: The spacing isn't necessary for the generated code, so it gets cleaner, but its a cosmetic issue nevertheless. Thanks!

daxpedda: Would you mind removing it then?
The JS output isn't meant to be beautiful at the moment, that's a whole other can of worms.

spigaz: Sure, done.

PR #3992, opened 2024-06-18
Update and stablize the Clipboard API

### LINKED ISSUE #3991: Stabilize the Clipboard API
### Motivation

Currently the Clipboard API is in the unstable folder: https://github.com/rustwasm/wasm-bindgen/blob/main/crates/web-sys/webidls/unstable/Clipboard.webidl. Since Firefox 127 all browsers support this API in stable versions. There are still some difference like `unsanitized` where Safari/Firefox won't follow Chrome: https://github.com/w3c/clipboard-apis/issues/206#issue-2038753339.

Additionally this would be useful for Ruffle so we don't need to opt-in to unstable for https://github.com/ruffle-rs/ruffle/pull/16692.

### Proposed Solution

I am going to make it stable.

### Alternatives

n/a

### Additional Context

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility

evilpie:
Fixes #3991.


crates/web-sys/webidls/Clipboard.webidl
```diff
@@ -1,43 +1,29 @@
...
- * W3C Working Draft, 4 June 2021
+ * Editor’s Draft, 4 June 2024
+ *
  * The origin of this IDL file is:
- * https://www.w3.org/TR/2021/WD-clipboard-apis-20210604/
+ * https://w3c.github.io/clipboard-apis/#idl-index
  */
 
 partial interface Navigator {
   [SecureContext, SameObject] readonly attribute Clipboard? clipboard;
```
evilpie, line 11: I left this as optional `?`, but I am not sure if that is required. i.e. what happens in browser without `navigator.clipboard` ?

crates/web-sys/webidls/Clipboard.webidl
```diff
@@ -46,6 +32,20 @@ dictionary ClipboardItemOptions {
...
+
+[SecureContext, Exposed=Window]
+interface Clipboard : EventTarget {
+  Promise<ClipboardItems> read(optional ClipboardUnsanitizedFormats formats = {});
+  Promise<DOMString> readText();
+  Promise<undefined> write(ClipboardItems data);
+  Promise<undefined> writeText(DOMString data);
+};
+
+dictionary ClipboardUnsanitizedFormats {
```
evilpie, line 45: I wasn't sure if I should remove this, because it's Chrome-only.

crates/web-sys/webidls/Clipboard.webidl
daxpedda: It would panic, but this is problem we won't solve here, but would require a more general solution.

So please go ahead and remove it.

crates/web-sys/webidls/Clipboard.webidl
daxpedda: We generally tend to follow the spec regardless of browser supprt.

daxpedda (commented):
I don't believe we can stabilize this just yet. Not only is the spec still in draft, but the repo seems quite active as well. I couldn't find any usage data as well.

Do you have any more hints to this API not changing anymore? Preferably usage data, which usually prevents any breaking changes to the API.

crates/web-sys/webidls/Clipboard.webidl
evilpie: This is a bit of a special case, because this API was merged without consensus, but I don't want to relitigate this here :melting_face:

evilpie:
The base API for this is shipping in all three major engines, so I don't see how the current API could change.

crates/web-sys/webidls/Clipboard.webidl
daxpedda: Lets leave this part as unstable for sure then.

daxpedda:
> The base API for this is shipping in all three major engines, so I don't see how the current API could change.

Unfortunately this happens regularly and then we end up stuck with outdated APIs that we can't really fix without a breaking change.

Though I agree that the base API is unlikely to change any further, so what we can do is stabilize parts of it we are more certain then not aren't going to change.

Let me know if you need any help with that!

evilpie:
I've removed (commented out) everything that isn't supported by Chrome, Firefox and Safari currently.

crates/web-sys/src/features/gen_ClipboardItem.rs
```diff
@@ -34,4 +34,11 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `ClipboardItem`*"]
     pub fn get_type(this: &ClipboardItem, type_: &str) -> ::js_sys::Promise;
+    # [wasm_bindgen (static_method_of = ClipboardItem , js_class = "ClipboardItem" , js_name = supports)]
+    #[doc = "The `supports()` method."]
+    #[doc = ""]
+    #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/supports)"]
```
kjarosh, line 40: That link seems to be broken, shouldn't it be https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/supports_static ?

daxpedda: The link is generated, so that would have to be fixed in the generator.
It would be interesting to see if that is an exception for this specific static method or if this actually fixes a bug in the generator.

daxpedda: Addressed in #4010.

crates/web-sys/src/features/gen_Navigator.rs
```diff
@@ -12,6 +12,14 @@ extern "C" {
...
     #[doc = "*This API requires the following crate features to be activated: `Navigator`*"]
     pub type Navigator;
+    #[cfg(feature = "Clipboard")]
+    # [wasm_bindgen (structural , method , getter , js_class = "Navigator" , js_name = clipboard)]
+    #[doc = "Getter for the `clipboard` field of this object."]
+    #[doc = ""]
+    #[doc = "[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard)"]
+    #[doc = ""]
+    #[doc = "*This API requires the following crate features to be activated: `Clipboard`, `Navigator`*"]
+    pub fn clipboard(this: &Navigator) -> Clipboard;
```
alecmocatta, line 22: cc https://github.com/rustwasm/wasm-bindgen/issues/4085

PR #3997, opened 2024-06-22
Restrict to bumpalo v3.12.1

notgull:
This is the last version of bumpalo that compiles with wasm-bindgen's
stated MSRV of v1.57.

### DIFF
--- a/crates/backend/Cargo.toml
+++ b/crates/backend/Cargo.toml
@@ -17,7 +17,7 @@ spans = []
 extra-traits = ["syn/extra-traits"]
 
 [dependencies]
-bumpalo = "3.0.0"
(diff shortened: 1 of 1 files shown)


daxpedda:
I do understand where this is coming from and it affected crates I maintain myself and with others (Winit) as well, but I don't think its an option to limit the dependency for users who don't require the MSRV.

Unfortunately I'm not aware of any good fix for this except to wait for https://github.com/rust-lang/cargo/issues/9930, which obviously wouldn't solve this issue because it would only affect future Rust versions.

Let me know if you know a better way of handling this. I'm also open to debate the issue if you think I'm missing something here.

PR #3968, opened 2024-05-22
Added jitterBufferTarget to RtcRtpReceiver.

erik-moqvist:

PR #3998, opened 2024-06-23
Implement WebIDL `(Getter|Setter)Throws`

daxpedda:
I noticed in #3968 that we don't handle `GetterThrows` or `SetterThrows`, which are used to mark certain class members as potentially throwing but only for their respective getters or setters.

This would have been a breaking change so I went ahead and implemented a list of exclusions which we can remove on the next breaking change to `web-sys`. Apparently not a single unstable WebIDL was using those attributes, so actually no API changes were made.


Liamolucko (approved):
LGTM!

I'm guessing the lack of changelog entry is because no user-facing changes actually occurred?

daxpedda:
> I'm guessing the lack of changelog entry is because no user-facing changes actually occurred?

Indeed.

PR #3999, opened 2024-06-24
Added changelog entry instruction to web-sys readme

maccesch:


daxpedda (approved):
Documentation around these parts should be de-duplicated first probably, but until then this doesn't hurt.

PR #4005, opened 2024-07-04
Fix no-std compatibility

Systemcluster:
`no-std` support is currently non-functional. This PR does the following:

* Replace the unconditional `std` imports with `core` and `alloc` counterparts
* Remove most of the `if_std!` gates (`Box`, `String` and `Vec` are available in `alloc`)
* Allow exporting classes without `std`
* Add a `check --no-default-features` call to the check workflow

There are no behavioral changes. 

Note: I assume `wasm-bindgen` with `std` disabled was initially supposed to be alloc-free, but considering it hasn't been functional without `std` for a while at all, I don't think it's necessary to keep that limitation. If necessary, a separate `alloc` feature instead might make sense.

`externref`, `intern` and `take_last_exception`/`__wbindgen_exn_store` remain `std`-gated due to use of `std::thread_local!`.

### DIFF
--- a/src/externref.rs
+++ b/src/externref.rs
@@ -1,8 +1,9 @@
 use crate::JsValue;
-use std::cell::Cell;
-use std::slice;
-use std::vec::Vec;
-use std::cmp::max;
+
+use alloc::slice;
+use alloc::vec::Vec;
+use core::cell::Cell;
+use core::cmp::max;
 
 externs! {
     #[link(wasm_import_module = "__wbindgen_externref_xform__")]
@@ -98,10 +99,17 @@ impl Slab {
 }
 
 fn internal_error(msg: &str) -> ! {
-    if cfg!(debug_assertions) {
+    if cfg!(debug_assertions) || !cfg!(feature = "std") {
         super::throw_str(msg)
     } else {
-        std::process::abort()
+        #[cfg(feature = "std")]
+        {
+            std::process::abort()
+        }
+        #[cfg(not(feature = "std"))]
+        {
+            unreachable!()
+        }
     }
 }
 
--- a/src/externref.rs
+++ b/src/externref.rs
@@ -99,15 +99,17 @@ impl Slab {
 }
 
 fn internal_error(msg: &str) -> ! {
-    if cfg!(debug_assertions) || !cfg!(feature = "std") {
-        super::throw_str(msg)
-    } else {
-        #[cfg(feature = "std")]
-        {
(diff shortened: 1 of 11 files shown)


src/externref.rs
```diff
@@ -98,10 +99,17 @@ impl Slab {
...
     } else {
-        std::process::abort()
+        #[cfg(feature = "std")]
+        {
+            std::process::abort()
+        }
+        #[cfg(not(feature = "std"))]
+        {
+            unreachable!()
+        }
```
daxpedda, line 112: Lets use [`core::arch::wasm32::unreachable()`](https://doc.rust-lang.org/1.80.0/core/arch/wasm32/fn.unreachable.html) here instead.

daxpedda, line 112: Same as above.

src/lib.rs
```diff
@@ -1609,151 +1610,155 @@ pub mod __rt {
...
+            super::throw_str("invalid malloc request")
+        } else {
+            #[cfg(feature = "std")]
+            {
                 std::process::abort();
             }
+            #[cfg(not(feature = "std"))]
+            {
+                unreachable!()
+            }
```
daxpedda, line 1749: Same as above.

src/lib.rs
```diff
@@ -1790,31 +1795,36 @@ pub mod __rt {
...
     pub fn link_mem_intrinsics() {
-        crate::externref::link_intrinsics();
-    }
-
-    std::thread_local! {
-        static GLOBAL_EXNDATA: Cell<[u32; 2]> = Cell::new([0; 2]);
+        #[cfg(feature = "std")]
+        {
+            crate::externref::link_intrinsics();
+        }
```
daxpedda, line 1801: I'm incredibly mindblown by this hack and very surprised its needed in the first place.
The [`no_mangle` attribute](https://doc.rust-lang.org/1.80.0/reference/abi.html?highlight=no_mangle#the-no_mangle-attribute) should imply the [`used` attribute](https://doc.rust-lang.org/1.80.0/reference/abi.html?highlight=no_mangle#the-used-attribute), addressing this issue.

Without digging into this, the best we can do is make a separate module, as described in the commend, and move `link_intrinsics()` there.

daxpedda (changes requested):
Apologies for the delay!
Thank you for doing this!

daxpedda (approved):

Systemcluster:
Thanks for merging!

Issue #2370, opened 2020-11-25
Support for Option arguments of user types


fjarri:
Currently, having a binding that takes an optional argument of a custom struct
```
#[wasm_bindgen]
pub fn func(&self, param: Option<&MyStruct>) -> ...
```
requires one to implement `FromWasmAbi` and `OptionFromWasmAbi` for `&MyStruct`, which are not very well documented, and require `unsafe` code and a good understanding of `wasm-bindgen` internals. It looks like something that can be derived automatically.

Alternatively, one would think that the following would work:
```
#[wasm_bindgen]
pub fn func(&self, param: Option<MyStruct>) -> ...
```
This compiles, but leads to unexpected behavior on the JS side (I couldn't find a mention of it in the documentation, which, in fact, states that `Option<T>` parameters are not supported): the first invocation of the binding nullifies the pointer of the JS struct:
```
let param = new MyStruct();
console.log(param);
obj.func(param);
console.log(param);
```
Output:
```
MyStruct {ptr: <some_nonzero_number>}
MyStruct {ptr: 0}
```

RReverser:
For clarity, this issue should probably be rephrased that its specifically for passing `Option` with references. By-value consumption and returns are already implemented and work as expected (as your last example shows).

fjarri:
I must admit I certainly did not expect this behavior on JS side, but I understand how it makes sense if you want to make your JS code behave like Rust. Also, I couldn't find any warning about it in the docs (in fact, `Option<T>` parameters [are explicitly marked as not supported](https://rustwasm.github.io/docs/wasm-bindgen/reference/types/result.html))

RReverser:
@fjarri That table is for `Result<T, JsValue>` specifically. Like for other types in that section, what it means is that `Option<Result<T, JsValue>>` is not supported in params (I admit, it's confusing though, due to usage of `T` in both).

RReverser:
That said, [here](https://rustwasm.github.io/docs/wasm-bindgen/reference/types/exported-rust-types.html) it should certainly be marked as supported in both params and returns.

fjarri:
> Like for other types in that section, what it means is that Option<Result<T, JsValue>> is not supported in params (I admit, it's confusing though, due to usage of T in both).

Ah, I see, my mistake.

alexcrichton:
[...]

Issue #1906, opened 2019-12-09
Option<JsValue> and interoperate with js in a more js natural way


trivigy:
### Motivation

I have been trying to expose a rust method to javascript which accepts an optional arbitrary javascript object. Unfortunately so far most of my attempts to do it properly, and I put an emphasis on `properly`, are unsuccessful. The main issue I am running into can be described as the following error message.

```
  --> src/components/button.rs:67:1
   |
67 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ the trait `wasm_bindgen::convert::traits::OptionFromWasmAbi` is not implemented for `wasm_bindgen::JsValue`
   |
   = note: required because of the requirements on the impl of `wasm_bindgen::convert::traits::FromWasmAbi` for `std::option::Option<wasm_bindgen::JsValue>`
```
> I will address that I think "properly" is in the context of this entire issue.

From reading the other conversations across open and closed issues and documentations I understand that the case is that this feature is simply not implemented. I could not however understand whether this is because it is too complicated or not something that is sought after. I am including the following links for the record and minor explanations of what I had learned and where.

https://rustwasm.github.io/docs/wasm-bindgen/reference/types/jsvalue.html
Here one can see the table above that shows the state of what and where this type works and `Option<T>` says `NO`.

https://github.com/rustwasm/wasm-bindgen/issues/14#issuecomment-368305569
https://github.com/rustwasm/wasm-bindgen/issues/14#issuecomment-387143548
https://github.com/rustwasm/wasm-bindgen/issues/14#issuecomment-387157611
Here are additional mentions of this. This conversation was very interesting and seems to indicate that ultimately we want to have something like that in the library but at a much more complex capacity. Allowing for `Option<CustomType>`.

Now this seems to be as a great goal to have but I find it hard to reconcile with how typical javascript/typescript workflow is done. For example, following these instructions it is possible to expose a custom rust type to js [HERE](https://rustwasm.github.io/docs/wasm-bindgen/reference/types/exported-rust-types.html), although very limited in functionality. In particular the only public types that are currently supported are numbers. This means that something along the lines of the following is not possible.

```
#[wasm_bindgen]
pub struct ExportedNamedStruct {
    pub inner: String,
}
```

There are workaround that I have seen being suggested that recommend utilizing setters/getters, however I see the workaround approach as very antipattern to a typical js/ts code. Let me explain better in a form of an example.

#### Example
[...]

Issue #14, opened 2018-02-11
Support for Option


nickbabcock:
```rust
#[no_mangle]
#[wasm_bindgen]
pub extern fn optional() -> Option<i32> {
    None
}
```

Ideally this would be returned as `null` on the JS side

```
the trait `wasm_bindgen::convert::WasmBoundary` is not implemented for `std::option::Option<i32>`
```

(this seems like lower hanging fruit than `Result`)

---

The good news is that I'm getting my head around WebAssembly so I may be able to help in the future 🤞 

Edit: Oof, maybe it's best to tackle the general generics problem before specializing for the `Option` use case.

alexcrichton:
Nah I think this should definitely be possible! It may be a little tricky getting optional integers working, but for sure optional structs/objects should work.

s3bk:
I have been thinking about the orthogonal approach of creating an Option type in JS.

Option.unwrap() -> returns value or raises an exception
Option.or_null() -> returns value or null
Option.or_undefined() -> returns value or undefined
Option.is_some() -> true if Some, false if None
Option.is_none() -> true if None, false if Some

alexcrichton:
Hm yeah that's definitely possible! I was thinking though we'd probably interpret `Option<T>` as `T` or `null` in JS though?

s3bk:
… that would then map to `*mut T` or `NULL`, which also makes sense.

Maybe I was trying too much to make the interfaces for ?Option<T>` and `Result<T, E>` similar.

RReverser:
Just 5c - semantically, `undefined` would be more appropriate representation for `None` as it's intended to represent a missing value, while `null` is more like `ptr::null` (a valid value for object pointer). For example, this would allow to correctly represent `Some(ptr::null())` vs `None` in `Option<*const T>` types.

alexcrichton:
@s3bk er sorry what I mean is that `Option<JsValue>` in Rust would, for example, be either that JS value in JS or `null` (or something like that, maybe `undefined`). It does mean that `Some(JsValue::null())` wouldn't work (it'd come out as `null`) but there's not much we can do about that.

Another example is `Option<i32>` which would be either a number in JS or `null`/`undefined`

s3bk:
I don't like that some cases would not work. (Then you also have to remember which cases don't work, or you will run into strange bugs.)
[...]

PR #4007, opened 2024-07-06
Add OptionIntoWasmAbi and OptionFromWasmAbi implementations for JsValue

Icemic:
OptionIntoWasmAbi: This will allow adding `#[wasm_bindgen]` to a function returning `Result<Option<JsValue>, ...>` which is useful for Rust code when it needs to return nothing and not want to create a `undefined` JsValue.

OptionFromWasmAbi: To be pair to `OptionIntoWasmAbi`, it allows transform a undefined or null in js to `None` if you define the return type of js method to `Result<Option<JsValue>, ...>`

P.S. I am not sure why the none check inside `OptionFromWasmAbi` is `*js == 0` instead of `*js == JSIDX_UNDEFINED`, and why it cannot distinguish undefined and null (both are 0). So the impl of `OptionFromWasmAbi` for JsValue could be wrong.

### DIFF
--- a/src/convert/impls.rs
+++ b/src/convert/impls.rs
@@ -323,12 +323,26 @@ impl IntoWasmAbi for JsValue {
     }
 }
 
+impl OptionIntoWasmAbi for JsValue {
+    #[inline]
+    fn none() -> u32 {
+        JsValue::undefined().into_abi()
+    }
+}
+
 impl FromWasmAbi for JsValue {
     type Abi = u32;
 
     #[inline]
     unsafe fn from_abi(js: u32) -> JsValue {
         JsValue::_new(js)
+    }
+}
+
+impl OptionFromWasmAbi for JsValue {
+    #[inline]
+    fn is_none(js: &u32) -> bool {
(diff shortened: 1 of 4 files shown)


daxpedda:
Unfortunately the lack of these implementations is on purpose and has a lot of history: #14, #1906, #2370.
Basically it boils down to it not mapping very well, e.g. `Some(JsValue::UNDEFINED)` is still possible.

A proper use-case would help a lot!
In the meantime I will close this, but please reply here and I can re-open it!

> P.S. I am not sure why the none check inside `OptionFromWasmAbi` is `*js == 0` instead of `*js == JSIDX_UNDEFINED`, and why it cannot distinguish undefined and null (both are 0). So the impl of `OptionFromWasmAbi` for JsValue could be wrong.

`JSIDX_UNDEFINED != 0` so I'm not sure what exactly you mean here.

Also the translation is done in JS, so distinguishing between null or undefined happens in JS beforehand.
In this case its not distinguished in Rust, basically the same problem as above.

PR #4008, opened 2024-07-08
Fix nightly CI

daxpedda:
Nightly CI was broken because the version it was using was too old and a new version of Serde started failing.
See https://github.com/serde-rs/serde/pull/2767 (probably).


upsuper:
@daxpedda 👋 It seems the build starts failing again. Should we do another version bump on the CI?

PR #4015, opened 2024-07-17
Fix byte order for big-endian platforms

magic-akari:
> [!NOTE]  
> WebAssembly memory is always in little-endian format, regardless of the platform it's run on. Therefore, for portability, you should read and write multi-byte values in JavaScript using [DataView](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView).

https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Memory

Related issues:

- swc-project/swc#9259
- https://github.com/nodejs/node/pull/53725#issuecomment-2229226306

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1766,47 +1766,47 @@ impl<'a> Context<'a> {
     }
 
     fn expose_int8_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Int8", memory)
+        self.memview("Int8Array", memory)
     }
 
     fn expose_uint8_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Uint8", memory)
+        self.memview("Uint8Array", memory)
     }
 
     fn expose_clamped_uint8_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Uint8Clamped", memory)
+        self.memview("Uint8ClampedArray", memory)
     }
 
     fn expose_int16_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Int16", memory)
+        self.memview("Int16Array", memory)
     }
 
     fn expose_uint16_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Uint16", memory)
+        self.memview("Uint16Array", memory)
     }
 
     fn expose_int32_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Int32", memory)
+        self.memview("Int32Array", memory)
     }
 
     fn expose_uint32_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("Uint32", memory)
+        self.memview("Uint32Array", memory)
     }
 
     fn expose_int64_memory(&mut self, memory: MemoryId) -> MemView {
-        self.memview("BigInt64", memory)
+        self.memview("BigInt64Array", memory)
     }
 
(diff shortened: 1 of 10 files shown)


magic-akari:
Could you review this pull request? @daxpedda

daxpedda (commented):
I'm a bit confused why you chose not to re-use `expose_*_memory()`.
The only place it still seems to be used is for importing arrays, which you actually haven't covered here.

I will have more time tomorrow to look into this properly.

> Could you review this pull request? @daxpedda

I apologize for the delay, my backlog from my vacation was quite big and I'm just getting to it.

magic-akari:
> I'm a bit confused why you chose not to re-use `expose_*_memory()`. The only place it still seems to be used is for importing arrays, which you actually haven't covered here.

Could you please specify which part you're referring to? Marking it on the relevant code would be helpful.

crates/cli-support/src/js/mod.rs
```diff
@@ -1769,15 +1767,21 @@ impl<'a> Context<'a> {
...
         if !self.should_write_global(view.name.clone()) {
             return view;
         }
         let mem = self.export_name_of(memory);
 
+        let is_dataview = kind == "DataView";
+
         let cache = format!("cached{}Memory{}", kind, view.num);
-        let resized_check = if self.module.memories.get(memory).shared {
+        let resized_check = if self.module.memories.get(memory).shared || is_dataview {
```
daxpedda, line 1784: This should be explained in a comment.
Additionally, I think we can do a cheaper check by testing [`ArrayBuffer.prototype.detached`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/detached).

crates/cli-support/src/js/mod.rs
```diff
@@ -1795,11 +1799,17 @@ impl<'a> Context<'a> {
 
         self.global(&format!("let {cache} = null;\n"));
 
+        let kind = if is_dataview {
+            kind
+        } else {
+            &format!("{}Array", kind)
+        };
```
daxpedda, line 1806: Lets just pass the `Array` part in all calling functions instead.

daxpedda (changes requested):
My earlier assumptions where wrong, when getting arrays out of Wasm to JS land, we can't actually use `DataView`, this is a responsibility the user will have to take when extracting data from `DataView`.

crates/cli-support/src/js/mod.rs
magic-akari: Yes, we check its detached status, but unfortunately, node.js currently does not support this.

daxpedda: Yes, apparently this is reproduced in CI as well.
Still trying to figure out the details, because I'm unable to re-produce it locally.

crates/cli-support/src/js/mod.rs

crates/cli-support/src/js/mod.rs
magic-akari: image

daxpedda: Yes, so the issue is that Node v20 is broken in this regard, I will skim their changelog.
But in any case, the current LTS version is v20, so we would want to support that.

I found that Node v22 works.

daxpedda: This doesn't just affect Node, Safari also only implemented this since v17.4 and we definitely support older versions then that.

So its good that CI caught it, I added a bit of an ugly fallback, but it works.

daxpedda (approved):

PR #4024, opened 2024-07-29
Add alloc feature

Systemcluster:
Follow up to #4005. This keeps a path open to add a no-alloc implementation without breaking changes in the future, making `alloc` a required feature for now.

(This makes `default-features=false` without adding `std` or `alloc` break again like in the current release, but prevents sudden breakage from adding a new default feature in a future release.)

Note: I don't feel too strongly about this, and I'm not sure how useful wasm-bindgen can be without `alloc`, this is mostly a precautionary change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
     - run: cargo check --all
-    - run: cargo check --no-default-features
+    - run: cargo check --no-default-features --features alloc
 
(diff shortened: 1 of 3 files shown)


Systemcluster:
Alternatively to errorring without `alloc` feature, the `build.rs` could print a future-compat warning.

@daxpedda What do you think?

daxpedda:
I don't think compiling without `alloc` is a worthwhile goal for `wasm-bindgen`.
For sure there are uses for `wasm32-unknown-unknown` without `alloc`, none of those need to interface with JS though.

Thank you for doing this however, its good that we discussed it at least and gave it a good think, appreciated!

PR #4027, opened 2024-07-30
Add node module output format

Systemcluster:
This PR adds a node module output target. Node stabilized node modules with release 12, a lot of the JS ecosystem moved towards publishing modules, and the other targets apart from `nodejs` all generate modules.

There were already code-paths for this target, but it wasn't hooked up or finished. I changed the naming away from "experimental" and finished the implementation.

Support for the format would also need to be added to `wasm-pack` after this is merged.


crates/cli-support/src/lib.rs
```diff
@@ -687,11 +663,7 @@ impl Output {
 
         // And now that we've got all our JS and TypeScript, actually write it
         // out to the filesystem.
-        let extension = if gen.mode.nodejs_experimental_modules() {
-            "mjs"
-        } else {
-            "js"
-        };
```
daxpedda, line 694: Do Node ES modules not use `mjs`?

daxpedda (commented):
Even though the changes are very small, personally I can't review this as I have no clue about Node.

Best case scenario I can take a look at it later when I have more time to learn whats needed here, but even then I would keep it experimental until a maintainer who knows whats going on here can review this.

crates/cli-support/src/lib.rs
Systemcluster: Both `mjs` and `js` are used: https://nodejs.org/api/packages.html#determining-module-system

None of the other targets use `mjs` even though they all output modules, and the `node` target uses `js` instead of `cjs` even though it is the only non-module output. I can undo this change, it was for consistency more than for correctness. `mjs` would make things easier in a mixed-esm-and-cjs package too.

(Personally I would prefer to rename the `node` output to `cjs` too, but I don't think that's doable in a patch release. You currently have to [rename](https://github.com/Systemcluster/kitoken/blob/wip5/packages/javascript/package.json#L28) it manually when necessary.)

Systemcluster:
Thanks for checking, I'll annotate some of the changes later to make it easier to review.

crates/cli-support/src/lib.rs
daxpedda: Considering this was experimental to begin with, I don't think we have to adhere to Semver.

crates/cli-support/src/js/mod.rs
```diff
@@ -217,26 +213,29 @@ impl<'a> Context<'a> {
...
-                ));
-            } else if self.config.mode.nodejs_experimental_modules() {
-                shim.push_str(&format!("imports['{}'] = import{};\n", module, i));
-            } else {
-                shim.push_str(&format!("imports['{0}'] = require('{0}');\n", module));
+            for (i, module) in imports.iter().enumerate() {
+                if module.as_str() != PLACEHOLDER_MODULE {
+                    shim.push_str(&format!("imports['{}'] = import{};\n", module, i));
+                }
+            }
```
Systemcluster, line 228: The import steps are added to the index file now, so no circular import, whether that would be an issue or not.

Why it generates an import for `{name}_bg.js` as shown in the snippet in the PR I'm not sure, but it's correct. That's what the import table looks like:

```
Import[2]:
 - func[0] sig=1 <./node_bg.js.__wbindgen_string_new> <- ./node_bg.js.__wbindgen_string_new
 - func[1] sig=2 <./node_bg.js.__wbindgen_throw> <- ./node_bg.js.__wbindgen_throw
```

Systemcluster (commented):
I've split up the changes into separate commits, so it should be more obvious what changes what. The core changes are in 4d63628d8f17f84d61bd0e7b459507beb0a465dd. In general, the new target mostly uses the `bundler` target paths, plus the existing `node` import logic as noted below.

crates/cli-support/src/lib.rs
Systemcluster: It seems like there is other code that depends on the extension being `.js`, multiple tests fail with `.mjs`. I don't want to spend more time investigating those issues right now.

crates/cli-support/src/js/mod.rs
```diff
@@ -442,13 +445,10 @@ impl<'a> Context<'a> {
...
-            // the wasm file as if it were an ES module and let the
-            // bundler/runtime take care of it.
-            OutputMode::Bundler { .. }
-            | OutputMode::Node {
-                experimental_modules: true,
-            } => {
+            // With Bundlers we can simply import the wasm file as if it were an ES module
+            // and let the bundler/runtime take care of it.
+            // With Node we manually read the wasm file from the filesystem and instantiate it.
+            OutputMode::Bundler { .. } | OutputMode::Node { module: true } => {
```
Systemcluster, line 451: Unfortunately this didn't turn out to be the case, [not yet anyways](https://nodejs.org/api/esm.html#wasm-modules).

Systemcluster:
My earlier question resolved itself after gaining some more familiarity with the codebase, I've updated my comments where I was being unclear.

I've uploaded some outputs here: https://github.com/Systemcluster/kitoken/tree/temp/packages/javascript/dist
The `module` files are from the new target. It's very similar to the `bundler` target output. `node ./test.js` in the parent directory can be used as a test.

@daxpedda I think you should be able to review this in this state.

daxpedda (changes requested):
LGTM!

I'm happy to move this forward with the following changes:
- Name the option `experimental-nodejs-module`.
  - I don't feel entirely comfortable otherwise, considering my lack of knowledge with this tech.
  - After some digging I do believe we want to produce `.mjs` files, so until we figure that out we shouldn't stabilize the option.
- Make sure its being tested in CI.
  The CI is a bit of a mess and I have no guidance on where to add it, unfortunately you will have to dig into it yourself.

Systemcluster:
> I don't feel entirely comfortable otherwise, considering my lack of knowledge with this tech.

I'm fine with experimental as long as it's merged. 🙂

> After some digging I do believe we want to produce .mjs files, so until we figure that out we shouldn't stabilize the option.

That will require a few more changes than 4d63628d8f17f84d61bd0e7b459507beb0a465dd if I've seen right, so far I was trying to change as little as possible. If I go through and generalize the extension already, I'll also add a command line option for overwriting the default.

Just to make clear, currently there are the following targets:

|Target| Type| Extension|
|--------|--------|--------|
|`bundler`, `deno`, `web`| ESM | `.js`|
|`nodejs`| CJS | `.js`|
|`no-modules`| Integrated | `.js`|

Changing it for this target would add another row:

|Target| Type| Extension|
|--------|--------|--------|
|`nodejs-module`| ESM | `.mjs`|

While I don't disagree that a `.mjs` extension would make sense for the target itself, it would be surprising with the other targets being as they are. Consider if an option to manually set the extension and leaving the default consistent with the others it isn't better here at this point.

In either case, since this remains experimental for now we can do that in a follow-up PR.

> The CI is a bit of a mess and I have no guidance on where to add it, unfortunately you will have to dig into it yourself. 

I didn't find any relevant tests on first glance and was hoping for more info, but I'll take digging in. 😄

daxpedda:
> While I don't disagree that a `.mjs` extension would make sense for the target itself, it would be surprising with the other targets being as they are. Consider if an option to manually set the extension and leaving the default consistent with the others it isn't better here at this point.

Is there a reason file extensions should be consistent across targets?
I certainly would not have expected so, because targets are in no way equal, they have to be handled very differently.

> In either case, since this remains experimental for now we can do that in a follow-up PR.

Agreed.

---

Your work and input so far is really appreciated, it has been very valuable and you put a lot of effort into making this as easy to review as possible.

Thank you!

Systemcluster:
> Is there a reason file extensions should be consistent across targets?

Principle of least surprise is the main one, but I don't have reasons beyond that.

> because targets are in no way equal, they have to be handled very differently.

Technically yes, but only to some degree. If you want to distribute your wasm-bindgen generated files on NPM you always have to package it with a `package.json`, and resolution through `package.json` by default goes through Node. TypeScript and other tools also adhere to the Node resolution strategy, either by default or as an option, and that resolution will consider the `"type"` field in the `package.json` to decide whether `.js` files are ESM or CJS. That part is the same with all targets.

I'll think about it more until the follow-up. I actually don't mind using `.mjs` here, I just wish we could make the `nodejs` target output `.cjs` instead.

Systemcluster:
I added tests in the place where I think it's most apppropriate, but I'm not sure where in the CI they are called. (I think the latest test failure is unrelated.)

Systemcluster:
It looks like for use-cases without a `package.json` the explicit extension is really useful 😄 

Test is running now, I added a `package.json` temporarily to make the test work. It can be removed with the extension change.

crates/test/src/lib.rs
```diff
@@ -76,6 +76,12 @@ macro_rules! wasm_bindgen_test_configure {
         pub static __WBG_TEST_RUN_IN_SERVICE_WORKER: [u8; 1] = [0x04];
         $crate::wasm_bindgen_test_configure!($($others)*);
     );
+    (run_in_node $($others:tt)*) => (
+        #[link_section = "__wasm_bindgen_test_unstable"]
+        #[cfg(target_arch = "wasm32")]
+        pub static __WBG_TEST_RUN_IN_NODE: [u8; 1] = [0x05];
+        $crate::wasm_bindgen_test_configure!($($others)*);
+    );
```
daxpedda, line 84: This should be renamed to contain `experimental` as well.

Not required, but it would be nice if this is documented here:
https://github.com/rustwasm/wasm-bindgen/blob/4596d4a633900a3f474c715aa51937ed0282e205/guide/src/wasm-bindgen-test/browsers.md#L17-L26

.github/workflows/main.yml
```diff
@@ -113,6 +113,16 @@ jobs:
...
 
+  test_node:
+    name: "Run wasm-bindgen node tests"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - run: rustup update --no-self-update stable && rustup default stable
+    - run: rustup target add wasm32-unknown-unknown
+    - run: |
+          cargo test --target wasm32-unknown-unknown --test headless_no_threads -p js-sys
```
daxpedda, line 124: Why does this require a separate job?
It should be covered here:
https://github.com/rustwasm/wasm-bindgen/blob/4596d4a633900a3f474c715aa51937ed0282e205/.github/workflows/main.yml#L188-L202

crates/cli-support/src/js/mod.rs
```diff
@@ -476,7 +485,10 @@ impl<'a> Context<'a> {
                 );
 
                 if needs_manual_start {
-                    start = Some("\nwasm.__wbindgen_start();\n".to_string());
+                    _start.push_str("\nwasm.__wbindgen_start();\n");
+                }
+                if !_start.is_empty() {
+                    start = Some(_start);
                 }
```
daxpedda, line 492: I'm assuming this logic is a refactoring leftover?
I assume this doesn't have to happen before `self.imports_post.push_str()`, so it could all be at the bottom, not requiring the `_start` variable.

Let me know if I'm missing something.

crates/cli/src/bin/wasm-bindgen.rs
```diff
@@ -18,7 +18,7 @@ Options:
     --out-dir DIR                Output directory
     --out-name VAR               Set a custom output filename (Without extension. Defaults to crate name)
     --target TARGET              What type of output to generate, valid
-                                 values are [web, bundler, nodejs, no-modules, deno],
+                                 values are [web, bundler, nodejs, no-modules, deno, experimental-nodejs-module],
```
daxpedda, line 21: Because its experimental its not required, but it would be nice if we add an entry here:
https://github.com/rustwasm/wasm-bindgen/blob/4596d4a633900a3f474c715aa51937ed0282e205/guide/src/reference/deployment.md#L11-L23

crates/js-sys/tests/headless_no_threads.rs
daxpedda, line 1: Whats going on with the "no_threads" in the file name?

If the test is otherwise the same as the browser one, lets make it a shared file between these two tests.

But before doing that: if your intention was to just run any available test in the new target, I would recommend you to put it in `/tests`. `/tests/headless` would seem appropriate to me.
Sharing the same file between multiple targets there should be the way to go as well, if it makes sense.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
```diff
@@ -88,7 +98,14 @@ pub fn execute(
...
     );
 
-    let js_path = tmpdir.join("run.js");
+    let js_path = if module_format {
+        // fixme: this is a hack to make node understand modules
+        let package_json = tmpdir.join("package.json");
+        fs::write(&package_json, r#"{"type": "module"}"#).unwrap();
+        tmpdir.join("run.mjs")
+    } else {
+        tmpdir.join("run.cjs")
```
daxpedda, line 107: Please enlighten me: the output we currently produce with the "node" (not the experimental module) target is CommonJS?

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
```diff
@@ -88,7 +98,14 @@ pub fn execute(
     ",
     );
 
-    let js_path = tmpdir.join("run.js");
+    let js_path = if module_format {
+        // fixme: this is a hack to make node understand modules
+        let package_json = tmpdir.join("package.json");
+        fs::write(&package_json, r#"{"type": "module"}"#).unwrap();
+        tmpdir.join("run.mjs")
```
daxpedda, line 105: > It looks like for use-cases without a `package.json` the explicit extension is really useful 😄
> 
> Test is running now, I added a `package.json` temporarily to make the test work. It can be removed with the extension change.

So coming back to the discussion from earlier in https://github.com/rustwasm/wasm-bindgen/pull/4027#issuecomment-2263042667:
> > Is there a reason file extensions should be consistent across targets?
> 
> Principle of least surprise is the main one, but I don't have reasons beyond that.

What I was trying to say is: consistency only makes sense when the treatment of the output is desired to be similar. But it seems to me that you won't use the modules target *and* the no-modules target, you will be using either.

Consistency would be more desirable if, lets say, you want to use *both* targets with the same configuration.

Please keep in mind that I barely touched this ecosystem in my life and I'm clueless about the tools. I might have some assumptions that are just wrong.

---

I'm in favor of switching the target right now if we can avoid the `package.json`.
You mentioned some issues before with switching to the new file, are these resolved at this point seeing that the test runs successfully?

daxpedda (changes requested):

crates/js-sys/tests/headless_no_threads.rs
Systemcluster: There is a test called `headless` that's called with threads support here: https://github.com/rustwasm/wasm-bindgen/pull/4027/files/80b29068e8d6586faef0896011306c7f01549265#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R113

Hence you can't just add `run_in_node` to the existing one.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
Systemcluster: Yes.

daxpedda: Well, another thing to consider in the next breaking change.
Thanks!

crates/js-sys/tests/headless_no_threads.rs
daxpedda: So the idea is that you have a shared Rust file, because the tests are otherwise equal, and then use this shared test file by two separate test files that only have `wasm_bindgen_test_configure!(run_in_browser);` or `wasm_bindgen_test_configure!(run_in_node);` in them.
That way the threading test can still continue run only in the browser target (by using `--test browser`) while not being executed in Node.

So in the end you will have three files:
- `browser.rs` (containing `wasm_bindgen_test_configure!(run_in_browser); pub mod common;`)
- `node.rs` (containing `wasm_bindgen_test_configure!(run_in_node); pub mod common;`)
- `common/mod.rs` (containing all the tests that were in `headless.rs` before)

See https://github.com/rustwasm/wasm-bindgen/tree/4596d4a633900a3f474c715aa51937ed0282e205/crates/test/sample for inspiration.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
Systemcluster: I just didn't consider the use-case of running bare scripts without `package.json` as very common, the vast majority of scripts I've seen are distributed as packages, but of course it's a use-case that exists. In a package it just doesn't matter as long as all files are in the same format. But I don't disagree.

> I'm in favor of switching the target right now if we can avoid the package.json. [...]

There is still a bunch of code that assumes a `.js` extension, I haven't checked how much. Right now only the `run.js` is renamed, since it's not used anywhere else.

crates/test/src/lib.rs
Systemcluster: The `__WBG_TEST_RUN_IN_NODE` here is not actually used anywhere and is only there for the `#[link_section]`.

Systemcluster: Do you mean the `run_in_node` test specifier?

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
daxpedda: > There is still a bunch of code that assumes a `.js` extension, I haven't checked how much. Right now only the `run.js` is renamed, since it's not used anywhere else.
> 
> I think the `package.json` is fine here for now since it's only in a temporary directory and only in the test runner, and it's only created when running tests for the new target. I can get back to it during the week.

No worries, lets do it in a follow-up then!

crates/test/src/lib.rs
daxpedda: Yes!

crates/cli-support/src/js/mod.rs
```diff
@@ -475,9 +475,20 @@ impl<'a> Context<'a> {
...
+                        start.push_str(&self.generate_node_wasm_loading(Path::new(&format!(
+                            "./{}_bg.wasm",
+                            module_name
+                        ))));
+                    }
+                    if needs_manual_start {
+                        start.push_str("\nwasm.__wbindgen_start();\n");
+                    }
+                    (!start.is_empty()).then_some(start)
+                };
```
daxpedda, line 491: Apologies, this is what I was proposing:
```suggestion
                if matches!(self.config.mode, OutputMode::Node { module: true }) {
                    let start = start.get_or_insert_with(String::new);
                    start.push_str(&self.generate_node_imports());
                    start.push_str(&self.generate_node_wasm_loading(Path::new(&format!(
                        "./{}_bg.wasm",
                        module_name
                    ))));
                }
                if needs_manual_start {
                    start
                        .get_or_insert_with(String::new)
                        .push_str("\nwasm.__wbindgen_start();\n");
                }
```

.github/workflows/main.yml
daxpedda, line 1: Can we remove the changes here?

crates/cli-support/src/js/mod.rs
Systemcluster: That works too, I guess it saves an allocation in case neither matches. 😄

daxpedda: FWIW, I just found it easier to reason about the code.
[`String::new()`](https://doc.rust-lang.org/1.80.0/std/string/struct.String.html#method.new) doesn't actually allocate.

Systemcluster: Yeah I can see it.

daxpedda (approved):
Great work, thank you!

Systemcluster:
It took a bit of back and forth but we got there, thanks for the review and guidance!

Systemcluster:
Thanks for linking that issue, that's a feature I want too. I wanted to experiment with some approaches during the week.

PR #4033, opened 2024-08-01
Add semicolon to "export { initSync }" in OutputMode::Web

Awpteamoose:
The rest of the generated js glue doesn't rely on implicit semicolons, but this line does. This causes some minifiers (one of which I happen to use) to output erroneous code.


daxpedda (approved):
You should probably report that to your minifier as well!

PR #4035, opened 2024-08-01
improved error reporting for UnwrapThrowExt for Result

sokorototo:
Since `Result<T, E: fmt::Debug>` has access to `E: fmt::Debug`, this PR modifies the default implementation of `UnwrapThrowExt::unwrap_throw()` to include this extra information. This is my first PR, and I'm open to criticism

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1382,6 +1382,32 @@ where
     E: core::fmt::Debug,
 {
     #[cfg_attr(debug_assertions, track_caller)]
+    fn unwrap_throw(self) -> T {
+        if cfg!(all(
+            target_arch = "wasm32",
+            not(any(target_os = "emscripten", target_os = "wasi"))
+        )) {
+            match self {
+                Ok(val) => val,
+                Err(err) => {
+                    let loc = core::panic::Location::caller();
+                    let msg = alloc::format!(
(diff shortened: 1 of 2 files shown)


daxpedda (approved):
Wow, nice!
Thank you!

daxpedda:
I will add a changelog entry.

src/lib.rs
```diff
@@ -1381,6 +1381,32 @@ impl<T, E> UnwrapThrowExt<T> for Result<T, E>
 where
     E: core::fmt::Debug,
 {
+    #[cfg_attr(debug_assertions, track_caller)]
+    fn unwrap_throw(self) -> T {
+        if cfg!(all(
+            target_arch = "wasm32",
+            not(any(target_os = "emscripten", target_os = "wasi"))
+        )) {
```
Liamolucko, line 1389: This needs to be gated behind `cfg(debug_assertions)`, the same as the default `unwrap_throw` implementation; see https://github.com/rustwasm/wasm-bindgen/pull/2995#pullrequestreview-1041858759.

(The `cfg(feature = "std")` might not be needed anymore after #4005.)

daxpedda: Thank you!
Addressed in #4042.

PR #4038, opened 2024-08-02
Introduce MSRV policy

### LINKED ISSUE #3114: Document MSRV
### Motivation

I'm adding a feature to `wasm-bindgen` (opening a PR soon), but don't know against which version of Rust I should check my changes. Clearly documenting the Minimum Supported Rust Version would help me test my changes.

### Proposed Solution

As  `.github/workflows/main.yml` uses Rust 1.56 for one of the tests, maybe we could start with this version? Then we could document it in the README, and/or by adding a [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) field to `Cargo.toml` (which incidentally is supported on Rust 1.56 and up).

### Alternatives

There may be other ways to document the MSRV, but "MSRV" itself has become a fairly common keyword to look for that information, so I'd like for it to be surfaced in the README and in `Cargo.toml`.

### Additional Context

N/A.

### LINKED ISSUE #3918: wasm-bindgen is broken and untested at MSRV (1.57)
`webauthn-rs` is a downstream user of `wasm-bindgen`, and we run tests in CI at our MSRV (currently 1.70) and on Rust stable. 

We've recently started getting build failures at MSRV due to [this library's dependency on `bumpalo`]((https://github.com/rustwasm/wasm-bindgen/blob/d25a68eaa778f339ea9deb147ddc76a1facfc625/crates/backend/Cargo.toml#L19-L20)):

```
% cargo tree -i bumpalo
bumpalo v3.15.4
└── wasm-bindgen-backend v0.2.92
    └── wasm-bindgen-macro-support v0.2.92
        └── wasm-bindgen-macro v0.2.92 (proc-macro)
            └── wasm-bindgen v0.2.92
```

[`bumpalo 3.15.1` and later have an MSRV of `1.73`](https://github.com/fitzgen/bumpalo/issues/238) – and [this was noticed and bumped in `wasm-bindgen`'s CI configuration](https://github.com/rustwasm/wasm-bindgen/pull/3850). However, [`wasm-bindgen` has a stated MSRV of 1.57](https://github.com/rustwasm/wasm-bindgen/blob/d25a68eaa778f339ea9deb147ddc76a1facfc625/Cargo.toml#L15).

Going back in history, [`bumpalo 3.14.0` has an MSRV of 1.60](https://crates.io/crates/bumpalo/versions), but there's *no* version of `bumpalo` which says it supports `wasm-bindgen`'s current *stated* MSRV (1.57) – probably because this metadata was never collected.

[`wasm-bindgen-backend` depends on `bumpalo 3.0.0`](https://github.com/rustwasm/wasm-bindgen/blob/d25a68eaa778f339ea9deb147ddc76a1facfc625/crates/backend/Cargo.toml#L19-L20), so there's a theoretical solution, but [Cargo doesn't consider the MSRV when selecting a dependency version](https://github.com/rust-lang/cargo/issues/9930).

As `webauthn-rs` is a collection of libraries and binaries (so we don't always control `Cargo.lock`), and `bumpalo` is a transitive dependency for us, we're very limited as to what we can do to mitigate the issue, aside from telling users on Rust `< 1.73` and building for WASM targets to run `cargo update -p bumpalo --precise 3.14.0` first.

## Digging deeper

The version of Rust used in `wasm-bindgen`'s CI config has been bumped several times to "fix build failures", *without* updating the MSRV metadata:

* from 1.57 to 1.59: https://github.com/rustwasm/wasm-bindgen/pull/3293 (the first divergence)
* from 1.59 to 1.60: https://github.com/rustwasm/wasm-bindgen/pull/3409
* from 1.60 to 1.69: https://github.com/rustwasm/wasm-bindgen/pull/3414
* from 1.69 to 1.76: https://github.com/rustwasm/wasm-bindgen/pull/3850 (`bumpalo`)

Because these changes were all made to work around build and/or test failures in CI on those lower Rust versions, they've become *de facto* MSRV changes. This means the MSRV metadata has been unverified (and quite probably false) for a bit over a year (since #3293).

In fairness to those involved, this was an incident waiting to happen – [there is no documented MSRV policy](https://github.com/rustwasm/wasm-bindgen/issues/3114), and none of the CI configuration makes reference to MSRVs.

## Proposed solution

Assuming a *new* MSRV target of 1.60, I propose:

* Update `wasm-bindgen` packages' MSRV metadata to 1.60.

  I'd be happy with `wasm-bindgen`'s MSRV going as high as 1.70 – that's at least what I'm confident we've actually tested, and shouldn't impact our users.

* Test `wasm-bindgen` in CI on the new MSRV **and** the current stable version of Rust.

  Make it clear in the CI config that changing the tested Rust versions is a *de facto* MSRV change.

* Pin `bumpalo` to `>=3.14.0,<3.15.0` (which has an MSRV of 1.60), as well as any other dependencies which have higher MSRV requirements – at least until Cargo can solve this itself and the new resolver is available to the MSRV.

* Publish SemVer-appropriate version bumps for all packages with the corrected MSRV.

Going forward, this should make `wasm-bindgen`s published MSRV metadata accurate, put guard rails around the CI config, and give downstream users a way to manage their own MSRV bumps.

daxpedda:
This PR introduces a MSRV policy.

Library crates have an MSRV of v1.57 and changes to the MSRV are accompanied by a minor version bump.
The following crates are included in this policy:
- `wasm-bindgen`
- `js-sys`
- `web-sys`
- `wasm-bindgen-futures`
- `wasm-bindgen-test`
- `wasm-bindgen-macro`
- `wasm-bindgen-test-macro`
- `wasm-bindgen-macro-support`
- `wasm-bindgen-backend`
- `wasm-bindgen-shared`

CLI crates and their support libraries have an MSRV of v1.76, due to being dependent on the [`wasm-tools` ecosystem](https://github.com/bytecodealliance/wasm-tools). The MSRV is only guaranteed for a specific version, so it is only accompanied by a patch version bump.
The following crates are included in this policy:
- `wasm-bindgen-cli`
- `wasm-bindgen-cli-support`
- `wasm-bindgen-wasm-interpreter`
- `wasm-bindgen-threads-xform`
- `wasm-bindgen-multi-value-xform`
- `wasm-bindgen-externref-xform`
- `wasm-bindgen-wasm-conventions`

Fixes #3114.
Fixes #3918.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -438,6 +438,28 @@ jobs:
         name: doc_api
         path: docs.tar.gz
 
+  msrv-lib:
+    name: Check MSRV for libraries
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: crates/msrv/lib
+    steps:
+    - uses: actions/checkout@v4
+    - run: rustup update --no-self-update 1.57 && rustup default 1.57
+    - run: cargo build
+      
+  msrv-cli:
+    name: Check MSRV for CLI tools
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: crates/msrv/cli
+    steps:
+    - uses: actions/checkout@v4
+    - run: rustup update --no-self-update 1.76 && rustup default 1.76
+    - run: cargo build
+      
 
   deploy:
     permissions:
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ yarn.lock
 .vscode
 webdriver.json
 benchmarks/pkg
+/crates/msrv/*/target
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,9 @@
 * Added support for WebIDL records. This added new methods to various APIs, notably `ClipboardItem()`, `GPUDeviceDescriptor.requiredLimits` and `Header()`.
   [#4030](https://github.com/rustwasm/wasm-bindgen/pull/4030)
 
+* Added an official MSRV policy. Library MSRV changes will be accompanied by a minor version bump. CLI tool MSRV can change with any version bump.
+  [#4038](https://github.com/rustwasm/wasm-bindgen/pull/4038)
+
 ### Changed
 
 * Stabilize Web Share API.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ members = [
   "examples/synchronous-instantiation",
   "tests/no-std",
 ]
+exclude = ["crates/msrv"]
 resolver = "2"
 
 [patch.crates-io]
--- a/README.md
+++ b/README.md
@@ -108,6 +108,12 @@ You can find general documentation about using Rust and WebAssembly together
 - [web-sys](https://docs.rs/web-sys)
 - [wasm-bindgen-futures](https://docs.rs/wasm-bindgen-futures)
 
+## MSRV Policy
+
+Libraries that are released on [crates.io](https://crates.io) have a MSRV of v1.57. Changes to the MSRV will be accompanied by a minor version bump.
+
+CLI tools and their corresponding support libraries have a MSRV of v1.76. Changes to the MSRV will be accompanied by a patch version bump.
+
 ## License
 
 This project is licensed under either of
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -3,7 +3,6 @@ name = "wasm-bindgen-benchmark"
 version = "0.1.0"
 authors = ["The wasm-bindgen Developers"]
 edition = "2018"
-rust-version = "1.57"
 
 [dependencies]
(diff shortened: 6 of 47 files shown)


Liamolucko:
> Library crates have an MSRV of v1.57 and changes to the MSRV are accompanied by a minor version bump.

That's a lot stronger of a policy than I would have expected: since all of those crates are at 0.x.y, a minor version bump is effectively a major version bump, and so we'll be stuck at an MSRV of 1.57 until we make a breaking release.

daxpedda:
> > Library crates have an MSRV of v1.57 and changes to the MSRV are accompanied by a minor version bump.
> 
> That's a lot stronger of a policy than I would have expected: since all of those crates are at 0.x.y, a minor version bump is effectively a major version bump, and so we'll be stuck at an MSRV of 1.57 until we make a breaking release.

Yep.
I don't see us really raising the MSRV considering we don't really have any dependencies.

Do you disagree?

PR #4047, opened 2024-08-04
Add snapshot tests

Systemcluster:
Instead of understanding the existing test infrastructure, I added snapshot tests. I think snapshot tests are a great choice here since they make both the current state as well as future changes easily observable.

The snapshot tests are currently not ideal due to the way the output is assembled, but I want to keep refactoring for another PR.

(Depends on #4027, only the last commit is new.)


Systemcluster:
I don't know what's causing this test to randomly fail, did it ever fail with debug mode enabled on retry?

```
---- duck_typed_interfaces stdout ----
Error: error running file server

Caused by:
    error shutting down connection
```

crates/cli-support/tests/test_add.rs
daxpedda, line 1: I didn't check, but both these files seem equal to me except the `ADD_WASM` constant.
Can we split the common code up in a module again so we don't have this much duplicate code?

crates/cli-support/Cargo.toml
```diff
@@ -28,3 +28,10 @@ wasm-bindgen-threads-xform = { path = '../threads-xform', version = '=0.2.92' }
 wasm-bindgen-wasm-conventions = { path = '../wasm-conventions', version = '=0.2.92' }
 wasm-bindgen-wasm-interpreter = { path = "../wasm-interpreter", version = '=0.2.92' }
 unicode-ident = "1.0.5"
+
+[dev-dependencies]
+insta = "1.39.0"
+
+[profile.dev.package]
+insta.opt-level = 3
+similar.opt-level = 3
```
daxpedda, line 37: Can you leave a comment here saying why this is necessary?

daxpedda (changes requested):
A big issue with the current testing infrastructure in `wasm-bindgen` is that is terribly fragmented.
While I believe this PR is a great change, it further increases that fragmentation.
Currently similar tests are done in `/crates/cli/tests/reference`.

Would it be possible to migrate the current similar tests  to use `insta` as well?
I didn't look to closely into `insta` yet, so let me know if its just not possible or why you believe it shouldn't be done.

Again, thank you!
I believe this is a great improvement for `wasm-bindgen`.

crates/cli-support/Cargo.toml
Systemcluster: It's not necessary, but it's from the recommended setup: https://insta.rs/docs/quickstart/#optional-faster-runs

Rust compiles all build dependencies without any optimizations by default.

daxpedda: That's a known fact I believe, this is only necessary when running your tests takes much longer than your compilation time. I think it this case its the reverse: compiling with optimizations just makes our compilation time longer with barely a benefit during runtime, because our test runtime should be around a second or so.

So I would remove this unless you think its not making the job actually run slower in CI?

Systemcluster: >  That's a known fact I believe

I didn't want to assume, Rust has a lot of lesser known corners! 😄

.github/workflows/main.yml
```diff
@@ -349,14 +363,16 @@ jobs:
...
-              (./build.sh && mkdir -p ../../exbuild/$dir && cp -r ./* ../../exbuild/$dir && rm ../../exbuild/$dir/**/.gitignore))
+        function build() {
+          (cd examples/$1 &&
+            (corepack pnpm run build -- --output-path ../../exbuild/$1 ||
+              (./build.sh && mkdir -p ../../exbuild/$1 && cp -r ./* ../../exbuild/$1 && rm ../../exbuild/$1/**/.gitignore))
           ) || exit 1;
-        done
+        }
+        export -f build
+        parallel -j 8 build ::: `ls examples | grep -v README | grep -v raytrace | grep -v deno | grep -v wasm-audio-worklet`
```
Systemcluster, line 375: It turned out to be really simple.

.github/workflows/main.yml
```diff
@@ -369,8 +385,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - run: rustup default nightly-2024-07-06
+    - run: rustup target add x86_64-unknown-linux-gnu
     - run: rustup target add wasm32-unknown-unknown
-    - run: rustup component add rust-src
+    - run: rustup component add rust-src --target x86_64-unknown-linux-gnu
+    - run: rustup component add rust-src --target wasm32-unknown-unknown
```
Systemcluster, line 391: I don't know why `build_nightly` fails now, and this change doesn't help.

Systemcluster:
> Currently similar tests are done in /crates/cli/tests/reference.
> Would it be possible to migrate the current similar tests to use insta as well?

There are other similar ones in the `xform` tests, but I'm not sure. I don't know what the `BLESS` environment variable is supposed to do exactly. From a glance I think `insta` can be used for the tests that compare against a reference; I'd prefer not to get into it though.

.github/workflows/main.yml
```diff
@@ -363,11 +363,11 @@ jobs:
     - run: |
         cargo build -p wasm-bindgen-cli
         ln -snf `pwd`/target/debug/wasm-bindgen $(dirname `which cargo`)/wasm-bindgen
-    - run: mv _package.json package.json && npm install && rm package.json
+    - run: mv _package.json package.json && corepack pnpm install && rm package.json
     - run: |
         function build() {
           (cd examples/$1 &&
-            (npm run build -- --output-path ../../exbuild/$1 ||
+            (corepack pnpm run build -- --output-path ../../exbuild/$1 ||
```
Systemcluster, line 370: `pnpm` resolves and installs generally much faster than `npm`.

.npmrc
```diff
@@ -0,0 +1,3 @@
+dedupe-direct-deps=true
+fund=false
+audit=false
```
Systemcluster, line 3: And this saves another second. Node package installation at 4 seconds should be safe to ignore now.

.github/workflows/main.yml
```diff
@@ -385,20 +402,30 @@ jobs:
...
+    - uses: actions/checkout@v4
     - uses: actions/download-artifact@v4
       with:
         name: examples2
         path: exbuild
     - run: rustup update --no-self-update stable && rustup default stable
-    - run: cargo test -p example-tests
+    - run: cargo test -p example-tests --no-default-features --features nightly
       env:
         EXBUILD: ${{ github.workspace }}/exbuild
```
daxpedda, line 430: This could be in a matrix as well.

.github/workflows/main.yml
daxpedda: In general, introducing any new dependencies have to be done in their own dedicated PRs.

daxpedda (changes requested):
> > Currently similar tests are done in /crates/cli/tests/reference.
> > Would it be possible to migrate the current similar tests to use insta as well?
> 
> There are other similar ones in the `xform` tests, but I'm not sure. I don't know what the `BLESS` environment variable is supposed to do exactly. From a glance I think `insta` can be used for the tests that compare against a reference; I'd prefer not to get into it though.

`BLESS` fixes the references to adhere to the output instead of checking against it. This is used when the output is actually modified and you want to update all the references. Its like `INSTA_UPDATE="always"` I assume.

Unless there is a good reason to introduce a second testing system instead of using the one in place, I don't want to further fragment our testing setup.
I am happy too replace the system we currently have in place, but having multiple once in place at the same time introduces friction into the contributor experience.

---

CI improvements should be done in separate PR.

.github/workflows/main.yml
Systemcluster: It's not a new dependency, it ships with node (as does `yarn`) pre-installed.

.github/workflows/main.yml
```diff
@@ -122,6 +122,20 @@ jobs:
...
+      WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
+    steps:
+    - uses: actions/checkout@v4
+    - run: rustup update --no-self-update stable && rustup default stable
+    - run: rustup target add wasm32-unknown-unknown
+    - uses: actions/setup-node@v4
+      with:
+        node-version: '20'
+    - uses: ./.github/actions/setup-geckodriver
     - run: cargo test --target wasm32-unknown-unknown --features enable-interning
```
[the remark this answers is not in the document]
Systemcluster: Will do!

Systemcluster:
I pulled out the other CI changes, will send a PR with less duplication.

> Unless there is a good reason to introduce a second testing system instead of using the one in place, I don't want to further fragment our testing setup. [..]

I'll leave that to you or anyone else that wants to pick it up from here in that case, this would just be a small addition to make reviewing refactoring and target addition PRs easier anyway.

PR #4051, opened 2024-08-04
Add sccache to GitHub workflow

Systemcluster:
A lot of the jobs compile the same crates, both within the workflow and across PRs, so sccache might speed things up quite a bit.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,11 +18,17 @@ concurrency:
 permissions:
   contents: read #  to fetch code (actions/checkout)
 
+env:
+  SCCACHE_GHA_ENABLED: "true"
+  RUSTC_WRAPPER: "sccache"
+
 jobs:
   # Check Code style quickly by running `rustfmt` over all code
   rustfmt:
     name: Rustfmt
     runs-on: ubuntu-latest
+    env:
+      RUSTC_WRAPPER: ""
     steps:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
@@ -33,6 +39,8 @@ jobs:
   check:
     name: Check
     runs-on: ubuntu-latest
+    env:
+      RUSTC_WRAPPER: ""
     steps:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
@@ -43,6 +51,8 @@ jobs:
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
+    env:
+      RUSTC_WRAPPER: ""
     steps:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
@@ -82,6 +92,7 @@ jobs:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
     - run: rustup target add wasm32-unknown-unknown
+    - uses: mozilla-actions/sccache-action@v0.0.5
     - uses: actions/setup-node@v4
       with:
         node-version: '20'
@@ -108,6 +119,7 @@ jobs:
     - run: rustup default nightly-2024-07-06
     - run: rustup target add wasm32-unknown-unknown
     - run: rustup component add rust-src
+    - uses: mozilla-actions/sccache-action@v0.0.5
     # Note: we only run the browser tests here, because wasm-bindgen doesn't support threading in Node yet.
     - run: |
         RUSTFLAGS='-C target-feature=+atomics,+bulk-memory,+mutable-globals' \
@@ -143,6 +155,7 @@ jobs:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
     - run: rustup target add wasm32-unknown-unknown
+    - uses: mozilla-actions/sccache-action@v0.0.5
     - uses: actions/setup-node@v4
       with:
         node-version: '20'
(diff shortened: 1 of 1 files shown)


daxpedda:
While I appreciate the effort, I'm have to be cautious about adding new dependencies, even if just CI dependencies, that I know nothing about.

I believe that most of the slowness of the CI comes from a lack of parallelization. Some jobs compile a lot of things between every step, if this could be adjusted with a GitHub Actions matrix, it would speed things up tremendously.
Only then I would consider pulling in foreign dependencies to speed it up further.

With that in mind, I will go ahead and close this.

---

On a side note, it would help tremendously if you put an explanation in the OP, otherwise I will have to spend significant time researching whats going in this PR, even though you've already done that, which increases the already pretty high maintenance burden of `wasm-bindgen` for me.

Systemcluster:
> On a side note, it would help tremendously if you put an explanation in the OP

I went with the assumption that sccache is a well-known entity, apologies for the missing explanation - There wasn't a lot of effort in this PR. 

Sccache stores compilation sub-artifacts and stores them in the GitHub cache to speed up the upcoming compilations. I'm using it in all my GitHub workflows, using my [custom setup-rust action](https://github.com/Systemcluster/actions/tree/main/actions/setup-rust) not the official action, but the official action seemed like the least amount of change. I'm also using it on all my local machines.

> Only then I would consider pulling in foreign dependencies to speed it up further.

It's a fair point. I think of sccache as trusted - a 11 year old Mozilla project that is consistently maintained and updated. But I understand.

> I believe that most of the slowness of the CI comes from a lack of parallelization.

There's a lot of parallelization already, but more would of course help more. Another option that would also help would be to simply cache the Rust directories, which is also a very common approach.

Could you re-open this PR (as draft) so I can kick off a few test commits to compare runtime changes?

daxpedda:
> > I believe that most of the slowness of the CI comes from a lack of parallelization.
> 
> There's a lot of parallelization already, but more would of course help more.

If you take a look at the [last run](https://github.com/rustwasm/wasm-bindgen/actions/runs/10239220852), you can see that most jobs are around the 2 minute mark.

Taking a look at anything above that:
- [Clippy](https://github.com/rustwasm/wasm-bindgen/blob/09e0e6d28b769141fde6100b6785dfdb40d2fe8f/.github/workflows/main.yml#L43-L74): could easily be parallelized dropping it below the 2 minute mark. In this case not all steps require re-compilation because they have similar dependencies, but e.g. mixing `wasm-bindgen-cli` with `web-sys` is a terrible idea. I would at least split up the different targets. Examples should also be tested in a separate job.
- [wasm-bindgen crate tests](https://github.com/rustwasm/wasm-bindgen/blob/09e0e6d28b769141fde6100b6785dfdb40d2fe8f/.github/workflows/main.yml#L76-L101): this is the worst offender with 11 minutes, could easily be put into a matrix to get it below 3 minutes.
- [web-sys crate tests](https://github.com/rustwasm/wasm-bindgen/blob/09e0e6d28b769141fde6100b6785dfdb40d2fe8f/.github/workflows/main.yml#L159-L177): Same as with the `wasm-bindgen` tests.
- [build_examples](https://github.com/rustwasm/wasm-bindgen/blob/09e0e6d28b769141fde6100b6785dfdb40d2fe8f/.github/workflows/main.yml#L255-L282): The main slowness here comes from running Webpack. Unfortunately I know nothing about Webpack, maybe a solution would be to run it in parallel on different threads with a bunch of Bash.
- [dist_windows](https://github.com/rustwasm/wasm-bindgen/blob/09e0e6d28b769141fde6100b6785dfdb40d2fe8f/.github/workflows/main.yml#L396-L407): Not much we can do here, probably exploring cross compilation to bring it in line with the Linux targets.

I believe this should bring our CI time down from 11.5 minutes to 3 minutes, if we can solve the [`dist_windows`] job, if not its down to 4.5 minutes. This is what I believe it comes down to: unless we can shorten the time of the slowest job, making faster jobs run even faster doesn't change the CI runtime.

> Could you re-open this PR (as draft) so I can kick off a few test commits to compare runtime changes?

Unfortunately I'm unable too:
image

Systemcluster:
I've been experimenting with a lot¹ of different setups here: https://github.com/Systemcluster/wasm-bindgen/actions

I was about to clean up the branches and send a PR with a more detailed analysis, but you beat me to it 😄 

I got it down to a pretty much [exactly 6 minutes](https://github.com/Systemcluster/wasm-bindgen/actions/runs/10263328389), although on the GitHub Actions runners it fluctuates a bit. If you got it down to 3 it was wasted effort, otherwise I can clean it up and send a PR so we can compare. You have much more knowledge about the project as a whole, I've just been going off the runtime numbers.

The 6 minutes come from `build_examples` taking 4 minutes and then `test_examples`, which depends on it, taking about two. `build_examples` is a bash loop that I didn't touch. I assume GNU `parallel` could work there, but I never used it before.

¹I deleted the older runs, but the last build is `#55`

> Unfortunately I'm unable too:

I didn't know, but good to be aware!

daxpedda:
> If you got it down to 3 it was wasted effort, otherwise I can clean it up and send a PR so we can compare.

Apologies for the misunderstanding, I got nothing!
I was just theorizing what should be possible :sweat_smile:!

> The 6 minutes come from `build_examples` taking 4 minutes and then `test_examples`, which depends on it, taking about two. `build_examples` is a bash loop that I didn't touch.

Yeah this has become the longest running job indeed. Parallelizing it would probably be the thing to do at this point, no clue about GNU `parallel` as well.

After that the second longest one is "[Run wasm-bindgen crate tests with different features](https://github.com/Systemcluster/wasm-bindgen/actions/runs/10263328389/job/28394917776#logs)", which could easily be split up as well.

PR #4052, opened 2024-08-06
Fix flaky tests

Systemcluster:
I don't know much about `hyper` but searching for the "error shutting down connection" error message in its code led me to attempt the `graceful_shutdown` addition, and I haven't had an example run fail since.

### DIFF
--- a/crates/example-tests/src/lib.rs
+++ b/crates/example-tests/src/lib.rs
@@ -354,14 +354,17 @@ pub async fn test_example(
             let conn = builder.serve_connection(TokioIo::new(stream), &service);
             tokio::pin!(conn);
 
-            tokio::select! {
+            let ret = tokio::select! {
                 res = conn.as_mut() => {
                     res.map_err(|e| anyhow::Error::msg(e.to_string()))
                 }
                 _ = rx => {
                     Ok(())
                 }
-            }
+            };
+
+            conn.graceful_shutdown();
+            ret
         },
         async {
             #[derive(Deserialize)]


crates/js-sys/src/lib.rs
```diff
@@ -2835,6 +2835,7 @@ macro_rules! number_try_from {
             #[inline]
             fn try_from(x: $x) -> Result<Number, Self::Error> {
                 let x_f64 = x as f64;
+                #[allow(clippy::manual_range_contains)]
                 if x_f64 >= Number::MIN_SAFE_INTEGER && x_f64 <= Number::MAX_SAFE_INTEGER {
```
Systemcluster, line 2839: This errored when reorganizing test runs in the CI, I don't know why it doesn't error normally.

daxpedda: I'm assuming this doesn't affect the CI run of this PR?
If that's the case, could you address it when we actually get there?

I was already working separately on making sure Clippy runs properly on CI. Currently the Clippy run on CI skips a lot of code that otherwise would throw warnings, that should definitely be addressed.

daxpedda (changes requested):
Wow, thanks for looking into it, that's really helpful!

crates/js-sys/src/lib.rs
Systemcluster: I was wondering about that since I saw other clippy warnings that didn't error in the CI either, I just happened to run into this one. This PR doesn't need it.

daxpedda (approved):

Systemcluster:
It looks like this wasn't 100% successful, I'm [still getting test failures](https://github.com/Systemcluster/wasm-bindgen/actions/runs/10271950946/job/28423243152), albeit much less. I don't know enough about `hyper` and `tokio` to investigate this further, but maybe this can serve as a start.

PR #4056, opened 2024-08-06
Improve the CI runtime

Systemcluster:
Follow-up to #4051, see there for prior discussion.

This splits up the longest-running CI jobs into smaller jobs to improve the CI runtime.

The overall _usage_ didn't go down much, it's still around 100 minutes, but now it's spread out over more jobs.
(It's gotten a bit more stable slightly under the 100 minute mark with this PR, but not much.)

Additional changes are commented below.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,8 +124,8 @@ jobs:
     - run: cargo test --target wasm32-unknown-unknown --features serde-serialize
     - run: cargo test --target wasm32-unknown-unknown --features enable-interning
 
-  test_wasm_bindgen_envs:
-    name: "Run wasm-bindgen crate tests with various environment variables"
+  test_wasm_bindgen_wasm:
+    name: "Run wasm-bindgen wasm test"
     runs-on: ubuntu-latest
     env:
       WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
@@ -139,6 +139,19 @@ jobs:
     - run: cargo test --target wasm32-unknown-unknown --test wasm
       env:
         WASM_BINDGEN_NO_DEBUG: 1
+
+  test_wasm_bindgen_envs:
+    name: "Run wasm-bindgen crate tests with various environment variables"
+    runs-on: ubuntu-latest
+    env:
+      WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
+    steps:
+    - uses: actions/checkout@v4
+    - run: rustup update --no-self-update stable && rustup default stable
+    - run: rustup target add wasm32-unknown-unknown
+    - uses: actions/setup-node@v4
+      with:
+        node-version: '20'
     - run: cargo test --target wasm32-unknown-unknown
       env:
         WASM_BINDGEN_EXTERNREF: 1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -548,6 +548,7 @@ jobs:
     permissions:
       contents: write #  to push changes in repo (jamesives/github-pages-deploy-action)
 
+    if: github.repository == 'rustwasm/wasm-bindgen'
     needs:
       - doc_api
       - doc_book
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,7 @@ on:
     tags-ignore: [dev]
   pull_request:
     branches: [main]
+  workflow_dispatch:
 defaults:
   run:
     shell: bash
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,6 @@ jobs:
     - run: cargo clippy --no-deps --all-features -p example-tests -- -D warnings
     - run: cargo clippy --no-deps --all-features -p wasm-bindgen-externref-xform -- -D warnings
     - run: cargo clippy --no-deps --all-features --target wasm32-unknown-unknown -p wasm-bindgen-futures -- -D warnings
-    - run: cargo clippy --no-deps --all-features --target wasm32-unknown-unknown -p js-sys -- -D warnings
     - run: cargo clippy --no-deps --all-features -p wasm-bindgen-macro -- -D warnings
     - run: cargo clippy --no-deps --all-features -p wasm-bindgen-macro-support -- -D warnings
     - run: cargo clippy --no-deps --all-features -p wasm-bindgen-multi-value-xform -- -D warnings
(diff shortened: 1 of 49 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -2835,6 +2835,7 @@ macro_rules! number_try_from {
             #[inline]
             fn try_from(x: $x) -> Result<Number, Self::Error> {
                 let x_f64 = x as f64;
+                #[allow(clippy::manual_range_contains)]
                 if x_f64 >= Number::MIN_SAFE_INTEGER && x_f64 <= Number::MAX_SAFE_INTEGER {
```
Systemcluster, line 2839: Here it's necessary 😄

.github/workflows/main.yml
```diff
@@ -5,6 +5,7 @@ on:
     tags-ignore: [dev]
   pull_request:
     branches: [main]
+  workflow_dispatch:
```
Systemcluster, line 8: I added this to be able to test the CI easier but I don't see a reason not to have this trigger for this workflow.

crates/js-sys/src/lib.rs
daxpedda: Lets address it while we are here!
I think it should be `(Number::MIN_SAFE_INTEGER..=Number::MAX_SAFE_INTEGER).contains(&x_f64)`?

.github/workflows/main.yml
daxpedda: Sure!

daxpedda:
I'm happy to give it a final review already as-is, up to you when its ready!

I believe the last three things we could do are:
- Parallelize `build_examples` via bash somehow (or if you know your way around webpack maybe there is something there?).
- Remove the `test --all-features -p web-sys`, it can be tested without all features, but we have to add the required features to dev-dependencies. Checking building with all features is already done in another part of the CI.
- Cross-compile `dist_windows` from Linux, which should finally get us down to under 3 minutes (if the other two points work out!).

Could all be done in a follow-up btw.

crates/example-tests/src/lib.rs
```diff
@@ -364,6 +364,7 @@ pub async fn test_example(
             };
 
             conn.graceful_shutdown();
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
Systemcluster, line 367: It looks like it didn't work, maybe it's not (just) here after all?

.github/workflows/main.yml
```diff
@@ -258,6 +346,11 @@ jobs:
     - uses: actions/checkout@v4
     - run: rustup update --no-self-update stable && rustup default stable
     - run: rustup target add wasm32-unknown-unknown
+    - uses: Swatinem/rust-cache@v2
+      with:
+        key: build-examples
+        cache-targets: |
+          debug
```
Systemcluster, line 353: This improved the runtime of the `build_benchmarks` job from `3m 57s` to `50s` in the [last run](https://github.com/Systemcluster/wasm-bindgen/actions/runs/10272820093/usage).

_package.json
```diff
@@ -5,12 +5,11 @@
...
-    "text-encoding": "^0.7.0",
-    "webpack": "^5.49.0",
-    "webpack-cli": "^4.7.2",
-    "webpack-dev-server": "^3.11.2"
+    "css-loader": "^7.1.2",
+    "html-webpack-plugin": "^5.6.0",
+    "mini-css-extract-plugin": "^2.9.0",
+    "webpack": "^5.93.0",
+    "webpack-cli": "^5.1.4",
+    "webpack-dev-server": "5.0.4"
```
Systemcluster, line 13: I updated them in an attempt to improve the time, it didn't help _much_, but I don't think it hurts much to be on the recent versions.

examples/add/webpack.config.js
```diff
@@ -14,12 +14,6 @@ module.exports = {
         new WasmPackPlugin({
             crateDirectory: path.resolve(__dirname, ".")
         }),
-        // Have this example work in Edge which doesn't ship `TextEncoder` or
-        // `TextDecoder` at this time.
-        new webpack.ProvidePlugin({
-          TextDecoder: ['text-encoding', 'TextDecoder'],
-          TextEncoder: ['text-encoding', 'TextEncoder']
-        })
```
Systemcluster, line 22: This shim has been marked deprecated and hasn't been useful for a while.

Systemcluster (commented):
I didn't get it under 6 minutes, but the later commits touch up some details that might still be worthwile.

I'm done with the CI for a while, feel free to modify and force push this branch if something should change. I'll leave any additional improvements for you or someone else (or for me at some later time 🙂).

examples/add/webpack.config.js
```diff
@@ -14,12 +14,6 @@ module.exports = {
...
     ],
     mode: 'development',
```
Systemcluster, line 18: Maybe it would be faster in `release` mode?

Systemcluster: It's called `production` and it's [about the same](https://github.com/Systemcluster/wasm-bindgen/actions/runs/10273924466/usage). Both `test_examples` as well as `build_examples` stays the same within fluctuations.

_package.json
daxpedda: Thats fine!
Thank you!

.github/workflows/main.yml
```diff
@@ -459,12 +552,13 @@ jobs:
...
     - run: rustup update --no-self-update 1.76 && rustup default 1.76
     - run: cargo build
-      
+
 
   deploy:
     permissions:
       contents: write #  to push changes in repo (jamesives/github-pages-deploy-action)
 
+    if: github.repository == 'rustwasm/wasm-bindgen'
```
daxpedda, line 561: Why?
This didn't make any difference in the CI time.

.github/workflows/main.yml
daxpedda: Lets add this in a dedicated PR please.
In any case, as long as we don't get the other jobs down in time this won't make a difference.

.github/workflows/main.yml
Systemcluster: It fails when running in other repositories: https://github.com/Systemcluster/wasm-bindgen/actions/runs/10274428735/job/28431248685

daxpedda: Ah, you are checking the repo, apologies, didn't read carefully enough.

.github/workflows/main.yml
daxpedda: In that follow-up, I would say just slap it on everything, because we are running into the open-source runner limit already.

daxpedda (changes requested):
"[Run wasm-bindgen crate tests with various environment variables](https://github.com/rustwasm/wasm-bindgen/actions/runs/10274290866/job/28431141019#logs)" could still be split up, that ones the longest now.

I'm happy to merge it as-is if we can split out the caching.

.github/workflows/main.yml
Systemcluster: That might be on me, I'm sorry for that. I didn't consider that runs in forks count against the origin limits. 
I wish there was a better alternative to test the CI, `act` is a bit awkward for this repo (and in general). I guess next time I'll run the CI in a private reuplod when it comes up. I hope it's only short term?

daxpedda: > I didn't consider that runs in forks count against the origin limits.

Wow ... that would be surprising ...

> I hope it's only short term?

Sorry, what is only short term?

Systemcluster: > Sorry, what is only short term?

Runner limits? I don't know which limits you meant.

> Wow ... that would be surprising ...

I'm trying to find a reference but I can't, I must have misremembered. I've stared at the CI for too long today I think. 😄

daxpedda: > > Sorry, what is only short term?
> 
> Runner limits? I don't know which limits you meant.

Ah, the limit is basically how many runners can run at the same time, that's always present unless you pay.

daxpedda (approved):
Thanks for all the work!

.github/workflows/main.yml
Systemcluster: There's another limit that's usage hours per month, which would be a bit more uncomfortable to hit. Good if it's the other one.

Systemcluster:
> Parallelize build_examples via bash somehow (or if you know your way around webpack maybe there is something there?).

I know that there are many faster webpack alternatives (`rollup`, `rspack` or `parcel` as bundlers, or maybe bare `swc` or `esbuild` can do it as well) that might be easy to switch to, I haven't looked at the webpack setup in detail to be able to tell.

I assume just running the loop in parallel is a bit easier though. `parallel` or something [like this](https://github.com/aaronriekenberg/rust-parallel) would surely work.

Systemcluster:
I need to add (with this as a follow-up, and it only showing in linked CI runs) that adding sccache really doesn't seem to help much with the `wasm-bindgen` CI. The jobs it speeds up aren't a bottleneck, and it doesn't seem to speed up the slow jobs for various reasons. It also doesn't reduce the 100 minute usage much. 
@daxpedda's comment in the closed PR is on point, other gains should be prioritized, unless I missed some configuration that would improve sccache here.

PR #4058, opened 2024-08-08
Improve the CI runtime, take two

Systemcluster:
Split from #4047, see there for some earlier review.

It looks like the CI takes about 5 minutes with this PR, one minute down.


.github/workflows/main.yml
```diff
@@ -349,14 +302,16 @@ jobs:
     - run: |
         cargo build -p wasm-bindgen-cli
         ln -snf `pwd`/target/debug/wasm-bindgen $(dirname `which cargo`)/wasm-bindgen
-    - run: mv _package.json package.json && npm install && rm package.json
+    - run: mv _package.json package.json && corepack pnpm install && rm package.json
```
Systemcluster, line 305: I left the `packageManager` field out of the `package.json` files compared to the commit in #4047 so it always uses the latest one. It prints this in the CI now, which I think is ok:
> The local project doesn't define a 'packageManager' field. Corepack will now add one referencing [current version]

.github/workflows/main.yml
```diff
@@ -349,14 +302,16 @@ jobs:
...
-              (./build.sh && mkdir -p ../../exbuild/$dir && cp -r ./* ../../exbuild/$dir && rm ../../exbuild/$dir/**/.gitignore))
+        function build() {
+          (cd examples/$1 &&
+            (corepack pnpm run build -- --output-path ../../exbuild/$1 ||
+              (./build.sh && mkdir -p ../../exbuild/$1 && cp -r ./* ../../exbuild/$1 && rm ../../exbuild/$1/**/.gitignore))
           ) || exit 1;
-        done
+        }
+        export -f build
+        ls examples | grep -v README | grep -v raytrace | grep -v deno | grep -v wasm-audio-worklet | parallel -j4 build
```
Systemcluster, line 314: As mentioned in the other PR, more parallelization doesn't help much because Cargo just gets blocked by other Cargo instances. I haven't fine-tuned the parallelization but `-j8` was slower than `-j4`.

Systemcluster:
Latest run was actually slower, but it looks like `cargo build -p wasm-bindgen-cli` in the latest `build_examples` run was just uncharacteristically slow.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - "examples/*"
```
Systemcluster, line 2: Having all examples in a shared workspace allows resolving all packages at once.

.github/workflows/main.yml
```diff
@@ -385,20 +338,30 @@ jobs:
...
+    - uses: actions/checkout@v4
     - uses: actions/download-artifact@v4
       with:
         name: examples2
         path: exbuild
     - run: rustup update --no-self-update stable && rustup default stable
-    - run: cargo test -p example-tests
+    - run: cargo test -p example-tests --no-default-features --features nightly
       env:
         EXBUILD: ${{ github.workspace }}/exbuild
```
daxpedda, line 366: I think this would benefit from a matrix:
```yml
  test_examples:
    matrix:
    - needs: build_examples
      artifact: examples1
    - needs: build_nightly
      artifact: examples2
      args: --no-default-features --features nightly
    needs:
    - ${{ matrix.needs }}
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v4
    - uses: actions/download-artifact@v4
      with:
        name: ${{ matrix.artifact }}
        path: exbuild
    - run: cargo test -p example-tests ${{ matrix.args }}
      env:
        EXBUILD: ${{ github.workspace }}/exbuild
```

Might not work because of `needs`.

.github/workflows/main.yml
```diff
@@ -215,61 +215,44 @@ jobs:
...
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-    - run: rustup update --no-self-update stable && rustup default stable
-    - run: cd crates/web-sys && cargo run --release --package wasm-bindgen-webidl -- webidls src/features ./Cargo.toml
-    - run: git diff --exit-code
-
-  test_js_sys:
-    name: "Run js-sys crate tests"
+  test_node:
```
daxpedda, line 218:
```suggestion
  test_crates:
```
Can't come up with anything better.

.github/workflows/main.yml
```diff
@@ -215,61 +215,44 @@ jobs:
...
+    strategy:
+      matrix:
+        runs:
+          - name: "web-sys"
+            run: |
+              cargo build --manifest-path crates/web-sys/Cargo.toml --target wasm32-unknown-unknown
+              cargo build --manifest-path crates/web-sys/Cargo.toml --target wasm32-unknown-unknown --features Node
+              cargo build --manifest-path crates/web-sys/Cargo.toml --target wasm32-unknown-unknown --features Element
+              cargo build --manifest-path crates/web-sys/Cargo.toml --target wasm32-unknown-unknown --features Window
+          - name: "js-sys (all features)"
```
daxpedda, line 228:
```suggestion
          - name: "web-sys (all features)"
```

.github/workflows/main.yml
```diff
@@ -215,61 +215,44 @@ jobs:
...
+            run: cargo test --manifest-path crates/web-sys/Cargo.toml --target wasm32-unknown-unknown --all-features
+          - name: "js-sys (unstable, all features)"
```
daxpedda, line 230:
```suggestion
          - name: "web-sys (unstable, all features)"
```

.github/workflows/main.yml
```diff
@@ -215,61 +215,44 @@ jobs:
...
-        RUSTFLAGS: --cfg=web_sys_unstable_apis
-
-  check_web_sys:
-    name: "Verify that web-sys is compiled correctly"
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-    - run: rustup update --no-self-update stable && rustup default stable
-    - run: cd crates/web-sys && cargo run --release --package wasm-bindgen-webidl -- webidls src/features ./Cargo.toml
-    - run: git diff --exit-code
```
daxpedda, line 269: This test got a bit lost in translation.

daxpedda (changes requested):
Again, thank you for doing this, its a much needed improvement.

.github/workflows/main.yml
Systemcluster: These all use the webdriver so maybe `test wasm-bindgen with webdriver (test name)`?

.github/workflows/main.yml
Systemcluster: I think `-j4` saved about 20 to 30 seconds, but I'll compare directly.

Otherwise I think it's more readable as a function even if we remove the `parallel` for now.

.github/workflows/main.yml
```diff
@@ -349,14 +299,17 @@ jobs:
     - run: |
         cargo build -p wasm-bindgen-cli
         ln -snf `pwd`/target/debug/wasm-bindgen $(dirname `which cargo`)/wasm-bindgen
-    - run: mv _package.json package.json && npm install && rm package.json
+    - run: echo 'packages:\n  - "examples/*"' > pnpm-workspace.yaml && sed -i -e 's/\\n/\n/g' pnpm-workspace.yaml
```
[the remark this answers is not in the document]
Systemcluster: As far as I understand it, what it is currently doing is just installing the top level `_package.json` dependencies, and when running the packages it just locates their dependencies one level up. Meaning before #4056 it was using `webpack-dev-server` 3 even if the packages specified 4. It works because all their dependencies are covered by the top package dependencies.
I added `-r` here to also recursively install dependencies of packages in subdirectories, and with the workspace we restrict it to the `examples` directory and allow them to share the same `node_modules`, basically intentionally creating the behavior of before, but actually resolving their listed dependencies. (You could say this is a "fix" and technically you don't need the top level `_package.json` anymore after this.)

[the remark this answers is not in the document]
Systemcluster: Unless you mean the use of `sed`, then sure. I didn't find a way to directly input newlines into the file - with `zsh` it just works to use `\n` but `bash` always escaped it. I'll add that to the comment too, unless you know a way to do this without `sed`.

.github/workflows/main.yml
Systemcluster: Doesn't work because of `needs` if I can trust the VSCode extension's validator.

.github/workflows/main.yml
Systemcluster: > I think this just needs to be split up between Cargo and Webpack runs.

It looks like most of that happens through `wasm-pack` so that's not that simple.

.github/workflows/main.yml
```diff
@@ -215,61 +215,44 @@ jobs:
...
+            env:
+              RUSTFLAGS: --cfg=js_sys_unstable_apis
+          - name: "wasm-bindgen-webidl"
+            run: cargo test -p wasm-bindgen-webidl
+          - name: "webidl-tests"
+            run: cargo test -p wasm-bindgen-webidl --target wasm32-unknown-unknown
+            env:
+              WBINDGEN_I_PROMISE_JS_SYNTAX_WORKS_IN_NODE: 1
+          - name: "webidl-tests (unstable)"
+            run: cargo test -p wasm-bindgen-webidl --target wasm32-unknown-unknown
```
daxpedda, line 247:
```suggestion
            run: cargo test -p webidl-tests --target wasm32-unknown-unknown
            env:
              WBINDGEN_I_PROMISE_JS_SYNTAX_WORKS_IN_NODE: 1
          - name: "webidl-tests (unstable)"
            run: cargo test -p webidl-tests --target wasm32-unknown-unknown
```

daxpedda (changes requested):

daxpedda (approved):
Thanks!

Systemcluster:
Thanks for the review!

PR #4060, opened 2024-08-10
Experimental test coverage support

### LINKED ISSUE #2276: Code Coverage in wasm-bindgen-test?

### LINKED ISSUE #3774: [Proposal] Capture coverage information in wasm-pack test

daxpedda:
This PR adds experimental test coverage support facilitated by [`minicov`](https://crates.io/crates/minicov) and by [Hacken OÜ](https://hknio.github.io/wasmcov) finding out all the necessary hacks to make this work.

The feature requires nightly and is enabled by using `cfg(wasm_bindgen_unstable_test_coverage)`, for more information see the guide I added in the PR.

Thank you @aDogCalledSpot for the initial implementation which I adopted here.
Thank you @njelich for pioneering and documenting all this.

Fixes #2276.
Fixes #3774.
Replaces #3782.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,9 @@
 * Added importing strings as `JsString` through `#[wasm_bindgen(thread_local, static_string)] static STRING: JsString = "a string literal";`.
   [#4055](https://github.com/rustwasm/wasm-bindgen/pull/4055)
 
+* Added experimental test coverage support for `wasm-bindgen-test-runner`, see the guide for more information.
+  [#4060](https://github.com/rustwasm/wasm-bindgen/pull/4060)
+
 ### Changed
 
 * Stabilize Web Share API.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,11 @@ serde_derive = "1.0"
 wasm-bindgen-test-crate-a = { path = 'tests/crates/a' }
 wasm-bindgen-test-crate-b = { path = 'tests/crates/b' }
 
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(wasm_bindgen_unstable_test_coverage)',
+] }
+
 [workspace]
 members = [
   "benchmarks",
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -220,6 +220,7 @@ impl ToTokens for ast::Struct {
         (quote! {
             #[automatically_derived]
             impl #wasm_bindgen::describe::WasmDescribe for #name {
+                #[cfg_attr(wasm_bindgen_unstable_test_coverage, coverage(off))]
                 fn describe() {
                     use #wasm_bindgen::__wbindgen_if_not_std;
                     use #wasm_bindgen::describe::*;
@@ -293,6 +294,7 @@ impl ToTokens for ast::Struct {
                 #[doc(hidden)]
                 // `allow_delayed` is whether it's ok to not actually free the `ptr` immediately
                 // if it's still borrowed.
+                #[cfg_attr(wasm_bindgen_unstable_test_coverage, coverage(off))]
                 pub unsafe extern "C" fn #free_fn(ptr: u32, allow_delayed: u32) {
                     use #wasm_bindgen::__rt::alloc::rc::Rc;
 
@@ -401,6 +403,7 @@ impl ToTokens for ast::Struct {
             }
 
             impl #wasm_bindgen::describe::WasmDescribeVector for #name {
+                #[cfg_attr(wasm_bindgen_unstable_test_coverage, coverage(off))]
                 fn describe_vector() {
                     use #wasm_bindgen::describe::*;
                     inform(VECTOR);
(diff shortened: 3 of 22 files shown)


aDogCalledSpot:
Thanks for taking the time for this! I unfortunately had to move to a different project and was never able to finish this. Good to see it being picked up though <3

njelich:
@daxpedda Nice work. Happy to see this implemented :)

Issue #2265, opened 2020-07-30
One generated wasm to rule them all


holygits:
### Motivation

Create a single, isomorphic JS package using wasm-bindgen and tooling.

### Proposed Solution

Generate the wasm import table in a consistent or configurable way so that the wasm binary may be used
by any JS target runtime.

Maybe there is a way to provide a core wasm file with the main logic and generate only "glue" wasm files for each target which provide the right import shims.

### Alternatives

It is possible to do this [using wasm-pack or manual build scripts and other glue code](https://github.com/rustwasm/wasm-pack/pull/705) the downside of this, in my understaind, is it requires multiple wasm blobs, one for each target, which unnecessary bloats the final package size.

### Additional Context

Idea came from the observation that the generated wasm for `--target=node` and `--target=browser` produce
the same binary apart from imports table (converted to wat):
[<img width="1151" alt="Screen Shot 2020-07-31 at 10 14 11 AM" src="https://user-images.githubusercontent.com/5133901/88980424-ccafc000-d317-11ea-9f73-b4773a529c5b.png">](url)

I don't know dark js arts so maybe there's any easier way to do this from JS code but I could not find that in my research into this.

alexcrichton:
This is theoretically possible but not a lot of effort has gone into this. There's various subtle interactions with how the target and the generated JS affects the resulting wasm module, for example precisely what intrinsics are exported and not. Currently there's not a pass in wasm-bindgen that attempts to export everything it would need for all targets, so support would need to be added for that.

torfmaster:
FYI: I created a draft PR here which might touch this issue: https://github.com/rustwasm/wasm-bindgen/pull/4065

Issue #3790, opened 2024-01-17
Sharing the same WebAssembly file across multiple targets


magic-akari:
### Summary

Compiling wasm for multiple targets is a common scenario. 
Is it possible to share the same wasm and only have different JavaScript glue code?

### Additional Details

related issue:
- JohnnyMorganz/StyLua#803

daxpedda:
I honestly don't know as I only use the Web target, but I don't remember any Wasm changes related to the target being made by `wasm-bindgen-cli`.

A very simple way to find out is just give it a try, if the Wasm file produced by `wasm-bindgen-cli` is byte-equivalent between different targets it should be fine.

Swolebrain:
I looked at implementing this but since most people use wasm-bindgen through wasm-pack, it looked like it would be a huge rippling change. 

So I opted to hack around it by creating a build script that builds all 3 targets, moves them into subdirectories of my `pkg` folder, and then arranges for the root `package.json` to have subpackages. You can take a look at this for reference https://github.com/cedar-policy/cedar/blob/main/cedar-wasm/build-wasm.sh

magic-akari:
This is an attempt to share a wasm file across multiple compilation targets.

[wasm-bindgen-proposal.tar.gz](https://github.com/user-attachments/files/16216076/wasm-bindgen-proposal.tar.gz)


## 1. Project Structure

Under the src directory, `lib.rs` is the source code, and the file content is quite simple.

```rust
use wasm_bindgen::prelude::*;
#[wasm_bindgen]
pub fn err() -> Result<(), u8> {
    Err(1)
}
```

This is the most basic example I've found where the WASM imports include an additional function, which we need to address next.

## 2. Build

Executing `build.sh` will generate the wasm file, and the content of `build.sh` is as follows.

```bash
# web
wasm-pack build --target=web --out-dir=web
wasm2wat web/wasm_bindgen_proposal_bg.wasm -o web/wasm_bindgen_proposal_bg.wat

# nodejs
wasm-pack build --target=nodejs --out-dir=nodejs
wasm2wat nodejs/wasm_bindgen_proposal_bg.wasm -o nodejs/wasm_bindgen_proposal_bg.wat

# bundler
wasm-pack build --target=bundler --out-dir=bundler
wasm2wat bundler/wasm_bindgen_proposal_bg.wasm -o bundler/wasm_bindgen_proposal_bg.wat
```

[...]

PR #4065, opened 2024-08-12
PoC for inline target

torfmaster:
This PR adds an "isomorphic" or "inline" target next to the nodejs target as a proof of concept. It targets simple use cases as using small snippets of Rust Code in larger typescript codebases. In detail it allows (with a proposal that I will create for wasm-pack) creating npm packages that ship wasm modules inline.

# Background

Using Rust code via wasm in existing nodejs typescript applications is challenging. It usually involves changing the bundler configuration and currently involves handling wasm artifacts.
The motivating example for a typescript lambda built via nx and bundled via esbuild (which is the de-facto standard for AWS Lambda in CDK). After a lot of experimentation (involving using different bundlers and writing plugins) I found no obvious way for my rather easy use case (a few lines of Rust Code embedded via wasm). Hence I

# Relation to other issues

This touches at least https://github.com/rustwasm/wasm-bindgen/issues/2265, https://github.com/rustwasm/wasm-pack/issues/1334, https://github.com/rustwasm/wasm-pack/issues/831.

# Testing

I currently only did some rough manual testing. Once I know that this PR is somehow useful I can invest more.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -10,6 +10,7 @@ use crate::wit::{JsImport, JsImportName, NonstandardWitSection, WasmBindgenAux};
 use crate::{reset_indentation, Bindgen, EncodeInto, OutputMode, PLACEHOLDER_MODULE};
 use anyhow::{anyhow, bail, Context as _, Error};
 use binding::TsReference;
+use file_util::create_load_inline_bytes_snippet;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt;
@@ -19,6 +20,7 @@ use std::path::{Path, PathBuf};
 use walrus::{FunctionId, ImportId, MemoryId, Module, TableId, ValType};
 
 mod binding;
+pub(crate) mod file_util;
 
 pub struct Context<'a> {
     globals: String,
@@ -308,19 +310,14 @@ impl<'a> Context<'a> {
     fn generate_inline_wasm_loading(&mut self) -> String {
         let mut shim = String::new();
 
-        let buf = self.module.emit_wasm();
+        let wasm = self.module.emit_wasm();
 
-        let mut serialized = "const bytes = new Uint8Array([".to_string();
-        let (last, bytes) = buf.split_last().unwrap();
-        for byte in bytes {
-            serialized.push_str(&format!("{},", byte));
-        }
-        serialized.push_str(&format!("{}", last));
-        serialized.push_str("]);");
+        let serialized = create_load_inline_bytes_snippet(&wasm, "bytes".into());
+
         shim.push_str(&serialized);
         shim.push_str(
             "
-                const wasmModule = new WebAssembly.Module(bytes.buffer);
+                const wasmModule = new WebAssembly.Module(bytes);
                 const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
                 wasm = wasmInstance.exports;
                 module.exports.__wasm = wasm;
--- a/crates/cli-support/src/js/file_util.rs
+++ b/crates/cli-support/src/js/file_util.rs
@@ -1,0 +1,17 @@
+use base64::{prelude::BASE64_STANDARD_NO_PAD, Engine};
+
+pub(crate) fn create_load_inline_bytes_snippet(bytes: &[u8], variable_name: String) -> String {
+    format!(
+        "
+        let {variable_name};
+        const base64 = \"{base64}\";
+        if (typeof Buffer === 'undefined') {{
+            {variable_name} = Uint8Array.from(atob(base64), c => c.charCodeAt(0));
+        }} else {{
+            {variable_name} = Buffer.from(base64, 'base64');
+        }}
+        ",
+        variable_name = variable_name,
+        base64 = BASE64_STANDARD_NO_PAD.encode(bytes),
+    )
+}
--- a/crates/cli-support/src/wasm2es6js.rs
+++ b/crates/cli-support/src/wasm2es6js.rs
@@ -1,8 +1,9 @@
 use anyhow::{bail, Error};
-use base64::{prelude::BASE64_STANDARD, Engine as _};
 use std::collections::HashSet;
 use std::fmt::Write;
 use walrus::Module;
+
+use crate::js::file_util::create_load_inline_bytes_snippet;
 
 pub struct Config {
     base64: bool,
@@ -232,18 +233,7 @@ impl Output {
         let wasm = self.module.emit_wasm();
         let (bytes, booted) = if self.base64 {
             (
-                format!(
-                    "
-                    let bytes;
-                    const base64 = \"{base64}\";
-                    if (typeof Buffer === 'undefined') {{
-                        bytes = Uint8Array.from(atob(base64), c => c.charCodeAt(0));
-                    }} else {{
-                        bytes = Buffer.from(base64, 'base64');
-                    }}
-                    ",
-                    base64 = BASE64_STANDARD.encode(&wasm)
-                ),
+                create_load_inline_bytes_snippet(&wasm, "bytes".into()),
                 inst,
             )
         } else if let Some(ref path) = self.fetch_path {


crates/cli/src/bin/wasm-bindgen.rs
```diff
@@ -110,6 +110,7 @@ fn rmain(args: &Args) -> Result<(), Error> {
             "nodejs" => b.nodejs(true)?,
             "deno" => b.deno(true)?,
             "experimental-nodejs-module" => b.nodejs_module(true)?,
+            "isomorphic" => b.isomorphic(true)?,
```
Systemcluster, line 113: Since this is a proof-of-concept I think `experimental-isomorphic` for the name makes more sense.

crates/cli-support/src/js/mod.rs
```diff
@@ -500,6 +525,32 @@ impl<'a> Context<'a> {
...
             }
+
+            OutputMode::Isomorphic => {
+                js.push_str(&self.generate_node_imports());
+
+                js.push_str("let wasm;\n");
+
+                for (id, js) in crate::sorted_iter(&self.wasm_import_definitions) {
+                    let import = self.module.imports.get_mut(*id);
+                    footer.push_str("\nmodule.exports.");
```
Systemcluster, line 536: The use-cases all expect CommonJS?

crates/cli-support/src/js/mod.rs
```diff
@@ -142,7 +142,7 @@ impl<'a> Context<'a> {
             self.globals.push_str(c);
         }
         let global = match self.config.mode {
-            OutputMode::Node { module: false } => {
+            OutputMode::Isomorphic | OutputMode::Node { module: false } => {
```
Systemcluster, line 145: Why the name `Isomorphic`?

Systemcluster (commented):
See my PR #4027 for the Node target and #2176/#3990 for the Deno target in regards to tests and other target-addition changes.

I was going to refactor the target generation a bit, but this change is small enough where it doesn't matter whether this PR or mine gets merged first. I'll ping you once I open it in either case.

crates/cli-support/src/js/mod.rs
torfmaster: Good point. Technically it also makes sense for ecmascript modules but I can't think of a use case.
[ ] check whether and how to support ecmascript modules

crates/cli/src/bin/wasm-bindgen.rs
torfmaster: - [ ] change cli option

crates/cli-support/src/js/mod.rs
torfmaster: It was introduced in https://github.com/rustwasm/wasm-pack/issues/1334 and I didn't question it, it was just a catchy name. The idea seems to be that the code should run unchanged (without polyfiils or other modification) in any context (node, browser, ...). But this goal is beyond what I am doing here. Here I just inline the assets that need (and effectively cannot atm) to be processed by a bundler.
I would rather call it `node-universal`. Basically because that's basically what it should do: You can build your Rust code with `wasm-pack` and include it in your typescript application (e.g. react, aws lambda, node express) without worrying about the bundling.
What do you think?

magic-akari: Sorry for disturbing you here.

I previously facilitated the sharing of cjs/esm/bundler for WASM through manual adjustments in this https://github.com/JohnnyMorganz/StyLua/pull/848 .
I also shared my concept in https://github.com/rustwasm/wasm-bindgen/issues/3790, which supports various environments, including browsers, Node, Deno, Bun, and bundlers.

I'm not certain about your reference to `include it in your typescript application (e.g. react, ` but include CJS into the web project is challenging. We faced this issue in https://github.com/swc-project/swc-playground/issues/42, and finallay, we had to create a separate esm to fulfill the requirements.

Inline WASM is viable, but I don't believe it can serve as a direct alternative to the current web/node/deno/bundler option. Instead, this option should complement them, being usable independently and applicable across all environments.

torfmaster: > Sorry for disturbing you here.
Thank you for your valuable insights!

> I previously facilitated the sharing the same WebAssembly file across of cjs/esm/bundler through manual adjustments in this [JohnnyMorganz/StyLua#848](https://github.com/JohnnyMorganz/StyLua/pull/848) . I also shared my concept in #3790, which supports various environments, including browsers, Node, Deno, Bun, and bundlers.
> 
> I'm not certain about your reference to `include it in your typescript application (e.g. react, ` but include CJS into the web project is challenging. We faced this issue in [swc-project/swc-playground#42](https://github.com/swc-project/swc-playground/issues/42), and finallay, we had to create a separate esm to fulfill the requirements.

Maybe I am missing something here (I am actually using wasm only in AWS Lambdas for now). Is there an issue with using the npm package generated by the `Isomorphic` target in an arbitrary react app?

> Inline WASM is viable, but I don't believe it can serve as a direct alternative to the current web/node/deno/bundler option. Instead, this option should complement them, being usable independently and applicable across all environments.
I totally agree. My focus is also on the inline part not on the "usability across targets". I have been watching the asset handling issue in `esbuild`, `webpack`, `rollup`, `wasm-pack`, `wasm-bindgen` and never came closer to the solution than writing custom plugins plus a bunch of scripts. So as an intermediate solution I would propose inline wasm until every target just speaks wasm as esm module either directly or with any bundler of my choice without even having to change configuration.

Summing this up, I think I should just focus on an "inline node" target in this PR, moving the portability issues to the future when someone actually needs this.

lgarron:
I'm sorry this isn't a very substantial contribution, but the name "isomorphic" is rather confusing to me. If I understand, the idea is to make the code identical for browsers and other JS engines (which I'd really love to see), whereas "isomorphic" refers in general to things that are equivalent but not identical? Would it make more sense to call the code something more like "cross-compatible"?

cryptoquick:
"cross-compatible" is not very specific. I would prefer not to invent a new industry term. This behavior is generally known within the developer community as isomorphic builds. Consider it a term of art.

lgarron:
Hmm, it seems you're right: https://en.wikipedia.org/wiki/Isomorphic_JavaScript

That also states:

> The naming of the term 'Isomorphic JavaScript' has been a matter of controversy.[[1]](https://en.wikipedia.org/wiki/Isomorphic_JavaScript#cite_note-:0-1) The term 'isomorphic' was first coined by Charlie Robbins from [Nodejitsu](https://en.wikipedia.org/wiki/Nodejitsu), in one of the company's blog posts.[[2]](https://en.wikipedia.org/wiki/Isomorphic_JavaScript#cite_note-2) Spike Brehm, a software engineer from [Airbnb](https://en.wikipedia.org/wiki/Airbnb), wrote another blog post using the same term.[[3]](https://en.wikipedia.org/wiki/Isomorphic_JavaScript#cite_note-3) However, others have proposed to use the term Universal JavaScript instead.[[1]](https://en.wikipedia.org/wiki/Isomorphic_JavaScript#cite_note-:0-1)[[4]](https://en.wikipedia.org/wiki/Isomorphic_JavaScript#cite_note-4)[[5]](https://en.wikipedia.org/wiki/Isomorphic_JavaScript#cite_note-5)

I certainly think "universal" is much less of a misnomer, but thanks for the context!

torfmaster:
I changed the target name to "inline"

* because it is less controversial
* it is more precise. I don't claim this code will work with web browsers at all

It would be great to have an "universal" target (or if you want to call it like this: isomorphic). But the required changes would be much bigger than this tiny pull request.

crates/cli-support/src/js/mod.rs
torfmaster: I decided not to use ecmascript modules for now as I don't see a use case for this. The right way of mixing ecmascript modules with wasm should be importing wasm as esm.

torfmaster:
Out of curiosity I also checked browser compatibility: after a tiny fix basic functionality (meaning you can go the happy path and npm i a wasm project in your react app) seems to be there in browsers as well. A PoC is here: https://github.com/torfmaster/inline-wasm-react.

crates/cli-support/src/js/mod.rs
```diff
@@ -291,6 +291,31 @@ impl<'a> Context<'a> {
...
+    fn generate_inline_wasm_loading(&mut self) -> String {
+        let mut shim = String::new();
+
+        let buf = self.module.emit_wasm();
+
+        let mut serialized = "const bytes = new Uint8Array([".to_string();
+        let (last, bytes) = buf.split_last().unwrap();
+        for byte in bytes {
+            serialized.push_str(&format!("{},", byte));
+        }
```
RunDevelopment, line 303: I know that this PR is a PoC, so please take this comment as me throwing around ideas.

---

I wonder if there's a better way to store the WASM bytes inline.

Especially in the web context, the time it takes to *parse* a JS file is significant. JS is an interpreted language, so the total time it takes to run a script is `parse time + exec time`. JS is a complex language and JS parsers are equally complex. This is why often times, storing JSON data inside JS files as 
```js
const data = {...}
```
is *slower* than 
```js
const data = JSON.parse('{...}')
```

JSON is a simpler language and so it can be parsed much faster than JS. Basically, the time it takes to parse a JS object literal with all of JS's complexity is often more than the time it takes to parse a JS string literal and then parsing the string with a JSON parser.

So I believe that using `new UInt8Array(JSON.parse("[1,2,3,...]"))` would be faster than the current approach. However, benchmarks need to be done.

There might also be more efficient ways (both in terms of runtime and data compression). Base64 was my first thought, but I'm not aware of any fast and compatible way to do that in pure JS. NodeJs has `Buffer.from("...", "base64")`, but then generated files would be truly nodejs-only.

RunDevelopment: I just found this, so base64 might be an option.

https://github.com/rustwasm/wasm-bindgen/blob/4765448545d7bfc0223fd52f25d966b0bfbf8004/crates/cli-support/src/wasm2es6js.rs#L235-L246

torfmaster: Thank you for your reply. I agree with your observation and with your fix. Hence, I would

 - write a polyfill for base64 decode
 - use this to decode the wasm binary blob

It shouldn't be to hard to accomplish and I will write the fix as soon as I find the time (essentially the time for testing).

torfmaster: I implemented the base64 alternative (re-using the existing code). It works well in nodejs as in the browser. I don't know yet how to easily measure the startup performance as for usual wasm payloads the contribution of parsing the js code seems to be below noise. Maybe, I'll do some experiments with larger (non-wasm) payloads.

jlucaso1:
Nice proposal! I've done something similar in https://github.com/WhiskeySockets/whatsapp-rust-bridge/blob/master/ts/index.ts

PR #4068, opened 2024-08-12
Make sure deploy CI job runs last

daxpedda:
Follow-up to #4058, where we forgot to adjust the deploy CI job to run last.


daxpedda:
Never mind, this was on purpose and its better this way, the deploy job only relies on what it needs to deploy.

Issue #4081, opened 2024-08-19
Sign is printed twice when formatting a negative `BigInt`


ryangoree:
### Describe the Bug

If you format a negative `BigInt` instance, the sign (`-`) gets printed twice.

### Steps to Reproduce

1. Create a negative `BigInt`
   ```rs
   let int = BigInt::from(-1);
   ```
2. Use `console_log!` macro to see it's formatted with 2 signs:
   ```rs
   console_log!("{}", int); // --1
   ```

Minimal test repo: https://github.com/ryangoree/js-sys-bigint-display-bug

### Expected Behavior

The sign would be printed once (`-1`).

### Actual Behavior

The sign is printed twice (`--1`).

ryangoree:
Looks like the problem is here:
https://github.com/rustwasm/wasm-bindgen/blob/ddceac747876d67fe52dafc90964f95741ae467d/crates/js-sys/src/lib.rs#L1496

The [docs for `pad_integral`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.pad_integral) state:
> str. The str should not contain the sign for the integer, that will be added by this method.

However `to_string_unchecked` calls `toString` on the bigint which returns a string *with* the sign:
https://github.com/rustwasm/wasm-bindgen/blob/ddceac747876d67fe52dafc90964f95741ae467d/crates/js-sys/src/lib.rs#L1350

PR #4082, opened 2024-08-19
Fix negative `BigInt` formatting

ryangoree:
Closes: #4081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * Fixed many proc-macro generated `impl` blocks missing `#[automatically_derived]`, affecting test coverage.
   [#4078](https://github.com/rustwasm/wasm-bindgen/pull/4078)
 
-* Fixed negative `BigInt` values being incorrectly formatted with 2 leading dashes.
+* Fixed negative `BigInt` values being incorrectly formatted with two minus signs.
   [#4082](https://github.com/rustwasm/wasm-bindgen/pull/4082)
 
 --------------------------------------------------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 * Fixed many proc-macro generated `impl` blocks missing `#[automatically_derived]`, affecting test coverage.
   [#4078](https://github.com/rustwasm/wasm-bindgen/pull/4078)
 
+* Fixed negative `BigInt` values being incorrectly formatted with 2 leading dashes.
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -1493,21 +1502,33 @@ impl fmt::Debug for BigInt {
 impl fmt::Display for BigInt {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad_integral(self >= &BigInt::from(0), "", &self.to_string_unchecked(10))
+        f.pad_integral(
+            self >= &BigInt::from(0),
+            "",
+            &self.abs().to_string_unchecked(10),
+        )
```
daxpedda, line 1509: The check `self >= &BigInt::from(0)` is being done twice now, probably the easiest thing here is to let `abs()` also return a `bool`.

daxpedda (changes requested):
This needs an entry in the changelog.
Thank you!

ryangoree:
> This needs an entry in the changelog. Thank you!

I added an entry to the top-level changelog which seemed to be the convention, but lmk otherwise.

crates/js-sys/src/lib.rs
```diff
@@ -1388,12 +1388,12 @@ impl BigInt {
             .unchecked_into()
     }
 
-    /// Returns the absolute value of this BigInt value.
-    fn abs(&self) -> Self {
+    /// Returns a tuple of this BigInt's absolute value along with a boolean indicating whether the BigInt was negative.
```
daxpedda, line 1391:
```suggestion
    /// Returns a tuple of this [`BigInt`]'s absolute value along with a
    /// [`bool`] indicating whether the [`BigInt`] was negative.
```

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
 * Fixed many proc-macro generated `impl` blocks missing `#[automatically_derived]`, affecting test coverage.
   [#4078](https://github.com/rustwasm/wasm-bindgen/pull/4078)
 
+* Fixed negative `BigInt` values being incorrectly formatted with 2 leading dashes.
```
daxpedda, line 17:
```suggestion
* Fixed negative `BigInt` values being incorrectly formatted with two minus signs.
```

daxpedda (approved):
Thank you!

PR #4088, opened 2024-08-21
Fix `BigInt` formatting negative check

ryangoree:
This is a follow-up to #4082 in which I mistakingly used `is_neg` for an argument called `is_nonnegative`.

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1391,10 +1391,10 @@ impl BigInt {
     /// Returns a tuple of this [`BigInt`]'s absolute value along with a
     /// [`bool`] indicating whether the [`BigInt`] was negative.
     fn abs(&self) -> (Self, bool) {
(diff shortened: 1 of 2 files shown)


crates/js-sys/src/lib.rs
```diff
@@ -1391,10 +1391,10 @@ impl BigInt {
     /// Returns a tuple of this [`BigInt`]'s absolute value along with a
     /// [`bool`] indicating whether the [`BigInt`] was negative.
     fn abs(&self) -> (Self, bool) {
-        if self >= &BigInt::from(0) {
-            (self.clone(), false)
-        } else {
+        if self < &BigInt::from(0) {
             (-self, true)
+        } else {
+            (self.clone(), false)
```
ryangoree, line 1397: I swapped these to make it more readable. Now if the condition is `true`, then the bool is `true`.

daxpedda (approved):
`is_nonnegative` ...
Thank you!

PR #4091, opened 2024-08-25
Fix emitted package.json structure

kawaemon:
Currently it emits incorrectly structured `package.json` like this:
```json
{
  "foo": "^0.21.1",
  "bar": "^0.21.4"
}
```

This PR fixes it to match [the spec](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#dependencies).
```json
{
  "dependencies": {
    "foo": "^0.21.1",
    "bar": "^0.21.4"
  }
}
```


daxpedda (approved):
Thanks!

PR #4094, opened 2024-08-26
Consistently capitalize “Wasm”

mgeisler:
According to https://webassembly.org/, WebAssembly should be abbreviated as “Wasm”, not “wasm” or “WASM”. So this is just the result of a bit of search and replace.


daxpedda (approved):
Thank you!

PR #4111, opened 2024-09-10
Support the Tail Call Proposal

CryZe:
This adds support for the Tail Call Proposal.


daxpedda (changes requested):
That looks great, thank you!
Could you add an entry into the changelog as well?

daxpedda (approved):

Issue #4117, opened 2024-09-18
Enhancement: generate Symbol.dispose on exported structs


H-Plus-Time:
### Motivation

The end-goal is drop-in integration with the Explicit Resource Management proposal (at Stage 3), with the nearer term goal of smoothing integration with tools that polyfill that.

The syntax it enables:
```javascript
const func = () => {
  // declare an instance of our wasm-bindgen'd struct, specifying that it should be cleaned up
  // at the end of the scope it's declared in.
  using foobar = new MyStruct();
  // do a bunch of stuff...
}
func();
// When func() (or any other scope that keyword is used in) finishes, MyStruct::free is called
```

At the moment, Deno, Bun and Typescript >= 5.2 support the above directly; Usage without the above sugar is a little underwhelming, but probably still useful as a standardized target:
```javascript
class MyStruct {
  free() {} // the usual wasm-bindgen methods elided for brevity
  [Symbol.dispose]() {
    this.free();
  }
}

const foo = new MyStruct();
foo[Symbol.dispose]();
// or, more usefully, use a disposable stack (supplied by a polyfill) to orchestrate 
// more elaborate resource management
const stack = new DisposableStack();
const bar = stack.use(new MyStruct());
const baz = stack.use(new OtherStruct(bar));
// free the above in reverse order
stack[Symbol.dispose](); 
``` 

### Proposed Solution

Include the `[Symbol.dispose]() { this.free() }` pattern on all generated classes, polyfilling via `Symbol.dispose ??= Symbol("Symbol.dispose");`.

Async dispose is *probably* irrelevant

### Alternatives

It's not terribly difficult for 3rd party authors to run the generated bindings through code transformation of their own.

### Additional Context

The ecmascript proposal, for reference: https://github.com/tc39/proposal-explicit-resource-management

daxpedda:
Closed by #4118.

PR #4118, opened 2024-09-18
Generate Symbol.dispose on classes

H-Plus-Time:
Implements https://github.com/rustwasm/wasm-bindgen/issues/4117 (Explicit Resource Management support).

The main things I can think of for additional tests here are:
1. A basic Deno-only test involving `using`.
2. The polyfill mechanism (not worth it given the simplicity)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 * Add bindings to `RTCRtpTransceiverDirection.stopped`.
   [#4102](https://github.com/rustwasm/wasm-bindgen/pull/4102)
 
+* Added experimental support for `Symbol.dispose` via `WASM_BINDGEN_EXPERIMENTAL_SYMBOL_DISPOSE`.
+  [#4118](https://github.com/rustwasm/wasm-bindgen/pull/4118)
+
 ### Fixed
 
 * Fixed linked modules emitting snippet files when not using `--split-linked-modules`.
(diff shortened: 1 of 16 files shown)


daxpedda (changes requested):
I didn't fully review this yet, but considering its not standard yet, this should be behind an experimental flag.

Otherwise I'm happy to include this.

H-Plus-Time:
Gating behind an experimental flag I think would be more trouble than it's worth, for a few reasons:
1. At worst, it is a no-op (Symbol.dispose exists uniquely, or not at all).
2. Integrating tools (wasm-pack, basically) would need to account for said flag.
3. While the overall proposal is at stage 3, Symbol.dispose is on the far side (in the sense that the symbol name and meaning have been settled for several years).

daxpedda:
> 1. At worst, it is a no-op (Symbol.dispose exists uniquely, or not at all).

My worry is that if something changes, for whatever reason, we would need a breaking change.

> 2. Integrating tools (wasm-pack, basically) would need to account for said flag.

Indeed. We could do it via an environment variable, like many other features.

> 3. While the overall proposal is at stage 3, Symbol.dispose is on the far side (in the sense that the symbol name and meaning have been settled for several years).

Unfortunately I'm not well versed in the JS ecosystem at all, my experience so far has been that things change a lot. From the looks of it I agree that this is very unlikely to change, but I don't want to bet on it.

H-Plus-Time:
Oh, right, environment variables - yep, I'm fine with that, provided it's discoverable.

daxpedda (approved):
I explicitly added `EXPERIMENTAL` to the name and added a changelog entry.

Thank you!

PR #4120, opened 2024-09-18
Firefox supports import in Worklet

padenot:
And has for some time now: https://bugzilla.mozilla.org/show_bug.cgi?id=1572644


padenot:
(the compile example works as is, obviously)

daxpedda (approved):
Thank you!

PR #4133, opened 2024-09-28
Enable transformations when detecting target features

daxpedda:
This is especially relevant for Rust v1.82, where the reference type and multivalue proposal will be enabled by default.


danielhjacobs:
How can I disable these transformations on stable Rust 1.82?

danielhjacobs:
See #4227

PR #4135, opened 2024-10-01
Improved doc comment formatting

RunDevelopment:
Changes:
- The stars have aligned. The standard way to format `/**`-style doc comments is to align the starting `*` of subsequent lines with the first `*` in `/**`. This is now the case everywhere.
- Do not emit empty doc comments.
- Properly indent doc comments in `.d.ts` files.

This PR only changes code for comment generation and indentation.

Motivation: I just did it because (as a user of wasm-bindgen) the generated doc comments irked me and because (as a first-time contributor) it seemed like an easy fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,9 @@
 * Fixed enum variant name collisions with object prototype fields.
   [#4137](https://github.com/rustwasm/wasm-bindgen/pull/4137)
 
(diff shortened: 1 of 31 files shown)


RunDevelopment:
Uhm, why are elements generated in a different order in CI? I used `BLESS=1` to get the reference files, so I don't understand why the CI gets a different result.

PR #4136, opened 2024-10-02
derive(Debug) for JsError

### LINKED ISSUE #4099: `wasm_bindgen::JsError` doesn't implement `Debug`
### Summary

`wasm_bindgen::JsError` doesn't implement `Debug`

### Additional Details

When I write unit test, I can't direct use unwrap or expect func, so I use match handle it, It's too long and ugly.
I want ask it can implement `Debug` or can some one tell me the best practice

rcrisanti:
This should allow using `.unwrap()` & `.expect()` on it, in unit tests for example.

Resolves #4099.


daxpedda (approved):
Thanks!

PR #4141, opened 2024-10-04
Add `TextDecoder.decode` to `IMMUTABLE_SLICE_WHITELIST`

bdbai:
According to [Encoding Standard](https://encoding.spec.whatwg.org/#dom-textdecoder-decode), the input buffer is not supposed to be modified by implementations, hence the input slice do not need to be mutable.


daxpedda (approved):
Thank you!

PR #4143, opened 2024-10-05
Add `MathMLElement` interface and mixins `ElementCSSInlineStyle`, `HTMLOrForeignElement`

Philipp-M:
This aims to add the `MathMLElement` interface, from https://searchfox.org/mozilla-central/source/dom/webidl/MathMLElement.webidl

I noticed it used mixins that were not there yet, which are also used for the interfaces `SVGElement` and `HTMLElement`, I've added them as well and removed the redundant attributes from interfaces that inherit from these. I can keep it more focused on only the `MathMLElement` addition if that is desired. Was the omittance of `ElementCSSInlineStyle` intentional?


daxpedda (changes requested):
> I noticed it used mixins that were not there yet, which are also used for the interfaces `SVGElement` and `HTMLElement`, I've added them as well and removed the redundant attributes from interfaces that inherit from these.

This is unfortunately a breaking change, so we can't do this yet.

---

We prefer getting the bindings from the specification instead of from browser implementations. See https://www.w3.org/TR/MathML2/appendixe.html#dom-bindings.IDLBinding.

A changelog entry is missing as well.

Philipp-M:
> This is unfortunately a breaking change, so we can't do this yet.

What's the policy here? Accumulate more breaking changes at once?

Though I think it's a "minor breaking change" if at all, everything should be working as expected, as the methods just shift towards the parent interfaces, and coercion should do the rest, I'm not sure if that's even considered a breaking change?

We could also leave the methods on the child interfaces there, and accept the duplicated methods, so that it's not a breaking change.

> We prefer getting the bindings from the specification instead of from browser implementations. See https://www.w3.org/TR/MathML2/appendixe.html#dom-bindings.IDLBinding.

Thanks for the link, I missed that (although I studied the specs as well). AFAIK this browser implementation is a subset of the specs.

The motivation for this change, is to have a `mathml_element.style()`, which interestingly is not in that specs. It's here though: https://www.w3.org/TR/cssom-1/#the-elementcssinlinestyle-mixin
I guess I'll just combine these then?

I think I'll just add the sub-interfaces as well, so that we have a complete and more accurate representation of MathML then?

Philipp-M:
> I'm not sure if that's even considered a breaking change

Well I guess it is, when it's used explicitly: `HtmlInputElement::set_autofocus(el, true)`

So that leaves this suggestion:
> We could also leave the methods on the child interfaces there, and accept the duplicated methods, so that it's not a breaking change.

Philipp-M:
I've updated this to not be a breaking change anymore.

Well yet another time it turns out how messy the specification and its implementation in browsers is, e.g. https://www.w3.org/TR/MathML3/chapter2.html#fund.globatt specifies `style` directly, while e.g. `mathElementStyle` doesn't exist in the browsers I have checked (Firefox, Chrome). All the other methods seemingly not as well.
So I guess what is already in this PR, is at least compliant to the specs, and should cover all major currently used browsers...

crates/web-sys/webidls/enabled/Element.webidl
```diff
@@ -263,3 +263,18 @@ partial interface Element {
   [NeedsCallerType]
   undefined requestPointerLock();
 };
+
+interface mixin HTMLOrForeignElement {
+  [SameObject] readonly attribute DOMStringMap dataset;
+
+  [CEReactions, SetterThrows, Pure] attribute boolean autofocus;
+  [CEReactions, SetterThrows, Pure] attribute long tabIndex;
+  [Throws, NeedsCallerType] undefined focus(optional FocusOptions options = {});
```
daxpedda, line 272:
```suggestion
  attribute DOMString nonce; // intentionally no [CEReactions]

  [CEReactions, SetterThrows] attribute boolean autofocus;
  [CEReactions, SetterThrows] attribute long tabIndex;
  [Throws] undefined focus(optional FocusOptions options = {});
```
Lets follow the specification as close as possible here.
(also adding the new `nonce` attribute)

crates/web-sys/webidls/enabled/MathMLElement.webidl
```diff
@@ -0,0 +1,19 @@
...
+ * and permissive document license rules apply.
+ */
+
+[Exposed=Window]
+interface MathMLElement : Element { };
+MathMLElement includes GlobalEventHandlers;
+MathMLElement includes HTMLOrForeignElement;
+MathMLElement includes ElementCSSInlineStyle;
+MathMLElement includes TouchEventHandlers;
+MathMLElement includes OnErrorEventHandlerForNodes;
```
daxpedda, line 19: I mentioned this before, but we follow the specification nowadays instead of browser implementations.
```idl
  interface MathMLElement : Element
  {
             attribute DOMString         className;
             attribute DOMString         mathElementStyle;
             attribute DOMString         id;
             attribute DOMString         xref;
             attribute DOMString         href;
    readonly attribute MathMLMathElement ownerMathElement;
  };
```

I found `MathMLElement includes ElementCSSInlineStyle` in [CSSOM](https://drafts.csswg.org/cssom/#ref-for-elementcssinlinestyle%E2%91%A2), each one of those needs a reference to the spec it comes from.

daxpedda (changes requested):
Thank you for cleaning up the mixin interfaces!

crates/web-sys/webidls/enabled/Element.webidl
```diff
@@ -263,3 +263,18 @@ partial interface Element {
...
+  [Throws] undefined blur();
+};
```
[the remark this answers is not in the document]
Philipp-M: Sure we can do that for now I considered it anyway. There was already https://github.com/whatwg/html/pull/5248 and I thought it would be more accurate since it is not either HTML or SVG (and I guess that sooner or later that rename will be done, but we can adjust that when it's official).

crates/web-sys/webidls/enabled/MathMLElement.webidl
Philipp-M: I mean I can certainly add those attributes, but I'm not sure what the target group is, when it's not browser implementations (i.e. what other wasm target supports something like MathML?).

The issue I see with adding something like `mathElementStyle` is that there's no to me known wasm implementation that supports it, and it may be tempting to use, but results in exceptions then, I'm not sure if that's really what we want?

IME the spec(s) can be kinda messy and not really represent the reality (probably grown historically), that's why I think an intersection between being compliant to the specs and support of actual implementations is a good compromise.

I think [this spec](https://www.w3.org/TR/mathml-core/#dom-and-javascript) is more up-to-date and better represents the reality (funnily they mention `HTMLOrForeignElement` there and link to the `HTMLOrSVGElement`...)

> each one of those needs a reference to the spec it comes from.

I mean each one of those (mixins) has a link that specifies this indirectly in the mixin (but the `HTMLOrSVGElement` as you noticed, and I added a link to the `TouchEventHandlers`), should I do anymore here?

crates/web-sys/webidls/enabled/Element.webidl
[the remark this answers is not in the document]
daxpedda: I see.
Yeah, I would prefer not to be pre-emptive, in my experience so far things are extremely fluid in this place.

crates/web-sys/webidls/enabled/MathMLElement.webidl
daxpedda: > I mean I can certainly add those attributes, but I'm not sure what the target group is, when it's not browser implementations (i.e. what other wasm target supports something like MathML?).

No that's fine.
I was just trying to point out that all the `includes` aren't mentioned there, but I can see how this was confusing!

> > each one of those needs a reference to the spec it comes from.
> 
> I mean each one of those (mixins) has a link that specifies this indirectly in the mixin (but the `HTMLOrSVGElement` as you noticed, and I added a link to the `TouchEventHandlers`), should I do anymore here?

Basically I just want a URL next to each one of these `includes` to the spec that tells us where we got that from.
E.g. the link you added to `TouchEventHandlers` is great, but I can't find the line `MathMLElement includes TouchEventHandlers;` in it.

Philipp-M: >  but I can't find the line MathMLElement includes TouchEventHandlers; in it.

It's kind of implied in:

> The following section describes extensions to the existing [GlobalEventHandlers](https://www.w3.org/TR/html5/#globaleventhandlers) mixin, defined in [[HTML5](https://w3c.github.io/touch-events/#bib-html5)], to facilitate the event handler registration. For user agents where [expose legacy touch event APIs](https://w3c.github.io/touch-events/#dfn-expose-legacy-touch-event-apis) is false, this mixin must not be implemented.

```
partial interface mixin [GlobalEventHandlers](https://w3c.github.io/touch-events/#dom-globaleventhandlers) {
                    attribute [EventHandler](https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler) [ontouchstart](https://w3c.github.io/touch-events/#dom-globaleventhandlers-ontouchstart);
                    attribute [EventHandler](https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler) [ontouchend](https://w3c.github.io/touch-events/#dom-globaleventhandlers-ontouchend);
                    attribute [EventHandler](https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler) [ontouchmove](https://w3c.github.io/touch-events/#dom-globaleventhandlers-ontouchmove);
                    attribute [EventHandler](https://html.spec.whatwg.org/multipage/webappapis.html#eventhandler) [ontouchcancel](https://w3c.github.io/touch-events/#dom-globaleventhandlers-ontouchcancel);
};
```

So to be more close to the spec, `TouchEventHandlers` should just be the above partial mixin `GlobalEventHandlers`.

I've added some additional comments, as I don't think it hurts.

daxpedda: > It's kind of implied in:
> 
> > The following section describes extensions to the existing [GlobalEventHandlers](https://www.w3.org/TR/html5/#globaleventhandlers) mixin, defined in [[HTML5](https://w3c.github.io/touch-events/#bib-html5)], to facilitate the event handler registration. For user agents where [expose legacy touch event APIs](https://w3c.github.io/touch-events/#dfn-expose-legacy-touch-event-apis) is false, this mixin must not be implemented.
> 
> ```
> partial interface mixin GlobalEventHandlers {
>     attribute EventHandler ontouchstart;
>     attribute EventHandler ontouchend;
>     attribute EventHandler ontouchmove;
>     attribute EventHandler ontouchcancel;
> };
> ```
> 
> So to be more close to the spec, `TouchEventHandlers` should just be the above partial mixin `GlobalEventHandlers`.

I'm not following, if this should be `GlobalEventHandlers` instead, then why are we using `TouchEventHandlers` instead?
But I can't find any `MathMLElement includes GlobalEventHandlers` in any specification, so I'm not sure what you mean with implied.

I understand that this is just copied from the Mozilla source, but as I said, we don't do this anymore.
We only take from the specification.
> I've added some additional comments, as I don't think it hurts.

To clarify: this isn't optional, its the *requirement* I'm asking for here!

---

My ask here is that we don't add any WebIDL bindings outside from specifications.
Any added binding should contain a link to the specification its taken from.
If it doesn't exist in the specification, we don't add it.

Please let me know if anything is unclear!

Philipp-M: > I'm not following, if this should be GlobalEventHandlers instead, then why are we using TouchEventHandlers instead?

It's just the same as in `SvgElement` and `HtmlElement`, they have the same specs (`partial mixin GlobalEventHandlers`). We can of course also refactor these in the process as well, by using the partial mixin instead of the non-partial mixin `TouchEventHandlers`. The effect should be the same though.

> But I can't find any MathMLElement includes GlobalEventHandlers in any specification, so I'm not sure what you mean with implied.

Uh? I linked it in my previous comment (which is also in the (first) comment of the webidl code of the `MathMLElement`):

> I think [this spec](https://www.w3.org/TR/mathml-core/#dom-and-javascript) is more up-to-date and better represents the reality (funnily they mention `HTMLOrForeignElement` there and link to the `HTMLOrSVGElement`...)

> To clarify: this isn't optional, its the requirement I'm asking for here!

Right... but I don't think anything is missing anymore, i.e. all `includes` have a link to the relevant spec that defines it?

daxpedda: Ah, I see now!

> Right... but I don't think anything is missing anymore, i.e. all `includes` have a link to the relevant spec that defines it (apart of `OnErrorEventHandlerForNodes` which has a comment in `EventHandler`, not sure about that)?

I will leave a separate comment on that.

crates/web-sys/webidls/enabled/MathMLElement.webidl
```diff
@@ -0,0 +1,22 @@
...
+
+[Exposed=Window]
+interface MathMLElement : Element { };
+// https://www.w3.org/TR/mathml-core/#dom-and-javascript
+MathMLElement includes GlobalEventHandlers;
+MathMLElement includes HTMLOrSVGElement;
+// https://drafts.csswg.org/cssom/#the-elementcssinlinestyle-mixin
+MathMLElement includes ElementCSSInlineStyle;
+// https://w3c.github.io/touch-events/#extensions-to-the-globaleventhandlers-mixin
+MathMLElement includes TouchEventHandlers;
```
daxpedda, line 21:
```suggestion
// TODO: Deprecated, add to `GlobalEventHandlers` via mixin:
// https://w3c.github.io/touch-events/#extensions-to-the-globaleventhandlers-mixin
MathMLElement includes TouchEventHandlers;
```

crates/web-sys/webidls/enabled/MathMLElement.webidl
```diff
@@ -0,0 +1,22 @@
...
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * The origin of this IDL file is
+ * https://mathml-refresh.github.io/mathml-core/
+ *
+ * Copyright © 2019 W3C® (MIT, ERCIM, Keio, Beihang). W3C liability, trademark
+ * and permissive document license rules apply.
+ */
```
daxpedda, line 11:
```suggestion
// https://www.w3.org/TR/mathml-core
```

crates/web-sys/webidls/enabled/MathMLElement.webidl
```diff
@@ -0,0 +1,22 @@
...
+[Exposed=Window]
+interface MathMLElement : Element { };
+// https://www.w3.org/TR/mathml-core/#dom-and-javascript
+MathMLElement includes GlobalEventHandlers;
+MathMLElement includes HTMLOrSVGElement;
+// https://drafts.csswg.org/cssom/#the-elementcssinlinestyle-mixin
+MathMLElement includes ElementCSSInlineStyle;
+// https://w3c.github.io/touch-events/#extensions-to-the-globaleventhandlers-mixin
+MathMLElement includes TouchEventHandlers;
+MathMLElement includes OnErrorEventHandlerForNodes;
```
daxpedda, line 22: I see now whats going on, unfortunately I didn't know about all this mess, apologies!
```suggestion
// TODO: See `OnErrorEventHandlerForNodes` definition in `EventHandler.webidl`.
MathMLElement includes OnErrorEventHandlerForNodes;
```

crates/web-sys/webidls/enabled/MathMLElement.webidl
daxpedda: I guess you assumed I know whats going on with `TouchEventHandlers` and `OnErrorEventHandlerForNodes`, but unfortunately I didn't.

Sorry for the unnecessary back and forth!
Should be good to go after we resolve my suggestions below!

crates/web-sys/webidls/enabled/MathMLElement.webidl
daxpedda: Again, apologies about that.
You were just faithfully going by how the rest looked like.

crates/web-sys/webidls/enabled/MathMLElement.webidl
Philipp-M: Should be done anything else here, or do we leave it as is for now (i.e. keep that modification for a later PR)?

Philipp-M:
(force push, because CI didn't run successfully, but I don't think that failed test is related to the PR)

crates/web-sys/webidls/enabled/MathMLElement.webidl
daxpedda: No, thank you for your patience!

daxpedda (approved):
Thank you for the patience!

PR #4152, opened 2024-10-09
Add support for C-style enums with implicit discriminants

### LINKED ISSUE #2273: Enum: Enable annotating discriminant of only some variants
### Motivation
In the codebase I am working on are a couple of C-style enums, which have some variants annotated with discriminants and others not. _(see for example [`enum ColorPrimaries`](
https://github.com/xiph/rav1e/blob/6a4e7eb3445896023a7fd65df5feaefcfd53de10/src/api/color.rs#L41-L73))_

I'd like to export these enums to js, to be able to use them as input parameters for some functions.

But, currently following code doesn't work, because one`must either annotate discriminant of all variants or none`.

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
#[derive(Debug)]
pub enum Boat {
    Red,
    Blue,
    Green = 42,
    Yellow,
}

#[wasm_bindgen]
pub fn run_it(b: Boat) -> String {
    format!("{:?}", b)
}
```

### Proposed Solution
I'd like `wasm-bindgen` to generate following js-glue-code:
```ts
export const Boat = Object.freeze({
    Red: 0,     "0":  "Red",
    Blue: 1,    "1":  "Blue",
    Green: 42,  "42": "Green",
    Yellow: 43, "43": "Yellow",
});
```
... and following type annotations:
```ts
export enum Boat {
  Red,
  Green,
  Blue,
  Yellow,
}
```

### Alternatives

I could also create a copy of the needed enums, remove all discriminants and then match the copied and actual enum in the function. But this sounds not like a clean and maintainable solution...

### Additional Context

I am happy to implement these changes, but would need a bit of mentoring :smile: 

<details>
<summary>Edit</summary>

Initially I proposed to generate following js-Object:
```js
export const Boat = Object.freeze({
    Red: 0,     "0":  "Red",
    Blue: 1,    "1":  "Blue",
    Green: 42,  "42": "Green",
    Yellow: 2,  "2":  "Yellow",
});
```

But playing around with rust I saw that following code:
```rust
pub enum Boat {
    Red,
    Blue,
    Green = 42,
    Yellow,
}

fn main() {
    let a = Boat::Red as u32;
    let b = Boat::Blue as u32;
    let c = Boat::Green as u32;
    let d = Boat::Yellow as u32;

    for i in [a, b, c, d].iter() {
        eprintln!("{:?}", i);
    }
}
```
... actually outputs:
```
0
1
42
43
```
So I guess it is better to stick with that also for the js code.

</details>

RunDevelopment:
fixes #2273

Changes:
1. Allow implicit discriminant by using the same algorithm as rustc. This behavior is [explicitly documented](https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants) and guaranteed by the compiler.
2. Simpler algorithm for finding holes.

### 1. Implicit discriminants

I won't go into the algorithm for determining implicit discriminants, just [read the docs](https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants). I just want to talk about a few details.

Firstly, while the behavior of implicit discriminants is consistent with rustc, everything would still work correctly even if it wasn't. Since we generate both the Rust and JS/TS enum with all discriminants explicitly specified, there will never a case where Rust and JS/TS disagree on what number a certain variant is. The worst that can happen is that we assign implicit discriminants numbers that users don't expect. (But again, we do it the same way rustc does, so it should be what users expect.)

Secondly, [TS enums](https://www.typescriptlang.org/docs/handbook/enums.html) use the same algorithm for assigning implicit discriminants. The only difference is that TS allows duplicate discriminants while Rust doesn't. So JS/TS devs copying over TS enums with duplicate discriminants will result in a compiler error with and without `#[wasm_bindgen]`.

### 2. Holes

Since I keep track of all discriminant values for better error messages, I implemented a simpler algorithm for finding holes. It's guaranteed to find the hole with the smallest numeric value and doesn't have issues with arithmetic overflow.

Note that the old algorithm didn't always find the hole with smallest numeric value, so Wasm bindgen might use different holes now. E.g. `enum E { A = 1, B = 3 }` previously used the hole 2 but now uses 0. I don't think this is an issue, but I wanted to mention it.


crates/macro-support/src/parser.rs
```diff
@@ -1454,21 +1470,9 @@ impl<'a> MacroParse<(&'a mut TokenStream, BindgenAttrs)> for syn::ItemEnum {
...
-                    None
-                }
-            })
-            .unwrap_or(*values.last().unwrap() + 1);
-        for value in values {
-            assert!(hole != value);
-        }
+        let hole = (0..u32::MAX)
+            .find(|v| !discriminate_map.contains_key(v))
+            .unwrap_or(u32::MAX);
```
daxpedda, line 1475: If an enum has `u32::MAX` variants, this will output `u32::MAX`, which would be wrong.
Considering that this should be practically impossible, I would argue using `unwrap()` or `expect()` here.

RunDevelopment: Certainly an unlikely edge case, but you're right.

Related question: Should we hard-limit the number of enum variants?

daxpedda: > Related question: Should we hard-limit the number of enum variants?

I don't think its necessary, it should actually be quite impossible to make such a big enum to begin with.
In any case, panicking should keep us on the sane side of this thought-experiment.

daxpedda (approved):
Thanks!

The changelog entry needs to be moved to a new "Unreleased" section.

RunDevelopment:
> The changelog entry needs to be moved to a new "Unreleased" section.

Done :)

daxpedda (approved):

PR #4153, opened 2024-10-09
Properly indent the doc comments of enum variants

RunDevelopment:
Very small PR. The doc comments of enum variants weren't indented properly in `.d.ts` files. This PR fixes that.

The enum objects in `.js` aren't formatted sensibly with comments anyway, so I just left it as is.


daxpedda (approved):
Thank you!

PR #4155, opened 2024-10-09
Replace all `Self` identifiers in method argument and return types

### LINKED ISSUE #3105: Returning Option<Self> in #[wasm_bindgen] impl block gives error

RunDevelopment:
fixes #3105
closes #3631

This PR is based on #3631 by @snOm3ad. I looked at it to see what needed to be changed and that I stole tests from there. I tagged @snOm3ad as the co-author in one commit, but (*speaking to the maintainer*) please also tag them in the merge commit of this PR in case GH doesn't.

The main approach of this PR is different from #3631. I used [`syn`'s mut visitor](https://docs.rs/syn/latest/syn/visit_mut/index.html) to replace all `Self` identifiers with the identifier for the type. This makes the code very simple and guarantees that no occurrences of `Self` are left.


daxpedda (commented):
I would appreciate if you could leave some code comments here, I always have to spend non-trivial time digging into this when it comes up.

crates/macro-support/src/parser.rs
```diff
@@ -911,24 +912,24 @@ fn function_from_decl(
...
+                    if i == "Self" {
+                        *i = self.0.clone();
+                    }
+                }
+            }
+
+            let mut replace = SelfReplace(self_ty.clone());
+            replace.visit_type_mut(&mut t);
+        }
+        t
```
daxpedda, line 932: Apologies, I wasn't being specific enough.
This is the part I think that needs to be explained.

Whats happening here?
How is this used to replace all `Self` regardless of depth? (is that the case?)
Why is `VisitMut` necessary?

This might need more comments in the surrounding code as well.
You didn't write all the surrounding code, I'm sorry this is being pushed onto you!

RunDevelopment: Ah, sorry about that. I did **a lot** of stuff with ASTs before, so using visitors is almost second nature to me at this point.

I'll add some more comments to answer your questions.

> You didn't write all the surrounding code, I'm sorry this is being pushed onto you!

No worries. I'm currently working on another PR that changes some code in this area, so I can add a few comments to explain the big picture and how all the pieces for together.

RunDevelopment: Okay, I added a few comments **for now**.

I actually wrote a few more, but discovered some minor bugs while doing that. So I'll fix those bugs and explain the fixed code in a follow-up PR.

daxpedda (approved):
Thank you!
(needs a rebase again)

Issue #3864, opened 2024-03-01
enum in return position have wrong type


sigmaSd:
### Describe the Bug

rust enums in return position are typed as object, but at runtime they're just a number

### Steps to Reproduce

```rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub enum A {
    B,
    C,
}

#[wasm_bindgen]
pub fn b() -> A {
    return A::B;
}
```

```ts
import { b } from "./dist/b.js";

console.log(typeof b()); //number
console.log(b() === 0); 
```
```
error: TS2367 [ERROR]: This comparison appears to be unintentional because the types 'Readonly<{ B: 0; "0": "B"; C: 1; "1": "C"; }>' and 'number' have no overlap.
console.log(b() === 0);
            ~~~~~~~~~
```

wasm-bindgen 0.2.91

Liamolucko:
Weird. I think this is the bit of the generated code causing the issue:

```js
/**
* @returns {A}
*/
export function b() {
    const ret = wasm.b();
    return ret;
}

/**
*/
export const A = Object.freeze({ B:0,"0":"B",C:1,"1":"C", });
```

It seems like TypeScript is interpreting the `@returns {A}` as `@returns {typeof A}` for some reason.

Is this happening in Deno? I can reproduce it with Deno, but not with plain TSC, I think because TSC automatically checks for `b.d.ts` and Deno doesn't. If it is, you can get Deno to use the proper type annotations instead of the JSDoc ones with `@deno-types`:

```ts
// @deno-types="./dist/b.d.ts"
import { b } from "./dist/b.js";

console.log(typeof b()); //number
console.log(b() === 0); 
```

There are 2 approaches to solving this properly:

1. Add `@enum {number}` to `export const A`. This isn't as good as it being a real TypeScript enum, it basically just declares `type A = number` but it fixes the error.
2. Add a `/// <reference types="..." />` to the JS file when using `--target deno` so that Deno can pick up the proper type declarations in the first place.

Probably we should do both, since the JSDoc declarations should work if necessary (e.g. `--no-typescript` is passed) but then the TypeScript declarations should be used if available.

sigmaSd:
Yes it's in Deno, I didn't think it's related thank fo investigating, is Deno behavior correct here or should I open a bug report ?

Liamolucko:
[...]

PR #4157, opened 2024-10-09
Add JSDoc annotations to ensure enums type check even without `.d.ts` file

RunDevelopment:
This is a partial fix for #3864.

I added additional JSDoc annotations to ensure that enums type check even without explicit type definition (`.d.ts` file). 

C-style enums get an `@enum {1|2|3|4}` annotation which (1) marks the frozen enum object as an enum and (2) ensures that type annotations properly resolve. A comment in #3864 suggested using `@enum {number}`, but I defined the enum as the union of all possible numeric values instead to make the type more specific.

String enums get `@typedef {"a" | "b" | "c"} StringEnum` which is exactly the same as TS `type StringEnum = "a" | "b" | "c"`. This means that string enums have exact type definitions with and without a `.d.ts` file. I also added `@type {StringEnum[]}` to arrays of string enum variants to ensure that the internal uses of those arrays don't cause type errors.

With those 2 changes, all of `enum.js` type checks without its `.d.ts` file.

One remaining problem is that I don't know how to test for the lack of type errors. While making this PR, I just added a `// @ts-check` at the top of `enum.js` so I could see type errors in my IDE. Any ideas how I could add tests for this?


daxpedda:
@RunDevelopment if you rebase this PR, I'm happy to give it a review.

> One remaining problem is that I don't know how to test for the lack of type errors. While making this PR, I just added a `// @ts-check` at the top of `enum.js` so I could see type errors in my IDE. Any ideas how I could add tests for this?

Unfortunately not.
I have almost no experience with JS/TS linting, but ideally all our output testing should be covered by them.

RunDevelopment:
Rebased. Thank you for reviewing!

crates/cli/tests/reference/enums.js
```diff
@@ -78,9 +79,16 @@ Yellow:1,"1":"Yellow",
...
  */
 Red:2,"2":"Red", });
-
+/**
+ * @enum {0 | 1 | 42 | 43}
+ */
 export const ImplicitDiscriminant = Object.freeze({ A:0,"0":"A",B:1,"1":"B",C:42,"42":"C",D:43,"43":"D", });
 
+/**
+ * The name of a color.
```
daxpedda, line 88: Considering this isn't exported, I believe doc comments shouldn't be exposed as well.
WDYT?

crates/cli/tests/reference/enums.js
```diff
@@ -78,9 +79,16 @@ Yellow:1,"1":"Yellow",
...
+ * @typedef {"green" | "yellow" | "red"} ColorName
```
daxpedda, line 89: Does this get registered even though the user only gets to use and see `get_name()`?

crates/cli/tests/reference/enums.js
```diff
@@ -78,9 +79,16 @@ Yellow:1,"1":"Yellow",
...
+ * @type {ColorName[]}
```
daxpedda, line 90: > I also added `@type {StringEnum[]}` to arrays of string enum variants to ensure that the internal uses of those arrays don't cause type errors.

This belong in #4189 then.

crates/cli/tests/reference/enums.js
RunDevelopment: I think that's reasonable. I'll remove them.

daxpedda (changes requested):
LGTM!
Missing a changelog entry.

RunDevelopment:
Done, although I'm not sure whether the explanation I picked for this change is good...

daxpedda (approved):
Would appreciate if you could do the rebase!

RunDevelopment:
Sure thing!

PR #4178, opened 2024-10-11
Fixed methods with `self: &Self` consuming the object

### LINKED ISSUE #3958: JS Pointer gets nulled if rust method defined as self: &Self instead of &self
### Describe the Bug

When defining methods on a rust struct that you want to be exported with wasm-bindgen, if you define the arguments as `self: &Self` then when called from JS, the JS pointer gets nulled out after calling that method instead of the expected behaviour of the pointer staying valid. If you intead define the arguments as `&self` then it works as expected.

### Steps to Reproduce

Here is a minimal example of the code not working.

The rust code:
```rust
#[wasm_bindgen]
pub struct Test {

}

#[wasm_bindgen]
impl Test {
    #[wasm_bindgen(constructor)]
    pub fn new() -> Test {
        Test {}
    }
    pub fn method(self: &Self) -> String {
        "Test".to_string()
    }
    pub fn method_2(self: &Self) -> String {
        "Test_2".to_string()
    }
}
```
JS Code:
```js
const test = new wasm.Test()
console.log(test)
test.method()
console.log(test)
test.method_2()
console.log(test)
```
When this is run, the JS code errors at `test.method_2()` "Error importing `index.js`: Error: null pointer passed to". We can see from the log that the internal wasm pointer goes from a valid number to 0 after the `test.method()` call.

If we change the rust code to:
```rust
#[wasm_bindgen]
pub struct Test {

}

#[wasm_bindgen]
impl Test {
    #[wasm_bindgen(constructor)]
    pub fn new() -> Test {
        Test {}
    }
    pub fn method(&self -> String {
        "Test".to_string()
    }
    pub fn method_2(&self -> String {
        "Test_2".to_string()
    }
}
```
(Just replacing `self: &Self` with `&self`)
Then both methods pass and no pointers are nulled out.

### Expected Behavior

I would expect both pieces of rust code to behave the same, and the first definition shouldn't null pointers after calling each method.

### Actual Behavior

The first method nulls the pointer after each method called on the struct.

### Additional Context

This was run with wasm-bindgen version of "0.2.84" on Linux Ubuntu. I started with the [hello wasm-pack template](https://rustwasm.github.io/docs/wasm-pack/)

RunDevelopment:
fixes #3958

Rust allows type annotations for the `self` parameter, so it can have many forms. E.g. `self`, `&self`, `&mut self`, `self: &Self`, `self: Self`, `self: Box<Self>`, etc. Wasm bindgen needs to know whether a value is consumed (pass by value) or referenced to generate the correct JS shim.

Previously, the parser only looked at whether the self parameter had an `&` before `self`. This works for `self`, `&mut self`, and `&self`, but not for `self: &Self`. So it previously interpreted `self: &Self` (a reference) as just `self` (pass by value), because it ignored any and all type annotations, which caused #3958.

This PR fixes the issue by looking at the type of `self` instead. Helpfully, syn always gives us a type even for `&self`. [See their docs](https://docs.rs/syn/latest/syn/struct.Receiver.html) for more details. So instead of checking whether `self` has an `&`, it now checks whether the type of `self` is a reference type. This is the same approach wasm bindgen uses for all other arguments to figure out whether they are passed by referece or by value.


daxpedda (changes requested):
LGTM!

daxpedda (approved):

crates/macro-support/src/parser.rs
```diff
@@ -886,6 +886,26 @@ pub(crate) fn is_js_keyword(keyword: &str, skip: Option<&[&str]>) -> bool {
...
 }
 
+/// Returns whether `self` is passed by reference or by value.
+fn get_self_method(r: syn::Receiver) -> ast::MethodSelf {
+    // The tricky part here is that `r` can have many forms. E.g. `self`,
+    // `&self`, `&mut self`, `self: Self`, `self: &Self`, `self: &mut Self`,
+    // `self: Box<Self>`, `self: Rc<Self>`, etc.
+    // Luckily, syn always populates the `ty` field with the type of `self`, so
+    // e.g. `&self` gets the type `&Self`. So we only have check whether the
+    // type is a reference or not.
```
daxpedda, line 896: I thought I left a comment about this before:
Excellent! Thank you!

PR #4173, opened 2024-10-10
Add `skip_typescript` to add string enums in `web-sys`

RunDevelopment:
Temporary solution for #4163.

I changed the webidl generator code and added a test case to see whether the fix works.

However, running the command from the `web-sys` README fails on my system, so I cannot re-generate the rs files.

```
PS C:\Users\micha\Git\wasm-bindgen> cargo run --release --package wasm-bindgen-webidl -- webidls src/features ./Cargo.toml
    Finished `release` profile [optimized] target(s) in 0.30s
     Running `target\release\wasm-bindgen-webidl.exe webidls src/features ./Cargo.toml`
Error: reading webidls directory

Caused by:
    Das System kann den angegebenen Pfad nicht finden. (os error 3)
error: process didn't exit successfully: `target\release\wasm-bindgen-webidl.exe webidls src/features ./Cargo.toml` (exit code: 1)
```

Issue #4163, opened 2024-10-10
Unnecessary Typescript definitions are created for web_sys string enums


davids-work:
### Describe the Bug
After updating to wasm_bindgen 0.2.94 / web_sys 0.3.71, I noticed that the generated Typescript definition file created by wasm-pack contained definitions for a bunch of string enums stemming from web_sys, e.g.
```typescript
/**
 *The `MediaSourceReadyState` enum.
 *
 **This API requires the following crate features to be activated: `MediaSourceReadyState`*
 */
export type MediaSourceReadyState = "closed" | "open" | "ended";
```
These were not there with wasm_bindgen 0.2.93.

### Steps to Reproduce
Cargo.toml:
```toml
[package]
name = "wasm-bindgen-unnecessary-bindings"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib"]

[dependencies]
wasm-bindgen = { version = "0.2.94" }

[dependencies.web-sys]
version = "0.3.71"
features = [
    "console",
    "MediaSourceReadyState",
]
```

src/lib.rs
```rust
use wasm_bindgen::prelude::wasm_bindgen;
use web_sys::{console::info, js_sys::Array};

// Note how MediaSourceReadyState is not used anywhere!

#[wasm_bindgen]
pub fn foo() {
    // Just to create some reference to web_sys so that it's actually pulled in
    info(&Array::new());
}
```
Build: `wasm-pack build --release --target web`

### Expected Behavior

Definitions are only generated for types that are part of my crate's interface

### Actual Behavior
Definitions are generated for the `web_sys::MediaSourceReadyState` enum despite it not being part of my library at all.

More string enum definitions are generated if other web_sys features are enabled

daxpedda:
Stemming from #4147.

I didn't check yet, but this might justify a yank.
Cc @RunDevelopment.

RunDevelopment:
I reproduced the issue and *technically* it is working as intended.

(Sorry in advance for the long comment. I tried to make it as short as possible, but it still ended up like this.)

## How `web-sys` defines string enums

`web-sys` main purpose is define *Rust bindings* for JS web APIs. Since JS APIs are *imported* into Rust, nothing is *exported* on the JS/TS side, so `.d.ts` don't include these types. Example:

```rs
// web-sys code
#[wasm_bindgen]
extern "C" {
[...]

PR #4179, opened 2024-10-11
Add test for code gen with `web-sys`

RunDevelopment:
This re-adds the test for web-sys from #4173 to make sure #4163 doesn't repeat.


daxpedda (approved):
More tests :heart:.

PR #4180, opened 2024-10-11
Improved string enum typing try 2

RunDevelopment:
Alright, let's try again. This PR reverts most of #4174 and restores most of #4147. I just made 2 changes:

1. String enum types are only generated when used.
2. String enum types are never exported.

This ensures that we don't repeat #4163 and puts us in a good position to make a fix for #2153, because we just have to decide when to export the generated type (but this will be its own PR).

Code-wise, not much is different from #4147. The only new part is how usage is detected. I hocked into the function that generates function signatures to keep track of which string enums are referenced. Since we always generate the type definition for exported functions *before* the type definitions for string enums, this works out beautifully and ensures that we know exactly which string enums are used.

---

If possible, please merge #4179 before this PR, so I can update this PR and have tests to ensure that #4163 isn't repeated.


crates/cli-support/src/js/binding.rs
```diff
@@ -76,6 +78,14 @@ pub struct JsFunction {
     pub log_error: bool,
 }
 
+/// A references to an (likely) exported symbol used in TS type expression.
+///
+/// Right now, only string enum require this type of anaylsis.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum TSReference {
```
daxpedda, line 85:
```suggestion
pub enum TsReference {
```
See the [API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

daxpedda (changes requested):
Nice work!
Missing a changelog entry.

Gonna wait for those tests as well like you said!

daxpedda:
Also reverts #4175.

CHANGELOG.md
```diff
@@ -3,9 +3,14 @@
 
 ## Unreleased
 
+### Changed
+
+* String enums now have private TypeScript types only if used.
```
```suggestion
* String enums now generate private TypeScript types but only if used.
```

crates/cli/tests/reference/enums.js
```diff
@@ -83,6 +83,10 @@ export const ImplicitDiscriminant = Object.freeze({ A:0,"0":"A",B:1,"1":"B",C:42
 
 const __wbindgen_enum_ColorName = ["green", "yellow", "red"];
 
+const __wbindgen_enum_FooBar = ["foo", "bar"];
```
daxpedda, line 86: This isn't used so it should not be generated as well.
I don't believe this is an issue this PR caused though.

daxpedda (approved):

crates/cli/tests/reference/enums.js
RunDevelopment: Yes, this is a side effect of #4147 (point 3), though I also only noticed this yesterday. I'll make a PR fixing this tomorrow. I just wanted for this PR to be in, because I'll need to implement a similar usage analysis. Nothing complex, I just wanted to see would be acceptable in the first place.

PR #4186, opened 2024-10-11
Use `isLikeNone` to detect null/undefined for string enums

RunDevelopment:
This is a very minor thing I noticed. All conversions except the string enum one use `isLikeNone` to detect none-like JS values (`undefined` and `null`). So I change the JS string enum to WASM string enum conversion to use `isLikeNone` as well. All conversions new use `isLikeNone` to detect none-like values.


daxpedda (approved):
Good catch!

PR #4187, opened 2024-10-11
Remove JSDoc annotations in TS files

RunDevelopment:
This PR can be seen as a continuation of #4135.

I removed the JSDoc annotations in `.d.ts` files.

Before:
```ts
/**
 * Some docs.
 * @param {number} a
 * @param {number} b
 * @returns {number}
 */
export function add_u32(a: number, b: number): number;
```
After:
```ts
/**
 * Some docs.
 */
export function add_u32(a: number, b: number): number;
```

I did this because JSDoc annotations (so `@param` and `@returns`) are just unnecessary noise. They just repeat the function signature and contain *zero* information.

Note that this **only** affects the generates TS files. The JSDoc annotations are wanted in JS files and are remain unchanged.

---

With this PR, I would actually consider the form of generated `.d.ts` files to be good. At least for my use case, they are now exactly what I want.


daxpedda (commented):
Unfortunately I'm not really familiar with the TS ecosystem.
Is there some guidelines where something like this is outlines?

If you just point me into the right direction I'm happy to go ahead with this.

RunDevelopment:
Here's an [SO post explaining doc comments in TypeScript](https://stackoverflow.com/questions/23072286/where-is-the-syntax-for-typescript-comments-documented). The top answer explains it quite well.

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## Unreleased
+
+### Changed
+
+* Remove unnecessary JSDoc annotations from generated `.d.ts` files
```
```suggestion
* Remove unnecessary JSDoc type annotations from generated `.d.ts` files
```
Is that the right wording, or is it more confusing?

daxpedda (approved):
LGTM!

CHANGELOG.md
RunDevelopment: No you're right. There could be user-defined JSDoc annotations and those aren't removed.

Issue #3330, opened 2023-02-24
Snippets generated even when not needed


daxpedda:
The PR #3069 introduced linked modules, `link_to!(module = "/test.js")`. Currently this is meant to use snippets only if the `--split-linked-modules` is enabled, but not otherwise.

Currently, snippets are always generated, even in the no-modules target, they can be safely deleted and ignored, but they shouldn't be generated in the first place.

This can be easily tried by using `link_to!(module = "/test.js")` and trying out the following commands.
- `wasm-bindgen --target no-modules`
- `wasm-bindgen --target web`
- `wasm-bindgen --target no-modules --split-linked-modules`

Happy to give a more comprehensive, easy to use example if desired.

Cc @lukaslihotzki.

lukaslihotzki:
I understand your issue.

There isn't a clean, small fix for it, because we don't know the target in the function where we could remove the files. Maybe we could disable the write-to-disk of all snippets on no-modules without `--split-linked-modules`, because snippets cannot be imported modules or split linked modules (as wasm-bindgen would error earlier), so they are embedded linked modules, so their write-to-disk can be skipped. I'll look if that can be a small fix. I wouldn't like to introduce new complexity to fix this.

RReverser:
link_to can actually help me remove couple of manual hacks from wasm-bindgen-rayon, but only if this behaviour is preserved. Now that it exists, I'm doing something like:

```rs
fn _ensure_worker_emitted() {
    // Just ensure that the worker is emitted into the output folder, but don't actually use the URL.
    wasm_bindgen::link_to!(module = "/src/workerHelpers.worker.js");
}
```

This is because I need the Worker to be emitted alongside the snippet JS so that I can use static construct `new Worker(new URL('/src/workerHelpers.worker.js', import.meta.url), { type: 'module' })` that is recognised by bundlers only in this shape.

If snippet wasn't emitted and I could only get a URL value from wasm-bindgen without the statically emitted `new Worker(...)` wrapper, that would break serveral bundlers.

RReverser:
Another alternative would be to extend `link_to!` to allow custom wrappers around it so that devs could do `link_to!(module = "...", wrapper = "new Worker({}, {{ type: 'module' }})")` and such, but I imagine that would be a bigger change.

daxpedda:
@RReverser I plan to address this soon and would like to understand your use case a bit more.

[...]

PR #4188, opened 2024-10-12
Add `null` to argument types of optional parameters

RunDevelopment:
All functions with `Option<T>` arguments take both `undefined` and `null` and generally treat them as equivalent ([see `isLikeNone`](https://github.com/rustwasm/wasm-bindgen/blob/3a8da7cb8842d4cb9918871179b6b7d3c77df5a0/crates/cli-support/src/js/mod.rs#L2181)). Despite both `undefined` and `null` being accepted, the generated type definitions only allowed `undefined`. Example:

```rs
#[wasm_bindgen]
pub fn echo(a: Option<String>) -> Option<String> {
    a
}
```

```ts
export function echo(a?: string): string | undefined
```

This is unnecessarily restrictive, as `echo(null)` would also work (see `isLikeNone`) and return `undefined` just like `echo(undefined)`. This also results in worse ergonomics as many DOM APIs return `value | null` (e.g. [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)), which currently requires dev to unnecessarily map `null` to `undefined` (e.g. `echo(element.textContent ?? undefined)`).

To bring the generated type definitions in line with the actual runtime behavior of the JS glue code, I changed the type of optional parameters from `T | undefined` to `T | undefined | null`. E.g. the function signature of `echo` is the following with this PR:
```ts
export function echo(a?: string | null): string | undefined
```

This accurately represent the runtime behavior of the function and results in better ergonomics.

Note: This PR only affects function arguments. The return type is still `T | undefined`, as it should be.

### Tests

To test that this (and other features I have planned) are implemented correctly, I added a single large test file: `echo.rs`. This test file just contains echo/identity functions for various types to check the glue code of various input/output types.


crates/cli-support/src/js/binding.rs
```diff
@@ -1417,7 +1419,23 @@ impl Invocation {
     }
 }
 
-fn adapter2ts(ty: &AdapterType, dst: &mut String) {
+#[derive(Debug, Clone, Copy)]
+enum NoneType {
+    Undefined,
+    Null,
```
daxpedda, line 1425: This is never built, right?

Instead, why don't we do something like this:
```rust
pub enum TypePosition {
    Argument,
    Return,
}
```

daxpedda (changes requested):
LGTM!

Missing a changelog entry.

crates/cli-support/src/js/binding.rs
RunDevelopment: I used to use it, but not anymore after some code changes before I made the PR.

But your suggestion is better. I'm planing to do more changes that make inputs more accepting in the future, so this is perfect.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * String enums now generate private TypeScript types but only if used.
   [#4174](https://github.com/rustwasm/wasm-bindgen/pull/4174)
 
+* Optional parameters are now typed as `T | undefined | null` to reflect the JS behavior.
```
daxpedda, line 11:
```suggestion
* Optional parameters are now typed as `T | undefined | null` to reflect the actual JS behavior.
```

crates/cli/tests/reference/echo.d.ts
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * @param {(Foo)[]} a
+ * @returns {(Foo)[]}
+ */
+export function echo_vec_struct(a: (Foo)[]): (Foo)[];
+/**
+ * @param {number | undefined | null} [a]
+ * @returns {number | undefined}
+ */
+export function echo_option_u8(a?: number): number | undefined;
```
daxpedda, line 97: Something happened in the last rebase and the `| null` is missing.

crates/cli/tests/reference/enums.js
```diff
@@ -36,7 +36,7 @@ function isLikeNone(x) {
     return x === undefined || x === null;
 }
 /**
- * @param {Color | undefined} [color]
+ * @param {Color | undefined | null} [color]
```
daxpedda, line 39:
```suggestion
 * @param {Color | null} [color]
```
I think it should not include `undefined` because that's already covered by the `[]` around `color` which signify an optional parameter, right?

crates/cli/tests/reference/echo.d.ts
RunDevelopment: Oh god, thanks for catching this!

crates/cli/tests/reference/enums.js
RunDevelopment: Depends on what you mean by "should".

You are correct in that `| undefined` is redundant because of the `[]` and can be removed. So for good form, it should be removed. However, it's also not wrong, since it doesn't do anything. So if you meant "it should be removed for correctness", that would be wrong.

I can remove the `| undefined` if you want, but I'll need to add a code branch to do it (maybe 5 LOC). I don't mind either way, so please decide.

daxpedda: Why would we leave it if its redundant?
If there isn't a reason, I would go ahead with removing it.

RunDevelopment: After thinking about it some more, I think we should keep `| undefined` in JSDoc.

Someone not as familiar with JSDoc might not immediately realize that `[]` means that the parameter accepts `undefined`. Consider full function `option_enum_echo`:

```js
/**
 * @param {Color | null} [color]
 * @returns {Color | undefined}
 */
export function option_enum_echo(color) {
    const ret = wasm.option_enum_echo(isLikeNone(color) ? 3 : color);
    return ret === 3 ? undefined : ret;
}
```

Nothing in the function body makes it immediately obvious that `color` can be `undefined`. So I think it would be better to explicitly spell out `| undefined` instead of relying on `[]` to imply it.

I know that this is not a good source, but [I asked chatgpt about this](https://chatgpt.com/share/670b93ed-5f00-800f-aeee-cfabdd5c1c44), and it also recommended including the `| undefined` so it's clear to people new to JS/JSDoc.

---

I also think that the same argument does not apply to TS type definitions. Every TS dev should know what `arg?: T` means, because it's a commonly-used language feature, so I don't worry about that causing confusion.

daxpedda: Unfortunately I'm not qualified to make a decision like this.

The way I usually approach this kind of stuff is to follow the specification.
If a specification is not present, only then I would follow community consensus.
Only in a case where the specification and the community consensus conflict, I would dare to make a judgement call.

I don't know what the community consensus is, but the specification is clear to me.
While I would not rely on ChatGPTs output, I'm not seeing a recommendation here.
If what I'm reading there is correct, I would actually go with removing `| undefined`.

So unless you feel quite strongly about this, seeing that you are more qualified then myself here, lets go ahead with removing `| undefined`.

RunDevelopment: Alright, then let's remove it. I don't have any strong feelings about this, and it's a minor thing anyway since it's correct either way.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 * Remove unnecessary JSDoc type annotations from generated `.d.ts` files
   [#4187](https://github.com/rustwasm/wasm-bindgen/pull/4187)
 
+* Optional parameters are now typed as `T | undefined | null` to reflect the actual JS behavior.
```
daxpedda, line 14:
```suggestion
* Optional parameters are now typed as `T | null` to reflect the actual JS behavior.
```

crates/cli-support/src/js/binding.rs
```diff
@@ -1450,8 +1470,11 @@ fn adapter2ts(ty: &AdapterType, dst: &mut String, refs: Option<&mut HashSet<TsRe
         AdapterType::Bool => dst.push_str("boolean"),
         AdapterType::Vector(kind) => dst.push_str(&kind.js_ty()),
         AdapterType::Option(ty) => {
-            adapter2ts(ty, dst, refs);
-            dst.push_str(" | undefined");
+            adapter2ts(ty, position, dst, refs);
+            dst.push_str(match position {
+                TypePosition::Argument => " | undefined | null",
```
daxpedda, line 1475:
```suggestion
                TypePosition::Argument => unreachable!(),
```
We always handle this case manually, right?

daxpedda (changes requested):
Small nit left, otherwise this LGTM.

CHANGELOG.md
RunDevelopment: This would be incorrect. We may *write* an optional parameter as `arg?: T | null` since `undefined` is implied by `?`, but the type of `arg` *is* `T | undefined | null`.

crates/cli-support/src/js/binding.rs
RunDevelopment: We don't for variadic arguments (`bindings.rs` line 419). Though I'm not sure how relevant that is, since varargs probably don't allow `Option<T>`.

But I also don't like the idea of `adapter2ts` knowing with what types it is called. This feels like a bit of global reasoning to me.

RunDevelopment:
I just realized that return and argument types not being the same is a problem for generated getter-setter pairs.

RunDevelopment:
Fixed it in #4202. After #4202 is in, I need to update this PR again, so I'll mark it as a draft for the time being.

CHANGELOG.md
daxpedda: Ah, right!

crates/cli-support/src/js/binding.rs
daxpedda: Hm, aren't we handling all `Option<T>` manually before they ever arrive in `adapter2ts()`?
I'm just concerned because this code is actually incorrect, this function can't handle this case correctly.

Ideally we move `Option<T>` handling entirely into this function, but this could be done in a separate PR.

So unless I'm missing something, I think the whole `AdapterType::Option(ty) => { ... }` statement should yield `unreachable!()` imo.

daxpedda: > I'm just concerned because this code is actually incorrect, this function can't handle this case correctly.

Lets say varargs do support this, would the output be correct?

RunDevelopment: Yes, the output would be correct for [documenting varargs in JSDoc](https://stackoverflow.com/questions/4729516/correct-way-to-document-open-ended-argument-functions-in-jsdoc) AFAICT. E.g. `Option<u32>` would produce:

```
@param {...number | null | undefined} name
```

RunDevelopment: After thinking about it some more, let's go with your suggestion. Varargs are pretty under-tested and under-documented right now, so I have no idea whether they even allow `Option<T>`. 

I will need to re-add this in the future (because I plan to add support for arrays soon), but I can re-add it when there are tests for it then.

RunDevelopment:
Alright, getters and setters are now correct, so this PR is pretty much finished.

CI doesn't pass, because CI is currently broken. See #4210, #4213, and #4216.

RunDevelopment:
@daxpedda You know when we talked about `unreachable!()`? We both made the mistake of thinking that `Option`s were always handled in a special branch. We should have [read the code](https://github.com/rustwasm/wasm-bindgen/blob/b45cfb6ced0f591acdeb9578edc993cc13a09b57/crates/cli-support/src/js/binding.rs#L323) more carefully :) See that `if omittable`? Yeah, it's super reachable, but our tests didn't cover it. I'll add a test for it.

RunDevelopment:
Okay, there is one more thing we have to talk about. I noticed that one of our TS tests failed. This test defined Rust struct like this:

```rs
#[wasm_bindgen]
pub struct Fields {
    pub hallo: Option<bool>,
    pub spaceboy: bool,
}
```

And checked whether this type checks:

```ts
import * as wbg from '../pkg/typescript_tests';

const fields: wbg.Fields = { spaceboy: true, free: () => { } };
```

Note that the assigned object has no `hallo` field.

This previously worked, because the generated type of the `Fields` class *was* this:

```ts
class Fields {
    hallo?: boolean;
    spaceboy: boolean;
    free(): void;
}
```

Note that `hallo` *used to be* an optional field, meaning that it could be omitted.

This is no longer the case. Since fields of type `Option<T>` now have properly typed getters and setters, the type definition of the fields class now looks like this:

```ts
class Fields {
    get hallo(): boolean | undefined;
}
```

RunDevelopment:
Okay, there is one more thing I need to put out there. I noticed that one of our TS tests failed. This test defined Rust struct like this:

```rs
#[wasm_bindgen]
pub struct Fields {
    pub hallo: Option<bool>,
    pub spaceboy: bool,
}
```

And checked whether this type checks:

```ts
import * as wbg from '../pkg/typescript_tests';

const fields: wbg.Fields = { spaceboy: true, free: () => { } };
```

Note that the assigned object has no `hallo` field.

This previously worked, because the generated type of the `Fields` class *was* this:

```ts
class Fields {
    hallo?: boolean;
    spaceboy: boolean;
    free(): void;
}
```

Note that `hallo` *used to be* an optional field, meaning that it could be omitted. This is why the above TS code compiled.

This is no longer the case. Since fields of type `Option<T>` now have properly typed getters and setters, the type definition of the fields class now looks like this:

```ts
class Fields {
    get hallo(): boolean | undefined;
    set hallo(value: boolean | null | undefined);
    spaceboy: boolean;
    free(): void;
}
```

The important bit here is that getters and setters can never be omitted. This means that the above TS code no longer compiles, since the `hallo` field is missing. So I adjusted the tests to make the CI pass.

I am brining this up here, because this might break (type-checked) user code, just like it broke the test. However, I am not too concerned about this. `Fields` is a class, but the old test [used it as an interface](https://stackoverflow.com/questions/51716808/when-use-a-interface-or-class-in-typescript). So I can only hope that current user code doesn't abuse TS's structural type system to misuse classes as interfaces.

I just wanted to mention all this in case someone opens an issue about this.

daxpedda:
> This is no longer the case. Since fields of type `Option<T>` now have properly typed getters and setters, the type definition of the fields class now looks like this:
> 
> ```ts
> class Fields {
>     get hallo(): boolean | undefined;
>     set hallo(value: boolean | null | undefined);
>     spaceboy: boolean;
>     free(): void;
> }
> ```

I'm a bit confused, why is this happening? I can't seem to find the code changes that cause these to be getters/setters instead of fields. If this is caused by another PR then which and how?

RunDevelopment:
#4202 implemented this. You can see similar behavior in the tests of that PR in the `Foo#weird` property. Previously, WBG always emitted fields to type properties, but that was plain wrong.

In JS, getters and setter can return/accept data of different types, so the general way to write a getter-setter pair is

```ts
    get prop(): GetType;
    set prop(value: SetType);
```

But since `GetType` and `SetType` are often the same, it's simpler to write it such properties with the field syntax

```ts
    prop: GetAndSetType; // assumes GetAndSetType == GetType == SetType
```

So when it comes to typing properties, the field syntax is just a common special case.

Since the getter type and setter type of `Option<T>` properties is no longer the same, they can no longer use the field syntax and fall back to the general form.

daxpedda:
> Since the getter type and setter type of `Option<T>` properties is no longer the same, they can no longer use the field syntax and fall back to the general form.

Ah, completely overlooked that, thank you!

---

> I am brining this up here, because this might break (type-checked) user code, just like it broke the test. However, I am not too concerned about this. `Fields` is a class, but the old test [used it as an interface](https://stackoverflow.com/questions/51716808/when-use-a-interface-or-class-in-typescript). So I can only hope that current user code doesn't abuse TS's structural type system to misuse classes as interfaces.

I'm concerned that this is exactly whats happening. I wish I could get more sign-offs on this, but unfortunately I don't expect any.

I'm gonna ask around and see if I get any responses before merging it.
Sorry for the continuous delays!

daxpedda (approved):

RunDevelopment:
> I'm concerned that this is exactly whats happening.

Yeah, the fact that we had a test for this is a bad omen. So a third opinion would be nice.

However, the good news is that this at most causes type errors in code that misuses classes as interfaces. And we recently talked about type errors [not being breaking changes](https://github.com/rustwasm/wasm-bindgen/issues/4223#issuecomment-2457290892) on the side, so maybe it's not an issue at all.

daxpedda:
In theory this is definitely not an issue, especially because it would be an API misuse.

But my concern is in practice, I don't want to break maybe a whole ecosystem out there without even a warning.
E.g. what happened with `wasm-bindgen-rayon` (https://github.com/rustwasm/wasm-bindgen/issues/3330#issuecomment-1870338632).

RunDevelopment:
> But my concern is in practice, I don't want to break maybe a whole ecosystem out there without even a warning.

In general, we should probably add docs clarifying what guarantees WBG makes.

Also, maybe something like an RC/nightly/canary process could help? If nothing else, maybe a week or two before a new release, make an RC and a public call for testing, so people can report bugs before the actual release is made. No idea to what level cargo/crates.io supports this though (I'm still somewhat new to Rust). I'm just speaking from my experience in the npm ecosystem.

daxpedda:
> > But my concern is in practice, I don't want to break maybe a whole ecosystem out there without even a warning.
> 
> In general, we should probably add docs clarifying what guarantees WBG makes.

These kind of docs are in great need for improvement indeed, any help here would be appreciated!

> Also, maybe something like an RC/nightly/canary process could help? If nothing else, maybe a week or two before a new release, make an RC and a public call for testing, so people can report bugs before the actual release is made. No idea to what level cargo/crates.io supports this though (I'm still somewhat new to Rust). I'm just speaking from my experience in the npm ecosystem.

Would be nice for releases like this. Unfortunately the maintainer bandwidth isn't really there for that.
But yes, crates.io support is there for RC releases.

RunDevelopment:
> These kind of docs are in great need for improvement indeed, any help here would be appreciated!

Then I'll try to write down TypeScript guarantees somewhere the next few days.

> Would be nice for releases like this. Unfortunately the maintainer bandwidth isn't really there for that.

Are there no automated tools for that in Rust? 

E.g. there's [changesets](https://github.com/changesets/changesets) which automates changelog creating, publishing, versioning etc. for JS/TS monorepos. Admittedly, I only worked on projects that use `changesets` but haven't used to release anything myself yet. I just heard from people that do use it that it makes releasing a new version as simple as running a single command.

daxpedda:
> > Would be nice for releases like this. Unfortunately the maintainer bandwidth isn't really there for that.
> 
> Are there no automated tools for that in Rust?
> 
> E.g. there's [changesets](https://github.com/changesets/changesets) which automates changelog creating, publishing, versioning etc. for JS/TS monorepos. Admittedly, I only worked on projects that use `changesets` but haven't used to release anything myself yet. I just heard from people that do use it that it makes releasing a new version as simple as running a single command.

While making a new `wasm-bindgen` release is not a single command, its an effort of a couple minutes, its not the issue. The concern is more general, I'm more concerned with finding the time to review PRs and fixing issues then making beta releases and collecting feedback. But maybe I should get my priorities straight :sweat_smile:!

RunDevelopment:
> The concern is more general, I'm more concerned with finding the time to review PRs and fixing issues then making beta releases and collecting feedback.

Wouldn't a beta release process ease that burden? I mean, if something breaks, you'll get reports anyway, but there is less pressure to fix it in a beta release (and fewer duplicates), since it doesn't affect the whole ecosystem. As I see it, the main thing you buy yourself with a beta release is time and ease of mind. If the beta is broken and you're on vacation, most users won't even notice since they're on the latest stable release. Just fix it a month later (or wait until someone does a PR), make another beta, and repeat until it passes the scream test. This might slow down releases, but people that need the latest features can use beta and bear the cutting edge. But maybe my view on this all is a bit naive since I've been in the position of maintaining a project of the size and reach of WBG.

daxpedda:
No you are absolutely right, which is why I said earlier I should get my priorities straight :sweat_smile:!

But just as a reality check: I don't believe a lot of users will get reached by a beta. Unfortunately is is incompatible with any other dependency. So in practice, users will have to use `[patch.crates-io]` to try it out, I suspect few will. In any case, it should still be worth a shot!

I sent you a private Email about further discussing stuff like that, did you get it?

RunDevelopment:
> I sent you a private Email about further discussing stuff like that, did you get it?

Straight to spam. Thanks gmail 😆

daxpedda:
I'm gonna go ahead, rebase this and merge it.

daxpedda (changes requested):
This is good to go.
I wanted to rebase it myself and merge it, but it seems more involved then I thought.
So rebase it when you get to it please!

RunDevelopment:
Merge conflicts resolved. Thanks for merging this!

daxpedda (approved):

PR #4192, opened 2024-10-13
Fixed 2 minor bugs in JS glue code

RunDevelopment:
This separates the 2 bug fixes from #4189 into their own PR.

Changes:

1. `debugString` (this is the JS glue code version of `format!("{:?}", value)`) had a bug where it checked for a regex mismatch incorrectly. This would lead to dereferencing `null` at runtime, which throws a `TypeError`.
2. `_assertClass` returned `instance.ptr`. Here's the full code of the function:
    ```js
    function _assertClass(instance, klass) {
        if (!(instance instanceof klass)) {
            throw new Error(`expected instance of ${klass.name}`);
        }
        return instance.ptr;
    }
    ```
    Returning `instance.ptr` is weird for 2 reasons:
    - It's not used anywhere in the code. `_assertClass` is only used in one place and its return value is ignored. As @daxpedda [found out](https://github.com/rustwasm/wasm-bindgen/pull/4189#issuecomment-2408872554), the return value hasn't been used for 6 years now.
    - There is no `ptr` field. The `ptr` field was [renamed to `__wbg_ptr` one year ago](https://github.com/rustwasm/wasm-bindgen/pull/4189#issuecomment-2408715540) and this function wasn't updated.
    
    So I just removed the return statement and `instance.ptr`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 * Fixed `#[should_panic]` not working with `#[wasm_bindgen_test(unsupported = ...)]`.
   [#4196](https://github.com/rustwasm/wasm-bindgen/pull/4196)
 
-* Fixed a minor bug with regexes when printing debug information.
+* Fixed potential `null` error when using `JsValue::as_debug_string()`.
   [#4192](https://github.com/rustwasm/wasm-bindgen/pull/4192)
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


daxpedda (changes requested):
Thank you!
LGTM.

But missing a changelog entry for the regex issue.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 * Fixed unused string enums generating JS values.
   [#4193](https://github.com/rustwasm/wasm-bindgen/pull/4193)
 
+* Fixed a minor bug with regexes when printing debug information.
```
daxpedda, line 22:
```suggestion
* Fixed potential `null` error when using `JsValue::as_debug_string()`.
```

daxpedda (approved):

PR #4193, opened 2024-10-13
Only generate Js values for string enums if used

RunDevelopment:
This fixes the issue we talked about [here](https://github.com/rustwasm/wasm-bindgen/pull/4180#discussion_r1797780272). Basically, in #4147 I changed the bindings of string enums to use a variable storing an array (point 3). However, this array was always generated regardless of whether it was actually used or not.

This PR fixes the issue by marking which enums are used as the bindings are generated. Since we always generate the bindings of function that can use string enums before the variable of a string enum, this works out. I used the same `expose_*` pattern for the function that is commonly used for this type of task.


daxpedda (approved):
Thank you, this is great!

PR #4200, opened 2024-10-15
Added missing space in doc comments

RunDevelopment:
This PR just fixes something super minor that bugged. When a Rust doc comment doesn't start with a space, the JSDoc comment won't either. None of the web-sys doc comments do, so web-sys string enums looked horrible.

This PR fixes this by inserting a space if the doc comment doesn't start with one. This means that properly-formatted Rust doc comments remain unchanged, but bad ones get "auto fixed".


daxpedda (approved):
Thank you!

PR #4201, opened 2024-10-15
Support regular JS numbers for incoming 64-bit integers

### LINKED ISSUE #4113: Can't return `Number` values as `u64`
### Describe the Bug

When creating bindings to a function with a `u64` return value, the returned value is assumed to be of type `BigInt` in the generated bindings. It seems like it would be possible to get `Number` values containing integers between 2^32 and 2^53-1 as an integer type directly in the bindings, but it doesn't appear to be possible currently.

I couldn't find any documentation explicitly saying "use f64 if you want integers over u32 and convert on the Rust side if you need, it's the same size anyway" (I may have just missed it), but that seems to be the approach that's been taken eg for `Date.now()` in js_sys.

### Steps to Reproduce

1. Create a simple alternative binding to `Date.now()` expecting a `u64` and a test that calls it:
```rs
#[wasm_bindgen]
extern "C" {
    pub type Date;

    #[wasm_bindgen(static_method_of = Date)]
    pub fn now() -> u64;
}

#[cfg(test)]
pub mod tests {
    use super::Date;
    use wasm_bindgen_test::*;

    #[wasm_bindgen_test]
    pub fn u64_now() {
        Date::now();
    }
}
```
2. Run `wasm-pack test --node`

### Expected Behavior

A binding is produced that can handle a `Number` or `BigInt` return value

### Actual Behavior

Test fails with `wasm-bindgen: imported JS function that was not marked as `catch` threw an error: expected a bigint argument, found number`

RunDevelopment:
fixes #4113

This adds support for regular JS numbers (think `typeof x === 'number'`) as values for `u64` and `i64` parameters. This should makes Rust APIs taking large numbers more convenient and allows bindings for JS functions that return large numbers (e.g. `Date.now()`).

For details on the `number` -> `bigint` conversion, see the doc comment for `number_to_bigint`. Otherwise, the implementation of this PR is quite simple. The correctness of the conversion is tested in the `optional_primitives` test.

The reason this PR is still a draft is that I haven't updated the code for function signature types yet. E.g. `arg: u64` is still `arg: bigint` in TS, even thought it should be `arg: bigint | number` now. I'm waiting for #4188, because #4188 add exactly the functionality I need to implement this.


guybedford:
WebAssembly core modules are defined such that i64 types always require BigInt values to be passed, throwing a TypeError otherwise.

I think we should likely stick with this design in aligning with core, instead of trying to add more sugar in this case unfortunately.

PR #4202, opened 2024-10-15
Fixed generated types for getters and setters

RunDevelopment:
While working on #4201, I had the idea to test how wasm bindgen handles getter-setter pairs with different types. Not well was the answer.

This PR completely changes how the type definitions for getters and setters are generated and fixes several bugs:

1. Static getters/setters with the same name as instance getters/setters would interfere with each other and only result in one generated property. Whether this property was static depended on the order in which the getters/setters were declared in Rust.
2. Getters and setters with different types only produced one property with one of the types. Which type the property had depended on the order in which the getter and setter were declared in Rust.
3. (minor bug) Write-only properties (setter but no getter) produced incorrect type definition that allowed reading the property.

There were also some minor issues:
- Fields were sorted by name and not written in the order in which they were declared like everything else.
- Doc comments weren't indented properly.

This PR fixes all of those issues and adds a test to verify correctness.

Fixing these issues is a requirement for #4188 and #4201.


CHANGELOG.md
```diff
@@ -28,6 +28,12 @@
 * Fixed potential `null` error when using `JsValue::as_debug_string()`.
   [#4192](https://github.com/rustwasm/wasm-bindgen/pull/4192)
 
+* Fixed generated types when the getter and setter of a property had different types.
+  [#4202](https://github.com/rustwasm/wasm-bindgen/pull/4202)
+
+* Fixed generated types when a static getter/setter the same name as an instance getter/setter.
```
daxpedda, line 34:
```suggestion
* Fixed generated types when a static getter/setter has the same name as an instance getter/setter.
```

crates/cli-support/src/js/mod.rs
```diff
@@ -1164,6 +1165,124 @@ __wbg_set_wasm(wasm);"
...
+                ts_dst.push_str(is_static);
+                ts_dst.push_str("set ");
+                ts_dst.push_str(name);
+                ts_dst.push_str("(value: ");
+                ts_dst.push_str(&setter.ty);
+                if setter.is_optional {
+                    ts_dst.push_str(" | undefined");
+                }
+                ts_dst.push_str(");\n");
+            };
```
daxpedda, line 1215: None of those have to be closures, right?
How about methods on `FieldAccessor`?

daxpedda (changes requested):
Just some nits.

crates/cli-support/src/js/mod.rs
RunDevelopment: `write_[gs]etter` capture the local variable `is_static`, but I really just wanted to have the printing logic all in one place. I could make them methods of `FieldAccessor`, but that would spread out the printing logic over multiple functions and 2 types. I prefer to have related things close together.

daxpedda: Ah, somehow missed that capture.

> Also, is there a reason not to use closures for helper functions?

No, I just find it harder to read.
My experience so far in Rust has been that most people tend to avoid closures when possible.

No worries though!

CHANGELOG.md
```diff
@@ -28,6 +28,12 @@
 * Fixed potential `null` error when using `JsValue::as_debug_string()`.
   [#4192](https://github.com/rustwasm/wasm-bindgen/pull/4192)
 
+* Fixed generated types when the getter and setter of a property had different types.
```
daxpedda, line 31:
```suggestion
* Fixed generated types when the getter and setter of a property have different types.
```

daxpedda (commented):
LGTM, but I will give it another review with a fresh set of eyes tomorrow.
Thank you, this is great work!

daxpedda (approved):
Apologies for the delay, was sick and had to play catch-up with my other projects before getting back to this again.

LGTM!

RunDevelopment:
No worries. I'm glad you're feeling better!

And thanks for the review!

PR #4204, opened 2024-10-16
Add support for enums with negative discriminants

### LINKED ISSUE #2313: A negative custom discriminant in a fieldless number enumeration seems unsupported
In Rust, it's OK to have a a negative discriminant for a number enum variant:

```
pub enum NumberEnum {
    Foo = 0,
    Bar = -1,
    Qux = 2,
}
```

When compiled with wasm-bindgen on Mac OS, however, the following code:

``` 
#[wasm_bindgen]
pub enum NumberEnum {
    Foo = 0,
    Bar = -1,
    Qux = 2,
}
```

Produces the following error

```
error: enums with #[wasm_bidngen] may only have number literal values
   --> src/trade_manager.rs:704:11
    |
704 |     Bar = -1,
    |           ^^
```

RunDevelopment:
fixes #2313

This adds support for enums with negative discriminants and slightly improves error messages.

Honestly, implementing this was pretty simple. I don't have too much to say about this, just a few notes on implementation details:

1. Within the serialized data for ASTs and descriptors, variants are still passed around as `u32`. Instead of representing the actual value, they now represent the bits of the `i32` or `u32` value. Which one it is depends on the new `signed: bool` field. I wanted to use `i64`, but this doesn't work well with `inform` or other serialization, so I opted to "decode" the variant values to i64 on the CLI side just before JS code gen.
2. Holes. The parser will only search for non-negative holes. This means that a signed enum with 2^31 variants will fail to find a hole, but I think that's okay. Not having to worry about the sign of a hole also makes the downstream code handling holes simpler.


crates/backend/src/codegen.rs
```diff
@@ -1571,13 +1576,13 @@ impl ToTokens for ast::Enum {
             #[automatically_derived]
             impl #wasm_bindgen::convert::OptionFromWasmAbi for #enum_name {
                 #[inline]
-                fn is_none(val: &u32) -> bool { *val == #hole }
+                fn is_none(val: &Self::Abi) -> bool { *val == #hole as #underlying }
```
daxpedda, line 1579: The `as` conversion here shouldn't be necessary, right?

crates/backend/src/codegen.rs
```diff
@@ -1571,13 +1576,13 @@ impl ToTokens for ast::Enum {
...
             }
 
             #[automatically_derived]
             impl #wasm_bindgen::convert::OptionIntoWasmAbi for #enum_name {
                 #[inline]
-                fn none() -> Self::Abi { #hole }
+                fn none() -> Self::Abi { #hole as #underlying }
```
daxpedda, line 1585: As above.

crates/macro-support/src/parser.rs
```diff
@@ -1498,15 +1520,40 @@ impl<'a> MacroParse<(&'a mut TokenStream, BindgenAttrs)> for syn::ItemEnum {
...
+        }
+
+        // To make all the code handling holes simpler, we only consider
+        // non-negative holes. This allows us to use `u32` to represent holes.
+        let max = if signed {
+            i32::MAX as i64
+        } else {
+            u32::MAX as i64
+        };
+        let hole = (0..max + 1)
```
daxpedda, line 1547:
```suggestion
        let hole = (0..=max)
```

crates/macro-support/src/parser.rs
daxpedda, line 1: Could we move all the checks (parsing, range, `i32`/`u32`) to one place?
E.g. after we collect all the `syn::Expr`?

daxpedda (changes requested):
Thank you for doing this!

crates/macro-support/src/parser.rs
RunDevelopment: It kinda hurts my soul to know that the iteration is less efficient with `..=`, but I agree that this is more readable.

daxpedda: Why would it be :open_mouth:?

crates/backend/src/codegen.rs
RunDevelopment: No, it's necessary because `Self::Abi` is either `u32` or `i32` while `hole` is always `u32`. So this would be a type error without the cast. This is because `quote` preserves the type of interpolated values. I.e. [this is how tokens are generated for u32](https://docs.rs/quote/latest/src/quote/to_tokens.rs.html#176-180).

daxpedda: I see, thanks!

crates/macro-support/src/parser.rs
RunDevelopment: When you say "checks" you mean everything that bails, right? Otherwise, I'm not sure what you're referring to, since this comment was added to the file and not a specific line of code.

But if I guessed right, then this might be difficult to do. We at least have to check the upper range (`u32::MAX`) while parsing all variants to prevent implicit variants from overflowing, and range checks have to be done for parsed and implicit discriminants.

daxpedda: > When you say "checks" you mean everything that bails, right?

Yes.

> But if I guessed right, then this might be difficult to do. We at least have to check the upper range (`u32::MAX`) while parsing all variants to prevent implicit variants from overflowing, and range checks have to be done for parsed and implicit discriminants.

I'm unsure what you are trying to say, it sounds like I suggested to remove those "checks".
So while you are looping through the discriminants, you already parse them or calculate implicit discriminants. Then you do upper bound checks for `i32::MIN` and `u32::MAX`. After you collect those results, you check again for upper bounds ranges against `i32` or `u32`.

I'm suggesting that you just collect `syn::Expr`, or the lack thereof, directly, and do all that afterwards.
Because right now all these validations are spread out.

RunDevelopment: > After you collect those results, you check again for upper bounds ranges against i32 or u32.

You are referring to the `if *max.0 > i32::MAX as i64` check, right? This check is there to ensure that enums like this are handled correctly:

```rs
enum E {
  A = u32::MAX,
  B = -1,
}
```

The tricky thing with `E` is that we cannot truly check the allowed range of discriminants in the first pass where we parse the discriminants, because we need to know the underlying type (`i32` or `u32`) to do that. But we only figure out the underlying type (stored in the `signed` variable) by looking at the parsed values of discriminants. So we have to process `E::B` to know that the enum is `i32` and therefore `E::A` is out of range.

So the validation will always be spread out over 2 passes.

I'm sorry, but I just don't understand how collecting the `syn::Expr`s would improve this, so I might still not understand what you are suggesting.

daxpedda: > I'm sorry, but I just don't understand how collecting the `syn::Expr`s would improve this, so I might still not understand what you are suggesting.

My suggestion is to avoid checking in multiple places!
Right now you are doing some checks inside the collection loop, and then some afterwards.

What I'm suggesting is just do them all in one place, after the collection loop.
I agree, this is still requires two passes, which can still be done in one place.

I hope this makes more sense :sweat_smile:!

RunDevelopment: I'm sorry, but I still don't see how what you're suggesting works. What would the collection loop do then? Or put another way, what is the type of the items the loop collects? Right now, the loop collects `ast::Variant`, but that requires a fully parsed discriminant (and implicit discriminants computed). So what's the shape of the data you're collecting?

daxpedda: I took a stab at what I'm proposing. Basically it boils down to:
```rust
        let signed = self.variants.iter().any(|v| {
            if let Some((_, expr)) = &v.discriminant {
                matches!(
                    get_expr(expr),
                    syn::Expr::Unary(syn::ExprUnary {
                        op: syn::UnOp::Neg(_),
                        ..
                    })
                )
            } else {
                false
            }
        });
```
Which would move all `bail!`s into one section. Now instead of having separate error messages for 32-bit integer and signed/unsigned we can merge them. The code complexity is going to increase, but we get better UX imo.

RunDevelopment: Okay, now I get it. Thank you for your patience... 

I also agree that this would be better since the error message would be more specific, so let's go with that. Thank you for the suggestion!

RunDevelopment: I implemented your approach with the `NumericValue` struct. This allows use to use the same logic for the first pass that figures out whether the enum is signed and the second pass that actually parses the discriminant value. This might be a tiny bit overengineered, but it feels like a cleaner solution than passing the sign+digits value around as a tuple.

daxpedda (approved):
Nice!

PR #4205, opened 2024-10-16
Improved indentation and enum formatting

RunDevelopment:
Enums are now formatted with each variant on its own line. This makes enum definitions easier to read. It also allowed me to remove the hard-coded enum detection in `reset_indentation`, because it wasn't necessary anymore.

I also fixed a minor bug in `reset_indentation`. If a line started or ended with `}`, it decreased the indentation. This is the wrong condition, but it will lead to code after a braced(?) single-line statement (e.g. `if (cond) { doStuff() }`) to be indented incorrectly. This type of braced single-line statement is generated rarely, but it is used in a few places ([example](https://github.com/rustwasm/wasm-bindgen/blob/76776ef54cce17f1c7442b010e26d7eadaea3cd0/crates/cli-support/src/js/binding.rs#L924)). Rare or not, I believe `reset_indentation` should handle this case correctly.

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -502,8 +502,8 @@ fn reset_indentation(s: &str) -> String {
     for line in s.lines() {
         let line = line.trim();
 
+        // handle doc comments separately
         if is_doc_comment(line) {
-            // handle doc comments separately
             for _ in 0..indent {
                 dst.push_str("    ");
             }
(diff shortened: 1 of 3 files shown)


crates/cli-support/src/lib.rs
```diff
@@ -498,9 +498,22 @@ fn reset_indentation(s: &str) -> String {
 
     for line in s.lines() {
         let line = line.trim();
-        if line.starts_with('}') || (line.ends_with('}') && !is_doc_comment(line)) {
+
+        if is_doc_comment(line) {
+            // handle doc comments separately
```
daxpedda, line 503:
```suggestion
        // handle doc comments separately
        if is_doc_comment(line) {
```

daxpedda (approved):
LGTM!
Thank you!

Issue #4207, opened 2024-10-17
Functions returning Result<T,E> result in incorrect typescript function output type `Array`


wandomPewlin:
### Describe the Bug

In version 0.2.95, any function returning a `Result<T,E>` will cause the generated typescript function to return `Array` (note that it's not `Array<T>`, just `Array`). The problem doesn't occur in version 0.2.93.

### Steps to Reproduce

Using the code from `wasm-bindgen` [doc](https://rustwasm.github.io/wasm-bindgen/api/wasm_bindgen/struct.JsError.html)

```rust
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn throwing_function() -> Result<(), JsError> {
    Err(JsError::new("message"))
}
```

### Expected Behavior

in version 0.2.93, the generated function signature in typescript would be 

```typescript
export function throwing_function(a: number): void;
```

### Actual Behavior

but in 0.2.95, it becomes.

```typescript
export function throwing_function(): Array;
```

### Additional Context

The problem doesn't seem to be `JsError` itself, I originally run into the problem while using `serde-wasm-bindgen` and the function returns `Result<i32,JsValue>`. The generated ts function also returns `Array` as output.

RunDevelopment:
Cannot reproduce.

I just tested your code on `main` and the function returns `void` as expected (in JS and in the type definitions). So this is probably an issue with `serde-wasm-bindgen`.

davids-work:
This is dependent on the rust version! Something has happened between Rust 1.81 and 1.82 that triggers this behavior.

Using the code from the original post, I can reproduce it like this:
```
rustup toolchain add 1.82.0
rustup toolchain add 1.81.0
rustup override set 1.81.0
wasm-pack build --target web
cp pkg/array.d.ts array-1.81.d.ts
rustup override set 1.82.0
wasm-pack build --target web
diff -u pkg/array.d.ts array-1.81.d.ts
```
Gives me
```
 export interface InitOutput {
   readonly memory: WebAssembly.Memory;
-  readonly throwing_function: (a: number) => void;
-  readonly __wbindgen_add_to_stack_pointer: (a: number) => number;
+  readonly throwing_function: () => Array;
+  readonly __wbindgen_export_0: WebAssembly.Table;
+  readonly __externref_table_dealloc: (a: number) => void;
+  readonly __wbindgen_start: () => void;
 }
```

[...]

PR #4208, opened 2024-10-17
More tests for result type

RunDevelopment:
I renamed `result-string.rs` -> `result.rs` and added 2 more functions for testing.

This also shows that #4207 is not a `wasm_bindgen` issue per se.


daxpedda:
PR just adding tests :heart:
Thank you!

PR #4210, opened 2024-10-18
Fixed invalid type annotation for multivalue returns

### LINKED ISSUE #4207: Functions returning Result<T,E> result in incorrect typescript function output type `Array`

RunDevelopment:
fixes #4207

`Array` itself is not a valid type because it requires one generic parameter. I replaced it with `number[]`. I chose `number` as the element type, because that's what the surrounding code is using for everything.

Using `number` for everything is probably wrong, since WASM uses `bigint` for 64-bit ints AFAIK, but it shouldn't matter too much.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,9 @@
 * Fixed generated types when a static getter/setter has the same name as an instance getter/setter.
   [#4202](https://github.com/rustwasm/wasm-bindgen/pull/4202)
 
+* Fixed invalid TypeScript return types for multivalue signatures.
+  [#4210](https://github.com/rustwasm/wasm-bindgen/pull/4210)
+
 --------------------------------------------------------------------------------
 
 ## [0.2.95](https://github.com/rustwasm/wasm-bindgen/compare/0.2.94...0.2.95)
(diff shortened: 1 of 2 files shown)


RunDevelopment:
CI fails because CI uses the latest stable compiler (Rust 1.82 now), which uses multi-value returns by default. This means that the generated WASM and JS code will be different in all tests now.

daxpedda (approved):
Yes, we should probably read out the `ValueType`, but this is already a bug, so this doesn't degrade the current state.

PR #4215, opened 2024-10-20
Add `rename_all = "camelCase"` option to automatically rename methods and functions

### LINKED ISSUE #1818: Automatically camelify names
### Motivation

Right now you can use `js_name` to convert a Rust `foo_bar` into JS `fooBar` (or vice versa). But 99% of the time you want it to be camel case, so this just ends up being repetitive boilerplate:

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_name = fooBar)]
    fn foo_bar(...);

    #[wasm_bindgen(js_name = quxCorge)]
    fn qux_corge(...);

    #[wasm_bindgen(js_name = yesNo)]
    fn yes_no(...);
}
```

### Proposed Solution

wasm-bindgen should *automatically* camelify JS names, so `foo_bar` will be translated into `fooBar` on import/export.

It will still be possible to use `js_name` to manually override that if the user does *not* want camel-case.

This is a breaking change, so we should batch it together with other breaking changes.

RunDevelopment:
fixes #1818

This adds support for `rename_all = "camelCase"` to camelify names inside a block. 

Under the hood, I added support for a more general renaming system mimicking serde's renaming system. If we want to camelify methods by default in the next breaking change, we just have to change the default values for those renaming rules.

Examples:

```rs
#[wasm_bindgen(rename_all = "camelCase")]
extern "C" {
    #[wasm_bindgen(js_name = foo_bar)] // explicit js name
    fn foo_bar();
    fn qux_corge(); // js_name = quxCorge
    #[wasm_bindgen(js_namespace = Baz)]
    fn yes_no(); // js_name = yesNo
}


#[wasm_bindgen(rename_all = "camelCase")]
pub struct RustStruct {
    pub foo: u32, // js_name = foo
    pub some_cool_field: u32, // js_name = someCoolField
    #[wasm_bindgen(js_name = "another_field_for_you")]
    pub another_field: u32,
}

#[wasm_bindgen] // renaming is opt in, so nothing changes for existing code
impl RustStruct {
    pub fn i_dont_get_renamed() {}
}

#[wasm_bindgen(rename_all = "camelCase")]
impl RustStruct {
    // js_name = incrementFoo
    pub fn increment_foo(&mut self, amount: Option<u32>) {
        self.foo += amount.unwrap_or(1);
    }

    // js_name = setFoo
    pub fn set_foo(&mut self, foo: u32) {
        self.foo = foo;
    }

    #[wasm_bindgen(js_name = get_another)] // explicit js name
    pub fn another(&self) -> u32 {
        self.another_field
    }
}
```

### Open questions

I mostly don't like the name `rename_all`, because it's not clear that it:

1. Only affects methods/functions and fields,
2. Don't affect explicitly given `js_name`s.


crates/macro-support/src/lib.rs
```diff
@@ -104,9 +104,70 @@ pub fn expand_class_marker(
...
+}
+impl RenameRule {
+    fn snake_case_to_camel_case(name: String) -> String {
+        let mut camel = String::new();
+        let mut capitalize = false;
+        for c in name.chars() {
+            if c == '_' {
+                capitalize = true;
+            } else if capitalize {
+                camel.push(c.to_ascii_uppercase());
```
daxpedda, line 124:
```suggestion
                camel.push(c.to_uppercase());
```
I'm not sure if we even allow non-ASCII symbols.

crates/macro-support/src/lib.rs
```diff
@@ -104,9 +104,70 @@ pub fn expand_class_marker(
...
+    }
+
+    /// Applies the rule to the field of a struct or enum variant data.
+    fn apply_to_field(self, name: String) -> String {
+        self.rename_snake_case(name)
+    }
+    /// Applies the rule to the method of a struct or enum.
+    fn apply_to_method(self, name: String) -> String {
+        self.rename_snake_case(name)
+    }
```
daxpedda, line 147: Why are we separating those?

daxpedda (changes requested):
> I mostly don't like the name `rename_all`, but it's not clear that it:
> 1. Only affects methods/functions and fields,
> 2. Don't affect explicitly given `js_name`s.

I think your current implementation sounds good to me: affects methods, static methods and fields and explicit `js_name`s aren't overwritten.

`camelify_by_default` sounds a bit too funny to me. How about `default_rename_camel_case`? Kinda terrible as well, let me know if you have any better ideas.

crates/macro-support/src/lib.rs
RunDevelopment: We don't.

```rs
#[wasm_bindgen]
pub fn nöp() {}
```
```
        error[E0754]: `#[no_mangle]` requires ASCII identifier
         --> C:\Users\micha\Git\wasm-bindgen\crates\cli\tests/reference\nop.rs:3:1
          |
        3 | #[wasm_bindgen]
          | ^^^^^^^^^^^^^^^
          |

        For more information about this error, try `rustc --explain E0754`.
        error: could not compile `reference-test` (lib) due to 1 previous error

```

Since I know a bit about how WBG handles name mangling because of #4230, I thought about supporting arbitrary strings as JS names after that PR is through. This would naturally add support for non-ASCII identifiers as well.

But I'm not sure whether this means we should use `to_ascii_uppercase` or `to_uppercase`. What do you think?

crates/macro-support/src/lib.rs
daxpedda: Lets not overextend, this can always be expanded when we actually get there.

RunDevelopment:
> `camelify_by_default` sounds a bit too funny to me. How about `default_rename_camel_case`? Kinda terrible as well, let me know if you have any better ideas.

[This comment](https://github.com/rustwasm/wasm-bindgen/issues/1818#issuecomment-543216831) suggests `auto_camel_case`, which I kinda like.

Also, maybe we could use something like `experimental_auto_camel_case`? Right now, this feature feels really underspecified to me, so I think it would be good to get some actual user feedback on it before we commit to a specific behavior. Especially since changing the behavior would be a breaking change.

What do you think?

---

On a different note: static properties. Static properties and readonly variables sometimes use SCREAMING_SNAKE_CASE similar Rust `const` and `static` variables. We should probably detect that, right?

In general, the renaming assumes snake_case, but doesn't verify it. Should we only rename names that we can verify to be snake_case?

daxpedda:
> On a different note: static properties. Static properties and readonly variables sometimes use SCREAMING_SNAKE_CASE similar Rust `const` and `static` variables. We should probably detect that, right?
> 
> In general, the renaming assumes snake_case, but doesn't verify it. Should we only rename names that we can verify to be snake_case?

Hm, good point. In which case wouldn't we be looking at something like `auto_jsify` doing it all in one go instead of having multiple ones? Which in the future would be enabled by default.

RunDevelopment:
> In which case wouldn't we be looking at something like auto_jsify doing it all in one go instead of having multiple ones?

Not really. My point was the JS devs sometimes use camelCase and sometimes SCREAMING_SNAKE_CASE for constants/static variables/getters, so we probably shouldn't make everything camelCase.

Thinking about this, JS devs generally don't use lower_snake_case for anything. So how about we make the behavior of `auto_camel_case` (bike shed pending) that it converts all lower_snake_case names to camelCase, regardless of whether it's a method, field, or something else? This means that there is only one simple rule that people have to learn and it applies equally everywhere. This makes the behavior predictable and consistent, which is pretty important for something that might be on by default in the future.

What do you think?

crates/macro-support/src/lib.rs
RunDevelopment: I replaced with all with a single `snake_case_to_camel_case` function.

daxpedda:
> > In which case wouldn't we be looking at something like auto_jsify doing it all in one go instead of having multiple ones?
> 
> Not really. My point was the JS devs sometimes use camelCase and sometimes SCREAMING_SNAKE_CASE for constants/static variables/getters, so we probably shouldn't make everything camelCase.

Is there no standard? If the goal is to do this automatically in the next breaking change, what are we going to convert it to then?

> Thinking about this, JS devs generally don't use lower_snake_case for anything. So how about we make the behavior of `auto_camel_case` (bike shed pending) that it converts all lower_snake_case names to camelCase, regardless of whether it's a method, field, or something else? This means that there is only one simple rule that people have to learn and it applies equally everywhere. This makes the behavior predictable and consistent, which is pretty important for something that might be on by default in the future.
> 
> What do you think?

Basically boils down to the question above. We should figure out what we want to do automatically in the future and implement this here. But yes, I agree we should "fix" *all* names, not just methods.

But I think we should do it regardless of what the user names their items, explicit names should be set with the `js_name` attribute.

RunDevelopment:
> Is there no standard?

No. In JS, the best we have are [common](https://www.robinwieruch.de/javascript-naming-conventions/) [naming](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript#function_names) [conventions](https://google.github.io/styleguide/jsguide.html#naming). JS itself uses SCREAMING_SNAKE_CASE for constants (e.g. `Math.PI` and `Number.MAX_SAFE_INTEGER`), but camelCase for constants in common in the wider ecosystem as well. But aside from constants, almost everyone uses camelCase for function/methds/variables and PascalCase for classes/interfaces/types.

This means that names that are already in PascalCase, camelCase, or SCREAMING_SNAKE_CASE, are most likely correct. So we only really have to rename snake_case names, because snake_case is most likely "wrong".

> But I think we should do it regardless of what the user names their items, explicit names should be set with the `js_name` attribute.

I don't think that what you're saying is wrong, but I think it might not be practical - for us and our users. 

If you want to convert more than just snake_case names, then you have to be able to detect and convert different naming conversions. That's not easy, because there are many many edge cases (e.g. a constant called `MATRIX_4x3`). Once set, we can't really change the behavior of renaming anymore. Any change to renaming, no matter how minor, will a breaking change. And the more complex cases renaming has to handle, the higher the chance for bugs and edge cases to sneak in.

For users, it's also less practical. Renaming is essentially guessing the correct name for the user. If we guess wrong, the user has to fix it with `js_name`. I think that the way we guess should be simple and predictable for the user, so that they can rely on it. I don't want the renaming behavior to be so complex that users essentially just have to "try and find out" what WBG renames more difficult names to.

RunDevelopment:
> > No. In JS, the best we have are [common](https://www.robinwieruch.de/javascript-naming-conventions/) [naming](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript#function_names) [conventions](https://google.github.io/styleguide/jsguide.html#naming).
> 
> Hm, that's bad.

Yep. But if we take SCREAMING_SNAKE_CASE for static constants, naming isn't too different from Rust:

| | Rust | JavaScript/TypeScript |
| --- | --- | --- |
| variables/parameters | snake_case | camelCase |
| classes/enums/types | PascalCase | PascalCase |
| enum variants | PascalCase | n/a (or PascalCase if [implemented like tsify](https://github.com/madonoharu/tsify?tab=readme-ov-file#enum)) |
| instance/static methods | snake_case | camelCase |
| instance fields/properties | snake_case | camelCase |
| class static fields/properties | n/a | camelCase |
| (static) constants | SREAMKING_SNAKE_CASE | SREAMKING_SNAKE_CASE |

> I don't think I'm following here. It seems much more straightforward to me that we simply rename everything regardless, instead of making any exceptions in specific cases that users have to understand in the first place. No guessing required. It seems more predictable to me.

My point was that to convert everything to one naming convention, you first need to detect the casing of the identifier. This is why I said that `MATRIX_4x3` is an edge case, but it's neither snake_case nor SCREAMING_SNAKE_CASE strictly speaking.

But that's not really a problem IF we assume Rust naming conventions. I just didn't want to make that assumption, because there are a lot of WBG users that use camelCase for methods in Rust *right now* due to the lack of automatic renaming. Even if we discourage not following Rust naming conventions, I think we should still support it to some degree for compatibility with existing camelCase identifiers.

That said, it feels like we're not 100% talking about the same thing here. I'm not quite sure what you mean by "rename everything regardless". As you can see in the above table, we only need to rename snake_case identifiers, so what's everything here?

daxpedda:
> > > No. In JS, the best we have are [common](https://www.robinwieruch.de/javascript-naming-conventions/) [naming](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Writing_style_guide/Code_style_guide/JavaScript#function_names) [conventions](https://google.github.io/styleguide/jsguide.html#naming).
> > 
> > 
> > Hm, that's bad.
> 
> Yep. But if we take SCREAMING_SNAKE_CASE for static constants, naming isn't too different from Rust:

Sure, lets just go with that. However, we might want to consider not converting anything that matches naming conventions. E.g. Struct and enum types should be pascal case in both languages, so if the user doesn't use pascal case in Rust, we can take that as being set explicit. So no conversion necessary. The same would apply to static constants.

That might just resolve your concern of practicality here.

> My point was that to convert everything to one naming convention, you first need to detect the casing of the identifier. This is why I said that `MATRIX_4x3` is an edge case, but it's neither snake_case nor SCREAMING_SNAKE_CASE strictly speaking.

We only need to detect the casing and spacing, not necessarily which convention its following. I think its pretty clear how `MATRIX_4x3` will be converted to lets say SCREAMING_SNAKE_CASE: `MATRIX_4X3`.

> But that's not really a problem IF we assume Rust naming conventions. I just didn't want to make that assumption, because there are a lot of WBG users that use camelCase for methods in Rust _right now_ due to the lack of automatic renaming. Even if we discourage not following Rust naming conventions, I think we should still support it to some degree for compatibility with existing camelCase identifiers.

Compatibility with current code is not a concern, right? Users still need to explicitly set the attribute.

> That said, it feels like we're not 100% talking about the same thing here. I'm not quite sure what you mean by "rename everything regardless". As you can see in the above table, we only need to rename snake_case identifiers, so what's everything here?

I was explicitly responding to this:
> This means that names that are already in PascalCase, camelCase, or SCREAMING_SNAKE_CASE, are most likely correct. So we only really have to rename snake_case names, because snake_case is most likely "wrong".

So regardless of what naming convention a user has used, we should convert every property and method to the naming convention according to the table you set above.

Though keep in mind my suggestion at the start of the post, where we could just not do any conversion for static constants, as they should be SCREAMING_SNAKE_CASE for both languages. In which case if a user is using PascalCase for their static constants, we won't convert it to SCREAMING_SNAKE_CASE. They should get a Rust warning though.

daxpedda:
> > we might want to consider not converting anything that matches naming conventions. E.g. Struct and enum types should be pascal case in both languages, so if the user doesn't use pascal case in Rust, we can take that as being set explicit. So no conversion necessary.
> 
> I agree. This is also what I wanted to do. I also agree that users not following Rust naming conventions is a strong signal that they are doing it on purpose.
> 
> So to summarize, the renaming strategy for identifiers will be:
> 
>     1. If the identifier conforms to the Rust naming convention and JS/TS uses a different naming convention (as define in the previous table), convert the Rust identifier to the JS/TS naming convention.
> 
>     2. Otherwise, use the Rust identifier as is in JS.
> 
> 
> This is what you mean, right? If so, then I 100% agree with it.

Yes, I see now that there was a misunderstanding on my part before. I was somehow assuming that we need to figure out which naming convention e.g. enum types are to check if they follow the Rust convention. But because there is nothing for us to do if they don't, we don't actually need to that.

> On a related note: should we rename function parameters? They fulfill the criteria, but parameter names aren't really observable in JS (kinda, you can observe them with hacks, but you must not rely on them). Renaming them might also cause problems with user-defined JSDoc `@param` tags.

Yes. Lets move the user-defined JSDoc conversation to real-time because this needs a bit of a longer back and forth.

> We may also want to change the attribute from `auto_camel_case` to something like `auto_rename`. It only "camelCases" right now, but I don't want to exclude the possibility of having different conversions in the future. E.g. we may want to generate getters for public associated constants on structs/enums in the future.

Yes ... finally a good name.

> > In which case if a user is using PascalCase for their static constants, we won't convert it to SCREAMING_SNAKE_CASE. They should get a Rust warning though.
> 
> Clippy already emits a warning for not following Rust naming conventions. So we don't have to do anything, I think.

Yeah, we just need to check via UI tests, but I know this is hard to get right, so its not a requirement for a PR.

mmgeorge:
@RunDevelopment just wondering if you are still working on this - would be a really great feature to have! Potentially interested in helping out.

RunDevelopment:
@mmgeorge No, I'm not working on WBG anymore. Feel free to fork/copy my PR and pick up where I left off though!

guybedford:
This would be really nice to have. Is there still interest from the author in finishing it?

RunDevelopment:
There isn't, so I'm going to close this now. Feel free to use a copy of the branch and finish it if need be.

guybedford:
Ok, I'll keep this open so we can continue to track it, but hopefully we can resolve it soon.

PR #4219, opened 2024-10-23
wasm-bindgen-webidl: use `clap` instead of `structopt`

ognevny:
structopt is unmaintained, it uses very old version of clap, which pulls some old dependencies


ognevny:
CI failure seems not related to this change

daxpedda (approved):
Thank you, its really appreciated!

PR #4221, opened 2024-10-24
Add `AbstractRange` and `StaticRange` interfaces

behrenle:
This pull request adds support for the `AbstractRange` and `StaticRange` interfaces to wasm-bindgen, enabling Rust developers to interact with these DOM range objects for improved web API compatibility.


crates/web-sys/webidls/enabled/Range.webidl
```diff
@@ -10,10 +10,10 @@
...
  * Copyright © 2012 W3C® (MIT, ERCIM, Keio), All Rights Reserved. W3C
  * liability, trademark and document use rules apply.
- */
+ */ 
 
 [Constructor]
-interface Range {
+interface Range : AbstractRange {
   [Throws]
   readonly attribute Node startContainer;
```
RunDevelopment, line 18: (I'm not a reviewer and this isn't a review.) I've never worked with this before, so this might be a stupid question, but why is this attribute still defined here? 

The [IDL for `Range`](https://dom.spec.whatwg.org/#range) seems to rely on inheritance to get this attribute, so is the IDL parser wasm bindgen uses not smart enough to get inheritance or could this attribute (and the others defined by `AbstractRange`) be removed from this file?

behrenle: Thanks for the question! Here’s a bit more context:
- I added the inheritance from `AbstractRange` to `Range` in the IDL but left the existing attributes in place.
- I didn’t modify any bindings or attributes in `Range` itself because I wasn’t sure if removing the attributes would lead to breaking changes, especially with our current wasm-bindgen setup.
- The IDL parser that wasm-bindgen uses might not fully handle inheritance in this case, so removing the attribute definitions could potentially break the interface. 
That’s why I opted to keep the attributes defined in `Range` for now while adding the inheritance from `AbstractRange` to ensure compatibility.
Hope that clarifies things!

crates/web-sys/webidls/enabled/AbstractRange.webidl
```diff
@@ -0,0 +1,23 @@
...
+
+interface AbstractRange {
+  [Throws]
+  readonly attribute Node startContainer;
+  [Throws]
+  readonly attribute unsigned long startOffset;
+  [Throws]
+  readonly attribute Node endContainer;
+  [Throws]
+  readonly attribute unsigned long endOffset;
```
daxpedda, line 21: I can't find anything in the specification saying that this could throw.
I assume you simply copied this from the Firefox webidl directory, but we tend to follow the specification as close as possible.

Let me know if you have different information.

crates/web-sys/webidls/enabled/AbstractRange.webidl
daxpedda, line 1: EOL missing.

crates/web-sys/webidls/enabled/Range.webidl
```diff
@@ -10,10 +10,10 @@
  *
  * Copyright © 2012 W3C® (MIT, ERCIM, Keio), All Rights Reserved. W3C
  * liability, trademark and document use rules apply.
- */
+ */
```
daxpedda, line 13:
```suggestion
 */
```

crates/web-sys/webidls/enabled/StaticRange.webidl
daxpedda, line 1: EOL missing.

daxpedda (changes requested):

behrenle:
@daxpedda I have incorporated your requested changes and would appreciate another review. Unfortunately, I can't remove the 'waiting for author' label myself.

kartonrad:
Ah, this PR is just what i need, thank you @behrenle

kartonrad:
Oh, unfortunately i can't use this as a git dependency due to the "links" tag - 
if anything depends on wasm-bindgen, and i import this version of web_sys, dependency resolution fails.

It's not impossible to get it to work temporarily, but i would definetely appreciate if this were merged sometime in the coming months

guybedford (approved):
I've rebased this to the latest main branch, would be great to land along with the CSS Highlight API WebIDL updates for the next release.

daxpedda (approved):
I apologize for never getting back to this!
I'm glad @guybedford is picking up the slack here.

kartonrad:
Thank you <3 @guybedford @daxpedda

PR #4222, opened 2024-10-24
Add support for i128 and u128

### LINKED ISSUE #2822: Implement WASM ABI conversion for `u128` and `i128`

RunDevelopment:
Fixes #2822

This adds WASM ABI support for `i128` and `u128`. Both are split into 2 `u64` and passed as such in the ABI. 

The following types can now be used in argument and return positions: `i128`, `u128`, `Option<i128>`, and `Option<u128>`. Return positions can also use `Result<i128, E>` and `Result<u128, E>`.

It's important to note that `Result<Option<i128>, E>` is **not** supported. This is because the `WasmAbi` trait supports at most 4 primitives, but `Result<Option<i128>, E>` would require 5 primitives (2 for `i128`, 1 for the option tag, 1 for the result tag, and 1 for the error variant value). Future `Result<Option<T>, E>` ABI optimizations (either combining the result and option tag into one primitive, or raising the number of supported primitives to 5) will make `Result<Option<i128>, E>` possible, but this should be a separate PR IMO.

---

Related to #4201.


crates/cli-support/src/js/binding.rs
```diff
@@ -982,11 +1029,7 @@ fn instruction(
             js.push(format!("!isLikeNone({0})", val));
             js.push(format!(
                 "isLikeNone({val}) ? {zero} : {val}",
-                zero = if *ty == ValType::I64 {
-                    "BigInt(0)"
-                } else {
-                    "0"
-                }
+                zero = if *ty == ValType::I64 { "0n" } else { "0" }
```
MattiasBuelens, line 1032: This might regress on #3164. We may want to avoid bigint literals everywhere else too.

RunDevelopment: Good catch. Thank you! This is definitely a blocker.

After looking at #3164 and the error it intended to fix, I now believe that #3164 shouldn't have been accepted. I'll open an issue about where I'll explain my reasoning in detail.

daxpedda:
Setting on draft until #4246 is resolved.

RunDevelopment:
> Converting to draft until #4246 is resolved.

Either that, or I just use `BigInt(value)` everywhere for now. If we decide to bigint literals in #4246 later, I can just make another PR to change it later. Grepping for `BigInt(` isn't hard.

daxpedda:
> > Converting to draft until #4246 is resolved.
> 
> Either that, or I just use `BigInt(value)` everywhere for now. If we decide to bigint literals in #4246 later, I can just make another PR to change it later. Grepping for `BigInt(` isn't hard.

Yep, please go ahead and just use the constructor!

crates/cli-support/src/js/binding.rs
RunDevelopment: I'll just use `BigInt` constructors for now.

tests/wasm/optional_primitives.js
```diff
@@ -102,6 +104,22 @@ exports.js_works = () => {
...
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_min()), -170141183460469231731687303715884105728n);
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_max()), 170141183460469231731687303715884105727n);
+    assert.strictEqual(wasm.optional_i128_identity(0x3_1415_9265_3598_7932_3846n), 0x3_1415_9265_3598_7932_3846n);
+    assert.strictEqual(wasm.optional_i128_identity(-0x3_1415_9265_3598_7932_3846n), -0x3_1415_9265_3598_7932_3846n);
+
+    assert.strictEqual(wasm.optional_u128_identity(wasm.optional_u128_none()), undefined);
+    assert.strictEqual(wasm.optional_u128_identity(wasm.optional_u128_zero()), BigInt('0'));
+    assert.strictEqual(wasm.optional_u128_identity(wasm.optional_u128_one()), BigInt('1'));
+    assert.strictEqual(wasm.optional_u128_identity(wasm.optional_u128_min()), BigInt('0'));
+    assert.strictEqual(wasm.optional_u128_identity(wasm.optional_u128_max()), 340282366920938463463374607431768211455n);
```
daxpedda, line 120:
```suggestion
    assert.strictEqual(wasm.optional_u128_identity(wasm.optional_u128_max()), 340282366920938463463374607431768211455n);
```
Can't we just use hexadecimal representation here and in other similar ones as well?

tests/wasm/optional_primitives.js
```diff
@@ -102,6 +104,22 @@ exports.js_works = () => {
     assert.strictEqual(wasm.optional_u64_identity(wasm.optional_u64_min()), BigInt('0'));
     assert.strictEqual(wasm.optional_u64_identity(wasm.optional_u64_max()), BigInt('18446744073709551615'));
 
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_none()), undefined);
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_zero()), BigInt('0'));
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_one()), BigInt('1'));
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_neg_one()), BigInt('-1'));
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_min()), -170141183460469231731687303715884105728n);
+    assert.strictEqual(wasm.optional_i128_identity(wasm.optional_i128_max()), 170141183460469231731687303715884105727n);
+    assert.strictEqual(wasm.optional_i128_identity(0x3_1415_9265_3598_7932_3846n), 0x3_1415_9265_3598_7932_3846n);
```
daxpedda, line 113: Whats that number? Just random?

daxpedda (changes requested):
Great, thank you!

Just some nits and a changelog entry missing.

tests/wasm/optional_primitives.js
RunDevelopment: It's pi. I just wanted to test a large random number.

tests/wasm/optional_primitives.js
RunDevelopment: Good point.Hex would be a bit easier to understand here.

tests/wasm/optional_primitives.js
daxpedda: Oh, right!

daxpedda (approved):
Thanks!

PR #4224, opened 2024-10-25
Fix generic type 'Array<T>' requires 1 type argument

nikgraf:


RunDevelopment:
Duplicate of #4210. Related to #4207.

Unfortunately, it seems that the maintainers of this repo are quite busy this week, so it might take a little while until this issue is fixed.

nikgraf:
thanks @RunDevelopment - then I close mine

PR #4225, opened 2024-10-25
Fixed imports with late binding and `this`

### LINKED ISSUE #4214: 0.2.95 breaks imports that rely on context when using `js_namespace`
### Describe the Bug

I have the following imports in my Rust code:
```rust
#[wasm_bindgen(js_namespace = document, js_name = createTextNode)]
pub(crate) fn text_node(t: &str) -> Node;

#[wasm_bindgen(js_namespace = document, js_name = createTextNode)]
pub(crate) fn text_node_num(t: f64) -> Node;
```

These follow the [examples given in the docs](https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-js-imports/js_namespace.html) so I don't think I'm doing anything weird.

The codegen for 0.2.93 looks as follows:

```js
imports.wbg.__wbg_createTextNode_9e1923f20a1a5e7c = function(arg0, arg1) {
    const ret = document.createTextNode(getStringFromWasm0(arg0, arg1));
    return addHeapObject(ret);
};
imports.wbg.__wbg_createTextNode_91b927b5f0a34bb9 = function(arg0) {
    const ret = document.createTextNode(arg0);
    return addHeapObject(ret);
};
```

This works as intended.

Since 0.2.94 got yanked I can't check that, however in 0.2.95 the codegen has changed:

```js
imports.wbg.__wbg_createTextNode_9e1923f20a1a5e7c = function(arg0, arg1) {
    const ret = document.createTextNode(getStringFromWasm0(arg0, arg1));
    return ret;
};
imports.wbg.__wbg_createTextNode_91b927b5f0a34bb9 = typeof document.createTextNode == 'function' ? document.createTextNode : notDefined('document.createTextNode');
```

Ignoring the change to reference passing (not sure why that happens yet but it's not the issue) the new codegen grabs a reference to the JS function directly without its context, which leads to errors when the function is a method and requires its context on call site.

You can easily reproduce the same error in the browser console:
image

### Steps to Reproduce

Use this example from the docs:

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = ["window", "document"])]
    fn write(s: &str);
}

write("hello, document!");
```

Change the parameter from `&str` to a primitive such as `u32`.

### Expected Behavior

The codegen should maintain the callsite as it did in 0.2.93, or it should use the `.bind` method:

```js
imports.wbg.__wbg_createTextNode_91b927b5f0a34bb9 = typeof document.createTextNode == 'function' ? document.createTextNode.bind(document) : notDefined('document.createTextNode');
```

### Additional Context

I'm filing this as a bug since upgrading the version did break my code, _however_ I'm not sure if the thing I'm doing (and the docs in question) really should be the correct way to do this. Having to provide my own little snippets for function calls like that is annoying, but in the long run probably conveys the intent clearer.

I reckon alternative to reverting any changes here would be to change the docs stating that `js_namespace` shouldn't be used on built-in objects with methods. Sadly JS is extremely inconsistent about this, while you can't grab methods from `document` without the context, `console` methods (such as `console.log`) generally don't care about context at all.

RunDevelopment:
Fixes #4214

The problem in #4214 was that function like `document.write` can't be imported as a function reference into WASM directly, because `document.write` does not capture the `this` argument (here `document`). Or in other words, there is a difference between:

```js
export function __wbg_reload_90d82b22b83c1d99(text) {
    document.write(text); // `write` has `document` as the `this` arg
};
```

and 

```js
// there is no `this` arg, so `write` will fail when invoked in WASM
export const __wbg_reload_90d82b22b83c1d99 = typeof document.write === "function" ? document.write(text) : notDefined('document.write');
```

However, aside from `this` not being captured, the (let's called it) direct import style also didn't support late binding. If the object or function is only defined *after* the WASM module is instantiated, then the direct import style will cause an error.

This PR fixes both issues by always generating a wrapper function for imported function from global scope.


daxpedda (changes requested):
Nice!

Missing a changelog entry!

RunDevelopment:
Added changelog.

daxpedda (approved):

PR #4226, opened 2024-10-25
Fix docs on `Option<char>`

RunDevelopment:
The docs said that `Option<char>` is not support in argument or return position, but it very much is.


daxpedda (approved):
Thanks!

PR #4230, opened 2024-10-28
Add support for symbols in `js_name` and `getter`/`setter`

### LINKED ISSUE #4149: Support symbols for fields and functions of exported structs
### Motivation

Certain JS protocols require symbols to identify a certain method or field. E.g. for #4142, we need to use [`Symbol.iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator) to define an iterable object.

### Proposed Solution

Support the syntax `#[wasm_bindgen(js_name = Symbol.name)]` to define use symbols instead of regular (string) identifiers for struct fields and functions. A different syntax to identify symbols may also be used, e.g. `@@name`.

Whatever syntax is used, I think the syntax should be made exclusive. So current code using `js_name = Symbol.name` (or whatever syntax is chosen) should fail to compile for non-class member bindings. Example:

```rs
#[wasm_bindgen(js_name = Symbol.iterator)] // error
fn foo() { }

#[wasm_bindgen]
struct Foo;

#[wasm_bindgen]
impl Foo {
    #[wasm_bindgen(js_name = Symbol.iterator)] // ok
    pub fn iterator(&self) -> js_sys::Iterator { todo!() }
}
```

### Alternatives

1. Don't support symbols via `#[wasm_bindgen]` and rely on reflection. This is quite painful.
2. Add a new option, e.g. `js_symbol`, separate from `js_name` for symbols. I don't think this is a good idea. A symbol is intended to be a unique name, so using `js_name` makes sense.

### Additional Context

none

RunDevelopment:
Fixes #4149.

This PR adds basic support for `js_name = "[Symbol.<name>]"`. It generally works, but still needs to be polished.

TODO list:

- [x] Resolve conflicts. This PR conflicts with #4225, #4202, and #4188. Those PR should be dealt with before this one.
- [x] Fix the bug that `getter = name` doesn't rename imported static getters. See `import_static_symbol_getter` in `crates/cli/tests/reference/symbol.rs`. This might be fixed by #4225.
- [x] Add a symbol name filter. Right now, any name is allowed, but we should only allow known symbol names. This should also blacklist `Symbol.dispose` as [pointed out here](https://github.com/rustwasm/wasm-bindgen/issues/4149#issuecomment-2400851626).
- [x] More testing, especially for the errors users get when using `js_name = Symbol.name` in invalid positions. Users should ideally get an error from the `#[wasm_bindgen]` macro, so users get errors in the IDEs.

Leaving for follow-ons:
- Automatically polyfill missing symbols. Just like #4118, we should polyfill symbols that aren't defined by the JS runtime. We could probably just generalize the `expose_symbol_dispose` method to do this.


sonthonaxrk:
This would be extremely useful for a lot of things, I hope the maintainers take another look at this soon

guybedford (commented):
I like the approach here a lot. Would be happy to review and land further.

In theory the same could apply for private fields or other special cases of compute properties. And if we hit naming conflicts, we could rely on `js_name = "Symbol.iterator"` to disambiguate.

* Do we properly error if attempting to define the same name twice? 
* Can we explicitly ban `Symbol.dispose` being defined on classes, as we define it separately?
* Can we include a test that `js_name = "Symbol.iterator"` can still be used to set the string `MyClass.prototype["Symbol.iterator"]` as another edge case? I believe js_name can be quoted?

Twey:
Rather than special-casing on `Symbol`, it would be nice if the syntax for this were `js_name = [Symbol.toPrimitive]` (or whatever colour of bikeshed — `js_raw_name = "Symbol.toPrimitive"` would also be fine) to allow arbitrary JavaScript expressions as keys (as it works in JavaScript).

Polyfilling could be done `js_name = [Symbol.toPrimitive || new Symbol("Symbol.toPrimitive")]`.

guybedford:
Rebased this branch onto current main and force-pushed (RunDevelopment/wasm-bindgen@symbols). Latest pushed commit: RunDevelopment/wasm-bindgen@72ae73d6e.

Changes from the original commit:

* Resolved conflicts in `crates/macro-support/src/{ast,parser}.rs` (the old `crates/backend/` is now merged into macro-support), `crates/shared/src/lib.rs`, `crates/cli-support/src/{js,wit}/mod.rs`.
* Integrated with HEAD's additions: `JsNamespace` struct, third `invalid_unused` arg in `attrgen!`, `RegularThis`, `qualified_name`, `private`, `reexport`, the new `FieldLocation`/`FieldInfo` typescript-fields refactor in js/mod.rs, etc.
* Made `BindgenAttrs::js_name()` return `Option<(&JsName, Span)>` and added a small helper macro arm to keep the existing macro structure.
* Reused `property_definition`/`property_accessor`/`property_key` for symbol-aware codegen across class methods, getters, setters, struct fields, and TS field types.
* Properly disambiguate `Symbol.<name>` in optional-chained field access (`?.x` vs `?.[Symbol.x]`).
* Bumped `APPROVED_SCHEMA_FILE_HASH` (no `SCHEMA_VERSION` bump - that's reserved for releases).
* Re-blessed the reference test outputs to match current main's codegen (now generates `.bg.js` for bundler target, etc).
* Added guide coverage for symbol names in `js_name`, `getter`, and `setter` on both imports and Rust exports.
* Added macro UI coverage for invalid symbol positions and invalid computed-key forms.

Status of the prior TODO list:

* Conflicts: resolved and rebased.
* `getter = name` / static getter behavior: existing `import-getter-setter` and `symbol` reference coverage now passes on this branch, including `js_namespace + getter` cases.
* Symbol name filtering: intentionally not added; this keeps accepting any `Symbol.<ident>` rather than maintaining a whitelist.
* `Symbol.dispose`: intentionally not rejected; this keeps the behavior permissive.
* Error coverage: added `crates/macro/ui-tests/symbol-invalid-positions.rs` to pin macro-time diagnostics for invalid positions.

Test status after the latest force-push:

* `cargo fmt --check`
* `cargo test -p wasm-bindgen-macro --test ui`
* `cargo test -p wasm-bindgen-cli --test wasm-bindgen reference`

Generated output for the test case looks like:

```ts
export class Foo {
    private constructor();
    free(): void;
    [Symbol.dispose](): void;
    [Symbol.toPrimitive](): string;
    readonly [Symbol.toStringTag]: string;
}
```

```js
[Symbol.toPrimitive]() { ... }
get [Symbol.toStringTag]() { ... }
// imports:
arg0[Symbol.iterator]()
SomeClass[Symbol.toPrimitive](...)
SomeClass[Symbol.iterator]
```

At this point I believe the review TODO list is handled.

guybedford (approved):

crates/cli-support/src/js/mod.rs
```diff
@@ -5917,27 +5934,79 @@ fn define_export(
...
+                name.to_string()
+            } else {
+                format!("[\"{}\"]", name.escape_default())
+            }
+        }
+        AuxName::Symbol(name) => format!("[Symbol.{name}]"),
+    }
+}
+
+/// Similar the runtime string or symbol value of a property of the given name.
```
```suggestion
/// The runtime string or symbol value of a property of the given name.
```

crates/shared/src/lib.rs
```diff
@@ -242,33 +248,70 @@ pub fn unwrap_function(struct_name: &str) -> String {
     name
 }
 
-pub fn free_function_export_name(function_name: &str) -> String {
-    function_name.to_string()
+#[derive(Debug, Hash)]
+pub enum NameRef<'a> {
+    Identifier(&'a str),
+    Symbol(&'a str),
+}
```
hoodmane: Why do we need `NameRef`? Can't we just use `&Name` everywhere?

guybedford: Refactored to use a pure string convention instead.

crates/macro-support/src/ast.rs
```diff
@@ -572,7 +572,15 @@ impl Export {
             generated_name.push_str(class);
         }
         generated_name.push('_');
-        generated_name.push_str(&self.function.name.to_string());
+        // The JS-side name may contain characters that aren't valid in a
+        // Rust identifier (notably the `[Symbol.<name>]` computed-key form
+        // accepted by `js_name`). Filter to a valid identifier suffix; this
+        // is a no-op for plain identifier names.
```
hoodmane, line 578: Doesn't this just delete characters that aren't alphanumeric or underscore? I don't know what is a suffix here.

hoodmane (approved):

PR #4236, opened 2024-10-30
Add bindings for `MediaStreamTrack.getCapabilities`

DanielleHuisman:
Add bindings for [`MediaStreamTrack.getCapabilities`](https://www.w3.org/TR/mediacapture-streams/#dom-mediastreamtrack-getcapabilities) and stabilize [`MediaTrackCapabilities`](https://www.w3.org/TR/mediacapture-streams/#media-track-capabilities) (which it returns).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
 * Added support for C-style enums with negative discriminants.
   [#4204](https://github.com/rustwasm/wasm-bindgen/pull/4204)
 
+* Added bindings for `MediaStreamTrack.getCapabilities`.
+  [#4236](https://github.com/rustwasm/wasm-bindgen/pull/4236)
+
 ### Changed
 
 * String enums now generate private TypeScript types but only if used.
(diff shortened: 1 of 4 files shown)


daxpedda (changes requested):
We don't stabilize proposals that aren't in the standard class.
But the addition of `MediaStreamTrack.getCapabilities()` is nice, but has to be added unstably.

Let me know if you have any trouble doing that or if you have any questions!

daxpedda:
> Just curious: what makes `MediaStreamTrack` stable, but not `MediaTrackCapabilities`?

`MediaStreamTrack` should never have been made stable in the first place. Hindsight is 20/20 ofc.
> What specification or part of the specification indicates this?

Basically we don't stabilize anything unless it has reached the last stage: "Recommended".

There is some nuance to this, where we can stabilize earlier if we are fairly confident that no API changes are coming. This mostly depends on evaluating the proposals issue tracker and if its so widely used API changes are very unlikely as they would break the Internet.

daxpedda (approved):
Thanks!

PR #4245, opened 2024-11-02
Added more documentation for `bool`, `char`, and numeric types

RunDevelopment:
A small PR where I added documentation for `bool`, `char`, and the numeric types to explain how they are represented in JS and how the conversion between JS values and Rust values works. I also added some links and formatting to a few other pages.

My goal was to answer all the questions I had about number conversions when I first used WBG. E.g. when I first used WBG, I thought that `u64` would map to a JS `number` as well, since that's what the docs said: "JavaScript representation: A JavaScript number value". So I used `js_sys::BigInt` for the `u64` in my project since I needed the full 64 bits and added some custom parsing logic to convert `js_sys::BigInt` -> `u64`. Clunky and unnecessary.

---

Relates to #4222 and #4201, since both PRs need to update the docs from this PR.

This PR is technically independent of #4226, but I can combine the two if you like.

### DIFF
--- a/guide/src/reference/types/boxed-number-slices.md
+++ b/guide/src/reference/types/boxed-number-slices.md
@@ -4,7 +4,7 @@
 |:---:|:---:|:---:|:---:|:---:|:---:|:---:|
 | Yes | No | No | Yes | Yes | Yes | A JavaScript `TypedArray` of the appropriate type (`Int32Array`, `Uint8Array`, etc...) |
 
-Note that the contents of the slice are copied into the JavaScript `TypedArray`
+Note that the contents of the slice are copied into the JavaScript [`TypedArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
 from the Wasm linear memory when returning a boxed slice to JavaScript, and vice
 versa when receiving a JavaScript `TypedArray` as a boxed slice in Rust.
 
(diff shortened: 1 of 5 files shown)


guide/src/reference/types/boxed-number-slices.md
```diff
@@ -4,7 +4,7 @@
 |:---:|:---:|:---:|:---:|:---:|:---:|:---:|
 | Yes | No | No | Yes | Yes | Yes | A JavaScript `TypedArray` of the appropriate type (`Int32Array`, `Uint8Array`, etc...) |
 
-Note that the contents of the slice are copied into the JavaScript `TypedArray`
+Note that the contents of the slice are copied into the JavaScript [`TypedArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
```
daxpedda, line 7:
```suggestion
Note that the contents of the slice are copied into a JavaScript [`TypedArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray)
```
Note: not an issue introduced by this PR.

daxpedda, line 7: Lets do the whole `> **Note:**` for consistency here as well?
Note: again, the inconsistency was already here before this PR.

daxpedda (approved):
Amazing ... thank you!

PR #4249, opened 2024-11-05
Make it clearer which versions are affected by the constructor bug

RunDevelopment:
The form of the section informing users about an important bug was weird. It started with "versions `>=X` are affected", explained the bug and how it affects users, and then told users "actually, only versions `>=X, <Y` are affected" (literally: "This is no longer required as of v0.2.88"). This is bad form, because we are making users read through and understand a bug + workarounds only tell them at the very end that this isn't relevant to them.

This PR mentions the full affected version range at the start of the section.

### DIFF
--- a/guide/src/reference/attributes/on-rust-exports/constructor.md
+++ b/guide/src/reference/attributes/on-rust-exports/constructor.md
@@ -35,7 +35,7 @@ console.log(f.get_contents());
 
 ## Caveats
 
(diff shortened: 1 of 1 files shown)


daxpedda (approved):
Thank you!

I was considering removing the bottom notice, seeing that we just moved it up. But it doesn't really hurt and hopefully prevents users from overlooking that this isn't required anymore.

PR #4253, opened 2024-11-07
Revert bundler target specific default transformations

daxpedda:
`webpack` now supports the reference type proposal, so there is no need to not enable reference type transformations when enabled via the `target_features` section.

See #4235.


daxpedda:
So I did think that this is resolved now in `webpack`, but I have no clue how to make `wasm-pack` make use of that fix.
See https://github.com/webpack/webpack/issues/15566.

daxpedda:
Fixed in https://github.com/webpack/webpack/pull/18979. Going to wait for it to merge and release.

I also went ahead and removed the `wasm2js` example, as it doesn't support reference-types and there are no plans to add support.

daxpedda:
I decided not to wait for any longer for a `webpack` release to cut a new release for `wasm-bindgen`.

denis-sh:
> I decided not to wait for any longer for a `webpack` release to cut a new release for `wasm-bindgen`.

[webpack v5.97.0](https://github.com/webpack/webpack/releases/tag/v5.97.0) (released 2024-12-03, 1.5 months after [Rust 1.82.0 release](https://releases.rs/docs/1.82.0/)) finally added support for WebAssembly Reference types.

(added for historical record)

PR #4255, opened 2024-11-08
Faster local testing with `reference.rs`

RunDevelopment:
This is basically a follow-up to #4242 to make it faster to run reference tests locally. The main problem is that tests take time to run and that `references.rs` had no progress indication and only reported failures after all tests went through.

This PR:

1. Add progress indication by simply printing the current test file to stdout. Super simple and helps a lot.
2. Stops after the first test failed, so you don't have to wait for the other tests to see what went wrong.
3. Runs tests in order of modification date. So test files that were recently modified get tested first. This makes it very fast to iterate on tests, since my usual workflow is: edit one test file, run all tests, see the results.

Change 2 and 3 only take effect locally. CI workflows use the old strategy of running all tests in order of name and printing all error messages (if any) at the end.

With this, writing and running tests is now quite convenient. Before #4242, I had to wait roughly 2 minutes to see the results of a test I added. After #4242, it was down to around 20\~30 seconds. With this PR, it's down to 2\~4 seconds.


crates/cli/tests/reference.rs
```diff
@@ -57,15 +62,33 @@ fn main() -> Result<()> {
...
+        runtest(t).err().map(|e| (t, e))
+    });
+
+    let first_error = errs_iter.next();
+    if first_error.is_none() {
         println!("{} tests passed", tests.len());
         return Ok(());
     }
+
+    let mut errs = vec![first_error.unwrap()];
```
daxpedda, line 85:
```suggestion
    let Some(first_error) = errs_iter.next() else {
        println!("{} tests passed", tests.len());
        return Ok(());
    };

    let mut errs = vec![first_error];
```

daxpedda (changes requested):
Fun! Thank you for doing this!

daxpedda (approved):

PR #4256, opened 2024-11-08
Update testing docs for `wasm-bindgen-macro`

RunDevelopment:
Changes:
1. The command in the docs was wrong. `cargo test -p ui-tests` doesn't work because there is `ui-tests` crate. The correct command is `cargo test -p wasm-bindgen-macro --test ui`.
2. The way for updating snaps was pretty much linux-only because it required running a bash script. I added a note explaining how you can do it on Windows with `TRYBUILD=overwrite`.
3. I removed the note linking to #601. That issue has long been closed and the note isn't relevant anymore.


daxpedda (approved):
Thank you!

PR #4258, opened 2024-11-09
Add support for static enum methods via TS namespaces

### LINKED ISSUE #1715: Enum impl fn gives name conflict error
Based on the comment at https://github.com/rustwasm/wasm-bindgen/issues/1496#issuecomment-519255857, this gives an error:

```rust
#[wasm_bindgen]
#[derive(Copy, Clone)]
pub enum ImageFormat {
    PNG,
    JPEG,
    GIF,
}

#[wasm_bindgen]
impl ImageFormat {
    #[wasm_bindgen]
    pub fn from_str(s: &str) -> Result<ImageFormat, JsValue> {
        match s {
            "PNG" => Ok(ImageFormat::PNG),
            "JPEG" => Ok(ImageFormat::JPEG),
            "GIF" => Ok(ImageFormat::GIF),
            _ => Err(JsValue::from(js_sys::Error::new(&format!("Unknown format: {}", s)))),
        }
    }
}
```

The error is:

    error: cannot shadow already defined class `ImageFormat`

The `wasm-bindgen` version in my lock file is `0.2.48`.

### LINKED ISSUE #2045: Imperfect enum of TypeScript.

RunDevelopment:
Fixes #1715
Resolves #2045

This PR adds support for static methods on enums (string and C-style). It uses the approach I outlined [here](https://github.com/rustwasm/wasm-bindgen/issues/1715#issuecomment-2439973635) using TS `namespace`s for typing.

Example:

```rs
#[wasm_bindgen]
pub enum ImageFormat { PNG, JPEG, GIF }

#[wasm_bindgen]
impl ImageFormat {
    pub fn from_str(s: &str) -> ImageFormat { todo!() }
}
```
Now generates the following JS and type definitions:

```js
/**
 * @enum {0 | 1 | 2}
 */
export const ImageFormat = {
    PNG: 0, "0": "PNG",
    JPEG: 1, "1": "JPEG",
    GIF: 2, "2": "GIF",
};

(function(ImageFormat) {
    /**
     * @param {string} s
     * @returns {ImageFormat}
     */
    ImageFormat.from_str = function from_str(s) { /* ... */ };
})(ImageFormat);
```
```ts
export enum ImageFormat {
  PNG = 0,
  JPEG = 1,
  GIF = 2,
}
export namespace ImageFormat {
  export function from_str(s: string): ImageFormat;
}
```

Note that C-style enums no longer use `Object.freeze`. This is required since the functionality of namespaces lies in adding properties to the namespaced object. If `Object.freeze` was kept, the namespace would result in a runtime error when the JS file is run. This resolves #2045.

Another thing to note is how string enums are handled. String enums don't have an exported object the namespace could latch onto, so an empty object is automatically exported.

---

Changes:
1. Exported methods with a JS class that points to an enum or string enum will now be exported via a namespace.
2. Error if users try to add instance methods/getters/setters or static getters/setters.
    
    This is technically a **breaking change**, because string enums technically supported all this before. Technically, because them supporting it was a bug. Both string enums and C-style enums had the bug where they tried to generate a class for methods and getters/setters. For C-style enums, with resulted in a runtime error in JS, because both the enum and class were exported under the same name. However, string enums don't export anything, so the class didn't cause any problems. While buggy (there were no type definitions for the generated class) and unintended, this somewhat worked. Users could successfully define and use static methods for string enums.

    Why is this a breaking change? The namespaces I implemented don't support static getters and setter. While it is trivial to implement static getter/setters in JS via [`Object.defineProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#creating_a_property), the problem is typing them in TS. TS only support getters/setters on classes. The best `namespace`s can do is to `export const/let` like [this](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAOwIYFtgGcxIMbDgMQgjgG8AoOSuUSWObCBdeAQTgF44BOAbgqprR4AG2DwAQhzgBGAAx8AvmTIB6FXDYBLdHCjAkAE0bCAngBo46TSk3CkUODGJI4AczExgUMoQgA6Nk4AFgAmPjU4SW04AHcoTU8kACNgYQsrGzsHJzgXdxhPKABadA8vOCxNb18-SWCwoA).

   So the question is: are we okay with breaking user code relying on a bug, or should I add support for static getters/setters on namespaces?

3. Explicitly type what type of expressions can be exported in `Context::export`. This is done via the new `ExportJs` enum, which supports classes, functions, expressions, and namespaces. Instead of making the `Context::export` detect what is being exported by analyzing the string content of the JS expression, the ones generating the JS expression are now required to follow a certain format. This makes the code for exporting a lot more explicit and the API a lot less stringly typed.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -281,6 +281,9 @@ impl<'a> Context<'a> {
                     // exported without an existing object of the same name.
                     // In that case, we need to create the object before
                     // initializing the namespace.
+                    //
+                    // It's only correct to do it like this, because namespaces
+                    // are always the last things to be exported.
                     let mut definition = String::new();
                     if !self.defined_identifiers.contains_key(export_name) {
                         definition = format!("export const {} = {{}};\n", export_name)
@@ -4774,15 +4777,12 @@ impl ExportedNamespace {
         ts: Option<&str>,
     ) {
         self.contents.push_str(js_docs);
-        self.contents.push_str("function ");
-        self.contents.push_str(function_name);
-        self.contents.push_str(js);
-        self.contents.push('\n');
         self.contents.push_str(&self.name);
         self.contents.push('.');
         self.contents.push_str(function_name);
-        self.contents.push_str(" = ");
+        self.contents.push_str(" = function ");
         self.contents.push_str(function_name);
+        self.contents.push_str(js);
         self.contents.push_str(";\n");
 
         if let Some(ts) = ts {
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -61,6 +61,8 @@ pub struct Context<'a> {
 
     exported_classes: Option<BTreeMap<String, ExportedClass>>,
 
+    exported_namespaces: Option<BTreeMap<String, ExportedNamespace>>,
+
     /// A map of the name of npm dependencies we've loaded so far to the path
     /// they're defined in as well as their version specification.
     pub npm_dependencies: HashMap<String, (PathBuf, String)>,
@@ -120,6 +122,35 @@ struct FieldAccessor {
     is_optional: bool,
 }
 
+struct ExportedNamespace {
+    name: String,
+    contents: String,
+    /// The TypeScript for the namespace's methods.
+    typescript: String,
+    /// Whether TypeScript for this namespace should be emitted (i.e., `skip_typescript` wasn't specified).
+    generate_typescript: bool,
+}
+
+enum ClassOrNamespace<'a> {
+    Class(&'a mut ExportedClass),
+    Namespace(&'a mut ExportedNamespace),
+}
+
+/// Different JS constructs that can be exported.
+enum ExportJs<'a> {
+    /// A class of the form `class Name {...}`.
+    Class(&'a str),
+    /// An anonymous function expression of the form `function(...) {...}`.
+    ///
+    /// Note that the function name is not included in the string.
+    Function(&'a str),
+    /// An arbitrary JS expression.
+    Expression(&'a str),
+    /// A namespace as a function expression for initiating the namespace. The
+    /// function expression is of the form `(function(Name) {...})`.
+    Namespace(&'a str),
+}
+
 const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
 // Must be kept in sync with `src/lib.rs` of the `wasm-bindgen` crate
 const INITIAL_HEAP_OFFSET: usize = 128;
@@ -143,6 +174,7 @@ impl<'a> Context<'a> {
             typescript_refs: Default::default(),
             used_string_enums: Default::default(),
             exported_classes: Some(Default::default()),
+            exported_namespaces: Some(Default::default()),
             config,
             threads_enabled: config.threads.is_enabled(module),
             module,
@@ -163,38 +195,72 @@ impl<'a> Context<'a> {
     fn export(
         &mut self,
         export_name: &str,
-        contents: &str,
+        export: ExportJs,
         comments: Option<&str>,
     ) -> Result<(), Error> {
-        let definition_name = self.generate_identifier(export_name);
-        if contents.starts_with("class") && definition_name != export_name {
+        // The definition is intended to allow for exports to be renamed to
+        // avoid conflicts. Since namespaces intentionally have the same name as
+        // other exports, we must not rename them.
+        let definition_name = if matches!(export, ExportJs::Namespace(_)) {
+            export_name.to_owned()
+        } else {
+            self.generate_identifier(export_name)
+        };
+
+        if matches!(export, ExportJs::Class(_)) && definition_name != export_name {
             bail!("cannot shadow already defined class `{}`", export_name);
         }
 
-        let contents = contents.trim();
+        // write out comments
         if let Some(c) = comments {
             self.globals.push_str(c);
         }
+
+        fn namespace_init_arg(name: &str) -> String {
+            format!("{name} || ({name} = {{}})", name = name)
+        }
+
         let global = match self.config.mode {
-            OutputMode::Node { module: false } => {
-                if contents.starts_with("class") {
-                    format!("{}\nmodule.exports.{1} = {1};\n", contents, export_name)
-                } else {
-                    format!("module.exports.{} = {};\n", export_name, contents)
+            OutputMode::Node { module: false } => match export {
+                ExportJs::Class(class) => {
+                    format!("{}\nmodule.exports.{1} = {1};\n", class, export_name)
                 }
-            }
-            OutputMode::NoModules { .. } => {
-                if contents.starts_with("class") {
-                    format!("{}\n__exports.{1} = {1};\n", contents, export_name)
-                } else {
-                    format!("__exports.{} = {};\n", export_name, contents)
+                ExportJs::Function(expr) | ExportJs::Expression(expr) => {
+                    format!("module.exports.{} = {};\n", export_name, expr)
                 }
-            }
+                ExportJs::Namespace(namespace) => {
+                    format!(
+                        "{}({});\n",
+                        namespace,
+                        namespace_init_arg(&format!("module.exports.{}", export_name))
+                    )
+                }
+            },
+            OutputMode::NoModules { .. } => match export {
+                ExportJs::Class(class) => {
+                    format!("{}\n__exports.{1} = {1};\n", class, export_name)
+                }
+                ExportJs::Function(expr) | ExportJs::Expression(expr) => {
+                    format!("__exports.{} = {};\n", export_name, expr)
+                }
+                ExportJs::Namespace(namespace) => {
+                    format!(
+                        "{}({});\n",
+                        namespace,
+                        namespace_init_arg(&format!("__exports.{}", export_name))
+                    )
+                }
+            },
             OutputMode::Bundler { .. }
             | OutputMode::Node { module: true }
             | OutputMode::Web
-            | OutputMode::Deno => {
-                if let Some(body) = contents.strip_prefix("function") {
+            | OutputMode::Deno => match export {
+                ExportJs::Class(class) => {
+                    assert_eq!(export_name, definition_name);
+                    format!("export {}\n", class)
+                }
+                ExportJs::Function(function) => {
+                    let body = function.strip_prefix("function").unwrap();
                     if export_name == definition_name {
                         format!("export function {}{}\n", export_name, body)
                     } else {
@@ -203,14 +269,25 @@ impl<'a> Context<'a> {
                             definition_name, body, definition_name, export_name,
                         )
                     }
-                } else if contents.starts_with("class") {
-                    assert_eq!(export_name, definition_name);
-                    format!("export {}\n", contents)
-                } else {
-                    assert_eq!(export_name, definition_name);
-                    format!("export const {} = {};\n", export_name, contents)
                 }
-            }
+                ExportJs::Expression(expr) => {
+                    assert_eq!(export_name, definition_name);
+                    format!("export const {} = {};\n", export_name, expr)
+                }
+                ExportJs::Namespace(namespace) => {
+                    assert_eq!(export_name, definition_name);
+
+                    // In some cases (e.g. string enums), a namespace may be
+                    // exported without an existing object of the same name.
+                    // In that case, we need to create the object before
+                    // initializing the namespace.
+                    let mut definition = String::new();
+                    if !self.defined_identifiers.contains_key(export_name) {
+                        definition = format!("export const {} = {{}};\n", export_name)
+                    }
+                    format!("{}{}({});\n", definition, namespace, export_name)
+                }
+            },
         };
         self.global(&global);
         Ok(())
@@ -224,6 +301,9 @@ impl<'a> Context<'a> {
         // glue for all classes as well as finish up a few final imports like
         // `__wrap` and such.
         self.write_classes()?;
+
+        // Write out generated JS for namespaces.
+        self.write_namespaces()?;
 
         // Initialization is just flat out tricky and not something we
         // understand super well. To try to handle various issues that have come
@@ -1005,6 +1085,22 @@ __wbg_set_wasm(wasm);"
         Ok((js, ts))
     }
 
+    fn require_class_or_namespace(&mut self, name: &str) -> ClassOrNamespace {
+        if self.aux.enums.contains_key(name) || self.aux.string_enums.contains_key(name) {
+            ClassOrNamespace::Namespace(self.require_namespace(name))
+        } else {
+            ClassOrNamespace::Class(self.require_class(name))
+        }
+    }
+
+    fn require_class(&mut self, name: &str) -> &'_ mut ExportedClass {
+        self.exported_classes
+            .as_mut()
+            .expect("classes already written")
+            .entry(name.to_string())
+            .or_default()
+    }
+
     fn write_classes(&mut self) -> Result<(), Error> {
         for (class, exports) in self.exported_classes.take().unwrap() {
             self.write_class(&class, &exports)?;
@@ -1152,10 +1248,10 @@ __wbg_set_wasm(wasm);"
 
         self.write_class_field_types(class, &mut ts_dst);
 
-        dst.push_str("}\n");
+        dst.push('}');
         ts_dst.push_str("}\n");
 
-        self.export(name, &dst, Some(&class.comments))?;
+        self.export(name, ExportJs::Class(&dst), Some(&class.comments))?;
 
         if class.generate_typescript {
             self.typescript.push_str(&class.comments);
@@ -1281,6 +1377,57 @@ __wbg_set_wasm(wasm);"
                 }
             };
         }
+    }
+
+    fn require_namespace(&mut self, name: &str) -> &'_ mut ExportedNamespace {
+        self.exported_namespaces
+            .as_mut()
+            .expect("namespaces already written")
+            .entry(name.to_string())
+            .or_insert_with(|| {
+                let _enum = self.aux.enums.get(name);
+                let string_enum = self.aux.string_enums.get(name);
+
+                let generate_typescript = _enum.map_or(true, |e| e.generate_typescript)
+                    && string_enum.map_or(true, |e| e.generate_typescript);
+
+                ExportedNamespace {
+                    name: name.to_string(),
+                    contents: String::new(),
+                    typescript: String::new(),
+                    generate_typescript,
+                }
+            })
+    }
+
+    fn write_namespaces(&mut self) -> Result<(), Error> {
+        for (class, namespace) in self.exported_namespaces.take().unwrap() {
+            self.write_namespace(&class, &namespace)?;
+        }
+        Ok(())
+    }
+
+    fn write_namespace(&mut self, name: &str, namespace: &ExportedNamespace) -> Result<(), Error> {
+        if namespace.contents.is_empty() {
+            // don't emit empty namespaces
+            return Ok(());
+        }
+
+        let dst = format!(
+            "(function({name}) {{\n{contents}}})",
+            contents = namespace.contents
+        );
+        self.export(name, ExportJs::Namespace(&dst), None)?;
+
+        if namespace.generate_typescript {
+            let ts_dst = format!(
+                "export namespace {name} {{\n{ts}}}\n",
+                ts = namespace.typescript
+            );
+            self.typescript.push_str(&ts_dst);
+        }
+
+        Ok(())
     }
 
     fn expose_drop_ref(&mut self) {
@@ -2461,11 +2608,11 @@ __wbg_set_wasm(wasm);"
     }
 
     fn require_class_wrap(&mut self, name: &str) {
-        require_class(&mut self.exported_classes, name).wrap_needed = true;
+        self.require_class(name).wrap_needed = true;
     }
 
     fn require_class_unwrap(&mut self, name: &str) {
-        require_class(&mut self.exported_classes, name).unwrap_needed = true;
+        self.require_class(name).unwrap_needed = true;
     }
 
     fn add_module_import(&mut self, module: String, name: &str, actual: &str) {
@@ -2831,11 +2978,15 @@ __wbg_set_wasm(wasm);"
                             self.typescript.push_str(";\n");
                         }
 
-                        self.export(name, &format!("function{}", code), Some(&js_docs))?;
+                        self.export(
+                            name,
+                            ExportJs::Function(&format!("function{}", code)),
+                            Some(&js_docs),
+                        )?;
                         self.globals.push('\n');
                     }
                     AuxExportKind::Constructor(class) => {
-                        let exported = require_class(&mut self.exported_classes, class);
+                        let exported = self.require_class(class);
 
                         if exported.has_constructor {
                             bail!("found duplicate constructor for class `{}`", class);
@@ -2850,7 +3001,7 @@ __wbg_set_wasm(wasm);"
                         receiver,
                         kind,
                     } => {
-                        let exported = require_class(&mut self.exported_classes, class);
+                        let mut exported = self.require_class_or_namespace(class);
 
                         let mut prefix = String::new();
                         if receiver.is_static() {
@@ -2859,6 +3010,17 @@ __wbg_set_wasm(wasm);"
                         let ts = match kind {
                             AuxExportedMethodKind::Method => ts_sig,
                             AuxExportedMethodKind::Getter => {
+                                let class = match exported {
+                                    ClassOrNamespace::Class(ref mut class) => class,
+                                    ClassOrNamespace::Namespace(_) => {
+                                        bail!(
+                                            "the getter `{}` is not supported on `{}`. Enums only support static methods on them.",
+                                            name,
+                                            class
+                                        );
+                                    }
+                                };
+
                                 prefix += "get ";
                                 // For getters and setters, we generate a separate TypeScript definition.
                                 if export.generate_typescript {
@@ -2873,14 +3035,25 @@ __wbg_set_wasm(wasm);"
                                         is_optional: false,
                                     };
 
-                                    exported.push_accessor_ts(location, accessor, false);
+                                    class.push_accessor_ts(location, accessor, false);
                                 }
                                 // Add the getter to the list of readable fields (used to generate `toJSON`)
-                                exported.readable_properties.push(name.clone());
+                                class.readable_properties.push(name.clone());
                                 // Ignore the raw signature.
                                 None
                             }
                             AuxExportedMethodKind::Setter => {
+                                let class = match exported {
+                                    ClassOrNamespace::Class(ref mut class) => class,
+                                    ClassOrNamespace::Namespace(_) => {
+                                        bail!(
+                                            "the setter `{}` is not supported on `{}`. Enums only support static methods on them.",
+                                            name,
+                                            class
+                                        );
+                                    }
+                                };
+
                                 prefix += "set ";
                                 if export.generate_typescript {
                                     let location = FieldLocation {
@@ -2893,13 +3066,27 @@ __wbg_set_wasm(wasm);"
                                         is_optional: might_be_optional_field,
                                     };
 
-                                    exported.push_accessor_ts(location, accessor, true);
+                                    class.push_accessor_ts(location, accessor, true);
                                 }
                                 None
                             }
                         };
 
-                        exported.push(name, &prefix, &js_docs, &code, &ts_docs, ts);
+                        match exported {
+                            ClassOrNamespace::Class(class) => {
+                                class.push(name, &prefix, &js_docs, &code, &ts_docs, ts);
+                            }
+                            ClassOrNamespace::Namespace(ns) => {
+                                if !receiver.is_static() {
+                                    bail!(
+                                        "non-static method `{}` on namespace `{}`",
+                                        name,
+                                        ns.name
+                                    );
+                                }
+                                ns.push(name, &js_docs, &code, &ts_docs, ts);
+                            }
+                        }
                     }
                 }
             }
@@ -3957,7 +4144,7 @@ __wbg_set_wasm(wasm);"
 
         self.export(
             &enum_.name,
-            &format!("Object.freeze({{\n{}}})", variants),
+            ExportJs::Expression(&format!("{{\n{}}}", variants)),
             Some(&docs),
         )?;
 
@@ -4008,7 +4195,7 @@ __wbg_set_wasm(wasm);"
     }
 
     fn generate_struct(&mut self, struct_: &AuxStruct) -> Result<(), Error> {
-        let class = require_class(&mut self.exported_classes, &struct_.name);
+        let class = self.require_class(&struct_.name);
         class.comments = format_doc_comments(&struct_.comments, None);
         class.is_inspectable = struct_.is_inspectable;
         class.generate_typescript = struct_.generate_typescript;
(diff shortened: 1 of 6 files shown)


daxpedda:
> 2. Error if users try to add instance methods/getters/setters or static getters/setters.
>
>    This is technically a **breaking change**, because string enums technically supported all this before. Technically, because them supporting it was a bug. Both string enums and C-style enums had the bug where they tried to generate a class for methods and getters/setters. For C-style enums, with resulted in a runtime error in JS, because both the enum and class were exported under the same name. However, string enums don't export anything, so the class didn't cause any problems. While buggy (there were no type definitions for the generated class) and unintended, this somewhat worked. Users could successfully define and use static methods for string enums.
>
>    Why is this a breaking change? The namespaces I implemented don't support static getters and setter. While it is trivial to implement static getter/setters in JS via [`Object.defineProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#creating_a_property), the problem is typing them in TS. TS only support getters/setters on classes. The best `namespace`s can do is to `export const/let` like [this](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAOwIYFtgGcxIMbDgMQgjgG8AoOSuUSWObCBdeAQTgF44BOAbgqprR4AG2DwAQhzgBGAAx8AvmTIB6FXDYBLdHCjAkAE0bCAngBo46TSk3CkUODGJI4AczExgUMoQgA6Nk4AFgAmPjU4SW04AHcoTU8kACNgYQsrGzsHJzgXdxhPKABadA8vOCxNb18-SWCwoA).
>
>     So the question is: are we okay with breaking user code relying on a bug, or should I add support for static getters/setters on namespaces?

I don't think this was necessary unintended by users. After all these methods are used internally by Rust as well, so I think we shouldn't disallow them. We might want to require users to do this via a separate `impl` block which doesn't have the `#[wasm_bindgen]` attribute, but indeed, this is a breaking change.

If we want users to use a separate `impl` block, we can use warnings instead of errors. The upside is that its clear to users what is and what isn't exported. So AFAICS this is the best route.

Adding support for static getters and setters will still not solve the issue with non-static methods, where we still need to decide if we want users to have a separate `impl` block or not.

I'm unable to make a decision on implementing static getters and setters, I'm having a hard time getting a feel of how "okay this is according to TS". If you believe this doesn't look too strange, it would sounds like a good feature to me to have.

daxpedda:
> Another thing to note is how string enums are handled. String enums don't have an exported object the namespace could latch onto, so an empty object is automatically exported.

We should definitely not do this here, instead lets resolve #4260 first.

RunDevelopment:
> I don't think this was necessary unintended by users. After all these methods are used internally by Rust as well, so I think we shouldn't disallow them.

Uhm, please reconsider. Okay, so here are some details on how string enum methods """work""" in JS. Firstly, string enums don't implement `RefFromWasmAbi`, so you can only consume `self` in methods/getters/setters. Secondly, please look at the JS code gen of a string enum with a constructor and instance method:

<details>
<summary>Detailed example (it's a lot of code but necessary)</summary>

```rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub enum StringEnum {
    A = "a",
    B = "b",
}

#[wasm_bindgen]
impl StringEnum {
    #[wasm_bindgen(constructor)]
    pub fn new() -> Self {
        Self::A
    }
    pub fn consume(self) {}
}
```

```js
let wasm;
export function __wbg_set_wasm(val) {
    wasm = val;
}


const __wbindgen_enum_StringEnum = ["a", "b"];

const StringEnumFinalization = (typeof FinalizationRegistry === 'undefined')
    : new FinalizationRegistry(ptr => wasm.__wbg_stringenum_free(ptr >>> 0, 1));

export class StringEnum {

    __destroy_into_raw() {
        const ptr = this.__wbg_ptr;
        this.__wbg_ptr = 0;
        StringEnumFinalization.unregister(this);
        return ptr;
    }

    free() {
        const ptr = this.__destroy_into_raw();
        wasm.__wbg_stringenum_free(ptr, 0);
    }
    constructor() {
        const ret = wasm.stringenum_new();
        return __wbindgen_enum_StringEnum[ret];
    }
    consume() {
        const ptr = this.__destroy_into_raw();
        wasm.stringenum_consume(ptr);
    }
}
```
```wat
(module $reference_test.wasm
  (type (;0;) (func (result i32)))
  (type (;1;) (func (param i32)))
  (func $stringenum_new (;0;) (type 0) (result i32))
  (func $stringenum_consume (;1;) (type 1) (param i32))
  (memory (;0;) 17)
  (export "memory" (memory 0))
  (export "stringenum_new" (func $stringenum_new))
  (export "stringenum_consume" (func $stringenum_consume))
  (@custom "target_features" (after code) "\04+\0amultivalue+\0fmutable-globals+\0freference-types+\08sign-ext")
)
```

Please feast your eyes upon this beauty. A few things to note:

1. The constructor doesn't work. It doesn't initialize the `__wbg_ptr` field, meaning that class instances created by it are already in a freed/destroyed state. It instead just returns the string value of `StringEnum::A`. This means that there is no way to create valid instances of this class. \
   (This is minor, but the constructor also doesn't register instances in the generated `FinalizationRegistry`.)
3. The `consume` method doesn't work. `this.__destroy_into_raw()` always returns `undefined`, which is obviously an invalid pointer. Also, I'm not sure whether `wasm.stringenum_consume` even wants a pointer, since the WASM ABI of string enums is an i32/u32 that identifies the variant.
4. The `free` method and finalizer will always fail, because `wasm.__wbg_stringenum_free` doesn't exist.

So to actually use a class generated for a string enum, the user would have to either overwrite the constructor or manually initialize the `__wbg_ptr` field later. Either way, they would have to write the logic for allocating and returning a pointer to a Rust `StringEnum` type. And once they've done all that work to create a valid instance, they get to use it at most once (if it works at all. As I said, I'm not sure about the ABI).

</details>

So yeah, string enum methods are generated in a completely dysfunctional state. I highly highly doubt that anyone is using them on the JS side. This is why I can only see string enum methods as a bug.

That all said, I agree that some users might have accidentally included instance methods in an `impl` block with `#[wasm_bindgen]` and thought "if it compiles, it works!" So how about this: instead of a hard error, we just ignore instance methods/getters/setters and constructors for string enums and don't generate a JS class for it for now? This would be backwards compatible, since no one is using the generated JS class anyway. This does mean that we'll have a few unnecessarily exported functions in the wasm binary, but that's shouldn't cause any problems, right? 

Then in the next major version, we can make it hard error and tell people to use a separate `impl` block.

What do you think?

> I'm unable to make a decision on implementing static getters and setters, I'm having a hard time getting a feel of how "okay this is according to TS". If you believe this doesn't look too strange, it would sounds like a good feature to me to have.

On the TS side, it's pretty hacky. It's not that getting getters+setters to work on a `namespace` is difficult to implement in JS, it's just that getters/setters can only be typed with `classe`s and `interface`s.  So using `export const` and `export let` is pretty hacky. I also don't see TS adding a way to properly type this any time soon.

But I still think we should support them. They are just too useful and natural to not support. Even if the type definitions may be a bit hacky, I think it's still worth it. 

Or put another way, I don't like as the one that implements the feature, but I want it as someone that will use the feature.

> > Another thing to note is how string enums are handled. String enums don't have an exported object the namespace could latch onto, so an empty object is automatically exported.
> 
> We should definitely not do this here, instead lets resolve #4260 first.

Sorry, but I don't see why this is a problem? We necessarily need to export a "physical" JS object to create a namespace.

I also don't see how this relates to #4260. #4260 is about the TS type of string enums (not) being exported. This PR only exports a JS object and doesn't change the type or whether it's exported at all.

daxpedda:
> > I don't think this was necessary unintended by users. After all these methods are used internally by Rust as well, so I think we shouldn't disallow them.
> 
> Uhm, please reconsider. Okay, so here are some details on how string enum methods """work""" in JS. Firstly, string enums don't implement `RefFromWasmAbi`, so you can only consume `self` in methods/getters/setters. Secondly, please look at the JS code gen of a string enum with a constructor and instance method:
> Detailed example (it's a lot of code but necessary)
> 
> 
> So yeah, string enum methods are generated in a completely dysfunctional state. I highly highly doubt that anyone is using them on the JS side. This is why I can only see string enum methods as a bug.
> 
> That all said, I agree that some users might have accidentally included instance methods in an `impl` block with `#[wasm_bindgen]` and thought "if it compiles, it works!" So how about this: instead of a hard error, we just ignore instance methods/getters/setters and constructors for string enums and don't generate a JS class for it for now? This would be backwards compatible, since no one is using the generated JS class anyway. This does mean that we'll have a few unnecessarily exported functions in the wasm binary, but that's shouldn't cause any problems, right?
> 
> Then in the next major version, we can make it hard error and tell people to use a separate `impl` block.
> 
> What do you think?

I was saying that regular methods can be used in Rust as well, this is why we shouldn't disallow it. I didn't know that e.g. constructors were allowed in first place and that regular methods generated any JS code.

What we can do is make `#[wasm_bindgen(constructor)]` error, because it is simply not supported. Anything else should be allowed, but not generate any JS code as you were already suggesting. As an additional fix we should also just not export these functions in the Wasm binary as well.

I think it would also be nice if we emit a warning for every function in an `#[wasm_bindgen]` `impl` block that isn't being exported. But if this is too hard I don't want to make this a requirement for this PR.
> > I'm unable to make a decision on implementing static getters and setters, I'm having a hard time getting a feel of how "okay this is according to TS". If you believe this doesn't look too strange, it would sounds like a good feature to me to have.
> 
> On the TS side, it's pretty hacky. It's not that getting getters+setters to work on a `namespace` is difficult to implement in JS, it's just that getters/setters can only be typed with `classe`s and `interface`s. So using `export const` and `export let` is pretty hacky. I also don't see TS adding a way to properly type this any time soon.
> 
> But I still think we should support them. They are just too useful and natural to not support. Even if the type definitions may be a bit hacky, I think it's still worth it.
> 
> Or put another way, I don't like as the one that implements the feature, but I want it as someone that will use the feature.

You make it sound like its not very TSy, so my gut feeling is that we should not implement this. Lets move this discussion to a follow-up PR.
> > > Another thing to note is how string enums are handled. String enums don't have an exported object the namespace could latch onto, so an empty object is automatically exported.
> > 
> > 
> > We should definitely not do this here, instead lets resolve #4260 first.
> 
> Sorry, but I don't see why this is a problem? We necessarily need to export a "physical" JS object to create a namespace.
> 
> I also don't see how this relates to #4260. #4260 is about the TS type of string enums (not) being exported. This PR only exports a JS object and doesn't change the type or whether it's exported at all.

My thinking here is that we should not expose static methods of a type we don't export. This would apply to TS as well. You make it sound obvious that there is no relation at all, am I missing something?

RunDevelopment:
Since string enums seem to cause of few problems, maybe I should change the PR to only affect C-style enums? Basically, enable namespaces for C-style enums, fix up the situation around string enums, and then enable namespaces for string enums.

Or we fix up things around string enums beforehand and block this PR until then.

I don't mind either way. What do you think?

> What we can do is make `#[wasm_bindgen(constructor)]` error, because it is simply not supported. Anything else should be allowed, but not generate any JS code as you were already suggesting. As an additional fix we should also not export these functions in the Wasm binary.

Sounds good 👍 

As for not exporting those functions and emitting warning, I think this should be a separate PR.

> You make it sound like its not very TSy, so my gut feeling is that we should not implement this. Lets move this discussion to a follow-up PR.

I kind of agree, but the problem is that static getters and setters already work today for string enums.

> My thinking here is that we should not expose static methods of a type we don't export. This would apply to TS as well. You make it sound obvious that there is no relation at all, am I missing something?

Ah, that's what you meant. Yeah, I was thinking of namespaces as mostly independent of the type they add things too, but you're right that this is a bit confusing.

daxpedda:
> Since string enums seem to cause of few problems, maybe I should change the PR to only affect C-style enums? Basically, enable namespaces for C-style enums, fix up the situation around string enums, and then enable namespaces for string enums.
> 
> Or we fix up things around string enums beforehand and block this PR until then.
> 
> I don't mind either way. What do you think?

I don't mind either way as well, but I would go with fixing string enums first.

> > You make it sound like its not very TSy, so my gut feeling is that we should not implement this. Lets move this discussion to a follow-up PR.
> 
> I kind of agree, but the problem is that static getters and setters already work today for string enums.

Strange, I couldn't confirm this. What did you do to reproduce this exactly?

RunDevelopment:
> Strange, I couldn't confirm this. What did you do to reproduce this exactly?

<details>
<summary>Code</summary>

```rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub enum Foo {
    A = "a",
    B = "b",
}

#[wasm_bindgen]
impl Foo {
    pub fn parse(s: &str) -> Foo {
        todo!()
    }
    #[wasm_bindgen(getter)]
    pub fn favorite() -> Foo {
        Foo::A
    }
    #[wasm_bindgen(setter)]
    pub fn set_favorite(value: Foo) {}
}
```
```ts
/* tslint:disable */
/* eslint-disable */
type Foo = "a" | "b";

```
```js
const __wbindgen_enum_Foo = ["a", "b"];

const FooFinalization = (typeof FinalizationRegistry === 'undefined')
    : new FinalizationRegistry(ptr => wasm.__wbg_foo_free(ptr >>> 0, 1));

export class Foo {

    __destroy_into_raw() {
        const ptr = this.__wbg_ptr;
        this.__wbg_ptr = 0;
        FooFinalization.unregister(this);
        return ptr;
    }

    free() {
        const ptr = this.__destroy_into_raw();
        wasm.__wbg_foo_free(ptr, 0);
    }
    /**
     * @param {string} s
     * @returns {Foo}
     */
    static parse(s) {
        const ptr0 = passStringToWasm0(s, wasm.__wbindgen_malloc, wasm.__wbindgen_realloc);
        const len0 = WASM_VECTOR_LEN;
        const ret = wasm.foo_parse(ptr0, len0);
        return __wbindgen_enum_Foo[ret];
    }
    /**
     * @returns {Foo}
     */
    static get favorite() {
        const ret = wasm.foo_favorite();
        return __wbindgen_enum_Foo[ret];
    }
    /**
     * @param {Foo} value
     */
    static set favorite(value) {
        wasm.foo_set_favorite((__wbindgen_enum_Foo.indexOf(value) + 1 || 3) - 1);
    }
}
```

</details>

There are no types generated, but the JS code gen actually works. Static methods, getters, and setters are functional.

RunDevelopment:
So to summarize, the plan for now is to:

1. Disallow `constructor` methods and instance `getter`/`setter` methods for all enums (C-style and string).
2. Do not generate code for instance methods. (Maybe emit a warning nudging the user to use separate `impl` blocks.)
3. Add support for static getters/setters for all enums.

daxpedda:
Yes, except for:
> 3. Add support for static getters/setters for all enums.

Which we should do in a follow-up PR, first figuring out if this is something we want to support in the first place.

RunDevelopment:
Okay:

- I disallowed constructors, getters, and setters entirely for enums. 
- Static methods will be exported like before. 
- Instance method will be ignored when generating JS bindings, but are still included in the WASM binary. There's also warning that I `println!`ed. Not pretty.

---

I also just thought that we could maybe support instance methods by automatically making them static methods. My thought was that an instance method has the same ABI as a static method where the first argument is `arg: Self`. So we could do this:

<details>
<summary>Code</summary>

```rs
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
#[derive(Copy, Clone)]
pub enum ImageFormat {
    PNG,
    JPEG,
    GIF,
}

#[wasm_bindgen]
impl ImageFormat {
    pub fn from_str(s: &str) -> ImageFormat {
        match s {
            "PNG" => ImageFormat::PNG,
            "JPEG" => ImageFormat::JPEG,
            "GIF" => ImageFormat::GIF,
            _ => panic!("unknown image format: {}", s),
        }
    }

    pub fn is_lossless(self) -> bool {
        matches!(self, ImageFormat::PNG | ImageFormat::GIF)
    }
}
```
```ts
export enum ImageFormat {
  PNG = 0,
  JPEG = 1,
  GIF = 2,
}
export namespace ImageFormat {
  export function from_str(s: string): ImageFormat;
  export function is_lossless(self: ImageFormat): boolean;
}
```

</details>

This should probably be a follow-up PR though.

What do you think?

crates/cli/tests/reference/namespace.rs
```diff
@@ -0,0 +1,52 @@
...
+            "JPEG" => ImageFormat::JPEG,
+            "GIF" => ImageFormat::GIF,
+            _ => panic!("unknown image format: {}", s),
+        }
+    }
+
+    /// I will be ignored by the generated JS bindings.
+    pub fn is_lossless(self) -> bool {
+        matches!(self, ImageFormat::PNG | ImageFormat::GIF)
+    }
```
daxpedda, line 26: This should instead be outputted as-is by the proc-macro and not generate any `extern "C" fn` blocks.

Alternatively, instead of going through that work, you can go ahead and implement instance methods as you have suggested.

crates/cli-support/src/js/mod.rs
```diff
@@ -2873,33 +3048,63 @@ __wbg_set_wasm(wasm);"
...
+                                if !is_static {
+                                    let msg = format!(
+                                        "The enum `{}` cannot support the instance method `{}`. \
+                                        No binding will be generated for this method. \
+                                        Consider moving the method in an `impl` block with the `#[wasm_bindgen]` attribute to avoid exporting it, \
+                                        or making it a static method by replacing `self` with `value: Self`.",
+                                        ns.name,
+                                        name
+                                    );
+                                    println!("WARNING: {}", msg);
```
daxpedda, line 3102: The warning has to happen in the proc-macro, not in the CLI.
Warnings in proc-macros are a bit tough, so its not a requirement for the PR.

crates/cli-support/src/js/mod.rs
```diff
@@ -2831,11 +2981,24 @@ __wbg_set_wasm(wasm);"
...
                     AuxExportKind::Constructor(class) => {
-                        let exported = require_class(&mut self.exported_classes, class);
+                        let exported = self.require_class_or_namespace(class);
+                        let exported = match exported {
+                            ClassOrNamespace::Class(class) => class,
+                            ClassOrNamespace::Namespace(_) => {
+                                bail!(
+                                    "constructor is not supported on `{}`. Enums do not support exported constructors.",
+                                    class
+                                );
```
daxpedda, line 2999: This has to happen in the proc-macro instead. In which case this becomes `unreachable!()`.
A UI test would be great as well.

crates/cli-support/src/js/mod.rs
```diff
@@ -2850,21 +3013,33 @@ __wbg_set_wasm(wasm);"
...
                             AuxExportedMethodKind::Method => ts_sig,
                             AuxExportedMethodKind::Getter => {
+                                let class = match exported {
+                                    ClassOrNamespace::Class(ref mut class) => class,
+                                    ClassOrNamespace::Namespace(_) => {
+                                        bail!(
+                                            "the getter `{}` is not supported on `{}`. Enums only support static methods on them.",
+                                            name,
+                                            class
+                                        );
```
daxpedda, line 3033: Same as above.

crates/cli-support/src/js/mod.rs
```diff
@@ -2873,33 +3048,63 @@ __wbg_set_wasm(wasm);"
...
                             }
                             AuxExportedMethodKind::Setter => {
+                                let class = match exported {
+                                    ClassOrNamespace::Class(ref mut class) => class,
+                                    ClassOrNamespace::Namespace(_) => {
+                                        bail!(
+                                            "the setter `{}` is not supported on `{}`. Enums only support static methods on them.",
+                                            name,
+                                            class
+                                        );
```
daxpedda, line 3066: Same as above.

crates/cli/tests/reference/namespace.js
```diff
@@ -0,0 +1,125 @@
...
+     * @param {string} s
+     * @returns {ImageFormat}
+     */
+    ImageFormat.from_str = function from_str(s) {
+        const ptr0 = passStringToWasm0(s, wasm.__wbindgen_malloc, wasm.__wbindgen_realloc);
+        const len0 = WASM_VECTOR_LEN;
+        const ret = wasm.imageformat_from_str(ptr0, len0);
+        return ret;
+    };
+})(ImageFormat);
```
daxpedda, line 107: Same as before with my question on why TS compiles enums into JS so strangely.
Couldn't we do this directly instead of wrapping it into a function and then executing it?

daxpedda (changes requested):
> I also just thought that we could maybe support instance methods by automatically making them static methods. My thought was that an instance method has the same ABI as a static method where the first argument is `arg: Self`. 

> This should probably be a follow-up PR though.
> 
> What do you think?

Yes, lets do it, but I guess we can't do it for string enums until we resolve #4260, because they can't be instantiated.

crates/cli/tests/reference/namespace.js
```diff
@@ -0,0 +1,125 @@
...
+}
+/**
+ * C-style enum
+ * @enum {0 | 1 | 2}
+ */
+export const ImageFormat = {
+    PNG: 0, "0": "PNG",
+    JPEG: 1, "1": "JPEG",
+    GIF: 2, "2": "GIF",
+};
```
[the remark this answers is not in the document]
RunDevelopment: Because TS cannot control the order of enums and namespaces. Since TS has to merge enums and namespaces, and you can declare them in any order. So TS cannot create the enum object directly since it might already exist. It has to assign each property individually.

`ImageFormat[ImageFormat["PNG"] = 0] = "PNG"` is just a little trick to compactly write:

```js
ImageFormat["PNG"] = 0
ImageFormat[0] = "PNG"
```

Remember how in C assignment to variables is an expression that returns the assigned value? Same thing in JS.

crates/cli/tests/reference/namespace.js
```diff
@@ -0,0 +1,125 @@
...
+     */
+    ImageFormat.from_str = function from_str(s) {
+        const ptr0 = passStringToWasm0(s, wasm.__wbindgen_malloc, wasm.__wbindgen_realloc);
+        const len0 = WASM_VECTOR_LEN;
+        const ret = wasm.imageformat_from_str(ptr0, len0);
+        return ret;
+    };
+})(ImageFormat);
+
+export const Status = {};
```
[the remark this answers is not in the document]
RunDevelopment: This is necessary for namespaces to work. If we don't export the object of the namespace, then the whole namespace is inaccessible.

crates/cli-support/src/js/mod.rs
RunDevelopment: That will be difficult. We don't know whether a type is a string or an enum until the whole program is parsed.

crates/cli/tests/reference/namespace.js
RunDevelopment: See above. In general, whenever TS code gen seems weird, they do it to support some edge cases. TS generally only has one code gen strategy per language feature and that code gen strategy has to work with all other language features.

crates/cli/tests/reference/namespace.js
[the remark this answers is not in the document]
daxpedda: Ooooh :bulb:!
Thanks!

crates/cli-support/src/js/mod.rs
daxpedda: Yeah, that's bad ... I guess that explains why this was implemented how it was until now.

I guess the best we can do now is emit a warning in `wasm-bindgen`, but erroring out is not good. Not all library authors actually run `wasm-bindgen-cli` in their CI, they just `cargo build --target wasm32-unknown-unknown` and call it a day.

crates/cli/tests/reference/namespace.js
daxpedda: In this case, because we control the ordering, shouldn't we be able to just assign it directly?
Or what am I missing?

RunDevelopment: Done. I removed the IIFE from the ESM output, but it has to stay for `nodejs` and `no-modules`, because namespaces are generated without knowledge of how they will be exported. See the new tests.

crates/cli-support/src/js/mod.rs
daxpedda: Just remembered that this can actually be done via the `ClassMarker` (or w/e its called, on my phone rn).

RunDevelopment:
Okay, to emit proper warnings and later error, I need to analyze the whole parsed program. There is no other way around it.

I also found out that methods most likely take a pointer as `self` and not the enum value. So the easierst way to convert these methods into static methods is in a post-process step after the whole process is parsed and before code gen. In other words, same as when warnings are emitted.

So that's what needs to be done next, I guess.

RunDevelopment:
So I just realized that analysis that requires the whole program is impossible to do inside the proc_macro. A macro can only see the thing it annotates and nothing outside of it. So e.g. for:

```rs
#[wasm_bindgen]
pub enum Foo { A, B, C }

#[wasm_bindgen]
impl Foo {
  #[wasm_bindgen(constructor)]
  pub fn new(): Self { Foo::A }
}
```

there is no way for the attribute annotating the `impl` to see the definition of `Foo`. So it has no way of knowing whether `Foo` is an `enum` or a `struct`.

So there is no way for the proc macro to emit warning or no export enum instance methods, since we can't even detect them.

The best we could do would be to do some magic with traits to cause compiler errors (basically, like C++ `static_assert`s), but that wouldn't be backwards compatible.

---

The only thing that we can do right now, is to ignore everything invalid on the CLI side. The wasm binary will have a unused imports, but there is nothing we can do about it without breaking stuff.

daxpedda:
> I also found out that methods most likely take a pointer as `self` and not the enum value. So the easierst way to convert these methods into static methods is in a post-process step after the whole process is parsed and before code gen. In other words, same as when warnings are emitted.

I'm not sure what you mean here exactly. Considering that only `self` methods are allowed, we could and should pass the value and not a reference to it.

`#[wasm_bindgen]` on enums also prevents enums from implementing `Drop`, so this shouldn't be a concern as well.

> The best we could do would be to do some magic with traits to cause compiler errors (basically, like C++ `static_assert`s), but that wouldn't be backwards compatible.

Why would that not be backwards compatible? In any case, this is how its currently done with e.g. not allowing `&self` methods on enums.

crates/cli-support/src/js/mod.rs
daxpedda: See my comment in https://github.com/rustwasm/wasm-bindgen/pull/4258#issuecomment-2477492640, this is definitely possible.

RunDevelopment:
> I'm not sure what you mean here exactly. Considering that only self methods are allowed, we could and should pass the value and not a reference to it.

I might have to look at that again, but what I meant is that the ABI is a `i32` and that the generated TS type is `number`. The ABI of exported Rust structs for `self` parameters is a pointer to the Rust struct instance in wasm memory. So `self`, `&self`, and `&mut self` all have the same ABI for exported Rust structs.

But I haven't actually looked at the Rust code gen, so I might be wrong that it's a pointer.

> > The best we could do would be to do some magic with traits to cause compiler errors (basically, like C++ `static_assert`s), but that wouldn't be backwards compatible.
> 
> Why would that not be backwards compatible? In any case, this is how its currently done with e.g. not allowing `&self` methods on enums.

What I meant was that we could have internal marker traits (e.g. `CanHaveConstructor`, `CanHaveMethods`) that `#[wasm_bindgen]` automatically implements for `struct`s but not for `enum`s. Then the code gen for e.g. `#[wasm_bindgen(constructor)]` can include a compile-time check for `Self: CanHaveConstructor`.

While this would enforce correctness, it's also a breaking change. All enums currently compile with constructors and instance methods, it's just that CLI fails for C-style enums.

daxpedda:
> > I'm not sure what you mean here exactly. Considering that only self methods are allowed, we could and should pass the value and not a reference to it.
> 
> I might have to look at that again, but what I meant is that the ABI is a `i32` and that the generated TS type is `number`. The ABI of exported Rust structs for `self` parameters is a pointer to the Rust struct instance in wasm memory. So `self`, `&self`, and `&mut self` all have the same ABI for exported Rust structs.
> 
> But I haven't actually looked at the Rust code gen, so I might be wrong that it's a pointer.

You are describing how it works for structs, but it shouldn't work that way for enums, because enums can be instantiated in JS/TS without FFI'ing into Rust, they can't really be pointers. String enums were also optimized to just be a `u32` by #3915 and conversion is done in JS.

That said, I have no idea how enum methods work right now and it might be buggy. Please let me know if I missed something.

> > > The best we could do would be to do some magic with traits to cause compiler errors (basically, like C++ `static_assert`s), but that wouldn't be backwards compatible.
> > 
> > 
> > Why would that not be backwards compatible? In any case, this is how its currently done with e.g. not allowing `&self` methods on enums.
> 
> What I meant was that we could have internal marker traits (e.g. `CanHaveConstructor`, `CanHaveMethods`) that `#[wasm_bindgen]` automatically implements for `struct`s but not for `enum`s. Then the code gen for e.g. `#[wasm_bindgen(constructor)]` can include a compile-time check for `Self: CanHaveConstructor`.
> 
> While this would enforce correctness, it's also a breaking change. All enums currently compile with constructors and instance methods, it's just that CLI fails for C-style enums.

If the CLI fails, I would also say its a breaking change, even if maybe not technically by Rust standards. In any case, we have already decided that this is an acceptable breaking change in https://github.com/rustwasm/wasm-bindgen/pull/4258#issuecomment-2469860735. We have historically made technically breaking changes in `wasm-bindgen` for things that are actually broken. Which is currently the case for `#[wasm_bindgen(constructor/getter/setter)]`. Please correct if I'm wrong.

RunDevelopment:
> That said, I have no idea how enum methods work right now and it might be buggy. Please let me know if I missed something.

You probably didn't miss anything. I just saw that the TS type was as generated as `number` instead of the enum type. So I assumed that the underlying ABI was `NonNull<EnumType>` (which has the WASM ABI of `i32`).

But I again, I'll have to check the Rust code gen to be sure.

> If the CLI fails, I would also say its a breaking change, even if maybe not technically by Rust standards. In any case, we have already decided that this is an acceptable breaking change in https://github.com/rustwasm/wasm-bindgen/pull/4258#issuecomment-2469860735.

Oh, okay. I thought you only agreed in reference to making it breaking for CLI users. 

And just to clear, my proposed solution would result in a compiler error for all users. E.g. the following code:

```rs
#[wasm_bindgen]
pub enum Foo { A = "a", B = "b" }

#[wasm_bindgen]
impl Foo {
    #[wasm_bindgen(constructor)]
    pub fn new() -> Self { Foo::A }
}
```

would result in generated code like this:

```rs
pub enum Foo { A = 0, B = 1 }
impl wasm_bindgen::marker::CanHaveStaticMethod for Foo {}
// wasm_bindgen::marker::CanHaveConstrctor is not implemented

impl Foo {
    pub fn new() -> Foo {
        // compile-time assert
        const _: () = {
            fn check<T: wasm_bindgen::marker::CanHaveConstrctor>() {}
            check::<Foo>(); // will fail, because the trait is not implemented
        };

        // the usual code gen
        #[no_mangel]
        fn __wbg_export_foo_constructor() -> Foo::Abi { /* ... */ }
        /* ... */
    }
}
```

This would result in a compiler error for whoever is compiling the WASM binary.

I'll probably make that a separate PR to keep this PR focused on namespaces.

PR #4262, opened 2024-11-11
Added note for which versions of `wasm-bindgen` support i128

RunDevelopment:
I forgot to add this to the docs in #4222. Since numbers are pretty fundamental, I think we should be clear that i128 is only supported in newer versions of WBG.


daxpedda (approved):
Thanks!

PR #4266, opened 2024-11-13
Add `WASM_BINDGEN_DRIVER_TIMEOUT`

daxpedda:
This PR introduces the `WASM_BINDGEN_DRIVER_TIMEOUT` environment variable for `wasm-bindgen-test-runner`, which sets the timeout for connecting to the test driver.

Before this PR the timeout was fixed to 5 seconds, which failed often when testing Safari in GitHub Actions CI. My experience is that the MacOS ARM runners, nowadays the default, tend to stall quite a lot, which I believe was the reason why this triggered the timeout so often and made CIs fail consistently.

### DIFF
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
@@ -188,7 +188,7 @@ fn main() -> anyhow::Result<()> {
         return Ok(());
     }
 
-    let timeout = env::var("WASM_BINDGEN_TEST_TIMEOUT")
+    let test_timeout = env::var("WASM_BINDGEN_TEST_TIMEOUT")
         .map(|timeout| {
             timeout
                 .parse()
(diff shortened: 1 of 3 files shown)


crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -197,9 +197,17 @@ fn main() -> anyhow::Result<()> {
...
+        println!("Set timeout to {} seconds...", test_timeout);
     }
 
+    let driver_timeout = env::var("WASM_BINDGEN_DRIVER_TIMEOUT")
+        .map(|timeout| {
+            timeout
+                .parse()
+                .expect("Could not parse 'WASM_BINDGEN_DRIVER_TIMEOUT'")
+        })
+        .unwrap_or(5);
```
RunDevelopment, line 209: Maybe the default could also be raised if it causes timeouts? Maybe 10 or 20 seconds.

I mean, the worst that can happen by raising it is that tests take a bit longer to time out.

daxpedda:
Turns out this wasn't the issue after all. I'm gonna make a new PR shortly.
@RunDevelopment I appreciate you taking a look :heart:!

PR #4267, opened 2024-11-13
Restart WebDriver on failure

daxpedda:
Currently, if starting the WebDriver process fails, the test runner simply spends its remaining timeout period trying to connect to the failed WebDriver process, which obviously doesn't work.

Instead this PR changes the behavior to continuously check if the WebDriver process didn't fail and only then wait until a connection is established. The timeout period remains 5 seconds in total.

I've tested this quite extensively now with `web-time` and I only got one unrelated failure. So this is a huge improvement from before, where almost every single CI run has failed.


RunDevelopment (approved):
LGTM!

Issue #2338, opened 2020-10-25
Unable to export js_name = round unless type is f64


jramsley:
### Describe the Bug

```
#[wasm_bindgen(js_name = round)]
pub fn js_math_round(a: f32) -> f32
```
 returns a signature mismatch of
```
rust-lld: error: function signature mismatch: round
          >>> defined as (f64) -> f64
```

### Steps to Reproduce

1. Create public function accessible in `lib.rs`
```
#[wasm_bindgen(js_name = round)]
pub fn js_math_round(a: f32) -> f32 {
    return a;
}
```
2. `wasm-pack build`

circumvented in my repo by naming `js_round` and post-processing
https://github.com/jramsley/gl-matrix/blob/main/src/vec2.rs#L108

### Expected Behavior

Generates an exported javascript function called `round`

### Actual Behavior

```
rust-lld: error: function signature mismatch: round
          >>> defined as (f64) -> f64 in ~/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libcompiler_builtins-a30046ba00655967.rlib(compiler_builtins-a30046ba00655967.compiler_builtins.ara2xh13-cgu.0.rcgu.o)
          >>> defined as (f32) -> f32 in ~/gl-matrix/target/wasm32-unknown-unknown/release/deps/gl_matrix.gl_matrix.7mocmiy6-cgu.1.rcgu.o
```

### Additional Context

First time writing something real in Rust, I may be missing something obvious.

pauldorehill:
Jake,

I couldn't get your bug to reproduce with your the repo posted or in a new project?

jramsley:
Hi @pauldorehill,

I removed my `target` and `pkg` folders and tried again.

First thing was to check for any `round`, `js_sys`, or `Math` imports via grep
```
$ grep -i -r round .
./src/vec2.rs:macro_rules! js_round {
./src/vec2.rs:#[wasm_bindgen(js_name = round)]
./src/vec2.rs:pub fn js_round(out: Float32Array, a: &Float32Array) -> Float32Array {
./src/vec2.rs:    out.set_index(0, js_round!(a.get_index(0)));
./src/vec2.rs:    out.set_index(1, js_round!(a.get_index(1)));

$ grep -i -r js_sys .
./src/vec2.rs:use js_sys::Float32Array;
./src/vec2.rs:use js_sys::JsString;

$ grep -i -r math .
```

[...]

PR #4269, opened 2024-11-13
Hash full signature for imported functions

### LINKED ISSUE #4268: Failed to import the same js functions with same rust fn name under different modules
### Describe the Bug

If I use `#[wasm_bindgen(js_name = parseFloat)]` to import the same js function in two modules, with the same rust fn name, but with different signatures, things don't work well.

```rust
use wasm_bindgen::JsValue;
use wasm_bindgen_test::*;

mod a {
    use wasm_bindgen::prelude::*;

    #[wasm_bindgen]
    extern "C" {
        #[wasm_bindgen(js_name = parseFloat)]
        pub fn parse_float(text: &JsValue) -> f64;
    }
}

mod b {
    use wasm_bindgen::prelude::*;

    #[wasm_bindgen]
    extern "C" {
        #[wasm_bindgen(js_name = parseFloat)]
        pub fn parse_float(text: &str) -> f64;
    }
}

#[wasm_bindgen_test]
pub fn pass_a() {
    let v = JsValue::from_str("1");
    assert_eq!(a::parse_float(&v), 1.);
}

#[wasm_bindgen_test]
pub fn pass_b() {
    assert_eq!(b::parse_float("2"), 2.);
}
```

The above test failed.

```
const wasmModule = new WebAssembly.Module(bytes);
                   ^

CompileError: WebAssembly.Module(): Compiling function #838:"f64::a::parse_float::hfd12f3d441bd8268" failed: not enough arguments on the stack for call (need 2, got 1) @+435657
```

### Steps to Reproduce

1. Test the above code or just clone https://github.com/EqualMa/example-wasm-bindgen-import-with-same-name
2. Run `wasm-pack test --node`

### Expected Behavior

Imports with the same js name and rust name should all work.

### Actual Behavior

If the imports have the same js name and rust name, it seems that they all *link* to the same bindgen js function. So if they have different signatures, calling one of them failed at runtime.

RunDevelopment:
Fixes #4268 

The problem in #4268 was the short hash added to disambiguate imported functions only hashed the Rust identifier of the function. This is enough to ensure uniqueness within a single Rust module, but not across modules as in #4268. This resulted in 2 different wasm functions being exported under the same name.

I fixed this by hashing not just the function name, but the whole signature (name + inputs + outputs + some extra). This means that 2 function will only have the same hash if they have the same name, input, and outputs. If this happens, the functions also have the same ABI and exporting them both under the same name actually works.

(Note: I don't like that 2 functions with the same signature have the same hash, so if you want to, I can hash the source code position of the function as well to make collisions very unlikely.)

Consequently, the hashes of all imported functions are now different. A lot of tests were updated, but the main new test is called `modules.rs`. 

I also had to enable the `'extra-traits'` feature on `syn`. This feature enables implementing `Hash` for `syn::Signature` and other types.

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -618,7 +618,7 @@ impl<'a> ConvertToAst<(&ast::Program, BindgenAttrs, &'a Option<ast::ImportModule
                 ast::ImportFunctionKind::Normal => (0, "n"),
                 ast::ImportFunctionKind::Method { ref class, .. } => (1, &class[..]),
             };
-            let data = (ns, &self.sig.ident, module);
+            let data = (ns, &self.sig, module);
             format!(
                 "__wbg_{}_{}",
                 wasm.name
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -618,7 +618,7 @@ impl<'a> ConvertToAst<(&ast::Program, BindgenAttrs, &'a Option<ast::ImportModule
                 ast::ImportFunctionKind::Normal => (0, "n"),
                 ast::ImportFunctionKind::Method { ref class, .. } => (1, &class[..]),
             };
-            let data = (ns, &self.sig, module);
+            let data = (ns, self.sig.to_token_stream().to_string(), module);
             format!(
                 "__wbg_{}_{}",
                 wasm.name
--- a/crates/cli/tests/reference/anyref-import-catch.js
+++ b/crates/cli/tests/reference/anyref-import-catch.js
@@ -52,7 +52,7 @@ export function exported() {
     }
 }
 
-export function __wbg_foo_fa0172cdeb670adb() { return handleError(function () {
+export function __wbg_foo_4e8309b1aa95a4ac() { return handleError(function () {
     foo();
 }, arguments) };
 
--- a/crates/cli/tests/reference/anyref-import-catch.js
+++ b/crates/cli/tests/reference/anyref-import-catch.js
@@ -52,7 +52,7 @@ export function exported() {
     }
 }
 
-export function __wbg_foo_95fe1a04017077db() { return handleError(function () {
+export function __wbg_foo_fa0172cdeb670adb() { return handleError(function () {
     foo();
 }, arguments) };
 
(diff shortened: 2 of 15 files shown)


daxpedda:
> (Note: I don't like that 2 functions with the same signature have the same hash, so if you want to, I can hash the source code position of the function as well to make collisions very unlikely.)

I'm a bit surprised why this doesn't cause issues already. I will take a deeper look.

> I also had to enable the `'extra-traits'` feature on `syn`. This feature enables implementing `Hash` for `syn::Signature` and other types.

That's probably very undesirable, I will take a look if we can avoid that.

crates/macro-support/src/parser.rs
```diff
@@ -618,7 +618,7 @@ impl<'a> ConvertToAst<(&ast::Program, BindgenAttrs, &'a Option<ast::ImportModule
                 ast::ImportFunctionKind::Normal => (0, "n"),
                 ast::ImportFunctionKind::Method { ref class, .. } => (1, &class[..]),
             };
-            let data = (ns, &self.sig.ident, module);
+            let data = (ns, &self.sig, module);
```
daxpedda, line 621:
```suggestion
            let data = (ns, self.sig.to_token_stream().to_string(), module);
```
I know this is a bit messed up ... but its worth it avoiding `extra-traits`.

daxpedda (changes requested):
Thank you for looking into it!

Unfortunately I couldn't find a way to get the line number or column from the `Span` or anywhere else, so I don't think this is possible outside of Rust nightly. I'm a bit perplexed how Rust is just ignoring the fact that we are exporting a function with the same exact name twice. I would have guessed a linker error or something.

I should mention that this is most likely buggy with other things as well, but I guess we will get there when we do.

RunDevelopment:
> I'm a bit perplexed how Rust is just ignoring the fact that we are exporting a function with the same exact name twice. I would have guessed a linker error or something.

Only if the two functions have different ABIs. They explained this when they talked about [why `no_mangle` is unsafe](https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#unsafe-attributes) in the 2024 edition. https://github.com/rust-lang/rust/issues/28179

I think this is also the cause of #2338, but I'm not 100% sure.

crates/macro-support/src/parser.rs
RunDevelopment: Huh, alright then.

daxpedda (approved):

PR #4270, opened 2024-11-14
Add test for imported getters and setters

RunDevelopment:
This PR just adds tests for imported getters and setters. It shows bugs that I want to fix.

1. Renaming for static getters and setters does not work correctly. See that the setter is always called `set_<name>`.
2. Static getters ignore the name set in the `getter` attribute. Same for setters.
3. The code gen for static getters/setters is not correct. Note that they are called as functions.

I already have a fix ready for 1 and 2, but I wanted to add the tests before that PR, so that it's easier to see how the fix PR changes the existing code gen.


daxpedda (approved):
This is awful ...
Thank you for doing this!

PR #4273, opened 2024-11-14
Make `getter`/`setter` names consistent with `js_name`

RunDevelopment:
This is the fix I was talking about in #4270. Please merge #4270, so the effects of this PR become obvious in git diff.

Changes:
- The names of getter/setters in now derived from the name of the underlying function, just like it is for everything else. This makes the function the single source of truth for the name of an exported function/method/getter/setter and prevents them from getting out of sync in the future. All AST structures have been adjusted to reflect this change.
- If a name is specified via both `getter` and `js_name`, `getter` will be chosen. This reflects the old behavior of exported getters. (Also applies to `setter`.)
- Imported static getters now don't ignore the name defined `getter`. This is a breaking change, since it potentially changes the import name. (Also applies to `setter`.)

This PR resolves TODO 2 from #4230.


RunDevelopment:
Interesting errors. The good news is that everything around getters and setters seems to be well-tested. I'll fix them over the coming days.

guybedford:
I've rebased this work to the main, with #4230 now landed.

PR #4276, opened 2024-11-17
Improved errors for invalid `self` arguments

RunDevelopment:
While writing tests, I forgot to add `#[wasm_bindgen]` on an `impl` block and the proc macro panicked with the message "arguments cannot be `self`". Since this was a panic, I tried to debug the proc macro for like 10 minutes before I noticed the missing `#[wasm_bindgen]` in my test.

This PR improves the error messages for `self` arguments in invalid positions. Instead of panics, users now get proper errors with helpful error messages.

Before:
```
error: custom attribute panicked
  --> ui-tests/invalid-self.rs:10:5
   |
10 |     #[wasm_bindgen(js_name = bar)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: message: arguments cannot be `self`
```

After:
```
error: the `self` argument is only allowed for functions in `impl` blocks.

       If the function is already in an `impl` block, did you perhaps forget to add `#[wasm_bindgen]` to the `impl` block?
  --> ui-tests/invalid-self.rs:11:17
   |
11 |     pub fn foo(&self) {}
   |                 ^^^^
```

For imported/`extern` functions with a `self` argument, I added a similar error message that suggests using `this` instead of `self`.

These new error messages should (1) make these issues easier to fix for our users and (2) make it clear that it's an error with user code and not the macro.

---

To implement the better error messages, I initially added a new parameter to `function_from_decl` to hold the function position (free, extern, impl), but realized that this single parameter could represent `allow_self: bool`, `self_ty: Option<&Ident>`, and `is_from_impl: bool` all in a single parameter. So I did that. In total, I added 1 new parameter and removed 3 old ones from `function_from_decl`. This makes the diff a bit larger, but I couldn't resist when it cleaned up the API of `function_from_decl` so nicely.

### DIFF
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -1016,14 +1016,14 @@ fn function_from_decl(
                 match position {
                     FunctionPosition::Free => {
                         bail_span!(
-                            r,
+                            r.self_token,
                             "the `self` argument is only allowed for functions in `impl` blocks.\n\n\
                             If the function is already in an `impl` block, did you perhaps forget to add `#[wasm_bindgen]` to the `impl` block?"
                         );
                     }
                     FunctionPosition::Extern => {
                         bail_span!(
-                            r,
+                            r.self_token,
                             "the `self` argument is not allowed for `extern` functions.\n\n\
                             Did you perhaps mean `this`? For more information on importing JavaScript functions, see:\n\
                             https://rustwasm.github.io/docs/wasm-bindgen/examples/import-js.html"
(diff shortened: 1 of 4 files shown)


crates/macro-support/src/parser.rs
```diff
@@ -910,6 +906,17 @@ fn get_self_method(r: syn::Receiver) -> ast::MethodSelf {
...
+enum FunctionPosition<'a> {
+    Extern,
+    Free,
+    Impl { self_ty: &'a Ident },
+}
+impl FunctionPosition<'_> {
+    fn is_impl(&self) -> bool {
+        matches!(self, FunctionPosition::Impl { .. })
+    }
+}
```
daxpedda, line 918: This seems unnecessary.

daxpedda (changes requested):
Good stuff!

Just a small nit. I will go ahead and address it myself.

daxpedda (approved):

PR #4278, opened 2024-11-18
Make `getter`s, `setter`s, and `constructor`s compiler errors for enums

RunDevelopment:
This PR implements a compile-time check to verify that enums have no methods with `#[wasm_bindgen(getter)]`, `#[wasm_bindgen(setter)]`, and `#[wasm_bindgen(constructor)]`. This was previously accepted and resulted in incorrect JS code gen.

The check is implemented like I outlined in [this comment](https://github.com/rustwasm/wasm-bindgen/pull/4258#issuecomment-2481306456). There are 3 new internal marker traits (`SupportsConstructor`, `SupportsInstanceProperty`, `SupportsStaticProperty`) that are implemented by `struct`s with `#[wasm_bindgen]` and not `enum`s. Exported function then generate a `const _: ()`  that checks whether the `Self` type (if any) implements those traits. I used `diagnostic::on_unimplemented` to make the trait errors easily understandable for users. 

Example error for using `#[wasm_bindgen(constructor)]` on an enum called `RustEnum`:

```
error[E0277]: JavaScript constructors are not supported for `RustEnum`
  --> ui-tests/unsupported-options.rs:56:12
   |
56 |     pub fn new() -> Self {
   |            ^^^ this function cannot be the constructor of `RustEnum`
   |
   = help: the trait `SupportsConstructor` is not implemented for `RustEnum`
   = note: `#[wasm_bindgen(constructor)]` is generally only supported for `struct`s with `#[wasm_bindgen]` and cannot be used for `enum`s.
   = note: Consider removing the `constructor` option and using a regular static method instead.
   = help: the trait `SupportsConstructor` is implemented for `RustStruct`
note: required by a bound in `__wasm_bindgen_generated_RustEnum_new::_::assert_supports_constructor`
  --> ui-tests/unsupported-options.rs:49:1
   |
49 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ required by this bound in `assert_supports_constructor`
```

As we can see, the error message is clear and even offers a suggestion to fix the error.

---

This PR is virtually a minimal implementation of a technique we could use a lot more. E.g. it can be used to verify that the `Self` type actually has a `#[wasm_bindgen]` macro (the test file I added includes an example where exactly this case does *not* cause an error). However, I kept it minimal so we can concentrate on the "how" this PR does things.

As for the how: I initially wanted to add just a single trait (something like `IsStruct`) and then let constructors/getters/setters/etc. assert that they are on a struct. This did work, but resulted in error messages that were a lot worse. By using traits that describe the capabilities of the type and not what the type is, error messages can be a lot more specific. Plus, adding/removing/changing capabilities seems easier at scale.


crates/backend/src/codegen.rs
```diff
@@ -780,12 +787,43 @@ impl TryToTokens for ast::Export {
...
+        let mut checks = Vec::new();
+        if self.start {
+            checks.push(quote! { const _ASSERT: fn() = || -> #projection::Abi { loop {} }; });
         };
 
+        if let Some(class) = self.rust_class.as_ref() {
+            // change span of class, so it points to the location of the
+            // function causing the assert to fail
+            let mut class = class.clone();
+            class.set_span(self.rust_name.span());
```
RunDevelopment, line 799: I'm not sure whether this is hygienic (or even correct). I did this so that the error message would point to the function causing the error instead of the type name on the `impl` block. This seems hacky, so please let me know if there are better ways of doing this.

RunDevelopment:
So that's unfortunate. [`#[diagnostic::on_unimplemented]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute) was only stabilized in Rust 1.78.0. So we can't use with our MSRV.

So either we give up on good error messages, or we conditionally enable the attribute only for Rust versions that support it [like zerocopy does](https://github.com/google/zerocopy/blob/main/Cargo.toml) (see their [`build.rs`](https://github.com/google/zerocopy/blob/main/build.rs)).

daxpedda:
We established over chat that we want to use [`rustversion`](https://crates.io/crates/rustversion) behind a crate feature enabled by default.

RunDevelopment:
Alright, I added a new `disagnostic` feature to gate the `rustversion` dependency, and enable the `diagnostic::on_unimplemented` attributes only when supported and wanted.

(Btw, `rustversion` itself is a joy to work with. It's so easy to use, I love it.)

Another change is that I raised the rust version we use to run UI tests in CI from 1.76 to 1.78. This is necessary for the CI to actually test the custom trait impl error messages. However, this also caused a few other UI tests to slightly change their output, so I had to update their output as well. Hence this PR touches a few unrelated test files.

Cargo.toml
```diff
@@ -22,12 +22,20 @@ features = ["serde-serialize"]
...
 spans = ["wasm-bindgen-macro/spans"]
 std = ["wasm-bindgen-macro/std", "once_cell/std"]
 
+# Enables the use of Rust's `diagnostic` macro for several traits to provide
+# better error messages when those traits are not implemented.
+#
+# Diagnostic messages are only available since Rust 1.78. Enabling this feature
+# for older compilers will NOT improve error message and will NOT result in a
+# compilation error.
+diagnostic = ["rustversion"]
```
daxpedda, line 37:
```suggestion
msrv = ["rustversion"]
```
We should just say here that this enables features that require a higher MSRV. The part where you clarify that this doesn't actually make compilation fail is great. Maybe use the word "compile-time detection" so users understand that they are opting into something here that will increase compilation time.

What this feature will actually do, until now only improve diagnostics, should be documented somewhere else and here we just leave a link to there.

crates/backend/src/codegen.rs
```diff
@@ -783,12 +790,44 @@ impl TryToTokens for ast::Export {
...
+            // little helper function to make sure the check points to the
+            // location of the function causing the assert to fail
+            let mut add_check = |token_stream| {
+                checks.push(respan(token_stream, &self.rust_name));
+            };
+
+            match &self.method_kind {
+                ast::MethodKind::Constructor => {
+                    add_check(quote! {
+                        struct CheckSupportsConstructor<T: #wasm_bindgen::__rt::marker::SupportsConstructor>(T);
```
daxpedda, line 808: We should avoid defining a bunch of new structs everywhere. `CheckSupportsConstructor` can be defined in `__rt` as well and reused.

crates/macro/ui-tests/main-debug.stderr
```diff
@@ -3,3 +3,11 @@ error: the main function has to be called main
...
 18 | fn fail() {}
    |    ^^^^
+
+warning: unused variable: `f`
+  --> ui-tests/main-debug.rs:12:19
+   |
+12 |     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+   |                   ^ help: if this is intentional, prefix it with an underscore: `_f`
+   |
+   = note: `#[warn(unused_variables)]` on by default
```
daxpedda, line 13: Lets fix that actually. We don't want to pollute those outputs.

crates/macro/ui-tests/main-infallible.stderr
```diff
@@ -3,3 +3,20 @@ error: the main function has to be called main
...
+  | unreachable expression
+  | any code following this expression is unreachable
+  |
+note: this expression has type `Infallible`, which is uninhabited
+ --> ui-tests/main-infallible.rs:4:1
+  |
+4 | #[wasm_bindgen(main)]
+  | ^^^^^^^^^^^^^^^^^^^^^
+  = note: `#[warn(unreachable_code)]` on by default
+  = note: this warning originates in the attribute macro `wasm_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)
```
daxpedda, line 22: Fun. That's a bug. We should avoid emitting this warning somehow.
Not related to this PR.

src/lib.rs
```diff
@@ -1582,6 +1583,10 @@ pub mod __rt {
     use alloc::boxed::Box;
     use alloc::rc::Rc;
 
+    pub mod marker {
+        pub use super::super::marker::*;
+    }
```
daxpedda, line 1588:
```suggestion
    pub use super::marker;
```
Right?

src/marker.rs
```diff
@@ -0,0 +1,85 @@
+//! This module contains marker traits used by `wasm-bindgen` to verify its
+//! generated code.
+//!
+//! NOTE: This module is not exported as is. It is re-exported under
+//! `wasm_bindgen::__rt::marker`.
+//!
+//! # ⚠️ Unstable
+//!
+//! This is an internal module, no stability guarantees are provided. Use at
+//! your own risk.
```
daxpedda, line 10: This isn't exposed, the notice is unnecessary.
Applies to the traits as well.

src/marker.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[cfg(feature = "diagnostic")]
+extern crate rustversion;
```
daxpedda, line 13:
```suggestion
#[cfg(feature = "diagnostic")]
use rustversion::rustversion;
```
This doesn't work?

src/marker.rs
```diff
@@ -0,0 +1,85 @@
...
+/// stability guarantees** are provided. Use at your own risk. See its
+/// documentation for more details.
+#[cfg_attr(
+    feature = "diagnostic",
+    rustversion::attr(
+        since(1.78),
+        diagnostic::on_unimplemented(
+            message = "JavaScript constructors are not supported for `{Self}`",
+            label = "this function cannot be the constructor of `{Self}`",
+            note = "`#[wasm_bindgen(constructor)]` is generally only supported for `struct`s with `#[wasm_bindgen]` and cannot be used for `enum`s.",
```
daxpedda, line 32:
```suggestion
            note = "`#[wasm_bindgen(constructor)]` is only supported for `struct`s and cannot be used for `enum`s.",
```
Applies to the others as well.

daxpedda (changes requested):
Amazing improvement, great work!

Missing a changelog entry.

Cargo.toml
RunDevelopment: And where should I put the diagnostics docs?

src/marker.rs
RunDevelopment:
```
error[E0432]: unresolved import `rustversion::rustversion`
 --> src\marker.rs:2:5
  |
2 | use rustversion::rustversion;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^ no `rustversion` in the root
```

src/lib.rs
RunDevelopment: That was my first thought too, but...

```
error[E0365]: `marker` is only public within the crate, and cannot be re-exported outside
    --> src\lib.rs:1593:13
     |
1593 |     pub use super::marker;
     |             ^^^^^^^^^^^^^ re-export of crate public `marker`
     |
     = note: consider declaring type or module `marker` with `pub`
```

The issue is that `pub use super::marker;` requires the original `marker` module to be public, but this would mean that people can import it via `wasm_bindgen::marker` as well (AFAIK), which we don't want.

So I created a new module of the same name and re-exported only the public elements inside the original `marker` module.

Cargo.toml
daxpedda: I have no idea! Ideally we have a place where we document all crate features, but I don't think we do!

I think the most appropriate place is top-level Rustdoc, so lets just start by at least documenting `msrv`.

src/lib.rs
daxpedda: Okay lets do this actually ... just move `__rt` to its own file.
```rust
#[path = "rt/mod.rs"]
pub mod __rt;
```
Then we can just split it up there and you can use `pub mod marker`.

src/marker.rs
daxpedda: Ah! The macro isn't actually called `rustversion`, `attr` is already the name of the proc-macro. So accessing it via `rustversion::attr` works just fine.
This line should be fine to remove entirely!

crates/backend/src/codegen.rs
```diff
@@ -798,7 +834,10 @@ impl TryToTokens for ast::Export {
                     export_name = #export_name,
                 )]
                 pub unsafe extern "C" fn #generated_name(#(#args),*) -> #wasm_bindgen::convert::WasmRet<#projection::Abi> {
-                    #start_check
+                    #[automatically_derived]
+                    const _: () = {
```
daxpedda, line 838:
```suggestion
                    const _: () = {
```
`#[automatically_derived]` only works on `impl` blocks, it has no effect when placed over any other item. But `const _` already has the same effect.

We (I) might have been overzealous and wrongly applying this in the past as well.

daxpedda (changes requested):
I just remembered: `crates/msrv` `lib` and `resolver` need to be fixed:

```toml
default = ["std"]
```
Should instead be:
```toml
default = [
  "wasm-bindgen/default",
  "js-sys/default",
  "wasm-bindgen-futures/default",
  "web-sys/default",
  "wasm-bindgen-test/default",
]
```

RunDevelopment:
> I just remembered: `crates/msrv` `lib` and `resolver` need to be fixed:

Please do that in a separate PR. This has nothing to do with the changes in this PR.

daxpedda:
> > I just remembered: `crates/msrv` `lib` and `resolver` need to be fixed:
> 
> Please do that in a separate PR. This has nothing to do with the changes in this PR.

`default = ["std"]` worked because the only relevant default crate feature was `std`. This PR adds a new default crate feature with new code paths. That has to be included in the tests.

src/lib.rs
```diff
@@ -14,6 +14,16 @@
 //! This feature currently enables the `std` feature, meaning that it is not
 //! compatible with `no_std` environments.
 //!
+//! ### `msrv` (default)
+//!
+//! Enables Rust language features that requires a higher MSRV. Enabling this
```
daxpedda, line 19:
```suggestion
//! Enables Rust language features that require a higher MSRV. Enabling this
```

src/lib.rs
```diff
@@ -14,6 +14,16 @@
...
+//! feature on older compilers will NOT result in a compilation error, the newer
+//! language features will simply not be used.
+//!
+//! Right now, this feature only enables the use of
+//! `diagnostic::on_unimplemented` (Rust 1.78 or later) to provide better error
+//! messages.
```
daxpedda, line 25:
```suggestion
//! When compiling with Rust v1.78 or later, this feature enables better error messages for invalid methods on structs and enums.
```
Lets leave implementation details out of public documentation. Feel free to re-word, you get the gist.

daxpedda (changes requested):
Missing a changelog entry and a couple of nits, otherwise LGTM!

Great work!

RunDevelopment:
Okay, together with #4345, I think I addressed all review relevant comments now.

Cargo.toml
```diff
@@ -22,12 +22,19 @@ features = ["serde-serialize"]
...
 serde-serialize = ["serde", "serde_json", "std"]
 spans = []
 std = ["wasm-bindgen-macro/std", "once_cell/std"]
 
+# Opt-in for Rust language features that require a higher MSRV.
+#
+# The current rustc version is detected at compile-time, so enabling this
+# feature for older compilers will NOT result in a compilation error. Instead,
+# any unsupported language feature will not be used.
+msrv = ["rustversion"]
```
daxpedda, line 36: When https://github.com/rustwasm/wasm-bindgen/pull/4345 is merged, add this crate feature to these crates as well and then add it to the test matrix.

daxpedda: Done.

daxpedda (approved):

RunDevelopment:
Thank you for the help @daxpedda!

PR #4280, opened 2024-11-19
Improved variable names in generated JS

RunDevelopment:
This PR implements a change I wanted to do for a while. The bindings defined JS variables inconsistently, using `const`, `let`, and `var` all over the place. As it turns out, most variables can be declared as `const` and everything else can be `var`. So I create a simple API for creating variables. This API also takes care of resolving naming conflicts, so we'll never have to worry about temporary variables conflicting with argument names ever again. In the future, this system may also be used to resolve other naming conflicts, but I kept it simple for now.

For those needing a refresher on `var`, `let`, and `const` in JS:

- `const name = value`: A block-scoped variable that cannot be reassigned.
- `let name = value`: A block-scoped variable that can be reassigned.
- `var name = value`: A function-scoped variable that can be reassigned.

Changes:

- New API for `JsBinding` that handles the creating of variables. The new function automatically resolves name conflicts and ensure that temporary variable names never conflict with parameter names. The main functions of the new API are:
    - `alias(name, expr)`: This creates a JS `const` variable. The variable is immutable.
    - `var(name, initial)`: This creates a JS `var` variable. The variable is mutable AND visible in the `finally` block.
    - `guarnatee_alias`: Same as alias, but panics if the name is taken. This is necessary, because one instruction assumes that a variable called `retptr` has been created, so we can't rename that variable.
- Use the new API everywhere. All variables are now created with the new API.


RunDevelopment:
Please merge #4281 before this PR, because it includes code gen for most of the binding instructions this PR changes.

crates/cli-support/src/js/binding.rs
```diff
@@ -448,6 +452,62 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
...
+    /// Creates a new variable with the given name and returns the name.
+    ///
+    /// If the name is already taken, this function panics.
+    pub fn guarantee_alias(&mut self, name: &str, expression: String) -> String {
+        let can_use = self.used_identifiers.insert(name.to_string());
+        assert!(
+            can_use,
+            "identifier {} already in use in {}",
+            name, self.debug_name
+        );
```
daxpedda, line 469: Lets `bail!` instead.

crates/cli-support/src/js/binding.rs
```diff
@@ -819,12 +863,12 @@ fn instruction(
...
         Instruction::Retptr { size } => {
             js.cx.inject_stack_pointer_shim()?;
-            js.prelude(&format!(
-                "const retptr = wasm.__wbindgen_add_to_stack_pointer(-{});",
-                size
-            ));
+            let return_ptr = js.guarantee_alias(
+                "retptr",
+                format!("wasm.__wbindgen_add_to_stack_pointer(-{size})"),
+            );
```
daxpedda, line 869: Any idea why this just has to be called `retptr`?

daxpedda (changes requested):
I don't know what to think of the change from `let` to `var`. I always read to not use `var`. After some research, it also seems to be the recommendation I find most. Personally, `var` make it just harder to read code compared to `let`. Is there some guidance somewhere we could follow?

The `deferred` stuff makes sense to me to use `var` on, but on anything else I would prefer to stick with `let`, `const` taking precedence over both ofc.

crates/cli-support/src/js/binding.rs
RunDevelopment: Because `Instruction::LoadRetptr` has `retprt` hard-coded. `LoadRetptr` can't access the return pointer variable through the stack, so it's hard-coded.

I added a comment explaining this.

crates/cli-support/src/js/binding.rs
```diff
@@ -661,11 +706,10 @@ fn instruction(
                 // If the call is deferred, the arguments to the function still need to be
                 // accessible in the `finally` block, so we declare variables to hold the args
                 // outside of the try-finally block and then set those to the args.
-                for (i, arg) in js.stack[js.stack.len() - params..].iter().enumerate() {
-                    let name = format!("deferred{tmp}_{i}");
-                    writeln!(js.pre_try, "let {name};").unwrap();
-                    writeln!(js.prelude, "{name} = {arg};").unwrap();
-                    args.push(name);
+                let mut slice = Vec::new();
+                slice.extend(js.stack[js.stack.len() - params..].iter().cloned());
```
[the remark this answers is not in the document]
RunDevelopment: Yeah, probably not.

RunDevelopment:
The reason I didn't use `let` is to minimize the API of `JsBuilder`. 

The only difference between `var` and `let` inside function bodies is that `var` is function-scoped while `let` is block-scoped. Both can be re-assigned, but only `var` can be accessed from anywhere within the function. So `let` is strictly less powerful than `var`. That power is also the reason `let` was added to JS. Turns out, people make coding mistakes and `var` makes it easy to use a variable before it is defined. Example:

```js
try {
  throw new Error();
  var foo = 42;
} catch {
  console.log(foo);
}
```

Accessing `foo` inside the `catch` block before it has been assigned a value is valid and will log `undefined`. Did you expect this? If no, then you now know the reason people say you shouldn't use `var`. People are very used to block-scoped variables from other programming languages, and so `var` goes against their intuition.

Btw the reason the above logs `undefined` is that `var foo` is not block-scoped. So the above is executed by JS like this:
```js
let foo = undefined;
try {
  throw new Error();
  foo = 42;
} catch {
  console.log(foo);
}
```

Now the behavior is a bit more clear, right?

Why did I use `var` everywhere, then? Again, to minimize the API, and because most of the reasons people recommend `let` over `var` don't apply to us. The code is generated, so human error in writing code doesn't really apply, and scoping rules of `var` are actually useful to us because they make code gen a bit easier.

I just don't see the point in adding support for `let` when `var` already does everything `let` can do and more.

daxpedda:
> Now the behavior is a bit more clear, right?

I was already aware of all that. AFAIU to me it all boils down to scoping and hoisting vs TDZ. My personal preference is `let` because I find function level scoping horrible to read and if I've had to choose between temporal dead zoning and hoisting I would pick TDZ.

> Why did I use `var` everywhere, then? Again, to minimize the API, and because most of the reasons people recommend `let` over `var` don't apply to us. The code is generated, so human error in writing code doesn't really apply, and scoping rules of `var` are actually useful to us because they make code gen a bit easier.

I agree that we should prioritize WBG over the output. But the readability of the generated output is not worth nothing. E.g. removing the complexity around `deferred` in the implementation is a plus in my book. But the cost of using `let` in all the other cases seems incredibly small to me especially because it is very self-contained with the new changes you made. I would definitely not like to decrease the readability of the code unless the tradeoff makes sense, even if its not a priority.

So unless there are other reasons to use `var`, lets stick with `let`.

RunDevelopment:
Done. I used `let` wherever possible.

I also added reference tests for closures, because they had some code gen unique to them.

RunDevelopment:
I'm going insane. Code gen for closures isn't consistent between compilers. I'm not having another `raw.js`. Bye bye closure reference tests.

crates/cli-support/src/js/binding.rs
```diff
@@ -1291,85 +1311,71 @@ fn instruction(
...
             let len = js.pop();
             let ptr = js.pop();
             let f = js.cx.expose_get_vector_from_wasm(kind.clone(), *mem)?;
-            let i = js.tmp();
             let free = js.cx.export_name_of(*free);
-            js.prelude(&format!("let v{};", i));
+            let val = js.r#let("v", "undefined".to_string());
             js.prelude(&format!("if ({} !== 0) {{", ptr));
-            js.prelude(&format!("v{} = {}({}, {}).slice();", i, f, ptr, len));
+            js.prelude(&format!("{val} = {f}({ptr}, {len}).slice();"));
```
daxpedda, line 1372: Shouldn't we just be able to use `var` here?

daxpedda (changes requested):

crates/cli-support/src/js/binding.rs
RunDevelopment: With the introduction of `let`, the purpose of `var` has now become to declare variables as accessible from within the `finally` block.

The goal is to generate correct JS code, not code golfing.

daxpedda: In lack of a proper response: please use `var` here.

PR #4281, opened 2024-11-19
Add test for most data conversions

RunDevelopment:
This PR adds the echo test from #4188. This test includes simple functions for most data types WBG supports, so it essentially tests the code gen converting those data types to and from WASM.


daxpedda (approved):
Great!

Just rebased and will probably regenerate if it fails.

PR #4282, opened 2024-11-19
Use private constructors in type definitions of structs without an exported constructor

RunDevelopment:
## Motivation

A simple empty struct was previously typed as follows:

```rs
#[wasm_bindgen]
struct Foo;
```
```ts
export class Foo {
  free(): void;
}
```

This typing is incorrect. In JavaScript, a missing constructor is equivalent to an empty constructor. <details><summary>Example:</summary>

```js
// omitting the constructor...
class Example { }

// ... is the same as this
class Example {
  constructor() {}
}
```

</details> 

So the above type for `Foo` is actually just a short form for this:

```ts
export class Foo {
  constructor();
  free(): void;
}
```

Obviously, this is not correct. The `Foo` class is not supported to be instantiate-able using the class constructor. In debug mode, WBG even generates a constructor that always throws.

## Changes in this PR

This PR solves this problem by adding a `private constructor` for structs with a `#[wasm_bindgen(constructor)]`. E.g. the above `Foo` struct would get this type definitions:

```ts
export class Foo {
  private constructor();
  free(): void;
}
```

A private constructor in TS declares the constructor as inaccessible to anyone besides the class itself. In particular, a private constructor prevents users in TypeScript from:

1. Instantiating the class directly. E.g. `new Foo()`.
2. Making derived classes. E.g. `class Bar extends Foo {}`.
3. Using `Foo` in generic functions that require a constructor.

See [this TS playground for proof](https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEkoGdnwGIHtPwN4Ch4j4AHGASwDcoAXBMTAO2RpgFcwbMYAKASgDchYgDM4IfgC54lTOWBCAvvnwB6VfACMAOkRRGjTDXjlmNfTXK0EWTPgZn4I6bfgBeeIxAB3DNn5CItpiIBKCKurwAEy6YPqGxiAAHnSMwH52SKjwAEKw8MmpwGiuuPDKahoAzLHxRvBsyDbYJozwAOYgXhRgiEx0KWjeABYgcPBQfWbsnNwmaHAAjmzkcMD4ImyMnORMiHDWADwAKgUpXcV4nj5SE4wAnuUAfDwA1tCo0sd80gCSZvowCBjvcSCATk88MIiHAaGwYG0vL53ihkAF8MowAc6DxbHwgA).

With those changes, the type definitions now accurately represent the correct usage of class `Foo` and correctly cause errors on incorrect usage.

## Is this a breaking change?

Probably not. 

Since this is a type-only change, nothing will break at runtime (more than it already is). Users using the implicitly defined constructor most likely isn't intended by WBG since WBG even generates a constructor that always throws in debug mode to prevent exactly this.

So this PR will at most cause new TS compiler errors for incorrect code. From that perspective, I would argue that this PR is not a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,9 @@
 * Fixed imports for functions using `this` and late binding.
   [#4225](https://github.com/rustwasm/wasm-bindgen/pull/4225)
 
+* Don't expose non-functioning implicit constructors to classes when none are provided.
+  [#4282](https://github.com/rustwasm/wasm-bindgen/pull/4282)
+
 --------------------------------------------------------------------------------
 
 ## [0.2.95](https://github.com/rustwasm/wasm-bindgen/compare/0.2.94...0.2.95)
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1016,14 +1016,19 @@ __wbg_set_wasm(wasm);"
         let mut dst = format!("class {} {{\n", name);
         let mut ts_dst = format!("export {}", dst);
 
-        if self.config.debug && !class.has_constructor {
-            dst.push_str(
-                "
-                    constructor() {
-                        throw new Error('cannot invoke `new` directly');
-                    }
-                ",
-            );
+        if !class.has_constructor {
+            // declare the constructor as private to prevent direct instantiation
+            ts_dst.push_str("  private constructor();\n");
+
+            if self.config.debug {
+                dst.push_str(
+                    "
+                        constructor() {
+                            throw new Error('cannot invoke `new` directly');
+                        }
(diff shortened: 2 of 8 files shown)


RunDevelopment:
So that's cool. Turns out, 2 of our TS tests used the unintended default constructors.

daxpedda (changes requested):
Nice!

Fixing something that caused bugs when using it is not a breaking change in my book.

I was thinking that it might be reasonable to implement default constructors for unit structs, which would also mimic how it works in Rust. Something to think about for a follow-up.

Missing a changelog entry.
And please rebase as well, quite a lot has changed recently.

daxpedda (approved):

PR #4297, opened 2024-11-28
transition away from deprecated setters internally

nolanderc:
The old builder style setters are now deprecated, favoring the new `.set_field(...)` over the old `.field(...)`. We should make use of the new setters internall.

In `wgpu`, we vendor the WebGPU bindings from this crate into `wgpu` itself. This triggers the deprecation warnings, and upgrading to the latest `main`, we see about 100 of them strewn throughout the generated bindings. This change aims to remove those warnings.


daxpedda (approved):
Ha, nice catch!

PR #4299, opened 2024-11-29
Removed a few `#[allow(dead_code)]`

RunDevelopment:
Changes:

- Removed unused `nargs` field in `AuxImport::Closure`. The field wasn't used because the generated JS code used `...args` to capture any number of arguments. If this information is necessary in the future, `AuxImport::Closure` already includes the adapter ID of the closure function and the number of args can be retrieved from that.
- Removed unused `hole` field in `Instruction::OptionWasmToStringEnum`. The bindings have a comment explaining how the hole is handled implicitly as an "invalid" value, so it doesn't need to know the actual value of the hole.
- Changed how 8/16/32/64 bit ints are converted.

So about the integer conversion: Instead of removing the unused fields, I realized that `Instruction::IntToWasm` and `Instruction::WasmToInt` were kind of busted. The main issue is that they not only carried redundant data, but could also carry invalid data.

To fix this, I first split both into 64 and 32 variants. This immediately makes the JS -> WASM conversions nops. The WASM -> JS directions are also simpler, because they just need a single `bool` of data to select the correct conversion. 

This also fixes a minor bug in `representable_without_js_glue`. The function previously incorrectly determined that `WASM -> JS u64` and `WASM -> JS NonNull` did not require glue code. This was trivial to fix with the new representation.

---

This PR should not affect the output of the CLI in any way.


daxpedda (approved):
LGTM.
Will look into the other unused fields.

PR #4300, opened 2024-11-29
Remove dead code

daxpedda:
The code changes here should most definitely not cause any observable changes to runtime behavior.
However, instead they unfortunately cause memory allocation errors (for more info see CI), presumably caused by some UB somewhere.


daxpedda:
Figured it out: I removed a `bool` parameter from a function that had 2 `bool` parameters and deleted the wrong `bool` at the call site. :facepalm:

PR #4301, opened 2024-11-29
Add `no_export` attribute and export all enums by default

RunDevelopment:
This adds the `no_export` attribute we talked about. Enums with the `no_export` attribute will not be exported in JS and TS. This makes them consistent with `struct`s that are also exported by default and provides an option to opt out.

The `no_export` attribute is completely independent of the visibility of the Rust enums. E.g. an enum may be `pub` and non-exported, or private and exported. To make this consistent, string enums no longer use the enum's visibility to determine whether their type is exported.

Changes:
- Add `no_export` attribute.
- Changed behavior of private string enums. Their type will now be exported by default.
- Allow private numeric enums to be consistent with structs and string enums.
- Add `no_export` to all string enums in `web-sys`.

---

TODO:

- [ ] Numeric enums with `no_export` that are used in exported function are currently typed as non-exported TS enums in TS. This makes the exported functions with such enums are parameters a lot less useful, since they can't be invoked without either (1) TS compiler errors or (2) a type cast to `any` to work around the type checker.
- [ ] Documentation. We may want to think about the details here a bit more. I'll write the docs when we have everything figured out.
- [ ] Think about how users should deal with third-party crates with string enums that are now exported by default.


guybedford:
The `private` attribute added in https://github.com/wasm-bindgen/wasm-bindgen/pull/4788 covers this functionality now.

PR #4310, opened 2024-11-30
Smarter enhancement of JSDoc comments with a JSDoc parser

### LINKED ISSUE #1847: Support descriptions for JS/TS param/return comments

### LINKED ISSUE #4223: Synchronize type annotations of JSDoc comments and TypeScript to give users more control over types
### Motivation

JSDoc has `@param`, `@returns`, and `@type` annoations that have been used for years in the JS world to annotate the types of function parameters, return types, field types, and more. Right now, WASM bindgen automatically generates `@param` and `@return` annotations with types for functions (in JS #4187) or not with [`skip_jsdoc`](https://rustwasm.github.io/wasm-bindgen/reference/attributes/on-rust-exports/skip_jsdoc.html).

This currently creates the problems that the user-defined JSDoc type annotations are ignored in TS and that WASM-bindgen-generated annotations have to be explicitly disabled to not interfere with the user-defined ones.

### Proposed Solution

If the user provides a JSDoc comment for a function, parse the JSDoc comment and:

1. If a parameter or the return is not documented, generate an `@`-property with the type annotation for it.
2. If a parameter or the return is document, but has no type annotation, add the type annotation.
3. If a parameter or the return is document and has a type annotation, use the annotated type use the annotated type in both JSDoc and TS.

Example:

```rs
/// Documentation for the function.
/// @param b This gets its type inferred.
/// @param {100 | 200 | null} c Note that this parameter is *not* optional and does not accept `undefined`.
/// @returns {"success" | "failure"}
#[wasm_bindgen]
pub fn foo(a: &str, b: u32, c: Option<u32>) -> String { todo!() }
```
```ts
// bindings.js
/**
 * Documentation for the function.
 * @param {string} a
 * @param {number} b This gets its type inferred.
 * @param {100 | 200 | null} c Note that this parameter is *not* optional and does not accept `undefined`.
 * @returns {"success" | "failure"}
 */
export function foo(a, b, c) { ... }

// bindings.d.ts
/**
 * Documentation for the function.
 * @param b This gets its type inferred.
 * @param c Note that this parameter is *not* optional and does not accept `undefined`.
 */
export function foo(a: string, b: number, c: 100 | 200 | null): "success" | "failure";
```

The same idea of syncing types for can also be used for class fields and constants with `@type {...}`.

This solution not only solves the problem of making sure that typing is consistent between JSDoc and TS, it also:

1. Makes `skip_jsdoc` almost unnecessary, since generated annotations no longer interfere with user-defined ones.
2. Removes the possible error of users accidentally annotating the wrong type when they just wanted to document a parameter.
3. Allows users to define a custom type for every field, argument, return, and constant using JSDoc. 
 
So not only is this more convenient, ensures consistency, it also gives users more control. The last point is very important, because this fixes #1197, #1591, #1847, #2869, #3107, and #3953. 

I want to highlight here, that custom types are defined using a syntax already familiar to JS devs. Any JS dev that has used JSDoc for type annotations before will know what to do without us having to explain anything. Using JSDoc comments also has the advantage that type expressions are not limited by Rust syntax in any way.

### Complexity

I think the main challenge for implementing this will be to parse and interpret JSDoc. There is no guarantee that the JSDoc comments will be valid. JSDoc itself is also quite complex, so interpreting it will be tricky even in parsed form.

We might be able to use [swc's `jsdoc` crate](https://crates.io/crates/jsdoc) for something battle-tested. The main problem with this crate is that pulls in tons of dependencies, and it's entirely undocumented as well. If push comes to shove, we might have to write a simple parser ourselves. (I already worked on JSDoc syntax highlighter and formatter before, so I can do it if need be.)

### Backwards compatibility

Since this only affects comments and TS types, I'm not sure if this would be considered a breaking change. I think in most case, with would even be a fixing change, since the TS and JSDoc types finally match.

In any case, I think that invalid JSDoc comments or other problems (e.g. duplicate `@param` or `@return` tags) should not cause compiler errors but emit warnings. We can then decide to upgrade those warnings to errors in future releases.

### Additional Context

I'm willing to make a PR for this.

I also want to point out this approach of using JSDoc to define typing information naturally extends to other features as well. E.g. generics with `@template`, custom type definitions with `@typedef` and `@callback`, overloads with `@overload`, and more. By treating JSDoc comments as more than just comments, we have a path for supporting a lot of advanced typing features in a way that is intuitive to most JS/TS devs.

RunDevelopment:
Resolves #4223.
Resolves #1847.

This PR improves the type information is added to JSDoc comments. Instead of just adding `@param` and `@returns` tags to existing JSDoc comments, WBG will now parse the existing comment and add type information to existing tags (if any). This makes `skip_jsdoc` virtually unnecessary, since WBG won't generate tag that conflict with existing tags anymore. (Note that `skip_jsdoc` still has niche use cases.)

Example:

```rs
/// Some comment
/// @param a This is the first parameter.
#[wasm_bindgen]
pub fn foo(a: u32, b: &str) -> u32 { ... }
```
```js
/**
 * Some comment
 * @param {number} a This is the first parameter.
 * @param {string} b
 * @returns {number}
 */
export function foo(a, b) { ... }
```

This behavior can be turned off with `skip_jsdoc`

```rs
/// Some comment
/// @param a This is the first parameter.
#[wasm_bindgen]
pub fn foo(a: u32, b: &str) -> u32 { ... }
```
```js
/**
 * Some comment
 * @param a This is the first parameter.
 */
export function foo(a, b) { ... }
```

With the new smarter behavior, users documenting their code with standard JSDoc/TSDoc can now use `@param`/`@returns` tags without having WBG mess it up or having to using `skip_jsdoc` and duplicating the type info. WBG's JSDoc comment enhancement now works together with our users.

Moving forward, I believe that we should teach people to WBG will automatically add type information for them, and they should not manually add type info to `@param` and `@returns` tags at all. This has several advantages, one being that users don't get any ideas about trying to declare custom types as `@param {1 | 2 | 3} arg`, which does not affect the TS types. Once we have a way for users to declare custom TS type annotations, they will just work with the new JSDoc enhancements.

---

The main complexity of this PR is:

1. The new JSDoc parser. I made a custom JSDoc parser in around 600 LoC (including comments).
2. Functionality for modifying the parsed JSDoc AST. This is the logic for actually adding type info to existing JSDoc and is around 150 LoC (including comments).
3. Testing the parser. The parser needs to handle a lot of edge cases, so I added a new snapshot test for it.

The parser is written in a way that makes any string valid JSDoc. This can also be seen by the signature of the parsing function: `fn parse(comment: &str) -> JsDoc`. The doc comment on the function explains how this is done. In generally, I extensively documented the parser. If you have any question regarding how the parser works, the answer is probably in a comment.

For the tags that are supported, the implementation is fairly complete. The parser supports all sorts of mean constructs like multiple-line TS types, comments in TS types, default values, and argument paths (e.g. `@param {number} arg0.age`). So it's fairly complete for what it supports. See the snapshot test for many more examples.


crates/cli-support/src/js/jsdoc.rs
```diff
@@ -0,0 +1,970 @@
...
+        })
+    }
+}
+
+fn post_process_typescript_expression(expr: Option<String>) -> Option<String> {
+    expr.and_then(|e| if e.trim().is_empty() { None } else { Some(e) })
+}
+
+/// A version trim_start that ignores text direction.
+fn trim_left(s: &str) -> &str {
```
RunDevelopment, line 460: I made three functions for trimming, because I don't understand the text direction behavior of [`trim_start`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_start). JSDoc is inherently left to right, so I don't know how this will interact with Arabic text.

Am I just paranoid here? Should I just use the std functions?

daxpedda: The std documentation seems clear, even has some code examples. I don't think you are paranoid.

daxpedda (commented):
I can safely say I can't maintain this code. Just reviewing it so far has taken me a not-insignificant amount of time.

I've been thinking about it quite a lot by now and I can safely say I don't feel comfortable at all landing this.

Would love to land this if:
- The parser is in a maintained and trusted external crate. Just remembered [oxc](https://github.com/oxc-project/oxc), which gives me an orders of magnitude better impression so far than SWC. If you can do the research I'm happy to look into and make a decision on if a specific dependency is acceptable.
- It is opt-in, preferably via an attribute (alternatively I proposed just a line saying `wbg::jsdoc-start` or something like that). Again: I'm really not a fan of parsing doc comments by default. But I'm happy to discuss this point again and be convinced otherwise.

If this doesn't work out, the easy alternative is still to support third-party crates via a bunch of WBG attributes. Definitely the less appealing way.

RunDevelopment:
That's a shame.

> Just remembered [oxc](https://github.com/oxc-project/oxc), which gives me an orders of magnitude better impression so far than SWC. If you can do the research I'm happy to look into and make a decision on if a specific dependency is acceptable.

To repeat my previous findings: There are essentially 3 packages [on crates.io](https://crates.io/search?q=jsdoc) that need to be considered (all others aren't JSDoc parsers):

- [jsdoc](https://crates.io/crates/jsdoc/4.0.0). This is the swc dependency I was talking about before. The problem with it is that it pulls in 9 MB of transitive deps [according to libs.rs](https://lib.rs/crates/jsdoc).
- [tree-sitter-jsdoc](https://crates.io/crates/tree-sitter-jsdoc/0.23.2). This is a plugin for tree-sitter, so I don't understand the AST format and the crate requires a C compiler.
- [doctor](https://crates.io/crates/doctor). The AST format is too simplistic for our use case.

While `oxc` itself is a JS toolbox(?), I did miss that it also has a JSDoc parser. The reason I missed it is that the JSDoc isn't part of oxc's AST/parser crates, but [`oxc_semantic`](https://docs.rs/oxc_semantic/latest/oxc_semantic/struct.JSDoc.html), which deals with the semantic analysis of JS code. And befitting of that task, it pulls in an appropriate amount of transitive dependencies, since analyzing JS is a much harder task than just parsing a bit of JSDoc.

Anyway, the JSDoc parser itself seems [usable](https://docs.rs/oxc_semantic/latest/oxc_semantic/struct.JSDocTag.html#method.type_name_comment). I did some tests, and it's not 100% correct when parsing type expressions, but it should be good enough. The only issue is that their JSDoc AST is immutable, so we'll have to implement the modifying and printing ourselves.

So yeah, that's the state of JSDoc parsing in the Rust ecosystem. The choice is either `jsdoc` from swc or `oxc_semantic` from oxc. Nothing else meets our requirements. So please make a decision.

> It is opt-in, preferably via an attribute [...]

As I said before, the whole point is to change the default of WBG just adding `@param`/`@returns` tags blindly, causing it to potentially create nonsensical JSDoc. Put bluntly, the current default kinda sucks. So making it opt-in would be completely beside the point.

> If this doesn't work out, the easy alternative is still to support third-party crates via a bunch of WBG attributes. Definitely the less appealing way.

As I explained before, this is not possible. Only the CLI knows the final return/parameter types. No matter what attributes we may add, only the CLI has the necessary information to perform this task. There is no alternative.

daxpedda:
> So yeah, that's the state of JSDoc parsing in the Rust ecosystem. The choice is either `jsdoc` from swc or `oxc_semantic` from oxc. Nothing else meets our requirements. So please make a decision.

I've already made a decision on `swc` and its a no. 
`oxc_semantic` is too heave for us. Do you think it can be separated from `oxc_semantic` into its own crate so its more lightweight? Could you tell me more about what needs to be done on `wasm-bindgen`s side to make this work for us? If it defeats the point, can we contribute to `oxc` to make it work for us? 

> > It is opt-in, preferably via an attribute [...]
> 
> As I said before, the whole point is to change the default of WBG just adding `@param`/`@returns` tags blindly, causing it to potentially create nonsensical JSDoc. Put bluntly, the current default kinda sucks. So making it opt-in would be completely beside the point.

We have already discussed this, I deemed this a non-goal for WBG. As I said many times before: our target audience are developers who want to builds tools on top of `wasm-bindgen` for other end users. Not end users directly. Convenience will always be an important goal, but a secondary one.

In this case I have reasons I laid out already why I think this isn't an ideal situation to do by default. Which I find more important then convenience in this case.

If the point is not convenience, then I apparently missed the point.

> > If this doesn't work out, the easy alternative is still to support third-party crates via a bunch of WBG attributes. Definitely the less appealing way.
> 
> As I explained before, this is not possible. Only the CLI knows the final return/parameter types. No matter what attributes we may add, only the CLI has the necessary information to perform this task. There is no alternative.

The last time we have discussed this we established that this is possible via attributes. From your comment here I think there is a misunderstanding.

I'm talking about a WBG-CLI implementation. The JSDoc is just passed via attributes instead of parsing Rustdoc comments. A third-party crate can parse the JSDoc in the Rustdoc and fill in these attributes instead of WBG doing it. The difference is that attributes can be split up into attributes on arguments, a `js_doc_return` attribute on the function, and so on. This way WBG doesn't have to parse JSDoc because the attributes act as a form of AST already.

If I'm missing some specific interaction that can not be support this way then please elaborate.

daxpedda:
Replaced by #4394.

PR #4311, opened 2024-11-30
Use TypeScript to verify CLI reference output

RunDevelopment:
This PR adds a new CI test that uses the TypeScript compiler to verify that all JS and TS files in `cli/tests/references` are valid JS/TS. This ensures that invalid JS/TS will be caught even if human review misses some error.

Since I use the TS compiler to verify JS files, I also enabled type checking for them. So we will hopefully catch invalid glue code early now.

---

I'm still working out the kinks.


RunDevelopment:
Alright, I changed it to only check JS syntax for now. Fixing all the type errors in JS will require some more work, but that can be a follow-up PR.

For now, this PR just checks the types of `.d.ts` files and the syntax of `.js` files.

.github/workflows/main.yml
```diff
@@ -239,6 +239,22 @@ jobs:
...
 
+  # This checks that the output of the CLI is actually valid JavaScript and TypeScript
+  test_cli_reference_typescript:
+    name: Run CLI reference TypeScript check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: '20'
```
daxpedda, line 250: What is your thinking there?

I would either use the default one or preferably just the latest version.

.github/workflows/main.yml
```diff
@@ -239,6 +239,22 @@ jobs:
...
+    - run: npm i -g typescript
+    - run: npm i --save @types/node @types/deno
+    - name: Check TypeScript output
+      run: tsc --noEmit --skipLibCheck --lib esnext,dom $(echo crates/cli/tests/reference/*.d.ts)
```
daxpedda, line 254: Whats going on with the `echo` call here?

RunDevelopment: `tsc` doesn't support glob patterns. To pass multiple files to `tsc`, you either have to add a `tsconfig.json` (which isn't possible, because that would make it impossible to check `.js` and `.d.ts` separately) or pass in all files separated by spaces. The `echo` call does exactly that. It runs the glob pattern and returns the paths of all matching files separated by spaces.

.github/workflows/main.yml
RunDevelopment: I just copied the code from another config and 20 seemed like a fine version.

But we can go to latest LTS. I don't want to use the actual latest release, because nodejs has had multiple versions in the past that flat out broke certain npm packages, and I don't want to deal with that here.

daxpedda: Sounds fine.

daxpedda (approved):

daxpedda:
This is great btw, thank you!

PR #4316, opened 2024-12-05
Add support for number slices of type `MaybeUninit<T>`

daxpedda:
This PR adds support for numeric slice types with `MaybeUninit<T>`. While, generally in Rust, unsafety requirements have to be kept by the user, this isn't unsafe in Wasm to begin with, so there is no concern in this regard.

The use case here is interfacing with cross-platform Rust code where using this type probably made much more sense.


RunDevelopment (commented):
Please document the behavior of reading `MaybeUninit<T>` values in JS. I think a small paragraph saying that it's not UB, but that the value is unspecified would be enough.

Also, please add a CLI reference test, so we can see the generated JS and TS.

guide/src/reference/types/number-slices.md
```diff
@@ -1,9 +1,13 @@
-# Number Slices: `[u8]`, `[i8]`, `[u16]`, `[i16]`, `[u32]`, `[i32]`, `[u64]`, `[i64]`, `[f32]`, and `[f64]`
+# Number Slices: `[u8]`, `[i8]`, `[u16]`, `[i16]`, `[u32]`, `[i32]`, `[u64]`, `[i64]`, `[f32]`, `[f64]`, `[MaybeUninit<u8>]`, `[MaybeUninit<i8>]`, `[MaybeUninit<u16>]`, `[MaybeUninit<i16>]`, `[MaybeUninit<u32>]`, `[MaybeUninit<i32>]`, `[MaybeUninit<u64>]`, `[MaybeUninit<i64>]`, `[MaybeUninit<f32>]`, and `[MaybeUninit<f64>]`
 
 | `T` parameter | `&T` parameter | `&mut T` parameter | `T` return value | `Option<&T>` parameter | `Option<T>` return value | JavaScript representation |
 |:---:|:---:|:---:|:---:|:---:|:---:|:---:|
 | No | Yes | Yes | No | No | No | A JavaScript `TypedArray` view of the Wasm memory for the boxed slice of the appropriate type (`Int32Array`, `Uint8Array`, etc) |
 
+> **Note:** While reading from an uninitialized `MaybeUninit` is unsafe in
+> Rust, it is not in Wasm. Though they might contain unspecified values.
```
RunDevelopment, line 8: Minor corrections:

1. Reading uninitialized memory in Rust isn't just unsafe, it's UB.
2. WASM doesn't do the reading, the JS users do.

```suggestion
> **Note:** While reading from an uninitialized `MaybeUninit` is UB in Rust, it is safe in JavaScript. Though they might contain unspecified values.
```

RunDevelopment (commented):
One last comment, then gtg 👍

guide/src/reference/types/number-slices.md
RunDevelopment: You're right, that's a good idea. I think it would be good to say that data coming from JS is always initialized, so taking a `&[MaybeUninit<u8>]`, while supported, typically doesn't make sense.

daxpedda: Okay, after a lot of different tries I arrived at the current state. I was trying not to explain Rust or Wasm but stay concise to what users need to know only for this feature and nothing else.

daxpedda: Let me know if you want me to improve anything still, we can do that in a follow-up.

PR #4319, opened 2024-12-05
Support `undefined` static imports with `Option`

daxpedda:
```rust
extern "C" {
    type Crypto;
    #[wasm_bindgen(thread_local_v2, js_name = crypto)]
    static CRYPTO: Option<Crypto>;
}
```
This currently throws an error in JS if `crypto` is `undefined`.

This change would set `CRYPTO` to `None` instead.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -3274,7 +3274,13 @@ __wbg_set_wasm(wasm);"
                 if *optional {
                     writeln!(
                         prelude,
-                        "const result = typeof {js} === 'undefined' ? null : {js};"
+                        "\
+                        let result;
+                        try {{
+                            result = {js};
+                        }} catch (_) {{
+                            result = null;
+                        }}",
                     )
                     .unwrap();
                     Ok("result".to_owned())
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2586,7 +2586,7 @@ __wbg_set_wasm(wasm);"
 
         // After we've got an actual name handle field projections
         if optional {
-            name = format!("typeof {name} === 'undefined' ? null: {name}");
+            name = format!("typeof {name} === 'undefined' ? null : {name}");
 
             for field in import.fields.iter() {
                 name.push_str("?.");
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2578,6 +2578,30 @@ __wbg_set_wasm(wasm);"
         Ok(name)
     }
 
+    fn import_static(&mut self, import: &JsImport, optional: bool) -> Result<String, Error> {
+        let mut name = self.import_name(&JsImport {
+            name: import.name.clone(),
+            fields: Vec::new(),
+        })?;
+
+        // After we've got an actual name handle field projections
+        if optional {
+            name = format!("typeof {name} === 'undefined' ? null: {name}");
+
+            for field in import.fields.iter() {
+                name.push_str("?.");
+                name.push_str(field);
+            }
+        } else {
+            for field in import.fields.iter() {
+                name.push('.');
+                name.push_str(field);
+            }
+        }
+
+        Ok(name)
+    }
+
     /// If a start function is present, it removes it from the `start` section
     /// of the Wasm module and then moves it to an exported function, named
     /// `__wbindgen_start`.
@@ -3269,24 +3293,7 @@ __wbg_set_wasm(wasm);"
                 assert!(kind == AdapterJsImportKind::Normal);
                 assert!(!variadic);
                 assert_eq!(args.len(), 0);
-                let js = self.import_name(js)?;
-
-                if *optional {
-                    writeln!(
-                        prelude,
-                        "\
-                        let result;
-                        try {{
-                            result = {js};
-                        }} catch (_) {{
-                            result = null;
-                        }}",
-                    )
-                    .unwrap();
-                    Ok("result".to_owned())
-                } else {
-                    Ok(js)
-                }
+                self.import_static(js, *optional)
             }
 
             AuxImport::String(string) => {
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2730,7 +2730,7 @@ __wbg_set_wasm(wasm);"
                 | AuxImport::Value(AuxValue::Setter(js, ..))
                 | AuxImport::ValueWithThis(js, ..)
                 | AuxImport::Instanceof(js)
-                | AuxImport::Static(js)
+                | AuxImport::Static { js, .. }
                 | AuxImport::StructuralClassGetter(js, ..)
                 | AuxImport::StructuralClassSetter(js, ..)
                 | AuxImport::IndexingGetterOfClass(js)
@@ -3265,11 +3265,22 @@ __wbg_set_wasm(wasm);"
                 Ok("result".to_owned())
             }
 
-            AuxImport::Static(js) => {
+            AuxImport::Static { js, optional } => {
                 assert!(kind == AdapterJsImportKind::Normal);
                 assert!(!variadic);
                 assert_eq!(args.len(), 0);
-                self.import_name(js)
+                let js = self.import_name(js)?;
+
+                if *optional {
+                    writeln!(
+                        prelude,
+                        "const result = typeof {js} === 'undefined' ? null : {js};"
+                    )
+                    .unwrap();
+                    Ok("result".to_owned())
+                } else {
+                    Ok(js)
+                }
             }
 
             AuxImport::String(string) => {
--- a/crates/cli/tests/reference/static.js
+++ b/crates/cli/tests/reference/static.js
@@ -1,0 +1,77 @@
+let wasm;
+export function __wbg_set_wasm(val) {
+    wasm = val;
+}
+
+
+function isLikeNone(x) {
+    return x === undefined || x === null;
+}
+
+function addToExternrefTable0(obj) {
+    const idx = wasm.__externref_table_alloc();
+    wasm.__wbindgen_export_1.set(idx, obj);
+    return idx;
+}
+
+const lTextDecoder = typeof TextDecoder === 'undefined' ? (0, module.require)('util').TextDecoder : TextDecoder;
+
+let cachedTextDecoder = new lTextDecoder('utf-8', { ignoreBOM: true, fatal: true });
+
+cachedTextDecoder.decode();
+
+let cachedUint8ArrayMemory0 = null;
+
+function getUint8ArrayMemory0() {
+    if (cachedUint8ArrayMemory0 === null || cachedUint8ArrayMemory0.byteLength === 0) {
+        cachedUint8ArrayMemory0 = new Uint8Array(wasm.memory.buffer);
+    }
+    return cachedUint8ArrayMemory0;
+}
+
+function getStringFromWasm0(ptr, len) {
+    ptr = ptr >>> 0;
+    return cachedTextDecoder.decode(getUint8ArrayMemory0().subarray(ptr, ptr + len));
+}
+
+export function exported() {
+    wasm.exported();
+}
+
+export function __wbg_static_accessor_NAMESPACE_OPTIONAL_c9a4344c544120f4() {
+    const result = typeof test.NAMESPACE_OPTIONAL === 'undefined' ? null : test.NAMESPACE_OPTIONAL;
+    const ret = result;
+    return isLikeNone(ret) ? 0 : addToExternrefTable0(ret);
+};
+
+export function __wbg_static_accessor_NAMESPACE_PLAIN_784c8d7f5bbac62a() {
+    const ret = test.NAMESPACE_PLAIN;
+    return ret;
+};
+
+export function __wbg_static_accessor_OPTIONAL_ade71b6402851d0c() {
+    const result = typeof OPTIONAL === 'undefined' ? null : OPTIONAL;
+    const ret = result;
+    return isLikeNone(ret) ? 0 : addToExternrefTable0(ret);
+};
+
+export function __wbg_static_accessor_PLAIN_c0f08eb2f0db194c() {
+    const ret = PLAIN;
+    return ret;
+};
+
+export function __wbindgen_init_externref_table() {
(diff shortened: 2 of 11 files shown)


RunDevelopment:
The title of this PR doesn't make sense to me. If the global `crypto` variable was set to `undefined`, then wouldn't the imported value be `JsValue::UNDEFINED` (or whatever the constant is called)? 

I think what you really mean is that it now supports undeclared variables, right? So for your example, `crypto` might not be defined by the browser/runtime, so just accessing it might give a `ReferenceError` at runtime. If so, then this is very different from the global variable being set to `undefined`. Please change the changelog entry if that's what you meant.

Also, please document this feature. Its behavior is not at all obvious. E.g. WBG doesn't support `Option<JsValue>` in functions at all, so it's surprising that it's allowed here. Globally defined variables set to `undefined` or `null` are also handled differently from just `static FOO: JsValue`.

daxpedda:
> The title of this PR doesn't make sense to me. If the global `crypto` variable was set to `undefined`, then wouldn't the imported value be `JsValue::UNDEFINED` (or whatever the constant is called)?

It would not, that's why `Option<JsValue>` isn't supported in the first place. I fixed OP.

> I think what you really mean is that it now supports undeclared variables, right? So for your example, `crypto` might not be defined by the browser/runtime, so just accessing it might give a `ReferenceError` at runtime. If so, then this is very different from the global variable being set to `undefined`. Please change the changelog entry if that's what you meant.

Right, the distinction between `undefined` and undeclared wasn't quite clear to me. It should be noted that in either case it the output would be `None`.

> Also, please document this feature. Its behavior is not at all obvious. E.g. WBG doesn't support `Option<JsValue>` in functions at all, so it's surprising that it's allowed here. Globally defined variables set to `undefined` or `null` are also handled differently from just `static FOO: JsValue`.

I might consider expanding this feature to cover `JsValue` as well, that would make sense.

daxpedda:
> > Also, please document this feature. Its behavior is not at all obvious. E.g. WBG doesn't support `Option<JsValue>` in functions at all, so it's surprising that it's allowed here. Globally defined variables set to `undefined` or `null` are also handled differently from just `static FOO: JsValue`.
> 
> I might consider expanding this feature to cover `JsValue` as well, that would make sense.

Or not actually? As you said, this feature is not about `undefined`, its about undeclared.

daxpedda:
Adjusted and added documentation.

So to be clear: this PR does not add support for `Option<JsValue>`!

crates/cli/tests/reference/static.js
```diff
@@ -0,0 +1,77 @@
...
+    ptr = ptr >>> 0;
+    return cachedTextDecoder.decode(getUint8ArrayMemory0().subarray(ptr, ptr + len));
+}
+
+export function exported() {
+    wasm.exported();
+}
+
+export function __wbg_static_accessor_NAMESPACE_OPTIONAL_c9a4344c544120f4() {
+    const result = typeof test.NAMESPACE_OPTIONAL === 'undefined' ? null : test.NAMESPACE_OPTIONAL;
```
RunDevelopment, line 42: This isn't correct. If the `test` namespace is not declared, this will still throw a `ReferenceError` at runtime. The code gen needs to be `typeof test === 'undefined' ? null : test.NAMESPACE_OPTIONAL`.

This also raises the question of how nested namespaces are handled. E.g. let's say I want to import the constant `A.B.C.D` (so the namespace is `A` -> `B` -> `C`). If `A` is undeclared, then the Rust value should be `None`, that's fine. But what if `B` is `undefined`? Then accessing `A.B.C.D` would result in a `TypeError` like this:

```
TypeError: Cannot read properties of undefined (reading 'C')
```

The easiest way around this would be to use [optional chaining `?.`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (CAREFUL! `?.` does not mean the same thing as it does in Rust). Like this:

```js
const result = typeof A === 'undefined' ? null : A?.B?.C?.D;
```

However... optional chaining is relatively new (at least newer than WASM), so we can't use it. E.g. for chrome, optional chaining is supported in chrome >=80, but WBG currently supports chrome >=74 (I'm going by the [sign extension operator](https://webassembly.org/features/) here).

So we probably need to implement the optional chaining ourselves here. While not perfect, we could implement it as follows:

```js
const result = typeof A === 'undefined' || A == null || A.B == null || A.B.C == null ? null : A.B.C.D;
```

Note: In JS, `expr == null` is the same as `expr === undefined || expr === null`. It's a quick way to check for both `null` and `undefined`.

I'm saying "not perfect", because we need to access the namespaces `B` and `C` multiple times. Since JS has getters (and proxies), the value returned by `A.B` might change between accesses. So for 100% correctness, we would have to do the same thing as transpilers and store the value of `A.B` is a temporary variable. Transpilers produce code like this:

```js
let t; // temporary variable
const result = typeof A === 'undefined' || (t=A) == null || (t=t.B) == null || (t=t.C) == null ? null : t.D;
```
This is 100% equivalent to 

```js
const result = typeof A === 'undefined' ? null : A?.B?.C?.D;
```

but needs more logic on the Rust side to generate.

guide/src/reference/static-js-objects.md
```diff
@@ -65,6 +65,20 @@ extern "C" {
 }
 ```
 
+## Undeclared
+
+When accessing an in JS undeclared value, it will throw in JS. This can be accounted for by using `Option`.
```
RunDevelopment, line 70:
```suggestion
## Undeclared vara

When accessing an in JS undeclared value, it will throw in JS. This can be accounted for by using `Option`.
```

crates/cli/tests/reference/static.js
daxpedda: I see. How about just using `try { .. } catch() { .. }`?

I should add a runtime test as well then.

RunDevelopment: Depends. If `A.B` is a getter and throws its own error, should we ignore the error?

daxpedda: > Depends. If `A.B` is a getter and throws its own error, should we ignore the error?

Probably not? I'm considering just not supporting namespaces in this case. The main use case, importing globals, would remain covered ...

Let me see how bad the implementation would be.

daxpedda: Its not terrible actually.
On that note, do we want to even support getters here? `static` is thread-local cached, so if this is a getter it doesn't make sense in the first place.

So after actually implementing it ... I'm currently tending towards just not. The documentation states "static objects", not getters.

daxpedda: Pushed the fix and added some runtime tests.
Let me know what you think on the getter problem.

RunDevelopment: IMO, this is not a question of whether we support getters. People are going to assume that we do, because they won't even think about getters when they need the constant `A.B`. 

So I think it's fine to say that we don't support getters that throw in that it will cause an error, but I don't think it's fine to hide the error. Failing silently here is a problem IMO, because it hides a real bug in user code.

daxpedda: I do consider our checks on bindings best effort only. From the perspective of Rust this is actually unsafe, which is why in [edition 2024 it will be written `unsafe extern "C" { ... }`](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html).

Quoting:
> Adding the `unsafe` keyword helps to emphasize that it is the responsibility of the author of the `extern` block to ensure that the signatures are correct. If the signatures are not correct, then it may result in undefined behavior.

But if we can detect that these are getters, then we should do that. I will add some documentation to that effect as well.

RunDevelopment: > I do consider our checks on bindings best effort only.

Then we should do our best :)

`unsafe extern` isn't really relevant here, because it's not like users can use statics to express the correct signature for throwing getters. They have to use `getter` functions. (Ignore that they are broken right now.)

My point wasn't that we should support throwing getters in statics. I don't think we should. My point was that the way in which we do not support them should be better. If a getter throws for a static, it's a bug in user code either way. The question is just whether we should loudly panic and tell the user about it, or silently pretend nothing is wrong.

Note that I am assuming that JS errors thrown in the JS glue code for statics currently causes a panic. Is that correct?

daxpedda: To clarify: we don't support getters at all in `static`. We cache the value thread-locally, so a using `static` with a getter is a footgun already, it was already like that.

> Note that I am assuming that JS errors thrown in the JS glue code for statics currently causes a panic. Is that correct?

No, the throw isn't caught. It will cause re-entrancy and all sort of issues, but that's a whole different rabbit-hole.

---

> > I do consider our checks on bindings best effort only.
> 
> Then we should do our best :)

What are you suggesting? *Just* detecting throwing getters is an improvement, but only a very minor one. It tells users that they are using a getter and they shouldn't. In *every other case* it will not tell users anything. I'm not happy about the code/benefit trade-off.

So I would prefer if we can detect getters to begin with. Is that possible?

daxpedda: Thinking about this a bit more, I consider ignoring throwing getters an improvement actually. Because causing re-entrency is pretty bad imo.

So to actually address this, we need to do the following: catch the error, return it to Rust and panic there.

Still: ideally, we detect if its a getter in the first place.

RunDevelopment: > So I would prefer if we can detect getters to begin with. Is that possible?

Yes, it's possible with [`Object.getOwnPropertyDescriptor`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor). 

However, it's also highly undesirable. Sometimes people use static properties, because the constant is computed and/or to cache the computation. In general, you really shouldn't differentiate between plain old property access and getters, because no one in JS land expects you to do that.

Just because something is implemented as a getter, doesn't mean that its value changes.

> What are you suggesting?

IMO, the ideal behavior would be to cause a panic in Rust. So it's unfortunate that there isn't any kind of machinery for it in place already...

My main problem with ignoring the exception is that I already know that people will start relying on it. So if we allow this behavior now, there will never be a chance to change it. So I would even prefer whatever issues not catching the exception is causing right now. At least then we get to talk about fixing it.

Also, catching and ignoring errors would be inconsistent with old plain (non-`Option<T>`) statics. So that's fun.

daxpedda: > > So I would prefer if we can detect getters to begin with. Is that possible?
> 
> Yes, it's possible with [`Object.getOwnPropertyDescriptor`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor).
> 
> However, it's also highly undesirable. Sometimes people use static properties, because the constant is computed and/or to cache the computation. In general, you really shouldn't differentiate between plain old property access and getters, because no one in JS land expects you to do that.

Well, that certainly puts a wrench into my thinking ...

> > What are you suggesting?
> 
> IMO, the ideal behavior would be to cause a panic in Rust. So it's unfortunate that there isn't any kind of machinery for it in place already...

There is: `catch`. The integration is just more work I would have preferred to leave for later.

There also more design concerns with this: currently this problem concerns all bindings, not just statics. E.g. if you import a function that can throw and you don't mark it with `catch`, it will also throw in JS and may cause unexpected reentrancy into Rust.

> My main problem with ignoring the exception is that I already know that people will start relying on it. So if we allow this behavior now, there will never be a chance to change it. So I would even prefer whatever issues not catching the exception is causing right now. At least then we get to talk about fixing it.
>
> Also, catching and ignoring errors would be inconsistent with old plain (non-`Option<T>`) statics. So that's fun.

Puh, I was less concerned about it before, but if we should support getters then its definitely a concern.

---

I will go ahead then and do throw detection and adjust the documentation.

daxpedda: Done.

guide/src/reference/static-js-objects.md
```diff
@@ -65,6 +70,20 @@ extern "C" {
...
+
+```rust
+extern "C" {
+    type Crypto;
+    #[wasm_bindgen(thread_local_v2, js_name = crypto)]
+    static CRYPTO: Option<Crypto>;
+}
+```
+
+If `crypto` is undeclared in JS, it will simply return `None` in Rust.
```
RunDevelopment, line 85: I wrote a bit more about the feature. The main things I added are to mention that `null` and `undefined` also map to `None` and that namespaces work as well.

I also gave this feature a name: optional statics. This will hopefully make it easier to teach the feature.

Unfortunately, I can't use GH suggestions to suggest the change, because those suggestions use MD syntax... so here it is:

    ## Optional statics

    Sometimes, the JavaScript value you're trying to access may not be available, e.g. because the browser does not support it. In this case, you can use `Option<T>` to handle this. Example:

    ```rust
    #[wasm_bindgen]
    extern "C" {
        type Crypto;
        #[wasm_bindgen(thread_local_v2, js_name = crypto)]
        static CRYPTO: Option<Crypto>;
    }
    ```

    If `crypto` is not defined in JavaScript, it will simply return `None` in Rust.

    Note that `None` will also be returned if the value of the `crypto` variable is nullish (`null` or `undefined`).

    Namespaces are handled similarly. Example:

    ```rust
    #[wasm_bindgen]
    extern "C" {
        #[wasm_bindgen(thread_local_v2, js_name = port, js_namespace = server)]
        static PORT: Option<u32>;
    }
    ```

    `PORT` will only return a `Some` value in Rust if `server.port` is set to a number in JavaScript. If `server` is not defined or has a nullish value, it will return `None`. If the `port` property is not defined on the `server` object or is set to a nullish value, it will also return `None`.

crates/cli-support/src/js/mod.rs
```diff
@@ -2578,6 +2578,30 @@ __wbg_set_wasm(wasm);"
...
 
+    fn import_static(&mut self, import: &JsImport, optional: bool) -> Result<String, Error> {
+        let mut name = self.import_name(&JsImport {
+            name: import.name.clone(),
+            fields: Vec::new(),
+        })?;
+
+        // After we've got an actual name handle field projections
+        if optional {
+            name = format!("typeof {name} === 'undefined' ? null: {name}");
```
RunDevelopment, line 2589: Super minor nit

```suggestion
            name = format!("typeof {name} === 'undefined' ? null : {name}");
```

RunDevelopment (approved):
LGTM! Thank you @daxpedda!

PR #4323, opened 2024-12-05
Support raw identifiers for enums and enum variants

RunDevelopment:
Using raw identifiers for enum names and enum variant names resulted in syntactically invalid JS. This PR fixes this issue by unraw-ing the identifiers.


crates/backend/src/ast.rs
```diff
@@ -475,8 +475,10 @@ pub struct Enum {
 #[cfg_attr(feature = "extra-traits", derive(Debug, PartialEq, Eq))]
 #[derive(Clone)]
 pub struct Variant {
-    /// The name of this variant
-    pub name: Ident,
+    /// The name of this variant in Rust
+    pub rust_name: Ident,
+    /// The name of this variant in JS
+    pub js_name: String,
```
daxpedda, line 481: Could we just keep storing the `Ident` and `unraw()` it when we need it?

crates/backend/src/encode.rs
```diff
@@ -236,23 +236,19 @@ fn shared_function<'a>(func: &'a ast::Function, _intern: &'a Interner) -> Functi
...
         comments: e.comments.iter().map(|s| &**s).collect(),
         generate_typescript: e.generate_typescript,
     }
 }
 
-fn shared_variant<'a>(v: &'a ast::Variant, intern: &'a Interner) -> EnumVariant<'a> {
+fn shared_variant(v: &ast::Variant) -> EnumVariant {
     EnumVariant {
-        name: intern.intern(&v.name),
+        name: v.js_name.as_str(),
```
daxpedda, line 251: Do you know what the `Interner` does exactly?

RunDevelopment: The entire encoded AST is "borrowed" in that all strings are `&str`. (I think this is because the encoder doesn't support anything else, but I'm not sure.) The job of `intern` to turn a temporary `String` into `&'a str`. Basically, the temporary string from an identifier doesn't live long enough, so `intern` stores it somehow and returns a reference to it that does live long enough.

crates/backend/src/ast.rs
daxpedda: This is not how something like that should be solved though imo.

If you see this as such a problem, then wrap the `Ident` into a new type, e.g. `Name`, and make two methods: `rust_name()` and `js_name()`.

RunDevelopment: Most places where the `unraw` is missing are where an `Ident` is taken directly from the syn AST and then converted to a string for our backend AST. A newtype wouldn't solve that.

daxpedda: I'm not sure what you are trying to say.

I imagined that you can encapsulate the state when `rust_name` and `js_name` is the same and when its not. How is that related to where a conversion happens?

RunDevelopment: Okay, so this place here is the only place in the AST whether the JS identifiers are always only derived from Rust identifiers. This is only the case, because enum variants don't support `js_name` (yet). For all other cases, we already store the rust name as an `Ident` and the JS name as a `String`.

When I said that there are still many places left that don't `unraw`, I meant that there are still code paths deriving the JS name that don't account for this. Example [1](https://github.com/rustwasm/wasm-bindgen/blob/c60e8071a174489ac9b63ef40fa234f334034f85/crates/macro-support/src/parser.rs#L355) [2](https://github.com/rustwasm/wasm-bindgen/blob/c60e8071a174489ac9b63ef40fa234f334034f85/crates/macro-support/src/parser.rs#L580).

daxpedda: > Okay, so this place here is the only place in the AST whether the JS identifiers are always only derived from Rust identifiers. This is only the case, because enum variants don't support `js_name` (yet). For all other cases, we already store the rust name as an `Ident` and the JS name as a `String`.

I see, a pointless optimization then.
Thank you for explaining.

guybedford (approved):

PR #4325, opened 2024-12-06
Add `wasm_bindgen::is_memory_shared()`

daxpedda:
This PR adds `wasm_bindgen::is_memory_shared()`, which simply determines if the memory module is shared or not.

Retrieving that information via the `WebAssembly.Memory` interface is difficult because it requires checking if the `buffer` is an instance of `SharedArrayBuffer`. But `SharedArrayBuffer` is only available with COOP + COEP. So a workaround could be going through the constructor name.

However, the information is already available inside the Wasm module without going through JS at all, so this saves the user from trying to figure out how to do this.


daxpedda:
After some thinking, I might want to close this after all.

This was only done with the thought in mind that some users like to use a shared `WebAssembly.Memory` even when not compiling with `atomics`. But maybe this is too niche and will never be supported by `wasm-bindgen` to begin with.

daxpedda:
After some thinking and input from others I decided to close this. It was a good exercise however and the same implementation strategy could be used in the future for other things.

PR #4329, opened 2024-12-06
Fix handling of JS keyword for imported functions, types and statics, and when causing invalid code gen

### LINKED ISSUE #4317: Import "new" becomes "_new"
I am trying to import `window.__TAURI__.menu.Menu.new` from Tauri.

```rust
#[wasm_bindgen(js_namespace = ["window", "__TAURI__", "menu"])]
extern "C" {
    #[wasm_bindgen]
    pub type Menu;

    #[wasm_bindgen(static_method_of = Menu)]
    pub async fn new(options: JsValue) -> Menu;
}
```

The generated JS is:

```js
imports.wbg.__wbg_new_bcb7a76eea7b9ef8 = function() { return logError(function (arg0) {
    const ret = window.__TAURI__.menu.Menu._new(arg0);
    return ret;
}, arguments) };
```

`new` has become `_new`, which does not exist. Using `js_name = "new"` does not affect the generated JS. I have made an ugly workaround by aliasing `_new` to `new`.

Presumably this behaviour is deliberate because `new` is a JS keyword. However, it is also a valid function name ("valid" != "good"). 

Would it be possible to force the `js_name` to be retained, even for keywords? Maybe something like `force_js_name = true` to avoid breaking any existing code that relies on this? I have browsed around the code a bit but I can't find where this substitution is made. Are other keywords treated similarly?

Alternatively, it would be useful if the guide mentioned this behaviour. It took me ages to figure out what was going on. I can have a go at making that change myself if needed.

### LINKED ISSUE #4328: Using `js_namespace = []` causes a crash in the CLI
### Describe the Bug

```rs
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = [])]
    fn my_function();
}
```

The proc macro allows this, but the CLI crashes with:

```
    status: exit code: 101
    stderr:
        thread 'main' panicked at crates\cli-support\src\wit\mod.rs:1087:34:
        range start index 1 out of range for slice of length 0
```

### Steps to Reproduce

Compile the code and run the CLI.

### Expected Behavior

I'm not sure, whether the proc macro should just not allow this, or whether this should be interpreted as "no namespace".

### Actual Behavior

Crash.

### Additional Context

We currently internally represent the namespace of an imported item as `Option<Vec<String>>`. The CLI apparently assumes that the inner `Vec` is non-empty.

RunDevelopment:
Fixes #4317
Fixes #4328

As I said in #4317, keyword handling was kind of a mess. Keywords are now handled correctly for both imports and exports AFAICT.

Changes:

1. Stricter parsing for `js_namespace`. Now, the first identifier cannot be a JS keyword and empty lists are no longer allowed. This is not a breaking change, because keyword identifiers resulted in invalid JS code gen, and empty lists crashed the CLI.
2. Keywords are now generally allowed for imported functions, types, and statics. However, they will cause an error if the resulting JS code gen would be invalid.
3. Keywords are now always correctly escaped when used as function parameters.
4. Exported classes and enums named after keywords now result in an error instead of invalid code gen. Since all JS keywords are lower-case, I don't export anyone to have used keywords for them anyway.
5. I fixed the keyword list. Whoever decided that we don't need to check Rust keywords forgot that raw identifiers and `js_name` exist. The list now includes all JS keywords.

About the keyword list update: I also add keywords that are only keywords in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode). JavaScript has 2 execution modes: strict mode and sloppy mode. Sloppy mode is the legacy mode and shouldn't be used anymore. The key takeaway here is that we should only generate code that works in strict mode, so we must not allow identifiers that are keywords in strict mode. This is important, because all JS files that use ESM (`import`/`export`) implicitly use strict mode. I don't want WBG to be in the situation where it only generates valid JS for *some* CLI targets.

Lastly, `await`. `await` is a keyword ONLY is `async` functions in JS. So it can generally be used as an identifier, but this can cause problems. If a class is named `await`, then this class cannot be used inside `async` functions. It's a weird edge case, and I'm not sure whether we should consider `await` a reserved keyword too. Right now, I just allowed it.


crates/macro-support/src/parser.rs
```diff
@@ -886,25 +942,28 @@ impl ConvertToAst<BindgenAttrs> for syn::ItemFn {
...
             self.attrs,
             self.vis,
             FunctionPosition::Free,
-            Some(&["default"]),
         )?;
         attrs.check_used();
-        Ok(ret.0)
+
+        // Due to legacy behavior, we need to escape all keyword identifiers as
+        // `_keyword`, expect `default`
```
daxpedda, line 956:
```suggestion
        // `_keyword`, except `default`
```

crates/macro-support/src/parser.rs
```diff
@@ -1720,6 +1773,52 @@ impl MacroParse<ForeignItemCtx> for syn::ForeignItem {
...
         };
 
+        // check for JS keywords
+
+        // We only need to check if there isn't a JS namespace or module. If
+        // there is namespace, then we already checked the namespace while
+        // parsing. If there is a module, we can rename the import symbol to
+        // avoid using keywords.
+        let needs_check = js_namespace.is_none() && module.is_none();
+        match &kind {
```
daxpedda, line 1783: The whole `match` could be gated behind `needs_check`, right?

crates/macro-support/src/parser.rs
```diff
@@ -19,29 +19,59 @@ use crate::ClassMarker;
...
+    "class",
+    "const",
+    "continue",
     "debugger",
     "default",
     "delete",
+    "do",
+    "else",
+    "enum", // always reserved, but not used
+    // "eval", reserved in strict mode, but behaves like a functions in JS, so we allow it
```
daxpedda, line 39: I believe we should put this in a separate list allowing this for functions only, e.g. classes can't be called `eval`.

crates/macro-support/src/parser.rs
```diff
@@ -19,29 +19,59 @@ use crate::ClassMarker;
...
 thread_local!(static ATTRS: AttributeParseState = Default::default());
 
-/// Javascript keywords which are not keywords in Rust.
-const JS_KEYWORDS: [&str; 20] = [
-    "class",
+/// Javascript keywords.
+///
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words
+const JS_KEYWORDS: [&str; 44] = [
+    "arguments", // reserved in strict mode, which we are in
```
daxpedda, line 26:
```suggestion
    "arguments",
```
We don't need this comment everywhere, just leave a note on the top-level comment.

crates/macro-support/src/parser.rs
```diff
@@ -19,29 +19,59 @@ use crate::ClassMarker;
...
     "debugger",
     "default",
     "delete",
+    "do",
+    "else",
+    "enum", // always reserved, but not used
+    // "eval", reserved in strict mode, but behaves like a functions in JS, so we allow it
     "export",
     "extends",
+    // "false", false is a keyword, but behaves like a literal in JS
```
daxpedda, line 42: Same as `eval`. But needs yet another type of list.
Could be inline as well, how you implement this is at your discretion.

crates/macro-support/src/parser.rs
```diff
@@ -19,29 +19,59 @@ use crate::ClassMarker;
...
+    "private",   // reserved in strict mode, which we are in
+    "protected", // reserved in strict mode, which we are in
+    "public",    // reserved in strict mode, which we are in
+    "return",
+    "static", // reserved in strict mode, which we are in
+    "super",
     "switch",
     "this",
     "throw",
+    // "true", true is a keyword, but behaves like a literal in JS
```
daxpedda, line 65: Same as `false`.

daxpedda (changes requested):
Good stuff!

Missing a changelog entry.

crates/macro-support/src/parser.rs
RunDevelopment: I agree. I'll add something for it.

RunDevelopment:
Done. Thanks for the suggestions!

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## Unreleased
+
+### Fixed
+
+- Fixed JS keyword-like identifiers not being handled correctly for imports and exports.
+  [#4329](https://github.com/rustwasm/wasm-bindgen/pull/4329)
```
daxpedda, line 9: Considering this is a breaking change, even if an acceptable one, this needs to be expanded.
Mention at least all the things that are now a compile error that weren't before and why they are now.

crates/macro-support/src/parser.rs
```diff
@@ -21,9 +21,12 @@ thread_local!(static ATTRS: AttributeParseState = Default::default());
 
 /// Javascript keywords.
 ///
+/// Note that some of these keywords are only reserved in strict mode. Since we
+/// generate strict mode JS code, we treat all of these as reserved.
+///
 /// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words
-const JS_KEYWORDS: [&str; 44] = [
-    "arguments", // reserved in strict mode, which we are in
+const JS_KEYWORDS: [&str; 47] = [
```
daxpedda, line 28: Hm, seeing that this is a superset of `VALUE_LIKE_JS_KEYWORDS` I would have preferred it to include it programmatically. But I'm afraid Rust doesn't give us the tools for that.

WDYT of removing the corresponding keywords from that list and include them programmatically correctly in `is_js_keyword()`?

daxpedda (changes requested):

crates/macro-support/src/parser.rs
RunDevelopment: Not a fan. The idea of a list that contains all keywords, except all keywords in this other list, seems weird to me. I also don't think the code duplication is bad enough to justify it.

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 * Optional parameters are now typed as `T | undefined | null` to reflect the actual JS behavior.
   [#4188](https://github.com/rustwasm/wasm-bindgen/pull/4188)
 
+### Fixed
+
+- Fixed JS keyword-like identifiers not being handled correctly. Keywords on imports will no longer be escaped. Using keyword identifiers in places that caused invalid JS code gen will now error at compile time.
```
daxpedda, line 13: This needs to be detailed. A list of all the things that are now a compile error is the minimum. Instead of listing all the keywords here link to it. But "in places" needs to be expanded in detail.

crates/macro-support/src/parser.rs
daxpedda: Sure, leave a comment at `VALUE_LIKE_JS_KEYWORDS` then saying that `JS_KEYWORDS` is a superset.

CHANGELOG.md
RunDevelopment: Done.

daxpedda (approved):
LGTM! Thanks for all the work!

alockey:
Thanks for the prompt fix - much appreciated!

dylanowen:
I believe this is a breaking change. Something like:
```rust
#[wasm_bindgen(module = "dialog-polyfill")]
extern "C" {
    #[wasm_bindgen(js_namespace = default, js_name = registerDialog)]
    fn register_dialog(dialog: &HtmlDialogElement);
}
```
would generate:
```javascript
import { default as default1 } from 'dialog-polyfill';
```
but now it breaks with these changes. (I'm not saying this style of import should be supported, but the documentation doesn't make it immediately obvious on how I can import [`dialog-polyfill`](https://www.npmjs.com/package/dialog-polyfill))

Spxg:
> how I can import [`dialog-polyfill`](https://www.npmjs.com/package/dialog-polyfill)

Maybe like this?

```rust
#[wasm_bindgen(module = "./node_modules/dialog-polyfill/dist/dialog-polyfill.esm.js")]
extern "C" {
    #[wasm_bindgen(js_namespace = default, js_name = registerDialog)]
    fn register_dialog(dialog: &HtmlDialogElement);
}
```

Perhaps script could also be used:
```html
<body><script src="dist/dialog-polyfill.js"></script></body>
```

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = ["globalThis", "dialogPolyfill"], js_name = registerDialog)]
    fn register_dialog(dialog: &HtmlDialogElement);
}
```

guybedford:
It actually seems like my update in https://github.com/wasm-bindgen/wasm-bindgen/pull/4744 inadvertantly fixed this bug as well, so https://github.com/wasm-bindgen/wasm-bindgen/pull/4329#issuecomment-2672840111 will be resolved in the next release.

PR #4333, opened 2024-12-07
Make API for exporting JS content more explicit

RunDevelopment:
This PR brings the improvements to the `export` function from #4258. Having this change will make implementing another PR easier, so I wanted to make a separate PR for them now.

Change:
- Instead of trying to detect what kind of expression is being exported from the string value of the JS expression, an enum is used to explicitly spell out what types of expressions the `export` function supports. It is the responsibility of the caller to ensure the string matches the contract laid out by the `ExportJs` type. The makes the `export` function a lot more explicit and easier to change.

This PR is purely internal refactoring and doesn't affect the generated JS in any way.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -120,6 +120,18 @@ struct FieldAccessor {
     is_optional: bool,
 }
 
+/// Different JS constructs that can be exported.
+enum ExportJs<'a> {
+    /// A class of the form `class Name {...}`.
+    Class(&'a str),
+    /// An anonymous function expression of the form `function(...) {...}`.
+    ///
+    /// Note that the function name is not included in the string.
+    Function(&'a str),
+    /// An arbitrary JS expression.
+    Expression(&'a str),
+}
+
 const INITIAL_HEAP_VALUES: &[&str] = &["undefined", "null", "true", "false"];
 // Must be kept in sync with `src/lib.rs` of the `wasm-bindgen` crate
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -163,38 +175,46 @@ impl<'a> Context<'a> {
...
+        export: ExportJs,
         comments: Option<&str>,
     ) -> Result<(), Error> {
         let definition_name = self.generate_identifier(export_name);
-        if contents.starts_with("class") && definition_name != export_name {
+        if matches!(export, ExportJs::Class(_)) && definition_name != export_name {
             bail!("cannot shadow already defined class `{}`", export_name);
         }
 
-        let contents = contents.trim();
```
daxpedda, line 174: I assume this wasn't needed anymore because you removed the `\n` on the other end?

crates/cli-support/src/js/mod.rs
```diff
@@ -1169,10 +1187,10 @@ __wbg_set_wasm(wasm);"
 
         self.write_class_field_types(class, &mut ts_dst);
 
-        dst.push_str("}\n");
+        dst.push('}');
```
daxpedda, line 1190: Lets assume this won't break any output we missed because its not covered by our tests.
:crossed_fingers:

crates/cli-support/src/js/mod.rs
daxpedda: Figured it out.

daxpedda (approved):

PR #4335, opened 2024-12-07
Fixed webidl output formatting on Windows

RunDevelopment:
I found the reason formatting the output files from `webidl` didn't work on Windows. The `rustfmt` command was invoked with all files given as command arguments. On Windows, there's a 32k character limit to command strings, so it failed to execute the command, because it was too long.

My solution was to invoke the command multiple times with 400 files each. In my testing ~840 files was the limit, but since the command length obviously depends on the file names, I gave us a lot of breathing room by picking 400.

---

I also removed the note about formatting failing, since it isn't relevant anymore. I'll also make another PR to address the newline issue.


daxpedda (approved):
Ah, and here I thought its an issue about long paths.

Thank you for fixing this!

PR #4336, opened 2024-12-07
Use LF for newlines everywhere

RunDevelopment:
Related to #4335

I'm addressing the comment [here](https://github.com/rustwasm/wasm-bindgen/pull/4302#discussion_r1874003568) in 2 ways:

1. I added a `rustfmt.toml` to force LF newlines.
2. I added `* text=auto eol=LF` to `.gitattributes`. This causes git to automatically normalize all line ends to LF. Contributors on Windows now don't have to worry about having to configure git to checkout LF or CRLF.

    I have this line in the `.gitattributes` in a bunch of project. It generally makes it easier for Windows and Unix people to co-operate, since everyone checks out/in the exact same files. No newline weirdness. So I can only recommend this.

---

If you think that forcing everyone to use LF is too harsh/could create problems, I can also think of a different way to make the rustfmt call in webidl use the correct newlines. The problem is that the rustfmt tries to detect the existing newlines of a file to determine which ones to use, but the raw output of webidl is a `.rs` file with all tokens in a single line. There is no newline to detect. So it defaults to the native newline, which is the cause of the problem. So the solution could be to detect the newlines from another existing file, and then pass what we detected to rustfmt.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
+# Normalize line endings to LF
+* text=auto eol=LF
+
 *.webidl linguist-vendored
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,2 @@
+edition = "2021"
+newline_style = "Unix"
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -998,8 +998,6 @@ pub fn generate(from: &Path, to: &Path, options: Options) -> Result<String> {
     fn rustfmt(paths: impl IntoIterator<Item = PathBuf>) -> Result<()> {
         // run rustfmt on the generated file - really handy for debugging
         let result = Command::new("rustfmt")
-            .arg("--edition")
-            .arg("2021")
             .args(paths)
             .status()
             .context("rustfmt failed")?;


rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+edition = "2021"
```
daxpedda, line 1: `cargo fmt` doesn't need it, because it gets the edition from `Cargo.toml`. I'm assuming calling `rustfmt` directly doesn't do that?

Lets just leave a comment to that effect here:
```suggestion
// Required to make direct calls to `rustfmt` work for our WebIDL generation.
edition = "2021"
```

.gitattributes
daxpedda, line 1: I'm against enforcing this on a Git level.
I've had enough issues with that in the past which I don't want to repeat here.

daxpedda (changes requested):
Great, thank you!

> If you think that forcing everyone to use LF is too harsh/could create problems, 

I have honestly no idea how that would create problems on a Rust level :open_mouth:.

rustfmt.toml
RunDevelopment: > I'm assuming calling rustfmt directly doesn't do that?

I have no idea. I just wanted to get rid of the `.arg("edition").arg("2021")`

RunDevelopment:
> I'm against enforcing this on a Git level.

That means I can close this PR. There's no point in having a `rustfmt.toml`, because we cannot specify the correct newline format there. Manually detecting it is, but that will be a separate PR.

daxpedda:
> > I'm against enforcing this on a Git level.
> 
> That means I can close this PR. There's no point in having a `rustfmt.toml`, because we cannot specify the correct newline format there. Manually detecting it is, but that will be a separate PR.

I'm not following exactly. We should still apply the `rustfmt.toml` so the call from `rustfmt` does the correct thing. The fact that git might interfere can still be up to users to fix on their end. Or what am I missing here?

RunDevelopment:
> I'm not following. We should still apply the rustfmt.toml so the call to rustfmt does the correct thing. The fact that git might interfere can still be up to users to fix on their end. Or what am I missing here?

Having `new_line = Unix` in `rustfmt.toml` will cause problems for anyone that checks out the repo with CRLF on Windows. Saying that users have to fix it on their end in nonsense. If you don't use `* text=auto eol=LF`, then you are saying "both line ends work." To then turn around and require LF anyway is just horrible. You are just pushing the task of configuring git for this repo onto the user after they've run into issues caused by this. I've run into this crap multiple times, and it sucks every time. I just want to make a quick PR to OSS, CI fails because of formatting, I run the formatter, 5k files changed. Great, now I can look up the git commands for re-configuring line ends, renormalize everything, and hope I didn't forget anything.

Right now, git is configured so that users are allowed to have CRLF and rustfmt naturally supports that. I'm not going to implement a line ending change that I know will just cause issues for contributors.

daxpedda:
> Saying that users have to fix it on their end in nonsense. If you don't use `* text=auto eol=LF`, then you are saying "both line ends work." To then turn around and require LF anyway is just horrible. You are just pushing the task of configuring git for this repo onto the user after they've run into issues caused by this. I've run into this crap multiple times, and it sucks every time. I just want to make a quick PR to OSS, CI fails because of formatting, I run the formatter, 5k files changed. Great, now I can look up the git commands for re-configuring line ends, renormalize everything, and hope I didn't forget anything.

Please refrain from non-technical statements or arguments, this is not the place for it.

Clearly you have a very negative experience around this issue. I didn't. This is why we discuss this, so you can lay out your point of view and I mine.

My experience was the opposite. Enforcing line endings on a git level caused various issues for me where it wasn't possible to commit files that purposefully had different line endings in them. So the OSS culture that I've interacted with refrains from this kind of override. Commits should always happen **as-is**, conversions happening on the git level is considered bad in my circles.

> Having `new_line = Unix` in `rustfmt.toml` will cause problems for anyone that checks out the repo with CRLF on Windows.

How so?

This is an area I have no experience in as a Linux user. I do remember that back then when I used Windows I set up line endings globally to address these issues, but I have no idea anymore what the problem was exactly in the first place.
> If you don't use `* text=auto eol=LF`, then you are saying "both line ends work."

That's not what Git says this means. Please see [the documentation](https://git-scm.com/docs/gitattributes#_eol). I have no idea if this is used to signify that popularly though.

> You are just pushing the task of configuring git for this repo onto the user after they've run into issues caused by this.

I am because AFAIK there is no way to tell Git to tell users that this repo is LF. The best setting I know that is representative is `core.autocrlf=false`, which just says checkout and commit as-is. But AFAIK this isn't recommended because it interferes with what users might actually want and I have no idea if that actually solves the problem you are trying to describe.

RunDevelopment:
> Commits should always happen as-is, conversions happening on the git level is considered bad in my circles.

When installing git on Windows, this is what you see as step 8/11:

image
(Screenshot from a [yt video](https://www.youtube.com/watch?v=2j7fD92g-gE), because I couldn't find documentation for the installer.)

The default for `core.autocrlf` [is `false` on Linux](https://stackoverflow.com/questions/39408793/git-core-autocrlf-line-ending-default-setting), but here you can see what they are for Windows. You see the problem? They even call it the "recommended setting on Windows."

This is why I said that it's "up to users to fix on their end" is nonsense. If you *require* checkout and commit as-is, you already messed things up anyone not having `core.autocrlf=false`, which are most people on Windows. Git supports all of these configurations, so if your repo doesn't, then your repo is the problem, not how users configure git.

> > If you don't use `* text=auto eol=LF`, then you are saying "both line ends work."
> 
> That's not what Git says this means. Please see [the documentation](https://git-scm.com/docs/gitattributes#_eol). I have no idea if this is used to signify that popularly though.

Nope, this is pretty much what it means. To quote the docs:

> Unspecified
> If the eol attribute is unspecified for a file, its line endings in the working directory are determined by the core.autocrlf or core.eol configuration variable (see the definitions of those options in [git-config[1]](https://git-scm.com/docs/git-config)). If text is set but neither of those variables is, the default is eol=crlf on Windows and eol=lf on all other platforms.

If you don't define `eol`, then it's up to the user to decide how to check out files. Ergo, users can decide whether they want LF or CRLF. You gave them this choice.

If your project only works with one type of line endings, then specify `eol=<value>`. If you want checkout and commit as is, then make it `binary`. This is also [how GitHub says](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings?platform=windows#per-repository-settings) you should do per-repo line end settings. (Along with again recommending `core.autocrlf true` on Windows.)

Note specifying `eol` but requiring a specific checkout is just a trap for users to fall into. And that trap will disproportionately affect Windows users.

---

Btw, my `core.autocrlf` is set to `input`, because I got tired of so many repo's tooling assuming LF line ends. 

But guess what? Now I have the problem of another kind of bad tooling that assumes that everyone on Windows uses CRLF. It just doesn't end. This is also the reason why most modern tooling has given up on platform-specific line ends and instead of just detects line ends from file content and uses that. That's also what `rustfmt` does by default.

daxpedda:
> This is why I said that it's "up to users to fix on their end" is nonsense. If you _require_ checkout and commit as-is, you already messed things up anyone not having `core.autocrlf=false`, which are most people on Windows. Git supports all of these configurations, so if your repo doesn't, then your repo is the problem, not how users configure git.

I really don't know what justifies this kind of wording choice. Again, lets keep this conversation civil and *technical*. I have a hard time finding anything useful in that paragraph to work with.

> > > If you don't use `* text=auto eol=LF`, then you are saying "both line ends work."
> > 
> > 
> > That's not what Git says this means. Please see [the documentation](https://git-scm.com/docs/gitattributes#_eol). I have no idea if this is used to signify that popularly though.
> 
> Nope, this is pretty much what it means. To quote the docs:
> 
> > Unspecified
> > If the eol attribute is unspecified for a file, its line endings in the working directory are determined by the core.autocrlf or core.eol configuration variable (see the definitions of those options in [git-config[1]](https://git-scm.com/docs/git-config)). If text is set but neither of those variables is, the default is eol=crlf on Windows and eol=lf on all other platforms.
> 
> If you don't define `eol`, then it's up to the user to decide how to check out files. Ergo, users can decide whether they want LF or CRLF. You gave them this choice.

We are talking about two different things. You are talking about the check out mechanism, but I was talking about telling the users what kind of file ending this file should have. These are two different things, even if Git is recommending settings that try to gloss over what kind of line-endings files really have and give something different to the user.

---

I'm tired of this conversation as its not really constructive. I'm gonna make my stance as clear as I possible can:
- I don't know how all this works.
- I had bad experiences with this in the past and I'm suspicious of repeating the same mistakes here again. As I don't know how it all actually works, my suspicions might be totally unfounded.
- I don't really know how the experience is for a Windows user and what kind of problems they actually encounter.

If you are interested in resolving this, you have to:
- Explain to me exactly what the problem is. You have been going off on a rant rather then explaining it to me.
- Engage the concerns that I have and try to address them.
- Explain how the solution works exactly and how it would resolve the problem.
- Keep the conversation civil and technical. There's no way we will have a constructive conversation if you call things "nonsense", "your repo is the problem", "this crap". They simply don't contain any useful information apart from making other people not want to engage in this conversation anymore.

RunDevelopment:
Yeah, no. I'm not going to continue this as is. All of this could have been resolved by reading up on the basics of how to do newlines cross-platform in git and doing some research on `autocrlf`. Heck, I even linked to the GH docs on the former.

Also, if you actually want me to engage with your concerns, then say what they are. So far, you haven't clearly listed any of your "experiences", "mistakes", and "suspicions" (quoted to highlight plural). The only thing I found you mentioning was "it wasn't possible to commit files [with] different line endings", which is about as specific as saying "my rust code doesn't run." Similarly, if you want me to explain something, then say what you want explained. Telling me to "explain [...] the problem" doesn't mean much, especially when you just reductively dismissed a previous explanation as a "rant".

PR #4340, opened 2024-12-08
Add `copy_to_uninit()` to all `TypedArray`s

daxpedda:
Currently copying from a `TypedArray`, e.g. `Uint8Array`, to a `[MaybeUninit<T>]` is not possible without going through `raw_copy_to_ptr()`, making it unsafe.

This PR adds a method `copy_to_uninit()` to all `TypedArray`s, making it safe to copy to a `[MaybeUninit<T>]` and also returns a `[T]` because all its values are guaranteed to be initialized now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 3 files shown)


RunDevelopment (approved):
Nice feature!

PR #4342, opened 2024-12-08
Document all features of `wasm-bindgen`

RunDevelopment:
This documents all features and puts them into 3 categories:

1. User-facing features. These are features users are allowed to use and that we actively support. E.g. `strict-macro`.
2. Deprecated features. These are features users are not supposed to use anymore, but nonetheless allowed.
3. Internal features. Features for testing and debugging that happen to public.

I documented all user-facing and deprecated feature and put a note on all internal features.


daxpedda (approved):
Great!

PR #4345, opened 2024-12-08
Fix features on `wasm-bindgen-msrv` crate

RunDevelopment:
Implements the change requested [here](https://github.com/rustwasm/wasm-bindgen/pull/4278#pullrequestreview-2486688591). I made this its own PR, because this is a fix for the misconfigured features of the `msrv` crate and has nothing to do with marker traits.

### DIFF
--- a/crates/msrv/lib/Cargo.toml
+++ b/crates/msrv/lib/Cargo.toml
@@ -5,13 +5,12 @@ publish = false
 version = "0.0.0"
 
 [features]
-default = ["std"]
-std = [
-  "wasm-bindgen/std",
-  "js-sys/std",
-  "wasm-bindgen-futures/std",
-  "web-sys/std",
-  "wasm-bindgen-test/std",
+default = [
+  "wasm-bindgen/default",
+  "js-sys/default",
+  "wasm-bindgen-futures/default",
+  "web-sys/default",
+  "wasm-bindgen-test/default",
 ]
 
 [dependencies]
--- a/crates/msrv/resolver/Cargo.toml
+++ b/crates/msrv/resolver/Cargo.toml
@@ -6,13 +6,12 @@ resolver = "1"
 version = "0.0.0"
 
 [features]
-default = ["std"]
-std = [
-  "wasm-bindgen/std",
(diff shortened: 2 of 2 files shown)


crates/msrv/lib/Cargo.toml
```diff
@@ -5,13 +5,12 @@ publish = false
...
-  "wasm-bindgen-futures/std",
-  "web-sys/std",
-  "wasm-bindgen-test/std",
+default = [
+  "wasm-bindgen/default",
+  "js-sys/default",
+  "wasm-bindgen-futures/default",
+  "web-sys/default",
+  "wasm-bindgen-test/default",
 ]
```
daxpedda, line 14: Don't replace the `std` feature, that's still needed.

daxpedda (changes requested):

crates/msrv/lib/Cargo.toml
RunDevelopment: Is it? I did a simple text search and couldn't find it being used anywhere.

daxpedda: Its not used because it was under `default` before. If we intend to add more features (`msrv`) we have to test these separately, in which case it will be used.

daxpedda: Or, if you want to, we can merge this as-is and re-add the `std` crate feature in #4278. Let me know how you want to proceed.

RunDevelopment: Could you please do that? I have no idea how to set that up properly.

daxpedda:
Closing in favor of https://github.com/rustwasm/wasm-bindgen/pull/4278.

crates/msrv/lib/Cargo.toml
daxpedda: Did it in https://github.com/rustwasm/wasm-bindgen/pull/4278/commits/e1d2358abab72afb030710ad90047a4f9d574239.

RunDevelopment: Thank you!

PR #4346, opened 2024-12-09
Better TS type formatting for array types

RunDevelopment:
Small PR. If the TS item type of an array is a valid identifier, we don't need to the parens.

FYI, the parens are necessary for imported types. Imported types can specify an arbitrary TS type expression as their TS type.

### DIFF
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -1,5 +1,7 @@
 use std::char;
 
+use crate::js::identifier::is_valid_ident;
+
 macro_rules! tys {
     ($($a:ident)*) => (tys! { @ ($($a)*) 0 });
     (@ () $v:expr) => {};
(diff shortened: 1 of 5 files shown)


daxpedda (commented):
> FYI, the parens are necessary for imported types. Imported types can specify an arbitrary TS type expression as their TS type.

What is an "arbitrary TS type expression"?
Why can that contain invalid Unicode (if I'm interpreting the code correctly)?
Can we have a test for something like this or is there one already?

RunDevelopment:
> What is an "arbitrary TS type expression"?

Anything that isn't just a plain identifier. E.g. `number | string`. Absent the ability to specify custom types for parameters and return types, people are currently using this workaround:

```rs
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(typescript_type = "number | string")]
    type CustomType;
}

#[wasm_bindgen]
pub fn single(a: CustomType) {}
```

I added a test.

> Why can that contain invalid Unicode (if I'm interpreting the code correctly)?

By the guarantees of `str`, obviously no. What in the code gave you that idea? (Because I just moved the stuff that we already had.)

daxpedda (approved):
Thanks!

> > Why can that contain invalid Unicode (if I'm interpreting the code correctly)?
> 
> By the guarantees of `str`, obviously no. What in the code gave you that idea? (Because I just moved the stuff that we already had.)

I didn't understand what `is_valid_ident()` actually does. But I had a proper read now.

PR #4347, opened 2024-12-09
Simplify `no_std` implementation

daxpedda:
By removing the usage of `thread_local!` entirely, we can simplify the code around `no_std`.


daxpedda:
> Does that mean string interning works on `no_std` now?

It does not.

It uses a `HashMap`, so we either stop using the `std` one or find a different implementation.
In any case, its unrelated to this PR.

PR #4349, opened 2024-12-10
Import memory and enable `module()` for Node.js

daxpedda:
Revert #4318 in favor of actually importing the memory. This is getting us closer to actually be able to support multi-threading on Node.js.

I briefly attempted to get this into the bundler target as well. But quickly encountered https://github.com/webpack/webpack/issues/8157.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@
 * Add test coverage support for Node.js.
   [#4348](https://github.com/rustwasm/wasm-bindgen/pull/4348)
 
+* Support importing memory and using `wasm_bindgen::module()` in Node.js.
+  [#4349](https://github.com/rustwasm/wasm-bindgen/pull/4349)
+
 ### Changed
 
 * Optional parameters are now typed as `T | undefined | null` to reflect the actual JS behavior.
@@ -50,7 +53,7 @@ Released 2024-12-07
 
 ### Added
 
-* Add support for multi-threading in Node.js.
+* Add support for compiling with `atomics` for Node.js.
   [#4318](https://github.com/rustwasm/wasm-bindgen/pull/4318)
 
 * Add `WASM_BINDGEN_TEST_DRIVER_TIMEOUT` environment variable to control the timeout to start and connect to the test driver.
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -261,7 +261,12 @@ impl<'a> Context<'a> {
 
     fn generate_node_imports(&self) -> String {
         let mut imports = BTreeSet::new();
-        for import in self.module.imports.iter() {
+        for import in self
+            .module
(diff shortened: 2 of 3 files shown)


daxpedda:
Alright, I confirmed that multi-threading actually works with Node.js and WBG. Its a bit messed up because of two things:
- Node.js doesn't actually support the Web Worker API, but has its own custom one: [`worker_threads`](https://nodejs.org/api/worker_threads.html#class-worker). See https://github.com/nodejs/node/issues/43583.
- For some reason I'm not aware of and totally lack the context, we rename the module of all our internal imports to name of the file. I'm afraid to change anything there because I don't know how people are deploying Node.js stuff. Definitely not impossible to work around right now.

Here is some sample code for the interested:
```rust
use js_sys::{Array, Object};
use wasm_bindgen::prelude::wasm_bindgen;
use wasm_bindgen::{JsCast, JsValue};
use web_sys::console;

fn main() {
    let url = META.with(Meta::url);
    let import_module = format!("./{}", url.rsplit_once('/').unwrap().1);
    let script = include_str!("worker.js").replacen("@shim.js", &url, 1);
    let script = script.replacen("@import_module", &import_module, 2);
    let options = WorkerOptions::new();
    options.set_eval(true);
    options.set_worker_data(&Array::of2(
        &wasm_bindgen::module(),
        &wasm_bindgen::memory(),
    ));

    Worker::new(script, &options).unwrap();
}

#[wasm_bindgen]
pub fn worker_entry() {
    console::log_1(&"Hello from Worker!".into());
}

#[wasm_bindgen]
extern "C" {
    type Meta;

    #[wasm_bindgen(thread_local_v2, js_namespace = import, js_name = meta)]
    static META: Meta;

    #[wasm_bindgen(method, getter)]
    fn url(this: &Meta) -> String;
}

#[wasm_bindgen(module = "node:worker_threads")]
extern "C" {
    type Worker;

    #[wasm_bindgen(constructor, catch)]
    fn new(filename: String, options: &WorkerOptions) -> Result<Worker, JsValue>;

    type WorkerOptions;

    #[wasm_bindgen(method, setter, js_name = eval)]
    fn set_eval(this: &WorkerOptions, value: bool);

    #[wasm_bindgen(method, setter, js_name = workerData)]
    fn set_worker_data(this: &WorkerOptions, value: &JsValue);
}

impl WorkerOptions {
    fn new() -> Self {
        Object::new().unchecked_into()
    }
}
```

```js
// `worker.js` file.
import { workerData } from 'node:worker_threads'
import * as import0 from '@shim.js'

const [module, memory] = workerData

const imports = { '@import_module': import0, wbg: { memory: memory } }
const wasmInstance = new WebAssembly.Instance(module, imports)
const wasm = wasmInstance.exports

imports['@import_module'].__wbg_set_wasm(wasm, module)
wasm.__wbindgen_start()

wasm.worker_entry()
```

PR #4351, opened 2024-12-11
Add support for cfg_attr for struct fields

### LINKED ISSUE #2703: wasm_bindgen(skip) in cfg_attr
Hello
I tried to use `#[cfg_attr(feature = "wasm-bindgen", wasm_bindgen(skip))]` but I get this:

```   |
99 |     #[cfg_attr(feature = "wasm-bindgen", wasm_bindgen(skip))]
   |                                          ^^^^^^^^^^^^ not a non-macro attribute
```

But it works for other macros like derive. Is this a bug or does a workaround exist?

fl0rek:
closes #2703

Overview

- During #[wasm_bindgen] attribute macro expansion on struct, instead of generating ast for Program, it puts a #[derive(BindgenedStruct)] on it and preserves all the wasm_bindgen atributes as-is.
- When #[derive(BindgenedStruct)] is expanded, cfg_attr are already evaluated, so macro can just processes the leftover attributes and generate the glue code required, as it's done now. Attributes are then automatically removed by the compiler, since macro is defined as #[proc_macro_derive(BindgenedStruct, attributes(wasm_bindgen))].

### DIFF
--- a/crates/macro-support/src/lib.rs
+++ b/crates/macro-support/src/lib.rs
@@ -15,6 +15,7 @@ pub use crate::parser::BindgenAttrs;
 use crate::parser::{ConvertToAst, MacroParse};
 use backend::{Diagnostic, TryToTokens};
 use proc_macro2::TokenStream;
+use quote::quote;
 use quote::ToTokens;
 use quote::TokenStreamExt;
 use syn::parse::{Parse, ParseStream, Result as SynResult};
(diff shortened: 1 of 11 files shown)


tests/wasm/classes.rs
```diff
@@ -480,6 +480,19 @@ fn renamed_field() {
     js_renamed_field();
 }
 
+#[cfg_attr(
+    target_arch = "wasm32",
+    wasm_bindgen(inspectable, js_name = "ConditionalSkipClass")
```
daxpedda, line 485:
```suggestion
    wasm_bindgen(js_name = "ConditionalSkipClass")
```
Or is there some reason this is here?

tests/wasm/classes.rs
```diff
@@ -480,6 +480,19 @@ fn renamed_field() {
...
+)]
+pub struct ConditionalSkip {
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen(skip))]
```
daxpedda, line 488: We should check on the JS side if this is properly applied.

tests/wasm/classes.rs
```diff
@@ -480,6 +480,19 @@ fn renamed_field() {
...
+    pub skipped_field: [u8; 8],
+
+    /// cfg_attr annotated field
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen(getter_with_clone))]
+    pub needs_clone: String,
```
daxpedda, line 493: We might want to say in the comment instead that this wouldn't compile otherwise, to explain how we are testing this.

src/lib.rs
```diff
@@ -101,6 +101,7 @@ pub mod prelude {
     pub use crate::JsCast;
     pub use crate::JsValue;
     pub use crate::UnwrapThrowExt;
+    pub use wasm_bindgen_macro::BindgenedStruct;
```
daxpedda, line 104: Lets put it in `__rt` instead. It shouldn't be visible after all.

crates/macro-support/src/lib.rs
```diff
@@ -24,7 +24,40 @@ mod parser;
...
     let item = syn::parse2::<syn::Item>(input)?;
+    if let syn::Item::Struct(mut s) = item {
+        s.attrs.insert(
+            0,
+            syn::Attribute {
+                pound_token: Default::default(),
+                style: syn::AttrStyle::Outer,
+                bracket_token: Default::default(),
+                meta: syn::parse_quote! {
+                    derive(wasm_bindgen::prelude::BindgenedStruct)
```
daxpedda, line 38: We have to parse the attributes to extract the `#[wasm_bindgen(wasm_bindgen = <path>)]` attribute and use it here.

daxpedda (changes requested):
Thank you, this is a great improvement!

fl0rek (commented):
@daxpedda thanks for the review! Pushed the update after the prolonged xmas break

daxpedda (approved):
Thank you!

daxpedda:
Ah, apparently this requires a rebase! Otherwise LGTM!

PR #4354, opened 2024-12-15
Migrate `wasm-bindgen-cli` to `clap`

daxpedda:
This PR migrates `wasm-bindgen-cli`, `wasm-bindgen-test-runner` and `wasm2es6js` to [`clap`](https://crates.io/crates/clap).

`wasm-bindgen-test-runner` was parsing arguments in JS, this is now moved into the CLI via `clap` and is passed into JS fully parsed.


crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
```diff
@@ -139,8 +137,7 @@ pub fn execute(
             .env("NODE_PATH", env::join_paths(&path).unwrap())
             .arg("--expose-gc")
             .args(&extra_node_args)
-            .arg(&js_path)
-            .args(args),
```
daxpedda, line 143: I assume at some point this was used to be able to pass arguments into Node.js/Deno. But this was already not functional because when `wasm-bindgen-test-runner` started parsing arguments in JS it would fail during runtime.

PR #4368, opened 2024-12-19
Remove unnecessary setup in CI job

daxpedda:
Noticed some unnecessary setup in the CI.


daxpedda:
Right, false alarm! While we are only running native tests here, the native tests are executing Cargo with `wasm32-unknown-unknown`.

PR #4378, opened 2024-12-25
Use alloc::vec::Vec instead of Vec for no_std support

Dinnerbone:
Whilst `no_std` seems _technically_ supported, in practice it's unusable with a bunch of web-sys features due to relying on the implicit `std::vec::Vec` in places. This fixes that by using `alloc::vec::Vec` explicitly, which works in both std and no_std.

A quick test would be to `cargo check -p web-sys --no-default-features --features ImageData` - fails before this PR, succeeds after it.

After this change, we don't actually use `std` anywhere within `web-sys` - it could be permanently `no_std` if so desired. That'd certainly help keep it no_std compatible in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@
 * Internal functions are now removed instead of invalidly imported if they are unused.
   [#4366](https://github.com/rustwasm/wasm-bindgen/pull/4366)
 
-* `web-sys` crate now fully works on `no_std`.
+* Fixed `no_std` support for all APIs in `web-sys`.
   [#4378](https://github.com/rustwasm/wasm-bindgen/pull/4378)
 
 --------------------------------------------------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@
 * Internal functions are now removed instead of invalidly imported if they are unused.
   [#4366](https://github.com/rustwasm/wasm-bindgen/pull/4366)
 
+* `web-sys` crate now fully works on `no_std`.
+  [#4378](https://github.com/rustwasm/wasm-bindgen/pull/4378)
+
 --------------------------------------------------------------------------------
 
 ## [0.2.99](https://github.com/rustwasm/wasm-bindgen/compare/0.2.98...0.2.99)
--- a/crates/web-sys/src/lib.rs
+++ b/crates/web-sys/src/lib.rs
@@ -12,7 +12,7 @@
 //! require.
 
 #![doc(html_root_url = "https://docs.rs/web-sys/0.3")]
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
 #![allow(deprecated)]
 
 extern crate alloc;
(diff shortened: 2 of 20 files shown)


daxpedda (changes requested):
Exposing an `std` crate feature in `web-sys` was a mistake to begin with I believe. Too late now unfortunately. This also prevents us from properly testing it in CI.

In any case, thank you for catching this!
Could you just add a changelog entry?

Dinnerbone:
> Exposing an `std` crate feature in `web-sys` was a mistake to begin with I believe. Too late now unfortunately. This also prevents us from properly testing it in CI.

In that case; how about we just switch web-sys to be `no_std` always, and have the `std` feature only pass through to `wasm-bindgen` which does need it? We can do the same for `js-sys`, which doesn't need std either.

daxpedda:
> In that case; how about we just switch web-sys to be `no_std` always, and have the `std` feature only pass through to `wasm-bindgen` which does need it? We can do the same for `js-sys`, which doesn't need std either.

`js-sys` actually has a single use for `std`: implementing `Error` for `TryFromIntError`. But the issue with testing in CI is that we can't use `all-features` for `web-sys` without including `std` and listing all other `web-sys` features is a no-go. The same issue doesn't apply to `js-sys` so that's not the problem.

So making `web-sys` always `no_std` is a great idea! Feel free to go ahead and do that, I can do the change to CI in a follow-up.

Dinnerbone:
> `js-sys` actually has a single use for `std`: implementing `Error` for `TryFromIntError`

That's stable in core now - but I just noticed the MSRV is `1.57` which is too early. Oh well, someone in the future can do that :D

Added changelog and made web_sys `no_std` unconditional! And have a great Christmas! <3

CHANGELOG.md
```diff
@@ -67,6 +67,9 @@
 * Internal functions are now removed instead of invalidly imported if they are unused.
   [#4366](https://github.com/rustwasm/wasm-bindgen/pull/4366)
 
+* `web-sys` crate now fully works on `no_std`.
```
daxpedda, line 70:
```suggestion
* Fixed `no_std` support for all APIs in `web-sys`.
```
It sounds like the `std` `Vec` usage was on purpose, it was just an oversight :sweat_smile:!

Just for context: this is always a bug, because crate features have to be additive and they are not supposed to break compilation when you add a crate feature!

daxpedda (approved):
Thank you and merry Christmas as well!

daxpedda:
Self note: this is already covered by CI, because running `cargo build -p web-sys --all-features` is now enough to trigger a build-failure if the Std prelude is used anywhere.

crates/web-sys/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 //! require.
 
 #![doc(html_root_url = "https://docs.rs/web-sys/0.3")]
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
```
bushrat011899, line 15: This bit me while working on `no_std` for Bevy (you can read about the efforts [here](https://github.com/bevyengine/bevy/issues/15460)). For future reference, if you do this:

```rust
#![no_std]

#[cfg(feature = "std")]
extern crate std;
```

You can have an `std` feature without toggling the [implicit prelude](https://doc.rust-lang.org/reference/names/preludes.html#standard-library-prelude), which catches these kinds of import issues when compiling with all features.

daxpedda: Uh, that's pretty neat!
Thanks, will keep this in mind!

PR #4384, opened 2024-12-27
Add binding for `Date.to_locale_time_string_with_options`

### LINKED ISSUE #4383: Missing "options" parameter from Date.to_locale_time_string
According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#options), ToLocalTimeString takes an options parameter but it's missing from js-sys 
[js-sys::Date.to_locale_time_string](https://docs.rs/js-sys/latest/js_sys/struct.Date.html#method.to_locale_time_string)

paul-hansen:
Closes #4383

### DIFF
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3393,7 +3393,11 @@ extern "C" {
     pub fn to_locale_time_string(this: &Date, locale: &str) -> JsString;
 
     #[wasm_bindgen(method, js_name = toLocaleTimeString)]
-    pub fn to_locale_time_string_with_options(this: &Date, locale: &str, options: &JsValue) -> JsString;
+    pub fn to_locale_time_string_with_options(
+        this: &Date,
+        locale: &str,
+        options: &JsValue,
+    ) -> JsString;
 
     /// The `toString()` method returns a string representing
     /// the specified Date object.
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -3390,7 +3390,7 @@ extern "C" {
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString)
     #[wasm_bindgen(method, js_name = toLocaleTimeString)]
-    pub fn to_locale_time_string(this: &Date, locale: &str) -> JsString;
+    pub fn to_locale_time_string(this: &Date, locale: &str, options: &JsValue) -> JsString;
 
     /// The `toString()` method returns a string representing
     /// the specified Date object.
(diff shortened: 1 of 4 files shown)


paul-hansen:
I haven't been able to run the tests for js-sys locally when following the [contributing guide](https://rustwasm.github.io/docs/wasm-bindgen/contributing/js-sys/testing.html), getting a wasm-bindgen-cli mismatch that I can't seem to resolve even when using `cargo install -f wasm-bindgen-cli --path ../cli/ --locked` to build wasm-bindgen-cli from the repo. The tests build though 🤷

paul-hansen:
This should be marked as a breaking change since it adds a parameter to an existing function. Looks there's a label for that but I can't add labels if someone can do that for me. I'll get a changelog entry added now.

crates/js-sys/src/lib.rs
```diff
@@ -3390,7 +3390,7 @@ extern "C" {
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString)
     #[wasm_bindgen(method, js_name = toLocaleTimeString)]
-    pub fn to_locale_time_string(this: &Date, locale: &str) -> JsString;
+    pub fn to_locale_time_string(this: &Date, locale: &str, options: &JsValue) -> JsString;
```
daxpedda, line 3393: Instead we can add a new method called `to_local_time_string_with_options()` that fills this gap. Making this also not a breaking change.

daxpedda (changes requested):

crates/js-sys/src/lib.rs
paul-hansen: I can do that but I should note I did it this way to match `to_locale_date_string` which takes options. Since these are related functions it would make it inconsistent to have one have a separate fn for options and one not. (Another option would be to change bith to have a with_options, but that's also breaking)

I don't have strong feelings about this personally though so just lmk how you want to handle it. Thanks!

daxpedda: I see. I was following the pattern from `web-sys` actually. But yeah, we should still stick with `*_with_options()` simply to avoid the breaking change.

Thank you for your attention on this!

crates/js-sys/src/lib.rs
```diff
@@ -3392,6 +3392,13 @@ extern "C" {
     #[wasm_bindgen(method, js_name = toLocaleTimeString)]
     pub fn to_locale_time_string(this: &Date, locale: &str) -> JsString;
 
+    #[wasm_bindgen(method, js_name = toLocaleTimeString)]
+    pub fn to_locale_time_string_with_options(
+        this: &Date,
+        locale: &str,
+        options: &JsValue,
```
daxpedda, line 3399:
```suggestion
        options: Option<&Object>,
```
Is probably more accurate, again, deviating more from the `to_local_time()`. So lets just leave it for the next breaking change.

daxpedda (approved):
Thank you!

Issue #2945, opened 2022-06-16
Futures created with `wasm_bindgen_futures::spawn_local` don't get executed


felipellrocha:
I'm spinning a thread using `wasm_thread` (https://github.com/chemicstry/wasm_thread). Inside of it, I'm running some async functions inside of a `wasm_bindgen_futures::spawn_local` as such:

```rs
wasm_thread::spawn(move || {
  log::info!("1: spawn");
  while Ok(message) = channel_receiver.recv() {
    log::info!("2: message {:?}", &message);
    wasm_bindgen_futures::spawn_local(async move {
      log::info!("3: am i here?");
      download.asset().await;
    })
  }
});
```
Problem is that neither the 3rd `log::info!()` nor the `download.asset().await` get called.

Any idea why?

Liamolucko:
The way the JS event loop works is that other tasks (such as your futures) only get run once the current task finishes. That's never happening because you're blocking the thread waiting for messages over the channel.

To get around that, you could use an async channel (e.g. from `futures-channel`) instead of a blocking one, which will yield to the event loop instead of blocking (if you also spawn it using `wasm_bindgen_futures::spawn_local).

felipellrocha:
I have modified it to this:
```rs
wasm_thread::spawn(move || {
  log::info!("1: spawn");
  wasm_bindgen_futures::spawn_local(async move {
    log::info!("2: message {:?}", &message);
    while Ok(message) = channel_receiver.recv().await {
      log::info!("3: am i here?");
      download.asset().await;
    }
  })
});
```

I still get nothing

Liamolucko:
Hmm - my suspicion is that because the thread entrypoint returns immediately, the worker is getting closed. There isn't really a good way to get around that unless `wasm_thread` were to support an async entrypoint directly (and only close the thread once it resolves).

felipellrocha:
is there a safe way for me to keep it open? I suspect I can't just loop infinitely here since that would have the same issue?

Liamolucko:
> throw errors on loop, or just the once?

Just the once, after spawning the future.

felipellrocha:
Worked, thanks! Would've never figured this out on my own! :D

Davidster:
Hello! I am currently encountering this issue and am wondering if it is even possible to fix this in wasm-bindgen? For now I plan on creating a wrapper along these lines:

```rust
fn spawn_thread<F, T>(f: F)
where
    F: FnOnce() -> T,
[...]

PR #4391, opened 2025-01-03
Add additional documentation to spawn_local

drewcrawford:
This adds documentation to help troubleshoot why spawn_local might not work as expected.

see https://github.com/rustwasm/wasm-bindgen/issues/2945 and https://github.com/chemicstry/wasm_thread/issues/6

### DIFF
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -85,6 +85,12 @@ mod task {
 /// # Panics
 ///
 /// This function has the same panic behavior as `future_to_promise`.
+///
+/// # Special considerations
+///
+/// If this function is called from a worker (or non-main thread), the worker must remain running
+/// in order for the passed future to make forward progress.  For more information see
+/// [this](https://github.com/rustwasm/wasm-bindgen/issues/2945) issue.
 #[inline]
 pub fn spawn_local<F>(future: F)
 where


daxpedda (changes requested):
Events, `queueMicrotask`, `setTimeout` and `Scheduler.postTask()` are also affected by this so the issue isn't specific to `spawn_local()`. I think there needs to be a general understanding that blocking in Wasm is much more often than not a bad idea compared to native.

So its probably better to add documentation to functions that block instead to all possible functions that are affected by blocking. In the case of `wasm-bindgen` its `js_sys::Atomics::wait*`. Additionally some top-level documentation somewhere would help make this information be conveyed much more generally.

Thank you for looking into this and going through the effort of a PR!

drewcrawford:
The issue being addressed here is not caused by blocking, so wider awareness of blocking does not address the concern. 

Rather, the issue is caused by *not* blocking: worker threads that never block never have their tasks executed (or stop executing suddenly). 

If this is not appropriate to document then the PR should be closed.

daxpedda:
Ah, I misunderstood the original issue!

Indeed, the issue that you are talking about has nothing to do with `spawn_local()` or `wasm_bindgen`. The issue is that `wasm_thread` decided to close workers after the main task is finished, to more closely mimic the behavior of `std::thread`. So this is a documentation issue on `wasm_thread`s side.

I thought that the issue was that blocking a worker prevents tasks spawned with `spawn_local` & co. to never be run. Which is indeed something related to how JS works and probably could use some documentation in `wasm_bindgen`. But as you said, this is unrelated to this PR.

With that said: closing this PR then.

PR #4393, opened 2025-01-05
Add guide for running doctest

Spxg:
This is what I learned recently when trying to run doctests:
image

image

### DIFF
--- a/guide/src/wasm-bindgen-test/usage.md
+++ b/guide/src/wasm-bindgen-test/usage.md
@@ -128,14 +128,8 @@ Run the tests by passing `--target wasm32-unknown-unknown` to `cargo test`:
 cargo test --target wasm32-unknown-unknown
 ```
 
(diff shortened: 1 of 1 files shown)


guide/src/wasm-bindgen-test/usage.md
```diff
@@ -127,3 +127,15 @@ Run the tests by passing `--target wasm32-unknown-unknown` to `cargo test`:
...
+
+```
+cargo -Zdoctest-xcompile test --target wasm32-unknown-unknown
+```
+
+If you are in the stable channel, you can switch to nightly like this:
+
+```
+cargo +nightly -Zdoctest-xcompile test --target wasm32-unknown-unknown
+```
```
daxpedda, line 141: Instead we should note that his is an unstable feature that requires Rust nightly and merge these two blocks.

daxpedda (changes requested):

daxpedda (approved):
Thank you!

Issue #4385, opened 2024-12-29
Rename arguments of a function


Vexcited:
### Summary

Let's say I have the following method : 
```rust
#[wasm_bindgen]
pub fn update (some_url: &str) -> Self {
  // whatever
}
```

How can I rename the argument `some_url` to, for example, `someUrl` for JS ?

It's possible to do so with structs, functions name, why not arguments individually ?
At least I did not find anything about it in the documentation. 

Thanks you in advance !

daxpedda:
It isn't possible. We might want to consider adding it with together with #1847.

Issue #4377, opened 2024-12-25
allow specifying return type for functions/methods when the rust side is JsValue


rouzwelt:
### Motivation
Atm, because of rust orphan rule we cant directly implement `IntoJsResult` trait for a struct/type wrapper in Result<> and this issue prevents us to use typed returns on async functions as:
```rust
// `A` and `Error` both have all needed trait implementations manualy except `IntoJsResult` for a wrapped `Result<>` of them

#[derive(Serialize, Deserialize, Tsify)]
#[tsify(into_wasm_abi, from_wasm_abi)]
struct A; // notice there is no #[wasm_bindgen] used for this struct as we just wanna get a typed obj in js/ts and not a class obj

#[wasm_bindgen]
pub fn some_function() -> Result<A, Error> {
    /// this compiles ok and works ok
};

#[wasm_bindgen]
pub async fn some_function() -> Result<A, Error> {
    /// this doesnt compile with error complaining about IntoJsResult not implemented for Result<A, Error>
};
```

Same situation for wrapped type of T for `ReturnWasmAbi`:
```rust
pub type B = Result<A, Error>

#[wasm_bindgen]
pub fn some_function() -> Result<B, Error> {
    /// this doesnt compile with error complaining about ReturnWasmAbi not implemented for Result<A, Error>
    /// and same error for async fn along with IntoJsResult error
};
```

### Proposed Solution
- Prefered solution (my personal choice) can be to allow custom ts return type specified like other wasm_bindgen artts like `js_name`, where it would apply that to the return type of the typescript generated output for an standalone fn or a class method, because this issue also effects class methods where you cant return a typed type for a single method and it ends up being `any`, and if you wanna override that, you'd needing to override the whole typescript of the class and all its methods using custom typescript section, instead of just being able to override that single class method. Probably this solution is easiest of the two to implement, as it would just affect the generated .d.ts as the generated js cant really type the return type of a function with a typescript defined obj type.
So this solution can be adopted to provide typings for return types of any function and method for generated typescript for anything that returns JsValue (`any` in js) allowing users to fully type their output pkgs, ie functions/methods can return JsValue on rust side and handling conversion in their rust body, and still generated ts output can have a fully typed return type
```rust
#[derive(Serialize, Deserialize, Tsify)]
#[tsify(into_wasm_abi, from_wasm_abi)]
[...]

PR #4394, opened 2025-01-06
Function Attributes

rouzwelt:
# Motivation
resolves https://github.com/rustwasm/wasm-bindgen/issues/4223
resolves https://github.com/rustwasm/wasm-bindgen/issues/1847
resolves https://github.com/rustwasm/wasm-bindgen/issues/4385
resolves https://github.com/rustwasm/wasm-bindgen/issues/4377

At the moment, by default, exported Rust functions/methods generate function signature from equivalent rust types identifiers and without any documentations, unless completely handcoded using `skip-jsdoc` and/or `typescript-custom-section`, but this has some limitations as disscussed in details in issues mentioned above which have been requested to be addressed for long time, some examples of which are:
- Cannot specify documentation for each function argument or return individually unless handcoded the whole documentation block using `skip-jsdoc` and/or `typescript-custom-section`.
- Cannot return wrapped types (such as `Vec<(Foo, Bar)>`) as return type of async functions (and in some cases non async functions) without wrapping them in new struct and actually implementing `wasm_bindgen` for it.
- Cannot end up having a typed function signature in typescript (not having `any` in typescript) where arguments and return types are only typescript interfaces and not a js class, as `JsValue` in rust side where they are just serialized/deserialized using serde.
- Generated function arguments' names follow rust convention `snake_case`, but js/ts naming convention is `camelCase`, atm there is no way to convert function argument names to `camelCase` or optionally rename them.

# Proposed Solution
By introcusing function attributes these limitations and requested features can be addressed, these attributes will make it possible to override function arguments and return types and to write specific documentation for each of them individually as desired with relatively very minimal changes:
- `#[wasm_bindgen(unchecked_return_type)]` and `#[wasm_bindgen(return_description)]` used to override function's return type and to specify description for generated js/ts bindings.
- `#[wasm_bindgen(js_name)]`, `#[wasm_bindgen(unchecked_param_type)]` and `#[wasm_bindgen(param_description)]` applied to a rust function argument to override that argument's name and type and to specify description for generated js/ts bindings.

These attributes will provide a great level of control and customization over functions/methods generated bindings that are essential for creating well defined and structured bindings.
Let's look at some examples for more details:
```rust
/// Description for foo
#[wasm_bindgen(unchecked_return_type = "Foo", return_description = "some description for return type")]
pub async fn foo(
    #[wasm_bindgen(js_name = "firstArg", param_description = "some description for firstArg")]
    arg1: String,
    #[wasm_bindgen(js_name = "secondArg", unchecked_param_type = "Bar")]
    arg2: JsValue,
) -> Result<JsValue, JsValue> {
    // function body
}
```
This will generate the following js bindings:
```js
/**
* Description for foo
* @param {string} firstArg - some description for firstArg
* @param {Bar} secondArg
* @returns {Promise<Foo>} some description for return type
*/
export function foo(firstArg, secondArg) {
    // generated body
};
```
And will generate the following ts bindings:
```ts
/**
* Description for foo
* @param firstArg - some description for firstArg
* @param secondArg
* @returns some description for return type
*/
export function foo(firstArg: string, secondArg: Bar): Promise<Foo>;
```

Same thing applies to rust struct's (and enums) impl methods and their equivalent js/ts class methods:
```rust
/// Description for Foo
#[wasm_bindgen]
pub struct Foo {
    Foo: String,
}

#[wasm_bindgen]
impl Foo {
    /// Description for foo
    #[wasm_bindgen(unchecked_return_type = "Baz", return_description = "some description for return type")]
    pub fn foo(
        &self, // cannot use function attributes on "self" argument
        #[wasm_bindgen(param_description = "some description for arg1")]
        arg1: String,
        #[wasm_bindgen(unchecked_param_type = "Bar")]
        arg2: JsValue,
    ) -> JsValue {
        // function body
    }
}
```

This will generate the following js bindings:
```js
/**
* Description for Foo
*/
export class Foo {
    /**
    * Description for foo
    * @param {string} arg1 - some description for arg1
    * @param {Bar} arg2
    * @returns {Baz} some description for return type
    */
    foo(arg1, arg2) {
        // generated body
    };
}
```

And will generate the following ts bindings:
```ts
/**
* Description for Foo
*/
export class Foo {
    /**
    * Description for foo
    * @param arg1 - some description for arg1
    * @param arg2
    * @returns some description for return type
    */
    foo(arg1: string, arg2: Bar): Baz;
}
```

# High Level Implementation Details
These attributes can be implemented with very minimal changes, for function arguments attributes, we need to parse and extract them before passing it token stream, and for return type attributes they will be added to `attrgen!` macro in `crates/macro-support/src/parser.rs`, this guarantees that if they were used elsewhere other than a function or a method, it'll produce error or unused code warnings.
Once they are parsed and extracted, they can be stored in the `ast` as follows:
```rust
/// Information about a function's return
pub struct FunctionReturnData {
    /// Specifies the type of the function's return
    pub r#type: syn::Type,
    /// Specifies the JS return type override
    pub js_type: Option<String>,
    /// Specifies the return description
    pub desc: Option<String>,
}

/// Information about a function's argument
pub struct FunctionArgumentData {
    /// Specifies the type of the function's argument
    pub pat_type: syn::PatType,
    /// Specifies the JS argument name override
    pub js_name: Option<String>,
    /// Specifies the JS function argument type override
    pub js_type: Option<String>,
    /// Specifies the argument description
    pub desc: Option<String>,
}
```
and then will be passed around during encode/decode process of a `ast::Function` to `shared_api!` generated `Function` struct which now has a new property that holds those attrs with similar format.
These new structs can be easily extended when/if more features need to be supported.
Next it'll passed as new properties of `AuxExport` converted from `decode::Export`, which makes those attrs available for consumption at `crates/cli-support/js/bindings.rs` where the bindings get generated.
PS: validating `js_name` for an arg, or validating the provided type or description dont contain any illegal chars also takes place inside added logic of `crates/macro-support/parser.rs`.


rouzwelt:
@daxpedda

daxpedda:
Please refrain from pinging maintainers like this unless there is a more specific reason then just reviewing a PR.

guide/src/reference/attributes/on-rust-exports/function-attributes.md
daxpedda, line 1: We should also leave a `**Note**` mentioning that we don't check the type provided by the user, even if the conversion checking and method stays the same. So only because a user uses `param_type = "number"`, it doesn't mean its actually going to be a `number` type we are expecting here.

crates/macro-support/src/parser.rs
```diff
@@ -1153,11 +1158,108 @@ fn function_from_decl(
...
+                keep.push(false);
+                attr.parse_nested_meta(|meta| {
+                    if meta.path.is_ident("js_name") {
+                        if arg_attrs.name.is_some() {
+                            return Err(meta.error("duplicate attribute"));
+                        }
+                        let value = meta.value()?.parse::<syn::LitStr>()?.value();
+                        if is_js_keyword(&value) {
+                            return Err(meta.error("collides with js/ts keywords"));
+                        }
```
daxpedda, line 1196: We should also check that this is a valid JS ident, which I believe we have a function lying around of somewhere.

crates/macro-support/src/parser.rs
```diff
@@ -1153,11 +1158,108 @@ fn function_from_decl(
...
+                        return Ok(());
+                    }
+                    if meta.path.is_ident("param_type") {
+                        if arg_attrs.ty.is_some() {
+                            return Err(meta.error("duplicate attribute"));
+                        }
+                        let value = meta.value()?.parse::<syn::LitStr>()?.value();
+                        if is_js_keyword(&value) {
+                            return Err(meta.error("collides with js/ts keywords"));
+                        }
```
daxpedda, line 1207: I don't know if we have an easy way to check for a valid TS type. But I don't want this to be abused by users to do any funny shenanigans. e.g. `param_type = "\n console.log(\"injecting some JS code into the shim\")"`.

daxpedda (changes requested):
I believe we should not add support for `optional` without a good use-case. While `JsValue` is a good use case, we should instead discuss support for `Option<JsValue>` in a separate issue.

WDYT of wrapping `param_type` and `return_type` with `unsafe(...)`, to make sure users are aware that we can't check the type here?

Thank you for doing this, great work!

crates/backend/src/ast.rs
```diff
@@ -392,6 +392,32 @@ pub struct Function {
     pub generate_jsdoc: bool,
     /// Whether this is a function with a variadict parameter
     pub variadic: bool,
+    /// Function attributes used to provide extra information about function's components
+    pub fn_attrs: Option<FunctionAttributes>,
```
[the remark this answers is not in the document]
rouzwelt: > Can we instead not do the "attribute parsing" inside the proc-macro? So fold this information into `arguments` and `ret`? As we don't need the original argument name as well, we can instead replace `PatType` with `Type` and pass the name separately.

I originally wanted to somehow inject those info into already available props, but then I kinda thought maybe it wont serve the extendability aspect of it, but im totally ok with that if you choose so, just that can you elaborate more, im not sure i fully understand your suggestion, do you mean as some sort of a tuple? for example ret would become: `ret: (Option<syn::Type>, Option<String>, Option<String>)`, so same for `arguments`?

crates/cli-support/src/js/binding.rs
```diff
@@ -382,15 +446,28 @@ impl<'a, 'b> Builder<'a, 'b> {
         (ts, ts_arg_tys, ts_ret, ts_refs)
     }
 
-    /// Returns a helpful JS doc comment which lists types for all parameters
+    /// Returns a helpful JS/TS doc comment which lists types for all parameters
     /// and the return value.
-    fn js_doc_comments(
+    fn js_ts_doc_comments(
```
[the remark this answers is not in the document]
rouzwelt: makes total sense, I'll fix that

crates/macro-support/src/parser.rs
rouzwelt: well im not sure either, Im open to any suggestions, but I assume on js it will only end up in jsdoc, on ts side it can end up in fn signature, i might not be right, but I assume .d.ts has no runtime effect? so if any snippets injected it would be totally ineffective?

rouzwelt:
> I believe we should not add support for `optional` without a good use-case. While `JsValue` is a good use case, we should instead discuss support for `Option<JsValue>` in a separate issue.
> 
Makes sense, I'll remove that.

> WDYT of wrapping `param_type` and `return_type` with `unsafe(...)`, to make sure users are aware that we can't check the type here?
> 
Yea I understand the concerns here, I personally am in the camp of the responsibility of using these attrs are upon the user/dev, as these are not something everyone needs everytime, but are something you might need at times, so using them comes with the responsibility of making sure you know what you are doing, that said, I totally understand the concerns and am totally ok to impl this or any other syntax/format you see fit for this. 

<br>
@daxpedda  Lastly, thanks for identifying and correcting the typos and mistakes I had in documentation, and also I replied to other review comments/suggestion as well, some of which need your final call so I can go ahead to impl, so please let me know how should we go about them.

crates/macro-support/src/parser.rs
rouzwelt: I just asked ChatGPT about it 😄 :
Screenshot 2025-01-07 at 4 26 44 PM

guide/src/reference/attributes/on-rust-exports/function-attributes.md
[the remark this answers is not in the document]
rouzwelt: roger that

guide/src/reference/attributes/on-rust-exports/function-attributes.md
```diff
@@ -0,0 +1,139 @@
...
+It's fair to say that being able to write specific documentation for each argument and return variable of the function, or override an argument or return variable type to a custom one for generated js/ts bindings and many more use cases are essential for creating a well defined, structured and typed bindings.
+
+Function attributes addresses these issues and limitations by providing an ability to override a function's return type and arguments names and types for generated bindings as well as ability to write specific documentation for each of them individually as desired:
+- `#[wasm_bindgen(return_type)]` and `#[wasm_bindgen(return_description)]` used to override function's return type and to specify description for generated js/ts bindings.
+- `#[wasm_bindgen(js_name)]`, `#[wasm_bindgen(param_type)]` and `#[wasm_bindgen(param_description)]` applied to a rust function argument to override that argument's name and type and to specify description for generated js/ts bindings.
+- `#[wasm_bindgen(optional)]` used to tag a function's argument as optional (typescript `?` operator) for function's generated typescript signature.
```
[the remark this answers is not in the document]
rouzwelt: removed

crates/cli-support/src/decode.rs
```diff
@@ -94,6 +94,24 @@ impl<'src, T: Decode<'src>> Decode<'src> for Option<T> {
...
+}
+
+impl Debug for FunctionComponentAttributes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!(
+            "FunctionComponentAttributes {{ ty: {:?}, desc: {:?}, optional: {} }}",
+            self.ty, self.desc, self.optional
+        ))
+    }
+}
```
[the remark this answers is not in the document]
daxpedda: Right. I see we current just do a lot of replication with all of these `Aux` types ... that's horrible. Lets see how it goes after we resolve https://github.com/rustwasm/wasm-bindgen/pull/4394#discussion_r1905138631.

crates/macro-support/src/parser.rs
```diff
@@ -1153,11 +1158,108 @@ fn function_from_decl(
...
             generate_jsdoc: opts.skip_jsdoc().is_none(),
             variadic: opts.variadic().is_some(),
+            fn_attrs,
         },
         method_self,
     ))
 }
 
+/// Extracts function attributes
+fn extract_fn_attrs(
```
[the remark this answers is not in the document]
daxpedda: Please, this is way too much repetitive code otherwise. I believe the unused warning was a mistake, we should have always errored. I will see if we can easily turn these specific attributes into errors after we are done otherwise.

crates/macro-support/src/parser.rs
daxpedda: I unfortunately don't know much about TS, but you should be able to escape JSDoc by just using `param_type = "\n*/ console.log(\"injecting some JS code into the shim\")\n/*"`. Or what am I missing?

In any case, we can look at it in the end.

daxpedda:
> I can think of this format as well:
> `unchecked_param_type` and `unchecked_return_type`

This might be more appropriate actually. Let go with that, we can do more bikeshedding after we are done.

crates/macro-support/src/parser.rs
rouzwelt: oh yes you're right, very good catch, I'll think about it, I guess one way can be to restrict to only valid chars and possibly restrict the length, anyways we can come back to this once other stuff are done with.

rouzwelt: im wondering how it is enforced on `skip_jsdoc` and a manual jsdco through rust comments, I'll have a look, maybe there is something there

crates/macro-support/src/parser.rs
rouzwelt: using `BindgenAttrs::find` should already handle that, as it is using the attrgen `js_name` logic

crates/backend/src/ast.rs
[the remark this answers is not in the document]
rouzwelt: trying to do this, ended up being a massive rabbit hole, like it ends up needing bunch of chnages in bunch of crates, basically bunch of things downstream needs proper and carefull updating which I dont think Im the best person to do it, it def needs someone who is well familiar with all the details of this repo. It also breaks some parsing stuff as well in js-sys crate and codeine crate. 
I suggest we keep the current setup, ie the one I originally proposed, and maybe it can be refactored/handled in a sep PR and merged into `arguments` and `ret` props, @daxpedda what do you think?

crates/macro-support/src/parser.rs
rouzwelt: well actually this can also happen for `param_description` and `return_description`

crates/macro-support/src/parser.rs
rouzwelt: swicthed to useing BindgenAttrs::find, so it should already handle that, as it is using the attrgen js_name logic

crates/macro-support/src/parser.rs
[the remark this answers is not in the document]
rouzwelt: swicthed to using 'BindgenAttrs::find()'

crates/backend/src/ast.rs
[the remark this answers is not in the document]
rouzwelt: actually I was able to work it out, but slightly different, thats is, not changing the PatType to Type, but keeping it as is inside the new struct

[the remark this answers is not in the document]
rouzwelt: let me know your thoughts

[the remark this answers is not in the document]
daxpedda: I can take a stab at it when the PR is ready again.

> @daxpedda what do you think?

Please stop pinging me. Every time you write "@daxpedda" I get a priority notification.

crates/shared/src/lib.rs
```diff
@@ -141,6 +141,17 @@ macro_rules! shared_api {
             generate_typescript: bool,
             generate_jsdoc: bool,
             variadic: bool,
+            fn_attrs: Option<FunctionAttributes>,
+        }
+
+        struct FunctionAttributes {
+            ret: FunctionComponentAttributes,
+            args: Vec<FunctionComponentAttributes>,
```
daxpedda, line 149: Lets fold this into `arg_names` again.
The custom return type and description can be split into two fields instead of the struct which won't be reused with args anymore.

crates/macro/ui-tests/unused-attributes.rs
daxpedda, line 1: Lets check for duplicate attributes on the same argument as well. But probably in its own file.

guide/src/reference/attributes/on-rust-exports/function-attributes.md
```diff
@@ -0,0 +1,139 @@
...
+```ts
+export interface Person { name: string; }
+
+/**
+* @param personName - Specifies the person's name
+* @returns a Person object
+*/
+export function build_person(personName: string): Promise<Person>;
+```
+As you can see, using function attributes, we can now return a js/ts object without ever using `wasm_bindgen` macro on `Person` struct that would have generated a js/ts class object in the bindings which might not be the desired outcome for every case, you can see as well that instead of handcoding the full documentation for `build_person()` with all js_doc/ts_doc tags and syntax, we can just write specific docs for each individual component of the function and as a result end up with a fully typed and documented function in the bindings.
```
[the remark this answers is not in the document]
daxpedda: I meant the example and not just the reference to `tsify`. `serde_wasm_bindgen` is still an external crate.

daxpedda (changes requested):

crates/macro-support/src/parser.rs
daxpedda: Lets add a UI test for that as well while we are at it.

crates/backend/src/ast.rs
```diff
@@ -375,9 +375,9 @@ pub struct Function {
     /// Whether the function has a js_name attribute
     pub renamed_via_js_name: bool,
     /// The arguments to the function
-    pub arguments: Vec<syn::PatType>,
-    /// The return type of the function, if provided
-    pub ret: Option<syn::Type>,
+    pub arguments: Vec<FunctionArgumentData>,
+    /// The data of return type of the function
+    pub ret: FunctionReturnData,
```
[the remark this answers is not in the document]
rouzwelt: should it be? it has moved inside the struct, and is Option there, but the struct itself for `ret` isnt Option anymore, I can make it Option if that should be the case
Screenshot 2025-01-08 at 6 06 00 PM

crates/macro-support/src/parser.rs
rouzwelt: ok, I think we can enforce it by not allowing any comment open/close syntax, ie `/* */`, what do you think?

crates/backend/src/ast.rs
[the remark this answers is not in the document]
daxpedda: Yes, this should be an `Option` and the `Option` on `r#type` should be removed.

crates/shared/src/lib.rs
daxpedda: No two individual fields. No reason to make a new struct if its only used in a single place.

crates/backend/src/ast.rs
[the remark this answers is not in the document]
rouzwelt: ok so in that case, we should ignore the description and type override from attrs if the return of the function is not specified, because:
Screenshot 2025-01-08 at 8 55 31 PM
as you can see this returns a Option<Type> for ret, and yet still if it is None, one can still use description and type override using attrs, so if we only have Option on outter struct, then the inner possible description and type override will be ignored, is that ok with you?

crates/macro/ui-tests/unused-attributes.rs
rouzwelt: added

crates/backend/src/ast.rs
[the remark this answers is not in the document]
daxpedda: We should actually error when using `return_description` or `unchecked_return_type` if there is no return type. Right?

crates/macro-support/src/parser.rs
rouzwelt: I wasnt able to find an function that validates a js identifier besides `is_js_keyword()`, can you point that for me if it exists?

crates/backend/src/ast.rs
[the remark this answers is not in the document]
rouzwelt: yea makes sense

crates/macro-support/src/parser.rs
daxpedda: Apparently we don't. But I thought by using `BindgenAttrs::find()` we don't need to do this anymore? Looking at it more closely its because we use `syn::Ident`. While JS is a bit more lax, by allowing `_` at the start and `$` anywhere, its fine for now.

But if you really want to add full support, we should not use `regex`, but [`unicode-ident`](https://crates.io/crates/unicode-ident). See the [specification](https://tc39.es/ecma262/#sec-names-and-keywords) as well.

crates/macro/ui-tests/unused-attributes.rs
daxpedda, line 1: Now that we made those attributes error, we should move them to a separate test.

crates/shared/src/lib.rs
```diff
@@ -135,12 +135,20 @@ macro_rules! shared_api {
...
         struct Function<'a> {
-            arg_names: Vec<String>,
+            args: Vec<FunctionArgumentData>,
             asyncness: bool,
             name: &'a str,
             generate_typescript: bool,
             generate_jsdoc: bool,
             variadic: bool,
+            ret_ty_override: Option<String>,
+            ret_desc: Option<String>,
```
daxpedda, line 145:
```suggestion
            ret_ty_override: Option<&'a str>,
            ret_desc: Option<&'a str>,
```

crates/shared/src/lib.rs
```diff
@@ -135,12 +135,20 @@ macro_rules! shared_api {
...
+        }
+
+        struct FunctionArgumentData {
+            name: String,
+            ty_override: Option<String>,
+            desc: Option<String>,
```
daxpedda, line 151:
```suggestion
        struct FunctionArgumentData<'a> {
            name: String,
            ty_override: Option<&'a str>,
            desc: Option<&'a str>,
```
This will not make it work in `AuxExport` anymore, so we need a new type there. E.g. `AuxFunctionArgumentData`.

guide/src/reference/attributes/on-rust-exports/function-attributes.md
```diff
@@ -0,0 +1,102 @@
+# `function-attributes`
+
+By default, exported Rust functions and methods generate function signature from equivalent Rust types identifiers without any arguments and return variable documentations. However by using function attributes, it's possible to override a function's return type and arguments names and types for generated bindings, as well as the ability to write specific documentation for each of them individually as desired:
```
daxpedda, line 3:
```suggestion
By default, exported Rust functions and methods generate function signature from equivalent Rust types identifiers without any arguments and return variable documentations. However by using function attributes, it's possible to override a function's return type and argument names and types for generated bindings, as well as the ability to write specific documentation for each of them individually as desired:
```

daxpedda (changes requested):

crates/cli-support/src/decode.rs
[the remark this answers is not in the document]
daxpedda: See https://github.com/rustwasm/wasm-bindgen/pull/4394#discussion_r1908488825.

crates/macro-support/src/parser.rs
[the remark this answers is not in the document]
daxpedda: Fixed this in https://github.com/rustwasm/wasm-bindgen/pull/4394/commits/e7f2618b0d93236edb8468bf4210fbe301a21387.

crates/macro-support/src/parser.rs
```diff
@@ -166,6 +166,10 @@ macro_rules! attrgen {
             (static_string, StaticString(Span)),
             (thread_local, ThreadLocal(Span)),
             (thread_local_v2, ThreadLocalV2(Span)),
+            (unchecked_return_type, ReturnType(Span, String, Span)),
```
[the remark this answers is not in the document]
rouzwelt: use `Ident` for return type and possibly arg type? hmm, I dont think that really works, does it? because the whole idea and one of the main use cases is to be able to do stuff like `SomeType[]`, or `SomeType | undefined` etc.
I dont think `Ident` can cover that, can it?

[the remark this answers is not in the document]
daxpedda: Right, I think I meant for this to go to `js_name`, which doesn't make any sense, sorry!

crates/macro-support/src/parser.rs
daxpedda: No, for `js_name` we use `String`, not `Ident`.
But I found the relevant function: https://github.com/rustwasm/wasm-bindgen/blob/c9ee617e9a63340bfba0cecd7d53ba5262f25ddc/crates/cli-support/src/js/identifier.rs#L31-L42

crates/cli-support/src/js/binding.rs
```diff
@@ -222,17 +230,21 @@ impl<'a, 'b> Builder<'a, 'b> {
         // }
 
         let mut code = String::new();
+        let args_names = function_args
+            .iter()
+            .map(|v| v.name.as_str())
+            .collect::<Vec<&str>>();
```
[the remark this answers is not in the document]
rouzwelt: I think it is isnt it? `function_args` is now a vec of `FunctionArgumentData`, so the logic that comes right after this line works with vec of strings, I didnt want to really touch that logic, hence it is mapped into a vec of strings to be able to be used on that logic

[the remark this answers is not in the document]
rouzwelt: Screenshot 2025-01-09 at 4 10 35 PM

crates/cli-support/src/js/binding.rs
```diff
@@ -386,54 +432,133 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+            if let Some(v) = desc {
+                arg.push_str(" - ");
+                arg.push_str(v);
+            }
+
+            arg.push('\n');
+            ts_doc_args.push(arg);
+        }
+
+        let mut ret: String = ts_doc_args.into_iter().collect();
```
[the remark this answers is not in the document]
rouzwelt: yes ofc, we can, I personally tried to stick with the logic that was there in place for js_doc, just modify parts of it that were necessary for ts_doc, but yes ofc we can do as you suggested, I'll fix that

crates/cli-support/src/js/binding.rs
```diff
@@ -386,54 +432,133 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+    /// the return value descriptions.
+    fn ts_doc_comments(
+        &self,
+        args_data: &[FunctionArgumentData],
+        ret_desc: &Option<String>,
+    ) -> String {
+        let mut ts_doc_args = Vec::new();
+        // ofc we dont need arg type for ts doc, only arg name
+        for FunctionArgumentData { name, desc, .. } in args_data.iter() {
+            let mut arg = "@param ".to_string();
```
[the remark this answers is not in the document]
rouzwelt: as explained above, I'll fix that

crates/macro-support/src/parser.rs
rouzwelt: > No, for `js_name` we use `String`, not `Ident`. But I found the relevant function:
> 
> https://github.com/rustwasm/wasm-bindgen/blob/c9ee617e9a63340bfba0cecd7d53ba5262f25ddc/crates/cli-support/src/js/identifier.rs#L31-L42

ok awesome, as I see this is in cli-support crate, and not in macro-support crate, shall we copy/move this into macro-support crate and use it there for validation? or add a dep of cli-support to macro-support crate and use it there, or should the validation take place in cli-support crate, so I assume if thats the case, it would throw error on using cli bin and not at macro expand time where for example Rustanalyzer can show it

crates/cli-support/src/js/binding.rs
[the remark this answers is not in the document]
daxpedda: I assume you mean because of this line?
https://github.com/rustwasm/wasm-bindgen/blob/c9ee617e9a63340bfba0cecd7d53ba5262f25ddc/crates/cli-support/src/js/binding.rs#L235
This should should be easily replaced with a `iter().map().reduce()` or `fold()`.

I appreciate you trying to touch as little code as possible. Its really nice to review! But that can be done by separating commits, we don't want to degrade the code in pursuit of making the code easy to review.

While I think the `slice::join()` is very convenient, we should try to avoid unnecessary memory allocations if possible.

crates/macro-support/src/parser.rs
rouzwelt: I personally like to see it at macro expand time, so maybe this logic can be moved to backend or shared crate, so it can be accessed in multiple places, what do you think?

daxpedda: Yes, should definitely happen in the macro. Backend I believe is the right place.

daxpedda: Yep, `wasm-bindgen-shared` is only a dependency of the proc-macro (and CLI) but not of `wasm-bindgen` directly. So should be fine!

crates/macro-support/src/parser.rs
```diff
@@ -1027,6 +1036,7 @@ fn function_from_decl(
     attrs: Vec<syn::Attribute>,
     vis: syn::Visibility,
     position: FunctionPosition,
+    args_attrs: Option<Vec<FnArgAttrs>>,
```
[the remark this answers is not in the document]
daxpedda: I see, thank you for the explanation! Lets leave it at that then.

crates/shared/src/lib.rs
rouzwelt: is this necessary/needed? we can just use shared `FunctionArgumentData` here, instead of just replicating it as Aux again

crates/macro-support/src/parser.rs
rouzwelt: ok so shall I move it to `shared` crate?

crates/cli-support/src/js/binding.rs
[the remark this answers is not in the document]
rouzwelt: ok understandable, I'l fix that

crates/shared/src/lib.rs
daxpedda: Yes, we are reducing allocations and keeping consistent with the existing code. Ideally, we avoid allocations even in `AuxExport`, but I assume that would require a much bigger refactor.

FWIW: In big projects `wasm-bindgen` can get quite slow, I've seen up to 12s to post-process locally (on an Intel 13900K). The biggest bottleneck we can do anything about are allocations. While these won't save us any significant time, we shouldn't start making changes that make the situations actively worse either.

crates/macro-support/src/parser.rs
daxpedda: Yes please! Thanks!

crates/shared/src/lib.rs
rouzwelt: understood, I'll apply the chnages

crates/macro-support/src/parser.rs
rouzwelt: moved to `shared` crate

daxpedda (changes requested):
LGTM! Just small nits left.

The last thing is just a changelog entry.
Thank you for the great work!

crates/macro-support/src/parser.rs
```diff
@@ -1139,25 +1205,104 @@ fn function_from_decl(
...
+                                span,
+                                "collides with js/ts keyword",
+                            ));
+                        }
+                        if !is_valid_ident(js_name_override) {
+                            return Err(Diagnostic::span_error(
+                                span,
+                                "invalid js/ts argument identifier",
+                            ));
+                        }
```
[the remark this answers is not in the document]
rouzwelt: does this work? or you'd want functions perse?
Screenshot 2025-01-11 at 11 42 27 PM

crates/macro-support/src/parser.rs
```diff
@@ -1122,10 +1146,52 @@ fn function_from_decl(
...
+            js_type: ret_ty_override.as_ref().map_or(Ok(None), |(ty, span)| {
+                if is_js_keyword(ty) {
+                    return Err(Diagnostic::span_error(*span, "collides with js/ts keyword"));
+                }
+                if contains_js_comment_close(ty) {
+                    return Err(Diagnostic::span_error(
+                        *span,
+                        "contains illegal comment close syntax",
+                    ));
+                }
```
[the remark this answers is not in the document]
rouzwelt: check below

rouzwelt:
> LGTM! Just small nits left.
> 
> The last thing is just a changelog entry. Thank you for the great work!

Thank you too for very joyful collaboration.
Its been a great experience, and Im so thrilled by being a small contributer to this project.

crates/macro-support/src/parser.rs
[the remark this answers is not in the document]
daxpedda: Functions, I already pushed the change!

daxpedda (approved):
I re-worked the documentation a bit.
Again: thanks for the great work!

rouzwelt:
> I re-worked the documentation a bit.
> Again: thanks for the great work!

thank you.
one last question, when can we expect a release for this?

daxpedda:
Probably today.

rouzwelt:
> Probably today.

awesome, thanks again, been a pleasure.

PR #4402, opened 2025-01-09
Add a deprecation notice about async constructors

### LINKED ISSUE #3976: Disallow async constructors
### Describe the Bug

wasm_bindgen allows you to create an async constructor even though this is not supported by typescript.

### Steps to Reproduce

```rust

#[wasm_bindgen]
pub struct MyStruct {
	data: u32
}

#[wasm_bindgen]
impl MyStruct {
  #[wasm_bindgen(constructor)]
  pub async fn new(data: u32) -> Self {
     // do something async
     Self { data }
  }
}
```

This compiles and actually works, however seems like undefined behavior and should just fail to compile

### Expected Behavior

Unknown, probably fail to compile

### Actual Behavior

Works

danielronnkvist:
Adds a deprecation attribute to the method marked as constructor if it is async.

Appreciate suggestions on wording in the deprecation notice.

Resolves #3976

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## Unreleased
+
+### Changed
+
+* Deprecation warning when using an async function as a constructor
+  [#4402](https://github.com/rustwasm/wasm-bindgen/pull/4402)
+
+
 ## [0.2.100](https://github.com/rustwasm/wasm-bindgen/compare/0.2.99...0.2.100)
 
 Released 2025-01-12
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -811,6 +811,16 @@ impl TryToTokens for ast::Export {
                     add_check(quote! {
                         let _: #wasm_bindgen::__rt::marker::CheckSupportsConstructor<#class>;
                     });
+
+                    if self.function.r#async {
+                        (quote_spanned! {
+                            self.function.name_span =>
+                            #[deprecated(note = "async constructors are not supported")]
+                            fn constructor() {}
+                            constructor();
+                        })
+                        .to_tokens(into);
+                    }
                 }
                 ast::MethodKind::Operation(operation) => match operation.kind {
                     ast::OperationKind::Getter(_) | ast::OperationKind::Setter(_) => {
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -815,9 +815,11 @@ impl TryToTokens for ast::Export {
                     if self.function.r#async {
                         (quote_spanned! {
                             self.function.name_span =>
-                            #[deprecated(note = "async constructors are not supported")]
-                            fn constructor() {}
-                            constructor();
+                            const _: () = {
+                                #[deprecated(note = "async constructors produce invalid TS code and support will be removed in the future")]
+                                const fn constructor() {}
+                                constructor();
+                            };
                         })
                         .to_tokens(into);
                     }
--- a/crates/macro/ui-tests/unsupported-options.rs
+++ b/crates/macro/ui-tests/unsupported-options.rs
@@ -79,8 +79,4 @@ impl NonWasmType {
     pub fn static_method() {}
 }
 
-fn main() {
-    async {
-        RustStruct::new().await;
-    };
-}
+fn main() {}
--- a/crates/macro/ui-tests/unsupported-options.rs
+++ b/crates/macro/ui-tests/unsupported-options.rs
@@ -12,7 +12,7 @@ impl RustStruct {
     pub fn static_method() {}
 
     #[wasm_bindgen(constructor)]
-    pub fn new() -> Self {
+    pub async fn new() -> Self {
         Self { data: 0 }
     }
 
@@ -79,4 +79,8 @@ impl NonWasmType {
     pub fn static_method() {}
 }
 
-fn main() {}
+fn main() {
+    async {
+        RustStruct::new().await;
+    };
+}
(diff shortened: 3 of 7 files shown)


daxpedda (changes requested):
Please add a test in `crates/macro/ui-tests`. We can take a second look at the wording then.

daxpedda (changes requested):
Ah, unfortunately this doesn't seem to look right after all. Sorry!
The problem is we need a deprecation warning at the definition site, not the call site. Because realistically the call size will be in JS, not Rust.

This should work:
```diff
diff --git a/crates/backend/src/codegen.rs b/crates/backend/src/codegen.rs
index c8decb4a1..19bc75afd 100644
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -806,6 +806,14 @@ impl TryToTokens for ast::Export {
                     add_check(quote! {
                         let _: #wasm_bindgen::__rt::marker::CheckSupportsConstructor<#class>;
                     });
+
+                    if self.function.r#async {
+                        (quote_spanned! {
+                            self.function.name_span =>
+                            let async_constructors_deprecated: ();
+                        })
+                        .to_tokens(into);
+                    }
                 }
                 ast::MethodKind::Operation(operation) => match operation.kind {
                     ast::OperationKind::Getter(_) | ast::OperationKind::Setter(_) => {
```

I wanted to push this directly but didn't have permissions, but I can open a PR in your repo if this makes things easier.

danielronnkvist:
@daxpedda I see, good point! But what do you think about still using the deprecated attribute? In my opinion, that outputs a better type of warning for this scenario than an unused variable.

```diff
diff --git a/crates/backend/src/codegen.rs b/crates/backend/src/codegen.rs
index c8decb4a..e8ff0bcf 100644
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -806,6 +806,16 @@ impl TryToTokens for ast::Export {
                     add_check(quote! {
                         let _: #wasm_bindgen::__rt::marker::CheckSupportsConstructor<#class>;
                     });
+
+                    if self.function.r#async {
+                        (quote_spanned! {
+                            self.function.name_span =>
+                            #[deprecated(note = "async constructors are not supported")]
+                            fn constructor() {}
+                            constructor();
+                        })
+                        .to_tokens(into);
+                    }
                 }
                 ast::MethodKind::Operation(operation) => match operation.kind {
                     ast::OperationKind::Getter(_) | ast::OperationKind::Setter(_) => {
```

daxpedda:
Yes, definitely better!

tests/wasm/futures.rs
```diff
@@ -123,7 +123,6 @@ pub struct AsyncStruct;
 
 #[wasm_bindgen]
 impl AsyncStruct {
-    #[wasm_bindgen(constructor)]
```
daxpedda, line 126: We need to keep this test until we actually remove the feature.
```suggestion
    #[wasm_bindgen(constructor)]
    #[allow(deprecated)]
```

crates/macro/ui-tests/unsupported-options.rs
```diff
@@ -79,4 +79,8 @@ impl NonWasmType {
     pub fn static_method() {}
 }
 
-fn main() {}
+fn main() {
+    async {
+        RustStruct::new().await;
+    };
```
daxpedda, line 85: I assume this can be removed now.

crates/backend/src/codegen.rs
```diff
@@ -806,6 +806,16 @@ impl TryToTokens for ast::Export {
                     add_check(quote! {
                         let _: #wasm_bindgen::__rt::marker::CheckSupportsConstructor<#class>;
                     });
+
+                    if self.function.r#async {
+                        (quote_spanned! {
+                            self.function.name_span =>
+                            #[deprecated(note = "async constructors are not supported")]
+                            fn constructor() {}
+                            constructor();
```
daxpedda, line 815: This context is shared with user code, so we need to avoid naming collisions here. I tried this locally and it seems to still work just fine.
```suggestion
                            const _: () = {
                                #[deprecated(note = "async constructors are not supported")]
                                const fn constructor() {}
                                constructor();
                            };
```

crates/backend/src/codegen.rs
```diff
@@ -806,6 +806,16 @@ impl TryToTokens for ast::Export {
                     add_check(quote! {
                         let _: #wasm_bindgen::__rt::marker::CheckSupportsConstructor<#class>;
                     });
+
+                    if self.function.r#async {
+                        (quote_spanned! {
+                            self.function.name_span =>
+                            #[deprecated(note = "async constructors are not supported")]
```
daxpedda, line 813:
```suggestion
                            #[deprecated(note = "async constructors produce invalid TS code and support will be removed in the future")]
```

daxpedda (changes requested):
Apart from the outstanding issue it LGTM!
Thank you again!

daxpedda (approved):

CHANGELOG.md
```diff
@@ -43,6 +43,9 @@
 * `console.*()` calls in tests are now always intercepted by default. To show them use `--nocapture`. When shown they are always printed in-place instead of after test results, analogous to `cargo test`.
   [#4356](https://github.com/rustwasm/wasm-bindgen/pull/4356)
 
+* Deprecation warning when using an async function as a constructor
+  [#4402](https://github.com/rustwasm/wasm-bindgen/pull/4402)
```
daxpedda, line 47: One last change: the PR needs a rebase and the new changelog entry be moved to a new "Unreleased" section.

daxpedda (changes requested):

daxpedda (approved):
Thank you!

PR #4409, opened 2025-01-16
Expand `#[expect]` attributes into `#[allow]`

samueltardieu:
An `#[expect]` outer attribute must not be kept in the `#[automatically_derived]` code, as the expanded code will probably not meet the expectations. Transforming it into the equivalent `#[allow]` attribute is the safest way to ensure that expected attributes will not trigger a warning.

Noticed through a bug report entered at the Clippy repository at <https://github.com/rust-lang/rust-clippy/issues/14008>.


samueltardieu:
Ping?

samueltardieu:
Anyone alive here?

PR #4434, opened 2025-02-15
Add `WASM_BINDGEN_TEST_WEBDRIVER_JSON` var to override `webdriver.json` path

UebelAndre:
This is handy supporting various machines with different browsers and configurations installed. I'm able to update my CI to point to specific configs per machine that's matched that all contain unique settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 * Added `DocumentOrShadowRoot.adoptedStyleSheets`.
   [#4625](https://github.com/wasm-bindgen/wasm-bindgen/pull/4625)
 
+* Added ability to determine `webdriver.json` config location via
+  `WASM_BINDGEN_TEST_WEBDRIVER_JSON` environment variable to
+  `wasm-bindgen-test`.
+  [#4434](https://github.com/wasm-bindgen/wasm-bindgen/pull/4434)
+
 ### Changed
 
 * Fixed wrong method names for `GestureEvent` bindings.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * Added `DocumentOrShadowRoot.adoptedStyleSheets`.
   [#4625](https://github.com/wasm-bindgen/wasm-bindgen/pull/4625)
 
-* Added ability to determine `webdriver.json` config location via
+* Added ability to determine WebDriver JSON config location via
   `WASM_BINDGEN_TEST_WEBDRIVER_JSON` environment variable to
   `wasm-bindgen-test`.
   [#4434](https://github.com/wasm-bindgen/wasm-bindgen/pull/4434)
--- a/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
+++ b/crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
@@ -127,7 +127,7 @@ pub fn run(
     };
     println!("Try find `webdriver.json` for configure browser's capabilities:");
     let capabilities: Capabilities = match File::open(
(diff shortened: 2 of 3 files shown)


UebelAndre:
@daxpedda friendly ping on this one

UebelAndre:
@daxpedda one more ping here 🙏

crates/cli/src/bin/wasm-bindgen-test-runner/headless.rs
```diff
@@ -126,7 +126,9 @@ pub fn run(
         session: None,
     };
     println!("Try find `webdriver.json` for configure browser's capabilities:");
-    let capabilities: Capabilities = match File::open("webdriver.json") {
+    let capabilities: Capabilities = match File::open(
+        std::env::var("WASM_BINDGEN_WEBDRIVER_JSON").unwrap_or("webdriver.json".to_string()),
```
daxpedda:
```suggestion
        std::env::var("WASM_BINDGEN_TEST_WEBDRIVER_JSON").unwrap_or("webdriver.json".to_string()),
```

daxpedda (changes requested):
I think this is a good idea, lets get it in!

Still needs an entry in the guide and changelog.

UebelAndre:
> I think this is a good idea, lets get it in!
> 
> Still needs an entry in the guide and changelog.

@daxpedda Updated!

guide/src/wasm-bindgen-test/browsers.md
```diff
@@ -64,8 +64,9 @@ wasm-pack test --headless --chrome --firefox --safari
 
 ## Configuring Headless Browser capabilities
 
-Add the file `webdriver.json` to the root of your crate. Each browser has own 
-section for capabilities. For example:
+Either add the file `webdriver.json` to the root of your crate or ensure an environment
+variable `WASM_BINDGEN_TEST_WEBDRIVER_JSON` is set and points to a `webdriver.json` file.
```
daxpedda:
```suggestion
Either add the file `webdriver.json` to the root of your crate or ensure the environment
variable `WASM_BINDGEN_TEST_WEBDRIVER_JSON` points to one.
```

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 * Added `DocumentOrShadowRoot.adoptedStyleSheets`.
   [#4625](https://github.com/wasm-bindgen/wasm-bindgen/pull/4625)
 
+* Added ability to determine `webdriver.json` config location via
```
```suggestion
* Added ability to determine WebDriver JSON config location via
```

daxpedda (changes requested):
We should probably avoid talking specifically about `webdriver.json` for `WASM_BINDGEN_TEST_WEBDRIVER_JSON` as the environment variables can point to a file with an arbitrary name.

UebelAndre:
> We should probably avoid talking specifically about `webdriver.json` for `WASM_BINDGEN_TEST_WEBDRIVER_JSON` as the environment variables can point to a file with an arbitrary name.

@daxpedda Is there more I should do outside of your suggestions?

daxpedda:
> > We should probably avoid talking specifically about `webdriver.json` for `WASM_BINDGEN_TEST_WEBDRIVER_JSON` as the environment variables can point to a file with an arbitrary name.
> 
> daxpedda Is there more I should do outside of your suggestions?

No, we are all set!

(please avoid pinging on every message)

UebelAndre:
> > > We should probably avoid talking specifically about `webdriver.json` for `WASM_BINDGEN_TEST_WEBDRIVER_JSON` as the environment variables can point to a file with an arbitrary name.
> > 
> > 
> > daxpedda Is there more I should do outside of your suggestions?
> 
> No, we are all set!
> 
> (please avoid pinging on every message)

Fixed conflicts!

(And sorry, this is a common thing in the communities I normally engage in)

PR #4435, opened 2025-02-17
Generate dwarf for tests by default

Spxg:
Generate dwarf for tests by default, which is useful when debugging a lib crate with test cases.


guybedford:
Curious why this one was closed?

Spxg:
> Curious why this one was closed?

It was because

```rust
/// By default this flag is `false`. Note that any emitted DWARF is
/// currently wildly incorrect and buggy
```

But after thinking about it, the incorrect DWARF information should not affect the test run. I will open it again later.

guybedford:
Ahh, that actually changed in https://github.com/wasm-bindgen/walrus/pull/231. I mean, it's still potentially incorrect and buggy - but perhaps slightly less wildly these days.

PR #4437, opened 2025-02-17
`#[wasm_bindgen_test]`: Add an `#[allow(dead_code)]` to the original function

its-the-shrimp:
I've been working on a PR for [Yew](https://github.com/yewstack/yew), and our clippy tests were failing because `wasm-bindgen-test` would leave the test function unused if the compilation target wasn't Wasm. This PR attempts to address that with a simple fix.


its-the-shrimp:
How do I fix CI here? I'm not sure what I did wrong...

RReverser (approved):

PR #4443, opened 2025-02-28
add emscripten support

walkingeyerobot:
Adds `--target emscripten` option, which will output JS that Emscripten can consume and integrate into its own JS.

This is still a work-in-progress, but it's getting kind of beefy, so I thought it would be best to start review on this soon. There's still feature work to be done (i.e. I don't think futures work). My preference would be to submit this with only the current set of supporting features, mark `--target emscripten` as experimental, and then send smaller PRs to add more support.

Also adds a basic test. This test builds some rust code and links it with Emscripten, but the test runner invokes wasm-bindgen. It then tests to make sure the output looks reasonable. Normally, the wasm-bindgen cli tool is invoked by Emscripten at link time. Because cargo does not support binary dependencies, and because wasm-bindgen is effectively being used as a dependency of Emscripten, it's very difficult to do end-to-end testing in the wasm-bindgen repository. I'd like to add more comprehensive testing, but I think it's more appropriate to either Emscripten or to a new repository separate from wasm-bindgen and Emscripten that can depend on both of those repositories at HEAD.

It turns out wasm-bindgen also does not support Emscripten exceptions or PIC, so we've disabled both of those compiler options. This is mostly due to lack of features in the interpreter. Exception support is something I'd like to eventually add. PIC appears to be off by default for `wasm32-unknown-unknown` and on by default for `wasm32-unknown-emscripten`. PIC can be supported through changes to wasm-bindgen's interpreter, but that seemed a bit out of scope for this PR.

I've also made a corresponding PR to Emscripten: https://github.com/emscripten-core/emscripten/pull/23493

I've also made a PR to wasm-pack: https://github.com/rustwasm/wasm-pack/pull/1469. However, there has been no activity in wasm-pack by maintainers for ~4 months, and I suspect it's unlikely to get attention in the near future. The purpose of the wasm-pack PR is to provide a more straightforward workflow for Rust users already familiar with targeting Wasm.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,10 @@
 [target.'cfg(target_arch = "wasm32")']
 runner = 'cargo run -p wasm-bindgen-cli --bin wasm-bindgen-test-runner --'
+
+[build]
+[target.'cfg(all(target_arch = "wasm32", target_os = "emscripten"))']
+rustflags = [
+    "-Cllvm-args=-enable-emscripten-cxx-exceptions=0", 
+    "-Clink-arg=-Wno-undefined",    
+    "-Crelocation-model=static",                                                                                                                                                                                                     
+]
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -1661,7 +1661,7 @@ impl Invocation {
                     *log_error = false;
                 }
                 let ret = cx.invoke_import(import, kind, args, variadic, prelude, import_deps);
-                return ret
+                return ret;
             }
         }
     }
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -130,6 +130,7 @@ impl<'a, 'b> Builder<'a, 'b> {
         debug_name: &str,
         ret_ty_override: &Option<String>,
         ret_desc: &Option<String>,
+        import_deps: &mut Vec<String>,
     ) -> Result<JsFunction, Error> {
         if self
             .cx
@@ -194,6 +195,7 @@ impl<'a, 'b> Builder<'a, 'b> {
                 &instr.instr,
                 &mut self.log_error,
                 &self.constructor,
+                import_deps,
             )?;
         }
 
@@ -727,6 +729,7 @@ fn instruction(
     instr: &Instruction,
     log_error: &mut bool,
     constructor: &Option<String>,
+    import_deps: &mut Vec<String>,
 ) -> Result<(), Error> {
     fn wasm_to_string_enum(name: &str, index: &str) -> String {
         // e.g. ["a","b","c"][someIndex]
@@ -807,7 +810,7 @@ fn instruction(
             }
 
             // Call the function through an export of the underlying module.
-            let call = invoc.invoke(js.cx, &args, &mut js.prelude, log_error)?;
+            let call = invoc.invoke(js.cx, &args, &mut js.prelude, log_error, import_deps)?;
 
             // And then figure out how to actually handle where the call
             // happens. This is pretty conditional depending on the number of
@@ -1637,6 +1640,7 @@ impl Invocation {
         args: &[String],
         prelude: &mut String,
         log_error: &mut bool,
+        import_deps: &mut Vec<String>,
     ) -> Result<String, Error> {
         match self {
             Invocation::Core { id, .. } => {
@@ -1656,7 +1660,8 @@ impl Invocation {
                 if cx.import_never_log_error(import) {
                     *log_error = false;
                 }
-                cx.invoke_import(import, kind, args, variadic, prelude)
+                let ret = cx.invoke_import(import, kind, args, variadic, prelude, import_deps);
+                return ret
             }
         }
     }
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -10,6 +10,7 @@ use crate::wit::{JsImport, JsImportName, NonstandardWitSection, WasmBindgenAux};
 use crate::{reset_indentation, Bindgen, EncodeInto, OutputMode, PLACEHOLDER_MODULE};
 use anyhow::{anyhow, bail, Context as _, Error};
 use binding::TsReference;
+use regex::Regex;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::fmt;
@@ -18,7 +19,6 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use walrus::{FunctionId, ImportId, MemoryId, Module, TableId, ValType};
 use wasm_bindgen_shared::identifier::is_valid_ident;
-use regex::Regex;
 
 mod binding;
 
@@ -241,8 +241,7 @@ impl<'a> Context<'a> {
                     format!("{}\nModule.{} = {};\n", class, export_name, export_name)
                 }
                 ExportJs::Function(function) => {
-                    let body = function.strip_prefix("function")
-                                .unwrap();
+                    let body = function.strip_prefix("function").unwrap();
                     if export_name == definition_name {
                         format!("Module.{} = function{}\n", export_name, body)
                     } else {
@@ -662,9 +661,7 @@ __wbg_set_wasm(wasm);"
             }
         };
 
-        let set_to_list = |set: &HashSet<String>| -> Vec<String> {
-            set.iter().cloned().collect()
-        };
+        let set_to_list = |set: &HashSet<String>| -> Vec<String> { set.iter().cloned().collect() };
 
         if matches!(self.config.mode, OutputMode::Emscripten) {
             push_with_newline("var LibraryWbg = {\n");
@@ -672,16 +669,20 @@ __wbg_set_wasm(wasm);"
             push_with_newline(&init_js);
             push_with_newline("$initBindgen__deps: ['$addOnInit'],");
             push_with_newline("$initBindgen__postset: 'addOnInit(initBindgen);',");
-            push_with_newline("$initBindgen: () => {\n
-    wasmExports.__wbindgen_start();");
+            push_with_newline(
+                "$initBindgen: () => {\n
+    wasmExports.__wbindgen_start();",
+            );
             self.globals = self.globals.replace("wasm.", "wasmExports.");
             push_with_newline(&self.globals);
             push_with_newline("},");
-            let deps: Vec<String> = set_to_list(&self.emscripten_deps); 
-            push_with_newline(
-                &format!("}};\n
+            let deps: Vec<String> = set_to_list(&self.emscripten_deps);
+            push_with_newline(&format!(
+                "}};\n
                 extraLibraryFuncs.push('$initBindgen','$addOnInit',{});
-                addToLibrary(LibraryWbg);", deps.join(",")));
+                addToLibrary(LibraryWbg);",
+                deps.join(",")
+            ));
         } else {
             push_with_newline(&imports);
             push_with_newline(&self.imports_post);
@@ -928,16 +929,17 @@ __wbg_set_wasm(wasm);"
                 imports_init.push_str(&import.name);
                 if import.name == "__wbindgen_init_externref_table" {
                     imports_init.push_str(": () =>");
-                    imports_init.push_str(&js
-                        .trim()
-                        .strip_prefix("function()")
-                        .unwrap()
-                        .replace("wasm", "wasmExports"));
+                    imports_init.push_str(
+                        &js.trim()
+                            .strip_prefix("function()")
+                            .unwrap()
+                            .replace("wasm", "wasmExports"),
+                    );
                     imports_init.push_str(",\n");
                 } else {
-                   imports_init.push_str(": ");
-                   imports_init.push_str(&js.trim().replace("wasm", "wasmExports"));
-                   imports_init.push_str(",\n");
+                    imports_init.push_str(": ");
+                    imports_init.push_str(&js.trim().replace("wasm", "wasmExports"));
+                    imports_init.push_str(",\n");
                 }
             }
         }
@@ -984,7 +986,7 @@ __wbg_set_wasm(wasm);"
             }
         }
 
-         let js = match &self.config.mode {
+        let js = match &self.config.mode {
             OutputMode::Emscripten => format!(
             "\
                     {imports_init}",
@@ -1868,7 +1870,7 @@ __wbg_set_wasm(wasm);"
                     "if (typeof {} !== 'undefined') {{ {} }};",
                     s, init
                 )),
-                OutputMode::Emscripten => {},
+                OutputMode::Emscripten => {}
             }
         }
 
@@ -2661,7 +2663,6 @@ __wbg_set_wasm(wasm);"
                 }})`,\n
                 "
             ));
-           
         } else {
             self.global(&format!(
                 "
@@ -3188,16 +3189,19 @@ __wbg_set_wasm(wasm);"
                     )
                 } else {
                     if !import_deps.is_empty() {
-                        for dep in &import_deps{
+                        for dep in &import_deps {
                             self.emscripten_deps.insert(dep.clone());
                         }
-                        format!("function{},\n{}__deps:  [{}]", code, self.module.imports.get(core).name, import_deps.join(","))
-                    }
-                    else {
-                        format!("function{}\n",code)
+                        format!(
+                            "function{},\n{}__deps:  [{}]",
+                            code,
+                            self.module.imports.get(core).name,
+                            import_deps.join(",")
+                        )
+                    } else {
+                        format!("function{}\n", code)
                     }
                 };
-
 
                 self.wasm_import_definitions.insert(core, code);
             }
@@ -3209,7 +3213,8 @@ __wbg_set_wasm(wasm);"
                     self.emscripten_library.push_str("$");
                     self.emscripten_library.push_str(&self.adapter_name(id));
                     self.emscripten_library.push_str(": function");
-                    self.emscripten_library.push_str(&code.replace("wasm.", "wasmExports."));
+                    self.emscripten_library
+                        .push_str(&code.replace("wasm.", "wasmExports."));
                     self.emscripten_library.push_str(",\n\n");
                 } else {
                     self.globals.push_str("function ");
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -18,12 +18,14 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use walrus::{FunctionId, ImportId, MemoryId, Module, TableId, ValType};
 use wasm_bindgen_shared::identifier::is_valid_ident;
+use regex::Regex;
 
 mod binding;
 
 pub struct Context<'a> {
     globals: String,
     emscripten_library: String,
+    emscripten_deps: HashSet<String>,
     imports_post: String,
     typescript: String,
     exposed_globals: Option<HashSet<Cow<'static, str>>>,
@@ -148,6 +150,7 @@ impl<'a> Context<'a> {
         Ok(Context {
             globals: String::new(),
             emscripten_library: String::new(),
+            emscripten_deps: HashSet::new(),
             imports_post: String::new(),
             typescript: "/* tslint:disable */\n/* eslint-disable */\n".to_string(),
             exposed_globals: Some(Default::default()),
@@ -659,19 +662,26 @@ __wbg_set_wasm(wasm);"
             }
         };
 
+        let set_to_list = |set: &HashSet<String>| -> Vec<String> {
+            set.iter().cloned().collect()
+        };
+
         if matches!(self.config.mode, OutputMode::Emscripten) {
             push_with_newline("var LibraryWbg = {\n");
+            push_with_newline(&self.emscripten_library);
             push_with_newline(&init_js);
+            push_with_newline("$initBindgen__deps: ['$addOnInit'],");
             push_with_newline("$initBindgen__postset: 'addOnInit(initBindgen);',");
             push_with_newline("$initBindgen: () => {\n
     wasmExports.__wbindgen_start();");
-            self.globals = self.globals.replace("wasm", "wasmExports");
+            self.globals = self.globals.replace("wasm.", "wasmExports.");
             push_with_newline(&self.globals);
             push_with_newline("},");
+            let deps: Vec<String> = set_to_list(&self.emscripten_deps); 
             push_with_newline(
-                "};\n
-                extraLibraryFuncs.push('$initBindgen');
-                addToLibrary(LibraryWbg);");
+                &format!("}};\n
+                extraLibraryFuncs.push('$initBindgen','$addOnInit',{});
+                addToLibrary(LibraryWbg);", deps.join(",")));
         } else {
             push_with_newline(&imports);
             push_with_newline(&self.imports_post);
@@ -925,12 +935,9 @@ __wbg_set_wasm(wasm);"
                         .replace("wasm", "wasmExports"));
                     imports_init.push_str(",\n");
                 } else {
-                   imports_init.push_str(": (function() {\n");
-                   imports_init.push_str(&self.emscripten_library);
-                   imports_init.push_str("return ");
+                   imports_init.push_str(": ");
                    imports_init.push_str(&js.trim().replace("wasm", "wasmExports"));
-                   imports_init.push_str(";\n");
-                   imports_init.push_str("}()),\n");
+                   imports_init.push_str(",\n");
                 }
             }
         }
@@ -1893,10 +1900,10 @@ __wbg_set_wasm(wasm);"
         if matches!(self.config.mode, OutputMode::Emscripten) {
             self.emscripten_library.push_str(&format!(
                 "
-                function {}(ptr, len) {{
+                ${}(ptr, len) {{
                     ptr = ptr >>> 0;
                     return UTF8Decoder.decode(HEAP8.{}(ptr, ptr + len));
-                }}\n
+                }},\n
                 ",
                 ret, method
             ));
@@ -2501,34 +2508,68 @@ __wbg_set_wasm(wasm);"
         // while we invoke it. If we finish and the closure wasn't
         // destroyed, then we put back the pointer so a future
         // invocation can succeed.
-        self.global(&format!(
-            "
-            function makeMutClosure(arg0, arg1, dtor, f) {{
-                const state = {{ a: arg0, b: arg1, cnt: 1, dtor }};
-                const real = (...args) => {{
-                    // First up with a closure we increment the internal reference
-                    // count. This ensures that the Rust closure environment won't
-                    // be deallocated while we're invoking it.
-                    state.cnt++;
-                    const a = state.a;
-                    state.a = 0;
-                    try {{
-                        return f(a, state.b, ...args);
-                    }} finally {{
-                        if (--state.cnt === 0) {{
-                            wasm.{table}.get(state.dtor)(a, state.b);
-                            CLOSURE_DTORS.unregister(state);
-                        }} else {{
-                            state.a = a;
+
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            self.emscripten_library.push_str(&format!(
+                "
+                $makeMutClosure: function(arg0, arg1, dtor, f) {{
+                    const state = {{ a: arg0, b: arg1, cnt: 1, dtor }};
+                    const real = (...args) => {{
+                        // First up with a closure we increment the internal reference
+                        // count. This ensures that the Rust closure environment won't
+                        // be deallocated while we're invoking it.
+                        state.cnt++;
+                        const a = state.a;
+                        state.a = 0;
+                        try {{
+                            return f(a, state.b, ...args);
+                        }} finally {{
+                            if (--state.cnt === 0) {{
+                                wasmExports.{table}.get(state.dtor)(a, state.b);
+                                CLOSURE_DTORS.unregister(state);
+                            }} else {{
+                                state.a = a;
+                            }}
                         }}
-                    }}
-                }};
-                real.original = state;
-                CLOSURE_DTORS.register(real, state, state);
-                return real;
-            }}
-            ",
-        ));
+                    }};
+                    real.original = state;
+                    CLOSURE_DTORS.register(real, state, state);
+                    return real;
+                }},
+                $makeMutClosure__deps: ['$CLOSURE_DTORS'],\n
+                ",
+            ));
+            self.emscripten_deps.insert("'$CLOSURE_DTORS'".to_string());
+        } else {
+            self.global(&format!(
+                "
+                function makeMutClosure(arg0, arg1, dtor, f) {{
+                    const state = {{ a: arg0, b: arg1, cnt: 1, dtor }};
+                    const real = (...args) => {{
+                        // First up with a closure we increment the internal reference
+                        // count. This ensures that the Rust closure environment won't
+                        // be deallocated while we're invoking it.
+                        state.cnt++;
+                        const a = state.a;
+                        state.a = 0;
+                        try {{
+                            return f(a, state.b, ...args);
+                        }} finally {{
+                            if (--state.cnt === 0) {{
+                                wasm.{table}.get(state.dtor)(a, state.b);
+                                CLOSURE_DTORS.unregister(state);
+                            }} else {{
+                                state.a = a;
+                            }}
+                        }}
+                    }};
+                    real.original = state;
+                    CLOSURE_DTORS.register(real, state, state);
+                    return real;
+                }},\n
+                ",
+            ));
+        }
 
         Ok(())
     }
@@ -2547,32 +2588,61 @@ __wbg_set_wasm(wasm);"
         // executing the destructor, however, we clear out the
         // `this.a` pointer to prevent it being used again the
         // future.
-        self.global(&format!(
-            "
-            function makeClosure(arg0, arg1, dtor, f) {{
-                const state = {{ a: arg0, b: arg1, cnt: 1, dtor }};
-                const real = (...args) => {{
-                    // First up with a closure we increment the internal reference
-                    // count. This ensures that the Rust closure environment won't
-                    // be deallocated while we're invoking it.
-                    state.cnt++;
-                    try {{
-                        return f(state.a, state.b, ...args);
-                    }} finally {{
-                        if (--state.cnt === 0) {{
-                            wasm.{table}.get(state.dtor)(state.a, state.b);
-                            state.a = 0;
-                            CLOSURE_DTORS.unregister(state);
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            self.emscripten_library.push_str(&format!(
+                "
+                $makeClosure: function(arg0, arg1, dtor, f) {{
+                    const state = {{ a: arg0, b: arg1, cnt: 1, dtor }};
+                    const real = (...args) => {{
+                        // First up with a closure we increment the internal reference
+                        // count. This ensures that the Rust closure environment won't
+                        // be deallocated while we're invoking it.
+                        state.cnt++;
+                        try {{
+                            return f(state.a, state.b, ...args);
+                        }} finally {{
+                            if (--state.cnt === 0) {{
+                                wasmExports.{table}.get(state.dtor)(state.a, state.b);
+                                state.a = 0;
+                                CLOSURE_DTORS.unregister(state);
+                            }}
                         }}
-                    }}
-                }};
-                real.original = state;
-                CLOSURE_DTORS.register(real, state, state);
-                return real;
-            }}
-            ",
-        ));
-
+                    }};
+                    real.original = state;
+                    CLOSURE_DTORS.register(real, state, state);
+                    return real;
+                }},
+                $makeClosure__deps: ['$CLOSURE_DTORS'],\n
+                "
+            ));
+            self.emscripten_deps.insert("'$CLOSURE_DTORS'".to_string());
+        } else {
+            self.global(&format!(
+                "
+                function makeClosure(arg0, arg1, dtor, f) {{
+                    const state = {{ a: arg0, b: arg1, cnt: 1, dtor }};
+                    const real = (...args) => {{
+                        // First up with a closure we increment the internal reference
+                        // count. This ensures that the Rust closure environment won't
+                        // be deallocated while we're invoking it.
+                        state.cnt++;
+                        try {{
+                            return f(state.a, state.b, ...args);
+                        }} finally {{
+                            if (--state.cnt === 0) {{
+                                wasm.{table}.get(state.dtor)(state.a, state.b);
+                                state.a = 0;
+                                CLOSURE_DTORS.unregister(state);
+                            }}
+                        }}
+                    }};
+                    real.original = state;
+                    CLOSURE_DTORS.register(real, state, state);
+                    return real;
+                }}
+                ",
+            ));
+        }
         Ok(())
     }
 
@@ -2581,16 +2651,28 @@ __wbg_set_wasm(wasm);"
             return Ok(());
         }
         let table = self.export_function_table()?;
-        self.global(&format!(
-            "
-            const CLOSURE_DTORS = (typeof FinalizationRegistry === 'undefined')
-                ? {{ register: () => {{}}, unregister: () => {{}} }}
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            self.emscripten_library.push_str(&format!(
+                "
+                $CLOSURE_DTORS: `(typeof FinalizationRegistry === 'undefined')
+                    ? {{ register: () => {{}}, unregister: () => {{}} }}
                 : new FinalizationRegistry(state => {{
-                    wasm.{table}.get(state.dtor)(state.a, state.b)
-                }});
-            "
-        ));
-
+                    wasmExports.__indirect_function_table.get(state.dtor)(state.a, state.b)
+                }})`,\n
+                "
+            ));
+           
+        } else {
+            self.global(&format!(
+                "
+                const CLOSURE_DTORS = (typeof FinalizationRegistry === 'undefined')
+                    ? {{ register: () => {{}}, unregister: () => {{}} }}
+                    : new FinalizationRegistry(state => {{
+                        wasm.{table}.get(state.dtor)(state.a, state.b)
+                    }});
+                "
+            ));
+        }
         Ok(())
     }
     fn global(&mut self, s: &str) {
@@ -2956,7 +3038,7 @@ __wbg_set_wasm(wasm);"
             ContextAdapterKind::Export(e) => format!("`{}`", e.debug_name),
             ContextAdapterKind::Adapter => format!("adapter {}", id.0),
         };
-
+        let mut import_deps: Vec<String> = Default::default();
         // Process the `binding` and generate a bunch of JS/TypeScript/etc.
         let binding::JsFunction {
             ts_sig,
@@ -2980,6 +3062,7 @@ __wbg_set_wasm(wasm);"
                 &debug_name,
                 ret_ty_override,
                 ret_desc,
+                &mut import_deps,
             )
             .with_context(|| "failed to generates bindings for ".to_string() + &debug_name)?;
 
@@ -3104,8 +3187,17 @@ __wbg_set_wasm(wasm);"
                         code
                     )
                 } else {
-                    format!("function{}", code)
+                    if !import_deps.is_empty() {
+                        for dep in &import_deps{
+                            self.emscripten_deps.insert(dep.clone());
+                        }
+                        format!("function{},\n{}__deps:  [{}]", code, self.module.imports.get(core).name, import_deps.join(","))
+                    }
+                    else {
+                        format!("function{}\n",code)
+                    }
                 };
+
 
                 self.wasm_import_definitions.insert(core, code);
             }
@@ -3113,10 +3205,18 @@ __wbg_set_wasm(wasm);"
                 assert!(!catch);
                 assert!(!log_error);
 
-                self.globals.push_str("function ");
-                self.globals.push_str(&self.adapter_name(id));
-                self.globals.push_str(&code);
-                self.globals.push_str("\n\n");
+                if matches!(self.config.mode, OutputMode::Emscripten) {
+                    self.emscripten_library.push_str("$");
+                    self.emscripten_library.push_str(&self.adapter_name(id));
+                    self.emscripten_library.push_str(": function");
+                    self.emscripten_library.push_str(&code.replace("wasm.", "wasmExports."));
+                    self.emscripten_library.push_str(",\n\n");
+                } else {
+                    self.globals.push_str("function ");
+                    self.globals.push_str(&self.adapter_name(id));
+                    self.globals.push_str(&code);
+                    self.globals.push_str("\n\n");
+                }
             }
         }
         Ok(())
@@ -3340,6 +3440,7 @@ __wbg_set_wasm(wasm);"
         args: &[String],
         variadic: bool,
         prelude: &mut String,
+        import_deps: &mut Vec<String>,
     ) -> Result<String, Error> {
         let variadic_args = |js_arguments: &[String]| {
             Ok(if !variadic {
@@ -3356,6 +3457,18 @@ __wbg_set_wasm(wasm);"
                 }
             })
         };
+
+        // Emscripten needs the function name extracted from the full function call here
+        // to construct dependency lists for each import.
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            let re = Regex::new(r"([a-zA-Z_$][a-zA-Z0-9_$]*)\s*\(").unwrap();
+            for arg in args {
+                if let Some(result) = re.captures(arg) {
+                    import_deps.push(format!("'${}'", &result[1]));
+                }
+            }
+        }
+
         match import {
             AuxImport::Value(val) => match kind {
                 AdapterJsImportKind::Constructor => {
@@ -3467,10 +3580,12 @@ __wbg_set_wasm(wasm);"
                 assert_eq!(args.len(), 3);
 
                 let call = self.adapter_name(*adapter);
-
+                import_deps.push(format!("'${}'", call));
                 if *mutable {
                     self.expose_make_mut_closure()?;
-
+                    if matches!(self.config.mode, OutputMode::Emscripten) {
+                        import_deps.push("'$makeMutClosure'".to_string());
+                    }
                     Ok(format!(
                         "makeMutClosure({arg0}, {arg1}, {dtor}, {call})",
                         arg0 = &args[0],
@@ -3480,7 +3595,9 @@ __wbg_set_wasm(wasm);"
                     ))
                 } else {
                     self.expose_make_closure()?;
-
+                    if matches!(self.config.mode, OutputMode::Emscripten) {
+                        import_deps.push("'$makeClosure'".to_string());
+                    }
                     Ok(format!(
                         "makeClosure({arg0}, {arg1}, {dtor}, {call})",
                         arg0 = &args[0],
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -209,7 +209,8 @@ impl<'a> Context<'a> {
             OutputMode::Bundler { .. }
             | OutputMode::Node { module: true }
             | OutputMode::Web
-            | OutputMode::Deno => match export {
+            | OutputMode::Deno 
+            | OutputMode::Emscripten => match export {
                 ExportJs::Class(class) => {
                     assert_eq!(export_name, definition_name);
                     format!("export {}\n", class)
@@ -598,7 +599,7 @@ __wbg_set_wasm(wasm);"
             // browsers don't support natively importing Wasm right now so we
             // expose the same initialization function as `--target no-modules`
             // as the default export of the module.
-            OutputMode::Web => {
+            OutputMode::Web | OutputMode::Emscripten => {
                 self.imports_post.push_str("let wasm;\n");
                 init = self.gen_init(needs_manual_start, Some(&mut imports))?;
                 footer.push_str("export { initSync };\n");
@@ -698,7 +699,8 @@ __wbg_set_wasm(wasm);"
             OutputMode::Bundler { .. }
             | OutputMode::Node { module: true }
             | OutputMode::Web
-            | OutputMode::Deno => {
+            | OutputMode::Deno
+            | OutputMode::Emscripten => {
                 for (module, items) in crate::sorted_iter(&self.js_imports) {
                     imports.push_str("import { ");
                     for (i, (item, rename)) in items.iter().enumerate() {
@@ -1771,7 +1773,8 @@ __wbg_set_wasm(wasm);"
             OutputMode::Deno
             | OutputMode::Web
             | OutputMode::NoModules { .. }
-            | OutputMode::Bundler { browser_only: true } => {
+            | OutputMode::Bundler { browser_only: true } 
+            | OutputMode::Emscripten => {
                 self.global(&format!("const cached{0} = (typeof {0} !== 'undefined' ? new {0}{1} : {{ {2}: () => {{ throw Error('{0} not available') }} }} );", s, args, op))
             }
         };
@@ -1785,7 +1788,8 @@ __wbg_set_wasm(wasm);"
                 OutputMode::Deno
                 | OutputMode::Web
                 | OutputMode::NoModules { .. }
-                | OutputMode::Bundler { browser_only: true } => self.global(&format!(
+                | OutputMode::Bundler { browser_only: true }
+                | OutputMode::Emscripten => self.global(&format!(
                     "if (typeof {} !== 'undefined') {{ {} }};",
                     s, init
                 )),
@@ -3522,7 +3526,8 @@ __wbg_set_wasm(wasm);"
                         OutputMode::Web
                         | OutputMode::Bundler { .. }
                         | OutputMode::Deno
-                        | OutputMode::Node { module: true } => "import.meta.url",
+                        | OutputMode::Node { module: true }
+                        | OutputMode::Emscripten => "import.meta.url",
                         OutputMode::Node { module: false } => {
                             "require('url').pathToFileURL(__filename)"
                         }
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -23,6 +23,7 @@ mod binding;
 
 pub struct Context<'a> {
     globals: String,
+    emscripten_library: String,
     imports_post: String,
     typescript: String,
     exposed_globals: Option<HashSet<Cow<'static, str>>>,
@@ -146,6 +147,7 @@ impl<'a> Context<'a> {
     ) -> Result<Context<'a>, Error> {
         Ok(Context {
             globals: String::new(),
+            emscripten_library: String::new(),
             imports_post: String::new(),
             typescript: "/* tslint:disable */\n/* eslint-disable */\n".to_string(),
             exposed_globals: Some(Default::default()),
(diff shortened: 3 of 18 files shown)


guybedford:
@walkingeyerobot this is a really interesting initiative, thanks for your patience. I would love to review it further soon.

crates/macro-support/src/codegen.rs
```diff
@@ -281,12 +281,12 @@ impl ToTokens for ast::Struct {
                     let ptr = #wasm_bindgen::convert::IntoWasmAbi::into_abi(value);
 
                     #[link(wasm_import_module = "__wbindgen_placeholder__")]
-                    #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
+                    #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none", target_os = "emscripten")))]
```
RReverser: I'd say we should change all these checks to just `target_arch = "wasm32"`, no `target_os`.

Excluding Emscripten used to be the primary reason those were added, and we want to experiment with WASI too, so it doesn't make sense to have an allowlist that lists every possible Wasm OS at this point.

.cargo/config.toml
```diff
@@ -1,2 +1,12 @@
 [target.'cfg(target_arch = "wasm32")']
 runner = 'cargo run -p wasm-bindgen-cli --bin wasm-bindgen-test-runner --'
+
+[build]
+
+[target.'cfg(all(target_arch = "wasm32", target_os = "emscripten"))']
+rustflags = [
+  "-Cllvm-args=-enable-emscripten-cxx-exceptions=0",
```
RReverser: It's better to use a high-level setting `-sDISABLE_EXCEPTION_CATCHING` which should do the same.

That said, I wonder if we actually want this in face of recent unwinding support added by @hoodmane?

.cargo/config.toml
```diff
@@ -1,2 +1,12 @@
...
+  "-Clink-arg=-sERROR_ON_UNDEFINED_SYMBOLS=0",
+  "-Clink-arg=-Wno-undefined",
```
RReverser: Why do we need this - which bindings are still missing?

crates/cli-support/src/js/binding.rs
```diff
@@ -985,10 +986,20 @@ fn instruction(
             // Note that we always assume the return pointer is argument 0,
             // which is currently the case for LLVM.
             let val = js.pop();
-            let expr = format!(
+            let mut expr = format!(
                 "{mem}().{method}({} + {size} * {offset}, {val}, true);",
                 js.arg(0),
             );
+            if matches!(js.cx.config.mode, OutputMode::Emscripten) {
+                expr = format!(
```
RReverser: Let's change `expose_dataview_memory` to return `HEAP_DATA_VIEW` instead, that would be a smaller change as the rest of logic remains the same, and it allows us to get rid of a separate dataview binding just for wasm-bindgen.

crates/cli-support/src/js/binding.rs
```diff
@@ -1008,7 +1019,13 @@ fn instruction(
             // If we're loading from the return pointer then we must have pushed
             // it earlier, and we always push the same value, so load that value
             // here
-            let expr = format!("{mem}().{method}(retptr + {size} * {scaled_offset}, true)");
+            let mut expr = format!("{mem}().{method}(retptr + {size} * {scaled_offset}, true)");
+            if matches!(js.cx.config.mode, OutputMode::Emscripten) {
+                expr = format!(
```
RReverser: Same here - if we change `expose_dataview_memory` as suggested above, this diff can get reverted.

crates/cli-support/src/js/mod.rs
```diff
@@ -209,13 +220,50 @@ impl<'a> Context<'a> {
...
+        }
+
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            let actual_js_name: &str = js_name.unwrap_or(&name);
+            self.adapter_deps.insert(format!("\'${actual_js_name}\'"));
+
+            let mut content = val.replace("wasm.", "wasmExports.");
+
+            // Function signature replacement for Emscripten
+            // Replace all occurrences of "function name" -> "$name: function".
```
RReverser: Instead of string match & replacements, which can be pretty unreliable, it would be better to keep original code as-is and append something like

```js
addToLibrary({ func1, func2, ... });
```

for all intrinsics in one go at the end of the JS library script.

Or, equally fine, add it to the library right here after the declaration so you get output like

```js
function func1() {}

addToLibrary({ func1 });

...
```

for each function.

Either way, it should work equivalently - Emscripten preprocessor actually runs the library in Node and uses `func.toString()` for stringification, so it's okay to declare it first and add to the library object after.

crates/cli-support/src/js/mod.rs
```diff
@@ -561,9 +663,63 @@ impl<'a> Context<'a> {
...
+        if !self.emscripten_library.is_empty() {
+            imports.push_str(&self.emscripten_library);
+            imports.push_str(",\n");
+        }
+
+        // Inject Global Dependencies
+        for global_dep in self.emscripten_global_deps.iter() {
+            if global_dep == "'$WASM_VECTOR_LEN'" {
+                imports.push_str("$WASM_VECTOR_LEN: '0',\n");
+            } else if global_dep == "'$textEncoder'" {
```
RReverser: Can we forward these to Emscripten's built-in textencoder & textdecoder instances instead of creating new ones?

crates/cli-support/src/js/mod.rs
```diff
@@ -561,9 +663,63 @@ impl<'a> Context<'a> {
...
+            let import = self.module.imports.get_mut(*id);
+            import.module = "env".to_string();
+            imports.push_str(&import.name);
+            let trimmed_js = js.trim();
+            // Handle specific cases
+            if import.name == "__wbindgen_init_externref_table" {
+                let body = trimmed_js.strip_prefix("function()").unwrap_or(trimmed_js);
+                imports.push_str(&format!(
+                    ": () => {},\n",
+                    body.replace("wasm.", "wasmExports.")
```
RReverser: Similar comments here like above - I'd prefer clear separation between function declarations and adding them to the library object instead of doing somewhat obscure string replacements.

crates/cli-support/src/js/mod.rs
```diff
@@ -209,13 +220,50 @@ impl<'a> Context<'a> {
...
+        js_name: Option<&str>,
+        val: Cow<'static, str>,
+    ) {
+        if self.intrinsics.as_ref().unwrap().contains_key(&name) {
+            return;
+        }
+
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            let actual_js_name: &str = js_name.unwrap_or(&name);
+            self.adapter_deps.insert(format!("\'${actual_js_name}\'"));
```
RReverser: Looks like we manually add single-quotes around every dep when adding to `adapter_deps` collection, perhaps would be cleaner to add keys without those quotes and only add them when we stringify the whole list?

crates/cli-support/src/lib.rs
```diff
@@ -68,6 +68,7 @@ enum OutputMode {
     Node { module: bool },
     Deno,
     Module,
+    Emscripten,
```
RReverser, line 71: In the meetings we talked about Emscripten being just a cfg target instead of separate output mode, so that we could map all the output modes to Emscripten's `ENVIRONMENT`.

Is that still planned as a follow-up?

crates/cli-support/src/js/mod.rs
```diff
@@ -1487,161 +1701,214 @@ impl<'a> Context<'a> {
...
+        };
 
-                    for (; offset < len; offset++) {{
-                        const code = arg.charCodeAt(offset);
-                        if (code > 0x7F) break;
-                        mem[ptr + offset] = code;
-                    }}
-                ",
-            );
+        let encode_as_ascii = format!(
```
RReverser, line 2117: Looks like original comments got lost. Was this done by an AI agent by any chance? They tend to remove comments when refactoring.

crates/cli-support/src/js/mod.rs
```diff
@@ -1487,161 +1701,214 @@ impl<'a> Context<'a> {
...
 
-                    let len = arg.length;
-                    let ptr = malloc(len, 1) >>> 0;
+        {
+        let is_emscripten = matches!(self.config.mode, OutputMode::Emscripten);
 
-                    const mem = {mem}();
+        // Emscripten uses the global 'textEncoder' and a global heap HEAPU8.
+        let text_encoder_var = if is_emscripten { "textEncoder" } else { "cachedTextEncoder" };
+        let mem_formatted = if is_emscripten { format!("{}", mem.name) } else { format!("{mem}()") };
```
RReverser: Like with the `DataView`, it's better to encapsulate this into `expose_uint8_memory` and friends, so that we minimise diffs in usages like these and don't have to maintain two sets of views into the same Wasm memory.

crates/cli-support/src/js/mod.rs
```diff
@@ -1487,161 +1701,214 @@ impl<'a> Context<'a> {
...
+        ret.to_string().into(),
+        None,
 
-                    let len = arg.length;
-                    let ptr = malloc(len, 1) >>> 0;
+        {
+        let is_emscripten = matches!(self.config.mode, OutputMode::Emscripten);
 
-                    const mem = {mem}();
+        // Emscripten uses the global 'textEncoder' and a global heap HEAPU8.
```
RReverser: Can't we just rename the textEncoder to cachedTextEncoder in the Emscripten library? That's coming from our custom JS lib so we can choose any name, right?

crates/cli-support/src/js/mod.rs
```diff
@@ -1687,38 +1954,62 @@ impl<'a> Context<'a> {
...
+                if self.config.mode.emscripten() {
+                    self.adapter_deps.insert("'${add}'".to_string());
+                }
+
+                self.intrinsic(ret.to_string().into(), None, {
+                    let loop_body = if self.config.mode.emscripten() {
+                        format!("HEAP32[ptr + 4 * i >> 2] = {add}(array[i]);")
+                    } else {
+                        format!(
+                            "const add = {add}(array[i]); {mem}().setUint32(ptr + 4 * i, add, true);"
```
RReverser: Similarly here, if we change `expose_uint32_memory` to return the right view, we should be able to have single code here regardless of the target.

crates/cli-support/src/js/mod.rs
```diff
@@ -1750,11 +2041,23 @@ impl<'a> Context<'a> {
     }
 
     fn expose_text_encoder(&mut self, memory: MemoryId) {
-        intrinsic(&mut self.intrinsics, "text_encoder".into(), || {
-            let mut dst =
-                Self::write_text_processor(self.module, memory, "const", "TextEncoder", "()", None);
+        self.emscripten_global_deps
+            .insert("'$textEncoder'".to_string());
```
RReverser: Same here about single-quotes, let's add them only in the final processing.

crates/cli-support/src/js/mod.rs
```diff
@@ -1824,10 +2139,14 @@ impl<'a> Context<'a> {
             // creates just a view. That way in shared mode we copy more data but in
             // non-shared mode there's no need to copy the data except for the
             // string itself.
-            let is_shared = self.module.memories.get(memory).shared;
-            let method = if is_shared { "slice" } else { "subarray" };
-            let text_decoder_decode =
-                format!("cachedTextDecoder.decode({mem}().{method}(ptr, ptr + len))");
+            let text_decoder_decode = if matches!(self.config.mode, OutputMode::Emscripten) {
```
RReverser: I think this will also go away once we change the core memory stringification methods.

crates/cli-support/src/js/mod.rs
```diff
@@ -2132,12 +2455,17 @@ impl<'a> Context<'a> {
             name: name.into(),
             num: view.num,
         };
-        intrinsic(&mut self.intrinsics, name.into(), || {
+        let heap_view = if self.config.mode.emscripten() {
```
RReverser: This too.

crates/cli-support/src/js/mod.rs
```diff
@@ -2525,49 +2898,87 @@ impl<'a> Context<'a> {
...
+                            }}
+                        }};
+                        CLOSURE_DTORS.register(real, state, state);
+                        return real;
+                    }},
+                    $makeMutClosure__deps: ['$CLOSURE_DTORS'],\n
+                    ",
+                ).into()
+            } else {
+                format!(
```
RReverser: Same as with other intrinsics, instead of duplicating code, just add `addToLibrary()` after the declaration just for Emscripten.

.cargo/config.toml
google-yfyang: Done with changing it to `-sDISABLE_EXCEPTION_CATCHING `.

.cargo/config.toml
google-yfyang: Removed. It was needed during the development and seems okay to remove this now.

crates/cli-support/src/js/binding.rs
google-yfyang: Done. expose_dataview_memory actually already returns `HEAP_DATA_VIEW` in Emscripten. To account for a small syntactical difference, just added a conditional to string conversion for Emscripten.

crates/cli-support/src/js/binding.rs
google-yfyang: Done. ditto.

crates/cli-support/src/js/mod.rs
google-yfyang: I definitely agree the current implementation is brittle but the current layout is created to match those found in https://github.com/emscripten-core/emscripten/tree/main/src/lib. I will give this a shot and see how it goes.

crates/cli-support/src/js/mod.rs
google-yfyang: AFAIK, they are not always present in Emscripten and since the cost to create these new instances are pretty low does it  make sense to keep these global instances?

crates/cli-support/src/js/mod.rs
google-yfyang: Should be done now.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/lib.rs
google-yfyang: I think so. Once this change lands we can start looking into that.

crates/cli-support/src/js/mod.rs
google-yfyang: Sorry about this. Not sure how it gets here. Some automated tool was used to help me do the refactoring when syncing to upstream changes. Could also just be that I accidentally deleted them.

crates/cli-support/src/js/mod.rs
google-yfyang: AFAICT, it's not too easy to move this inside these methods directly as fmt for MemView type concatenates view with its name. I've created an `access` method that these methods call to remove duplications.

crates/cli-support/src/js/mod.rs
google-yfyang: Sure that is done.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
RReverser: Yeah I get the intent, just saying it doesn't have to match as Emscripten doesn't parse those modules statically and so doesn't care about the precise file structure.

Let me know how that goes, I spent fair share of time in Emscripten internals and happy to help if you run into issues.

crates/cli-support/src/js/mod.rs
RReverser: If you pass `-s TEXTDECODER=2`, the text decoder would always be present: https://github.com/emscripten-core/emscripten/blob/9c0f34ca00621514e80276d8002b44377f059acd/src/lib/libstrings.js#L16

There is no such setting for the encoder.

But yeah, either way not critical.

You might want to be aware of https://github.com/wasm-bindgen/wasm-bindgen/issues/4471 in this context too - AFAIK Emscripten doesn't have a corresponding fix, so maybe it's another reason to make sure they use a shared encoder instance so that we don't have to track and recreate each one when limit is reached.

crates/wasm-conventions/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
guybedford: This file no longer exists.

crates/threads-xform/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
```
guybedford: This file no longer exists.

crates/cli/src/wasm_bindgen.rs
```diff
@@ -14,6 +14,7 @@ enum Target {
     Deno,
     ExperimentalNodejsModule,
     Module,
+    Emscripten,
```
guybedford: Instead of requiring this to be passed to the Wasm Bindgen CLI, can we not auto-detect when a binary was compiled with Emscripten automatically? It seems to me like this should be fairly trivial even from the producers section?

We already do binary sniffing for various other internal features like the Wasm exception handling instructions, it should be possible to follow similarly here.

crates/cli/src/wasm_bindgen_test_runner.rs
```diff
@@ -378,6 +385,18 @@ fn rmain(cli: Cli) -> anyhow::Result<()> {
             node::execute(module, &tmpdir_path, cli, tests, !no_modules, benchmark)?
         }
         TestMode::Deno => deno::execute(module, &tmpdir_path, cli, tests)?,
+        TestMode::Emscripten => {
```
guybedford, line 388: Again, it would be nice to be able to sniff this from the binary.

tests/wasm32-emscripten/main.rs
```diff
@@ -0,0 +1,57 @@
+#![cfg(all(target_arch = "wasm32", target_os = "emscripten"))]
```
guybedford, line 1: Please can we add this test run to the GitHub CI to ensure we avoid regressions.

crates/cli-support/src/js/mod.rs
```diff
@@ -3253,16 +3740,38 @@ impl<'a> Context<'a> {
...
+                if matches!(self.config.mode, OutputMode::Emscripten)
+                    && !self.adapter_deps.is_empty()
+                {
+                    let mut import_deps_vec = self
+                        .adapter_deps
+                        .iter()
+                        .map(|s| s.to_string())
+                        .collect::<Vec<String>>();
+
+                    import_deps_vec.sort();
```
RReverser: Like for other similar datastructures in the codebase, let's instead use BTreeMap in the first place. That will ensure stable iteration order across any changes, and won't require extra conversion+sorting step here.

crates/cli-support/src/js/mod.rs
```diff
@@ -561,9 +668,63 @@ impl<'a> Context<'a> {
...
+        let mut imports = String::from("var LibraryWbg = {\n");
+
+        // Inject Intrinsics
+        if !self.emscripten_library.is_empty() {
+            imports.push_str(&self.emscripten_library);
+            imports.push_str(",\n");
+        }
+
+        // Inject Global Dependencies
+        for global_dep in self.emscripten_global_deps.iter() {
```
RReverser, line 724: Same here, best to use `BTreeMap` for `emscripten_global_deps` to ensure stable iteration order and reproducible output.

crates/cli-support/src/js/mod.rs
```diff
@@ -254,7 +307,7 @@ impl<'a> Context<'a> {
                     self.typescript.push_str(c);
                 }
                 // in nomodules, we output into a namespace, which is already ambient
-                if !self.config.mode.no_modules() {
+                if !self.config.mode.no_modules() && !self.config.mode.emscripten() {
```
RReverser, line 330: Can you update the comment please? From a quick look it's unclear why we skip `declare` for Emscripten too.

Is it because we don't output ESM there? If so, maybe better to enable that as the default until we have Emscripten support for all wasm-bindgen targets?

Alternatively, should `| Emscripten` be checked in the `OutputMode::no_modules` implementation like you did for the `TestMode::no_modules`? Having inconsistent behaviour in these two methods is confusing.

.cargo/config.toml
```diff
@@ -1,2 +1,11 @@
 [target.'cfg(target_arch = "wasm32")']
 runner = 'cargo run -p wasm-bindgen-cli --bin wasm-bindgen-test-runner --'
+
+[build]
+
+[target.'cfg(all(target_arch = "wasm32", target_os = "emscripten"))']
+rustflags = [
+  "-Clink-arg=-sDISABLE_EXCEPTION_CATCHING=1",
```
hoodmane, line 8: This is a bit weird, are we sure we want this?

hoodmane: Presumably this should be `-Cpanic=abort` if you want to turn off panic unwinding?

crates/threads-xform/Cargo.toml
google-yfyang: Deleted.

crates/wasm-conventions/Cargo.toml
google-yfyang: Deleted.

crates/cli-support/src/js/mod.rs
google-yfyang: Done now.

crates/cli-support/src/lib.rs
google-yfyang: As discussed, used a custom link section to handle the automatic detection.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli/src/wasm_bindgen.rs
google-yfyang: Done.

.cargo/config.toml
egasimus: Looks like the disabling is happening on the Emscripten site to allow regular panic unwinding to do its thing: https://emscripten.org/docs/porting/exceptions.html#emscripten-javascript-based-exception-support

crates/macro-support/src/codegen.rs
```diff
@@ -165,7 +162,7 @@ impl TryToTokens for ast::Program {
         let prefix_json_bytes = syn::LitByteStr::new(&prefix_json_bytes, Span::call_site());
 
         (quote! {
-            #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
+            #[cfg(all(target_arch = "wasm32"))]
```
RReverser: You can remove all these `all()` wrappers since there is just one condition now. `#[cfg(target_arch = "wasm32")]` should be enough.

But since you're changing all these cfgs anyway, I'd suggest to go a step further and replace them with `#[cfg(target_family = "wasm")]` - that will do the same thing as we want here, but will cover future extension to wasm64 as well without having to do another bulk-replace.

crates/cli-support/src/js/mod.rs
```diff
@@ -218,23 +217,65 @@ impl<'a> Context<'a> {
...
+
+        if !content.is_empty() {
+            if content.contains("addToLibrary(") {
+            } else if content.starts_with('$') {
+                content = format!("addToLibrary({{\n  {content}\n}});\n");
+            } else {
+                content = format!("{content}\naddToLibrary({{ '${js_name}': {js_name} }});\n");
+            }
+            
+            self.emscripten_library(&content);
```
RReverser: I was rather thinking of avoiding string manipulation altogether and something much simpler like:


```suggestion
        self.emscripten_library(raw_js);
        self.emscripten_library(&format!("addToLibrary({{ {js_name} }});"));
```

(might require adjustment but the idea is the same - just emit function first, and a statement to add it to the library second, no need to touch the raw JS)

crates/cli/src/wasm_bindgen_test_runner.rs
```diff
@@ -129,10 +129,15 @@ where
...
-    // Collect all tests that the test harness is supposed to run. We assume
-    // that any exported function with the prefix `__wbg_test` is a test we need
-    // to execute.
-    let wasm = fs::read(&cli.file).context("failed to read Wasm file")?;
+    let mut file_name_buf = PathBuf::from(cli.file.clone());
+
+    // Repoint the file to be read from "name.js" to "name.wasm" in the case of emscripten.
+    // Rustc generates a .js and a .wasm file when targeting emscripten. It lists the .js
+    // file as the primary executor which is inconsitent with what is expected here.
+    if file_name_buf.extension().unwrap() == "js" {
```
RReverser: This can panic, better to use


```suggestion
    if file_name_buf.extension().unwrap_or_default() == "js" {
```

crates/cli/src/wasm_bindgen.rs
```diff
@@ -171,7 +171,6 @@ fn rmain(args: &Args) -> Result<(), Error> {
     if let Some(ref name) = args.out_name {
         b.out_name(name);
     }
-
```
RReverser, line 174: Looks like accidental stray change?

crates/cli-support/src/lib.rs
```diff
@@ -232,6 +233,14 @@ impl Bindgen {
         Ok(self)
     }
 
+    /// This is only intended to be called within cli-support.
```
RReverser: If this is true, you can switch it to `pub(crate)` instead.

crates/cli-support/src/lib.rs
RReverser: I think it's a bit different - we talked about "cartesian product" style usage, where module being emscripten is independent from a specific output, so `emscripten` is just another boolean in output settings instead of a separate `OutputMode`.

We can leave this as a future improvement, just saying.

crates/cli-support/src/js/mod.rs
```diff
@@ -218,23 +217,65 @@ impl<'a> Context<'a> {
...
         })
     }
 
     fn has_intrinsic(&self, name: &str) -> bool {
         self.intrinsics.as_ref().unwrap().contains_key(name)
     }
 
+    /// A helper function to add any necessary addToLibrary wrappings for Emscripten
+    fn export_to_emscripten(&mut self, js_name: &str, raw_js: &str) {
+        let mut content = raw_js.replace("wasm.", "wasmExports.").trim().to_string();
```
RReverser: Could we just emit `const wasm = wasmExports;` somewhere when initializing Emscripten output? Then you wouldn't need all these replacements.

.cargo/config.toml
google-yfyang: @hoodmane  I am currently encountering this test failure when running a test in Emscripten mode and it looks like it's possibly related. 

When running `$ WASM_BINDGEN_SPLIT_LINKED_MODULES=1 EMCC_DEBUG_SAVE=1  RUST_BACKTRACE=1 cargo  test --test wasm32-emscripten --target wasm32-unknown-emscripten`
I got 
```
Error: executing `wasm-bindgen` over the Wasm file

Caused by:
    0: failed to generate catch wrappers
    1: __instance_terminated global required for catch wrappers

Stack backtrace:
   0: <anyhow::Error>::msg::<&str>
             at /usr/local/google/home/yfyang/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.100/src/backtrace.rs:27:14
   1: anyhow::__private::format_err
             at /usr/local/google/home/yfyang/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.100/src/lib.rs:695:13
   2: wasm_bindgen_cli_support::transforms::catch_handler::get_terminated_addr::{closure#2}
             at ./crates/cli-support/src/transforms/catch_handler.rs:154:13
   3: <core::option::Option<id_arena::Id<walrus::module::globals::Global>>>::ok_or_else::<anyhow::Error, wasm_bindgen_cli_support::transforms::catch_handler::get_terminated_addr::{closure#2}>
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1369:25
   4: wasm_bindgen_cli_support::transforms::catch_handler::get_terminated_addr
             at ./crates/cli-support/src/transforms/catch_handler.rs:153:10
   5: wasm_bindgen_cli_support::transforms::catch_handler::run
             at ./crates/cli-support/src/transforms/catch_handler.rs:82:27
   6: wasm_bindgen_cli_support::generate_wasm_catch_wrappers
             at ./crates/cli-support/src/lib.rs:832:18
   7: <wasm_bindgen_cli_support::Bindgen>::generate_output
             at ./crates/cli-support/src/lib.rs:472:9
   8: <wasm_bindgen_cli_support::Bindgen>::generate::<&std::path::PathBuf>
             at ./crates/cli-support/src/lib.rs:310:14
   9: wasm_bindgen_cli::wasm_bindgen_test_runner::rmain
             at ./crates/cli/src/wasm_bindgen_test_runner.rs:379:10
  10: wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args::<std::env::ArgsOs, std::ffi::os_str::OsString>
             at ./crates/cli/src/wasm_bindgen_test_runner.rs:128:5
  11: wasm_bindgen_test_runner::main
             at ./crates/cli/src/bin/wasm-bindgen-test-runner.rs:5:5
  12: <fn() -> core::result::Result<(), anyhow::Error> as core::ops::function::FnOnce<()>>::call_once
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
  13: std::sys::backtrace::__rust_begin_short_backtrace::<fn() -> core::result::Result<(), anyhow::Error>, core::result::Result<(), anyhow::Error>>
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs:166:18
  14: std::rt::lang_start::<core::result::Result<(), anyhow::Error>>::{closure#0}
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:206:18
  15: <&dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync as core::ops::function::FnOnce<()>>::call_once
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/core/src/ops/function.rs:287:21
  16: std::panicking::catch_unwind::do_call::<&dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync, i32>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:581:40
  17: std::panicking::catch_unwind::<i32, &dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:544:19
  18: std::panic::catch_unwind::<&dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync, i32>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panic.rs:359:14
  19: std::rt::lang_start_internal::{closure#0}
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/rt.rs:175:24
  20: std::panicking::catch_unwind::do_call::<std::rt::lang_start_internal::{closure#0}, isize>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:581:40
  21: std::panicking::catch_unwind::<isize, std::rt::lang_start_internal::{closure#0}>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:544:19
  22: std::panic::catch_unwind::<std::rt::lang_start_internal::{closure#0}, isize>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panic.rs:359:14
  23: std::rt::lang_start_internal
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/rt.rs:171:5
  24: std::rt::lang_start::<core::result::Result<(), anyhow::Error>>
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:205:5
  25: main
  26: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  27: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:360:3
  28: _start
error: test failed, to rerun pass `--test wasm32-emscripten`
```

And if I tried to replace the `sDISABLE_EXCEPTION_CATCHING ` with `-Cpanic=abort`, I need to add `-Z panic-abort-tests` and run the test with cargo +nightly; but the resulting failure is the same.

I am wondering if you are able to help me with addressing this error

crates/macro-support/src/codegen.rs
google-yfyang: Done.

crates/cli-support/src/lib.rs
google-yfyang: Done.

crates/cli/src/wasm_bindgen.rs
google-yfyang: reverted.

crates/cli/src/wasm_bindgen_test_runner.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
google-yfyang: This is just because Emscripten uses a different format to get typescript to work before. I am pretty sure typescript generation no longer works following the recent syncs to the upstream. Since I plan to address this in a separate change anyways, I will just remove it here to make it cleaner.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/cli-support/src/js/mod.rs
google-yfyang: Done.

crates/macro-support/src/codegen.rs
```diff
@@ -391,12 +391,12 @@ impl ToTokens for ast::Struct {
...
 
                     #[link(wasm_import_module = "__wbindgen_placeholder__")]
-                    #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
+                    #[cfg(target_family = "wasm")]
                     extern "C" {
                         fn #unwrap_fn(ptr: u32) -> u32;
                     }
 
-                    #[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
+                    #[cfg(not(all(target_arch = "wasm32")))]
```
RReverser: Missed `target_family` change here and in a few other places (all `not` expressions probably?).

src/externref.rs
```diff
@@ -106,8 +106,7 @@ fn internal_error(_msg: &str) -> ! {
         } else if #[cfg(feature = "std")] {
             std::process::abort();
         } else if #[cfg(all(
-            target_arch = "wasm32",
-            any(target_os = "unknown", target_os = "none")
+            target_arch = "wasm32"
```
RReverser: Also missed `all` / `target_family` change.

0001-Add-additional-licenses-to-futures-and-web-sys-crate.patch
```diff
@@ -0,0 +1,830 @@
+From f601a40e0b27e9b761af626468df61b742fe584f Mon Sep 17 00:00:00 2001
```
RReverser: Is this file committed accidentally?

google-yfyang:
> Nice, few more fixups mentioned above aside, no more major concerns from me. Thanks for working on this!
> 
> Marking as conditional approval similarly to Guy.

Thanks! I think the only thing that is left blocking is getting the newly added emscripten test to pass as I mentioned in the other thread. 

Specifically, I am running into this try and catch related error that seems related to tha recently introduced change.

When running $ WASM_BINDGEN_SPLIT_LINKED_MODULES=1 EMCC_DEBUG_SAVE=1  RUST_BACKTRACE=1 cargo  test --test wasm32-emscripten --target wasm32-unknown-emscripten
I got
```
Error: executing `wasm-bindgen` over the Wasm file

Caused by:
    0: failed to generate catch wrappers
    1: __instance_terminated global required for catch wrappers

Stack backtrace:
   0: <anyhow::Error>::msg::<&str>
             at /usr/local/google/home/yfyang/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.100/src/backtrace.rs:27:14
   1: anyhow::__private::format_err
             at /usr/local/google/home/yfyang/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.100/src/lib.rs:695:13
   2: wasm_bindgen_cli_support::transforms::catch_handler::get_terminated_addr::{closure#2}
             at ./crates/cli-support/src/transforms/catch_handler.rs:154:13
   3: <core::option::Option<id_arena::Id<walrus::module::globals::Global>>>::ok_or_else::<anyhow::Error, wasm_bindgen_cli_support::transforms::catch_handler::get_terminated_addr::{closure#2}>
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1369:25
   4: wasm_bindgen_cli_support::transforms::catch_handler::get_terminated_addr
             at ./crates/cli-support/src/transforms/catch_handler.rs:153:10
   5: wasm_bindgen_cli_support::transforms::catch_handler::run
             at ./crates/cli-support/src/transforms/catch_handler.rs:82:27
   6: wasm_bindgen_cli_support::generate_wasm_catch_wrappers
             at ./crates/cli-support/src/lib.rs:832:18
   7: <wasm_bindgen_cli_support::Bindgen>::generate_output
             at ./crates/cli-support/src/lib.rs:472:9
   8: <wasm_bindgen_cli_support::Bindgen>::generate::<&std::path::PathBuf>
             at ./crates/cli-support/src/lib.rs:310:14
   9: wasm_bindgen_cli::wasm_bindgen_test_runner::rmain
             at ./crates/cli/src/wasm_bindgen_test_runner.rs:379:10
  10: wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args::<std::env::ArgsOs, std::ffi::os_str::OsString>
             at ./crates/cli/src/wasm_bindgen_test_runner.rs:128:5
  11: wasm_bindgen_test_runner::main
             at ./crates/cli/src/bin/wasm-bindgen-test-runner.rs:5:5
  12: <fn() -> core::result::Result<(), anyhow::Error> as core::ops::function::FnOnce<()>>::call_once
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
  13: std::sys::backtrace::__rust_begin_short_backtrace::<fn() -> core::result::Result<(), anyhow::Error>, core::result::Result<(), anyhow::Error>>
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs:166:18
  14: std::rt::lang_start::<core::result::Result<(), anyhow::Error>>::{closure#0}
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:206:18
  15: <&dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync as core::ops::function::FnOnce<()>>::call_once
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/core/src/ops/function.rs:287:21
  16: std::panicking::catch_unwind::do_call::<&dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync, i32>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:581:40
  17: std::panicking::catch_unwind::<i32, &dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:544:19
  18: std::panic::catch_unwind::<&dyn core::ops::function::Fn<(), Output = i32> + core::panic::unwind_safe::RefUnwindSafe + core::marker::Sync, i32>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panic.rs:359:14
  19: std::rt::lang_start_internal::{closure#0}
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/rt.rs:175:24
  20: std::panicking::catch_unwind::do_call::<std::rt::lang_start_internal::{closure#0}, isize>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:581:40
  21: std::panicking::catch_unwind::<isize, std::rt::lang_start_internal::{closure#0}>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panicking.rs:544:19
  22: std::panic::catch_unwind::<std::rt::lang_start_internal::{closure#0}, isize>
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/panic.rs:359:14
  23: std::rt::lang_start_internal
             at /rustc/eda76d9d1d133effbf7facb28168fd78d75fd434/library/std/src/rt.rs:171:5
  24: std::rt::lang_start::<core::result::Result<(), anyhow::Error>>
             at /usr/local/google/home/yfyang/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs:205:5
  25: main
  26: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
  27: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:360:3
  28: _start
error: test failed, to rerun pass `--test wasm32-emscripten`
```

logan-gatlin:
This is awesome! I managed to get this working on CF workers with minimal issues and a lot of help from Claude. I did have to make a couple changes to get things working though. I am not super familiar with Emscripten, so please close any discussion that isn't actually an issue or is specific to just my problems. 

- My wasm-bindgen emscripten fork: https://github.com/logan-gatlin/wasm-bindgen-emscripten
- My workers-rs fork: https://github.com/cloudflare/workers-rs/tree/emscripten

crates/cli-support/src/js/mod.rs
```diff
@@ -591,9 +697,69 @@ impl<'a> Context<'a> {
...
+        for global_dep in self.emscripten_global_deps.iter() {
+            if global_dep == "WASM_VECTOR_LEN" {
+                imports.push_str("$WASM_VECTOR_LEN: '0',\n");
+            } else if global_dep == "cachedTextEncoder" {
+                imports.push_str("$cachedTextEncoder: \"new TextEncoder()\",\n");
+            } else if global_dep == "cachedTextDecoder" {
+                imports.push_str("$cachedTextDecoder: \"new TextDecoder()\",\n");
+            } else if global_dep == "heap" {
+                imports.push_str(&format!(
+                    "$heap: \"new Array({INITIAL_HEAP_OFFSET}).fill(undefined)\",\n\"heap.push({})\",\n",
```
logan-gatlin: The `"heap.push({})"` here looks like its just a dangling string inside of the object passed to `addToLibrary`. Changing this to `"$heap: \"new Array({INITIAL_HEAP_OFFSET}).fill(undefined)\",\n$heap__postset: \"heap.push({}); heap_next = heap.length;\",\n"` worked for me (AI suggested fix).

crates/cli-support/src/js/mod.rs
```diff
@@ -591,9 +697,69 @@ impl<'a> Context<'a> {
...
+        }
+        // Inject Intrinsics
+        if !self.emscripten_library.is_empty() {
+            imports.push_str(&self.emscripten_library);
+            imports.push('\n');
+        }
+
+        imports.push_str("addToLibrary({\n");
+        // Inject Global Dependencies
+        for global_dep in self.emscripten_global_deps.iter() {
```
logan-gatlin, line 724: Should `$wasm` be declared inside `addToLibrary`? `initBindgen` assigning `wasm = wasmExports` gave a reference error. Adding `imports.push_str("$wasm: \"null\",\n");` before this worked for me (AI suggested fix)

crates/cli-support/src/js/mod.rs
```diff
@@ -2862,8 +3230,20 @@ if (require('worker_threads').isMainThread) {{
...
-            )
-            .into()
+            );
+
+            if matches!(self.config.mode, OutputMode::Emscripten) {
+                js.push_str(
+                    "
+                addToLibrary({
+                    $makeClosure: makeClosure,
+                    $makeClosure_deps: ['$CLOSURE_DTORS']
```
logan-gatlin:
```suggestion
                    $makeClosure__deps: ['$CLOSURE_DTORS']
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2809,20 +3161,36 @@ if (require('worker_threads').isMainThread) {{
...
+            self.emscripten_global_deps
+                .insert("CLOSURE_DTORS".to_string());
+        }
         // For shared closures they can be invoked recursively so we
         // just immediately pass through `this.a`. If we end up
         // executing the destructor, however, we clear out the
         // `this.a` pointer to prevent it being used again the
         // future.
-        intrinsic(&mut self.intrinsics, "make_closure".into(), || {
+        self.intrinsic("make_closure".into(), "makeMutClosure".into(), {
```
logan-gatlin:
```suggestion
        self.intrinsic("make_closure".into(), "makeClosure".into(), {
```

crates/cli-support/src/js/mod.rs
```diff
@@ -1880,14 +2147,23 @@ if (require('worker_threads').isMainThread) {{
...
-                },
-            )
-            .into()
-        });
+            );
+
+
+        if is_emscripten {
+            format!(
+                "{func_decl}
```
logan-gatlin, line 2166: Some symbols [passStringToWam, makeMutClosure, makeClosure] declare their bodies outside of `addToLibrary`, while others declare them inline [CLOSURE_DTORS, PanicError]. I think the second way is correct, according to AI: 

> **Problem:** The function is declared at file scope (`function passStringToWasm0(...) { ... }`) and then registered with `addToLibrary`. But emscripten's `--js-library` files are evaluated at *compile time* by emcc's JS compiler. Free declarations at file scope execute during that phase, when library symbols like `$cachedTextEncoder`, `$WASM_VECTOR_LEN`, `$heap`, etc. don't exist yet. The function body references these symbols, so it either throws or silently captures `undefined`.

crates/cli-support/src/js/mod.rs
```diff
@@ -2441,9 +2753,29 @@ if (require('worker_threads').isMainThread) {{
...
+            let emscripten_heap: &'static str = match kind {
+                "Int8Array" => "HEAP8",
+                "Uint8Array" | "Uint8ClampedArray" => "HEAPU8",
+                "Int16Array" => "HEAP16",
+                "Uint16Array" => "HEAPU16",
+                "Int32Array" => "HEAP32",
+                "Uint32Array" => "HEAPU32",
+                "Float32Array" => "HEAPF32",
+                "Float64Array" => "HEAPF64",
+                "DataView" => "HEAP_DATA_VIEW",
```
logan-gatlin, line 2784: This causes errors for me because `HEAP_DATA_VIEW` is not defined. What worked for me (per ai):

> **Problem:** Emscripten only provides `HEAP_DATA_VIEW` when `SUPPORT_BIG_ENDIAN` is enabled (an uncommon flag). Without it, any wasm-bindgen code that uses `DataView` access (for `setInt32`, `getInt32` on the heap, which is used by virtually all wasm-bindgen glue) references an undefined variable.
> 
> **Fix:**
> 1. When `"DataView"` is mapped, insert `"HEAP_DATA_VIEW"` into `emscripten_global_deps`.
> 2. In the global deps block, emit a `$HEAP_DATA_VIEW` + `$HEAP_DATA_VIEW__postset` that monkey-patches `updateMemoryViews` to create/refresh `HEAP_DATA_VIEW = new DataView(wasmMemory.buffer)` whenever memory views are updated.

crates/cli-support/src/js/mod.rs
```diff
@@ -1199,11 +1369,47 @@ if (require('worker_threads').isMainThread) {{
...
+            ""
+        };
+
+        format!(
+            r#"
+            addToLibrary({{
+                $initBindgen__deps: ['$addOnInit'],
+                $initBindgen__postset: 'addOnInit(initBindgen);',
+                $initBindgen: () => {{
+                    wasm = wasmExports;
```
logan-gatlin, line 1401:
```suggestion
                    wasm = wasmExports;
                    if (wasmExports['_initialize']) {{
                        wasmExports['_initialize']();
                    }}
```
I think `_initialize` has to be called?

crates/cli-support/src/js/mod.rs
```diff
@@ -1199,11 +1369,47 @@ if (require('worker_threads').isMainThread) {{
...
+                    {start_logic}
+                    {classes_and_exports}
+                }}
+            }});
+            
+            extraLibraryFuncs.push('$initBindgen', '$addOnInit', {formatted_deps});
```
logan-gatlin:
```suggestion
            extraLibraryFuncs.push('$initBindgen', '$addOnInit', '$wasm', {formatted_deps});
```
Shouldn't `$wasm` be here as well?

crates/cli-support/src/js/mod.rs
```diff
@@ -1733,143 +1957,186 @@ if (require('worker_threads').isMainThread) {{
...
 
     fn expose_global_heap_next(&mut self) {
         self.expose_global_heap();
-        intrinsic(&mut self.intrinsics, "heap_next".into(), || {
-            "\nlet heap_next = heap.length;\n".into()
-        });
+        self.intrinsic(
+            "heap_next".into(),
+            None,
+            "\nlet heap_next = heap.length;\n".into(),
```
logan-gatlin, line 2000: > **Problem:** On emscripten, the `intrinsic()` path converts this to a file-scope `let heap_next = heap.length;` declaration. But `heap` is a library symbol (`$heap`) that doesn't exist at file scope during emcc's compile-time evaluation. `heap.length` would be evaluated against `undefined`.
> 
> **Fix:** Special-case emscripten to use `export_to_emscripten_with_deps("heap_next", ..., &["heap"])` so `heap_next` is emitted as a library symbol with an explicit `$heap` dependency.

(per AI)

crates/cli-support/src/js/mod.rs
```diff
@@ -228,13 +242,54 @@ impl<'a> Context<'a> {
...
+        js_name: Option<&str>,
+        val: Cow<'static, str>,
+    ) {
+        if self.intrinsics.as_ref().unwrap().contains_key(&name) {
+            return;
+        }
+
+        if matches!(self.config.mode, OutputMode::Emscripten) {
+            let actual_js_name: &str = js_name.unwrap_or(&name);
+            self.adapter_deps.insert(actual_js_name.to_string());
```
logan-gatlin, line 282:
```suggestion
            if !val.trim().is_empty() {
                self.adapter_deps.insert(actual_js_name.to_string());
            }
```
I was getting frivolous warnings about undefined symbols which traced back to this. Per AI:

> **Problem:** `intrinsic()` unconditionally adds the js_name to `adapter_deps`. For `textEncoder` on emscripten, the value passed is `""` (empty) — the actual encoder is handled via `$cachedTextEncoder` in global deps. But the empty-valued name still gets inserted into `adapter_deps`, which causes it to appear in `__deps` arrays as `'$textEncoder'`. Since no `$textEncoder` symbol is ever defined, emcc warns about an undefined symbol.

crates/cli-support/src/lib.rs
google-yfyang: Gotcha. Yeah we can revisit this once the current change gets checked in.

tests/wasm32-emscripten/main.rs
google-yfyang: Done.

src/externref.rs
google-yfyang: Done.

.cargo/config.toml
google-yfyang: This is resolved now with adding a `-Wl,--export=__instance_terminated` to force the symbol into the wasm module.

guybedford:
@google-yfyang some of the changes in https://github.com/logan-gatlin/wasm-bindgen-emscripten/commit/a21e426c5edc1b552ec221e7790dfc4a00659a28 that @logan-gatlin required here seem important. In particular:

* Ensuring heap_next is used in addToLibrary
* memory view updating
* undefined symbol handling
* reset after panic behaviour

We could, for some or any of these cases - land as-is, and have @logan-gatlin create a PR for these with clear tests, or integrate any of the above already now.

Let me know what suits you best here.

.cargo/config.toml
guybedford: I changed this to rather guard on the existence of the terminated symbol, since it seems the emcc files are using exception handling which triggers the detection.

guybedford:
@google-yfyang I was able to run this locally with the removal of the `--target emscripten` from the upstream PR and all is looking great, ready to land here.

I brought through from @logan-gatlin's work some of the wrapper fixes including the multiple case handling for the library calls. I didn't include any of the interpreter error recovery or missing adapter handling, as those sound like cases that correspond to bugs / test cases we should have coverage for to do that. I was also able to replace the regex usage and refactor the terminated binding issue based on detecting the terminated address presence.

crates/cli-support/src/js/mod.rs
```diff
@@ -253,19 +252,69 @@ impl<'a> Context<'a> {
...
+        if content.contains("addToLibrary") {
+            self.emscripten_library(&content);
+            return;
+        }
+
+        // Build the deps array string if we have deps
+        let deps_entry = if deps.is_empty() {
+            String::new()
+        } else {
+            let formatted: Vec<String> = deps.iter().map(|d| format!("'${d}'")).collect();
```
RReverser: These look like almost a revertal to the string manipulation we've been trying to avoid. Is this change necessary? @logan-gatlin @guybedford

crates/cli-support/src/js/mod.rs
```diff
@@ -727,11 +788,25 @@ impl<'a> Context<'a> {
...
+                // Emscripten only provides HEAP_DATA_VIEW with SUPPORT_BIG_ENDIAN.
+                // We monkey-patch updateMemoryViews to create/refresh it whenever
+                // memory views are updated (after wasm instantiation and on memory growth).
+                imports.push_str(concat!(
+                    "$HEAP_DATA_VIEW: \"null\",\n",
+                    "$HEAP_DATA_VIEW__postset: \"",
+                    "var _origUpdateMemoryViews = updateMemoryViews; ",
+                    "updateMemoryViews = function() { ",
+                    "_origUpdateMemoryViews(); ",
+                    "HEAP_DATA_VIEW = new DataView(wasmMemory.buffer); ",
```
RReverser: This looks odd too.

If Emscripten doesn't update its own DataView and this is a temporary workaround, we should at least make an upstream PR and clear TODO to remove this hack in the future.

crates/cli-support/src/js/mod.rs
```diff
@@ -2154,11 +2250,13 @@ if (require('worker_threads').isMainThread) {{
 
 
         if is_emscripten {
+            // Inline the function as a value inside addToLibrary instead of
```
RReverser: This is also a revertal of what @google-yfyang has been working on.

Execution at compile time is how Emscripten libraries work, we intentionally made leverage of it and function declarations to avoid brittle string manipulations.

crates/cli-support/src/js/mod.rs
google-yfyang: `HEAP_DATA_VIEW` will be actually available when SUPPORT_BIG_ENDIAN or WASM_BINDGEN is set. The relevant change to the Emscripten upstream will land once we can merge this PR. As the result, I don't think we will need this change.

google-yfyang:
> @google-yfyang some of the changes in [logan-gatlin/wasm-bindgen-emscripten@a21e426](https://github.com/logan-gatlin/wasm-bindgen-emscripten/commit/a21e426c5edc1b552ec221e7790dfc4a00659a28) that @logan-gatlin required here seem important. In particular:
> 
> * Ensuring heap_next is used in addToLibrary
> * memory view updating
> * undefined symbol handling
> * reset after panic behaviour
> 
> We could, for some or any of these cases - land as-is, and have @logan-gatlin create a PR for these with clear tests, or integrate any of the above already now.
> 
> Let me know what suits you best here.

I misunderstood what @logan-gatlin 's comment was there for. Sorry for not taking a look earlier. While I appreciate making these improvements to the Emscripten support, since this PR is already bloated as is, it will be cleaner and easier to review if we can pack further improvements as follow-up changes. 

I would also like to understand what initial errors were encountered that prompted relevant changes.

google-yfyang:
> @google-yfyang I was able to run this locally with the removal of the `--target emscripten` from the upstream PR and all is looking great, ready to land here.
> 
> I brought through from @logan-gatlin's work some of the wrapper fixes including the multiple case handling for the library calls. I didn't include any of the interpreter error recovery or missing adapter handling, as those sound like cases that correspond to bugs / test cases we should have coverage for to do that. I was also able to replace the regex usage and refactor the terminated binding issue based on detecting the terminated address presence.

Thanks! The removal of regex usage and addressing terminated binding issue look great to me.

logan-gatlin:
> I would also like to understand what initial errors were encountered that prompted relevant changes.

Getting this ready for a follow up PR. Didn't mean to delay this PR any further, sorry

guybedford:
Ok I've removed the patches pointed out. This now only leaves the following, which I've verified via end-to-end testing:

- expose_make_closure with wrong js_name ("makeMutClosure" instead of "makeClosure")
- Fixed `self.adapter_deps.insert("{add}".to_string());` to properly be an interpolated variable
- Double underscore `$makeClosure_deps` -> `$makeClosure__deps`
- Removed regex-based dep extraction, since all adapter cases are covered
- Guarded expose_text_encoder as requested by @RReverser via emscripten_global_deps insert behind Emscripten mode check
- Added `$wasm: "null"` declaration to generate_emscripten_imports
- Fixed `$heap__postset` (was broken inline push)
- Added _initialize() call in initBindgen for --no-entry builds (bug hit by @logan-gatlin in https://github.com/cloudflare/workers-rs/pull/951)
- Added $wasm to extraLibraryFuncs
- Skip catch wrapper generation when __instance_terminated is absent due to emscripten exception handling builds triggering incorrectly
- Reverted license changes from #4457

@RReverser PTAL

crates/cli-support/src/js/mod.rs
guybedford: Removed

crates/cli-support/src/js/mod.rs
guybedford: Removed

crates/cli-support/src/js/mod.rs
guybedford: Removed

crates/cli-support/src/js/mod.rs
google-yfyang: As I mentioned in the other thread, the change to provide HEAP_DATA_VIEW when the build detects wasm_bindgen is pending and will be landed soon after this PR is merged.

PR #4455, opened 2025-03-18
Typescript Generic Types

rouzwelt:
### Motivation

Currently if a function/method has a generic return or arguments type like:
```rust
pub fn some_fn(arg: SomeType<String>) -> SomeType<u8> {}
// or
pub fn some_fn(arg: SomeType<String>) -> Result<SomeType<u8>, JsValue> {}
```
where `SomeType` is just plain struct that impls wasm traits manually (not using wasm_bindgen macro on its definition, ie not js class) the generated typescript bindings wont correctly map the types and doesnt actually include the inner types:
```ts
// for both cases above
export function some_fn(arg: SomeType): SomeType
```
Although user can use `unchecked_return_type` and `unchecked_param_type` attribute to specify them, but this is not really ideal, however if the nested types can be processed regularly like the main type in recursive fashion, this can simply be resolved, so on typescript we would get:
```ts
// for both cases above
export function some_fn(arg: SomeType<string>): SomeType<number>
```
### Proposed Solution
Implement some logic to parse and map the types deeply, possibly recursively, until all nested types are parsed and accounted for in generated typescript.
There is a mapper currently that maps the rust return types to their equivalent js/ts types, but looks like it just checks 1 level deep, and doesnt go deeper than that, however with some tweaks, the same logic can be applied on nested types to correctly map the return type for the js/ts bindings:

## High-Level Implementation Details

We we would use native encoded/decoded `WasmDescribe::inform()` for return and arguments types, meaning we will encode their wasm informs in the `backend::codegen` for `Export` and then decode it at `cli-support::descriptor::Function` which now has 2 new more fields:
`inner_ret_map` and `args_ty_map`, these 2 field hold the `Descriptor` which now has 1 more variant called `Descriptor::Generic`:
```rust
Descriptor::Generic {
        ty: Box<Descriptor>,
        args: Box<[Descriptor]>,
}
```
 that can hold the type map for both args and return type, now at `cli-support::wit::register_export_adapter` we convert those `inner_ret_map` and `args_ty_map` fields into `AdapterType` using a newly added function called `describe_map_to_adapter` where it just maps the given `Descriptor` to its `AdapterType` by also handling of nested generic types.
Lastly when building the ts signature and js doc for the binding function/method, we will convert those generic `AdapterType`s into their corresponding js/ts types and append them to the main parent type.

### DIFF
--- a/crates/backend/src/codegen.rs
+++ b/crates/backend/src/codegen.rs
@@ -9,8 +9,8 @@ use quote::{quote, ToTokens};
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
 use syn::parse_quote;
-use syn::spanned::Spanned;
-use wasm_bindgen_shared as shared;
+use syn::{spanned::Spanned, PathArguments, PathSegment, Type};
+use wasm_bindgen_shared::{self as shared, identifier::TYPE_DESCRIBE_MAP};
 
 /// A trait for converting AST structs into Tokens and adding them to a TokenStream,
 /// or providing a diagnostic if conversion fails.
@@ -863,21 +863,29 @@ impl TryToTokens for ast::Export {
         })
         .to_tokens(into);
 
+        let mut args_ty_map = Vec::new();
         let describe_args: TokenStream = argtys
             .iter()
-            .map(|ty| match ty {
-                syn::Type::Reference(reference)
-                    if self.function.r#async && reference.mutability.is_none() =>
-                {
-                    let inner = &reference.elem;
-                    quote! {
-                        inform(LONGREF);
-                        <#inner as WasmDescribe>::describe();
+            .map(|ty| {
+                args_ty_map.push(map_type_describe(ty));
+                match ty {
+                    syn::Type::Reference(reference)
+                        if self.function.r#async && reference.mutability.is_none() =>
+                    {
+                        let inner = &reference.elem;
+                        quote! {
+                            inform(LONGREF);
+                            <#inner as WasmDescribe>::describe();
+                        }
                     }
+                    _ => quote! { <#ty as WasmDescribe>::describe(); },
                 }
-                _ => quote! { <#ty as WasmDescribe>::describe(); },
             })
             .collect();
+
+        // build the return type map and args type map
+        let ret_ty_map = map_type_describe(syn_ret);
+        let args_ty_map: TokenStream = args_ty_map.into_iter().collect();
 
         // In addition to generating the shim function above which is what
         // our generated JS will invoke, we *also* generate a "descriptor"
@@ -904,6 +912,10 @@ impl TryToTokens for ast::Export {
                 inform(#nargs);
                 #describe_args
                 #describe_ret
+                inform(#TYPE_DESCRIBE_MAP);
+                #ret_ty_map
+                inform(#TYPE_DESCRIBE_MAP);
+                #args_ty_map
             },
             attrs: attrs.clone(),
             wasm_bindgen: &self.wasm_bindgen,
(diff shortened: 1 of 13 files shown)


rouzwelt:
any @maintainers here?

rouzwelt:
@daxpedda hey there, sorry to ping you, this PR has been waiting on a review for some time now, would be nice if you or any other maintainers can take a look at it.
Thanks in advance

rouzwelt:
@alexcrichton @fitzgen hey guys, would be nice to get a review on this whenever you guys got some time.

rouzwelt:
@alexcrichton @fitzgen @daxpedda hey guys, would be nice if someone can take a look at this PR and maybe we can settle this, I think it would be a really good improvement that many would need/benefit from

rouzwelt:
@alexcrichton @fitzgen @daxpedda hey guys, it would be really nice if we can work this PR out together, looking forward to your response

guybedford:
@rouzwelt new maintainer here, this is an interesting direction, I would like to review it more carefully soon. It would be great if you could be clear about the limitations of the approach as well.

rouzwelt:
> @rouzwelt new maintainer here, this is an interesting direction, I would like to review it more carefully soon. It would be great if you could be clear about the limitations of the approach as well.

hey there, great to have you on
regarding limitations, I haven't really noticed any meaningful limitations so far, that's because for a type to be able to be compatible with wasm boundary, it should have full implementations of the traits, even generic ones, the inner generic types they themselves should have all the wasm traits implemented even if they don't get parsed/show up ins the final generated dts as generics, because for rust side of things to get compiled, those traits should be implemented for all of the inner types of a generic type,
so this PR just makes the generics to become available in the final generated dts

rouzwelt:
hopefully that explains my case here

rouzwelt:
@guybedford hopefully that explains my case here

rouzwelt:
@guybedford looking forward to your reply

daxpedda:
Please stop pinging maintainers repeatedly. Your PR hasn't been forgotten, we just don't have the necessary bandwidth.

FWIW we will be discussing generic support next week and try to come up with a design we want to support. Your PR will definitely help in that discussion so it is much appreciated!

crates/typescript-tests/src/ts_generic_sig.rs
```diff
@@ -0,0 +1,291 @@
...
+pub struct OtherGenericType<T, E> {
+    pub field1: T,
+    pub field2: E,
+}
+
+// some error type
+pub struct Error;
+
+#[wasm_bindgen]
+pub struct TestStruct;
```
guybedford, line 30: All types here seem to be _exported_ types. Can we extend the coverage here to imported types as well?

We would also need to cover interactions with `js_name` and `js_namespace` renaming to ensure aliasing is supported.

crates/backend/src/codegen.rs
```diff
@@ -904,6 +912,10 @@ impl TryToTokens for ast::Export {
                 inform(#nargs);
                 #describe_args
                 #describe_ret
+                inform(#TYPE_DESCRIBE_MAP);
+                #ret_ty_map
+                inform(#TYPE_DESCRIBE_MAP);
+                #args_ty_map
```
guybedford, line 918: I don't follow why we need two separate serializations here. The first-level cut-off of the generics is the function type, so we should just fold the data into describe_args and describe_ret instead of duplicating here.

crates/cli-support/src/wit/standard.rs
```diff
@@ -95,6 +97,11 @@ pub enum AdapterType {
     NamedExternref(String),
     Function,
     NonNull,
+    Unit,
+    Generic {
+        ty: Box<AdapterType>,
+        args: Box<[AdapterType]>,
+    },
```
guybedford, line 104: If we were writing this from scratch today, surely we wouldn't have a separate `Generic`, and would instead deeply nest all types properly? This seems like the right approach to properly integrate with other generics work in progress as well (such as generic enums - https://github.com/wasm-bindgen/wasm-bindgen/pull/4734, which had to deal with similar problems here in its own way).

I think we should try and generalize the approach a little better instead of stapling it on here.

So:
1. If using a separate generic type, generalizing it to not be function-specific?
2. Or, ideally, just making functions encode their deep tree to begin with?

rouzwelt:
@guybedford thanks for the review, well I submitted this a few months ago, and I believe there has been some changes and some generic features added since then as you mentioned, so I guess I first need to update myself with the new changes as I don't really remember the details of this PR 😄 
I read your review as you asked some questions, tbh I don't really remember why I made some of those decisions at the time of preparing this PR, so I need to review it myself again.

so I'll try to free up some time and resolve the conflicts first and then address your review if that's ok with you, plz let me know

guybedford:
@rouzwelt none of the generics changes have landed yet, but it would be nice to know work done here can align well with https://github.com/wasm-bindgen/wasm-bindgen/pull/4734 and https://github.com/wasm-bindgen/wasm-bindgen/pull/4746 for if/when they land.

Thanks for taking another look - just let me know when you're ready to discuss again and I'm happy to do another review.

Issue #4471, opened 2025-04-06
Safari: TextDecoder refuses to process data after 2GB causing RangeError (Bad value) and Out of bounds memory access


bes:
### Describe the Bug

After processing 2GB of data, Safari's TextDecoder refuses to produce any more strings, apparently some sort of (successful) security mitigation that had unintended consequences.

We are seeing lots (thousands per week) of crashes in Sentry from Safari versions 16 up to the latest 18.4, with crashes saying `Out of bounds memory access` and `RangeError(BadValue)`.

There is a [bug report](https://bugs.webkit.org/show_bug.cgi?id=280593) on the WebKit bugzilla, but it is unlikely to get fixed for older Safari versions (and they don't seem to be in a hurry to fix it, was reported 2024-09).

I initially started another [bug report](https://bugs.webkit.org/show_bug.cgi?id=286266) on WebKit bugzilla, but am now reasonably convinced that the `TextDecoder` is the culprit for our issues too.

### Steps to Reproduce

1. Open any Safari version 16 - 18.4
2. Open the Safari console on any site
3. Run the provided script from the [bug report](https://bugs.webkit.org/show_bug.cgi?id=280593)
```javascript
const text = new TextDecoder()
const buff = new ArrayBuffer(100)
for (let i = 0; i < 21474837; i++) {
    text.decode(buff)
}
```
4. See that RangeError: Bad value is produced
![Image](https://github.com/user-attachments/assets/8dea9e09-cdb5-4e01-afc8-a0a213e52be1)

When this happens in wasm-bindgen generated program (and it does happen for us, all the time) it's game over for that browser/tab instance.

### Expected Behavior

I expect Safari to fix this at the source, but I think wasm-bindgen could provide a workaround for Safari specifically, creating a new `TextDecoder` when decoding is about to pass 2GB of data.

### Actual Behavior

After processing 2GB of text, Safari's `TextDecoder` won't be able to decode anymore.

### Additional Context

This isn't an optimal situation for anyone, but I hope that we can reach consensus on what needs to be done here.

wasm-bindgen generates the following code for me:

```javascript
const lTextEncoder = typeof TextEncoder === 'undefined' ? (0, module.require)('util').TextEncoder : TextEncoder;

let cachedTextEncoder = new lTextEncoder('utf-8');

function getStringFromWasm0(ptr, len) {
    ptr = ptr >>> 0;
[...]

PR #4472, opened 2025-04-06
Workaround for a TextDecoder bug in Safari causing a RangeError to be thrown

bes:
`TextDecoder` in Safari has a limitation that causes it to throw `RangeError` after decoding more than 2GiB of data. This causes long running wasm programs that need to use `TextDecoder` to crash and start throwing `RuntimeError` with the message "Out of bounds memory access".

We work around the issue by tracking how much data has been decoded by any given `TextDecoder`, and replace it when it comes close to 2GiB, deducting a small margin of 1MiB which has been empirically shown to reduce the likelihood of miscounting (for unknown reasons) causing a `RangeError` to be thrown.

This commit also adds stricter handling of the kind of declaration used for TextDecoder and TextEncoder - TextDecoder always uses let because it needs to be mutable, and TextEncoder always uses const because it doesn't need to be mutable.

Fixes https://github.com/rustwasm/wasm-bindgen/issues/4471

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1689,10 +1689,10 @@ __wbg_set_wasm(wasm);"
         if !self.should_write_global("text_encoder") {
             return Ok(());
         }
-        self.expose_text_processor("TextEncoder", "encode", "('utf-8')", None)
+        self.expose_text_processor("const", "TextEncoder", "encode", "('utf-8')", None)
     }
 
-    fn expose_text_decoder(&mut self) -> Result<(), Error> {
+    fn expose_text_decoder(&mut self, mem: &MemView, memory: MemoryId) -> Result<(), Error> {
         if !self.should_write_global("text_decoder") {
             return Ok(());
         }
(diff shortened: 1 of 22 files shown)


crates/cli/tests/reference/getter-setter.wat
```diff
@@ -11,16 +11,16 @@
   (type (;9;) (func (param i32 f64)))
   (import "./reference_test_bg.js" "__wbindgen_init_externref_table" (func (;0;) (type 0)))
   (func $__wbindgen_realloc (;1;) (type 8) (param i32 i32 i32 i32) (result i32))
-  (func $__wbindgen_malloc (;2;) (type 6) (param i32 i32) (result i32))
```
bes, line 14: I am not sure why I needed these changes. I am using rustc 1.86.0, maybe I should be using another version?

bes:
Some unrelated clippy lints are failing, has CI rustc been updated without updating the code?

bes:
There don't seem to be any tests (generated .js-files) that use the non-browser path of `getText` 🤔

stephanemagnenat:
> Some unrelated clippy lints are failing, has CI rustc been updated without updating the code?

@daxpedda do you have any input on these? The PR works around a critical bug on Safari and would be very useful for many people!

stephanemagnenat:
@bes I opened an issue (#4482) as even `main` does not compile any more in CI.

bes:
> @bes I opened an issue (#4482) as even `main` does not compile any more in CI.
Thanks

bes:
We've been running this fix in production for a week now, and the results are encouraging. Screenshots from Sentry.

Ignore the first column, it's low because of a quirk in how Sentry does data cutoff...

range_error

safari_by_version


Can you guess which day the fix went out?

stephanemagnenat:
@bes the CI has been fixed in latest main commit, could you rebase your PR please? That should help it getting merged, now that there are more active maintainers (see #4533) ;-)

crates/cli-support/src/js/mod.rs
```diff
@@ -1729,32 +1729,82 @@ __wbg_set_wasm(wasm);"
...
+                    "TextDecoder",
+                    "decode",
+                    "('utf-8', { ignoreBOM: true, fatal: true })",
+                )?;
+
+                // Maximum number of bytes Safari can handle for one TextDecoder is 2GiB (2147483648)
+                // but empirically it seems to crash a bit before the end, so we remove 1MiB of margin.
+                // Workaround for a bug in Safari.
+                // See https://github.com/rustwasm/wasm-bindgen/issues/4471
+                const MAX_SAFARI_DECODE_BYTES: u32 = 2147483648 - 1048576;
```
daxpedda: Lets convert those to hex.

bes: Converted to hex

bes:
> I'm a tiny bit concerned about the 1MiB margin. This [comment](https://bugs.webkit.org/show_bug.cgi?id=280593#c2) shows that it fails at 2,147,483,600 bytes and succeeds at 2,147,483,500 bytes. So the margin should be between 148 and 47 bytes.
> 
> Considering how reproducable this test looks like, it would be nice to find out the exact number it fails at. The precise number is actually not that important, it would just be nice to once and for all establish that it actually fails at a certain number and that its not flaky or anything.
> 
> In any case, until somebody finds the time to actually try this out we will leave it at the current 1MiB margin considering that there is real-world data behind it.

I tried a small margin at first but could not get it to work on my machine, so I grabbed a margin which I thought was large enough to avoid problems but small enough to not have any significance (1 MiB is 0.049% of 2GiB)

crates/cli-support/src/js/mod.rs
stephanemagnenat: Tests still have the decimal version.

bes: That's generated JS, does it matter?

stephanemagnenat: I guess not ;-)

daxpedda:
> I tried a small margin at first but could not get it to work on my machine, so I grabbed a margin which I thought was large enough to avoid problems but small enough to not have any significance (1 MiB is 0.049% of 2GiB)

If you can actually try it, could you do the following:
```js
const decoder = new TextDecoder()
const buffer1 = new ArrayBuffer(100)

for (let i = 0; i < 21474836; i++) {
    decoder.decode(buffer1)
}

const buffer2 = new ArrayBuffer(1)
let i = 2147483601

try {
    for (; i <= 2147483648; i++) {
        decoder.decode(buffer2)
    }
} catch {
    console.err(`failed at ${i} number of bytes`)
}
```
And check if it fails consistently at the same number of bytes.

daxpedda (changes requested):
I keep forgetting this: please add an entry to the changelog as well!

bes:
Added a line in the changelog.

> If you can actually try it, could you do the following:

I'm on vacation right now, supposed to go on a day trip, so pushing these change kinda in between packing. I would appreciate if we can skip this for now :)

stephanemagnenat:
@bes thank you so much for your work!

@daxpedda thank you for merging! Any plan for making a release soon? This fix would be very useful to us!

daxpedda:
I not yet back to maintaining `wasm-bindgen` so my time is quite limited right now, I'll leave this to the other maintainers to decide.

My 2¢: making a release is relatively low cost so I favor making one as soon as requested.

stephanemagnenat:
> My 2¢: making a release is relatively low cost so I favor making one as soon as requested.

Is there an ideal way to make such a request? Should I open a new issue for that, or do other maintainers read closed PR?

daxpedda:
You asking for it here was the request.
But please open a tracking issue so it doesn't get lost!

PR #4488, opened 2025-04-22
Comment out test exports in generated .d.ts

### LINKED ISSUE #4449: wasm_bindgen_test syntax error at :: in .d.ts ("Expected type")
When I define a test like

```rs
#[wasm_bindgen_test]
async fn screen_example() {
    // ...
}
```

in my wasm-bindgen module `screening_wasm/src/lib.rs`, the resulting `screening_wasm/pkg/screening_wasm.d.ts` contains

```ts
// ...

export interface InitOutput {
  readonly __wbg_screenconfig_free: (a: number, b: number) => void;
  readonly __wbg_enumerationsettings_free: (a: number, b: number) => void;
  readonly __wbg_fixeddomains_free: (a: number, b: number) => void;
  readonly start: () => void;
  readonly screen: (a: any, b: any) => any;
  readonly __wbgt__screening_wasm::screen_example: (a: number) => void;

//                               ^ Expected a type but instead found ':'.
// ...
```

I'm not sure how this is supposed to work, but I guess the cause of this behavior is that wasm-bindgen-test-macro [generates](https://github.com/rustwasm/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/crates/test-macro/src/lib.rs#L100)

```rs
#[export_name = ::core::concat!("__wbgt_", #ignore_name, "_", ::core::module_path!(), "::", ::core::stringify!(#ident))]
```

and then [wasm2es6js.rs](https://github.com/rustwasm/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/crates/cli-support/src/wasm2es6js.rs) uses that name when making the .d.ts file, but it contains `::` which is not valid in a TypeScript identifier.

lynn:
Fixes #4449.

The change to crates/cli/tests/reference/wasm-export-types.rs reproduces the issue in a test. The change to wasm2es6js.rs fixes it.


crates/cli/tests/reference/wasm-export-types.d.ts
```diff
@@ -2,17 +2,99 @@
...
 
 export type InitInput = RequestInfo | URL | Response | BufferSource | WebAssembly.Module;
 
 export interface InitOutput {
   readonly memory: WebAssembly.Memory;
   readonly example: (a: number, b: bigint, c: any, d: number, e: number) => [number, number];
   readonly example_128: (a: bigint, b: bigint) => [number, bigint, bigint];
-  readonly __wbindgen_export_0: WebAssembly.Table;
+  readonly "__wbgt__reference_test::example_test": (a: number) => void;
+  readonly __wbgtest_cov_dump: () => [number, number];
```
lynn, line 76: The fix is visible here :) `__wbgt__reference_test::example_test` is quoted now. 🎉

guybedford:
This still seems to be failing CI unfortunately.

guybedford:
Finally just looks like the fixtures still just need to be rerun with a bless flag to update the expected output.

lynn:
> looks like the fixtures still just need to be rerun with a bless flag to update the expected output.

Sorry, I'm not totally sure what that means or (if it is something I'm supposed to do) how to achieve it. Is there a command I can try running?

guybedford:
Yes, you can run:

```
BLESS=1 cargo test -p wasm-bindgen-cli --test reference
```

crates/cli-support/src/wasm2es6js.rs
```diff
@@ -63,7 +63,11 @@ fn args_are_optional(name: &str) -> bool {
 pub fn interface(module: &Module) -> Result<String, Error> {
     let mut exports = String::new();
     module_export_types(module, |name, ty| {
-        writeln!(exports, "  readonly {name}: {ty};").unwrap();
+        if name.contains(':') {
+            writeln!(exports, "  readonly {name:?}: {ty};").unwrap();
```
daxpedda, line 69: Its not entirely clear on first glance whats going on here. I think a comment would help.

crates/cli/tests/reference/wasm-export-types.rs
```diff
@@ -15,3 +17,8 @@ pub fn example(a: u32, b: u64, c: JsValue, d: &str) -> String {
 pub fn example_128(a: u128) -> Option<u128> {
     None
 }
+
+#[wasm_bindgen_test]
+fn example_test() {
+    assert_eq!(example_128(1), None)
+}
```
daxpedda: I assume this is leftover from when you were debugging?

daxpedda (changes requested):
AFAICT we still need a test for this. Otherwise this is LGTM.

(if you can't get the reference tests to align dw about it, we can fix that before merging)

crates/cli/tests/reference/wasm-export-colon.d.ts
```diff
@@ -0,0 +1,118 @@
...
+   * `false` if at least one test failed.
+   */
+  run(tests: any[]): Promise<any>;
+}
+
+export type InitInput = RequestInfo | URL | Response | BufferSource | WebAssembly.Module;
+
+export interface InitOutput {
+  readonly memory: WebAssembly.Memory;
+  readonly "__wbgt__wasm_export_colon_reftest::colon_test": (a: number) => void;
```
lynn, line 72: Is this sufficient test of the feature? Previously, this `.d.ts` file would have caused a syntax error because there would be no quotes on this line.

guybedford: This looks great to me.

guybedford (approved):

crates/cli/tests/reference/wasm-export-colon.d.ts
lynn: Excellent! Could you take another look, @daxpedda?

Spxg:
changelog needs an update, it's currently in the release log for version 0.2.104.

PR #4503, opened 2025-05-05
Replace atomic operations with Once in backend/util

### LINKED ISSUE #4502: Race Condition and Performance Regression in wasm-bindgen util.rs
### Describe the Bug

There is a race condition and a related performance issue in the following section of the wasm-bindgen codebase:
https://github.com/rustwasm/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/crates/backend/src/util.rs#L135-L161

1. Race Condition:
The affected code attempts to lazily compute and cache a hash value, but the current implementation is not thread-safe. Consider the following interleaving of two threads:

--------------------------
|   Thread 1   | Thread 2 |
| -------------| -----------|
| check Hashed | check Hashed |
| recompute Hash | recompute Hash |

If two threads call the fmt function concurrently, they may both observe the Hashed flag as false and proceed to recompute the hash independently. This violates the intent of computing the hash only once and leads to redundant work. While the result is ultimately consistent (assuming a stable environment), this introduces unnecessary overhead.

2. Memory Ordering Issue
The code currently uses Ordering::SeqCst, which is overly strict for this scenario. A combination of Ordering::Acquire and Ordering::Release is sufficient to ensure correct synchronization while allowing better performance on most architectures. But for now OnceLock is the most concise implementation.

### Steps to Reproduce

Found from static analysis

### Expected Behavior

No race condition. Performance increase.

### Actual Behavior

In a benchmark on x86 Linux, replacing SeqCst atomics with OnceLock not only eliminated potential race conditions by ensuring the hash is computed only once, but also yielded a slight performance improvement.

### Additional Context

I will submit a PR with a fix

Chain-Fox:
Fix #4502 

The fix is to use ~~OnceLock~~ Once and atomic variables (for older Rust versions) to replace the original two atomic variables.
OnceLock guarantees the hash can only be initialized once.
This reduces the race condition and provides a slightly better performance.

I also add a test to it to ensure the result before and after patch is consistent.

### DIFF
--- a/crates/backend/src/util.rs
+++ b/crates/backend/src/util.rs
@@ -6,9 +6,8 @@ use std::env;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::iter::FromIterator;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::AtomicUsize;
-use std::sync::atomic::Ordering::SeqCst;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Once;
 
 use crate::ast;
 use proc_macro2::{self, Ident};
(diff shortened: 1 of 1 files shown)


guybedford:
Automated PRs are not appreciated.

PR #4508, opened 2025-05-06
webidl: webgpu: make the `size` argument to `copyBufferToBuffer` optional

andyleiserson:
This matches the latest spec.

There is also a 3-argument overload of `copyBufferToBuffer`, which I have not added, as it poses some additional challenges, and I do not immediately need it.

See https://www.w3.org/TR/webgpu/#commands-buffer-copies for context for the following explanation.

The logic in wasm-bindgen that generates unique names for the different overloads does not understand that the 3-argument overload is omitting the 2nd and 4th arguments to the 5-argument overload. So it generates names based on every argument starting with the 2nd, which (1) changes existing names for variants of the 5-argument overload, (2) includes `with_gpu_buffer` for the 3rd argument in the names even though that has no actual disambiguation purpose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 * Deprecate async constructors.
   [#4402](https://github.com/rustwasm/wasm-bindgen/pull/4402)
+* The `size` argument to `GPUCommandEncoder.copyBufferToBuffer` is now
+  optional. [#4508](https://github.com/wasm-bindgen/wasm-bindgen/pull/4508)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,9 @@
 
 * Deprecate async constructors.
   [#4402](https://github.com/rustwasm/wasm-bindgen/pull/4402)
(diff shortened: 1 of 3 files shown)


andyleiserson:
@daxpedda apologies for the ping, do you have a minute to review this? Excluding the generated code, it's a very small change, and having it upstream helps us to avoid tracking local changes to our vendored copy. Thanks!

daxpedda (changes requested):
LGTM otherwise!

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 * Deprecate async constructors.
   [#4402](https://github.com/rustwasm/wasm-bindgen/pull/4402)
+* The `size` argument to `GPUCommandEncoder.copyBufferToBuffer` is now
+  optional. [#4508](https://github.com/wasm-bindgen/wasm-bindgen/pull/4508)
```
```suggestion

* The `size` argument to `GPUCommandEncoder.copyBufferToBuffer` is now
  optional.
  [#4508](https://github.com/wasm-bindgen/wasm-bindgen/pull/4508)
```

daxpedda (approved):

PR #4512, opened 2025-05-16
Accept number array in TypeScript for number slice arguments

### LINKED ISSUE #4504: TypeScript type for number slices parameters should accept array of number
### Describe the Bug

For Rust type like `Box<[u32]>` in parameter, wasm-bindgen generates `Uint32Array` as type in the TypeScript declaration. However, the glue code generated uses [`TypedArray.set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set) internally to pass the content into the linear memory, and that method actually accepts array of number as well.

Having the type being only `TypedArray` forces consumer to create an extra `TypedArray` and copy content into it first, which should be unnecessary.

upsuper:
This PR fixes #4504.

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -1700,7 +1700,12 @@ fn adapter2ts(
         AdapterType::String => dst.push_str("string"),
         AdapterType::Externref => dst.push_str("any"),
         AdapterType::Bool => dst.push_str("boolean"),
-        AdapterType::Vector(kind) => dst.push_str(&kind.js_ty()),
+        AdapterType::Vector(kind) => {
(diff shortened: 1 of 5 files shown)


upsuper:
Closing as it's merged in our fork https://github.com/upsuper-forks/wasm-bindgen

guybedford:
Thank you for posting. I think you might want to restrict this to just actual number slices - so to exclude Uint8Array and its variants as well as other cases that don't correspond to number[] directly. In addition some types seem to output here like `any[] | number[]` which seems like a bug?

I'd recommend working through these test failures further as well as considering other cases as well if you want to upstream.

PR #4514, opened 2025-05-28
Remove experimental flag for Symbol.dispose

### LINKED ISSUE #4513: Remove experimental flag for Symbol.dispose
### Motivation

With the explicit resource management proposal moving to stage 4 (https://bsky.app/profile/robpalmer.bsky.social/post/3lqaieimpq222), and every major runtime and browser supporting it by default now (https://v8.dev/features/explicit-resource-management), we should remove the experimental flag `WASM_BINDGEN_EXPERIMENTAL_SYMBOL_DISPOSE` and generate `Symbol.dispose` definition by default.

ketanhwr:
Closes #4513

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1193,16 +1193,14 @@ __wbg_set_wasm(wasm);"
             wasm_bindgen_shared::free_function(name),
         ));
         ts_dst.push_str("  free(): void;\n");
-        if self.config.symbol_dispose {
-            dst.push_str(
-                "
-                [Symbol.dispose]() {{
-                    this.free();
-                }}
-                ",
-            );
-            ts_dst.push_str("  [Symbol.dispose](): void;\n");
-        }
+        dst.push_str(
+            "
+            [Symbol.dispose]() {{
+                this.free();
+            }}
+            ",
+        );
+        ts_dst.push_str("  [Symbol.dispose](): void;\n");
         dst.push_str(&class.contents);
         ts_dst.push_str(&class.typescript);
 
@@ -2724,7 +2722,7 @@ __wbg_set_wasm(wasm);"
     pub fn generate(&mut self) -> Result<(), Error> {
         self.prestore_global_import_identifiers()?;
         // conditionally override Symbol.dispose
-        if self.config.symbol_dispose && !self.aux.structs.is_empty() {
+        if !self.aux.structs.is_empty() {
             self.expose_symbol_dispose()?;
         }
 
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -39,7 +39,6 @@ pub struct Bindgen {
     multi_value: bool,
     encode_into: EncodeInto,
     split_linked_modules: bool,
-    symbol_dispose: bool,
 }
 
 pub struct Output {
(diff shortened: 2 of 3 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -2724,7 +2722,7 @@ __wbg_set_wasm(wasm);"
     pub fn generate(&mut self) -> Result<(), Error> {
         self.prestore_global_import_identifiers()?;
         // conditionally override Symbol.dispose
-        if self.config.symbol_dispose && !self.aux.structs.is_empty() {
+        if !self.aux.structs.is_empty() {
             self.expose_symbol_dispose()?;
```
ketanhwr, line 2726: I'm not completely sure if I should remove `expose_symbol_dispose()` as well.

guybedford:
`Symbol.dispose` is still not supported in Safari - https://caniuse.com/mdn-javascript_builtins_symbol_dispose.

I think a middle ground here might be to have `const symbolDispose = Symbol.dispose || Symbol('dispose')` and then use that throughout and removing the flag once that is proven out.

guybedford:
@daxpedda if I'm understanding you right are you saying the output should define `Symbol.dispose = Symbol.dispose || Symbol('dispose')` or to instead use `const symbolDispose = Symbol.dispose || Symbol('dispose')` in the implementation?

I do think we need to pick one of these to land this change. I think not overriding `Symbol.dispose` is problably best so the second option would be my recommendation.

daxpedda:
I actually meant to not change the implementation at all.

Why would we not want to overwrite `Symbol.dispose`?

guybedford:
> Why would we not want to overwrite Symbol.dispose?

It's never a good idea to mutate the environment in JS.

daxpedda:
> > Why would we not want to overwrite Symbol.dispose?
> 
> It's never a good idea to mutate the environment in JS.

I don't think that's a concern here. These are our own types we define. So in practice we aren't overriding anything.

Maybe you are thinking of imported types here?

crates/cli-support/src/js/mod.rs
```diff
@@ -2724,7 +2722,7 @@ __wbg_set_wasm(wasm);"
     pub fn generate(&mut self) -> Result<(), Error> {
         self.prestore_global_import_identifiers()?;
         // conditionally override Symbol.dispose
```
daxpedda, line 2724: I think the wording here is misleading, as its not actually overriding anything.

Am I missing something here?
Or is this just JS terminology?

guybedford (changes requested):
I'm referring to this line: https://github.com/wasm-bindgen/wasm-bindgen/blob/main/crates/cli-support/src/js/mod.rs#L1709

Basically, when we do `global.Symbol.dispose = Symbol('Symbol.dipose')` we are mutating the global JS environment.

Instead, we should do `const symbolDispose = Symbol.dispose || Symbol('Symbol.dispose')`.

Should be a straightforward fix, but marking clearly as a requirement for this to move out of experimental.

daxpedda:
I didn't know about this. We should definitely not do that, agreed.
Apologies, you were talking about the right thing, I was just confused and should have asked what you were talking about here exactly.

Do you know why is it necessary in the first place?
When I try adding an unsupported `Symbol` to a class, it seems to work just fine.
So I'm a bit confused why we have this workaround in place to begin with.

guybedford:
In Safari, when `Symbol.dispose` does not exist, we define it into the global, and then reference `Symbol.dispose` in all usage.

Instead, we should always reference a private local variable `symbolDispose` which is set from the global `Symbol.dispose` when available, or created from our private symbol `Symbol('Symbol.dispose')`.

One benefit of defining `Symbol.dispose` is that users can also call it, but the solution there is to use a global polyfill library before loading wasm bindgen code.

wasm-bindgen should not itself be polyfilling global features is the point as that's not the expectation users have.

daxpedda:
> wasm-bindgen should not itself be polyfilling global features is the point as that's not the expectation users have.

Totally agreed.

@ketanhwr let me know if anythings unclear.

guybedford:
I've posted up what was discussed here in https://github.com/wasm-bindgen/wasm-bindgen/pull/4666, let's move the conversation there.

PR #4520, opened 2025-06-09
Add some context to when node can`t be executed

dgnfu:
When node is missing or can`t be found by the test-runner the error presented is currently only `File not found (os error 2)` without giving any hint that it is node that is missing. To remove some confusion I suggest adding some context to the error message.


guybedford (approved):
Great context to add, thank you.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
```diff
@@ -146,7 +146,8 @@ pub fn execute(
         .arg("--expose-gc")
         .args(&extra_node_args)
         .arg(&js_path)
-        .status()?;
+        .status()
+        .context("failed to find or execute node")?;
```
guybedford:
```suggestion
        .context("failed to find or execute Node.js")?;
```

PR #4522, opened 2025-06-19
Error on invalid import attr position

RReverser:
The reference tests for imports used invalid syntax, and because we don't error on attributes in invalid position by default, it was silently passing anyway.

I've changed the import attributes to error by default (by the way, why do we not do this for all attributes? it seems like a pretty common footgun), fixed the test, changed it to `--target web` so that we could actually see imported files in the JS snapshot, and updated snapshots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 * Added isPointInFill and isPointInStroke methods for the SVGGeometryElement idl.
   [#4509](https://github.com/wasm-bindgen/wasm-bindgen/pull/4509)
+
 * Stricter checks for `module`, `raw_module` and `inline_js` attributes applied to inapplicable items.
   [#4522](https://github.com/wasm-bindgen/wasm-bindgen/pull/4522)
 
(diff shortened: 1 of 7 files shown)


RReverser:
Judging by other PRs CI status, quite a few of these errors are coming simply from new Rust version and would happen on main branch too. I guess that needs to be fixed upstream first.

RReverser:
> It would be good if you could go through blame and find out why errors were disabled in the first place. In any case, if we don't find a good reason (I predict we won't) I'm happy to break things like these and say sorry later.

I have a suspicion it's rather that wasm-bindgen used to not validate attributes at all in the beginning, so any validations became opt-in on case-by-case basis to avoid breaking existing code.

Worst case scenario, we can break it in the new major version whenever we make one, but yeah, if there's no good reason to keep them from validating today, it's worth exploring.

guybedford:
Rebase and changelog entry on this?

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@
 
 * Added isPointInFill and isPointInStroke methods for the SVGGeometryElement idl.
   [#4509](https://github.com/wasm-bindgen/wasm-bindgen/pull/4509)
+* Stricter checks for `module`, `raw_module` and `inline_js` attributes applied to inapplicable items.
```
daxpedda, line 24:
```suggestion

* Stricter checks for `module`, `raw_module` and `inline_js` attributes applied to inapplicable items.
```

PR #4526, opened 2025-06-30
feat: Stream to AsyncIterator

jpopesculian:
An attempt to implement the second part of #2399. I haven't written tests yet, but I wanted to have one of y'all take a look at it first before I invest more time into it. Of course, there's weirdness here because promises require 'static, but I figured since we're in a non-threaded context anyways it would be okay to just wrap the stream in an RefCell. However this does mean that there's no guarantees on order if you do something like

```JavaScript
let iter = async_iterable_from_rust();
let out = Promise.all([iter.next(), iter.next()])
```

To counteract that, you could probably just create a chain of oneshot channels but I wanted to get your feedback on the RefCell stuff before I did any more work on it.

### DIFF
--- a/crates/futures/src/stream.rs
+++ b/crates/futures/src/stream.rs
@@ -79,3 +79,68 @@ impl Stream for JsStream {
         }
     }
 }
+
+#[cfg(feature = "std")]
+mod to_iter {
+    use super::*;
+    use crate::future_to_promise;
+    use core::ops::DerefMut;
+    use js_sys::{Boolean, Object, Promise, Reflect};
+    use std::cell::RefCell;
+    use std::rc::Rc;
+
+    struct RefCellStreamFuture<S>(Rc<RefCell<S>>);
+
+    impl<S> Clone for RefCellStreamFuture<S> {
+        fn clone(&self) -> Self {
+            Self(self.0.clone())
+        }
+    }
+
+    impl<S> RefCellStreamFuture<S> {
+        fn new(stream: S) -> Self {
+            Self(Rc::new(RefCell::new(stream)))
+        }
+    }
+
+    impl<S> Future for RefCellStreamFuture<S>
+    where
+        S: Stream + Unpin,
+    {
+        type Output = Option<S::Item>;
+        fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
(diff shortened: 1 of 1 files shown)


crates/futures/src/stream.rs
```diff
@@ -79,3 +79,68 @@ impl Stream for JsStream {
...
+                            Reflect::set(&result, &"value".into(), &value).unwrap();
+                        }
+                        Ok(result.into())
+                    }
+                    Err(err) => Err(err),
+                }
+            })
+        });
+        let out = Object::new();
+        Reflect::set(&out, &"next".into(), &closure.into_js_value()).unwrap();
```
fl0rek, line 140: Just setting the `next` property isn't enough to then use `for async ( .. of ..)` syntax in js - I'm getting `iterator is not iterable` error. To make it work I've had to also set the [Symbol.asyncIterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncIterator/Symbol.asyncIterator) to a method which returns self. May not be the best way to do it (should it set the prototype instead?), but:
```
    let iterator = Object::new();
    let iterator_self = iterator.clone();
    let iterator_return_this =
        Closure::<dyn FnMut() -> JsValue>::new(move || iterator_self.clone().into())
            .into_js_value();

    Reflect::set(&iterator, &"next".into(), &next).expect("reflect shouldn't fail on Object");
    Reflect::set(&iterator, &Symbol::async_iterator(), &iterator_return_this)
        .expect("reflect shouldn't fail on Object");
```

guybedford: Ideally could this be done with https://github.com/wasm-bindgen/wasm-bindgen/pull/4230?

fl0rek: Actually, yes. 
Method I mentioned above is incorrect, I think cloning the JsValue causes a reference cycle, and returned value seems to never get garbage collected in js. 
Instead of that I resorted to creating a struct that has a method that returns self, that I then transplant with Reflect to Symbol.asyncIterator. Seems like adding above would drop the need for this weird move.

guybedford: Thanks for the update, I'm hoping to pick that one up in the coming weeks.

PR #4535, opened 2025-07-18
Fix wasm-bindgen-test compatibility with doctests in Node.js

drewcrawford:
This commit addresses crashes when running wasm-bindgen-test in doctests, which execute in Node.js environments with specific limitations.

Problem:
When running doctests with wasm-bindgen-test, async code using spawn_local() would crash with "not a shared typed array" error. This occurred because:
1. Doctests run in Node.js, not browsers
2. Node.js may not have SharedArrayBuffer enabled by default
3. Even with atomics compiled in, the runtime memory might not be shared

Solution:
1. SharedArrayBuffer detection: Before using Atomics.waitAsync(), we now check if the memory buffer is actually a SharedArrayBuffer. If not, we gracefully fall back to the polyfill implementation.

2. Worker availability handling: In Node.js doctests, Worker may not be available or may behave differently. The polyfill now:
   - Checks for Worker constructor availability before attempting to use it
   - Falls back to setTimeout-based polling when Worker is unavailable
   - Implements exponential backoff (1ms → 2ms → 4ms... up to 100ms)
   - Polls up to 10 times before timing out

Design decisions:
- The SharedArrayBuffer check uses inline JS to properly detect the type, as this can't be reliably done from Rust alone
- The polling fallback provides basic async behavior without Worker deps, making doctests functional even in restricted environments
- The implementation prioritizes compatibility over performance in fallback scenarios, as doctests are primarily for documentation/testing

This enables wasm-bindgen-test to work reliably in doctest environments while maintaining full functionality in production browser contexts.

### DIFF
--- a/crates/futures/src/task/multithread.rs
+++ b/crates/futures/src/task/multithread.rs
@@ -175,12 +175,23 @@ fn wait_async(ptr: &AtomicI32, current_value: i32) -> Option<js_sys::Promise> {
         ))
     } else {
         let mem = wasm_bindgen::memory().unchecked_into::<js_sys::WebAssembly::Memory>();
-        let array = js_sys::Int32Array::new(&mem.buffer());
-        let result = Atomics::wait_async(&array, ptr.as_ptr() as u32 / 4, current_value);
-        if result.async_() {
-            Some(result.value())
+        let buffer = mem.buffer();
+
+        // Check if the memory buffer is actually a SharedArrayBuffer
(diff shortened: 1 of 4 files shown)


drewcrawford:
Closing my 30+-day stale PR for lack of review.  Please reopen and ping me if you are interested in further contributions.

guybedford:
I'm happy to review this if you'd like to reopen. It would of course help if we can ensure simple test coverage.

PR #4541, opened 2025-07-25
fix: generated JS eqeqeq

nicoleneo:
The generated JavaScript code does not satisfy eslint's [eqeqeq rule](https://eslint.org/docs/latest/rules/eqeqeq). I changed the JavaScript snippets to use eqeqeq.

Compiler warning:
```
src/pkg/index.js

  Line 123:56:  Expected '!==' and instead saw '!='
```

Generated code `pkg/index.js`

```javascript
async function __wbg_load(module, imports) {
    if (typeof Response === 'function' && module instanceof Response) {
        if (typeof WebAssembly.instantiateStreaming === 'function') {
            try {

            } catch (e) {
                if (module.headers.get('Content-Type') != 'application/wasm') {
                    console.warn("`WebAssembly.instantiateStreaming` failed because your server does not serve Wasm with `application/wasm` MIME type. Falling back to `WebAssembly.instantiate` which is slower. Original error:\n", e);

                } else {
                    throw e;
                }
            }
        }

        const bytes = await module.arrayBuffer();
        return await WebAssembly.instantiate(bytes, imports);

    } else {
        const instance = await WebAssembly.instantiate(module, imports);

        if (instance instanceof WebAssembly.Instance) {
            return { instance, module };

        } else {
            return instance;
        }
    }
}
```

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -923,7 +923,7 @@ __wbg_set_wasm(wasm);"
                                 return await WebAssembly.instantiateStreaming(module, imports);
 
                             }} catch (e) {{
-                                if (module.headers.get('Content-Type') != 'application/wasm') {{
+                                if (module.headers.get('Content-Type') !== 'application/wasm') {{
                                     console.warn(\"`WebAssembly.instantiateStreaming` failed \
                                                     because your server does not serve Wasm with \
                                                     `application/wasm` MIME type. Falling back to \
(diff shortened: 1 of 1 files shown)


RReverser:
Autogenerated code should be excluded from linter configurations, it's definitely not meant to match any particular style (or, generally speaking, even be readable in the first place).

guybedford:
Thanks for the PR, closing per the comment, but happy to discuss further.

PR #4548, opened 2025-08-04
Unbreak CI + cleanup config a bit for newer Rust

RReverser:


daxpedda (approved):
Thanks!

daxpedda:
Feel free to merge it like this if you want, we can fix more of the CI in separate PRs. Up to you ofc.

RReverser:
I'd rather do everything in one go, will use this PR for more fixes.

RReverser:
Hm, a bit odd that the function order in `.wat` files I got from `BLESS=1` locally is different between my machine and CI expectations. Rust version is the same, the OS is not (Windows vs Linux) but surely that shouldn't affect the function order...

RReverser:
> the OS is not (Windows vs Linux) but surely that shouldn't affect the function order...

Apparently that's exactly what it is, got a different order when reran in WSL.

RReverser:
Aside from weird Firefox issue on CI, looks like example-tests have been broken since https://github.com/wasm-bindgen/wasm-bindgen/pull/4058 - can't run them locally due to `_package.json` being referenced but no longer existing after that PR.

daxpedda:
Examples have been very flaky for a long time unfortunately, even before #4058.
AFAIR its just an issue with the Firefox webdriver and not actually a bug on our side.

But yeah, can be retried a couple of times until it hits green ...

RReverser:
> Examples have been very flaky for a long time unfortunately, even before #4058.

Yeah aside from FF issue I'm referring not to the flake but to the example-tests/tests/webpack.rs being completely broken for local after that PR so you can't just do `cargo test -p example-tests`, because it tries to find `_package.json` which doesn't exist.

Working on a fix, I believe I just need to replicate CI-specific setup from https://github.com/wasm-bindgen/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/.github/workflows/main.yml#L368-L374...

RReverser:
Ah nice, merging as-is for now then.

PR #4551, opened 2025-08-04
Fix local runs of example-tests

RReverser:
Without this change, local `example-tests` were broken since #4058 and would only work on CI as CI had special setup that created pnpm-workspace.yaml.

We now have pnpm-workspace.yaml statically in the repo with some extra config to share dep versions in a centralised place and few mroe drive-by fixups to make tests work locally again.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -373,9 +373,9 @@ jobs:
         function build() {
           (cd $1 &&
             if [ -f package.json ]; then
-              echo ok # pnpm run build -- --output-path ../../exbuild/$1
+              pnpm run build -- --output-path ../../exbuild/$1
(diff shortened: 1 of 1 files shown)


daxpedda (commented):
It all looks reasonable to me, but unfortunately I've no idea about NPM so can't really review this. So feel free to merge at your leisure (or maybe somebody else will find the time to look at it).

Not sure what goings on with the `build_examples` CI job, but I think its an actual issue not caused by flakiness.

RReverser:
> Not sure what goings on with the `build_examples` CI job, but I think its an actual issue not caused by flakiness.

Yeah definitely real issue; I was debugging it yesterday, but gave up as it was getting late, will continue today.

PR #4554, opened 2025-08-05
Fix incorrect assertion when load and store memory

Spxg:
```wat
(func $init_pthread_self (type 1)
    (local i32)
    i32.const 0
    i32.const 1307316
    i32.store offset=1307436
    .....)
```
```
Executing bindgen...                              
thread 'main' panicked at crates/wasm-interpreter/src/lib.rs:245:21:
init_pthread_self: Read a negative address value from the stack. Did we run out of memory?
```
*After fixing this problem, it still fails due to insufficient memory, but we will not discuss it here.*

The address from stack can be 0.

https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Memory/Load
https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Memory/Store

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 * `TypedArray::from(&[T])` now works reliably across memory reallocations.
   [#4555](https://github.com/wasm-bindgen/wasm-bindgen/pull/4555)
 
-* Fix incorrect assertion when loading and storing memory
+* Fix incorrect memory loading and storing assertions during post-processing.
   [#4554](https://github.com/wasm-bindgen/wasm-bindgen/pull/4554)
 
 --------------------------------------------------------------------------------


Spxg:
Might be able to close  https://github.com/wasm-bindgen/wasm-bindgen/issues/4446 and https://github.com/cloudflare/workers-rs/issues/772

guybedford:
There's no reason to read from memory 0 though, so yes this won't be the root cause of anything.

Spxg:
> so yes this won't be the root cause of anything.

Actually, this example store 1307316 to address 0 + 1307436, but fails due to an incorrect assertion. It's same for load.

Spxg:
In addition, according to several recent reports, insufficient memory space can cause failures.
It is recommended to change it to configurable, may be via environment variable: 
https://github.com/wasm-bindgen/wasm-bindgen/blob/c58f64ebb962ea73d17857272d240a9b8054c80c/crates/wasm-interpreter/src/lib.rs#L74

Spxg:
> I think the right fix is to instead move the assertion down, after we calculate the full address with offset.

Yes, I also think this is more reasonable, but I am not sure what the intention of this contributor is, so I adopted a more conservative modification.

RReverser:
I think this would make the check basically useless, since it won't catch genuine loads/stores from address zero anymore. Pretty sure it was a mistake / bug in the original code, let's check the calculated address instead.

Spxg:
> Given this provided example

When I  link to the [SQLite3MC](https://github.com/utelle/SQLite3MultipleCiphers)  (debug profile), it included [init_pthread_self](https://github.com/emscripten-core/emscripten/blob/7f173272225e75fea27f0e3e2f3a4a39692e7865/system/lib/pthread/pthread_self_stub.c#L32). This is the logic inside it:
```wat
(func $init_pthread_self (type 1)
    (local i32)
    i32.const 0
    i32.const 1307316
    i32.store offset=1307436
    .....)
```

I disabled the interpreter for `__wasm_call_ctors` in another PR.

Spxg:
> let's check the calculated address instead.

done

RReverser (approved):
Thanks!

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 * Add a workaround for `TextDecoder` failing in older version of Safari when too many bytes are decoded through it over its lifetime.
   [#4472](https://github.com/wasm-bindgen/wasm-bindgen/pull/4472)
 
+* Fix incorrect assertion when load and store memory
```
RReverser:
```suggestion
* Fix incorrect assertion when load and store memory

```

daxpedda (approved):

PR #4555, opened 2025-08-05
Fix flaky `TypedArray::from(&[T])`

### LINKED ISSUE #4492: TypedArray::from(slice) sporadically fails
### Describe the Bug

Closely related to #4395 - `TypedArray::from` with slice input sporadically fails: "Cannot perform Construct on a detached ArrayBuffer". Implementation: https://github.com/rustwasm/wasm-bindgen/blob/c35cc9369d5e0dc418986f7811a0dd702fb33ef9/crates/js-sys/src/lib.rs#L6383

Note that the function is not marked `unsafe` itself.

### Steps to Reproduce

Add test case to `crates/js-sys/tests/wasm/TypedArray.rs`:
```rust
#[wasm_bindgen_test]
fn from_succeeds_when_heap_grows() {
    let _ = (0..10_000).map(|i| {
        Int32Array::from([i].as_slice())
    }).collect::<Vec<_>>();
}
```

Run test:
```
# cargo test --target wasm32-unknown-unknown from_succeeds_when_heap_grows
[...]
running 1 test
test TypedArray::from_succeeds_when_heap_grows ... FAIL

failures:

---- TypedArray::from_succeeds_when_heap_grows output ----
    error output:
        wasm-bindgen: imported JS function that was not marked as `catch` threw an error: Cannot perform Construct on a detached ArrayBuffer

        Stack:
        TypeError: Cannot perform Construct on a detached ArrayBuffer
            at new Int32Array (<anonymous>)
```

RReverser:
Instead of going via an intermediate unsafe view, we can add another safe binding that takes a slice "directly" and clones it into a typed array.

This is safer, generates slightly compact JS with fewer roundtrips, and fixes the flakiness due to allocations that used to happen between creating the slice and actually cloning it.

Fixes #4492.

### DIFF
--- a/crates/example-tests/src/lib.rs
+++ b/crates/example-tests/src/lib.rs
@@ -166,13 +166,16 @@ impl WebDriver {
 
         // Connect to the Firefox instance.
         let start = Instant::now();
+        let wait_sec = 60;
         let ws = loop {
             match tokio_tungstenite::connect_async(format!("ws://{driver_addr}/session")).await {
                 Ok((ws, _)) => break ws,
                 Err(e) => {
-                    if start.elapsed() > Duration::from_secs(20) {
-                        return Err(e).context("failed to connect to Firefox (after 20s)");
+                    if start.elapsed().as_secs() > wait_sec {
+                        return Err(e)
+                            .context(format!("failed to connect to Firefox (after {wait_sec}s)"));
                     }
+                    eprintln!("Failed to connect to Firefox: {e}, retrying...");
                 }
             }
         };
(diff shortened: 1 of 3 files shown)


Spxg:
[to_vec](https://docs.rs/js-sys/latest/js_sys/struct.Uint8Array.html#method.to_vec) has similar problem.

RReverser:
I know, I've seen your comments on like a dozen issues 😅

Feel free to send a PR - I think a reliable repro test would be the best first step here, e.g. the test from #4492 has been very helpful for this PR.

Spxg:
> I know, I've seen your comments on like a dozen issues 😅

emmm, I'm sorry about that.

daxpedda:
@RReverser would you mind adding a changelog entry for this?

PR #4556, opened 2025-08-05
make mime error detection more accurate

jrvidal:
`instantiateStreaming` might fail for a [variety of reasons](https://webassembly.github.io/spec/web-api/#dom-webassembly-instantiatestreaming). We can't just assume that an invalid MIME type is an oversight on the server. For instance, many 4xx or 5xx responses will be served as `text/html` so re-trying with `instantiate()` is just bound to cause more confusion.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -904,7 +904,11 @@ __wbg_set_wasm(wasm);"
                                 return await WebAssembly.instantiateStreaming(module, imports);
 
                             }} catch (e) {{
-                                if (module.headers.get('Content-Type') != 'application/wasm') {{
+                                const expectedTypes = new Set(['basic', 'cors', 'default']);
+
+                                const validResponse = module.ok && expectedTypes.has(module.type);
+
+                                if (validResponse && module.headers.get('Content-Type') !== 'application/wasm') {{
                                     console.warn(\"`WebAssembly.instantiateStreaming` failed \
                                                     because your server does not serve Wasm with \
                                                     `application/wasm` MIME type. Falling back to \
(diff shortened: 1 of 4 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -904,7 +904,11 @@ __wbg_set_wasm(wasm);"
                                 return await WebAssembly.instantiateStreaming(module, imports);
 
                             }} catch (e) {{
-                                if (module.headers.get('Content-Type') != 'application/wasm') {{
+                                const expectedTypes = new Set(['basic', 'cors', 'default']);
+
+                                const validResponse = module.ok && expectedTypes.has(module.type);
```
guybedford: Do we really need the expectedTypes as well here, or is the `module.ok` enough?

jrvidal: My main annoyance is indeed with failed requests, but when I actually read the [spec](https://webassembly.github.io/spec/web-api/#compile-a-potential-webassembly-response), I also found another possibility:

> * If response is not [CORS-same-origin](https://html.spec.whatwg.org/#cors-same-origin), [reject](https://webidl.spec.whatwg.org/#reject) returnValue with a [TypeError](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror) and abort these substeps.

guybedford: But even in the CORS case we need to validate that the response is Wasm as well, so effectively making the Wasm validation be checked first seems fine to me.

jrvidal: If the request failed due to an invalid type, there's probably nothing else to do, as the body of the response is not available for the other values of `type`.

guybedford: Right, there are no headers in this case too. Okay, in that case, let's at least make sure to hoist the set population though out of the inner body to be statically defined. Alternatively just split the statement into three string equality checks.

guybedford (approved):

PR #4562, opened 2025-08-06
Skip `__wasm_call_ctors` which we don't want to interpret

Spxg:
```rust
use wasm_bindgen::prelude::wasm_bindgen;

const _: () = {
    unsafe extern "C" fn __ctor() {
        let _ = 1 + 1;
    }

    #[unsafe(link_section = ".init_array")]
    static __CTOR: unsafe extern "C" fn() = __ctor;
};

#[wasm_bindgen]
pub fn add(x: i32, y: i32) -> i32 {
    x + y
}

#[cfg(test)]
mod tests {
    wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);

    use wasm_bindgen_test::wasm_bindgen_test;

    #[wasm_bindgen_test]
    fn it_works() {}
}
```
run with `wasm-pack build --target web --debug` or `wasm-pack test --headless`: 
```
Executing bindgen...                              
thread 'main' panicked at crates/wasm-interpreter/src/lib.rs:245:21:
wbgmem::_::__ctor::h5405e240fd081e6d: invalid binary op I32And
```

resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/2673
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/3315
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4446
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4491

It can be confirmed that the failure is caused by interpreting `__wasm_call_ctors`. Since the rust's `__wasm_call_ctors` was introduced in 1.85, there is no problem with 1.84. 

`wasm32-unknown-unknown` is now C-ABI, so programs linked with enscripten may also have this problem.

### DIFF
--- a/crates/wasm-interpreter/src/lib.rs
+++ b/crates/wasm-interpreter/src/lib.rs
@@ -80,7 +80,15 @@ fn skip_calls(module: &Module, id: FunctionId) -> HashSet<FunctionId> {
         .instrs
         .iter()
         .filter_map(|(instr, _)| match instr {
+            // There are only up to three calls for now:
+            //   1. __wasm_call_ctors (`#[link_section = ".init_array"]`)
+            //   2. __wbindgen_skip_interpret_calls (The original symbol, we don't care about it)
+            //   3. __wasm_call_dtors (This symbol may not be present in Rust program, but may be present if C program is linked)
             Instr::Call(Call { func }) | Instr::ReturnCall(ReturnCall { func }) => Some(*func),
+            // Typically, there are no other instructions, or only a return instruction.
+            //
+            // When coverage is turned on, there may be llvm coverage instrumentation
+            // instructions.
             _ => None,
         })
         .collect()


Spxg:
A more stable way to do this is needed, since there is no symbol name in release

Spxg:
If `__wasm_call_ctors` is not exported, `wasm-ld` will wrap each export symbol into `{export}.command_export`, ~~so we can skip the call of the first function (that is, `__wasm_call_ctors`).~~  It is not available when `strip = true`

If `__wasm_call_ctors` is exported, `__wasm_call_ctors` will not be inserted into the export function and this problem will not occur:
```rust
use wasm_bindgen::prelude::wasm_bindgen;

const _: () = {
    unsafe extern "C" fn __ctor() {
        let _ = 1 + 1;
    }

    #[unsafe(link_section = ".init_array")]
    static __CTOR: unsafe extern "C" fn() = __ctor;
};

#[wasm_bindgen]
pub fn __hack() {
    unsafe extern "C" {
        fn __wasm_call_ctors();
    }
    unsafe {
        __wasm_call_ctors();
    }
}

#[wasm_bindgen]
pub fn add(x: i32, y: i32) -> i32 {
    x + y
}
```

https://github.com/llvm/llvm-project/blob/a196281896de208fca1dde315e377a46ec9a2e66/lld/wasm/Writer.cpp#L1802

Spxg:
confirmed fix https://github.com/cloudflare/workers-rs/issues/772

Spxg:
"build_nightly" CI will fix by https://github.com/rust-lang/rust/pull/145096

src/lib.rs
```diff
@@ -73,6 +73,23 @@ use core::ptr::NonNull;
...
+    /// * it is not relocatable;
+    /// * it is not a position-independent executable;
+    /// * and it does not call `__wasm_call_ctors` or , directly or indirectly, from any
+    ///   exported function.
+    ///
+    /// If treated as "command-style linkage", we calculate the function address of
+    /// `__wasm_call_ctors` and `__wasm_call_dtors` based on this export, used to skip
+    /// interpret.
+    #[no_mangle]
+    pub extern "C" fn __wbindgen_skip_interpret_calls() {}
```
guybedford, line 90: Nice approach. And this definitely gets optimized out?

guybedford:
Thanks for putting together the upstream fix for CI as well. If you want to pull in https://github.com/wasm-bindgen/wasm-bindgen/pull/4571 here while that still hasn't landed feel free.

src/lib.rs
Spxg: Yes, It is the same as `__wbindgen_describe_`, it deletes the symbol after execution and does not appear in the final product

src/lib.rs
```diff
@@ -73,6 +73,23 @@ use core::ptr::NonNull;
 
 use crate::convert::{FromWasmAbi, TryFromJsValue, WasmRet, WasmSlice};
 
+const _: () = {
+    /// The Wasm linker may or may not insert a call to this function from the beginning
```
daxpedda, line 77: This was a bit confusing to me. Saying "this" here implies that `__wbindgen_skip_interpret_calls()` is the function that is called at the beginning of every exported function.

Spxg: This was totally my mistake, "this" refers to ` __wasm_call_ctors`.

Spxg: Fixed by https://github.com/wasm-bindgen/wasm-bindgen/pull/4575

PR #4564, opened 2025-08-06
Deduplicate Closure invocation implementation

### LINKED ISSUE #4119: Cannot wasm-pack build when using a js_sys array method (filter or map)
### Describe the Bug

When I run `wasm-pack build --target web --release` build fails if I use an Array method like `filter` or `map` (I havn't tried other methods). The strange part is that `wasm-pack build --target web --dev` runs perfectly

Cargo.toml
```toml
[package]
name = "transforms-wasm"
version = "0.1.0"
authors = ["Nicolas Bruère <nicolas.bruere@braincube.com>"]
edition = "2018"

[lib]
crate-type = ["cdylib"]

[features]
default = ["console_error_panic_hook"]

[dependencies]
wasm-bindgen = "0.2.93"
js-sys = "0.3.70"

# The `console_error_panic_hook` crate provides better debugging of panics by
# logging them with `console.error`. This is great for development, but requires
# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
# code size when deploying.
console_error_panic_hook = { version = "0.1.7", optional = true }

[dev-dependencies]
wasm-bindgen-test = "0.3.43"

[profile.release]
# Tell `rustc` to optimize for small code size.
opt-level = "s"
```

lib.rs
```rust
#[wasm_bindgen]
pub fn get_histogram(input: &Array, binsize: f64, min: f64, max: f64, dimension: i32) -> Array {
    let output = Array::new();

    let mut i = min;
    while i <= max {
        let arr = Array::new();
        arr.push(&JsValue::from_f64(i));
        let count = input.filter(&mut |item, _, _| {
            let row = Array::from(&item);
            let o = row.at(dimension).as_f64().unwrap();
            if o >= i && o < (i + binsize) {
                return true;
            }
            false
        }).length();
        arr.push(&JsValue::from(count));
        output.push(&arr);
        i += binsize;
    }

    output
}
```

The error output:
```
[INFO]: ⬇️  Installing wasm-bindgen...
error: failed to generates bindings for adapter

Caused by:
    no function table found in module
Caused by: Running the wasm-bindgen CLI
Caused by: failed to execute `wasm-bindgen`: exited with exit status: 1
  full command: "/Users/nicolas.bruere/Library/Caches/.wasm-pack/wasm-bindgen-507ae24572712ef8/wasm-bindgen" "/Users/nicolas.bruere/workspace/reco-explorer/transforms-wasm/target/wasm32-unknown-unknown/release/transforms_wasm.wasm" "--out-dir" "/Users/nicolas.bruere/workspace/reco-explorer/transforms-wasm/pkg" "--typescript" "--target" "web"
```

### Steps to Reproduce

1. Run `wasm-pack build --target web --release` => you get an error
2. Run `wasm-pack build --target web --dev` => you get 0 error

https://we.tl/t-V6pMfd5I4B

### Expected Behavior

I'd like to have the same behavior whether I'm building on dev or release. I'd be ok if both return an error since it might be a mistake I made ;)

### Actual Behavior

The build command outputs an error on release mode and no errors on dev mode

RReverser:
This is a branch I had lying around, just cleaned up.

Basically, I noticed that owned closures (represented via `wasm_bindgen::Closure`) and closure conversions (bindings for `&dyn Fn()/FnMut()`) duplicated the infrastructure for their invocations, even though they're naturally doing exactly same thing.

The only difference between them is in the JS bindings and presence of a destructor, so that's the only thing I'm keeping in `wasm_bindgen::Closure`'s descriptor.

I think there's a lot more that could be deduplicated between them, but for now just wanted to submit what I already had as it's already a noticeable cleanup.

Fixes #4119.

### DIFF
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -101,7 +101,6 @@ pub struct Function {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Closure {
-    pub shim_idx: u32,
     pub dtor_idx: u32,
     pub function: Function,
     pub mutable: bool,
@@ -263,12 +262,14 @@ fn get_string(data: &mut &[u32]) -> String {
 
 impl Closure {
     fn decode(data: &mut &[u32]) -> Closure {
-        let shim_idx = get(data);
         let dtor_idx = get(data);
-        let mutable = get(data) == REFMUT;
+        let mutable = match get(data) {
+            0 => false,
+            1 => true,
+            other => panic!("expected bool value, got {other}"),
+        };
         assert_eq!(get(data), FUNCTION);
         Closure {
-            shim_idx,
             dtor_idx,
             mutable,
             function: Function::decode(data),
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -199,7 +199,7 @@ impl<'a> Context<'a> {
                 let mut function = descriptor.function.clone();
                 function.arguments.insert(0, Descriptor::I32);
                 function.arguments.insert(0, Descriptor::I32);
-                let adapter = self.table_element_adapter(descriptor.shim_idx, function)?;
+                let adapter = self.table_element_adapter(descriptor.function.shim_idx, function)?;
                 self.aux.import_map.insert(
                     id,
                     AuxImport::Closure {
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -167,9 +167,7 @@ impl<'a> Context<'a> {
 
             // If any closures exist we need to prevent the function table from
             // getting gc'd
-            if !closure_imports.is_empty() {
-                self.aux.function_table = self.module.tables.main_function_table()?;
-            }
+            self.aux.function_table = self.module.tables.main_function_table()?;
 
             // Register all the injected closure imports as that they're expected
             // to manufacture a particular type of closure.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,11 @@
 * Fix incorrect memory loading and storing assertions during post-processing.
   [#4554](https://github.com/wasm-bindgen/wasm-bindgen/pull/4554)
 
-* Fix test `--exact` option not working as expected
+* Fix test `--exact` option not working as expected.
   [#4549](https://github.com/wasm-bindgen/wasm-bindgen/pull/4549)
+
+* Fix tables being removed even though they are used by stack closures.
+  [#4119](https://github.com/wasm-bindgen/wasm-bindgen/pull/4564)
 
 --------------------------------------------------------------------------------
 
--- a/crates/cli/tests/wasm-bindgen/main.rs
+++ b/crates/cli/tests/wasm-bindgen/main.rs
@@ -469,6 +469,32 @@ fn function_table_preserved() {
 }
 
 #[test]
+fn function_table_preserved_for_stack_closures() {
+    let (mut cmd, _out_dir) = Project::new("function_table_preserved_for_stack_closures")
+        .file(
+            "src/lib.rs",
+            r#"
+                use wasm_bindgen::prelude::*;
+
+                #[wasm_bindgen]
+                extern "C" {
+                    fn take_closure(closure: &dyn Fn());
+                }
+
+                #[wasm_bindgen]
+                pub extern fn pass_closure() {
+                    take_closure(&|| {
+                        // Noop, just ensure that the compilation succeeds.
+                        // See https://github.com/wasm-bindgen/wasm-bindgen/issues/4119.
+                    });
+                }
+            "#,
+        )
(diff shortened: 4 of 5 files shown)


crates/cli-support/src/descriptor.rs
```diff
@@ -263,12 +262,14 @@ fn get_string(data: &mut &[u32]) -> String {
 
 impl Closure {
     fn decode(data: &mut &[u32]) -> Closure {
-        let shim_idx = get(data);
         let dtor_idx = get(data);
-        let mutable = get(data) == REFMUT;
+        let mutable = match get(data) {
+            0 => false,
+            1 => true,
```
guybedford, line 268: Can we keep the `REFMUT` constant here?

crates/cli-support/src/wit/mod.rs
```diff
@@ -167,9 +167,7 @@ impl<'a> Context<'a> {
 
             // If any closures exist we need to prevent the function table from
             // getting gc'd
-            if !closure_imports.is_empty() {
-                self.aux.function_table = self.module.tables.main_function_table()?;
-            }
+            self.aux.function_table = self.module.tables.main_function_table()?;
```
guybedford, line 170: Will this prevent the function table from getting gc'd in the case of closures not existing?

daxpedda: Is this change related to the `Closure` code de-duplication? I would like to know in detail what the motivation behind this change is.

RReverser: See corresponding commit description: https://github.com/wasm-bindgen/wasm-bindgen/pull/4564/commits/bdc08885af1fe7b8169113a2b047b7d62db5cf26

It's not strictly related, but I did run into it while working on closure deduplication as mentioned in https://github.com/wasm-bindgen/wasm-bindgen/issues/4119#issuecomment-3044617365 - the issue is that tables are preserved only for `wasm_bindgen::Closure` but not for `&dyn Fn()` / `&mut dyn FnMut()` closures.

I think it should be safe to split out though, one change works without the other, I just happened to trigger the same bug in one of my manual tests. If there's a better fix, I'm happy to leave it to someone too.

crates/cli-support/src/descriptor.rs
RReverser: Yeah, I just didn't like that this place used a bit overly generic value (so in theory binary could have anything, not just REF and REFMUT).

At first I changed the `match` to just throw on anything else, but then figured might as well represent it as a proper boolean to make it clearly it can only hold two states.

crates/cli-support/src/wit/mod.rs
guybedford: Can we add a test case for what this fixes then?

RReverser: I spent about 2 hours trying to make one, but unfortunately it's somewhat tricky in the current setup, at least without writing some custom non-portable shell scripts or something, definitely not as part of `cargo test`.

Since this is post-linking processing, it's very susceptible to which crates end up linked in, and cargo invoking linker with wasm-bindgen-test or some of the other stray crates in the workspace is enough to hide the bug and make it pass even without a fix.

daxpedda: I suspect there was some reason why it was initially implemented this way, so we might now introduce an unnecessary table in certain scenarios? But I don't think this is a big problem really and fixing the bug takes precedence ofc.

> Since this is post-linking processing, it's very susceptible to which crates end up linked in, and cargo invoking linker with wasm-bindgen-test or some of the other stray crates in the workspace is enough to hide the bug and make it pass even without a fix.

Don't [these test](https://github.com/wasm-bindgen/wasm-bindgen/tree/771e1f068706c141701a6de0413862e2dd9eb030/crates/cli/tests/reference) avoid exactly this problem?

In any case, we shouldn't block on a test if you already tried for 2 hours!

daxpedda (approved):

crates/cli-support/src/wit/mod.rs
RReverser: > Don't [these test](https://github.com/wasm-bindgen/wasm-bindgen/tree/771e1f068706c141701a6de0413862e2dd9eb030/crates/cli/tests/reference) avoid exactly this problem?

Oh yeah, good point, I've tried multiple other approaches but not this one. I'll take another quick look.

RReverser: Yup, reference tests are perfect for this. No idea why I haven't thought of them, thanks @daxpedda.

RReverser: Oh and there is also this test that I could've easily duplicated

https://github.com/wasm-bindgen/wasm-bindgen/blob/771e1f068706c141701a6de0413862e2dd9eb030/crates/cli/tests/wasm-bindgen/main.rs#L453-L469

For some reason I got stuck trying to add top-level tests instead of thinking of CLI-specific ones yesterday, my bad 🤦‍♂️

daxpedda: On that note, would you mind adding something regarding #4119 to the changelog?

PR #4567, opened 2025-08-07
Properly share list of tys

RReverser:
As far as I can tell, wasm-bindgen crate already depends on wasm-bindgen-shared via macro -> macro-support, so this shouldn't introduce issues with publishing order, and makes sharing the opcodes less error-prone as it no longer relies on human factor.


guybedford (approved):

RReverser:
Ah, looks like this requires making wasm-bindgen-shared no-std... okay.

PR #4570, opened 2025-08-08
feat: allow imported future return function types for primitives & JsCast

guybedford:
I was finding it very limiting that imported async functions cannot have well-typed return values. This adds support for richer future return types in imported functions by applying a dynamic casting of primitive, optional, result and `JsCast` types.

This is effectively a form of runtime type conversion, since like closures these cases sit outside the static bindgen since Promise.then is typed by externref any. We thus need to replicate the WebAssembly `ToWebAssemblyValue` coercions that happen in core Wasm which we otherwise assume in the `FromAbi` layers. This is done by extending `TryFromJsValue` to all the primitive types and ensuring it mostly follows the `ToWebAssemblyValue` specifications.

`JsCast` is used to do these unchecked conversions, with a new `JsValueCast` type that is able to apply to primitives as well without requiring reference support.

All dynamic checks are unsafe in the sense that they will panic if not getting the expected value type. Better handling of conversion errors would be something worth thinking more about - perhaps via an attribute which throws a runtime error on conversion issues and unifies that with the `JsValue` error type as a JS error. I'm open to suggestions here.

In the process I also implemented a couple of helpers - (1) the `is_null_or_undefined()` intrinsic was useful to replicate the optional functionality on the JsValue "externref". I also added a `Justfile` for more easily running local operations, for example tests can now be run with `just test-wasm-bindgen`.


tests/wasm/futures.rs
```diff
@@ -170,13 +170,13 @@ async fn test_promise_ok_string() {
...
-        Err(Some(String::from("error")))
-    )
-}
+// #[wasm_bindgen_test]
+// async fn test_promise_err() {
+//     assert_eq!(
+//         call_promise_err().await.map_err(|j| j.as_string()),
+//         Err(Some(String::from("error")))
+//     )
+// }
```
guybedford:
```suggestion
#[wasm_bindgen_test]
async fn test_promise_err() {
    assert_eq!(
        call_promise_err().await.map_err(|j| j.as_string()),
        Err(Some(String::from("error")))
    )
}
```

src/lib.rs
```diff
@@ -369,6 +369,38 @@ impl JsValue {
...
+    /// assert_eq!(result, None);
+    /// ```
+    #[inline]
+    pub fn if_defined(&self) -> Option<JsValue> {
+        if self.is_null_or_undefined() {
+            None
+        } else {
+            Some(self.clone())
+        }
+    }
```
daxpedda: I'm hesitant to add more API surface to the `JsValue` function. What exactly is the motivation behind this?

justfile
daxpedda, line 1: I guess this has been a long time coming, as local testing was very unpleasant up until now.

My experience so far has been that adding more and more tool requirements to an open source project doesn't exactly improve the maintenance story. I've seen quite a lot of Rust projects that built up serious amount of technical debt by wanting remove all the e.g. python scripts from their repo. I would instead suggest to add something similar to `publish.rs` (which I assume was done like this here as well to avoid the same pitfall).

Currently, to run the whole testsuite we require, in addition to the Rust toolchain: NPM and Geckodriver. I would like to keep it at that.

I'm very interested in your opinion about all this.

daxpedda (commented):
Didn't review the code yet, but I think this was one of the most obvious missing features in `wasm-bindgen`s bindings generation ability.

src/lib.rs
guybedford: Calling `v.is_undefined() || v.is_null()` leads to two separate calls out through the Wasm boundary. So having a single `is_null_or_undefined` is a performance optimization for this kind of hot path.

With `is_null_or_undefined` then defined, the useful feature for optional types is an `if_defined` option map, which allows turning code like `if !v.is_null_or_undefined() { /* do something with v */ }` into a flattened optional pattern using more idiomatic Rust - `v.if_defined().map(|v| /* do something with v */ }`. The other benefit of this being that we only need to use the `v` binding reference once which I then used to optimize the output Rust bindgen to avoid declaring a separate local variable as we can replace `v` with a computed expression.

We could certainly refactor away from the `if_defined()` pattern though too and I could update the codegen to use a local variable to define `v` instead.

justfile
guybedford: Let me strip this down to a minimal justfile then that only has a few very simple commands.

The general approach I have seen taken to Justfile adoption in projects is that it is always optional, and simply used as a store of script invocations that one would otherwise need to remember. But that running everything without Just should always be possible if you remember the right incantations or have them written down.

Overall the productivity improvement is one for the core team though only.

src/lib.rs
daxpedda: I was only talking about `JsValue::if_defined()`. I'm totally in favor of `is_null_or_undefined()`.

> The other benefit of this being that we only need to use the `v` binding reference once which I then used to optimize the output Rust bindgen to avoid declaring a separate local variable as we can replace `v` with a computed expression.
> 
> We could certainly refactor away from the `if_defined()` pattern though too and I could update the codegen to use a local variable to define `v` instead.

I'm not really following. Why is optimizing the Rust bindgen to not use a local variable relevant/valuable? AFAICS its also just used in a single place, so we aren't really saving any source code lines as well.

I would also call it more idiomatic Rust if the type parameter in `map(|v| ...)` is actually a type representing a defined variable. But we don't have a type distinction for that. Ergo if you really need an `Option` here you can just call `v.is_null_or_undefined().then(|| { v; ... })`.

justfile
daxpedda: Sounds good enough to me! Would still be valuable to contributors though, would you mind adding something to the `CONTRIBUTING.md`?

src/convert/impls.rs
```diff
@@ -624,3 +624,163 @@ where
...
+    }
+}
+
+impl TryFromJsValue for f32 {
+    type Error = JsError;
+
+    fn try_from_js_value(js_value: JsValue) -> Result<Self, Self::Error> {
+        // WebAssembly ToNumber(v) algorithm with f32 conversion
+        let number = unsafe { crate::__wbindgen_as_number(js_value.idx) };
+        let result = if number.is_nan() {
```
RReverser: This shouldn't be necessary, `number as f32` already casts NAN too.

src/convert/impls.rs
```diff
@@ -624,3 +624,163 @@ where
...
+    }
+}
+
+impl TryFromJsValue for f64 {
+    type Error = JsError;
+
+    fn try_from_js_value(js_value: JsValue) -> Result<Self, Self::Error> {
+        // WebAssembly ToNumber(v) algorithm
+        let number = unsafe { crate::__wbindgen_as_number(js_value.idx) };
+        let result = if number.is_nan() { f64::NAN } else { number };
```
RReverser: Same here.

src/convert/impls.rs
```diff
@@ -624,3 +624,163 @@ where
...
+            .ok_or_else(|| JsError::new("String is empty"))
+    }
+}
+
+// Smaller integer types - use i32/u32 conversion then cast
+impl TryFromJsValue for i8 {
+    type Error = JsError;
+
+    fn try_from_js_value(js_value: JsValue) -> Result<Self, Self::Error> {
+        i32::try_from_js_value(js_value).map(|v| v as i8)
```
RReverser: This seems to circumvent the point of `TryFrom` methods - if you want fallible casts, wouldn't you want to go via `v.try_into()` to also catch range errors?

src/lib.rs
guybedford: Sure, I'll refactor it out.

src/convert/impls.rs
```diff
@@ -624,3 +624,163 @@ where
...
+    }
+}
+
+impl TryFromJsValue for u64 {
+    type Error = JsError;
+
+    fn try_from_js_value(js_value: JsValue) -> Result<Self, Self::Error> {
+        // WebAssembly ToBigUint64(v) algorithm
+        crate::bigint_get_as_i64(&js_value)
+            .map(|i64_val| i64_val as u64)
```
RReverser, line 674: Should probably also use `try_into()` if it's meant to be fallible.

guybedford: I'm refactoring these to rather be a new infallible trait implementation under the assumption that type system violations are panics by definition.

crates/cli-support/src/js/mod.rs
```diff
@@ -3638,6 +3638,11 @@ __wbg_set_wasm(wasm);"
                 format!("{} === null", args[0])
             }
 
+            Intrinsic::IsNullOrUndefined => {
+                assert_eq!(args.len(), 1);
+                format!("{} === null || {} === undefined", args[0], args[0])
```
RReverser: Nit: I'd go for `== null` which is equivalent but saves some code size.

src/convert/impls.rs
```diff
@@ -624,3 +624,163 @@ where
     }
     result.into_boxed_slice()
 }
+
+// WebAssembly ToWebAssemblyValue implementations for primitive types
+// These implement the exact WebAssembly specification algorithms
+
+impl TryFromJsValue for i32 {
```
RReverser: I think it would be nice to split those changes out into a separate PR, makes a bit confusing to see what is related to async import support specifically and what are these traits for.

RReverser:
> since like closures these cases sit outside the static bindgen since Promise.then is typed by externref any

The main complexity with closures is variadic arg signature, although I want to simplify that too.

With promises, I think it's enough to generate extra bindings. A common approach in wasm-bindgen is to generate an "extension" type that allows for extra specialization or extra methods, e.g. you can do:

```rust
#[wasm_bindgen]
extern "C" {
  type PromiseExt;

  #[wasm_bindgen(js_name = "then")]
  fn then_i32(callback: &Closure<dyn FnMut(i32)>);

  #[wasm_bindgen(js_name = "then")]
  fn then_bool(callback: &Closure<dyn FnMut(bool)>);

  ...
}
```

and just rely on the existing binding system to generate the necessary conversions.

guybedford:
I'm working on an update here that adds support for `TypedPromise<T>` as well, but didn't want to overload this PR with that work too. But the foundations require the `FromFromJsValue` work which I've now split out into a new trait.

Open to putting everything all in one PR or splitting up as suits.

src/convert/impls.rs
guybedford: See comment below re splitting up, this foundation is key to TypedPromise<T> and `TypedFunction<..???..>`.

src/convert/impls.rs
guybedford: Added, thanks.

justfile
guybedford: Thanks, added.

src/lib.rs
guybedford: Removed.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 
 ### Added
 
+* Allow richer runtime casts for imported async function return types (e.g., `JsString`, `Uint8Array`, `Object`, etc.) instead of being limited to only `JsValue`.
+  [#????](https://github.com/rustwasm/wasm-bindgen/pull/????)
```
guybedford:
```suggestion
  [#4570](https://github.com/wasm-bindgen/wasm-bindgen/pull/4570)
```

src/convert/impls.rs
```diff
@@ -624,3 +625,123 @@ where
     }
     result.into_boxed_slice()
 }
+
+// WebAssembly ToWebAssemblyValue implementations for primitive types
+// These implement the exact WebAssembly specification algorithms
```
guybedford, line 630: Continuing from the other comment - @RReverser mentioned an alternative approach here is to import a JS function with various types to implicitly handle the casting without an explicit implementation.

This would involve defining `function (x) -> x` and then importing it as taking an externref argument, but returning as each of the 4 core Wasm types in its imported variations, then using that to handle casting implicitly.

My argument remains this is more code and work overall than just having these simple checks which are well-defined.

src/cast.rs
daxpedda, line 1: If I'm understanding this correctly: because `JsFuture` is not typed and therefor only allows a single return type, `JsValue`, what we do here is applying the conversion to the type we want by going through `JsValue` in Wasm instead of e.g. being able to receive primitive types directly like we do in non-async functions. The downside here is that this isn't on par with the performance we usually give, because we have to make an extra roundtrip to JS here.

So in the end this future is purely a QoL feature extension of the proc-macro, to handle conversion for the user, we aren't actually adding new binding capabilities.

If I got all this right, I would instead suggest of starting implement a typed `JsFuture`/`Promise` before implementing this. Then we can provide proper binding support equal to non-async functions.

daxpedda: Aha, just saw #4573.

src/convert/impls.rs
daxpedda: Cc https://github.com/wasm-bindgen/wasm-bindgen/pull/4573#discussion_r2264177304.

guybedford:
Based on discussion with @daxpedda I'm going to move to starting from the first class runtime typed `Promise<T>` type, then rather come back to this later on as an addition to that foundational work instead.

RReverser:
@guybedford If possible, it would be great if you could look into making `#[wasm_bindgen]` itself support generic types and methods. I think it's pretty doable, just needs a bunch of work, and it would allow us to naturally make all relevant types generic - not only `Promise`, but also `Array`, `Map`, `Set` and so on - without introducing new APIs just for generic counterparts.

guybedford:
@RReverser yes this is exactly what I'm working on, as the more general cases @daxpedda pointed me too. Will share more details soon.

PR #4573, opened 2025-08-08
Typed futures and promises

guybedford:
This builds off of https://github.com/wasm-bindgen/wasm-bindgen/pull/4570 providing support for `TypedFuture` and `TypedPromise`.

Based to that PR.

See tests/wasm/typed_promise.rs for usage examples.


src/convert/impls.rs
```diff
@@ -624,3 +625,123 @@ where
     }
     result.into_boxed_slice()
 }
+
+// WebAssembly ToWebAssemblyValue implementations for primitive types
+// These implement the exact WebAssembly specification algorithms
```
RReverser, line 630: I still really don't think we should reimplement things which are already part of the JS/Wasm engines and are already provided "for free".

Those spec operations fall into that bucket - they are intended for engine implementors, not library consumers, and engines already provide them in form of typed function calls for us.

All you'd need to do is declare an intrinsic that generates literally a function like `function(x) { return x; }` and binds it with different types, under one name as `externref -> i32`, under another name as `externref -> char` and so on.

This would reduce both implementation complexity and generated code size - win-win.

RReverser: The other question is, you mentioned that custom casting is required here because it's similar to closures and requires dynamic / generic typing.

But isn't it _literally_ closures rather than just similar?

If you declare a single binding for `Promise::then_func(cb: js_sys::Function)` then you can implement any typed future via `Closure<dyn FnMut(T)>::into_js_function` and you don't need a new casting system at all.

guybedford: > under one name as externref -> i32, under another name as externref -> char and so on.

I don't think we'd do `char` - rather I think we'd want the Wasm core types in particular and then we'd have four new imports. We're talking about like 30 lines though. Introducing new intrinsics and imported functions is not free.

> Those spec operations fall into that bucket - they are intended for engine implementors, not library consumers, and engines already provide them in form of typed function calls for us.

I don't quite follow this argument. Are there specific maintenance aspects you are concerned about in expanding out https://www.w3.org/TR/wasm-js-api-2/#towebassemblyvalue in particular here?

> This would reduce both implementation complexity and generated code size - win-win.

I think once the new intrinsics are taken into account this would actually be more code not less in wasm bindgen. This area of the Wasm spec is also frozen and stable so it's not like we have to track anything.

Not that I'm completely against it, but just that I don't see why the work should be justified in the root PR (not this is introduced in #4570 not this PR so we should move discussion there rather).

guybedford: > If you declare a single binding for Promise::then_func(cb: js_sys::Function) then you can implement any typed future via Closure<dyn FnMut(T)>::into_js_function and you don't need a new casting system at all.

The problem being solved here is being able to type JS promises within Rust, so that you don't need dynamic casts or wrappers, not in representing Rust closures as typed JS promises. Or am I misunderstanding your question?

daxpedda: So RReverser was literally saying what I was thinking right now: the `JsFuture` implementation actually uses `Closure`s for the return value. We don't even need new intrinsics, we just need to make use of the already existing system to return the right type.

daxpedda: > Not that I'm completely against it, but just that I don't see why the work should be justified in the root PR (not this is introduced in #4570 not this PR so we should move discussion there rather).

I had some thoughts on why the PR order should be reversed in https://github.com/wasm-bindgen/wasm-bindgen/pull/4570#discussion_r2264687957.

daxpedda (commented):
I think we should thoroughly explore if we can make this work with default generic types, without a breaking change, before we are forced to introduce new types.

E.g. `struct JsFuture<T = JsValue>`.

src/convert/impls.rs
RReverser: > The problem being solved here is being able to type JS promises within Rust, so that you don't need dynamic casts or wrappers, not in representing Rust closures as typed JS promises. Or am I misunderstanding your question?

What @daxpedda and me are trying to say - JS promises in Rust are already implemented via essentially `then(_: Closure<dyn FnMut(JsValue)>)`, the only thing you'd need to change is make the binding generic so that you can pass any `Closure<dyn FnMut(T)>` and the rest falls in place automatically.

> Are there specific maintenance aspects you are concerned about in expanding out https://www.w3.org/TR/wasm-js-api-2/#towebassemblyvalue in particular here?

Main concern is all of these types are already implemented in wasm-bindgen, and we wouldn't want those types in async functions to potentially behave slightly different in some edge cases than they do in any other functions. To ensure that, it's better to work with the existing binding system and its traits rather than introduce yet another one.

That way we can both ensure that types behave the same across different bindings, _and_ avoid introducing even more code - that's what I meant by win-win.

guybedford: > So RReverser was literally saying what I was thinking right now: the JsFuture implementation actually uses Closures for the return value. We don't even need new intrinsics, we just need to make use of the already existing system to return the right type.

How to handle passing promises around in the non future case then? Eg, exporting functions that take types promises as parameters, or having typed promises as properties of objects etc?

This doesn't just support `async fn`, this supports a first class handle to `TypedPromise<T>` that can be used in arbitrary places.

My goal here is to create an isomorphism with the TypeScript type system - first class runtime types are a requirement for that. Static is of course important, but the point is the runtime type story here for better typing when working with JS.

guybedford:
> I think we should thoroughly explore if we can make this work with default generic types, without a breaking change, before we are forced to introduce new types.

This simply isn't possible to ensure comprehensive runtime typing through arbitrary interface usage, since not every type is an argument or return of a static function, and we need to support higher order types as export parameters and arbitrary return positions (e.g. an object returning a promise for an object with a promise property that resolves to a JS function that returns a promise for an integer - this is all runtime typed handles).

src/convert/impls.rs
RReverser: > How to handle passing promises around in the non future case then? Eg, exporting functions that take types promises as parameters, or having typed promises as properties of objects etc?

Exactly same approach - they'd simply take `Closure<dyn FnMut(T)>` instead of `Closure<dyn FnMut(JsValue)>` that the untyped `Promise` currently takes.

Let's talk next week, I'm happy to help with implementation.

guybedford: So you want to cast all JS functions into `Closure<dyn FnMut<T>`? Did you look at https://github.com/wasm-bindgen/wasm-bindgen/pull/4574 yet for how much better typed functions can be?

RReverser: > So you want to cast all JS functions into `Closure<dyn FnMut<T>`?

No, we are talking about the other way around here, how to pass Rust callbacks to JS.

I'm just saying, let's generally discuss design before making huge changes as there can be simpler and more atomic solutions to some of these problems.

guybedford: This PR is designed to support typing JS callbacks passed to Rust. Happy to discuss design further, we're on a call now if you're on Discord.

RReverser: Definitely not on a weekend 😅

src/convert/impls.rs
```diff
@@ -624,3 +625,123 @@ where
...
+
+impl JsValueCast for i32 {
+    fn unchecked_from_js_value(js_value: JsValue) -> Self {
+        // WebAssembly ToInt32(v) algorithm
+        let number = js_value.unchecked_into_f64();
+        if !number.is_finite() || number == 0.0 || number == -0.0 {
```
MaxGraey, line 636: `number == 0.0 || number == -0.0` is redundant (just use number == 0.0). Comparison with zero is sign-agnostic according to IEEE754

guybedford:
Based on discussion with @daxpedda I'm going to close this an reopen it in a form that can be defined as a lower level generic Promise primitive instead of TypedPromise, and also see if we can avoid JsValue casts (although no guarantees).

PR #4515, opened 2025-05-31
Implement array! and json! proc macros to make creation/instantiation easier

dcechano:
This PR merges two proc macros that make `js_sys::Array` and `js_sys::Object` creation and instantiation easier.

 Currently, to create an `Object`, you must use the `Reflect` API. Retrieving a property requires a `Reflect::get` call for every attempted retrieval. Likewise, to add a key-value pair to an `Object`, we must make a `Reflect::set` call for every property we would like to set. If we are adding an `Object` to another `Object`, that `Object` must be instantiated the same way before finally setting it in the target `Object`. This process is very cumbersome and quickly clutters the code.

The `json!` and `array!` proc macros address this problem. `json!` allows the creation of a `js_sys::Object` using JavaScript object literal syntax:

```rust
let john = json! {
    name: "John",
    age: 26
};

let has_name = Object::has_own_property(&john, &"name".into());
let has_age = Object::has_own_property(&john, &"age".into());
assert!(has_name);
assert!(has_age);
```  

Similarly, the `array!` macro facilitates the creation of `js_sys::Array`s using JavaScript array literal syntax:

```rust
let arr = array![1, 2, 3];
let expected = Array::of3(&1.into(), &2.into(), &3.into());
assert_eq!(arr.get(0).as_f64().unwrap(), expected.get(0).as_f64().unwrap());
assert_eq!(arr.get(1).as_f64().unwrap(), expected.get(1).as_f64().unwrap());
assert_eq!(arr.get(2).as_f64().unwrap(), expected.get(2).as_f64().unwrap());
```

Both macros support the use of variables in the JSON as long as the underlying type implements `Into<JsValue>`. Comments are also supported. I have been thinking about how useful a macro like this could be for quite some time, and I finally decided to implement it. Please check out the README.md for more information, as I've only covered the basics here. Let me know your thoughts!

PR #4574, opened 2025-08-09
Dynamically Typed functions

guybedford:
This adds support for `TypedFunction<Args, Ret, This?>` which is useful for e.g. return values of imported functions that return JS functions that we want to have typing applied for.

Example:

```rust
#[wasm_bindgen]
fn return_typed_function() -> TypedFunction<(), i32>;
```

defined a function that returns a typed function taking no arguments and returning an `i32`. It can then be used with:

```rust
let result: i32 = returned_fn.call();
assert_eq!(result, 20);
```

where `call()` automatically infers the argument length, so we don't need `calln` variations etc.

Based to https://github.com/wasm-bindgen/wasm-bindgen/pull/4573.


daxpedda (commented):
I think this is a great addition. However, I'm asking myself if we couldn't introduce a `wasm-bindgen-util` crate or something like that where we can dump all of this non-bindings related stuff. #4515 could live there as well.

Let me also add another motivation to move this off to a separate crate, as this topic seems to be a reoccurring theme: we can much more easily make breaking changes to another crate then we can to `wasm-bindgen`, which is stuck until we solve much bigger problems. E.g. the same applies to `wasm-bindgen-futures`, `js-sys` and `web-sys`, where we can make breaking changes without breaking the whole ecosystem.

PR #4579, opened 2025-08-11
Add helper macro for "generic" casts

RReverser:
The comment on the macro is longer than implementation itself, but TLDR this hacky macro allows us to significantly reduce number of specially supported intrinsics and amount of unsafe code in favour of just going via JS + wasm-bindgen's existing type conversions.

It also provides basis that allows us to remove even more intrinsics in future PRs.

### DIFF
--- a/src/cache/intern.rs
+++ b/src/cache/intern.rs
@@ -31,11 +31,14 @@ cfg_if! {
 
         fn intern_str(key: &str) {
             CACHE.with(|cache| {
-                let mut cache = cache.entries.borrow_mut();
+                let entries = &cache.entries;
 
                 // Can't use `entry` because `entry` requires a `String`
-                if !cache.contains_key(key) {
-                    cache.insert(key.to_owned(), JsValue::from(key));
+                if !entries.borrow().contains_key(key) {
+                    // Note: we must not hold the borrow while we create the `JsValue`,
+                    // because it will try to look up the value in the cache first.
+                    let value = JsValue::from(key);
+                    entries.borrow_mut().insert(key.to_owned(), value);
                 }
             })
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,6 +166,35 @@ const JSIDX_TRUE: u32 = JSIDX_OFFSET + 2;
 const JSIDX_FALSE: u32 = JSIDX_OFFSET + 3;
 const JSIDX_RESERVED: u32 = JSIDX_OFFSET + 4;
 
+// This macro is a hack to implement "generic" casts and reduce number of
+// boilerplate intrinsics. The implementation generates a no-op JS adapter that
+// simply takes an argument in one type, decodes it from the ABI, does nothing
+// with it on the JS side (by declaring function name as empty, so instead of
+// generating typical JS call that does `ret = foo(arg);` we end up with
+// just `ret = (arg);` and then encoding same value back with a different type.
+//
+// Note that we intentionally keep this macro undocumented as it's not meant
+// for public consumption - it relies on our implementation details, and
+// is intended only for simplifying our own implementation details as well.
+//
+// Someday we'll support generics in #[wasm_bindgen] macro, in which case
+// this can be replaced with a proper generic intrinsic.
+#[macro_export]
+#[doc(hidden)]
+macro_rules! wbg_cast {
+    ($value:expr, $from:ty, $to:ty) => {{
+        use $crate::prelude::wasm_bindgen;
+
+        #[wasm_bindgen(wasm_bindgen = $crate)]
+        extern "C" {
+            #[wasm_bindgen(js_name = "/* cast */")]
+            fn __wbindgen_cast(value: $from) -> $to;
+        }
+
+        __wbindgen_cast($value)
+    }};
+}
+
 impl JsValue {
     /// The `null` JS value constant.
     pub const NULL: JsValue = JsValue::_new(JSIDX_NULL);
@@ -194,7 +223,7 @@ impl JsValue {
     #[allow(clippy::should_implement_trait)] // cannot fix without breaking change
     #[inline]
     pub fn from_str(s: &str) -> JsValue {
-        unsafe { JsValue::_new(__wbindgen_string_new(s.as_ptr(), s.len())) }
+        wbg_cast!(s, &str, JsValue)
     }
 
     /// Creates a new JS value which is a number.
@@ -203,7 +232,7 @@ impl JsValue {
     /// allocated number) and returns a handle to the JS version of it.
     #[inline]
     pub fn from_f64(n: f64) -> JsValue {
-        unsafe { JsValue::_new(__wbindgen_number_new(n)) }
+        wbg_cast!(n, f64, JsValue)
     }
 
     /// Creates a new JS value which is a bigint from a string representing a number.
@@ -533,7 +562,7 @@ impl JsValue {
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus)
     #[inline]
     pub fn unchecked_into_f64(&self) -> f64 {
-        unsafe { __wbindgen_as_number(self.idx) }
+        wbg_cast!(self, &JsValue, f64)
     }
 }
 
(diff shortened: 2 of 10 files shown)


src/lib.rs
```diff
@@ -1034,13 +1063,7 @@ macro_rules! try_from_for_num128 {
 try_from_for_num128!(i128, i64);
 try_from_for_num128!(u128, u64);
 
-big_numbers! {
-    |n|,
-    i64 = __wbindgen_bigint_from_i64(n),
-    u64 = __wbindgen_bigint_from_u64(n),
-    i128 = __wbindgen_bigint_from_i128((n >> 64) as i64, n as u64),
-    u128 = __wbindgen_bigint_from_u128((n >> 64) as u64, n as u64),
```
guybedford, line 1042: Did you drop this `(n >> 64) as x64` step? Note that v128 throws in the `ToJSValue()` here https://www.w3.org/TR/wasm-js-api-2/#tojsvalue so we need to make sure we're still not trying to pass it directly.

src/lib.rs
```diff
@@ -166,6 +166,35 @@ const JSIDX_TRUE: u32 = JSIDX_OFFSET + 2;
...
+// generating typical JS call that does `ret = foo(arg);` we end up with
+// just `ret = (arg);` and then encoding same value back with a different type.
+//
+// Note that we intentionally keep this macro undocumented as it's not meant
+// for public consumption - it relies on our implementation details, and
+// is intended only for simplifying our own implementation details as well.
+//
+// Someday we'll support generics in #[wasm_bindgen] macro, in which case
+// this can be replaced with a proper generic intrinsic.
+#[macro_export]
```
guybedford: Even though this is hidden I'd still be weary of adding something to the public API...

Could we just replicate this definition in the other crate usage instead to avoid any risk of reliance?

src/lib.rs
```diff
@@ -166,6 +166,35 @@ const JSIDX_TRUE: u32 = JSIDX_OFFSET + 2;
...
+#[doc(hidden)]
+macro_rules! wbg_cast {
+    ($value:expr, $from:ty, $to:ty) => {{
+        use $crate::prelude::wasm_bindgen;
+
+        #[wasm_bindgen(wasm_bindgen = $crate)]
+        extern "C" {
+            #[wasm_bindgen(js_name = "/* cast */")]
```
guybedford: What about making the cast type part of the comment?

```suggestion
            #[wasm_bindgen(js_name = concat!("/* cast ", stringify!($from), " -> ", stringify!($to), " */"))]
```

src/lib.rs
RReverser: We don't need these custom steps, because, like with other removed conversions, wasm-bindgen already handles `i128` conversions for arbitrary bindings:

https://github.com/wasm-bindgen/wasm-bindgen/blob/a81a601efde4982446dbee3f21b3e0b77bb44b7e/src/convert/impls.rs#L38-L48

Essentially what I was trying to describe in the other issue, just letting the existing binding system work for us instead of duplicating same code.

src/lib.rs
RReverser: Hm we could but it doesn't seem any more risky than the rest of our ABI infrastructure (all the public-but-not-really traits like WasmDescribe / FromWasmAbi / etc.)? Or the entire `__rt` namespace.

src/lib.rs
RReverser: Yeah could definitely do that.

RReverser: Ah no, I forgot, I tried, but it won't work correctly because the proc-macro then receives tokens `concat`, `!`, `stringify` and so on individually and can't handle them.

It's possible to work around with more proc-macro hacks but eventually decided it's not worth the complexity.

RReverser: Hm although I see we already depend on `paste` crate... might as well use that.

src/lib.rs
guybedford: Those other traits are documented though? Don't have a strong opinion here, but just a suggestion.

src/lib.rs
RReverser: No, that one is only for identifiers. It's rather complicated, unfortunately.

guybedford: Makes sense. Yeah we don't want to start implementing compile time mangling...

src/lib.rs
RReverser: Yeah but they're also unstable implementation details.

The `__rt` namespace/module might be a better example. I can just move macro there to have those "private things for macro" colocated together.

RReverser: Moved.

src/cache/intern.rs
```diff
@@ -31,11 +31,14 @@ cfg_if! {
...
+                let entries = &cache.entries;
 
                 // Can't use `entry` because `entry` requires a `String`
-                if !cache.contains_key(key) {
-                    cache.insert(key.to_owned(), JsValue::from(key));
+                if !entries.borrow().contains_key(key) {
+                    // Note: we must not hold the borrow while we create the `JsValue`,
+                    // because it will try to look up the value in the cache first.
+                    let value = JsValue::from(key);
+                    entries.borrow_mut().insert(key.to_owned(), value);
```
daxpedda, line 41: Why the change here? Is it an improvement somehow that I'm not seeing?

src/rt/mod.rs
```diff
@@ -21,6 +20,34 @@ pub mod marker;
...
+// boilerplate intrinsics. The implementation generates a no-op JS adapter that
+// simply takes an argument in one type, decodes it from the ABI, does nothing
+// with it on the JS side (by declaring function name as empty, so instead of
+// generating typical JS call that does `ret = foo(arg);` we end up with
+// just `ret = (arg);` and then encoding same value back with a different type.
+//
+// Note that we intentionally keep this macro undocumented as it's not meant
+// for public consumption - it relies on our implementation details, and
+// is intended only for simplifying our own implementation details as well.
+//
```
daxpedda:
```suggestion
```
Everything in the `rt` module isn't meant to be public and is `doc(hidden)` already.

src/cache/intern.rs
RReverser: As the comment says, because now `JsValue::from` goes via generic conversion which tries to look up the string in cache (via `cache.borrow()`) while we're still holding the `RefMut` which results in panic.

The other alternative fix is to change lookup code from `cache.borrow()` to `cache.try_borrow()?`, perhaps that's cleaner.

src/rt/mod.rs
RReverser: Unfortunately, that doesn't apply to `macro_rules!`. They don't behave by normal scoping rules and always end up in the global scope of the crate, thus visible even if the module is not and not accessible via `crate::module::macro_name` like you'd expect. One of my personal pet peeves for many years 😅

src/cache/intern.rs
daxpedda: Aaah, I misunderstood the comment for just the regular Rust lifetime stuff because we already hold a mutable reference. So I was surprised why it worked before then.

That's fine then.

src/rt/mod.rs
daxpedda: Yes, right, it still requires `doc(hidden)`. What I was trying to say is that we have a bunch of macros inside the `rt` module, we don't need to tell developers on every macro that this isn't for public consumption.

So `rt` implies its not for public consumption so it shouldn't be here otherwise. The comment confused me for that reason.

RReverser: Ah ok. FWIW I only moved it to this module a few minutes ago after Guy's review above. Happy to shorten the comment.

src/rt/mod.rs
```diff
@@ -21,6 +20,34 @@ pub mod marker;
...
+// Someday we'll support generics in #[wasm_bindgen] macro, in which case
+// this can be replaced with a proper generic intrinsic.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! wbg_cast {
+    ($value:expr, $from:ty, $to:ty) => {{
+        #[$crate::prelude::wasm_bindgen(wasm_bindgen = $crate)]
+        extern "C" {
+            /// Foobar.
+            #[wasm_bindgen(js_name = "/* cast */")]
```
daxpedda, line 39: Whats actually wrong with just removing this line?
```suggestion
```

daxpedda: Ah ... never mind, my bad. Its in the comment.
I'm a bit confused why nothing changes though.

daxpedda: Nope, I take it all back, didn't run the tests properly.

PR #4580, opened 2025-08-11
Self-host functions that don't need to be real intrinsics

RReverser:
This makes intrinsics that are simple JS functions just bindings to those JS functions, so that 1) we have fewer intrinsics to specially support and 2) less unsafe code.

This intentionally doesn't touch intrinsics I'm already removing in #4579 to avoid duplicate work, even though some of them could be represented in form of function calls too.

### DIFF
--- a/crates/cli/tests/reference/result.js
+++ b/crates/cli/tests/reference/result.js
@@ -81,8 +81,8 @@ export function result_i32() {
     return ret[0];
 }
 
-export function __wbindgen_error_new(arg0, arg1) {
-    const ret = new Error(getStringFromWasm0(arg0, arg1));
+export function __wbg_Error_fcfdb1a705a32d74(arg0, arg1) {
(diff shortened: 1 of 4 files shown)


crates/cli/tests/reference/result.js
```diff
@@ -81,8 +81,8 @@ export function result_i32() {
     return ret[0];
 }
 
-export function __wbindgen_error_new(arg0, arg1) {
-    const ret = new Error(getStringFromWasm0(arg0, arg1));
+export function __wbg_Error_fcfdb1a705a32d74(arg0, arg1) {
+    const ret = Error(getStringFromWasm0(arg0, arg1));
```
RReverser, line 85: For the reviewer: we could bind it as a class to preserve `new`, but this is actually fine - `Error` is one of those APIs that, like `Number`, `String` etc. work as functions too.

daxpedda (approved):

PR #4581, opened 2025-08-11
Add justfile for dev convenience only

guybedford:
Includes the simple Justfile as previously discussed in https://github.com/wasm-bindgen/wasm-bindgen/pull/4570.


daxpedda (approved):
Mh ... maybe I will push against this again when we have `cargo script`.

guybedford:
Happy to reconsider this at any time.

PR #4584, opened 2025-08-14
Deduplicate closure macros

RReverser:
Next cleanup follow-up to #4564.

- Deduplicate the two different macros generating variadic impls for closures into one.
- Deduplicate separate implementations for `Fn` and `FnMut` into the same macro with different trait names as a param.
- Generate impls for `(&A)` in the same macro we generate all the variadic impls, just with different trait param.
- Simplify bounds where possible.
- Make `WasmClosureFnOnce` take a function signature as a param - this also allows to remove a weird inference hack.

What we get is a lot less of almost-but-not-quite duplicate boilerplate and all the closure impls defined from a single place.


src/closure.rs
```diff
@@ -460,10 +454,8 @@ impl Closure<dyn FnOnce()> {
 /// A trait for converting an `FnOnce(A...) -> R` into a `FnMut(A...) -> R` that
 /// will throw if ever called more than once.
 #[doc(hidden)]
-pub trait WasmClosureFnOnce<A, R>: 'static {
-    type FnMut: ?Sized + 'static + WasmClosure;
-
-    fn into_fn_mut(self) -> Box<Self::FnMut>;
+pub trait WasmClosureFnOnce<FnMut: ?Sized>: 'static {
```
guybedford: Realistically, what are the chances of this beaing a breaking change?

src/closure.rs
```diff
@@ -389,12 +388,7 @@ where
     pub fn forget(self) {
         drop(self.into_js_value());
     }
-}
 
-// NB: we use a specific `T` for this `Closure<T>` impl block to avoid every
-// call site having to provide an explicit, turbo-fished type like
-// `Closure::<dyn FnOnce()>::once(...)`.
-impl Closure<dyn FnOnce()> {
```
guybedford, line 397: How does this avoid the problem referenced by this comment?

src/convert/closures.rs
```diff
@@ -1,19 +1,38 @@
-#![allow(clippy::fn_to_numeric_cast)]
+#![allow(clippy::fn_to_numeric_cast, coherence_leak_check)]
```
daxpedda: Can we keep allowing `coherence_leak_check` as local as possible? We don't want to accidentally add more implementation of that in the future.

src/closure.rs
```diff
@@ -422,9 +416,9 @@ impl Closure<dyn FnOnce()> {
     /// // is `FnMut`, even though `f` is `FnOnce`.
     /// let closure: Closure<dyn FnMut() -> String> = Closure::once(f);
     /// ```
-    pub fn once<F, A, R>(fn_once: F) -> Closure<F::FnMut>
+    pub fn once<F>(fn_once: F) -> Self
     where
-        F: 'static + WasmClosureFnOnce<A, R>,
```
daxpedda: I think I understand what you did here overall. But can we really remove generics from a public API like this without a breaking change?

src/closure.rs
daxpedda: Generally speaking `#[doc(hidden)]` is not public API. But I'm also concerned about weird inference issues.
That said, https://github.com/wasm-bindgen/wasm-bindgen/pull/4584#discussion_r2279674296, very much looks like a breaking change to me.

src/convert/closures.rs
RReverser: Literally all the traits in this module need it due to implementations for `Fn(&A)` potentially conflicting with `Fn(A)`, so it seemed to make sense to hoist it out, but sure, no problem with keeping it copy-pasting on each trait either.

src/closure.rs
RReverser: Hm technically, yeah, I can see how it could be a breaking change if someone wrote `Closure::once::<_ /* skipped because can't spell out closure type */, (i32, i32), i32>(|...| ...)`...

But it's so much nicer without the weird `::<dyn FnOnce()>` hack, just letting inference work as it's supposed to.

Oh well, if it's a concern, I guess I can revert this change and we can shove this into "future breaking changes" bucket. Ideally if we go there, I'd like to get rid of the `Closure<>` type altogether - it's rather weird and unnecessary, as it's semantically just `Box<dyn Fn(...)>` and the like, and we could accept those in signatures directly, plus have an `Into<JsValue>` implementation.

Btw I mentioned this to @guybedford before, I'd like us to start a separate `breaking` branch where we can start fixing API stuff that would be breaking changes - e.g. everything under https://github.com/wasm-bindgen/wasm-bindgen/pulls?q=is%3Aopen+is%3Apr+label%3Abreaking-change - so that we don't just delay things indefinitely and keep it all in once place that can be released at some point in future.

guybedford: The main question with regards to breaking is if there is potential real usage patterns affected.

A quick search on https://github.com/search?q=Closure%3A%3Aonce+language%3ARust&type=code doesn't show any turbofish syntax for usage.

We just need to ensure we have confidence - we don't need to be pedantic about breaks, but we must be careful to ensure backwards compatibility for the majority userbase through this transition.

Making that call I will trust to your judgement.

RReverser: Hm good point, we can even search for explicitly turbofish syntax, looks like there's no public code at least https://github.com/search?q=%22Closure%3A%3Aonce%3A%3A%22+language%3ARust&type=code https://sourcegraph.com/search?q=context:global+%22Closure::once::%22&patternType=keyword&sm=0 although that doesn't guarantee anything about private apps built with it.

I'm a bit torn, given that it's a very popular project and we are only updating patch versions at the moment, it's better to be safe than sorry, but also it seems extremely unlikely someone used this API with explicit types, I don't know of any good reason someone would, and the simplification of the type inference is kinda nice.

@daxpedda What do you think?

daxpedda: In my experience so far, this has turned out bad more often then not. Especially with a library like `wasm-bindgen` there is a whole lot of closed-source code out there we can't account for.

I would also say that if we would want to experiment with something like this, it should be the last commit before the release so we can easily revert it if necessary and not start building more code relying on that change.

On the other hand I really don't see a reason right now to do any of that. While the change is "nice", it doesn't enable a new feature or anything.

TLDR: I'm not in favor.

daxpedda: > Btw I mentioned this to @guybedford before, I'd like us to start a separate `breaking` branch where we can start fixing API stuff that would be breaking changes - e.g. everything under https://github.com/wasm-bindgen/wasm-bindgen/pulls?q=is%3Aopen+is%3Apr+label%3Abreaking-change - so that we don't just delay things indefinitely and keep it all in once place that can be released at some point in future.

I predict that maintaining such a branch would be "too much" work. Instead, I would suggest we open a tracking issue with all the potential breaking changes we would want to make. We keep loosing sight of ones we found before and these become harder to dig up again over time.

RReverser: > We keep loosing sight of ones we found before and these become harder to dig up again over time.

That's exactly why I think having a branch would be more helpful. Having bunch of issues that we need to fix "eventually" requires way too much tracking and reduces chances of us ever actually getting around to them, whereas a branch is something we can work on more actively as well as even automate publishing under some `dev` / `alpha` / `nightly` tag to crates.io for early experimentation.

daxpedda: I specifically meant a *single* tracking issue.

I'm fine with a branch, but I kinda doubt we are gonna be able to keep it up-to-date in any reasonable sense.

RReverser: > I specifically meant a _single_ tracking issue.

Yeah I got that, but in my experience it will only make it look even more insurmountable to tackle over time, whereas a branch, even one lagging behind, would at least have some things "done" already.

RReverser: > While the change is "nice", it enable a new feature or anything.

Ah right, one of the reasons I did this particular change is because otherwise I can't merge the implementations together.

Rust thinks `WasmClosureFnOnce<(A,), R>` and `WasmClosureFnOnce<(A&,), R>` might conflict with each other and it's a hard error but `WasmClosureFnOnce<dyn FnMut(A) -> R>` and `WasmClosureFnOnce<dyn FnMut(&A) -> R>` are "fine" (unless they break that warning in some new edition in the future) because functions get special treatment in the type system and reference args turn the whole type into `WasmClosureFnOnce<dyn for<'a> FnMut(&'a A) -> R>` which is distinct from the one for concrete lifetime.

I can either keep this particular implementation separate, but I want to see first if I can just leave those two type params as unused, so we change signature of `WasmClosureFnOnce` (which is a "private" trait anyway) but keep the extra type args on `Closure::once` for backward compat.

RReverser: I put the type params back, mostly unused except for weird inference cases.

Now the only case that wouldn't work the same as before is if someone used `Closure::<dyn FnOnce()>::once(/* a FnOnce with actually multiple params */)` which is still technically a breaking change but I'd say much less likely one and a fair compromise.

src/convert/closures.rs
daxpedda: Indeed, I would prefer keeping it as-is.

src/closure.rs
daxpedda: Inference is a gray area for breaking changes in Rust. So in theory it seems fine to me.

> Now the only case that wouldn't work the same as before is if someone used `Closure::<dyn FnOnce()>::once(/* a FnOnce with actually multiple params */)` which is still technically a breaking change but I'd say much less likely one and a fair compromise.

AFAIU that seems like a pretty bad breaking change. Can you give me an example here of what you mean?

---

In practice, I'm asking myself why we would be talking such "risky" moves here. AFAICS the motivation of this PR is to simplify the current internal code, with no changes to the public API or enabling any other feature we want to implement. While this is definitely nice and desirable, the trade-off seems "grave" to me.

As it currently stands I'm in favor of moving this to the next breaking change. Unless there is a deeper motivation to get this change in?

RReverser: > AFAIU that seems like a pretty bad breaking change. Can you give me an example here of what you mean?

Hm just what I shown. Someone needs to very explicitly use the `Closure::<dyn FnOnce()>::once(...)` (the inference hack we used where we put `once` and `once_into_js` into separate impl, which is no longer necessary) and they need to use it with a function that is not actually zero-argument one like `Closure::<dyn FnOnce()>::once(|a, b, c| {})`.

That feels like going out of your way to shoot yourself in the foot by mismatching types, and, unlike with explicit generics on `once` itself, I feel like this is a bugfix and should be fine to ship.

`Closure::once(|a, b, c| {})`, `Closure::once(|| {})` etc. all work fine.

> Unless there is a deeper motivation to get this change in?

Yeah it makes further changes much easier.

For example, I want to eventually get rid of the "you can only have closure with a single reference argument, if you have more than that, you're SOL" limitation and it's just not possible with `<A, R>` definition of fnonce but can be possible with one that takes the `dyn FnMut` as well.

RReverser: To clarify, it's not possible because the existing hack uses `WasmClosureFnOnce<&A, R>` (no tuple for first arg) vs `WasmClosureFnOnce<(A,), R>` (single-item tuple for first arg) to make type system distinguish between implementations for reference and non-reference case.

This hack simply doesn't scale to more than one argument, whereas the new definition always works with all tuples and arbitrarily variadic functions in the same way.

guybedford (approved):
If you are confident about this then I'm happy to land this.

But let's let @daxpedda give a final review further.

src/closure.rs
daxpedda: > > AFAIU that seems like a pretty bad breaking change. Can you give me an example here of what you mean?
> 
> Hm just what I shown. Someone needs to very explicitly use the `Closure::<dyn FnOnce()>::once(...)` (the inference hack we used where we put `once` and `once_into_js` into separate impl, which is no longer necessary) and they need to use it with a function that is not actually zero-argument one like `Closure::<dyn FnOnce()>::once(|a, b, c| {})`.

Ah, I see, apologies, kinda missed it in your earlier comment.

---

Seeing that there is a clear motivation to do this, I would say that we shouldn't do this until we get to that motivation first.

I have to say that this is very clearly a breaking change and not a grey area anymore, as it was with inference. For such a foundational crate as `wasm-bindgen` this seems to me like a bad step to take. However, I agree that it is unlikely that anybody misused the API in a way to trigger this breaking change, but it remains a dangerous game we would be playing here.

So I'm in favor to get back to this when we actually can discuss and see what we plan to do with this change, instead of making a breaking change just to improve the internal code.

RReverser: I think that's counter to the desired approach of having more isolated small PRs. If every PR has to have all the future context that it would benefit, we will simply deadlock.

And there are several features that I want to add to closure handling, type casts and even tuples that would greatly benefit from having all variadic handling done consistently (always over a tuple) and declared in one place instead of copy pasting. Surely that doesn't mean we'd want to lump all such features into one huge PR. 

In isolation, this still makes sense as a separate change that consolidated closure handling together, and I'd still argue the described edge case it does break, should be seen as a bugfix for mismatched types than a breaking feature change - the dyn FnOnce() thingy was clearly a hacky workaround than intended API. 

And I'm saying this even after having been on the receiving end of this when we discussed how `link_to!` broke wasm-bindgen-rayon, so I'm not making this claim lightly.

daxpedda: > I think that's counter to the desired approach of having more isolated small PRs. If every PR has to have all the future context that it would benefit, we will simply deadlock.

I hope it is clear that I was not suggesting we treat every PR like we treat PRs that do breaking changes.

> And there are several features that I want to add to closure handling, type casts and even tuples that would greatly benefit from having all variadic handling done consistently (always over a tuple) and declared in one place instead of copy pasting. Surely that doesn't mean we'd want to lump all such features into one huge PR.

Absolutely not. But maybe one of those PR can include these changes. Or if you want to do parallel PRs they can be based on this one. So it doesn't really have to all be in one PR.

> [..] I'd still argue the described edge case it does break, should be seen as a bugfix for mismatched types than a breaking feature change - the dyn FnOnce() thingy was clearly a hacky workaround than intended API.

Agreed. We just have to hope that nobody uses this hacky workaround.

---

When it comes to breaking changes, I would really like to stick to the "rules" here. It is tough for us to judge whats out there relying on weird things, especially considering how broad the usage of this library is.

We do reserve the right to do breaking changes on things that the "rules" allow. E.g. things that aren't API stable (e.g. `doc(hidden)`), we have our own custom MSRV policy, the resolver version, grey areas like type inference or undocumented behavior.

But generics very clearly fall under API stability "rules", hacky workaround or not. If we really classify this as a bug (which I agree it is), according to the "rules", we would have to yank every previous version containing it. I believe following these "rules" is very important if we want to uphold the Rust stability contract we have with our users.

That said: I'm still in favor of doing this breaking change, because I suspect it won't affect anybody. However, this is a risk we take, a risk we can't weigh here if we aren't able to discuss its merits.

We don't have to put everything in a single PR. But before doing a *breaking change*, we must at least be *able* to discuss *why* we are doing it so we can properly justify it.

I hope I'm not taking any controversial stance here.

RReverser: > I hope I'm not taking any controversial stance here.

No, it's a useful discussion to have, I just want to be more pragmatic here - that's why I'm trying to find a compromise and added generics back to `Closure::once` itself (even though I think it's almost equally unlikely anyone uses those explicitly) but still want to remove this one.

Rest of my comment would be duplication of what I said in https://github.com/wasm-bindgen/wasm-bindgen/pull/4588#discussion_r2314210277, so going to just link to it.

TL;DR let's try to make these cleanups. I totally get worry about breaking someone when you're the sole maintainer on a large project, but now that we have healthy number of maintainers, I think we can allow ourselves to risk these edge-cases and revert if we absolutely must. If there is no evidence anyone relies on this unintended hack and common sense also tells us nobody should use it, then I'd rather go ahead and merge cleanups to make our lives easier going forward.

src/convert/closures.rs
RReverser: I moved it to just the reference macro call instead, so we don't need to duplicate it yet it's explicitly allowed only on the one case where it's necessary.

RReverser:
I'm going to go ahead and merge this one with @guybedford's review above. I've addressed primary concerns, and happy to take personal responsibility if the edge case we were bikeshedding over does cause any problems in the wild (even though I'm 99% sure it won't), just want to unblock other work. We have been chatting about edge cases for a while now and I believe the compromise implementation it's at now is safe enough to ship.

PR #4587, opened 2025-08-19
Use test-each for fixture tests

RReverser:
This improves integration with cargo test's filtering, IDE test explorers, and, perhaps most excitingly, tools like [cargo nextest](https://nexte.st/) so we can run tests in parallel which makes them much faster (especially the `reference` tests which need to compile each `.rs` file).

It's also less custom code to maintain.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -40,13 +40,9 @@ wasm-bindgen-shared = { path = "../shared", version = "=0.2.100" }
 assert_cmd = "2"
 diff = "0.1"
 predicates = "3"
-rayon = "1.0"
+test-each = "0.3"
 wasmparser = "0.214"
 wasmprinter = "0.214"
-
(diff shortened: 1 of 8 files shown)


RReverser:
Marking as draft again as want to explore a few more options.

daxpedda (changes requested):
I think this is a great change!

However, even if this is a dev-dependency, we should be careful about adding ones that don't have a lot of eyes on them or are unvetted. So unless you are aware of an alternative, I'm in favor of simply pinning the dependency version. While this is annoying to maintain, its better then a bad surprise later on.

RReverser:
I looked at https://crates.io/crates/rstest while working on this as a potential alternative. It's a bit more verbose and has tons more features than we need, but also much more maintained - do you think that would be a better choice?

daxpedda:
Yes!

PR #4588, opened 2025-08-19
Make debug cfg a non-feature

RReverser:
I've been wondering why, when running e.g. `cargo check --all-features`, I'm getting tons of dumped noise in my terminal, until found this.

This clearly should be an internal cfg like the `wasm_bindgen_unstable_test_coverage` and not a feature that gets opted into automatically.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,10 @@ wasm-bindgen-test-crate-b = { path = 'tests/crates/b' }
 workspace = true
 
 [workspace.lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(wasm_bindgen_unstable_test_coverage)', 'cfg(xxx_debug_only_print_generated_code)'] }
+unexpected_cfgs = { level = "warn", check-cfg = [
+  'cfg(wasm_bindgen_unstable_test_coverage)',
+  'cfg(xxx_debug_only_print_generated_code)',
+] }
 
 [workspace.lints.clippy]
 large_enum_variant = "allow"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,9 @@ strict-macro = ["wasm-bindgen-macro/strict-macro"]
 # are handled correctly
 gg-alloc = ["wasm-bindgen-test/gg-alloc"]
 
+# INTERNAL ONLY and deprecated, to be removed in the next major version.
+xxx_debug_only_print_generated_code = []
+
 [dependencies]
 cfg-if = "1.0.0"
 once_cell = { version = "1.12", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,10 +43,6 @@ strict-macro = ["wasm-bindgen-macro/strict-macro"]
 # are handled correctly
 gg-alloc = ["wasm-bindgen-test/gg-alloc"]
 
-# INTERNAL ONLY: This is only for debugging wasm-bindgen! No stability guarantees, so enable
-# this at your own peril!
-xxx_debug_only_print_generated_code = ["wasm-bindgen-macro/xxx_debug_only_print_generated_code"]
-
 [dependencies]
 cfg-if = "1.0.0"
 once_cell = { version = "1.12", default-features = false }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -43,10 +43,6 @@ strict-macro = ["wasm-bindgen-macro/strict-macro"]
 # are handled correctly
 gg-alloc = ["wasm-bindgen-test/gg-alloc"]
 
-# INTERNAL ONLY: This is only for debugging wasm-bindgen! No stability guarantees, so enable
-# this at your own peril!
-xxx_debug_only_print_generated_code = ["wasm-bindgen-macro/xxx_debug_only_print_generated_code"]
```
daxpedda, line 48: Unfortunately we can't just remove crate features like this, as that would technically be a breaking change.

Instead, we can simply make this crate feature a no-op.

```suggestion
# No-op.
xxx_debug_only_print_generated_code = []
```

daxpedda (changes requested):

Cargo.toml
RReverser: But it literally says internal only. Why would we not want to remove even those? At this rate, we might say the same about doc(hidden) APIs which are also technically breaking changes, but surely we don't want to go there.

daxpedda: Documentation for crate feature is not exposed to the user, see [docs.rs](https://docs.rs/crate/wasm-bindgen/0.2.100/features).

`doc(hidden)` has a long established tradition to mean non-stable API, even if not documented in the official API guidelines. I think the general rule has been that stability follows the output of docs.rs. If something isn't documented in docs.rs, its not stable.

However, I think the name alone implies that this is an internal crate feature only. I seriously doubt anybody is relying on this crate feature also considering there is no stabilized behavior connected to it.

On the other hand it is very cheap to make this crate feature no-op. So I would rather be safe then sorry.

TLDR: I really see no reason to do breaking changes, even if only technical, unless there is an explicit benefit behind it that can justify it.

RReverser: > I really see no reason to do breaking changes

I think the tech debt cost should be taken into account. There is _tons_ of small things in wasm-bindgen that add up to a rather complex project - a usual case of death by thousand cuts, where individually these oddities (whether this PR or the closure PR) are pretty small, but add up to complexity that is not inherent to what we're trying to provide.

I think we need to clean them up as much as possible to make it easier to work on actual features, rather than leave them out of fear of breaking someone even when it's very, very unlikely.

We can always make a patch release if it turns out that the extremely unlikely case does break someone, but I'd rather still try and make these cleanups first.

daxpedda: I agree on all this.

But I think a no-op crate feature is a very small tech debt compared to the huge stuff we already have?
I also agree with the "death by thousand cuts" argument, but seriously: we are talking about a couple of lines of TOML code here. No inherent complexity here.

In my opinion, breaking changes is a very clear line we should only cross if absolutely necessary. I'm surprised this comes up so often.

RReverser: > In my opinion, breaking changes is a very clear line we should only cross if absolutely necessary. I'm surprised this comes up so often.

Having been involved on-and-off with wasm-bindgen for years, I just keep coming back to feeling that fear of breaking changes is exactly what led to its stagnation (just look at the number of issues labeled with "breaking-change" that would be very concrete improvements but were never implemented precisely for fear of breaking existing ecosystem).

While I agree we need to be careful with them, and that most of those will have to go into a new release, I also want to make sure that we break (no pun intended) that habit going forward, otherwise we won't be able to meaningfully improve DX and the project will remain in maintenance mode forever.

Nothing wrong with that mode for finished libraries, but I think we can all agree wasm-bindgen is anything but, and there is number of very real improvements that can be made over time and that will require breaking changes on a much more regular basis than it had in the past.

daxpedda: I agree that would be nice, but not like this. The way this is solved is by releasing new versions, not by doing breaking changes.

The current problem is that releasing a new version of `wasm-bindgen` comes with such huge costs that it isn't worthwhile. This is the problem we ought to solve. Not subvert the Rust ecosystem SemVer expectations.

RReverser: I put back the feature + emitting deprecation warning if anyone uses it. I'm still pretty unhappy with this approach to purely technical breaking changes - I believe we have a lot more wiggle room here - but in this particular PR it's not a hill I'm going to die for, as it's indeed just a TOML line and not blocking anything else.

PR #4589, opened 2025-08-20
Add GestureEvent (a WebKit proprietary interface for multitouch)

thatcomputerguy0101:
GestureEvent is a WebKit proprietary interface for receiving multitouch pinch and rotation gestures. The information for the WebIDL was gathered from https://developer.apple.com/documentation/webkitjs/gestureevent.

I assumed this needed to be in `unstable` and not `enabled` due to being a non-standard API, but that does limit its usability.


crates/web-sys/webidls/unstable/GestureEvent.webidl
daxpedda, line 1: Where did you actually get the WebIDL? I couldn't find it in WebKits source code.

daxpedda (changes requested):
LGTM, just missing a changelog entry.

crates/web-sys/webidls/unstable/GestureEvent.webidl
thatcomputerguy0101: https://developer.apple.com/documentation/webkitjs/gestureevent seemed to be generated from the WebIDL (which I assume Apple has in a private repo) so I copied all the fields described in that and assembled them into a single file with some minor edits to match similar WebIDLs.

daxpedda: Really appreciate the effort to align with the existing IDL!

Unfortunately we are trying to move on from this older IDL format and vision. In the past the IDL was just copied from Mozilla or other browser source code, nowadays we want to stick as close as possible to the specification. This means we don't want to add browser-specific attributes and also use the newer WebIDL specification (e.g. constructors).

I went ahead and adjusted it.

thatcomputerguy0101:
I thought the `[BinaryName="clientX"]` line allowed the JS name and the Rust name to be separate. I didn't realize it was just a browser-specific attribute. Those two fields should be named `clientX` and `clientY` instead of `x` and `y`, which I have changed on the branch this PR was created from. Should I create a new PR?

daxpedda:
> I thought the `[BinaryName="clientX"]` line allowed the JS name and the Rust name to be separate.

Why should they be separate in this case? Shouldn't they always be named `clientX/clientY`?

> Should I create a new PR?

Yes please!

I also noticed that we got the constructor wrong. This type doesn't seem to have a constructor, but `initGestureEvent` is just a method,  similar to [`initMouseEvent`](https://github.com/wasm-bindgen/wasm-bindgen/blob/8efddf0f819ab1cb030578377336e5a4fb052581/crates/web-sys/webidls/enabled/MouseEvent.webidl#L41-L58).

thatcomputerguy0101:
> Why should they be separate in this case? Shouldn't they always be named clientX/clientY?

They shouldn't be separate. I had inferred conventions that don't exist.

> I also noticed that we got the constructor wrong. This type doesn't seem to have a constructor, but initGestureEvent is just a method, similar to [initMouseEvent](https://github.com/wasm-bindgen/wasm-bindgen/blob/8efddf0f819ab1cb030578377336e5a4fb052581/crates/web-sys/webidls/enabled/MouseEvent.webidl#L41-L58).

I will also fix that in the next PR.

PR #4590, opened 2025-08-20
feat: Generics support for Imports

guybedford:
This adds generics support for imported JS functions in wasm-bindgen.

This PR explcitly then adds typing to `Promise<T>` and `Array<T>`, operating on `JsValue<T>`, in a way that is backwards compatible on the default `JsValue` generic.

* The macro parser is updated to support generics on imported wasm bindgen functions and types, supporting generics on function parameters and return types. The parser carefully validates that lifetimes are not used, other generic features like const are not used so we just have very limited simple generics with bounds, and these cases are tested in the macro tests. We also validate that method instrinsics require `&self<T...>` for all non-optional generics.
* The validated generics structure is stored using `syn::Generics` on the function and imported type, and then retained in the final generated function.
* To support extern "C" which does not support generics we make all generic types concrete on the ABI boundary by using JsValue itself as the concrete "top type" for all generics. That is generics are constrained to be commutative with JsValue through the introduction of JsValue casting.
* Code generation then applies these casts, carefully merging and applying generic bounds, and carefully separating `impl` bounds in the cases of static and non-static methods.

## JsValue Casts

The main casting cases to deal with for generating concrete ABI types from generics are in an imported function include the following. In all of these cases we need to obtain either `JsValue` or `&JsValue`:

* `T`, `Foo<T>` value parameters into `JsValue` via `Into<JsValue>` for simplicity
* `T`, `Foo<T>` value return into `JsValue` cast into the exact type via the new `wbg_cast`

Types that cannot be cast into JsValue cannot be supported in generics. Slices cannot be made generic, but `Array` can.

## Value types

For value typing we lean into the new typed `JsValue<T>` provided in https://github.com/wasm-bindgen/wasm-bindgen/pull/4746, which allow providing compile-time type hints that are runtime-validated through the generic type system.

### wbg_cast

Replacing the original `JsValueCast`, `wbg_cast` is used to cast return values from `JsValue` into the exact type, while following the same semantics of the static bindgen, thanks to the work here by @RReverser. The `From` type is always JsValue, while the `To` type is the type being cast into.

## Generic Types

With the codegen and casts implemented, we are then able to add generics support for the following here. More APIs can be covered in future, but I didn't want to go too far down this road here.

Some functions have generic types that can't be inferred, and in these cases they are left untyped. When a typed variant of the same function is needed I started a couple of cases like `Array::new()` and `Promise::new()` where `Array::new_t::<T>()` can be used with a `_t` suffix to access the typed variant. It seems like this could be a relatively sensible convention to me, but further feedback is welcome.

### Promise

All promise methods are now typed, with tests in `typed_promise.rs` values carry through full typing information from a `Promise.resolve()` through to its `future` and through to its then.

A new `then_map` method is added which supports a return value for then also being typed as two separate generic parameters for the promise then.

Because of the default `JsValue` type, these changes remain backwards compatible.

### Closures

Getting promises to work required solving closure generics where all transitive generic traits as required are now implemented for closures.

### Futures

When resolving promises into futures, futures are now all typed by default, and this is backwards compatible by having `JsValue` as the default generic parameter.

`future_to_promise` was not possible to implement typed without that being a breaking change, so instead a new `future_to_promise_typed` is introduced with the generic parameter.

### Array

Array and Array methods are typed where possible. Unfortunately `FnMut` is not possible to type without implementing that as a cast into an externref which changes the lifetime semantics, so generic function callbacks are not yet supported. I'd like to rather implement this as a follow-on.

### Option

When supporting generics on `Option<T>` we need to be able to obtain an `Option<T>` from a JsValue. This is done by treating null or undefined as the None case, and now implemented as a new feature.

When casting `Some<T>` there is of course intersection with `null` and `undefined`, where it might make sense to have a marker for non null JS Values further, I did explore this briefly but didn't want to go too far down this road.

At the end of the day, many functions will want generics on `pub fn foo(&Option<T>) -> Option<T>`, so supporting the generic conversion through JsValue casting for option is key to this.

### Type Errors

By using `JsValue<T>` type error handling can be deferred to the end user in whether they choose to unwrap via `.unwrap()` or `.try_unwrap()` to obtain the Rust type from the JS JsValue type.


crates/macro-support/src/codegen.rs
```diff
@@ -984,34 +982,74 @@ impl ToTokens for ast::ImportType {
             }
         };
 
+        let generic_defs;
+        let generics;
+        let phantom;
+        let phantom_init;
+        let default_as_ref;
+
+        let where_clause = generics::where_clause(&self.generics);
```
RReverser: These custom utils seem to duplicate what [`syn::Generics::split_for_impl`](https://docs.rs/syn/latest/syn/struct.Generics.html#method.split_for_impl) is doing.

You might want to use it instead for simplicity.

guybedford:
I've rebased this branch to the latest and refactored `JsValueCast` to instead use the new `wbg_cast`.

All seems to mostly work, except there is a new bug with stack handling which I'll investigate further in another week or so. If anyone wants to take a look at the bug in the meantime please feel free.

crates/macro-support/src/codegen.rs
guybedford: Thanks, will look into this.

src/cast.rs
```diff
@@ -154,3 +155,36 @@ where
...
+/// constant reference slots for JS values available only at the generic bindgen
+/// call interface, by referencing constant slots in __rt.
+///
+/// By setting the `USES_VALUE_SLOTS` constant, the bindgen is then able to inform
+/// the bindgen process that it is using these externref slots and that they should
+/// be cleared after the function call, guaranteeing the JsValue indices are
+/// live for as long as the function call only.
+///
+/// This trait is automatically implemented for any type supporting `AsRef<JsValue>`.
+pub trait JsUpcastRef<T = JsValue> {
```
RReverser: I remember you initially added this machinery for references when you thought we can't accept both references and plain values in the same signature.

Now that we figured it out, can we leave this trait and slots implementation out of this PR? It feels like a large chunk of complexity here comes just from this reference upcasting, even though I imagine in real world scenarios it's going to be relatively uncommon for someone to want to pass `&i32` instead of `i32` and so on, and removing it from this PR would reduce scope to just generics for types we support today.

And if it's not uncommon, we can always add it separately, as being able to pass references to more types is not strictly tied to generics support.

guybedford: I did investigate this further, but it turns out we do need the generic on `&T`. The reason for this is that backwards compatibility is a requirement for this API and `Promise.resolve(&T)` is the current API today. Therefore solving supporting a wide range of `&T` inputs is an important use case goal via `JsUpcastRef` (see the test cases for more details).

If we were designing from scratch today, we could just use `Promise.resolve(T)` with variation support, but even with a `Borrow<T>` trait there are backwards compat issues. I think being able to ship generics without requiring new layering and while retaining backwards compat is a really powerful approach to this. If we wanted to let backwards compat slide, that could give us some more room, but it is also simply not an option for us with workers-rs unfortunately.

RReverser: > and `Promise.resolve(&T)` is the current API today

I think we talked about this too. Isn't it only `Promise::resolve(&JsValue)` that is current API today?

Since both `&JsValue` and `JsValue` already implement `IntoWasmAbi`, it should be enough for you to use any `fn resolve<T: IntoWasmAbi>(value: T)` in the generic API to cover both existing scenarios with full backward compatibility as well as new ones.

Such generic constraint would accept `Promise::resolve(jsval)`, `Promise::resolve(&jsval)`, `Promise::resolve(42_i32)`, `Promise::resolve("foo")` and so on automatically - no need to single-out references just for generics.

guybedford: While that is possible, the problem is the requirement for type specialization on the generic - where a singular specialization is possible on `&T` as `&JsValue` requiring no further type hints making it backwards compatible.

guybedford: In case that wasn't clear - if `T` can be either `&JsValue` or `JsValue` then more type information is required from the usage context, and thus this requirement on the type is a breaking change for existing usage.

RReverser: I don't understand, can you show an example that would work today but break with such generic?

guybedford: Actually I think the problem is simpler:

```rust
extern "C" { 
    #[wasm_bindgen(static_method_of = Promise)]
    pub fn resolve(obj: &JsValue) -> Promise;
}
```

Can only be represented by the generic:

```rust
extern "C" { 
    #[wasm_bindgen(static_method_of = Promise)]
    pub fn resolve<T>(obj: &T) -> Promise<T>;
}
```

It cannot be represented by the generic:

```rust
extern "C" { 
    #[wasm_bindgen(static_method_of = Promise)]
    pub fn resolve<T>(obj: T) -> Promise<T>;
}
```

as when we generate the specialization by substituting JsValue that would be a breaking change.

RReverser: Ah wait I understand what you mean, thanks for the example.

Hmm. Need to think about this.

I'd really like to avoid having to use references for primitives. Maybe there is some solution with custom trait that would have T::Output that would unwrap reference automatically...

Let's discuss more next week.

guybedford: This doesn't create a requirement for using references for primitives - most APIs should use value types for primitives, and also support `&T` or `T` interchangeably. But we need to special case `&T` for this backwards compatibility.

Ensuring `T` supports a ref case could be something we could add for new APIs that want to support own or ref, but that would require a separate design. If that's the case you're interested in happy to discuss it further.

RReverser: > Ensuring `T` supports a ref case could be something we could add for new APIs that want to support own or ref, but that would require a separate design. If that's the case you're interested in happy to discuss it further.

No, no, not at all. What I'm thinking of is ways to special case just `&JsValue` so that we don't need `JsUpcastRef` for all these other types just to accommodate one pre-existing special case.

guybedford: It's worth noting this is the majority of js-sys and web-sys usage, not just one special case.

crates/macro-support/src/codegen.rs
guybedford: I've added using `split_for_impl` here instead.

guybedford:
This PR has now been rebased to https://github.com/wasm-bindgen/wasm-bindgen/pull/4746, removing the need for `JsUpcastRef` in turn.

I'll be refining the approach here next week, to get it ready to land while the pre-requisite PRs get reviewed.

src/cast.rs
guybedford: `JsUpcastRef` has now been removed and replaced with `&JsValue<T>`.

guybedford:
This generics implementation was based on a "double casting" for T types on arguments and return values which incurs a performance cost, effectively running a first call out to JS to do the cast, and then a second call to do the generic function call.

The mitigation for primtive type generics is NOT to support `T = i32` for primitives in generics, but instead use the JsRef implementation separately for primitives - when `T = i32` is needed as argument, instead use `T = JsRef<i32>`. This is treated as a separate feature. Generics on non-primitive types then become the goal for the main generics implementation, avoiding the pitfall of this PR that got too deep into solving base-level values.

The new generics PR in https://github.com/wasm-bindgen/wasm-bindgen/pull/4756 takes the more general approach, by not solving for primitives. This provides a better abstraction to then also handle future additions to the generics system.

PR #4594, opened 2025-08-23
Add error message when export symbol is not found

Spxg:
close https://github.com/wasm-bindgen/wasm-bindgen/issues/4390


daxpedda (changes requested):
And thank you for putting together https://github.com/Spxg/wasm-ld-symbol-variant!

The error message isn't very helpful to regular users, but I can't think of what else we can do here.

Would you mind adding a changelog entry? Otherwise LGTM!

PR #4602, opened 2025-09-01
Implement arbitrary generic casts

RReverser:
Persistent closure handling in wasm-bindgen transforms was mixing two different features:

1. Describing closure type itself.
2. Generating code to convert arbitrary closure with arbitrary Rust generics into a `JsValue` for `Closure::into_js_value`.

While first one is rather mundane, the second is actually a very powerful feature, because if you squint at it just right, it's a way to convert any Rust type in a generic context into a `JsValue` - or the other way around - which is exactly what we often want to support usecases like https://github.com/wasm-bindgen/wasm-bindgen/pull/4590, https://github.com/wasm-bindgen/wasm-bindgen/pull/4579 and others.

In this PR I'm splitting these two features up, and reusing the existing mechanism to support arbitrary casts, not just `Closure<...generics> -> JsValue`.

Aside from the new feature, this makes closure handling a lot less magical and removes a bunch of code that is no longer necessary.

I recommend reviewing this commit by commit, I tried to ensure each change make sense as a separate step.


src/lib.rs
```diff
@@ -1143,7 +1144,7 @@ externs! {
         fn __wbindgen_cb_drop(idx: u32) -> u32;
 
         fn __wbindgen_describe(v: u32) -> ();
-        fn __wbindgen_describe_closure(a: u32, b: u32, c: u32) -> u32;
+        fn __wbindgen_describe_cast() -> !;
```
daxpedda: Not sure I understood why this should return `!`.

daxpedda (commented):
Just took a quick glance, but it makes a lot of sense to me. Great work!

src/lib.rs
RReverser: Because we want to support functions like

```rust
fn wbg_cast_breaks_if_inlined<A, B>(a: A) -> B {
  // must return something at compile time, doesn't matter what, as we're going to run transform on this function later anyway
}
```

For it to compile before we do the transform, we need it to return some dummy value. One option could be to construct uninitialized value, but that's just UB.

Another option is to use the never type which in Rust is automatically compatible with any other type, because it's guaranteed to never return (e.g. in Wasm it compiles to the function call followed by `unreachable` instruction), so the type system happily accepts it in any context and it doesn't cause UB. That's the option I went for.

daxpedda: Yeah that makes a lot of sense.
Can we add some documentation around that or did I miss it somewhere else?

RReverser: Yeah I have it in the transform itself: https://github.com/wasm-bindgen/wasm-bindgen/pull/4602/files#diff-6da07b04827cd5dcb413608a7b2b1e5dca91c3d01a148cd03c110a0355cb4885R336-R339

RReverser:
Looks like I broke optional closures at some point in this work, adding them back now.

crates/cli-support/src/wit/mod.rs
```diff
@@ -244,6 +235,22 @@ impl<'a> Context<'a> {
...
+                }
+                ElementItems::Expressions(_, exprs) => {
+                    for expr in exprs {
+                        if let ConstExpr::RefFunc(func) = expr {
+                            replace.visit_function_id_mut(func);
+                        }
+                    }
+                }
+            }
+        }
```
daxpedda, line 253: I didn't get why this part is necessary now.

daxpedda: Sorry, got it now :smile:

src/rt/mod.rs
```diff
@@ -20,28 +22,75 @@ pub mod marker;
...
+        prim4: <From::Abi as WasmAbi>::Prim4,
+    ) -> WasmRet<To::Abi> {
+        inform(FUNCTION);
+        inform(0);
+        inform(1);
+        From::describe();
+        To::describe();
+        To::describe();
+        // Pass all inputs and outputs across the opaque FFI boundary to prevent
+        // LLVM from removing them as dead code.
```
daxpedda:
```suggestion
        // compiler from removing them as dead code.
```
Who knows what compiler backends `wasm-bindgen` has to account for in the future 😆

RReverser:
> I remembered again how pleasantly surprised I was about this simplification the first time I reviewed it.

Heh I was worried the change from `fn __wbindgen_describe() -> !`(which felt rather elegant when I wrote in in the first iteratio of the PR) to the pointer mumbo-jumbo to stop optimiser from optimising would take away from that feeling 😅

crates/cli-support/src/interpreter/mod.rs
```diff
@@ -388,18 +323,19 @@ impl Frame<'_> {
...
-                    self.interp.descriptor_table_idx = Some(val as u32);
-                    stack.push(0)
+                // If this function is calling the `__wbindgen_describe_cast`
+                // function then it's just a marker for the parent function
+                // to be treated as a cast.
+                } else if Some(func) == self.interp.describe_cast_id {
+                    log::trace!("__wbindgen_describe_cast()");
+                    // The parent (cast wrapper) can return arbitrary type.
+                    // To be compatible with anything, we declare return type of
+                    // `__wbindgen_describe_cast` as `!` (never type).
```
RReverser: Note to self - update this outdated comment before merging.

PR #4608, opened 2025-09-03
Merge internal crates into `wasm-bindgen-cli`

RReverser:
Currently we have way too complex crate graph in this repo, with many of the crates intended purely for internal consumption. This unnecessarily complicates publishing, reasoning about public APIs, version dependencies, testing and general maintenance.

This PR merges crates which are purely CLI dependencies into the `wasm-bindgen-cli` crate itself. This includes:

- `wasm-bindgen-*-xform`: Purely internal crates for individual transformation steps. No reason for those to be separate crates:
  - `wasm-bindgen-externref-xform`
  - `wasm-bindgen-multi-value-xform`
  - `wasm-bindgen-threads-xform`

- `wasm-bindgen-wasm-conventions`: Similarly, an internal crate for parsing things that transforms above and the CLI care about. No reason for this to be a separate crate.

- `wasm-bindgen-wasm-interpreter`: An internal crate for interpreting the descriptors, infamously strongly tied to the rest of wasm-bindgen. No reason for this to be a separate crate.

- `wasm-bindgen-cli-support`: This one is likely controversial because it actually has a few external users.
  
  I believe we should leave the bindgen API public, but there is no reason to have this as a separate crate. Instead, `wasm-bindgen-cli` now has both the library and binary targets, and can be used either way.

  It's not a semver-breaking change to stop publishing new versions, but Cargo doesn't have a way to mark packages as deprecated, so it might be also not a very discoverable change. To make it more visible, I think we might want to publish `wasm-bindgen-cli-support` one last time with a shim that re-exports everything from `wasm-bindgen-cli` with a deprecation attribute. This way downstream users will continue receiving the latest version, but will get a warning that they should switch to `wasm-bindgen-cli`.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -249,7 +249,7 @@ impl<'a> Context<'a> {
         // up we always remove the `start` function if one is present. The JS
         // bindings glue then manually calls the start function (if it was
         // previously present).
-        let needs_manual_start = unstart_start_function(&mut self.module);
+        let needs_manual_start = unstart_start_function(self.module);
 
         // Cause any future calls to `should_write_global` to panic, making sure
         // we don't ask for items which we can no longer emit.


RReverser:
(Note to self: ~~update changelog before merging, and~~ make sure we follow up on the deprecation suggestion if agreed on it.)

guybedford:
Please also update the `justfile` as well.

RReverser:
> Please also update the `justfile` as well.

From a cursory look, it doesn't need any changes?

guybedford:
Ahh, right we still have macro-support of course.

daxpedda (changes requested):
I think this is a great change!

However, I would like to keep the `wasm-bindgen-cli-support` crate seperate to remove any unrelated dependencies, like `clap` and all the stuff related to `wasm-bindgen-test`.

So instead I would suggest merging everything into `wasm-bindgen-cli-support`.

Alternatively, we could put all this stuff behind crate features, but I don't think this works out properly the way cargo binaries are distributed and such. But I don't know for sure.

RReverser:
@daxpedda Yeah that's a valid concern.

> Alternatively, we could put all this stuff behind crate features, but I don't think this works out properly the way cargo binaries are distributed and such. But I don't know for sure.

I like this idea more. Using `wasm-bindgen-cli-support` as a separate library is kinda an edge case, so I think it's fair not to optimise for it DX-wise.

Let's just add `binary` feature to `wasm-bindgen-cli` that is enabled by default, and downstream users can opt-out via `default-features = false` to exclude those deps.

It really would be nice if Cargo allowed per-target dependencies, but that's been a long time coming and clearly not there yet...

daxpedda:
> Let's just add `binary` feature to `wasm-bindgen-cli` that is enabled by default, and downstream users can opt-out via `default-features = false` to exclude those deps.

While this could be a solution, I think in comparison its worse. Having one extra crate for the binary seems like a worthwhile cost to me compared to requiring special attention from users to not miss something.

Is there some other reason for getting rid of `wasm-bindgen-cli-support`?

RReverser:
> Is there some other reason for getting rid of `wasm-bindgen-cli-support`?

No, like everything else in this PR it's about trying to simplify crate grapht as much as possible.

> Having one extra crate for the binary seems like a worthwhile cost to me compared to requiring special attention from users to not miss something.

It's not a big deal though. As in, we're not talking about a potential breakage here - if they leave default features in, their compilation will still succeed and the resulting library will be exactly the same, the only issue is extra build time for dependencies which are eventually unused.

But same is the case for any default Cargo features in any crate - it's pretty common for users to just leave whatever the default is, even if they end up not using certain APIs that _could_ be excluded as a build time optimisation.

RReverser:
I did the above for now, PTAL. By default it's just 4 extra deps (plus whoever depends on cli-support is likely building their own CLI, and equally likely will use clap anyway). I don't think it's a huge deal to justify the complexity of separate crate.

daxpedda:
I think the trade-off we are talking about here is reducing the number of crates by one, which is definitely desirable, but its maintenance cost is limited by release automation. Versus relying on users to notice and use `default-features = false`, which otherwise doesn't break anything but the dependencies we are talking about are pretty heavy.

IMHO, it seems pretty clear to me that maintaining one extra crate to improve UX is the better choice.

RReverser:
Idk, I'm of two minds about this discussion. I can see your point, but OTOH it seems the downside to merging is so low it's not worth the extra crate.

Like, yes, it's just one crate, but the earlier mentioned death by thousand cuts is everywhere in this project, so I want us to have a justification for every such "just one" that is much stronger than "nice to have".

@guybedford maybe you can chime in with your thoughts since workers-rs is a popular user of said cli-support library, so you might have a clearer opinion as an end user than either of us?

daxpedda:
> Like, yes, it's just one crate, but the earlier mentioned death by thousand cuts is everywhere in this project, so I want us to have a justification for every such "just one" that is much stronger than "nice to have".

But both are cuts. Having an extra crate to improve UX seems like a very valid workaround to me.
The `default-features = false` requirement here seems like the bigger cut to me and is also a workaround.

> guybedford maybe you can chime in with your thoughts since workers-rs is a popular user of said cli-support library, so you might have a clearer opinion as an end user than either of us?

While I don't think breaking `cli-support` users is a good argument here, its definitely a good *excuse* to keep the current setup. Wasm runners have become quite popular in the ecosystem since `wasm-pack` was not so well maintained. So this will definitely have a sizeable impact.

RReverser:
> Having an extra crate to improve UX seems like a very valid workaround to me.
> The `default-features = false` requirement here seems like the bigger cut to me and is also a workaround.

That's not UX though, that's merely build time. UX aka API usability is the same.

As I said above, if you want to opt out from extra deps to save build time, you have to do that in any crate in Cargo ecosystem, it's a pretty well-established workflow. So what I don't understand is why our case is special.

> While I don't think breaking `cli-support` users is a good argument here, its definitely a good _excuse_ to keep the current setup.

Actually... combined with my earlier suggestion above:

> * To make it more visible, I think we might want to publish `wasm-bindgen-cli-support` one last time with a shim that re-exports everything from `wasm-bindgen-cli` with a deprecation attribute. This way downstream users will continue receiving the latest version, but will get a warning that they should switch to `wasm-bindgen-cli`.

This last published `wasm-bindgen-cli-support` just be a wrapper for `wasm-bindgen-cli = { version = "0.2", default-features = false }`.

This way all downstream users who already depend on it, will continue working exactly as they did before, with no new deps, yet we don't have to keep it in git tree.

Spxg:
> `wasm-bindgen-cli-support`: This one is likely controversial because it actually has a few external users.

I'm actually maintaining a large Rust project that supports multiple platforms, including wasm. We maintain a common CLI tool for packaging products, which has no other CLI binary dependencies, `wasm-bindgen-cli-support` is also being used.

I think it would be reasonable for `wasm-bindgen-cli` to provide a `cli-support` crate, just like [nextest-runner](https://crates.io/crates/nextest-runner).

RReverser:
> I think it would be reasonable for `wasm-bindgen-cli` to provide a `cli-support` crate, just like [nextest-runner](https://crates.io/crates/nextest-runner).

Oh yeah, to be clear, I'm not suggesting to remove the library.

What we are discussing is just whether `wasm-bindgen-cli-support = "0.2"` is really that much easier to write in Cargo.toml than `wasm-bindgen-cli = { version = "0.2", default-features = false }` for end consumers like you.

guybedford:
One extra crate, at least for now seems fine to me. This is already a huge reduction.

We can always circle back to the CLI / API separation question as well if usage patterns change in future.

RReverser:
https://github.com/rust-lang/cargo/issues/4663 can't come soon enough... 8 years and counting 😭

It would make this a breeze.

Ok I'll revert the cli-support for now as per informal vote above.

Spxg:
> default-features = false

I think using `default-features = false` to indicate a library is a bit odd, even though it seems to be just to avoid importing other crates?

As you said, if https://github.com/rust-lang/cargo/issues/4663 is resolved, it will be much simpler.

daxpedda:
I would really like to shift the perspective here.
This is all coming from the idea that removing an extra crate is a huge maintenance cost reduction.

The argument of a thousand cuts is a very valid one, but it really shouldn't be applied here. This isn't some maintenance burden that doesn't fulfill a purpose other then preventing an extremely unlikely breaking change. This is a maintenance burden with a very clear purpose and goal.

If we just apply this argument to anything that has a maintenance cost, then any piece of code in this repo falls under the same umbrella.

Lets also talk about the cost in practice as well. The cost of maintaining one extra crate is not exactly tangible. Its extra work for the CI, its extra work for the publishing process. But the amount of extra work we are talking about here is a couple of lines of code here and there because luckily its all automated.

The usual cost of maintaining multiple crates though is having to deal with the API interface now that you don't just have full access to everything. But considering we have to do this anyway for public consumption, I think that having to deal with an extra crate feature is actually more work in practice then dealing with an extra crate that is just added as a dependency.

Lets also touch on what we gain here. Letting users deal with `default-features = false` is not really a big deal because, as we established, it doesn't break anything. Even though we would be shifting this maintenance burden onto our users, its probably fair to say the cost is close to zero. But if forgotten by `wasm-bindgen` runner implementers, the build time cost to every user installing and updating it is significant. This is why going with `default-features = false` is, simply put, a footgun. It definitely has worth avoiding it. On that note: to me this definitely falls under UX; user experience.

---

Reducing the maintenance cost in `wasm-bindgen` should really be one of our top priorities. I believe it is the reason why maintaining this project in the past basically failed. But it isn't our top priority. Our top priority is serving our users.

This doesn't justify *any* maintenance cost. Some things aren't worth taking on or don't fit our bandwidth. But this is how we should frame things.

IMHO, in this case, the maintenance cost of maintaining one extra crate is definitely worth it.

RReverser:
> If we just apply this argument to anything that has a maintenance cost, then any piece of code in this repo falls under the same umbrella.

I know you say that in jest, but... yes. Perhaps that's the web part of my background speaking - where any API can never ever be removed without breaking someone - but any public API is a maintenance liability, and needs to have a very high bar before being exposed.

Given that none of these crates were supposed to be depended upon by external consumers, ideally they shouldn't have been published in the first place because as a result at this point almost any subtle change is a breaking change. But of course we're a few years too late for that now, and there are tons of users depending on every possible aspect of every published API.

All we can do now is try and reduce the public surface as much as possible in the future versions to make life easier both for us and for whoever ends up maintaining this project in another 5-10 years.

I'm obviously not saying "remove all the code - no code, no problem", but I do think that there should be a clear purpose of the project. That might be "a CLI to generate bindings for x y z targets", might be some APIs and tools beyond that, but we should clearly flesh those out and write them somewhere. Then, every exposed piece of code in future versions should clearly serve that purpose.

Publishing additional crates or exposing additional APIs just because they might be nice-to-have for downstream users, rather than because they serve the project itself, is simply not sustainable and is an unjustified maintenance burden by definition.

RReverser:
This wider discussion really should be done outside this PR though, especially since I already reverted the change as per comments above.

Let's leave it for maintainer meetings.

RReverser:
I don't yet understand why the thread transform tests are failing now...

Printing the generated module, the start section does look corrupted (points to the wrong function), but not sure why as the logic hasn't changed...

Maybe some dependency shenanigans triggering a bug that only occurs in a specific version of a specific dependency (walrus / wasmparser / etc)?

daxpedda:
I didn't know that making this public API was even the discussion here, because the initial proposal was still making it public, just in another crate. That's a totally different topic.

Just my 2¢ on this: we do want to make this API public. External projects making use of it, like `wasm-bindgen` runners are a very important part of the ecosystem we do want to support. But because its still marked as an internal crate its API isn't stable. This is something we should change in the future.

But yes, the currently public API of `wasm-bindgen-cli-support` is a mess and needs some significant clean up before stabilizing it. Also, just for historical context, we have broken the API of `wasm-bindgen-cli-support` a lot of times in the past.

In addition, breaking changes to `wasm-bindgen-cli-support` aren't a problem, we can always just increase the version to make those changes. This isn't free, because all users have to upgrade, but this is a top-level dependency. Unlike `wasm-bindgen` itself, where a version change would break the whole Rust ecosystem for a significant amount of time.

README.md
```diff
@@ -112,7 +112,7 @@ You can find general documentation about using Rust and WebAssembly together
 
 Libraries that are released on [crates.io](https://crates.io) have a MSRV of v1.57. Changes to the MSRV will be accompanied by a minor version bump.
 
-CLI tools and their corresponding support libraries have a MSRV of v1.76. Changes to the MSRV will be accompanied by a patch version bump.
+CLI tools and their corresponding support libraries have a MSRV of v1.82. Changes to the MSRV will be accompanied by a patch version bump.
```
daxpedda, line 115: We should mention that MSRV increase in the changelog as well.

daxpedda (changes requested):
Otherwise LGTM.

README.md
RReverser: Btw, I wonder if we should just commit `Cargo.lock`. It's not recommended for libraries, since end user will get their own version anyway, but since we have CLIs mixed in the same workspace, I think it's pretty necessary.

The only reason for this MSRV bump here is that crate reshuffling causes Cargo to recalculate latest versions of dependencies differently and pick up newer versions of deps than before.

But that also means we're going to have MSRV updated all the time, whenever some leaf dependency that doesn't have pin own MSRV, publishes a new release that depends on new Rust features.

Hence, I see only two options to prevent that: 1) declare that we don't have MSRV for CLI tools, which is probably fine since most users would download prebuilt libraries anyway or 2) commit Cargo.lock and carefully update CLI deps only when we can verify it's still compatible with our current MSRV.

daxpedda: > Btw, I wonder if we should just commit `Cargo.lock`. It's not recommended for libraries, since end user will get their own version anyway, but since we have CLIs mixed in the same workspace, I think it's pretty necessary.

I agree that for the CLIs we actually should commit `Cargo.lock`. However, it is pretty annoying to maintain it over time. The only solution I see here is to move the CLIs out of the workspace to have their own `Cargo.lock` (which would still be annoying to maintain, but we can just ignore it until before the release). But having multiple workspaces is also pretty annoying to work with.

Really not sure where I stand on this.

> But that also means we're going to have MSRV updated all the time, whenever some leaf dependency that doesn't have pin own MSRV, publishes a new release that depends on new Rust features.
>
> Hence, I see only two options to prevent that: 1) declare that we don't have MSRV for CLI tools, which is probably fine since most users would download prebuilt libraries anyway or 2) commit Cargo.lock and carefully update CLI deps only when we can verify it's still compatible with our current MSRV.

It is pretty established in the Rust ecosystem at this point that dealing with MSRV issues because of dependencies is a user issue, not an author one. Not because that's a desirable system, but because very few libraries actually have an MSRV policy. Realistically, the only choice crates had was to pin every dependency, which isn't really sustainable. So I think we can stick with what we currently have, despite MSRV issues.

Fortunately, this whole problem has been resolved since Cargo v1.84 with the new MSRV-aware dependency resolver.

However, I don't know how exactly this works with `cargo install`. We can't use the new v3 resolver because of our workspace, but we could use [`resolver.incompatible-rust-versions`](https://doc.rust-lang.org/1.89.0/cargo/reference/config.html#resolverincompatible-rust-versions) specifically for the CLIs. Not sure if `cargo install` respects that.

Spxg:
> I don't yet understand why the thread transform tests are failing now..

Because CI did not run the `wasm-bindgen-threads-xform` test before
https://github.com/wasm-bindgen/wasm-bindgen/blob/8efddf0f819ab1cb030578377336e5a4fb052581/.github/workflows/main.yml#L251

The reason this test fails is that `thread-xform` modifies the start function and adds an i32 parameter (It seems to be used to set the stack size), causing the wasmparser check to fail.

https://github.com/wasm-bindgen/wasm-bindgen/blob/8efddf0f819ab1cb030578377336e5a4fb052581/crates/threads-xform/src/lib.rs#L278

https://github.com/bytecodealliance/wasm-tools/blob/544ac54d0d9d43bcfed2597af4293729dfb89854/crates/wasmparser/src/validator.rs#L919

This non-standard `start` will be removed after `wasm-bindgen` processing and renamed to `__wbindgen_start`

https://github.com/wasm-bindgen/wasm-bindgen/blob/8efddf0f819ab1cb030578377336e5a4fb052581/crates/cli-support/src/js/mod.rs#L2692

RReverser:
> Because CI did not run the `wasm-bindgen-threads-xform` test before

Lol, great... At least that explains it. Thanks for catching this.

@daxpedda just another motivation to simplify the crate graph as much as possible. It might look cheap maintenance-wise if we just assume it already works as intended, but there is an inherent mental cost to any abstractions. Things easily slip by.

README.md
RReverser: > It is pretty established in the Rust ecosystem at this point that dealing with MSRV issues because of dependencies is a user issue, not an author one.

Yeah but the problem is, we run those checks in CI too which then fail because Cargo picks up the latest version, and forces us to bump MSRV.

Committing `Cargo.lock` should fix that if that's the fix we want.

> The only solution I see here is to move the CLIs out of the workspace to have their own `Cargo.lock` (which would still be annoying to maintain, but we can just ignore it until before the release). But having multiple workspaces is also pretty annoying to work with.

I was actually thinking about this too, for other reasons. We have a clear split between set of crates that are intended to be compiled to Wasm and crates that are meant to be native.

Having two workspaces would mean we can just group them correspondingly, and have individual `.cargo/config` with `build.target = "wasm32-unknown-unknown"` for the Wasm ones.

Then, both the local testing and the CI config simplifies to just two lines (`cargo` command individually in each workspace) instead of manually going crate by crate with different options, which should also avoid issues like the one below where we accidentally didn't test some crates on CI at all.

Maybe let's just do that and see how much pain it is to work with 2 workspaces?

daxpedda: We can still use `resolver.incompatible-rust-versions` locally to fix any CI issues, so that's that.

But I'm actually in favor of splitting it into two workspaces. Especially for the `build.target` reason. Its pretty annoying to work with locally right now because we can only have one target at a time.

But that can happen in a follow-up PR, right?

RReverser:
There were few more issues in threads transform tests, all fixed now.

crates/cli-support/src/transforms/threads/tests/mod.rs
```diff
@@ -17,9 +17,14 @@ fn runtest(test: &Test) -> Result<String> {
...
         .parse(&wasm)?;
 
-    wasm_bindgen_threads_xform::run(&mut module)?;
+    super::run(&mut module)?;
     walrus::passes::gc::run(&mut module);
 
+    // We add an extra parameter to the start function, making it invalid for the start section.
+    // It's only valid in combination with wasm-bindgen's "unstart" step, but for the _unit_ test
+    // purposes we can just remove the start section for now.
+    module.start = None;
```
daxpedda: Shouldn't we just do the whole "unstart" step here instead?
Otherwise we aren't actually testing this relatively important part.

RReverser: Then we need to create the whole JS context, which would make it an integration test rather than an isolated unit test.

We already have integration tests for threads, so IMO this one should stay isolated for easier narrowing down of an issue.

README.md
RReverser: > But that can happen in a follow-up PR, right?

Oh yeah, I didn't mean as part of this PR, just as a general discussion.

crates/cli-support/src/transforms/threads/tests/mod.rs
RReverser: Although I guess I can just move out the `unstart_start_function` out of `Context` object, it's not really tied to it. I'll do that, one min.

RReverser: Done. I'd like to stop here as further improvements are getting out of initial scope of this PR.

daxpedda: Fun. I thought there would be more of a difference here.

RReverser: Hah, yeah, it's a pretty trivial step.

RReverser:
@daxpedda Do you want me to hold off on merging this until you make a release, or should it be included in 0.2.101?

daxpedda:
I don't really see a reason it shouldn't be included, feel free to go ahead.

PR #4609, opened 2025-09-03
Switch to Playwright for example tests

RReverser:
The webdriver client in example-tests is a pretty complicated bespoke implementation, and doesn't even work that well as evidenced by CI Firefox issues.

Those get bad enough that e.g. in the last https://github.com/wasm-bindgen/wasm-bindgen/pull/4602 runs I'm getting so many logs of "trying to reconnect to Firefox" that my browser literally hangs trying to load them.

In constrast, in the JS land browser automation is a pretty commonplace and well-solved problem with tools like Puppeteer and Playwright.

Since we rely on `npm` for building examples anyway, might as well switch to JS where all the necessary testing can be done in a few lines with Playwright, with automatic retries, well-established interception APIs, easy support for more browsers in the future etc etc.

As an evidence of extra benefits, it actually caught some errors that weren't caught by the previous implementation, namely that `julia_set` and `weather_report` examples were trying to load some assets they didn't include in the webpack config.

It can be seen on the public URLs of these examples too, e.g. the images on https://wasm-bindgen.github.io/wasm-bindgen/exbuild/weather_report/ are broken: https://wasm-bindgen.github.io/wasm-bindgen/exbuild/weather_report/


daxpedda (commented):
I will let others who have experience with the JS ecosystem leave an actual approval.

But otherwise this LGTM and fixing the CI was already a long standing issue!

RReverser:
Whew, finally... Debugging CIs is never getting easier 😴

I've removed all non-cross-platform `.sh` and `.py` scripts from `examples` in the process and integrated all tests together, so that anyone (including me on Windows) can now just run `pnpm test` in the `examples` folder and be sure it runs all the tests there are.

It also simplifies the CI config since it matches what's happening locally a lot more closely (everything being built into a shared `dist` folder).

PR #4611, opened 2025-09-04
Update team and governance docs

guybedford:
This updates the team info and governance docs removing old outdated references to the Rust wasm project, clarifying the current team, and updating the process as currently agreed upon.

### DIFF
--- a/guide/src/contributing/governance.md
+++ b/guide/src/contributing/governance.md
@@ -1,0 +1,62 @@
+# Governance of the `wasm-bindgen` Organization
+
+This document describes the lightweight governance structure used in the
+[`wasm-bindgen` organization][org].
+
+- [Code of Conduct](#code-of-conduct)
+- [Membership](#membership)
+  - [Core Team](#core-team)
+- [Changes Policy](#changes-policy)
+
+## Code of Conduct
+
+We abide by the [Rust Code of Conduct][conduct] and ask that you do as well,
+with the moderation role performed by the [core team].
+
+## Membership
+
+We welcome new collaborators in the project! The project is not owned by any
+specific entity or group and belongs to the community as a whole.
+
+If you make two or three significant contributions, you should seriously consider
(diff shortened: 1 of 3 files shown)


guide/src/contributing/governance.md
```diff
@@ -0,0 +1,62 @@
...
+
+We abide by the [Rust Code of Conduct][conduct] and ask that you do as well,
+with the moderation role performed by the [core team].
+
+## Membership
+
+We welcome new collaborators in the project! The project is not owned by any
+specific entity or group and belongs to the community as a whole.
+
+If you make two or three significant contributions, you should seriously consider
```
Spxg, line 21: Thanks for sorting out the rules! 

I have a question: what kind of contributions are considered significant?

guide/src/contributing/team.md
```diff
@@ -1,36 +1,34 @@
...
+* [Ingvar Stepanyan](https://github.com/RReverser)
+* [Guy Bedford](https://github.com/guybedford)
+
+## Collaborators
+
+* [Hood Chatham](https://github.com/hoodmane)
+
+## Former Members
+
+`wasm-bindgen` was initially developed under the now sunsetted [Rust Wasm initiative], which included team members:
```
RReverser: Perhaps better to link to https://github.com/orgs/rustwasm/people. There were a lot of people maintaining wasm-bindgen over years, might be better than singling anyone out here.

guide/src/contributing/governance.md
guybedford: Any landed PRs would be considered significant IMO, but note that this is also a suggestion not a rule, and by no means a requirement.

If you think that can be better clarified we could try to make that clearer too.

guide/src/contributing/team.md
guybedford: thanks, added.

PR #4615, opened 2025-09-04
Fix GestureEvent WebIDL

thatcomputerguy0101:
This is a fixed version of #4589 to correct for some erroneous assumptions I made while creating the GestureEvent WebIDL file. Specifically, `clientX`/`clientY` are returned to their correct names, and the nonexistent constructor was replaced with `initGestureEvent` which does exist.

One change that I made from the information in https://developer.apple.com/documentation/webkitjs/gestureevent that is still present is the return type of `initGestureEvent` was made `undefined` instead of `null`. This matches what is returned when using the method in the browser console as well as the other `initXEvent` methods.


daxpedda:
`initGestureEvent` wasn't generated. Looking into it.

PR #4622, opened 2025-09-05
Fix crash caused by allocations during `TypedArray` interactions

Spxg:
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4392
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4395
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4492
resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4532

This problem is difficult to reproduce in actual use, but the consequences are serious, https://github.com/wasm-bindgen/wasm-bindgen/pull/4555 solves the detached problem from `&[T]` to `TypedArray`, this commit will fix the `TypedArray` to `&mut [T]` conversion issue and related methods.

Use `#[wasm_bindgen(js_name = "xxxx")]`, borrowed from `wbg_cast!`https://github.com/wasm-bindgen/wasm-bindgen/pull/4613, to be honest this is a bit hack.

If the changelog description is not appropriate, please feel free to modify it.

### DIFF
--- a/crates/js-sys/tests/wasm/TypedArray.rs
+++ b/crates/js-sys/tests/wasm/TypedArray.rs
@@ -229,18 +229,17 @@ fn copy_to_heap_growth() {
 
 #[wasm_bindgen_test]
 fn copy_from_heap_growth() {
-    let _ = (0..10_000)
-        .map(|_| {
-            let x = Uint8Array::new_with_length(10);
-            x.copy_from(&[1; 10]);
-            x
-        })
-        .collect::<Vec<_>>();
+    let mut v = Vec::with_capacity(10_000);
+    for _ in 0..10_000 {
+        let x = Uint8Array::new_with_length(10);
+        x.copy_from(&[1; 10]);
+        v.push(x);
+    }
 }
 
 #[wasm_bindgen_test]
 fn raw_copy_to_ptr_heap_growth() {
-    let mut v = vec![];
+    let mut v = Vec::with_capacity(10_000);
     for _ in 0..10_000 {
         let x = Uint8Array::new_with_length(10);
         let mut y = [0; 10];
@@ -256,7 +255,7 @@ fn raw_copy_to_ptr_heap_growth() {
 
 #[wasm_bindgen_test]
 fn to_vec_heap_growth() {
-    let mut v = vec![];
+    let mut v = Vec::with_capacity(10_000);
     for _ in 0..10_000 {
         let x = Uint8Array::new_with_length(10);
         // When the externref table capacity is insufficient,


crates/js-sys/src/lib.rs
```diff
@@ -6517,11 +6517,13 @@ macro_rules! arrays {
...
             /// large enough to fit this array's contents.
             pub unsafe fn raw_copy_to_ptr(&self, dst: *mut $ty) {
-                let buf = wasm_bindgen::memory();
-                let mem = buf.unchecked_ref::<WebAssembly::Memory>();
-                let all_wasm_memory = $name::new(&mem.buffer());
-                let offset = dst as usize / mem::size_of::<$ty>();
-                all_wasm_memory.set(self, offset as u32);
+                #[wasm_bindgen]
+                extern "C" {
+                    #[wasm_bindgen(js_name = "/* copy_to_slice */ ((a, b) => b.set(a))")]
```
RReverser: I don't think this hack is necessary tbh, you can instead have proper method-level binding.

The correct binding / fix is pretty trivial, but the problem I had when trying to add it is that I could never reliably trigger the bug so it was hard to demonstrate that the fix changes anything.

Does the test you've added here reliably crash before the PR?

Spxg: > I don't think this hack is necessary tbh, you can instead have proper method-level binding.

`copy_from_slice` can be implemented like this:

```rust
#[wasm_bindgen(method, js_name = set)]
pub fn copy_from_slice(this: &$name, slice: &[$ty]);
```

But `copy_to_slice` can't do that.

> Does the test you've added here reliably crash before the PR?

Yes, it can crash stably

RReverser: > Yes, it can crash stably

Awesome. I'll play with it a bit, but you might be right this is necessary after all.

Spxg:
CI failure introduced by https://github.com/wasm-bindgen/weedle/pull/61, need to wait for https://github.com/wasm-bindgen/wasm-bindgen/pull/4625 to be merged

crates/js-sys/src/lib.rs
RReverser: So the implementations I was thinking of is just

```rust
pub unsafe fn view(rust: &[$ty]) -> $name {
    wasm_bindgen::wbg_cast!(rust, &[$ty], $name)
}

/// Efficiently copies the contents of this JS typed array into a new Vec.
pub fn to_vec(&self) -> Vec<$ty> {
    wasm_bindgen::wbg_cast!(self, &$name, Vec<$ty>)
}
```

As far as I can tell, that's enough to pass the added tests, and it's a bit more concise as well, but of course if there are issues with the "raw" copy to slice, then this wouldn't fix them.

It would be good to add tests for those too to see if this fixup is really enough.

Spxg: > if there are issues with the "raw" copy to slice, then this wouldn't fix them

`copy_to`, `copy_from`, `raw_copy_to_ptr` all have the same problem, I have added tests that can reproduce them.

daxpedda:
That's my bad, we should probably pin `weedle`.

crates/js-sys/src/lib.rs
Spxg: > wasm_bindgen::wbg_cast!(rust, &[$ty], $name)

Thanks, this really simplifies the view implementation.

crates/js-sys/src/lib.rs
```diff
@@ -6495,13 +6489,8 @@ macro_rules! arrays {
...
             pub unsafe fn view_mut_raw(ptr: *mut $ty, length: usize) -> $name {
-                let buf = wasm_bindgen::memory();
-                let mem = buf.unchecked_ref::<WebAssembly::Memory>();
-                $name::new_with_byte_offset_and_length(
-                    &mem.buffer(),
-                    ptr as u32,
-                    length as u32
-                )
+                let slice = core::slice::from_raw_parts_mut(ptr, length);
+                wasm_bindgen::wbg_cast!(slice, &mut [$ty], $name)
```
RReverser:
```suggestion
                Self::view(slice)
```

crates/js-sys/src/lib.rs
```diff
@@ -6568,17 +6559,23 @@ macro_rules! arrays {
             /// This function will panic if this typed array's length is
             /// different than the length of the provided `src` array.
             pub fn copy_from(&self, src: &[$ty]) {
+                #[wasm_bindgen]
+                extern "C" {
```
RReverser: Please reuse methods one in another as much as possible. No need to generate extra bindings where we can use an existing view / set helper.

Spxg: It is definitely unsafe to use `view` in rust. As long as the heap grows, the view buffer will be detached, so I implemented the copy process in js.

```rust
#[wasm_bindgen_test::wasm_bindgen_test]
fn it_works() {
    let _ = (0..10_000)
        .map(|_| unsafe {
            let x = Uint8Array::view(&[1, 2, 3]);
            // use `set` in rust
            x.set(&JsValue::from(Uint8Array::new_with_length(3)), 0);
            x
        })
        .collect::<Vec<_>>();
}
```

Spxg: > As long as the heap grows, the view buffer will be detached, so I implemented the copy process in js.

In other words, it must complete the copy process in JS at one time

CHANGELOG.md
```diff
@@ -8,11 +8,14 @@
 * Added `DocumentOrShadowRoot.adoptedStyleSheets`.
   [#4625](https://github.com/wasm-bindgen/wasm-bindgen/pull/4625)
 
-### Changed
+### Fixed
 
 * Fixed wrong method names for `GestureEvent` bindings.
   [#4615](https://github.com/wasm-bindgen/wasm-bindgen/pull/4615)
 
+* Fix crash caused by buffer detached during copy
```
```suggestion
* Fix crash caused by allocations during `TypedArray` interactions.
```
WDYT?

I'm trying to keep the changelog user-centric. So instead of discussing the underlying internal issue, we should talk about what the user experiences.

Spxg: I totally agree

crates/js-sys/src/lib.rs
RReverser: I didn't mean `view`. You mentioned earlier that `copy_from_slice` can use plain bindings and not this hack - so we can do that here, right?

crates/js-sys/src/lib.rs
```diff
@@ -6517,11 +6506,13 @@ macro_rules! arrays {
...
             /// large enough to fit this array's contents.
             pub unsafe fn raw_copy_to_ptr(&self, dst: *mut $ty) {
-                let buf = wasm_bindgen::memory();
-                let mem = buf.unchecked_ref::<WebAssembly::Memory>();
-                let all_wasm_memory = $name::new(&mem.buffer());
-                let offset = dst as usize / mem::size_of::<$ty>();
-                all_wasm_memory.set(self, offset as u32);
+                #[wasm_bindgen]
+                extern "C" {
+                    #[wasm_bindgen(js_name = "/* copy_to_slice */ ((a, b) => b.set(a))")]
```
RReverser: I'd suggest

```rust
#[wasm_bindgen(js_namespace = $name, js_name = ".prototype.set.call")]
```

and flip the arguments around, otherwise you're creating a fresh closure on each invocation. Even with the hack `js_name` should be "pure".

Spxg: Thanks, learned

RReverser (approved):

crates/js-sys/src/lib.rs
```diff
@@ -6447,6 +6447,20 @@ macro_rules! arrays {
...
+            /// This method is not expected to be public. It requires the length of the
+            /// TypedArray to be the same as the slice, use `self.copy_from(slice)` instead.
+            #[wasm_bindgen(method, js_name = set)]
+            fn copy_from_slice(this: &$name, slice: &[$ty]);
+
+            /// Copies this TypedArray's data to Rust slice;
+            ///
+            /// This method is not expected to be public. It requires the length of the
+            /// TypedArray to be the same as the slice, use `self.copy_to(slice)` instead.
+            #[wasm_bindgen(js_namespace = $name, js_name = "prototype.set.call")]
```
daxpedda, line 6471: We should add a comment here describing this workaround.

daxpedda (changes requested):

PR #4626, opened 2025-09-06
Bump dependencies for various `/crates` and `/examples`

abhillman:


crates/backend/Cargo.toml
```diff
@@ -17,7 +17,7 @@ version = "0.2.101"
 extra-traits = ["syn/extra-traits"]
 
 [dependencies]
-bumpalo = "3.0.0"
+bumpalo = "3.19.0"
```
daxpedda, line 20: Ideally we don't want to specify compatible versions unless we are setting the minimal version required.

Equally, we don't want to raise the minimal version required unless actually necessary.

```suggestion
bumpalo = "3"
```

crates/cli/Cargo.toml
```diff
@@ -40,14 +40,14 @@ assert_cmd = "2"
 diff = "0.1"
 predicates = "3"
 rstest = "0.26"
-wasmparser = "0.214"
-wasmprinter = "0.214"
+wasmparser = "0.238.1"
```
daxpedda, line 43: We don't really want to bump `wasmparser` & co. until we update Walrus. Otherwise these very heavy dependencies are duplicated.

daxpedda (changes requested):

daxpedda:
Closing as out-of-date.

PR #4631, opened 2025-09-09
Preserve imports in reference tests

RReverser:
These are useful to have in snapshots, and are otherwise collected, making all modules look almost the same.


RReverser:
Ah great, so imports are also sorted differently across OS... I'll add a fixup similar to #4605.

crates/cli/tests/reference.rs
```diff
@@ -219,16 +222,42 @@ fn sanitize_wasm(wasm: &Path) -> Result<String> {
...
+    module.elements.delete(temp_element_id);
+    // Sort imports for deterministic snapshot.
+    std::mem::take(&mut module.imports)
+        .iter()
+        .map(|i| ((&i.module, &i.name), i.kind.clone()))
+        .collect::<BTreeMap<_, _>>()
+        .into_iter()
+        .for_each(|((module_name, name), kind)| {
+            module.imports.add(module_name, name, kind);
+        });
```
guybedford, line 259: This feels like something that should be in Walrus itself and could be done more optimally there?

Just to mention - if you want to do an upstream feature here, I'd be happy to merge and release too.

RReverser: Yeah there's lots of stuff that is awkward to do in Walrus, but at least this one is only useful for testing. You mentioned we might switch to a different parser anyway, so I think it's worth waiting.

PR #4635, opened 2025-09-11
Generate DWARF for tests by default

Spxg:
Original PR: https://github.com/wasm-bindgen/wasm-bindgen/pull/4435

After force pushing, it couldn't be reopened, so I had to create a new one..


crates/cli/src/bin/wasm-bindgen-test-runner/main.rs
```diff
@@ -120,12 +120,19 @@ fn main() -> anyhow::Result<()> {
         .map(Path::new)
         .context("file to test is not a valid file, can't extract file name")?;
 
+    let debug = env::var("WASM_BINDGEN_NO_DEBUG").is_err();
+
+    let no_generate_dwarf = env::var("WASM_BINDGEN_NO_GENERATE_DWARF").is_ok();
```
daxpedda:
```suggestion
    let no_generate_dwarf = env::var("WASM_BINDGEN_TEST_NO_DWARF").is_ok();
```
WDYT?

Spxg:
> What environment are you using that is emitting better error information with DWARF on?

After installing https://goo.gle/wasm-debugging-extension, can debug in chrome devtools:

image

I think this will be helpful for debugging.

daxpedda:
> > What environment are you using that is emitting better error information with DWARF on?
> 
> After installing https://goo.gle/wasm-debugging-extension, can debug in chrome devtools:

Yeah, that's what I've been using as well. I was hoping that you finally found an environment with native support :sweat_smile:.

So in any case, I suggest the following: we can go ahead and enable it by default, but remove the flag for removing DWARF output. The idea is that ultimately we want to enable DWARF by default without needing all these control knobs. Rust users who don't want DWARF in their tests can turn it off via `RUSTFLAGS` or Cargo profiles, like with any other target.

Can you check that no DWARF is emitted if debug info is disabled via Rust? Ideally we would have a test, but it might be quite difficult to implement.

See https://github.com/wasm-bindgen/walrus/pull/284, https://github.com/wasm-bindgen/wasm-bindgen/pull/3876, https://github.com/wasm-bindgen/wasm-bindgen/issues/3698.

(sorry for making you write the documentation about the environment variable just to ask you to remove it afterwards :/, completely forgot that we don't actually need these knobs)

Spxg:
> I was hoping that you finally found an environment with native support 😅.

I think I understand what you said, it still needs tools like `wasm-debugging-extension` to generate sourceMap, so it's not natively supported.

> Can you check that no DWARF is emitted if debug info is disabled via Rust? Ideally we would have a test, but it might be quite difficult to implement.

I will check it out

RReverser:
> it still needs tools like `wasm-debugging-extension` to generate sourceMap

Just to be clear, those tools don't generate source map - that's the old approach supported by scripts like https://github.com/emscripten-core/emscripten/blob/main/tools/wasm-sourcemap.py.

What the extension does is runs (parts of) actual LLDB in Wasm, in order to also support things like executing C++ expressions from devtools console, pretty-print types, and so on.

RReverser:
> Yeah, that's what I've been using as well. I was hoping that you finally found an environment with native support 😅.

Also - might be not what you meant by native, but VSCode has extension based on the Chrome DevTools one: https://marketplace.visualstudio.com/items?itemName=ms-vscode.wasm-dwarf-debugging

RReverser:
> esp given we aren't seeing test run times increase or other negative effects - unless I'm missing anything here though

I suspect this is because we still build all tests with DWARF enabled in Cargo / Rust itself, just not enabled for wasm-bindgen postprocessing, so this change doesn't affect compilation time much.

If we disabled them proper, we might get some compilation speedup, depending on how noisy times in CI are otherwise.

Spxg:
> Can you check that no DWARF is emitted if debug info is disabled via Rust?

Yes, that no DWARF is emitted if debug info is disabled via Rust.  

Perhaps we could generate gwarf by default for all products, not just tests, just like https://github.com/wasm-bindgen/wasm-bindgen/pull/3876. Users could configure the level of debug-info they want to retain via a debug profile: https://doc.rust-lang.org/cargo/reference/profiles.html#debug

daxpedda:
> Perhaps we could generate gwarf by default for all products, not just tests, just like #3876.

Well we decided against that in #3876. It breaks some important post-wbg-processors. Most notably `wasm-opt`, which isn't very good at handling DWARF.

So I believe we should wait until we have proper source map support.

Spxg:
> It breaks some important post-wbg-processors. Most notably `wasm-opt`, which isn't very good at handling DWARF.

It's still a problem? From https://github.com/WebAssembly/binaryen/issues/6391, it should be fixed by https://github.com/wasm-bindgen/walrus/pull/284

Sorry, that's all I know

daxpedda:
> > It breaks some important post-wbg-processors. Most notably `wasm-opt`, which isn't very good at handling DWARF.
> 
> It's still a problem? From [WebAssembly/binaryen#6391](https://github.com/WebAssembly/binaryen/issues/6391), it should be fixed by [wasm-bindgen/walrus#284](https://github.com/wasm-bindgen/walrus/pull/284)

So the problem https://github.com/wasm-bindgen/walrus/pull/284 fixed isn't the DWARF output itself, it just ensures that no DWARF output is generated when there is no debug info to begin with, ergo it was disabled via Cargo/Rustc.

But the problem isn't with our DWARF output to begin with, Binaryen has stated a couple of times already that there handling of DWARF is broken, so that's not something we can fix on our side anyway.

The idea of us generating source maps would solve all these issues, because most tools in the ecosystem only support source maps anyway. Not to mention that no extension would be required to let browsers be able to read the debug information and the fact that debug information is split off from the Wasm module, which reduces load times.
> Sorry, that's all I know

No problem! I barely remember all the details, so I appreciate you questioning all the assumptions so we actually make sure we are making the proper decision here!

RReverser:
> because most tools in the ecosystem only support source maps anyway

I'm curious, which tools are you referring to that support source maps for WebAssembly? That was mostly an early experiment Aside from Chrome (which has much more robust DWARF debugging above), I only know of Firefox that implemented it back in the day, and it was still considered mostly an experiment as source maps are a pretty poor fit for binaries and used weird hacks to make Wasm work.

I really don't think we should try and invest any more time in source maps in this age.

daxpedda:
AFAIK all browsers support source maps for Wasm. I'm using it for Chrome and Firefox on the regular, but have seen improved stack traces with Safari in the past as well. `wasm-opt` does so as well, currently source maps are the only way to retain debug information after `wasm-opt` transformations.

Chrome doesn't officially support DWARF, it requires extensions. AFAIK no environment officially supports Wasm DWARF so far.

RReverser:
> AFAIK all browsers support source maps for Wasm.

Never heard about Safari adding it, but I could be well wrong.

> currently source maps are the only way to retain debug information after `wasm-opt` transformations

That doesn't sound right, they have a few bugs (as the linked issue mentioned) but for the most part `wasm-opt` worked fine with DWARF for years, and Emscripten relies on it itself.

> Chrome doesn't officially support DWARF, it requires extensions.

Having been part of that development (I built the early prototype of DWARF support when I was on the Chrome team), that _is_ the official support by the Chrome team, with DevTools itself having a dedicated flag that connects to said extension. The only reason DWARF support is shipped separately is to avoid blowing the size of DevTools for all users who don't care about Wasm.

daxpedda:
> > AFAIK all browsers support source maps for Wasm.
> 
> Never heard about Safari adding it, but I could be well wrong.

Don't want to lie: it was a long time ago, maybe I remember it wrong.
Its for sure in Chrome and Firefox because I'm using it right now.

> > currently source maps are the only way to retain debug information after `wasm-opt` transformations
> 
> That doesn't sound right, they have a few bugs (as the linked issue mentioned) but for the most part `wasm-opt` worked fine with DWARF for years, and Emscripten relies on it itself.

I was told in the past to use source maps instead of DWARF for `wasm-opt` by maintainers. If you search in the `binaryen` repo for issues mentioning "DWARF" you will find the same quite fast.

EDIT: from personal experience I can also tell you that `wasm-opt` is unable to produce valid DWARF with `wasm-bindgen` output right now. In fact, 95% of the time it just crashes outright. While DWARF to source map converters and feeding that into `wasm-opt` seems to work just fine.

> > Chrome doesn't officially support DWARF, it requires extensions.
> 
> Having been part of that development (I built the early prototype of DWARF support when I was on the Chrome team), that _is_ the official support by the Chrome team, with DevTools itself having a dedicated flag that connects to said extension. The only reason DWARF support is shipped separately is to avoid blowing the size of DevTools for all users who don't care about Wasm.

I probably shouldn't have used the word "official", because I've no idea what the official stance actually is.
That's a strange decision, because it kinda makes it harder for e.g. automated tests with WebDriver's and such.

---

In essence, I would prefer if we could stick with DWARF if possible, but AFAICS the support has factually shifted to source maps.

To summarize:
- Chrome: source maps, DWARF with an official extension.
- Firefox: source maps (it briefly had DWARF support in the past)
- Safari: source maps (we should double-check that)
- `wasm-opt`: source maps recommended, DWARF is factually broken (see e.g. https://github.com/WebAssembly/binaryen/issues/7503#issuecomment-2807019571 or just search for DWARF issues)
- Node: idk
- Deno: I assume they support source maps considering the use of V8, but no idea

RReverser:
> * `wasm-opt`: source maps recommended, DWARF is factually broken (see e.g. [Generate webassemby binary with DWARF information WebAssembly/binaryen#7503 (comment)](https://github.com/WebAssembly/binaryen/issues/7503#issuecomment-2807019571) or just search for DWARF issues)

Ah, focus on LLVM DWARF output makes sense, probably why we have different impressions of level of support. There was zero support a few years ago, then a big push, and I haven't had issues with C++ code for a while now, but I can see how custom produced DWARF can still run into limitations.

That said, for debugging we shouldn't run `wasm-opt` anyway (and e.g. wasm-pack doesn't), so shouldn't be an issue?

> That's a strange decision, because it kinda makes it harder for e.g. automated tests with WebDriver's and such.

You can install extensions in those too via `--load-extension`, eg here are docs for Playwright we're using for example tests now: https://playwright.dev/docs/chrome-extensions

---

Overall, I'm just finding source maps almost useless - you can get decent stacktraces, and some basic stepping, but that's about it. Even stepping can behave weirdly (I fixed a few bugs in devtools back in the day, but that's still browser-dependant), and not being able to see real values in C++/Rust sense instead of obscure integers in anonymous Wasm locals & stack makes it very far from proper debugging experience.

daxpedda:
> That said, for debugging we shouldn't run `wasm-opt` anyway (and e.g. wasm-pack doesn't), so shouldn't be an issue?

A lot of people are asking for better stacktraces for their production code. I'm unsure how e.g. Sentry is handling this sort of stuff, but any tool will need valid debug information.

> Overall, I'm just finding source maps almost useless - you can get decent stacktraces, and some basic stepping, but that's about it.

For sure, I'm using Chrome with DWARF support myself to do Wasm debugging. We should definitely support both DWARF and source maps.

I'm just of the mind that source map has to be the default unless the landscape changes. Mangled stack traces are one of the worst sore points of the WBG experience today. While `console_error_panic_hook` alleviates this somewhat, it is much improved with source maps.

RReverser:
> I'm unsure how e.g. Sentry is handling this sort of stuff, but any tool will need valid debug information.

Oh they use DWARF too. https://docs.sentry.io/platforms/native/guides/wasm/

RReverser:
> Mangled stack traces are one of the worst sore points of the WBG experience today.

If it's just about having source locations in stack traces, we could probably just add source file/line/column to each function name in the names section?

That would have much less size overhead than entire source map, which likely matters for prod code.

Spxg:
The above discussion is a bit off the topic of this PR, I opened an issue for everyone to discuss. https://github.com/wasm-bindgen/wasm-bindgen/issues/4643

Thanks for the above discussion, I learned a lot!

PR #4636, opened 2025-09-11
Self-host almost all intrinsic signatures

RReverser:
Instead of describing intrinsic signatures out of band in cli-support, manually describing their raw signatures in wasm-bindgen, and unsafely converting from one to another, we can just declare almost all the intrinsics as a module import via the `#[wasm_bindgen]` macro.

This way they're guaranteed to stay in sync, and we can get rid of a bunch of `unsafe` code.

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -492,10 +492,10 @@ where
     T: ?Sized,
 {
     fn drop(&mut self) {
-        unsafe {
-            // this will implicitly drop our strong reference in addition to
-            // invalidating all future invocations of the closure
-            if super::__wbindgen_cb_drop(ManuallyDrop::take(&mut self.js)) {
+        // this will implicitly drop our strong reference in addition to
+        // invalidating all future invocations of the closure
+        if super::__wbindgen_cb_drop(&self.js) {
+            unsafe {
                 ManuallyDrop::drop(&mut self.data.inner);
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1128,7 +1128,7 @@ extern "C" {
     fn __wbindgen_throw(msg: &str) /* -> ! */;
     fn __wbindgen_rethrow(js: JsValue) /* -> ! */;
 
-    fn __wbindgen_cb_drop(js: JsValue) -> bool;
+    fn __wbindgen_cb_drop(js: &JsValue) -> bool;
 
     fn __wbindgen_jsval_eq(a: &JsValue, b: &JsValue) -> bool;
     fn __wbindgen_jsval_loose_eq(a: &JsValue, b: &JsValue) -> bool;


src/closure.rs
```diff
@@ -495,7 +495,7 @@ where
         unsafe {
             // this will implicitly drop our strong reference in addition to
             // invalidating all future invocations of the closure
-            if super::__wbindgen_cb_drop(self.js.idx) != 0 {
+            if super::__wbindgen_cb_drop(ManuallyDrop::take(&mut self.js)) {
```
daxpedda: Shouldn't we just keep the current code and drop `unsafe` instead?

RReverser: I can change it to take `&JsValue` instead, but the `unsafe` would still be necessary for the next line.

RReverser:
> Does this have any real effects on the generated Wasm?

Aside from import names, no - from Wasm point of view it's still the same raw import with same raw ABI as before.

src/closure.rs
RReverser: I did this for now, at least it limits the `unsafe` scope a bit.

src/lib.rs
```diff
@@ -1095,71 +1079,80 @@ extern "C" {
...
+    fn __wbindgen_gt(a: &JsValue, b: &JsValue) -> bool;
+
+    fn __wbindgen_number_get(js: &JsValue) -> Option<f64>;
+    fn __wbindgen_boolean_get(js: &JsValue) -> Option<bool>;
+    fn __wbindgen_string_get(js: &JsValue) -> Option<String>;
+    fn __wbindgen_bigint_get_as_i64(js: &JsValue) -> Option<i64>;
+
+    fn __wbindgen_debug_string(js: &JsValue) -> String;
+
+    fn __wbindgen_throw(msg: &str) /* -> ! */;
```
daxpedda, line 1128: That's a bit unfortunate.
Do we even want to add support for `!` in the proc-macro one day?

RReverser: Hm do you mean to simulate never type support before it's stabilised in Rust?

Mayyybe, but given how extremely rare that type is outside of our own usecase, I wouldn't worry about it for now.

daxpedda: No, absolutely not.

Just wanted to know if you had a reason in mind, because I couldn't. The downside is only two new `unsafe` lines, which seems fine with me.

RReverser: Yeah I don't have any reason to. If/when never type is finally stabilized, we can add FromWasmAbi and friends for it, but specially handling it in macro is not worth it.

> The downside is only two new `unsafe` lines, which seems fine with me.

Also note that technically previous bindings for these functions were wrapped in `unsafe` too (because they were raw C-like bindings), so technically these 2 unsafes are not even new, just changed in scope.

PR #4637, opened 2025-09-11
Update documentation about doctest

Spxg:
https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#doctest-xcompile

`-Zdoctest-xcompile` is now stable in Rust 1.89, which means running cargo test will also run doctest.

### DIFF
--- a/guide/src/wasm-bindgen-test/usage.md
+++ b/guide/src/wasm-bindgen-test/usage.md
@@ -128,7 +128,12 @@ Run the tests by passing `--target wasm32-unknown-unknown` to `cargo test`:
 cargo test --target wasm32-unknown-unknown
 ```
 
-If you also need to run doctests, add the unstable [`-Zdoctest-xcompile`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#doctest-xcompile) flag. This requires using the Rust nightly channel like this:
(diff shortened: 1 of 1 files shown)


guide/src/wasm-bindgen-test/usage.md
```diff
@@ -128,8 +128,4 @@ Run the tests by passing `--target wasm32-unknown-unknown` to `cargo test`:
 cargo test --target wasm32-unknown-unknown
 ```
 
-If you also need to run doctests, add the unstable [`-Zdoctest-xcompile`](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#doctest-xcompile) flag. This requires using the Rust nightly channel like this:
-
-```
-cargo +nightly test --target wasm32-unknown-unknown -Zdoctest-xcompile
-```
+This will run all tests, including doctests (requires at least Rust 1.89).
```
daxpedda: I think we should mention that this is doctest specific and not `cargo test` specific. E.g. `cargo test --doc` won't work either below Rust v1.89.

```suggestion
Running doctests requires at least Rust v1.89.
```

WDYT?

Spxg: Great advice, thanks

daxpedda (approved):

RReverser (approved):

PR #4638, opened 2025-09-12
Support --target=module target

guybedford:
This adds support for a new target for ESM Integration source phase imports - `--target=module`.

When used, output is generated to output into the syntax `import source wasmModule from './mod.wasm'` for the Wasm loading, as defined in the latest ESM integration and source phase imports proposal (https://github.com/tc39/proposal-source-phase-imports).

The reason why this is a new target and not combined with the other targets is because source phase imports are designed to form a new type of portable target for Wasm output generation that can work across bundlers, Node.js and the web.


justfile
daxpedda, line 1: I believe `CONTRIBUTING.md` needs to be updated?

guybedford: Thanks, added.

guybedford:
> Does this have any relation to the experimental-nodejs-module target?
I feel like this could be an extension to the existing ESM target (whichever one that is) instead of a whole target of its own.

There are two modes of the Wasm ESM Integration - instance imports (which is what the experimental-nodejs-module is) and source phase imports.

In Node.js instance imports are still experimental, so exactly match up with experimental-nodejs-module support and its experimental status.

In Node.js, source phase imports are stable. In addition as a future portable target, they truly are a baseline target and not a target property. This is one of the driving motivations of the proposal - to be able to define a single cross-platform target for Wasm output.

daxpedda:
I'm still a bit perplexed, because it looks to me like source-phase imports require the Wasm ESM integration proposal?
Or am I getting this wrong?

Basically, I'm trying to understand why these are two different modes and not source-phase imports being an extension.
If it is an extension, I would rather stabilize Wasm ESM integration, which is in phase 3 already, and make source-phase imports an additional flag. Not too sure how cross-target compatibility comes into play here.

Apologies for my ignorance on this topic ...

guybedford:
Sure, happy to explain further, and definitely let's get terminology right to start.

1. The "ESM Integration" is any ability to import Wasm in JS - any use of `import 'file.wasm'`
2. The instance integration is `import { feat } from './file.wasm'` which causes execution and linking. This is still not fully stable.
3. The source phase integration is `import source mod from './file.wasm'` returning compiled Wasm. This is fully stable.

(3) is the target being defined here, not (2), even though they are both part of the "ESM Integration proposal".

For implementations - Node.js gives an experimental warning if you use (2) but not (3). Chrome is expected to land (3) quite soon, but not (2) for a while.

So to answer your question - implementations are shipping them effectively as separate proposals with different stability levels despite being one proposal.

guybedford:
The Node.js docs may also help here - https://nodejs.org/docs/latest/api/esm.html#wasm-modules

daxpedda:
What do you think of this proposal:
- (Optionally) remove the `experimental-nodejs-module` target.
- Rename `source-phase` to `module`, to mean the ESM integration proposal.

Or is there a good reason to keep instance integration (2) around if (3) is available?

guybedford:
Yes, the `experimental-nodejs-module` target is still important, and I think the long-term future will likely converge on there being two target options - source phase or module (with experimental instance).

daxpedda:
> Yes, the `experimental-nodejs-module` target is still important, and I think the long-term future will likely converge on there being two target options - source phase or module (with experimental instance).

Would you mind elaborating on that? Why will there be two target options in the long-term future?
It seems to me that source phase is a strict improvement.

guybedford:
Yes source phase is likely the default recommended target going forward, and could be made the default in a major change.

But the instance phase will still exist and have its own benefits and use cases, and already exactly matches what `experimental-nodejs-module` already is today.

daxpedda:
After digging more into this I ended up with more questions then answers really.

So `experimental-nodejs-module` doesn't even make use of the new import semantic, it just uses the Wasm `import "./js_file.js"` semantic. Not sure why.

While I initially thought that source phase is a strict upgrade to "instance integration", it actually just seems to control over imports. I'm not sure I understand what benefits this gives us, apart from it being/becoming stable.

From what I can tell, if we use source-phase or instance integration seems to be an internal choice right now that doesn't affect users directly, apart from environment support. So I would still like to go with my initial proposal, of having a single target for ESM integration. "Instance integration" or source-phase imports would just be a modifier for that target, for environment compatibility purposes. Where source-phase is the default right now, because its stable.

I'm afraid I got something completely wrong here :/

guybedford:
I like the framing of a single target - `esm-integration`, where that target then has an instance modifier.

But I also don't see why we can't just have `--target=source-phase` or `--target=instance-phase` as two separate top-level targets. That seems easier to type for users, and easier to understand to me as well as they are distinct targets with distinct properties, even if conceptually they arise from the same spec, they have very different behaviours. The idea of target mutators seems more useful to manage compatibility within targets such as enabling or disabling new features or options that may or may not be available specifically within that target.

> While I initially thought that source phase is a strict upgrade to "instance integration", it actually just seems to control over imports. I'm not sure I understand what benefits this gives us, apart from it being/becoming stable.

The source phase imports proposal avoids the need for different targets in the first place - bundler / node / deno / web all exist because each of these has a different way of loading and compiling Wasm bytes, and slightly different code paths in how to do that in an optimized way (`instantiateStreaming` for browsers, Node.js builtins for Node etc). The source phase exactly eliminiates this distinction between targets providing a single streamlined way of obtaining the compiled Wasm module. It's not some major ground breaking thing, but it exactly does what we need to have a new target here and in this PR.

Please don't feel bad asking questions - when you feel bad it takes up both our time me having me tell you not to feel bad about it. We can just get straight to the point too and I trust these discussions are in good faith, and that is always assumed.

I think we are pretty close, it seems like there are maybe two paths forward here:

1, `--target=source-phase` (or maybe another name like `--target=module-source`) with a later `--target=module-instance`
2. `--target=module` with a `--target=module --instance-phase` modifier

Then let me know if that sums up where we are here? Would you then prefer we go with `--target=module-source` say for this PR?

Also to contrast, here's the emscripten docs for the same - https://emscripten.org/docs/compiling/Modularized-Output.html#source-phase-imports-experimental. Note they use `SOURCE_PHASE_IMPORTS` for the source phase and `ESM_INTEGRATION` for ths instance phase.

The instance phase has other interesting properties though like native tree-shaking which the source phase doesn't as easily support, which make it an attractive target as well when in one day has enough support and features to be usable.

So my preference remains for (1) over (2) though for the typing overhead, but it's certainly not a hill I'd choose to dig trenches on either.

guybedford:
I went ahead and pushed up a new commit which I believe is in line with your suggestion of just using `--target=module`. So effectively just doing (2). It might be simpler to understand for users, which is what swayed me. But let me know if this is what you were after or not.

daxpedda:
> The source phase exactly eliminiates this distinction between targets providing a single streamlined way of obtaining the compiled Wasm module.

I think this is exactly the point I'm confused about. AFAICS both source-phase imports and "instance integration" are able to do this via `import` (even though `experimental-nodejs-module` currently doesn't do that). Its just that source-phase imports allows you to specify imports in JS where "instance integration" does not.

Am I getting this part right?

> Please don't feel bad asking questions - when you feel bad it takes up both our time me having me tell you not to feel bad about it. We can just get straight to the point too and I trust these discussions are in good faith, and that is always assumed.

Appreciated!

> I think we are pretty close, it seems like there are maybe two paths forward here:
> 
> 1. `--target=source-phase` (or maybe another name like `--target=module-source`) with a later `--target=module-instance`
> 2. `--target=module` with a `--target=module --instance-phase` modifier
> 
> Then let me know if that sums up where we are here?

Yes.

> Would you then prefer we go with `--target=module-source` say for this PR?

My understanding so far is that going via source-phase imports or "instance integration" is an internal matter. The user is only so far concerned that they need feature support. If this is correct, I believe we should go with 2. If not, 1. seems the way to go to me.

I think the part I'm missing here is that I don't understand when users would choose between the two modes. Because WBG generates the JS in both cases that handles imports and loading the Wasm file, it really seems to me like the user has no role to play here (apart from the aforementioned feature support).

> So my preference remains for (1) over (2) though for the typing overhead, but it's certainly not a hill I'd choose to dig trenches on either.

I believe from the conversation so far we will probably land on the same preference when I finally understand how the end-game will look like for users.

guybedford:
> I think this is exactly the point I'm confused about. AFAICS both source-phase imports and "instance integration" are able to do this via import (even though experimental-nodejs-module currently doesn't do that). Its just that source-phase imports allows you to specify imports in JS where "instance integration" does not.

> Am I getting this part right?

Yes exactly. Source phase reifies just compilation, instance phase reifies compilation, instantiation and imports resolution.

> My understanding so far is that going via source-phase imports or "instance integration" is an internal matter. The user is only so far concerned that they need feature support. If this is correct, I believe we should go with 2. If not, 1. seems the way to go to me.

I think it's a user-facing decision for two reasons:

1. Source phase compat is expected to come to general availability as a recommended target over the coming year or so, while instance phase compat is still not as clear here.
2. If instance phase does become a recommended target itself, to be used instead of the source phase, there would still remain motivations to use the source phase - it has more flexibility in terms of being able to support multiple instantiations, custom mocking or virtualizations setting custom imports etc. This makes it inherently more configurable where users wanting more fine-grained control over linking will always want this "lower-level" ability.

In theory we could infer the right mode output, for example perhaps one way to frame it actually is that if "virtualization" configuration options are used, and we were hypothetically supporting an instance phase default output at any point in future if that become suitable, then the "dropping down" to the source phase could happen automatically and transparently as those virtualization flags get turned on.

So perhaps this does look like `Target::Module` where `Target::Module { virtualImports: Some(VirtualImportconfig), vmManagementFunctions: Some(VMControlOptions) }` auto enables source phase, whereas `Target::Module` could otherwise could in theory default to the instance phase if support warranted.

But then additionally from a compat perspective, in order to ensure the target is supported, we would likely also need to allow platform target options like `--platform node=22` or `--platform browsers="> .5% or last 2 versions"` say, then we could also automatically infer which underlying primitives to use, and whether we can use the source phase (more likely) or instance phase (less likely).

So while in theory we could encapsulate all this per the above three paragraphs, I think it would be very easy for it to be a leaky abstraction, such that `--target=module-source` and `--target=module-instance` might well be simpler on all sides overall. Then if you do `--target=module-instance --vm-controls=import-call-tracer,memory-functions` you'd get an error "must use --target=module-source target virtual machine control options" kind of thing, rather than automatic "magic" target switches (for some hypothetical vm feature of course!).

guybedford:
> I'm more of the mind that source-phase imports should be the default and going down to instance-phase is the knob users can turn for compatibility reasons. I believe this would address most of the situations you laid out.

Yes, this makes sense. So under this approach I'd like to recommend we land the source phase proposal as `--module`, and then if we ever add instance phase in future to land that as `--module-instance` or some other module mode modifier.

I've gone ahead and renamed the internal `OutputMode::SourcePhase` to `OutputMode::Module` as well, to then land that further.

PR #4640, opened 2025-09-14
Fix intrinsic typo

RReverser:
Eeek, not sure why I typed that in the first place but this slipped by in #4636.

Another reason to fix CI not running with the non-externref JS fallback (as in externref mode this intrinsic gets replaced with an inline function)...

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -139,7 +139,7 @@ impl<'a> Context<'a> {
             "__wbindgen_object_clone_ref",
             vec![Descriptor::I32],
             Descriptor::I32,
-            AuxImport::Intrinsic(Intrinsic::Throw),
+            AuxImport::Intrinsic(Intrinsic::ObjectCloneRef),
         )?;
         self.add_aux_import_to_import_map(
             "__wbindgen_object_drop_ref",


daxpedda (approved):
We should really add `wasm32v1-none` to the CI ...

RReverser:
> We should really add `wasm32v1-none` to the CI ...

I don't think that would be enough because a lot of tests depend on `std` and `wasm32v1-none` made an explicit choice to deviate from `wasm32-unknown-unknown` and not provide `std` with `panic` stubs :(

I think we need to instead bring back the `WASM_BINDGEN_EXTERNREF` environment variable, or a command-line option, that, when explicitly set to disabled, ignores the `reference-types` in the `target_features` section.

daxpedda:
We need to test `wasm32v1-none` anyway to make sure we continue to support this target.

The issue with bringing back some environment variable that ignores `reference-types` is that users will be unable to make use of it. The issue being that Rust already inserts `reference-types` specific instructions without us adding any in the post-processing.

Instead, we should test what users would do if they want to disable `reference-types` with `wasm32-unknown-unknown` in practice: use `-Zbuild-std`. We already do this for `atomics` as well.

RReverser:
> The issue being that Rust already inserts `reference-types` specific instructions

It only does that in table indices, but it doesn't use `externref` type itself, which is what causes the divergence for us - we're producing it with an extra transform.

Alternatively, we should just run tests additionally with Rust 1.81, or even 1.57, as it would cover both the JS output and generally more extensive MSRV checks.

daxpedda:
> > The issue being that Rust already inserts `reference-types` specific instructions
> 
> It only does that in table indices, but it doesn't use `externref` type itself, which is what causes the divergence for us - we're producing it with an extra transform.

What I meant is that we should test what users will use in the real world, to make sure our tests are as close as possible to a real world scenario. The only reason users want to disable `reference-types` is for feature support, e.g. old versions of Safari, in which case table indices breaks that already.

> Alternatively, we should just run tests additionally with Rust 1.81, or even 1.57, as it would cover both the JS output and generally more extensive MSRV checks.

Yeah that's probably the better *and* easier way!

PR #4642, opened 2025-09-15
Inline __externref_heap_live_count

RReverser:
Instead of declaring it as a C export in the externref runtime module, having the externref transform search for it and the "proper" `__wbindgen_externref_heap_live_count` intrinsic, and having the transform replace one with another, we can just use a `cfg(target_feature = "reference-types")` to choose the right implementation in the first place.

This allows to remove quite a bit of no longer used transitive code for such a small function.

### DIFF
--- a/src/externref.rs
+++ b/src/externref.rs
@@ -13,21 +13,14 @@ externs! {
     }
 }
 
-pub struct Slab {
+#[derive(Default)]
+struct Slab {
     data: Vec<usize>,
     head: usize,
     base: usize,
 }
 
 impl Slab {
-    fn new() -> Slab {
-        Slab {
-            data: Vec::new(),
-            head: 0,
-            base: 0,
-        }
-    }
-
     fn alloc(&mut self) -> usize {
         let ret = self.head;
         if ret == self.data.len() {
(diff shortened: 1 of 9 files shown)


src/externref.rs
```diff
@@ -118,19 +111,18 @@ fn internal_error(_msg: &str) -> ! {
 // Management of `externref` is always thread local since an `externref` value
 // can't cross threads in wasm. Indices as a result are always thread-local.
 #[cfg_attr(target_feature = "atomics", thread_local)]
-static HEAP_SLAB: crate::__rt::LazyCell<Cell<Slab>> =
-    crate::__rt::LazyCell::new(|| Cell::new(Slab::new()));
+static HEAP_SLAB: crate::__rt::LazyCell<Cell<Slab>> = crate::__rt::LazyCell::new(Default::default);
```
Spxg, line 114: I think it is OK to use `UnsafeCell` here, because it is always thread local, we can avoid creating an empty `Slab` to replace it every time.

RReverser: Yeah I'm not touching slab's implementation details in this PR :) Besides, it's safer this way, and empty Slab is zero-cost.

daxpedda: Agreed, as long as `Slab::new()` is zero cost, we shouldn't even consider going `unsafe`.

Alternatively we could use `RefCell` here, but unfortunately that's not going to work without `thread_local!`, which we didn't use to support `no_std`.

daxpedda (approved):
That's a nice improvement!

src/externref.rs
Spxg: Whether it is "zero cost" also depends on the opt-level, so I did some tests: https://github.com/Spxg/about-wasm-bindgen/tree/master/cell_vs

I'm not a fanatic about performance, but it's hard to convince myself not to use `UnsafeCell` when it's safe and makes the code faster and clearer (to be honest, I have thought more than once why this code is written this way).

daxpedda: Whoa, I'm seriously confused why there is such a big difference here.

But the performance data is pretty clear, we should definitely use `UnsafeCell` here.

Spxg: Thanks! I will submit a PR later

PR #4644, opened 2025-09-15
Support --generate-reset-state for --target module

guybedford:
This adds a new `--generate-reset-state` flag which is supported for `--target module`.

When used, a special `__wbg_reset_state` function is exported which when called will reset all bindings and Wasm state to their initial states.

To avoid stale pointers from being used where JS objects remain, finalizers and pointer management functions all check an additional `instanceId` that is only set and used when this feature is enabled.

Memviews, heap objects, the stack pointer, decoded byte counts and vector state variables are all conditionally reset when they are used only.

When trying to pas a stale object to Wasm bindgen, a new stale object error will be thrown in that case.


guybedford:
Yes, typically exports and imports need to be wrapped here to handle the outer world state management as well.

I've renamed to `--experimental-generate-reset-state`.

Agreed enabling exception handling would be the better path forward.

crates/cli-support/src/js/mod.rs
```diff
@@ -2485,11 +2514,26 @@ wasm = wasmInstance.exports;
         // while we invoke it. If we finish and the closure wasn't
         // destroyed, then we put back the pointer so a future
         // invocation can succeed.
+        let dtor_call = format!("wasm.{table}.get(state.dtor)(a, state.b);");
```
daxpedda: I don't see this being used more then once.
Why did we extract it?

crates/cli-support/src/js/mod.rs
```diff
@@ -2531,11 +2575,26 @@ wasm = wasmInstance.exports;
         // executing the destructor, however, we clear out the
         // `this.a` pointer to prevent it being used again the
         // future.
+        let dtor_call = format!("wasm.{table}.get(state.dtor)(state.a, state.b); state.a = 0;");
```
daxpedda: Same as above.

crates/cli-support/src/lib.rs
```diff
@@ -89,6 +90,7 @@ impl Bindgen {
             env::var("WASM_BINDGEN_ANYREF").is_ok() || env::var("WASM_BINDGEN_EXTERNREF").is_ok();
         let multi_value = env::var("WASM_BINDGEN_MULTI_VALUE").is_ok();
         let symbol_dispose = env::var("WASM_BINDGEN_EXPERIMENTAL_SYMBOL_DISPOSE").is_ok();
+        let generate_reset_state = env::var("WASM_BINDGEN_GENERATE_RESET_STATE").is_ok();
```
daxpedda:
```suggestion
        let generate_reset_state = env::var("WASM_BINDGEN_EXPERIMENTAL_GENERATE_RESET_STATE").is_ok();
```

We have it in the CLI, do we really need an environment variable?
I would prefer to not add more.

guide/src/reference/cli.md
```diff
@@ -111,3 +111,29 @@ containing the JS shim.
 On the no-modules target, `link_to!` won't work if used outside of a document,
 e.g. inside a worker. This is because it's impossible to figure out what the
 URL of the linked module is without a reference point like `import.meta.url`.
+
+### `--generate-reset-state`
```
daxpedda:
```suggestion
### `--experimental-generate-reset-state`
```

guide/src/reference/cli.md
```diff
@@ -111,3 +111,29 @@ containing the JS shim.
...
+
+Generates a `__wbg_reset_state()` function that allows reinitializing the
+entire library state for environments that wish to reuse and reset the same
+JavaScript execution context without reloading the entire library.
+
+This feature is currently only supported for `--target module`.
+
+When this flag is enabled, the generated code will also associate all objects
+with execution identity symbols to then validate and throws for stale references,
```
daxpedda:
```suggestion
with execution identity symbols that validate and throw for stale references,
```
If I'm getting this right.

crates/cli-support/src/js/mod.rs
guybedford: Thanks, fixed.

crates/cli-support/src/lib.rs
guybedford: Removed.

guide/src/reference/cli.md
guybedford: Added.

guide/src/reference/cli.md
guybedford: I actually renamed this now to `--experimental-generate-reset-state-function`.

PR #4645, opened 2025-09-15
Remove `JsObject`

RReverser:
This trait already has identical semantics to `JsCast` (except for the `WasmDescribe` bound, which we can easily add to relevant `impl`s), and, unlike some, it's `doc(hidden)` so we can safely remove it.


src/cast.rs
```diff
@@ -85,12 +85,6 @@ where
     /// Performs a zero-cost unchecked cast into a reference to the specified
     /// type.
     ///
-    /// This method will convert the `self` value to the type `T`, where both
-    /// `self` and `T` are simple wrappers around `JsValue`. This method **does
-    /// not check whether `self` is an instance of `T`**. If used incorrectly
-    /// then this method may cause runtime exceptions in both Rust and JS, this
-    /// should be used with caution.
```
daxpedda: Why would this not hold anymore?

src/cast.rs
```diff
@@ -12,7 +12,7 @@ use crate::{describe::WasmDescribe, JsValue};
 /// `#[wasm_bindgen]` `extern` block.
 ///
 /// [rfc]: https://github.com/rustwasm/rfcs/blob/master/text/002-wasm-bindgen-inheritance-casting.md
-pub trait JsCast
+pub trait JsCast: WasmDescribe + AsRef<JsValue> + Into<JsValue>
```
daxpedda: That is a problem.

In fact, I believe that there are weird libraries out there that make use of custom implementations of `JsCast`.

I will search for them in the meantime.

daxpedda: Well, [there are plenty](https://github.com/search?q=%22JsCast+for%22+language%3ARust&type=code&l=Rust).

Even [`workers-rs` makes use of it](https://github.com/cloudflare/workers-rs/blob/121cb1c088ae003a14c3cd4ce542c70fb28eaa2c/worker/src/r2/mod.rs#L152).

It was definitely a mistake not to seal traits like those without the clear intent to let users implement them.

src/cast.rs
RReverser: Woops, I meant to move this up to trait docs, but then decided against it and forgot to put this back in.

src/cast.rs
RReverser: Oh ffs, I always thought these are only used but not implemented.

Okay, I can just change corresponding impls to `T: JsCast + WasmDescribe` instead.

RReverser: Done.

PR #4647, opened 2025-09-15
Optimizing `HEAP_SLAB` calls using `UnsafeCell`

Spxg:
Discussion: https://github.com/wasm-bindgen/wasm-bindgen/pull/4642#discussion_r2349469794

### DIFF
--- a/src/externref.rs
+++ b/src/externref.rs
@@ -2,7 +2,7 @@ use crate::JsValue;
 
 use alloc::slice;
 use alloc::vec::Vec;
-use core::cell::Cell;
+use core::cell::UnsafeCell;
 use core::cmp::max;
 
 externs! {
@@ -111,13 +111,13 @@ fn internal_error(_msg: &str) -> ! {
 // Management of `externref` is always thread local since an `externref` value
 // can't cross threads in wasm. Indices as a result are always thread-local.
 #[cfg_attr(target_feature = "atomics", thread_local)]
-static HEAP_SLAB: crate::__rt::LazyCell<Cell<Slab>> = crate::__rt::LazyCell::new(Default::default);
+static HEAP_SLAB: crate::__rt::LazyCell<UnsafeCell<Slab>> =
+    crate::__rt::LazyCell::new(Default::default);
 
 fn with_slab<R>(f: impl FnOnce(&mut Slab) -> R) -> R {
-    let mut slot = HEAP_SLAB.take();
-    let ret = f(&mut slot);
-    HEAP_SLAB.replace(slot);
-    ret
+    // Safety: `externref` is always thread local
+    let slab = unsafe { &mut *HEAP_SLAB.get() };
+    f(slab)
 }
 
 #[no_mangle]


RReverser (changes requested):
How did you run benchmarks? If you had the browser console opened, keep in mind that disables Wasm compilation and it's executed in interpreted mode, which won't be reflective of real-world performance.

Also I'd recommend using something like Benchmark.js for proper warmup etc, as currently the code doesn't do any of that and it's a common issue with "manual" JS microbenchmarks.

I'm not strictly opposed to this change but I want to make sure we're measuring the right thing first, as getting benchmarks right in JS is tricky.

RReverser:
It's also worth noting that you didn't divide by the actual number of times (another factor of 100_000) so the absolute timings are at the order of `avg 0.000023545 ms` which is much closer to the noise margin.

RReverser:
I tried doing at least minimal warmup locally (but still without the statistical benchmarking Benchmark.js offers) + keeping console closed + dividing by total number and the difference I'm getting is

0.00000889800000011921ms

vs

0.000006668000000119209ms

that is, 8 nanoseconds vs 6 nanoseconds. A difference is still there, but it's so tiny I'm not too sure about it.

RReverser:
And on other reruns I'm getting 0.000008508999999761582ms even with unsafe-cell, which strongly suggest this is the scale of measurement margin.

Spxg:
> How did you run benchmarks? If you had the browser console opened, keep in mind that disables Wasm compilation and it's executed in interpreted mode, which won't be reflective of real-world performance.

Oh! I forgot about that, I'll retest.

> Also I'd recommend using something like Benchmark.js for proper warmup etc, as currently the code doesn't do any of that and it's a common issue with "manual" JS microbenchmarks.

I will check.

> It's also worth noting that you didn't divide by the actual number of times (another factor of 100_000)

Actually, I don't expect the accuracy of the preformance. Is there anything that can run natively? Maybe nodejs can do it, but I don't know much about it.

RReverser:
> Actually, I don't expect the accuracy of the preformance. Is there anything that can run natively? Maybe nodejs can do it, but I don't know much about it.

I'd definitely suggest benchmark.js + yeah, you can build for Node.js target. That's what I'm doing in serde-wasm-bindgen: https://github.com/RReverser/serde-wasm-bindgen/tree/main/benchmarks

Maybe there's finally way to use either Criterion or Divan with wasm-bindgen, it's been a while, but I don't know.

Spxg:
I might have made some progress (opt-level = 0):
`Cell`: `hz: 22620362`
`UnsafeCell`: `hz: 28655192`

I will go into more detail later

daxpedda:
Would also be interesting to see benchmarks with `RefCell`, to avoid `unsafe`.

Spxg:
I got stable results using nodejs benchmark: https://github.com/Spxg/about-wasm-bindgen/tree/master/cell_vs

||Cell|RefCell|UnsafeCell|
|-|-|-|-|
|opt-level=0|22,113,356 ops/sec ±0.23%|24,878,384 ops/sec ±0.44%|35,012,090 ops/sec ±0.41%|
|opt-level=1|70,824,816 ops/sec ±0.45%|72,064,993 ops/sec ±0.41%|76,436,707 ops/sec ±0.35%|
|opt-level=2|68,412,049 ops/sec ±0.28%|81,423,239 ops/sec ±0.60%|85,974,886 ops/sec ±0.37%|
|opt-level=3|70,689,223 ops/sec ±0.55%|83,553,134 ops/sec ±0.60%|84,443,275 ops/sec ±0.49%|
|opt-level="s"|57,638,158 ops/sec ±0.43%|73,181,241 ops/sec ±0.48%|80,080,127 ops/sec ±0.36%|
|opt-level="z"|58,498,860 ops/sec ±0.28%|74,129,376 ops/sec ±0.25%|75,057,844 ops/sec ±0.20%|

wasm-bindgen's externref bench (see `externref_bench`):

||Cell|RefCell|UnsafeCell|Improvement from Cell to UnsafeCell|
|-|-|-|-|-|
|opt-level=0|4,542,025 ops/sec ±0.37%|5,095,234 ops/sec ±0.30%|6,452,675 ops/sec ±0.41%|42%|
|opt-level=1|17,899,143 ops/sec ±0.44%|19,749,848 ops/sec ±0.37%|20,226,752 ops/sec ±0.17%|13%|
|opt-level=2|18,842,926 ops/sec ±0.28%|21,451,768 ops/sec ±0.40%|22,055,264 ops/sec ±0.38%|17%|
|opt-level=3|19,284,668 ops/sec ±0.44%|21,819,526 ops/sec ±0.37%|21,858,031 ops/sec ±0.51%|13%|
|opt-level="s"|15,597,994 ops/sec ±0.44%|19,656,787 ops/sec ±0.31%|20,452,012 ops/sec ±0.43%|31%|
|opt-level="z"|14,377,145 ops/sec ±0.24%|17,786,003 ops/sec ±0.26%|18,637,259 ops/sec ±0.18%|30%|

daxpedda:
Just converted the information from above to operation time and only `RefCell`.

| |RefCell|UnsafeCell|Improvement from `RefCell` to `UnsafeCell`|
|-|-|-|-|
|opt-level=3|46ns|46ns|0.2%|
|opt-level="s"|51ns|49ns|4%|
|opt-level="z"|56ns|54ns|4.8%|

Just looking at individual operations the difference seems irrelevant. But looking at it from the lens of throughput 5% is quite a lot.

I'm still in favor of going with `RefCell` here. My assumption here is that in practice `opt-level`'s like `s` and `z` experience serious performance degradation over many code paths in comparison to `3`. So if we are seeing *5%* in something like *this*, I assume the rest of their application is gonna be slowed down to such an extent (in comparison) that this isn't going to matter a whole lot.

In `opt-level=3`, the difference is negligible, so I'm happy to go with `RefCell`.

EDIT: totally open to be proven wrong here, I almost feel like I'm grasping at straws here to avoid `unsafe`.

RReverser:
Worth also noting that with either `RefCell` or `UnsafeCell` we can also remove the lazy wrapper and instead just have a newtype for `Send+Sync` impls.

`RefCell::new(Slab { ... })` is a const expression so can be put into static directly and forgo the lazy checks on each access.

daxpedda:
Now that I think about it again, I don't know what we are actually doing here.
We can actually just get a mutable reference anytime we want here, we have access anyway.

But yeah, lets just use wrappers like you suggest.

RReverser:
> We can actually just get a mutable reference anytime we want here, we have access anyway.

You'd need `static mut` for that, which is... strongly discouraged. I do think in our scenario (we know it's always single-threaded access) it's safe, but the official recommendation is to use alternatives like `UnsafeCell` / `SyncUnsafeCell` 🤷‍♂️

daxpedda:
Right, was just playing around with it and realized the same.
Will make a PR shortly with what I believe might be a safe abstraction.

EDIT: nope ... will put up some improvement anyway.

RReverser:
Can you take a look at low-hanging fruit in https://github.com/wasm-bindgen/wasm-bindgen/pull/4649 meanwhile please?

Spxg:
> EDIT: totally open to be proven wrong here, I almost feel like I'm grasping at straws here to avoid `unsafe`.

I still think that using unsafe code here is not a problem. After all, we are only implementing `Sync` without the atomic feature and use `thread_local` with the atomic feature. 

If there is a problem, it will cause compilation to fail. Unless someone later implements `Sync` unconditionally and removes the `thread_local` attribute both.

But considering that this possibility still exists, using `RefCell` is acceptable and except for `opt-level=0`, its performance is not much different from `UnsafeCell`.

daxpedda:
I think there is generally a culture in Rust to avoid `unsafe` as much as possible.

Personally, I had plenty of experience with thinking everything's fine with a single contained `unsafe` line just for it turning out not to be. So I understand where the culture is coming from.

But yeah, we can always get back to it if we think its worth it for our users.

daxpedda:
Closing in favor of #4650.

PR #4648, opened 2025-09-15
Cleaning up `Send`/`Sync` implementations for `LazyCell/Lock`

daxpedda:
Constraining the `unsafe` scope of `LazyCell`.

The way it `LazyCell` implements `Send + Sync` regardless of `target_feature = "atomics"` made it possible to unsafely use it in multi-threaded environments. This PR fixes that.

In the process I also removed the now unnecessary `Wrapper` type.

### DIFF
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -94,23 +94,23 @@ pub fn wbg_cast<From: IntoWasmAbi, To: FromWasmAbi>(value: From) -> To {
 }
 
 /// Wrapper around [`Lazy`] adding `Send + Sync` when `atomics` is not enabled.
-pub struct LazyCell<T, F = fn() -> T>(Wrapper<Lazy<T, F>>);
+pub struct LazyCell<T, F = fn() -> T>(Lazy<T, F>);
 
-struct Wrapper<T>(T);
+#[cfg(not(target_feature = "atomics"))]
+unsafe impl<T, F> Sync for LazyCell<T, F> {}
 
-unsafe impl<T> Sync for Wrapper<T> {}
-
-unsafe impl<T> Send for Wrapper<T> {}
+#[cfg(not(target_feature = "atomics"))]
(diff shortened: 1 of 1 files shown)


src/rt/mod.rs
```diff
@@ -128,7 +128,7 @@ pub use LazyCell as LazyLock;
 #[cfg(target_feature = "atomics")]
 pub struct LazyLock<T, F = fn() -> T> {
```
RReverser, line 129: We could just replace this with once_cell::sync::Lazy instead, right?

(or even std LazyLock if we assume  that atomics are always used with modern nightly Rust)

daxpedda: Its a bit complicated.

`once_cell` does blocking, which would panic in our case. To prevent this, we would have to implement `critical-section`. Because that is just all-in-all less then an ideal, we went with this solution.

See https://github.com/wasm-bindgen/wasm-bindgen/pull/4322.

RReverser: Interesting, thanks.

RReverser: Do you know if std LazyLock has the same issue? Looks like it's implemented via Once, which has couple of OnceState backends.

I haven't checked which one is used by Wasm threads build, but eg this backend looks promising, seems like pure atomics. https://github.com/rust-lang/rust/blob/f3fd3efe4f698ad9dc2ccd6b46a3b07e1bc911da/library/std/src/sys/sync/once/queue.rs#L71

daxpedda: No that would work. But unfortunately its not available in `no_std`.

daxpedda: Note: it does use `thread::park()`, but luckily that doesn't panic on Wasm, its just no-op.

RReverser: > But unfortunately its not available in `no_std`.

Ah :(

PR #4649, opened 2025-09-16
Use simple atomic for GLOBAL_EXNDATA

RReverser:
Exception `JsValue` index should never be zero (I think in general `JsValue` can use `NonZeroU32`, although I'm less sure and it'd be a bigger change), so we can reduce storage to just `u32`, and then we can also use `AtomicU32` which is `Send+Sync` on its own instead of the `LazyCell<Cell<>>` construct.

### DIFF
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -1,17 +1,17 @@
 use crate::convert::{FromWasmAbi, IntoWasmAbi, WasmAbi, WasmRet};
 use crate::describe::inform;
 use crate::JsValue;
+use alloc::alloc::{alloc, dealloc, realloc, Layout};
+use alloc::rc::Rc;
 use core::borrow::{Borrow, BorrowMut};
 use core::cell::{Cell, UnsafeCell};
 use core::convert::Infallible;
 use core::ops::{Deref, DerefMut};
 #[cfg(target_feature = "atomics")]
-use core::sync::atomic::{AtomicU8, Ordering};
-use wasm_bindgen_shared::tys::FUNCTION;
-
-use alloc::alloc::{alloc, dealloc, realloc, Layout};
-use alloc::rc::Rc;
+use core::sync::atomic::AtomicU8;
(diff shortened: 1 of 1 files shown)


daxpedda (changes requested):
> Exception `JsValue` index should never be zero, [..]

Remind me why?
> [..] and then we can also use `AtomicU32` which is `Send+Sync` on its own instead of the `LazyCell<Cell<>>` construct.

Addressed in #4650. Atomics aren't cheap in Wasm because we are stuck with sequential ordering.

RReverser:
> Remind me why?

I was going to say we have dedicated index for `undefined`, and JS "backend" reserves index 0 to represent `None` internally in some places, but actually I'm less sure whether externref does the same. I'm guessing not since I don't see anything like that in slab. Maybe it does need u64 atomic after all.

> Addressed in #4650. Atomics aren't cheap in Wasm because we are stuck with sequential ordering.

Hm I do think we'd need that wrapper anyway, but one nice thing about atomics is that they compile away to simple load/store on non-thread build safely, so we can avoid even `RefCell` checks without having to use unsafe APIs like `UnsafeCell`.

daxpedda:
> > Addressed in #4650. Atomics aren't cheap in Wasm because we are stuck with sequential ordering.
> 
> Hm I do think we'd need that wrapper anyway, but one nice thing about atomics is that they compile away to simple load/store on non-thread build safely, so we can avoid even `RefCell` checks without having to use unsafe APIs like `UnsafeCell`.

We are using `Cell` here however. But I would also guess that atomics are still easier to optimize then `Cell` in this case.

RReverser:
> We are using `Cell` here however. But I would also guess that atomics are still easier to optimize then `Cell` in this case.

I meant if we replace with `RefCell`. Otherwise `Cell` is probably comparable.

RReverser:
Anyway, I'll close this for now, might reopen with AtomicU64 depending on how other PRs go.

PR #4650, opened 2025-09-16
Re-introduce `ThreadLocalWrapper`

daxpedda:
This is partly reverting #4648 and using `Wrapper` (now `ThreadLocalWrapper`) in one more place.

We also now use `RefCell` instead of `Cell` to properly get mutable access instead of the shenanigans we had before.
@Spxg correctly identified that this caused measurable performance issues in https://github.com/wasm-bindgen/wasm-bindgen/pull/4642#discussion_r2349469794 and #4647.

Replaces #4647.

### DIFF
--- a/src/externref.rs
+++ b/src/externref.rs
@@ -2,7 +2,7 @@ use crate::JsValue;
 
 use alloc::slice;
 use alloc::vec::Vec;
-use core::cell::Cell;
+use core::cell::RefCell;
 use core::cmp::max;
 
 externs! {
@@ -13,7 +13,6 @@ externs! {
     }
 }
 
-#[derive(Default)]
 struct Slab {
     data: Vec<usize>,
     head: usize,
@@ -21,6 +20,14 @@ struct Slab {
 }
 
 impl Slab {
+    const fn new() -> Self {
+        Self {
+            data: Vec::new(),
+            head: 0,
+            base: 0,
+        }
+    }
+
     fn alloc(&mut self) -> usize {
         let ret = self.head;
         if ret == self.data.len() {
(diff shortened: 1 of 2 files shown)


src/rt/mod.rs
```diff
@@ -93,32 +93,34 @@ pub fn wbg_cast<From: IntoWasmAbi, To: FromWasmAbi>(value: From) -> To {
     unsafe { To::from_abi(breaks_if_inlined::<From, To>(prim1, prim2, prim3, prim4).join()) }
 }
 
-/// Wrapper around [`Lazy`] adding `Send + Sync` when `atomics` is not enabled.
-pub struct LazyCell<T, F = fn() -> T>(Lazy<T, F>);
+pub(crate) struct ThreadLocalWrapper<T>(pub(crate) T);
 
 #[cfg(not(target_feature = "atomics"))]
-unsafe impl<T, F> Sync for LazyCell<T, F> {}
+unsafe impl<T> Sync for ThreadLocalWrapper<T> {}
```
RReverser, line 99: I think we might want to hide it inside our own `thread_local` macro.

After all, it's `Send+Sync` _only_ if used in a variable that has `#[cfg_attr(target_feature = "atomics", thread_local)]`, and otherwise there's a risk of misuse.

RReverser: Actually, can't we use `__wbindgen_thread_local` here?

daxpedda: Well, the point of #4648 was supposed to make it misuse resistant by not implementing `Send + Sync` except if we are on `target_feature = "atomics"`.

But yeah, a macro would definitely be nice.

daxpedda: > Actually, can't we use `__wbindgen_thread_local` here?

Could be re-purposed? I don't think we want `JsThreadLocal` in the way, but didn't look too closely right now.

RReverser: I understand, but I mean on atomics target if someone just puts `ThreadLocalWrapper` on a variable that does not have `#[thread_local]` then it's not really thread-safe at all.

daxpedda: Well, it wouldn't compile either because its missing `Send + Sync`.
Or am I missing something here?

RReverser: But `ThreadLocalWrapper` would still have `Send + Sync` on `atomics` target, even if it's put in a place where it's not thread-safe.

daxpedda: No, it specifically says:
```rust
#[cfg(not(target_feature = "atomics"))]
unsafe impl<T> Send for ThreadLocalWrapper<T> {}
```

Otherwise it only implements `Send` if `T: Send`, in which case its okay.

RReverser: Oh right, I was thinking of `LazyLock` , should've just looked below🤦🏼‍♂️

src/externref.rs
```diff
@@ -111,18 +118,12 @@ fn internal_error(_msg: &str) -> ! {
...
 #[cfg_attr(target_feature = "atomics", thread_local)]
-static HEAP_SLAB: crate::__rt::LazyCell<Cell<Slab>> = crate::__rt::LazyCell::new(Default::default);
-
-fn with_slab<R>(f: impl FnOnce(&mut Slab) -> R) -> R {
-    let mut slot = HEAP_SLAB.take();
-    let ret = f(&mut slot);
-    HEAP_SLAB.replace(slot);
-    ret
-}
+static HEAP_SLAB: crate::__rt::ThreadLocalWrapper<RefCell<Slab>> =
```
RReverser, line 121: Heh I see @Spxg's point, it does feel pretty silly - especially now, with the descriptive name - to have to go through RefCell checks when we know it's a thread-local and we only borrow it to get/set the value... But better than nothing I guess.

RReverser (approved):

src/externref.rs
Spxg: I agree with you

Spxg (approved):

PR #4651, opened 2025-09-16
Prepare v0.2.102 release

daxpedda:


crates/shared/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub mod tys;
 // This gets changed whenever our schema changes.
 // At this time versions of wasm-bindgen and wasm-bindgen-cli are required to have the exact same
 // SCHEMA_VERSION in order to work together.
-pub const SCHEMA_VERSION: &str = "0.2.101";
+pub const SCHEMA_VERSION: &str = "0.2.102";
```
daxpedda, line 16: @RReverser we keep forgetting to bump this when we break compatibility. Ideally we should do this in the breaking PR instead of possibly forgetting it during the release.

guybedford (approved):

crates/shared/src/lib.rs
RReverser: I think we have a test that tells us when we need to bump it. I did have to bump it once when I changed schema but the test didn't trigger since?

RReverser: Or was the test just for the hash... If it is, then we probably want it to tell us to bump the version too.

daxpedda: AFAIK we have a test that checks the hash, which would trigger if we changed something in `wasm-bindgen-shared`. But we changed intrinsics and such outside of that crate, so it doesn't trigger.

PR #4656, opened 2025-09-17
Fix `__wbindgen_object_drop_ref` mapping

### LINKED ISSUE #4655: Panic at `args.len()` check in `ObjectDropRef` on wasm-bindgen-test v0.3.52
After updating wasm-bindgen-test to v0.3.52, the followig panic occurs.

```
Executing bindgen...
thread 'main' panicked at crates/cli-support/src/js/mod.rs:3952:17:
assertion `left == right` failed
  left: 0
 right: 1
```

The corresponding code is below

https://github.com/wasm-bindgen/wasm-bindgen/blob/8bfa10ae7b2e8c7507f1bb2bebd27ea13838b44c/crates/cli-support/src/js/mod.rs#L3951-L3954

The reproduce procedure is below

```
$ git clone https://github.com/veryl-lang/doc
$ cd doc/playground
$ wasm-pack test --node
```

I checked each commit between v0.3.51 and v0.3.52

* a30a65b2c3ab874e6c3f538e8a486253d0bbd07a and before it : it works fine
* 4e5fa176971c834e83f387064c3acfebf4d35930 to 0cc4c694f0d3b2a22f09f5f3021ed96186a5be69: the following error
```
Error: executing `wasm-bindgen` over the Wasm file

Caused by:
    import of `__wbindgen_describe_cast` doesn't have an adapter listed
error: test failed, to rerun pass `--lib`
```
* 8bfa10ae7b2e8c7507f1bb2bebd27ea13838b44c: the following error
```
Executing bindgen...
thread 'main' panicked at crates/cli-support/src/js/mod.rs:3952:17:
assertion `left == right` failed
  left: 0
 right: 1
```

daxpedda:
The intrinsics were no properly translated in #4636.

Fixes #4655.


Spxg (approved):

daxpedda:
I'm a bit disappointed that our integration tests don't properly test `wasm-bindgen-cli`, but just come with a bunch of custom stuff to produce the output. Its justified to the degree that we want reasonable WAT files to inspect by humans, but we really need to test the input with the actual way `wasm-bindgen-cli` works.

In any case, I'm still confused why so many tests compile with this bug. This should be triggered a lot more then it is.

daxpedda:
I will merge this and make a new release for now.
But would be nice if we can add a test to cover this later.

crates/cli-support/src/wit/mod.rs
```diff
@@ -137,13 +137,13 @@ impl<'a> Context<'a> {
         }
         self.add_aux_import_to_import_map(
             "__wbindgen_object_clone_ref",
-            vec![Descriptor::I32],
-            Descriptor::I32,
+            vec![Descriptor::Ref(Box::new(Descriptor::Externref))],
```
RReverser, line 140: Hm our original intrinsics were `u32` though, shouldn't externref pass be the one that changes them to `externref`?

Otherwise I'd think we're breaking the non-externref target.

RReverser:
> but we really need to test the input with the actual way `wasm-bindgen-cli` works

In theory that's what example tests are doing - full integration, with wasm-bindgen the CLI, the library, wasm-pack, complex code etc etc - like a proper integration.

The question is why they don't catch this.

crates/cli-support/src/wit/mod.rs
daxpedda: I'm really unsure of the details here. But the original code is:
https://github.com/wasm-bindgen/wasm-bindgen/blob/3046b1e7591a4568211f3fa6acaf3747a2d80412/crates/cli-support/src/intrinsic.rs#L180-L185

Did I miss anything here?

daxpedda:
> > but we really need to test the input with the actual way `wasm-bindgen-cli` works
> 
> In theory that's what example tests are doing - full integration, with wasm-bindgen the CLI, the library, wasm-pack, complex code etc etc - like a proper integration.

I think at least the majority of our tests should go through `wasm-bindgen-cli/support`. I'm concerned that these super specialized setups are giving us a false sense of security.

> The question is why they don't catch this.

I don't exactly understand why, but apparently this piece of code doesn't get triggered in any of our tests:
https://github.com/wasm-bindgen/wasm-bindgen/blob/6832c34d18ce3cd841f3f87f7e5b6417b47ca02b/crates/cli-support/src/js/mod.rs#L3951-L3954

We really need a reproducible. https://github.com/veryl-lang/doc is a bit too large to boil it down. But we could dig through the generated Wasm and see where it ends up being generated.

crates/cli-support/src/wit/mod.rs
RReverser: Hm probably you're right and I got confused, I guess I thought this has to be the raw signature here to match the original `fn __wbindgen_object_drop_ref(idx: u32) -> ();`.

I still don't get how integration tests haven't caught this, surely they use `__wbindgen_object_drop_ref` a lot.

RReverser:
> I don't exactly understand why, but apparently this piece of code doesn't get triggered in any of our tests:

That's so weird. I'd think it's one of the most popular intrinsics :/ Alright...

crates/cli-support/src/wit/mod.rs
RReverser: Nvm, saw the discussion below, let's continue there.

RReverser:
That intrinsic should be used whenever `JsValue` (or other imported type that wraps it) is dropped. Maybe none of our integration tests take `JsValue` by value?

RReverser:
> I think at least the majority of our tests should go through `wasm-bindgen-cli/support`. I'm concerned that these super specialized setups are giving us a false sense of security.

I thought about that before, but that would be the move _away_ from full integration though. The current `examples` setup tests workflow exactly the way our users would do it, but what you're suggesting would rely on `wasm-bindgen-cli-support` directly instead of having e.g. CLI itself in the pipeline.

daxpedda:
> That intrinsic should be used whenever `JsValue` (or other imported type that wraps it) is dropped. Maybe none of our integration tests take `JsValue` by value?

Yeah that's what I thought as well, but there are plenty of cases where we drop `JsValue`s inside Rust in our tests.

daxpedda:
> I thought about that before, but that would be the move _away_ from full integration though. The current `examples` setup tests workflow exactly the way our users would do it, but what you're suggesting would rely on `wasm-bindgen-cli-support` directly instead of having e.g. CLI itself in the pipeline.

No no. I used `/` as a means to say either is fine, as long as it actually mimics what `wasm-bindgen-cli` does. But yes, using `wasm-bindgen-cli` directly is the full integration test. Tests that do use that, should not be moved to `wasm-bindgen-cli-support`.

RReverser:
It looks like it's one of those intrinsics that gets replaced by the `externref` transform with an internal thingy instead of an import. Was the failure on pre-Rust 1.82 by any chance? At least that would explain it.

daxpedda:
No, it was with a recent Rust version.

RReverser:
> No no. I used `/` as a means to say either is fine, as long as it actually mimics what `wasm-bindgen-cli` does. But yes, using `wasm-bindgen-cli` directly is the full integration test. Tests that do use that, should not be moved to `wasm-bindgen-cli-support`.

Oh yeah then `examples` tests is exactly that already, we should just keep adding more.

RReverser:
> No, it was with a recent Rust version.

Then I have no idea :/

RReverser:
I wonder if it's just because of version mismatch then. Because that import really shouldn't be preserved in JS at all after the `externref` transform, and that path should only trigger if the import wasn't replaced with an externref intrinsic, so the only two explanations are either old Rust version (w/o reference-types) or the version mismatch causing the `externref` transform to not replace the intrinsic correctly.

daxpedda:
You should check with https://github.com/veryl-lang/doc. Pretty easy to see its not a version mismatch.

RReverser:
Hm but in the original report in https://github.com/wasm-bindgen/wasm-bindgen/issues/4655 you said it was version mismatch? (I mean CLI transforms vs wasm-bindgen library)

daxpedda:
No. The OP was trying to figure out which commit caused the regression. In that process, all the errors they found are probably version mismatch errors. But the original error is definitely reproducible with the current version and doesn't appear with the previous one.

RReverser:
Ok so even if that's not the reason in the OP case (which is still weird to me), I can at least reliably reproduce same issue by building a small example with eg Rust 1.73.

We should really add that MSRV branch to CI testing, but that's currently blocked on splitting up workspaces (https://github.com/wasm-bindgen/wasm-bindgen/pull/4660#issuecomment-3302638884) which I don't have time to do.

PR #4660, opened 2025-09-17
Try to run at least integration tests with 1.56

RReverser:


RReverser:
Ah right, because of shared workspace it's not as trivial as I had hoped.

Another reason to split into two workspaces for host & wasm as discussed earlier.

PR #4663, opened 2025-09-18
Fix the mismatch between definition and access of `encode_into`

Spxg:
Implementing `ValueEnum` for `EncodeInto` also can fix, but it requires importing `clap` crate into `cli-support`, which is a bit heavy.

resolves https://github.com/wasm-bindgen/wasm-bindgen/issues/4662

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 * Fixed incorrect function mapping during post-processing.
   [#4656](https://github.com/wasm-bindgen/wasm-bindgen/pull/4656)
 
+* Fixed wasm-bindgen-cli's `encode_into` argument not working.
+  [#4663](https://github.com/wasm-bindgen/wasm-bindgen/pull/4663)
+
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


Spxg:
The disadvantage of using string is that if you add an string option but don't add a match, an error will be reported at runtime. However, I think this is acceptable, after all, after adding an option, it is likely to be tested.

Or add a shell with the same definition as `EncodeInto` in `wasm-bindgen.rs`, and implement `From` trait both.

RReverser:
What the hell :/ Why doesn't clap report this statically...

RReverser:
> but it requires importing `clap` crate into `cli-support`, which is a bit heavy.

I think you can instead declare a local `enum` implementing `ValueEnum` and then we convert one to the other.

Spxg:
> I think you can instead declare a local enum implementing ValueEnum and then we convert one to the other.

Of course, as I said above:

> Or add a shell with the same definition as `EncodeInto` in `wasm-bindgen.rs`

The problem here is that we also need to make them implement `From` trait on each other. In this way, no matter which side adds a variant, it will require the other side to have a corresponding variant. However, this seems to me to be a bit redundant code. Otherwise, there are similar problems as using string.

PR #4666, opened 2025-09-18
Fix Symbol.dispose override, use Symbol.dispose by default

guybedford:
Follow-on to https://github.com/wasm-bindgen/wasm-bindgen/pull/4514, resolves https://github.com/wasm-bindgen/wasm-bindgen/issues/4513.

To avoid overwriting `Symbol.dispose` when unavailable we instead use `Symbol.for('wbg_symbol_dispose')` as the symbol instead.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1230,16 +1230,15 @@ wasm = wasmInstance.exports;
             wasm_bindgen_shared::free_function(name),
         ));
         ts_dst.push_str("  free(): void;\n");
-        if self.config.symbol_dispose {
-            dst.push_str(
-                "
-                [Symbol.dispose]() {{
-                    this.free();
-                }}
-                ",
-            );
-            ts_dst.push_str("  [Symbol.dispose](): void;\n");
-        }
+        self.expose_symbol_dispose()?;
+        dst.push_str(
+            "
+            [symbolDispose]() {{
+                this.free();
+            }}
+            ",
+        );
+        ts_dst.push_str("  [symbolDispose](): void;\n");
         dst.push_str(&class.contents);
         ts_dst.push_str(&class.typescript);
 
@@ -1735,7 +1734,7 @@ wasm = wasmInstance.exports;
         if !self.should_write_global("symbol_dispose") {
             return Ok(());
         }
-        self.global("if(!Symbol.dispose) { Symbol.dispose = Symbol('Symbol.dispose'); }");
+        self.global("const symbolDispose = Symbol.dispose || Symbol('Symbol.dispose');");
         Ok(())
     }
 
@@ -2937,11 +2936,6 @@ wasm = wasmInstance.exports;
 
     pub fn generate(&mut self) -> Result<(), Error> {
         self.prestore_global_import_identifiers()?;
-        // conditionally override Symbol.dispose
-        if self.config.symbol_dispose && !self.aux.structs.is_empty() {
-            self.expose_symbol_dispose()?;
-        }
-
         for (id, adapter, kind) in iter_adapeter(self.aux, self.wit, self.module) {
             let instrs = match &adapter.kind {
                 AdapterKind::Import { .. } => continue,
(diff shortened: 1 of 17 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -1735,7 +1734,7 @@ wasm = wasmInstance.exports;
         if !self.should_write_global("symbol_dispose") {
             return Ok(());
         }
-        self.global("if(!Symbol.dispose) { Symbol.dispose = Symbol('Symbol.dispose'); }");
+        self.global("const symbolDispose = Symbol.dispose || Symbol.for('wbg_symbol_dispose');");
```
daxpedda: This is what I was confused about: why do we need this in the first place?
We should be able to just use:
```js
    [Symbol.dispose]() {
        this.free();
    }
```
Regardless of support. It shouldn't trigger an error and we aren't overwriting anything either.

I'm suggesting:

```suggestion
```

daxpedda (changes requested):

crates/cli-support/src/js/mod.rs
guybedford: We wouldn't be able to make it the default since `Symbol.dispose` doesn't exist in Safari and would be `undefined` defining `MyClass.prototype.undefiend = ....`. https://caniuse.com/mdn-javascript_builtins_symbol_dispose

daxpedda: What do you mean with the "default" here? We have to keep `free()` anyway, so that's the default (if that is what you meant).

guybedford: I mean defaulting to enabling the Symbol.dispose output, whereas right now it is still flagged.

daxpedda: So whats wrong with `MyClass.prototype.undefined = ...`? From me trying it out it seems to not cause any errors.

guybedford: `'undefined'` is a perfectly valid string, nothing wrong with that. But I would not recommend it as an API.  There are perfectly other good strings to use of course. Or other symbols, like `Symbol.for('Symbol.dispose')` as we do here.

daxpedda: Oh ... I get it now. I thought in this case it would just not set anything at all. Instead it actually sets it on a field called `undefined`. That explains why it doesn't error.

What do you think of just defining it conditionally then?
```js
static {
    if (Symbol.dispose) {
        this.prototype[Symbol.dispose] = function() {
            this.free();
        };
    }
}
```

guybedford: We can do that as well, but then it's also not possible to call it in that case if users really wanted to in browsers that don't support the symbol.

daxpedda: Well, `free()` is still available and we can't remove it anyway until the next breaking change. So I don't think there is a reason to introduce another fallback?

guybedford: Yeah that makes sense sure.

guybedford: Added in the latest commit.

crates/cli-support/src/js/mod.rs
```diff
@@ -1248,6 +1239,12 @@ wasm = wasmInstance.exports;
         dst.push('}');
         ts_dst.push_str("}\n");
 
+        dst.push_str(&format!(
+            "
+                if (Symbol.dispose) {name}.prototype[Symbol.dispose] = {name}.prototype.free;
+            "
+        ));
```
daxpedda, line 1246: Why not put this inside the class definition with the help of `static { ... }`?

daxpedda (approved):
I think I mentioned this in the other PR, but should we update some of our code in the guide to use `using`?

crates/cli-support/src/js/mod.rs
guybedford: These still only have 92% browser support (https://caniuse.com/mdn-javascript_classes_static_initialization_blocks).

We can do it, but it seems like dropping multiple percentages of browser support for no reason?

daxpedda: Ah, good catch.

PR #4668, opened 2025-09-22
Switch release artifacts from `aarch64-unknown-linux-gnu` to `aarch64-unknown-linux-musl`

Spxg:


Spxg:
> Whats the reason for using Zig tools here?
> 
> Even if slightly more convenient, we should stick with the Rust tools if we can.

Actually, only the musl toolchain is needed, but musl.cc blocks connections from github. There may be a reliable mirror address, but I haven't found it.

daxpedda:
Is the `musl-tools` package not sufficient?

Spxg:
> Is the `musl-tools` package not sufficient?

Cross-compilation toolchain is required to compile the C library and specify the linker. Maybe I should try ubuntu-arm.

> musl.cc blocks connections from **github**

There is a detailed discussion here:
https://github.com/orgs/community/discussions/27906

Spxg:
> Maybe I should try ubuntu-arm.

Yes, it works perfectly, it just doesn't support using the `ubuntu-latest-arm` tag currently: https://github.com/orgs/community/discussions/148648#discussioncomment-11858287

.github/workflows/main.yml
```diff
@@ -407,6 +407,20 @@ jobs:
...
 
+  dist_linux_aarch64_musl:
+    runs-on: ubuntu-24.04-arm
+    steps:
+    - uses: actions/checkout@v5
+    - run: rustup update --no-self-update stable && rustup default stable
+    - run: rustup target add aarch64-unknown-linux-musl
+    - run: sudo apt update -y && sudo apt install musl-tools -y
+    - run: |
+        cargo build --manifest-path crates/cli/Cargo.toml --target aarch64-unknown-linux-musl --features vendored-openssl --release
```
daxpedda, line 418: Apparently we are stripping debug symbols from x86 MUSL for Linux as well ...
```suggestion
    - env:
        RUSTFLAGS: -C strip=debuginfo
      run: |
        cargo build --manifest-path crates/cli/Cargo.toml --target aarch64-unknown-linux-musl --features vendored-openssl --release
```

daxpedda (changes requested):
Small nit.

We should probably clean up the other releases to strip symbols there as well (in the same way).

.github/workflows/main.yml
Spxg: Release seems to have stripped debuginfo, which may be enabled by `debuginfo = true`, so the size is the same whether `strip="debuginfo"` is set or not.
https://doc.rust-lang.org/cargo/reference/profiles.html#debug

daxpedda: We still need to strip Std, which is pre-compiled and isn't affected by the profiles `debuginfo`.
Just tested it to double-check. While Std doesn't contain full debug information, it still contains some.
But to see the difference we actually have to max out the `strip` flag.
```suggestion
    - env:
        RUSTFLAGS: -C strip=symbols
      run: |
        cargo build --manifest-path crates/cli/Cargo.toml --target aarch64-unknown-linux-musl --features vendored-openssl --release
```

Spxg: > have to max out the `strip` flag.

Yes, I think we should keep symbols for panic backtraces.

daxpedda: That isn't the case. These symbols are only there for Std, but not for `wasm-bindgen`. If we want them for `wasm-bindgen` we would need to increase `debuginfo`.

Spxg: I am referring to the `RUST_BACKTRACE=1`. In the `strip=sybmols`, symbol info will be lost, making it difficult to troubleshoot the problem.

```
thread 'main' panicked at crates/cli-support/src/lib.rs:354:9:
explicit panic
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
vs
```                             
thread 'main' panicked at crates/cli-support/src/lib.rs:354:9:
explicit panic
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic_explicit
   3: wasm_bindgen_cli_support::Bindgen::generate_output::panic_cold_explicit
   4: wasm_bindgen_cli_support::Bindgen::generate_output
   5: wasm_bindgen_cli_support::Bindgen::generate
   6: wasm_bindgen_test_runner::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

daxpedda: I get that this is about the backtrace, but I thought the backtrace wouldn't work without symbols in `wasm-bindgen`, but apparently it does?

Spxg: > but apparently it does?
 
 It seems that just having the symbol is enough

jasonradford115-ui (approved):

PR #4669, opened 2025-09-22
Limit minimum glibc version for gnu artifacts

Spxg:


RReverser:
Do you mean making it the primary one? It's worth noting that musl is often chosen for its compactness, which is valuable in embedded environments and static builds, but it tends to be slower than glibc.

Maybe doesn't matter too much for our CLI but worth testing on some larger Wasm files.

daxpedda:
> Do you mean making it the primary one?

Yes.

> It's worth noting that musl is often chosen for its compactness, which is valuable in embedded environments and static builds, but it tends to be slower than glibc.
> 
> Maybe doesn't matter too much for our CLI but worth testing on some larger Wasm files.

We are using MUSL for x64 as well, anecdotally even in the biggest Wasm modules so far it just takes seconds.

In any case, I believe our CLI only minimally interacts with the filesystem and such, the biggest system interaction by far is going to be memory allocation. Not sure how much of a concern that is, because AFAIK we aren't multi-threading anything in the post-processor, so MUSLs allocator should be competitive.

RReverser:
> the biggest system interaction by far is going to be memory allocation

Yeah that's the thing, I think allocator in particular has been a problem in musl when I've read in the past. E.g. quick search in google:

> Both versions had the same number of syscalls. The main difference is the musl version spent 6.7 seconds in a futex while glibc only 0.5 seconds, a 13x penalty! This means there must be some contention for a shared lock in musl when allocating or de-allocating memory from multiple threads.

(c) [Default musl allocator considered harmful (to performance)](https://nickb.dev/blog/default-musl-allocator-considered-harmful-to-performance/)

Although as you said it sounds primarily related to multithreading, so maybe not too much of a concern but I'd still rather test before switching, and see if we might want to add mimalloc to the mix.

daxpedda:
> [..] but I'd still rather test before switching, and see if we might want to add mimalloc to the mix.

What I was mentioning earlier is that our x86 artifact is also *only* MUSL. I assume it won't differ much to ARM? So I think if we want to discuss providing more optimized binaries, we should do it for all of them.

RReverser:
> What I was mentioning earlier is that our x86 artifact is also _only_ MUSL.

Ah I misunderstood you. Yeah, in that case it's a bit out of scope.

Spxg:
By the way, is there any other suitable way to limit GLIBC's version?

daxpedda:
I'm not aware of one. Kinda one of the drivers for me to just move to MUSL and call it a day.

Spxg:
I think we need to decide whether to change the goal of this PR to remove gnu artifacts

daxpedda:
I think we should just keep the GNU artifact as long as `wasm-pack` still relies on it, but otherwise not change the current setup. Basically I'm in favor of closing the PR and otherwise not change anything.

PR #4670, opened 2025-09-22
Unconditionally use TextEncoder/TextDecoder

RReverser:
These have been available as globals even in Node.js for many years now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@
 * Fixed a bug in `--experimental-reset-state-function` support for heap reset.
   [#4665](https://github.com/wasm-bindgen/wasm-bindgen/pull/4665)
 
+### Changed
+
+* Unconditionally use `TextEncoder`/`TextDecoder` for string encoding/decoding.
+  [#TODO]
+
 --------------------------------------------------------------------------------
 
 ## [0.2.103](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.102...0.2.103)
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1506,9 +1506,7 @@ wasm = wasmInstance.exports;
         }
         self.expose_text_encoder()?;
 
-        // The first implementation we have for this is to use
-        // `TextEncoder#encode` which has been around for quite some time.
-        let encode = "function (arg, view) {
+        let polyfill_encode_into = "cachedTextEncoder.encodeInto = function (arg, view) {
             const buf = cachedTextEncoder.encode(arg);
             view.set(buf);
             return {
@@ -1517,41 +1515,24 @@ wasm = wasmInstance.exports;
             };
         }";
 
-        // Another possibility is to use `TextEncoder#encodeInto` which is much
-        // newer and isn't implemented everywhere yet. It's more efficient,
-        // however, because it allows us to elide an intermediate allocation.
-        let encode_into = "function (arg, view) {
-            return cachedTextEncoder.encodeInto(arg, view);
-        }";
-
-        // Looks like `encodeInto` doesn't currently work when the memory passed
+        // `encodeInto` doesn't currently work in any browsers when the memory passed
         // in is backed by a `SharedArrayBuffer`, so force usage of `encode` if
         // a `SharedArrayBuffer` is in use.
         let shared = self.module.memories.get(memory).shared;
 
         match self.config.encode_into {
-            EncodeInto::Always if !shared => {
-                self.global(&format!(
-                    "
-                    const encodeString = {encode_into};
-                "
-                ));
-            }
+            EncodeInto::Always if !shared => {}
             EncodeInto::Test if !shared => {
                 self.global(&format!(
                     "
-                    const encodeString = (typeof cachedTextEncoder.encodeInto === 'function'
-                        ? {encode_into}
-                        : {encode});
+                    if (!('encodeInto' in cachedTextEncoder)) {{
+                        {polyfill_encode_into}
+                    }}
                 "
                 ));
             }
             _ => {
-                self.global(&format!(
-                    "
-                    const encodeString = {encode};
-                "
-                ));
+                self.global(polyfill_encode_into);
             }
         }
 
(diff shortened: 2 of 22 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,11 @@
 * Fixed a bug in `--experimental-reset-state-function` support for heap reset.
   [#4665](https://github.com/wasm-bindgen/wasm-bindgen/pull/4665)
 
+### Changed
+
+* Unconditionally use `TextEncoder`/`TextDecoder` for string encoding/decoding.
```
daxpedda: Can we just mention the now minimum supported NodeJS version here instead?

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@
...
+  [#TODO]
```
daxpedda: ^

crates/cli-support/src/js/mod.rs
```diff
@@ -1732,7 +1713,7 @@ wasm = wasmInstance.exports;
         if !self.should_write_global("text_encoder") {
             return Ok(());
         }
-        self.expose_text_processor("const", "TextEncoder", "encode", "('utf-8')", None)
```
daxpedda, line 1735: Whats the story behind the `'utf-8'` parameter for this constructor? AFAICS the `TextEncoder` constructor doesn't even take a parameter.

daxpedda (changes requested):
Nice cleanup!

crates/cli-support/src/js/mod.rs
daxpedda: Okay, just a leftover from past refactoring: https://github.com/wasm-bindgen/wasm-bindgen/pull/1730.

CHANGELOG.md
RReverser: Sure

CHANGELOG.md
RReverser: Would be nice if there was some action that automatically fills in the PR number (since I don't know it at the time of commit)

RReverser:
Don't merge yet, I realised that the encodeInto polyfill will need to be adjusted / reassigned for the Safari bug workaround too (where we recreate the text encoder).

daxpedda:
I believe that's just the decoder, not encoder? AFAICS the output on the Safari workaround has adjusted correctly.

RReverser:
> I believe that's just the decoder, not encoder? AFAICS the output on the Safari workaround has adjusted correctly.

Oh you're right, it's just the decoder indeed. All good then.

kettle11:
@RReverser With this change I've started hitting errors "Uncaught ReferenceError: TextEncoder is not defined" when trying to use AudioWorklets. I found this while working on this PR to add an AudioWorklet based backend to cpal: https://github.com/RustAudio/cpal/pull/958

The simplest way to reproduce this is to run wasm-bindgen's wasm-audio-worklet example locally. For me that example hits the error locally in Chrome and Firefox but not in Safari. Prior to this change the code that checked if TextEncoder / Decoder exists before initializing them prevented this crash.

Puzzlingly the same example hosted on netlify does not hit this error, and I'm unsure as to why: https://wasm-bindgen.netlify.app/exbuild/wasm-audio-worklet/

RReverser:
> Puzzlingly the same example hosted on netlify does not hit this error, and I'm unsure as to why: https://wasm-bindgen.netlify.app/exbuild/wasm-audio-worklet/

It uses a simple polyfill to stub those out, which is likely something you should do in own audio worklets as well:

image

That said, I'm also puzzled that I can't find it in the source in the repo. It looks like a typical JS snippet referenced via `wasm_bindgen(module = "...")` but it's just not there.

I suspect we stopped publishing to Netlify sometime during org migration or maybe even further in the past, so that example shows older version of the code? @daxpedda

kettle11:
@RReverser Ah, I assumed the examples on netlify were up to date but that makes sense with the org shuffling. So that means the latest AudioWorklet example is hitting the same error I'm seeing.

The polyfill you're referencing uses a workaround. You can see it here when the AudioWorklet example was originally committed: https://github.com/wasm-bindgen/wasm-bindgen/blob/d881d9da64ae7b84b74f55b92f9dc0688367d013/examples/wasm-audio-worklet/src/wasm_audio.rs#L62-L68

Is going back to that the best approach? It'd be good to update the AudioWorklet example to reflect that.

RReverser:
> Is going back to that the best approach? It'd be good to update the AudioWorklet example to reflect that.

Yeah I guess we'll have to either put that back.

I wouldn't want to revert the PR, putting the shim that just throws errors just for the audio worklet would be pretty unfortunate in terms of output code size, given that wasm-bindgen promises to only output code for what you use...

It's a shame audio worklets are such a weird environment, keeps causing issues in both Emscripten and wasm-bindgen alike.

kettle11:
Personally I think it's tolerable if wasm-bindgen `AudioWorklet` users have to look at the example to learn about the weird shim, as long as it's still possible to make work and the workaround is documented. I can easily take that approach in the `cpal` PR I was working on.

Technically this was a breaking change, but I think me being the first to complain does indicate there aren't a ton of AudioWorklet users in the Rust ecosystem right now.

> It's a shame audio worklets are such a weird environment, keeps causing issues in both Emscripten and wasm-bindgen alike.

Agreed. I've never been a huge fan of how web APIs encourage "best practices" by disabling functionality. It leads to awkward situations like this one where nothing against best practices is actually being done but a bunch of effort is spent working around the situation anyways.

daxpedda:
> I suspect we stopped publishing to Netlify sometime during org migration or maybe even further in the past, so that example shows older version of the code? @daxpedda

Never touched or looked at that part, so no idea really.

> I wouldn't want to revert the PR, putting the shim that just throws errors just for the audio worklet would be pretty unfortunate in terms of output code size, given that wasm-bindgen promises to only output code for what you use...

I don't agree. I think we should support audio worklets with `wasm-bindgen`. It should be possible to use `TextEncoder` in other contexts but refrain from doing so in audio worklets, which is how it worked until now.

I completely forgot that this was the workaround when I approved the removal. But we can definitely decrease the output code. Another thing we could do is only emit this code when atomics is available, as without atomics you can and should just not use `TextEncoder` with audio worklets, in which case it wouldn't be present.

RReverser:
> Another thing we could do is only emit this code when atomics is available, as without atomics you can and should just not use `TextEncoder` with audio worklets, in which case it wouldn't be present.

FWIW what Emscripten does here is have a separate flag for `-s AUDIO_WORKLET`. Maybe we could just have that too?

Or we could even detect whether `AudioWorklet::add_module` is linked in, and only add the shim then.

PR #4671, opened 2025-09-22
Simplify Node.js & Deno loaders

RReverser:
- Leverage `WebAssembly.instantiateStreaming` in Deno.
- Leverage `file:` fetching which is supported in Deno since 2021.
- Use shorter but same `exports.foo` instead of `module.exports.foo` for Node.js (CommonJS).
- Replace platform-specific guessing in Node.js ESM with the built-in fs.readFile(url) support.

I'm a bit uncomfortable that we don't have integration tests for Node.js - I did run quick checks locally, but would appreciate another pair of eyes on the generated code.


daxpedda (commented):
LGTM! But somebody with actualy Deno/Node.js knowledge should approve this.

crates/cli-support/src/js/mod.rs
```diff
@@ -419,23 +401,9 @@ impl<'a> Context<'a> {
         // Deno removed support for .wasm imports in https://github.com/denoland/deno/pull/5135
         // the issue for bringing it back is https://github.com/denoland/deno/issues/5609.
```
guybedford: This comment is also out of date.

crates/cli-support/src/js/mod.rs
```diff
@@ -330,42 +330,24 @@ impl<'a> Context<'a> {
...
-            ",
-                path.file_name().unwrap().to_str().unwrap()
-            ));
-            shim.push_str(
-                "
-                const wasmModule = new WebAssembly.Module(bytes);
-                const wasmInstance = new WebAssembly.Instance(wasmModule, imports);
+                const wasmUrl = new URL('{module_name}_bg.wasm', import.meta.url);
+                const wasmBytes = await readFile(wasmUrl);
+                const {{ module: wasmModule, instance: wasmInstance }} = await WebAssembly.instantiate(wasmBytes, imports);
```
guybedford: This is a breaking change - and results in `require(mod)` not working. We shouldn't use TLA for Node.js, or it should be an option that is configurable.

guybedford (changes requested):

crates/cli-support/src/js/mod.rs
RReverser: This is specifically for the "experimental nodejs modules" target, not the primary Node.js one.

I think it's fair for an experimental target to change until it's stabilised?

RReverser: For extra context, that's the target we were talking about earlier. It was added a long time ago when JavaScript ESM itself was experimental in Node, but then never added to eg wasm-pack.

guybedford: While it's fair to change an experimental target, this change is not well-motivated and breaks valid usage patterns.

RReverser: TLA is really nice for Wasm scenario to avoid blocking the whole module graph, and aligns with Deno output closer (my hope is that eventually we'd align them enough that there won't be a need for a separate target) but I don't have that strong of an opinion at the moment, can switch back to readFileSync.

guybedford: A lot of Node.js use cases still rely on `require('lib')` working which has led to a temporary push away from TLA. This will resolve fairly quickly, but right now it's worth making this allowance rather I think.

crates/cli-support/src/js/mod.rs
RReverser: Ah I thought it's still the case but looks like they did implement it after all? Seems fairly recent though, I guess a bit early to switch for us. I'll update the comment.

PR #4672, opened 2025-09-23
Test more targets in import tests

RReverser:
Test more targets in import tests

import.rs tests a lot more of the module syntax than targets.rs (e.g. raw modules, inline_js, etc) and it will produce different output on different targets, so it's worth testing it as a matrix as well.

Also made the suffixes more readable as I'm tired of trying to remember which one of `-2`, `-3`, `-6` etc I need to look at heh.

In particular I want this to more easily see changes like https://github.com/wasm-bindgen/wasm-bindgen/pull/4671.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -230,7 +230,7 @@ jobs:
     - name: Check TypeScript output
       run: tsc --noEmit --skipLibCheck --lib esnext,dom $(echo crates/cli/tests/reference/*.d.ts)
     - name: Check JavaScript output
-      run: tsc --noEmit --skipLibCheck --lib esnext,dom --module esnext --allowJs $(echo crates/cli/tests/reference/*.js | sed 's/ crates\/cli\/tests\/reference\/targets-[56]\.js//g')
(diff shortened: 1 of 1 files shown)


crates/cli/tests/reference.rs
```diff
@@ -153,13 +153,17 @@ fn runtest(
...
 
-        // suffix the file name with the target
+        // suffix the file name with the sanitized flags
         let test = if all_flags.len() > 1 {
-            let base_file_name = format!(
-                "{}-{}.rs",
-                test.file_stem().unwrap().to_string_lossy(),
-                flags_index
-            );
+            let mut base_file_name = test.file_stem().unwrap().to_string_lossy().into_owned();
```
daxpedda, line 158:
```suggestion
            let mut base_file_name = test.file_stem().unwrap().to_str().unwrap().into_owned();
```
We shouldn't have any non UTF-8 file names to begin with.

RReverser:
> Didn't look at the failures

That's just TS step relying on hardcoded `-5.js` / `-6.js` suffixes. Updated.

PR #4675, opened 2025-09-23
Reuse Project for reference tests

RReverser:
wasm-bindgen CLI tests have a nifty `Project` test helper that takes care of setting up directories, invoking Cargo and wasm-bindgen.

WIth small modifications, we can reuse it for reference tests as well.

This does not only significantly simplify the reference test runner, but, more importantly, significantly speeds up running said tests as `Project` helper correctly sets up target directory to reuse that of the whole workspace, whereas our previous reference test runner used to create separate target directory for each test, which meant rebuilding same dependencies over and over.

### Timings on clean build

Before:

```
> cargo test -p wasm-bindgen-cli --test reference
...
test result: ok. 32 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 294.59s
```

After:
```
> cargo test -p wasm-bindgen-cli --test wasm-bindgen reference
...
test result: ok. 32 passed; 0 failed; 0 ignored; 0 measured; 16 filtered out; finished in 33.93s
```

### DIFF
--- a/crates/cli/tests/wasm-bindgen/reference.rs
+++ b/crates/cli/tests/wasm-bindgen/reference.rs
@@ -9,11 +9,6 @@
 //! compilation. Use `BLESS=1` in the environment to automatically update all
 //! tests.
 //!
-//! Note: Tests are run sequentially. In CI, tests are run ordered by name and
-//! all tests will be run to show all errors. Outside of CI, recently modified
-//! tests are run first and the runner will stop on the first failure. This is
-//! done to make it faster to iterate on tests.
-//!
 //! ## Dependencies
 //!
 //! By default, tests only have access to the `wasm-bindgen` and


crates/cli/tests/wasm-bindgen/main.rs
```diff
@@ -11,42 +11,58 @@
...
-    dir.pop(); // debug and/or release
-    dir
-}
-
-fn repo_root() -> PathBuf {
-    let mut repo_root = env::current_dir().unwrap();
-    repo_root.pop(); // remove 'cli'
-    repo_root.pop(); // remove 'crates'
-    repo_root
+lazy_static::lazy_static! {
```
daxpedda: We can't just use Std here?

RReverser: Yeah we could, I thought the macro provides a bit nicer syntax but the difference is not that big in the end.

Issue #4654, opened 2025-09-17
In version 0.2.102, using Rust prior to 1.82 will cause post-processing to fail in some scenarios.


Spxg:
This commit, https://github.com/wasm-bindgen/wasm-bindgen/pull/4642, changes the externref symbol to only exists when the `reference-types` target-feature is enabled.

Given that `wasm-bindgen` determines whether `reference-types` is enabled in the `target_feature` section, this means that if a `.o` file has it enabled, the final product will also have it enabled, causing `wasm-bindgen` to fail:

```shell
error: failed to find intrinsics to enable `clone_ref` function
error: failed to find the __wbindgen_externref_table_alloc function
```

This issue will be reproduced if you use `ring`, `zstd`, and other crates that require compiled C code, and use a compiler version prior to 1.82:

```rust
use wasm_bindgen::prelude::wasm_bindgen;

#[wasm_bindgen]
pub fn issue() -> u32 {
    unsafe { zstd_sys::ZDICT_isError(0) }
}
```
The simple solution is to add CFLAGS to disable:

```shell
export CFLAGS_wasm32_unknown_unknown="-mno-reference-types"
```

Perhaps this compatibility needs to be guaranteed by crate developers or `cc-rs`, but wasm-bindgen 0.2.102 does introduce a "breaking change" for this scenario.

daxpedda:
Hm, I think this is an acceptable breaking change?

Basically you can't simply mix-and-match target features between compilation files if those target features change the ABI.
In this case, the better solution might be to make sure that if you are using an old version of Rust, you also have to use an old version of Clang. The `wasm32-unknown-unknown` target between those too has to actually match up.

Spxg:
> is an acceptable breaking change

Probably not, especially considering the widely use of crates such as `ring` and `zstd-sys` that require compiled C code.

> In this case, the better solution might be to make sure that if you are using an old version of Rust, you also have to use an old version of Clang

Most users will hardly notice this, for example, which clang version has "reference-types" disable by default? Maybe there is a similar tool that can do this, I know nothing about clang version management.

daxpedda:
> > is an acceptable breaking change
> 
> Probably not, especially considering the widely use of crates such as `ring` and `zstd-sys` that require compiled C code.

Well, this doesn't outright break any compatibility with `ring` or other compiled C code. Just on target mismatch.

[...]

PR #4676, opened 2025-09-24
Detect `target_feature = "reference-types"` for Rust v1.82

daxpedda:
Rust v1.82 has enabled `target_feature = "reference-types"` by default for the `wasm32-unknown-unknown` target. However, because the target feature itself was still unstable until Rust v1.84, it can't be detected via `cfg(target_feature = "reference-types")`.

This PR introduces changes by detecting the relevant Rust versions and target so we can account for that. This works because users are unable to disable default target features without using the unstable `-Zbuild-std`.

To that end, `rustversion` is now enabled by default, which I've also moved to `build.rs`. This also gains us two new deprecated crate features: `msrv` and `rustversion`.

Resolves the Rust v1.82 issue discussed in #4654.


src/lib.rs
```diff
@@ -1087,7 +1079,7 @@ extern "C" {
 // standard wasm-bindgen ABI conversions.
 #[wasm_bindgen_macro::wasm_bindgen(wasm_bindgen = crate, raw_module = "__wbindgen_placeholder__")]
 extern "C" {
-    #[cfg(not(target_feature = "reference-types"))]
+    #[cfg(not(wbg_reference_types))]
     fn __wbindgen_externref_heap_live_count() -> u32;
```
Spxg, line 1083: Is this function still useful when externref is disabled?

daxpedda: AFAIK no, which is why its disabled without `reference-types`.

Spxg: What is the relationship with above
```rust
#[cfg(wbg_reference_types)]
use externref::__wbindgen_externref_heap_live_count;
```
Can we simply remove `fn __wbindgen_externref_heap_live_count() -> u32;`?

(not affect this PR merge)

daxpedda: Oops, sorry. Didn't read properly.

The non-reference-types version of this function returns the live count for the values saved on the JS side.
Its only used for tests, but these tests provide an important sanity check.

PR #4679, opened 2025-09-24
Add reference test for closures

RReverser:


RReverser:
Ah same issue as in https://github.com/wasm-bindgen/wasm-bindgen/pull/4652. Which is good, because I was surprised none of our existing reference tests caught it before.

daxpedda (approved):

RReverser:
So I sanitized all the hashed & indices, which seems to have done the trick for the most part, but the remaining diff is in table sizes between Windows and Linux.

I could easily sanitize this one as well, but I wonder if there's a genuine bug here.

Any ideas why it would differ between OS? (I have none)

guybedford:
I guess the question is if it is rustc or LLVM, and if an isolated case can be shared either way?

RReverser:
Hm I was a bit concerned because this is the only test that ran into this difference, but looks like it's because it's the only test that actually exports the function table, so that's probably fine.

All the other tests only export externref table, which we have control of and that's why it has a consistent size on all platforms.

I'm going to just set `initial = 0` for this one 🤞🏼

PR #4680, opened 2025-09-24
Run deploy & doc tasks only on main branch

RReverser:
This should speed up iteration on PRs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -375,6 +375,7 @@ jobs:
         path: benchmarks
 
   dist_linux_x86_64_musl:
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


RReverser:
Lol why does this remove Clippy & Rustfmt...

RReverser:
Ah hm looks like it's Github CI (or Rustup) that removed it as it now fails even on main branch even though worked fine earlier in the day.

RReverser:
And now it randomly started working again. Bizarre...

.github/workflows/main.yml
```diff
@@ -375,6 +375,7 @@ jobs:
         path: benchmarks
 
   dist_linux_x86_64_musl:
+    if: github.ref == 'refs/heads/main'
```
daxpedda, line 378: In practice nobody uses the nightly artifacts I believe, we could even do this only on a tag push, like deployment.

In any case, it doesn't hurt that we check if its not broken.

RReverser: Yeah I figured the time it takes on pushes to main is less critical than iterating on PRs.

PR #4681, opened 2025-09-24
Convert closure table elements to exports early in the pipeline

RReverser:
Closures can't be exported because Rust still lacks symbol visibility facilities (https://github.com/rust-lang/rust/issues/73295), so we describe them via invoke & destroy function pointers and look up the actual functions later in transforms.

The problem with this approach is that _all_ of the transforms need to deal with table elements just to support closures, even though we already always assume (via `get_function_table_entry`) that the function table entries are immutable.

Instead of having to deal with table entries during JS generation, we can export each function used by closures early in the process (in the "WIT" phase), so that from the point of view of JS generation and all the intermediate transforms it's just a regular export like any other.

That, in turn, allows to remove `Instruction::CallTableElement` with tons of associated code. It also allows to keep the function table private instead of exporting it just for these few callbacks.

Best to review with whitespace disabled.


RReverser:
What's going on with CI / Rustup :/

Clippy component keeps appearing and disappearing between reruns.

RReverser:
Huh looks like avoiding `rustup update stable` does the job. I've been meaning to do this regardless, since Github Actions track latest releases promptly, but if it actually avoids a bug with mismatched components, all the better.

RReverser:
There's definitely something wrong with CI still
image

.github/workflows/main.yml
```diff
@@ -26,7 +26,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v5
-    - run: rustup update --no-self-update stable && rustup default stable
```
daxpedda: I'm against removing this.
I've had many issues in the past with GitHub runners not really being up-to-date or doing other funny stuff.

I believe the issue you were hitting there is actually https://github.com/actions/runner-images/pull/13044, where GitHub decided to pin an older version of Rust. Because everybody does `rustup update` in their workflow, it overwhelmed Rusts infrastructure.

Now that might sound like an argument *for* removing this line, but its not. I think in practice if something weird like this is happening we *want* to know. So when CI fails because of this, I think its worth it.

RReverser: > I believe the issue you were hitting there is actually [actions/runner-images#13044](https://github.com/actions/runner-images/pull/13044), where GitHub decided to pin an older version of Rust.

I don't think this is it, because in CI runs I alwaya saw correct version 1.90, just sometimes without the necessary components.

Either way this was just me debugging the failures, unrelated to this PR specifically, and should be fixed separately, so I'm planning to revert this commit.

RReverser:
> We also need to update the schema version.

I guess I'm still confused about when we need to do that... I didn't change the descriptor schema, so why do we need to do that here?

daxpedda:
We always have to do it when the CLI becomes incompatible with the library.
In this case we are changing intrinsics, so that should break it.

RReverser:
Ohh I see, this is definitely wider than I thought. Can we test for this in CI somehow?

RReverser:
Alternatively, why don't we just bump the schema version as part of the release process? It might be a bit overzealous, but wasm-pack already downloads wasm-bindgen CLI to match one from Cargo.lock, so most users shouldn't even notice.

RReverser:
Also, looking at the diff again, I don't think I actually touched the library in an incompatible way here? Eg I moved destroy function to outer scope for a more readable name, but the CLI / lib communication doesn't rely on its name anyway, they just look it up by callback index.

daxpedda:
> Alternatively, why don't we just bump the schema version as part of the release process? It might be a bit overzealous, but wasm-pack already downloads wasm-bindgen CLI to match one from Cargo.lock, so most users shouldn't even notice.

We could, but no reason to break it if its compatible.

> Also, looking at the diff again, I don't think I actually touched the library in an incompatible way here? Eg I moved destroy function to outer scope for a more readable name, but the CLI / lib communication doesn't rely on its name anyway, they just look it up by callback index.

Just to make sure I tried it and tests seem to pass, so seems to be fine!

daxpedda:
> Ohh I see, this is definitely wider than I thought. Can we test for this in CI somehow?

That would be nice, but would need some scripting work. Rough sketch:
- Check if the version has changed since the last release.
- If not download the latest release binary.
- Run tests with that binary.

RReverser:
> We could, but no reason to break it if its compatible.

It mostly would make things easier for us. Having to download a new CLI after a clear error message seems less critical than stuff silently breaking because we forgot to update the version.

daxpedda:
> > We could, but no reason to break it if its compatible.
> 
> It mostly would make things easier for us. Having to download a new CLI after a clear error message seems less critical than stuff silently breaking because we forgot to update the version.

I think I would rather put the work into adding an appropriate test instead. But not against bumping it either.

PR #4684, opened 2025-09-25
Add reference test for target.rs + atomics

RReverser:
The output for atomics is a quite different (it requires importing memory, which is done in various ways on different targets) but so far we haven't had reference tests for it.


.github/workflows/main.yml
```diff
@@ -255,6 +255,10 @@ jobs:
     - uses: dtolnay/rust-toolchain@stable
       with:
         targets: wasm32-unknown-unknown
+    - uses: dtolnay/rust-toolchain@nightly
+      with:
+        targets: wasm32-unknown-unknown
+        components: rust-src
```
daxpedda, line 258: You have to switch them around. Otherwise the nightly toolchain is the default.

RReverser: Ah thanks, yeah that's why CI still failed.

daxpedda (approved):

PR #4685, opened 2025-09-25
Fix memory import on all targets

RReverser:
This moves memory import handling to target-independent portion of processing, which both simplifies code a bit and fixes the atomics JS output for multiple targets at once where previously they'd try to import memory from `env` JS module, which doesn't exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## [0.2.105](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.104...0.2.105)
+
+### Fixed
+
+* Fixed multithreading JS output for targets `bundler`, `deno` and `module`.
+
 ## [0.2.104](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.103...0.2.104)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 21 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## [0.2.105](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.104...0.2.105)
+
+### Fixed
+
+* Fixed multithreading JS output for targets `bundler`, `deno` and `module`.
```
daxpedda, line 8: I was actually under the impression until now that they weren't exactly broken, just unsupported. But looking at the reference output its clearly broken.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # `wasm-bindgen` Change Log
 --------------------------------------------------------------------------------
 
+## [0.2.105](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.104...0.2.105)
```
```suggestion
## Unreleased
```

daxpedda (changes requested):
Would be nice if we had multi-threading tests for all those different targets. In any case, the reference tests are better then nothing.

CHANGELOG.md
RReverser: Does it matter? It's not like we're gonna skip a patch version or release 0.3 anytime soon.

CHANGELOG.md
RReverser: Yup, they would literally fail trying to import a non-existing module.

CHANGELOG.md
daxpedda: Its just for people looking in the repo. So its understood that this isn't of yet unreleased and we don't have a broken link.

RReverser: Okay. Could be nice if publish script could add one automatically. Either way, accepted the suggestion.

daxpedda (approved):

PR #4690, opened 2025-09-26
Add reference tests for MVP

daxpedda:
I was a bit concerned after #4689 that we broke compiling with no reference-types.
While this doesn't actually run tests, at least we can see if something changed now.


RReverser:
> I was a bit concerned after #4689 that we broke compiling with no reference-types.

How so? Do you mean just concerned aka in theory, or that you actually see we broke something?

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -139,10 +139,36 @@ fn runtest_targets_atomics() -> Result<()> {
...
+            .env("RUSTUP_TOOLCHAIN", "nightly")
+            .env("RUSTFLAGS", "-C target-cpu=mvp")
+            .arg("-Zbuild-std=std,panic_abort");
+    })
 }
 
-fn runtest_with_opts(test: PathBuf, atomics: bool) -> Result<()> {
+fn runtest_with_opts(
+    test: PathBuf,
+    suffix: Option<&str>,
```
RReverser, line 169: Would be simpler to just change this to

```suggestion
    suffix: &str,
```

and use `""`, `"_atomics"`, `"_mvp"` explicitly at callsites. Removes allocation, dealing with `None` etc.

daxpedda:
> > I was a bit concerned after #4689 that we broke compiling with no reference-types.
> 
> How so? Do you mean just concerned aka in theory, or that you actually see we broke something?

No, just in theory!

RReverser:
Whew. But yeah I don't think it could've broken compiling with no reference-types since the transform I modified _only_ runs if reference-types are enabled.

crates/cli/tests/wasm-bindgen/reference.rs
daxpedda: I instead removed the allocation, because we once use it with `_` and once with `-`.

RReverser: I think it still makes sense for "suffix" to mean the whole thing together with underscore, but I don't care that strongly.

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -139,10 +139,36 @@ fn runtest_targets_atomics() -> Result<()> {
...
+    f: impl FnOnce(&mut Project),
```
RReverser, line 170: Another small nit: I'd limit this to just `&mut Command` until we actually need to pass the whole project.

RReverser (approved):
Approving with non-critical nits.

crates/cli/tests/wasm-bindgen/reference.rs
daxpedda: How would you do the `-` vs `_` handling?

daxpedda:
I should probably stop auto-merging.

crates/cli/tests/wasm-bindgen/reference.rs
RReverser: Either is fine by me.

PR #4692, opened 2025-09-26
Generate more meaningful export names

RReverser:


daxpedda (approved):
That's so nice!

RReverser:
One thing I wasn't sure about - any concerns about not having `__wbindgen_` prefix on those autogenerated names? We still generate unique names either way, just not sure if we want to separate them explicitly from user-provided names (but then we'd have to do the same for `memory` when it's not explicitly exported).

daxpedda (changes requested):
> One thing I wasn't sure about - any concerns about not having `__wbindgen_` prefix on those autogenerated names? We still generate unique names either way, just not sure if we want to separate them explicitly from user-provided names (but then we'd have to do the same for `memory` when it's not explicitly exported).

We should. This has an unfortunate history of issues. At some point it was switched to `__wbg_` as well, so that's what we should do.

But yeah, it has not been consistently applied until now.

RReverser:
Ugh this is not as easy as I thought, apparently in some cases we add, remove (through `unexport_intrinsics`) and re-add intrinsic exports so I end up with names like `__wbindgen_wbindgen_...` 🤦‍♂️

This will need a bit more tweaking, will take a look next week.

RReverser:
I realised I'm overthinking it a bit. We shouldn't worry about functions and globals - these symbols, when defined, are guaranteed to be unique (bar codegen-units duplication) via combination of mangling at Rust level & linker checking for duplicates - so the only names we need to worry about during generation are memories and tables.

For memory we've already generated `memory` name all this time, which we might want to change in cases when memory in core Wasm is imported and not exported, but I think that's fine to leave out of scope of this PR.

Now only tables remain, so I renamed that table to `__wbindgen_externrefs`.

guybedford (approved):

Issue #4653, opened 2025-09-17
Expose `wasm-bindgen-test-runner` as a Library


UebelAndre:
Howdy, I would like to include the `wasm-bindgen-cli` binary in some tooling and due to https://github.com/rust-lang/cargo/issues/9096 not yet being stabilized, getting access to the `wasm-bindgen` binary is pretty tricky and requires additional infrastructure or extra complex scripting to make sure it's available. Would it be acceptable by the maintainers to make [crates/cli/src/bin/wasm-bindgen.rs](https://github.com/wasm-bindgen/wasm-bindgen/blob/8bfa10ae7b2e8c7507f1bb2bebd27ea13838b44c/crates/cli/src/bin/wasm-bindgen.rs) and [crates/cli/src/bin/wasm-bindgen-test-runner/main.rs](https://github.com/wasm-bindgen/wasm-bindgen/blob/8bfa10ae7b2e8c7507f1bb2bebd27ea13838b44c/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs) into libraries where the only public function is `main`?

Spxg:
Maybe you can use [wasm-bindgen-cli-support](https://crates.io/crates/wasm-bindgen-cli-support).

daxpedda:
Let us know if `wasm-bindgen-cli-support` doesn't meet your need!

UebelAndre:
It notably doesn't have any of the `wasm-bindgen-test-runner` code. Maybe that's something that could be moved to `cli-support`?

daxpedda:
Yeah, that should definitely be in scope.

UebelAndre:
I'd be happy to open a PR for that if you had any guidance on how you'd like that to be done.

daxpedda:
I'm not too sure myself. The important part is to make sure that any further dependencies are hidden behind a crate feature and that we don't depend on `clap` inside `wasm-bindgen-cli-support`.

I think it is best if you take a closer look yourself and come up with a rough overarching plan that we can approve before investing too much time into code.

UebelAndre:
What is the stated purpose of `wasm-bindgen-cli-support`? That might help inform the design and what should go where. Not depending on `clap` is already a step beyond what I was gonna propose. Is the desire to avoid this dependency specifically or to avoid adding new dependencies to `wasm-bindgen-cli-support`?

daxpedda:
[...]

PR #4693, opened 2025-09-26
Migrate code from `cli` crate to `cli_support` crate.

UebelAndre:
Closes https://github.com/wasm-bindgen/wasm-bindgen/issues/4653


UebelAndre:
> We have to make sure that _any_ additional big dependencies in `wasm-bindgen-cli-support` are gated behind crate features, or in this case just one crate feature `test-runner` or something in that direction.
> 
> Otherwise this seems fine so far. Indeed environment variables shouldn't go into `wasm-bindgen-cli-support`, but should be handled and passed along in `wasm-bindgen-test-runner` itself.

Something like 39b986caa9390af65fd1b64750b06a1c230071fd ?

daxpedda:
Yes!

UebelAndre:
> Yes!

Sweet! Then let me know what you think of the changes and I can go address anything you find 😄

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
```diff
@@ -127,24 +129,15 @@ pub fn execute(
     };
     fs::write(&js_path, js_to_execute).context("failed to write JS file")?;
 
-    // Augment `NODE_PATH` so things like `require("tests/my-custom.js")` work
-    // and Rust code can import from custom JS shims. This is a bit of a hack
-    // and should probably be removed at some point.
-    let path = env::var("NODE_PATH").unwrap_or_default();
```
daxpedda, line 133: Could this be some sort of global environment variable and not `wasm-bindgen` specific? If so, we should probably keep it here.

crates/cli-support/src/testing.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub fn env(self) -> &'static str {
+        match self {
+            TestMode::Node { .. } => "WASM_BINDGEN_USE_NODE_EXPERIMENTAL",
+            TestMode::Deno => "WASM_BINDGEN_USE_DENO",
+            TestMode::Browser { .. } => "WASM_BINDGEN_USE_BROWSER",
+            TestMode::DedicatedWorker { .. } => "WASM_BINDGEN_USE_DEDICATED_WORKER",
+            TestMode::SharedWorker { .. } => "WASM_BINDGEN_USE_SHARED_WORKER",
+            TestMode::ServiceWorker { .. } => "WASM_BINDGEN_USE_SERVICE_WORKER",
+        }
+    }
```
daxpedda: Lets move this as a free-standing function into the binary.

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -291,7 +291,7 @@ fn sanitize_wasm(wasm: PathBuf) -> Result<String> {
     }
     for table in module.tables.iter_mut() {
         // The function table comes from LLVM and has different size between platforms.
-        if table.element_ty == walrus::RefType::Funcref {
+        if table.element_ty == wasm_bindgen_cli_support::testing::walrus::RefType::Funcref {
```
daxpedda: Lets just import `RefType`.

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -324,7 +324,7 @@ fn sanitize_wasm(wasm: PathBuf) -> Result<String> {
                 .collect(),
         ),
     );
-    walrus::passes::gc::run(&mut module);
+    wasm_bindgen_cli_support::testing::walrus::passes::gc::run(&mut module);
```
daxpedda: We can also just import `walrus` here.

crates/cli/tests/wasm-bindgen/reference.rs
UebelAndre: Done!

crates/cli/tests/wasm-bindgen/reference.rs
UebelAndre: Done!

crates/cli-support/src/testing.rs
UebelAndre: Done!

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -59,8 +59,10 @@ use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
+use wasm_bindgen_cli_support::testing::walrus::passes::gc::run as walrus_gc_run;
```
daxpedda: Not a fan of renaming imports if avoidable. Lets just import `walrus` so we can use it below.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
UebelAndre: I'm not super familiar with node environment variables but the value is overwritten by the modified value in the subprocess so I figured this was a good move that followed the spirit of clearing environment variables. I can leave it if you want but I don't think I can comment on the impact of any specific environment variables for node.

daxpedda: Yeah, that's fine. We have some experts here who will chime in eventually. Lets leave it as-is for now.

crates/cli/tests/wasm-bindgen/reference.rs
UebelAndre: Undone!

crates/cli/src/bin/wasm-bindgen-test-runner.rs
```diff
@@ -19,14 +19,13 @@ use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
 use std::thread;
+use wasm_bindgen_cli_support::testing::walrus;
+use wasm_bindgen_cli_support::testing::Test;
+use wasm_bindgen_cli_support::testing::TestMode;
+use wasm_bindgen_cli_support::testing::Tests;
+use wasm_bindgen_cli_support::testing::{deno, headless, node, server, shell};
```
daxpedda: Lets merge imports.

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -59,7 +59,8 @@ use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
-use walrus::{
+use wasm_bindgen_cli_support::testing::walrus;
+use wasm_bindgen_cli_support::testing::walrus::{
     ElementItems, ElementKind, ExportItem, FunctionKind, ImportKind, Module, ModuleConfig,
```
daxpedda:
```suggestion
use wasm_bindgen_cli_support::testing::walrus::{self,
ElementItems, ElementKind, ExportItem, FunctionKind, ImportKind, Module, ModuleConfig,
```

UebelAndre:
> I would have expected all the Wasm module parsing to move into `cli-support` as well. Can you describe your use-case so we can consider the design a bit more informed?

I want to write some tooling that wraps `wasm-bindgen-cli` and `wasm-bindgen-test-runner` for Bazel and due to the there being no library target in `wasm-bindgen-cli` or https://github.com/rust-lang/cargo/issues/9096 not being stabilized getting the binaries has required some clunky gymnastics. For my uses adding even a library target to `wasm_bindgen_cli` would get me what I need but if the desire is to have this in `wasm_bindgen_cli_support` then that also works for me and am willing to help contribute toward maintainers vision.

UebelAndre:
> I would have expected all the Wasm module parsing to move into `cli-support` as well.

If this is referring to kinda the bulk of `wasm-bindgen-test-runner.rs` then I would agree but don't quite know what the divide should be and figured it might be better to do that in a separate change. But I'd defer to you.

daxpedda (changes requested):
Thank you for the input! Thank you for going your contribution!

We definitely want `wasm-bindgen-cli-support` to support this kind of use-case. But we do want to keep control where its necessary and to ease user burden.

> > I would have expected all the Wasm module parsing to move into `cli-support` as well.
> 
> If this is referring to kinda the bulk of `wasm-bindgen-test-runner.rs` then I would agree but don't quite know what the divide should be and figured it might be better to do that in a separate change. But I'd defer to you.

We definitely don't want to leave the parsing to the user, so we will have to do this here.

But we shouldn't over-complicate anything without a compelling use-case first. So the divide should be straightforward: you can basically pass `Cli` along. This would need a new type because we don't want to depend on `clap` and there will be some small differences, like the `file`, which we shouldn't pass along. We can instead pass along `Vec<u8>` to leave the source entirely to the user.

Let me know if anything is unclear.

daxpedda:
After thinking about this a bit more:
In summary, we basically want to move all the code to `cli-support` apart from all the `clap` related code, the file loading (we do want to move the parsing) and environment variables.

UebelAndre:
> After thinking about this a bit more: In summary, we basically want to move all the code to `cli-support` apart all the `clap` related code, the file loading (we do want to move the parsing) and environment variables.

Is a56a5647c11ae23f283d9f0b110a3ebbc558a9ce kinda what you're thinking? This was a pretty crude code move but it sounded like what you were looking for might be a builder pattern here.

crates/cli-support/src/testing.rs
```diff
@@ -0,0 +1,526 @@
...
+        }
+
+        if self.list {
+            for test in tests.tests {
+                println!("{}: test", test.name);
+            }
+
+            // Returning cleanly has the strange effect of outputting
+            // an additional empty line with spaces in it.
+            std::process::exit(0);
```
daxpedda: We will have to do this in the runner instead.

crates/test-cli-support/src/lib.rs
daxpedda, line 1: Lets call this `test_runner`.

crates/cli-support/src/testing.rs
```diff
@@ -0,0 +1,526 @@
...
+
+pub mod deno;
+pub mod headless;
+pub mod node;
+pub mod server;
+pub mod shell;
+
+pub use walrus;
+
+pub struct Tests {
```
daxpedda: We can make all this stuff private again.

daxpedda (changes requested):
Yes, we want to model this like `Bindgen`, so lets go with `fn(&mut self) -> &mut Self`.
Whats left is as discussed:
- Pass file contents, not the path.
- Untangle all the environment variables.

crates/test-cli-support/src/lib.rs
UebelAndre: Done

crates/cli-support/src/testing.rs
UebelAndre: Done

crates/cli-support/src/testing.rs
UebelAndre: Done

UebelAndre:
> Yes, we want to model this like `Bindgen`, so lets go with `fn(&mut self) -> &mut Self`. Whats left is as discussed:
> 
>     * Pass file contents, not the path.
> 
>     * Untangle all the environment variables.

I think I've implemented what you're looking for. Let me know!

crates/cli-support/src/test_runner.rs
```diff
@@ -0,0 +1,627 @@
...
+            fallback_test_mode: None,
+            test_only_node: false,
+            test_only_web: false,
+            node_bin: None,
+            node_path: Vec::new(),
+            node_args: Vec::new(),
+        }
+    }
+
+    pub fn with_include_ignored(&mut self) -> anyhow::Result<&mut Self> {
```
daxpedda:
```suggestion
    pub fn include_ignored(&mut self) -> anyhow::Result<&mut Self> {
```
To align with `Bindgen`.

crates/cli-support/src/test_runner.rs
```diff
@@ -0,0 +1,627 @@
...
+        if self.ignored {
+            bail!("`--ignored` is mutually exclusive with `--include-ignored`");
```
daxpedda: Lets make sure we never return `Result` here. We can move those checks to `execute()`.

crates/test-cli-support/src/lib.rs
```diff
@@ -0,0 +1,627 @@
...
+                )
+                .context("failed to spawn server")?;
+                let addr = srv.server_addr();
+
+                // TODO: eventually we should provide the ability to exit at some point
+                // (gracefully) here, but for now this just runs forever.
+                if !headless {
+                    println!("Interactive browsers tests are now available at http://{addr}");
+                    println!();
+                    println!("Note that interactive mode is enabled because `NO_HEADLESS`");
```
daxpedda, line 573: Can you leave a `// TODO` here?

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 
 ## Unreleased
 
+### Changed
+
+* Migrated core `wasm-bindgen-test-runner` implementation into `wasm_bindgen_cli_support` library as the `testing` module.
```
daxpedda: Lets keep this user-facing.
```suggestion
* `wasm-bindgen-cli-support` now offers a way to execute the test-runner.
```

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 
 ## Unreleased
 
+### Changed
```
```suggestion
### Added
```

daxpedda (changes requested):
We still want to pass the file contents along instead of the path.
Otherwise this LGTM!

crates/cli-support/src/testing.rs
daxpedda: This is missing in the new runner binary now.

crates/cli/src/bin/wasm-bindgen-test-runner/node.rs
daxpedda: Cc @guybedford.

crates/test-cli-support/src/lib.rs
UebelAndre: What is the `TODO`? This was code that was copied:
https://github.com/wasm-bindgen/wasm-bindgen/blob/ba94c0dc77e6e6a911efe318d85d4df3f36c0cb2/crates/cli/src/bin/wasm-bindgen-test-runner/main.rs#L361-L372

daxpedda: Just that this mentions an environment variable that might or might not exist because we don't handle environment variables here directly anymore.

UebelAndre:
> We still want to pass the file contents along instead of the path.

Would you like `file_name` to then be a required argument?

crates/cli-support/src/test_runner.rs
UebelAndre: Done

crates/cli-support/src/test_runner.rs
UebelAndre: Done

crates/test-cli-support/src/lib.rs
UebelAndre: Are you saying to remove the prints in favor of a `TODO`?

crates/cli-support/src/test_runner.rs
```diff
@@ -0,0 +1,617 @@
...
+        self.debug = true;
+        self
+    }
+
+    pub fn split_linked_modules(&mut self) -> &mut Self {
+        self.split_linked_modules = true;
+        self
+    }
+
+    pub fn with_include_ignored(&mut self) -> &mut Self {
```
daxpedda: Remove `with_*` from all methods.

crates/cli-support/src/testing.rs
UebelAndre: If `execute` returns `Ok(())` then the process would exit with `0`, right?

crates/test-cli-support/src/lib.rs
daxpedda: No, just add a `TODO`, we will figure it out later.

crates/cli-support/src/testing.rs
daxpedda: Yes, but see the comment. We don't want those empty lines at the end because they break parsers like `cargo-nextest`.

crates/cli-support/src/test_runner.rs
UebelAndre: Done

crates/test-cli-support/src/lib.rs
UebelAndre: Done

crates/cli-support/src/testing.rs
UebelAndre: I'm not sure I understand the bug here. Should `wasm-bindgen-test-runner.rs` always finish with `proces::exit(0)`?

daxpedda:
> > We still want to pass the file contents along instead of the path.
> 
> Would you like `file_name` to then be a required argument?

Kinda funny its only used in the test coverage generation.
Its fine for now, this definitely will need some improvements in the future.

crates/cli-support/src/testing.rs
daxpedda: Yes.

Basically, by returning regularly form `fn main()`, e.g. via `Ok(())`, an extra empty line is inserted into the console. I have no idea why. But that extra line breaks tools trying to parse the output.

So we can just always return via `process::exit(0)` when `execute()` returns `Ok(())` to resolve the issue.

daxpedda:
Btw, something broke with the headless configuration. You can see the CI is stuck forever because it thinks we somehow enabled `NO_HEADLESS`.

crates/cli-support/src/testing.rs
UebelAndre: Done

UebelAndre:
> Btw, something broke with the headless configuration. You can see the CI is stuck forever because it thinks we somehow enabled `NO_HEADLESS`.

Oops, Should be fixed now!

crates/cli/src/bin/wasm-bindgen-test-runner.rs
```diff
@@ -0,0 +1,249 @@
...
+//! For more documentation about this see the `wasm-bindgen-test` crate README
+//! and source code.
+
+use anyhow::{bail, Context};
+use clap::Parser;
+use clap::ValueEnum;
+use std::env;
+use std::fs;
+use std::path::PathBuf;
+use std::process;
```
daxpedda:
```suggestion
use clap::{Parser, ValueEnum};
use std::{env, fs, process};
use std::path::PathBuf;´
```

daxpedda (approved):
LGTM!
Will give the other maintainers time to have a look over it.

daxpedda:
Needs a rebase as well.

RReverser (changes requested):
I'm not too sure about just dumping the whole test runner logic in the cli-support, especially after recent discussions we've had when I tried merging `cli-support` crate into` cli`.

In particular, I'm not very comfortable with the `cli-support` now spawning processes and depending on openssl.

Before these changes, it was almost entirely platform-independent library (~~and I think we could easily even make it `no_std`~~ actually no, because it still works with filesystem), but now these executable details "leak" into the library, so it spawns processes, HTTP server, `println`s stuff to console and so on, basically almost entire CLI logic not counting argument parsing.

Is it possible to split out these things back into CLI or is it too much work for user to replicate then?

If not, given your original usecase, it sounds like for this scenario it would make more sense to add `[lib]` target to `cli` crate itself, since the set of dependencies is almost the same as for the full thing and you want almost 1:1 logic anyway.

daxpedda:
> Before these changes, it was entirely platform-independent library that could even be compiled to Wasm (and I think we could easily even make it `no_std`), [..]

Its behind a crate feature.

> Is it possible to split out these things back into CLI or is it too much work for user to replicate then?

Unfortunately in its current state it doesn't make much sense. Its all so custom and has to be done in a very specific way to work correctly. We would need to put work into separating all these details away from the targets and for the user to be able to configure them correctly without too much work and footguns.

I think one of the biggest problems right now is that each target has special handling in how it treats the list of tests that aren't easily separated and packaged into a reasonable API.

> If not, given your original usecase, it sounds like for this scenario it would make more sense to add `[lib]` target to `cli` crate itself, since the set of dependencies is almost the same as for the full thing and you want almost 1:1 logic anyway.

If its just about the dependencies, ergo the crate feature isn't enough, we can just make it its own crate.
Why is the crate feature not enough or did you just miss it?

RReverser:
> Its behind a crate feature.

> If its just about the dependencies, ergo the crate feature isn't enough, we can just make it its own crate.
> Why is the crate feature not enough or did you just miss it?

I did see it, but at the point where it has to have almost every dependency from CLI, and then CLI forwards all of these features to cli-support, it just doesn't make sense for it to be in the cli-support crate.

Either making the CLI a library for this usecase, or adding a new crate like you said makes more sense. It's still pretty unusual for libraries to be spawning other processes, but at least if it's contained in a separate crate, that should be fine.

daxpedda:
I agree with the concerns. Moving it all into its own library sounds fine to me.
How about `wasm-bindgen-test-runner-support`? (actually serious :/)

RReverser:
> Either making the CLI a library for this usecase

Given that it's only `clap` and `env_logger` that remain in CLI if this code is split out, and given that the motivating usecase is to essentially just depend on and invoke this binary from other Rust code, I still think that simply adding a library target to CLI makes more sense.

It could even literally expose `fn run_cli_with_args(&[...])`, `fn run_test_with_args(&[...])` etc as a function for each available binary, which sounds like would be sufficient for the motivating problem.

RReverser:
> I agree with the concerns. Moving it all into its own library sounds fine to me.
> How about `wasm-bindgen-test-runner-support`? (actually serious :/)

Yeah, I don't care about naming, either that or `wasm-bindgen-test-cli-support` is fine.

RReverser:
> It could even literally expose `fn run_cli_with_args(&[...])`, `fn run_test_with_args(&[...])` etc as a function for each available binary, which sounds like would be sufficient for the motivating problem.

I'm curious now - the library API here is certainly nice to have! - but if we had functions like that in the first place, would it solve your problem from https://github.com/wasm-bindgen/wasm-bindgen/issues/4653 too @UebelAndre?

daxpedda:
> > Either making the CLI a library for this usecase
> 
> Given that it's only `clap` and `env_logger` that remain in CLI if this code is split out, and given that the motivating usecase is to essentially just depend on and invoke this binary from other Rust code, I still think that simply adding a library target to CLI makes more sense.

My concern regarding this is really just the dependencies. My hope is that this can be used similar to how `wasm-bindgen-cli-support` is used today. We aren't there yet, because right now we are just passing along all the CLI options, but that's the goal. So that is the reason why I would want it into its own library, so we don't force these dependencies on users which might not need them.

RReverser:
> so we don't force these dependencies on users which might not need them

Yeah, you've mentioned that before when I tried merging crates, and I reluctantly agreed, but when - after this change - the only remaining big dependency is just `clap`, it suddenly makes a lot less sense to keep it separate.

It's a lot of dependencies here that are only for the test-runner.

RReverser:
But yeah let's just go with separate crate for now.

UebelAndre:
> > It could even literally expose `fn run_cli_with_args(&[...])`, `fn run_test_with_args(&[...])` etc as a function for each available binary, which sounds like would be sufficient for the motivating problem.
> 
> I'm curious now - the library API here is certainly nice to have! - but if we had functions like that in the first place, would it solve your problem from #4653 too @UebelAndre?

If there was a library added to the `wasm-bindgen-cli` crate it would be something I could use to solve for my use cases. Those functions would be the dead simplest thing on my end but as I mentioned, I'm willing to contribute support for what maintainers think is best here (thus this PR) 😄

daxpedda:
> > so we don't force these dependencies on users which might not need them
> 
> Yeah, you've mentioned that before when I tried merging crates, and I reluctantly agreed, but when - after this change - the only remaining big dependency is just `clap`, it suddenly makes a lot less sense to keep it separate.
> 
> It's a lot of dependencies here that are only for the test-runner.

I still stand by that. `clap` is a pretty big dependency that we would force on every user trying to use this library. Trying to minimize dependencies is I think a really big deal and worth the effort. Especially when users are trying to review what they are actually installing there. I'm aware none of these arguments hit home properly, e.g. most likely anybody using this library is going to be using `clap` on their side as well.

---

@UebelAndre lets move the module into its own crate, lets go with `wasm-bindgen-test-cli-support`.
The crate would need to be added to `/publish.rs`.

RReverser:
> I'm aware none of these arguments hit home properly, e.g. most likely anybody using this library is going to be using `clap` on their side as well.

Indeed, you've hit it on the head here for my counterarguments / feeling about all of this.

If we were talking about some useless large dependency or tons of dependencies, I'd entirely agree, but when it's down to just clap, in a library that is intended to help other people to build their own CLIs, and even that dependency could easily be one `default-features=false` away from being disabled... well, then the effort we're putting into it just feels silly.

daxpedda:
> [..], and even that dependency could easily be one `default-features=false` away from being disabled ... [..]

Actually, we don't have the same issue we had before with moving `wasm-bindgen-cli-support`. So maybe we can do just that here as well.

@UebelAndre I think you work here is done here. We can move these files around ourselves. Thank you!

UebelAndre:
> @UebelAndre I think you work here is done here. We can move these files around ourselves. Thank you!

What's the next steps here? I'm happy to continue to help out since I would like some solution to land.

daxpedda:
We just discussed it briefly and we wanted to try https://github.com/wasm-bindgen/wasm-bindgen/pull/4693#issuecomment-3341998459 in a separate PR first and see how that works out.

So basically just expose two simple functions from `wasm-bindgen-cli` itself. This would involve moving the whole post-processor and test runner to the `[lib]`. No separation from `clap` or environment variables at all.

I didn't get to it yet and no idea how it should look like *exactly*, but if you want to experiment with it and make a PR that would certainly help!

UebelAndre:
closed in favor of https://github.com/wasm-bindgen/wasm-bindgen/pull/4710

PR #4695, opened 2025-09-27
Add integration test for experimental-nodejs-module

RReverser:
We can't use wasm-pack here like we do for all other tests, but at least with Cargo's message-format=json the custom script is pretty trivial.

### DIFF
--- a/examples/nodejs_and_deno/cargo-and-wasm-bindgen.js
+++ b/examples/nodejs_and_deno/cargo-and-wasm-bindgen.js
@@ -1,0 +1,27 @@
+import { execFileSync } from 'node:child_process';
+import * as assert from 'node:assert/strict';
+
+let { reason, filenames } = JSON.parse(execFileSync(
+  'cargo',
+  [
+    'build',
+    '--target',
+    'wasm32-unknown-unknown',
+    '--release',
+    '--message-format=json'
+  ],
+  {
+    stdio: ['inherit', 'pipe', 'inherit'],
+    encoding: 'utf-8'
+  }
+)
+  .trimEnd()
+  .split('\n')
+  .at(-2));
+
+assert.equal(reason, 'compiler-artifact', 'the message before the build-finished message should be a compiler-artifact');
+assert.equal(filenames.length, 1, 'there should be only one output filename');
+
+execFileSync('wasm-bindgen', [...filenames, ...process.argv.slice(2)], {
+	stdio: 'inherit'
+});
--- a/examples/nodejs_and_deno/package.json
+++ b/examples/nodejs_and_deno/package.json
@@ -1,12 +1,14 @@
 {
 	"type": "module",
 	"scripts": {
-		"build:nodejs": "wasm-pack build --target nodejs --out-dir ../dist/nodejs_and_deno --out-name nodejs",
-		"build:deno": "wasm-pack build --target deno --out-dir ../dist/nodejs_and_deno --out-name deno",
-		"build": "npm run build:nodejs && npm run build:deno",
+		"build:nodejs": "wasm-pack build --target nodejs --out-dir ../dist/nodejs_and_deno/nodejs",
+		"build:deno": "wasm-pack build --target deno --out-dir ../dist/nodejs_and_deno/deno",
+		"build:experimental-nodejs-module": "node cargo-and-wasm-bindgen.js --target experimental-nodejs-module --out-dir ../dist/nodejs_and_deno/experimental-nodejs-module",
+		"build": "npm run build:nodejs && npm run build:deno && npm run build:experimental-nodejs-module",
 
 		"test:nodejs": "node test.js nodejs",
 		"test:deno": "deno run --allow-read test.js deno",
-		"test": "npm run test:nodejs && npm run test:deno"
+		"test:experimental-nodejs-module": "node test.js experimental-nodejs-module",
+		"test": "npm run test:nodejs && npm run test:deno && npm run test:experimental-nodejs-module"
 	}
 }
--- a/examples/nodejs_and_deno/test.js
+++ b/examples/nodejs_and_deno/test.js
@@ -2,5 +2,5 @@ import { argv } from 'node:process';
 
 const targetName = argv[2];
 
-const { greet } = await import(`../dist/nodejs_and_deno/${targetName}.js`);
+const { greet } = await import(`../dist/nodejs_and_deno/${targetName}/node_and_deno.js`);
 greet(targetName);


examples/nodejs_and_deno/cargo-and-wasm-bindgen.js
```diff
@@ -0,0 +1,27 @@
...
+  .trimEnd()
+  .split('\n')
+  .at(-2));
+
+assert.equal(reason, 'compiler-artifact', 'the message before the build-finished message should be a compiler-artifact');
+assert.equal(filenames.length, 1, 'there should be only one output filename');
+
+execFileSync('wasm-bindgen', [...filenames, ...process.argv.slice(2)], {
+	stdio: 'inherit'
+});
```
guybedford: Since this is in the examples folder, let's just start to encourage direct workflows without wasm-pack:

```json
{
  "scripts": {
    "build:compile": "cargo build --target wasm32-unknown-unknown --release",
    "build:bindgen": "wasm-bindgen target/wasm32-unknown-unknown/foo.wasm --target experimental-nodejs-module --out-dir ...",
    "build": "npm run build:compile && npm run build:bindgen"
  }
}
```

RReverser: That won't work universally because target directory can depend on variety of configs - that's why either wasm-pack or the proposed `wasm-bindgen-ld` or a custom script like this is necessary to extract the actual final Wasm file.

Basically you can't rely on it being in `target/...` in current working directory, it's just the default. You need to get the actual path from Cargo, which is what I'm doing here.

guybedford: Sure, but even for our own test case this is fine since we don't customize target?

RReverser: We don't on CI, but e.g. I do have custom target directory configured in my local ~/.cargo/config.toml for better caching purposes, so for me this test would fail without correct extraction.

Same can happen to other contributors.

RReverser: Is there anything specific you dislike about this script implementation? I'm happy to address specific review concerns, but the functionality itself is definitely necessary for reproducible testing.

guybedford: The folder here is called "examples", but the technique is certainly not something I'd recommend to anyone.

Code in examples should set strong precedent for how to use wasm bindgen. If this is how we recommend folks use wasm bindgen, then I'm not sure what the hell we are doing to be quite honest.

guybedford: What about just adding `CARGO_TARGET_DIR=target` into the script invocation?

RReverser: > but the technique is certainly not something I'd recommend to anyone.

Until we have either this target in wasm-pack or full replacement for wasm-pack, some custom script for this experimental target is going to be unfortunately necessary. 

I'm definitely not a fan of having to do it either, but at least this way someone can even copy-paste same script to their build system for now 🤷‍♂️ 

> What about just adding `CARGO_TARGET_DIR=target` into the script invocation?

Then all of wasm-bindgen and its dependencies will be rebuilt just for this one target, which has a significant build time cost (eg having separate target dir is what I recently fixed in the reference tests, and shaved off 3 minutes off builds that way).

--message-format=json is specifically designed for tools invoking cargo so they can find the artifact and anything else (eg diagnostics) they want, the only question is whether wasm-pack does that or we do that in our script. Unfortunately, in this particular case wasm-pack doesn't do it.

RReverser: Before this comes up, I _could_ also use a shell script which does the same, but unlike a Node script, it wouldn't be cross-platform.

That's what we had  - custom shell and Python scripts - for half of examples before I did https://github.com/wasm-bindgen/wasm-bindgen/pull/4609 - so now having just one cross-platform script only for a target that really needs it doesn't seem like a very high cost.

And, of course, I'll be happy to remove it when either we have cargo-wasm-bindgen or wasm-pack integrates this target or whatever else allows us to do so without loss of functionality.

daxpedda: Why can't we assume in examples that the default target destination is used?
I agree that examples should be seen as recommendations how to use `wasm-bindgen`, they aren't there to account for every configuration out there.

I think presuming default configurations in examples is fine. We also shouldn't need to teach Cargo in examples either, unless its something Wasm specific.

daxpedda: After reading the conversation again, I hope I didn't misunderstand this: is this about reproducability because contributors running this on their machine might have a custom Cargo target configuration?

I really don't think assumptions like that should be present in examples, but besides the point, I hope we don't have to adjust tests to fit every contributors local configuration. I would still say that assuming defaults should be good enough.

RReverser: > because contributors running this on their machine might have a custom Cargo target configuration?

I mean, it's not just "might" - for me running integration tests would literally break without resolving correct target directory, and I hope I don't have to change the cargo config on my machine to accomodate one test in one specific example in one specific project.

> I really don't think assumptions like that should be present in examples

This is kinda the other way around - we shouldn't assume that target folder for everyone is in the same place.

RReverser: One slight simplification I could do is instead parse `target_dir` from `cargo metadata` command's JSON but I'd still rather not hardcode the directory.

Especially because it's an example, I believe we should show the best practice which is getting the correct path to the artifact from Cargo, not teach other people to hardcode directories in their projects as well.

daxpedda: > > because contributors running this on their machine might have a custom Cargo target configuration?
> 
> I mean, it's not just "might" - for me running integration tests would literally break without resolving correct target directory, and I hope I don't have to change the cargo config on my machine to accomodate one test in one specific example in one specific project.

Uhm, I understand where you are coming from. But I'm surprised there's an expectation from open source projects to accommodate non-default build systems/configurations? I've certainly never encountered that. It would have been fine if its free, but it isn't, *yet*.

> > I really don't think assumptions like that should be present in examples
> 
> This is kinda the other way around - we shouldn't assume that target folder for everyone is in the same place.

> Especially because it's an example, I believe we should show the best practice which is getting the correct path to the artifact from Cargo, not teach other people to hardcode directories in their projects as well.

Why not? I certainly haven't seen this being handled anywhere in public examples that need to access the Cargo output. Hell, even Rust documentation doesn't account for that.

Again: I believe assuming default configuration is sane. There are so many weird build systems/configurations we don't account for here, I don't know why we would make an exception for the target folder.

---

Still assuming we are talking about examples here. If its any other place I'm fine with it.

RReverser: > But I'm surprised there's an expectation from open source projects to accommodate non-default build systems/configurations?

> There are so many weird build systems/configurations we don't account for here, I don't know why we would make an exception for the target folder.

Maybe our experiences differ, but I find it's been always there. Usually incompatibility is treated like a bug, and it's fine to fix them on case-by-case basis as they arise, but any hardcoded assumption is still usually considered a bug.

> Hell, even Rust documentation doesn't account for that.

Can you show an example?

---

One overarching problem here is that, for better or worse, our examples folder both serves as an example and integration tests in one. If there is an easy way to separate those, that would be nice too, but for now I'm working with what we have.

If I won't be able to run integration tests locally or wouldn't be able to use the machine-wide cache anymore, I might as well close the PR as it would be a significant regression for local development.

I'm surprised doing the right thing - via the official JSON format Cargo provides specifically for build systems that invoke it - is so controversial though, just because it doesn't look as nice as the hardcoded path.

RReverser: > There are so many weird build systems/configurations we don't account for here, I don't know why we would make an exception for the target folder.

And, to be clear, this is *not* just target folder. Getting the artifact from Cargo messages also correctly accounts for others targets like `wasm32v1-none` / `wasm32-unknown-emscripten` / whatever we might add, whereas hardcoding directory only works for one very specific configuration.

RReverser: I found that folks behind [neon](https://github.com/neon-bindings/neon) have already published [cargo-cp-artifact](https://www.npmjs.com/package/cargo-cp-artifact) to npm, which does what I'm doing, but only for copying artifacts.

It does mean however that I can get rid of script for parsing JSON, and have commands purely inside package.json scripts, without losing the custom config support, so I did just that.

examples/nodejs_and_deno/package.json
```diff
@@ -1,12 +1,17 @@
 {
 	"type": "module",
 	"scripts": {
-		"build:nodejs": "wasm-pack build --target nodejs --out-dir ../dist/nodejs_and_deno --out-name nodejs",
-		"build:deno": "wasm-pack build --target deno --out-dir ../dist/nodejs_and_deno --out-name deno",
-		"build": "npm run build:nodejs && npm run build:deno",
-
+		"build:nodejs": "wasm-pack build --target nodejs --out-dir ../dist/nodejs_and_deno/nodejs",
+		"build:deno": "wasm-pack build --target deno --out-dir ../dist/nodejs_and_deno/deno",
```
guybedford, line 5: Could we even remove these now then?

guybedford (approved):
Very elegant, thank you for putting the thought to it!

examples/nodejs_and_deno/package.json
RReverser: Hm, no, they still test other JS targets. Or do you mean replacing wasm-pack everywhere in example tests?

PR #4699, opened 2025-09-30
Fix lib's CI build for MSRV(1.57)

Spxg:


RReverser:
It might be better to add it to definitions here, which I believe should have the same effect: https://github.com/wasm-bindgen/wasm-bindgen/blob/1614140824a257b1fc3890540972bb6281941286/crates/msrv/lib/Cargo.toml#L25

Spxg:
> It might be better to add it to definitions here, which I believe should have the same effect:

Sure, I did't notice that before, thanks.

PR #4700, opened 2025-09-30
Fix lib's CI build of MSRV(1.57) #2

Spxg:


Spxg:
Actually there is tiny difference here. In yaml, you can configure the version to be fixed only when the rustc version is 1.57. But this is a small problem.

RReverser:
> In yaml, you can configure the version to be fixed only when the rustc version is 1.57. But this is a small problem.

Yeah that shouldn't matter here, those projects are intentionally not part of the workspace and are only built by CI with Rust 1.57 for checking MSRV.

PR #4701, opened 2025-10-01
Add `get_raw` and `set_raw` for `WebAssembly.Table`

Spxg:
According to <https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/Table/set#value>, the type of value is now not just a wasm function.

Naming is hard, if you have any suggestions, please let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@
 
 ### Added
 
-* Added `new_with_value` and `grow_with_value` for for `WebAssembly.Table`.
+* Added `get_raw` and `set_raw` for `WebAssembly.Table`.
+  [#4701](https://github.com/wasm-bindgen/wasm-bindgen/pull/4701)
+
+* Added `new_with_value` and `grow_with_value` for `WebAssembly.Table`.
   [#4698](https://github.com/wasm-bindgen/wasm-bindgen/pull/4698)
 
 ### Fixed
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -4834,6 +4834,13 @@ pub mod WebAssembly {
         #[wasm_bindgen(method, catch, js_namespace = WebAssembly)]
         pub fn get(this: &Table, index: u32) -> Result<Function, JsValue>;
 
+        /// The `get()` prototype method of the `WebAssembly.Table()` object
+        /// retrieves a function reference stored at a given index.
+        ///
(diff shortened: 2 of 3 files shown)


daxpedda (approved):
This should probably be all moved to `web-sys` and the `js-sys` API could be depreacted.

hoodmane:
Ideally I'd think the externref table should have a different type than the funcref table. If it's a funcref table and you try to put anything except null or a wasm function into it it will fail with something like
```
TypeError: Argument 1 is invalid for table: function-typed object must be null (if nullable) or a Wasm function object
```
Maybe it's not worth worrying about though.

daxpedda:
> Ideally I'd think the externref table should have a different type than the funcref table.

Belongs to a higher-level API wrapper.
Here we are just trying to provide direct low-level bindings to the existing API.

Spxg:
> Maybe it's not worth worrying about though.

Sure, but here it is just because we cannot `set` or `get` the value of other type because it is hard-coded to be of type Function.

```rust
pub fn get(this: &Table, index: u32) -> Result<Function, JsValue>;
pub fn set(this: &Table, index: u32, function: &Function) -> Result<(), JsValue>;
```

Spxg:
> This should probably be all moved to `web-sys` and the `js-sys` API could be depreacted.

indeed

RReverser:
> This should probably be all moved to `web-sys` and the `js-sys` API could be depreacted.

I was thinking that we should have a separate crate for https://min-common-api.proposal.wintertc.org/#api-index. It's neither here nor there, it's a subset of APIs that are implemented by both browsers and popular non-browser engines alike.

Some of these APIs - like `WebAssembly` - don't belong in js-sys because they're not part of core language, but they also don't need the whole web-sys when users want to only use stuff that all runtimes have.

Of course, it could also be a Cargo feature on web-sys, just worried it might be lost among all the other features then.

daxpedda:
`web-sys` already hides every single API behind its own crate feature. So that seems like a minor problem to me.

However, I still think it might be useful to split `web-sys` into multiple crates. This would make breaking changes easier and also moves us away from relying on the crates.io team to make an exception for `web-sys`'s crate feature list, where we exceed the number of allowed crate features.

RReverser:
> and also moves us away from relying on the crates.io team to make an exception for `web-sys`'s crate feature list, where we exceed the number of allowed crate features.

Lol what? 😂 I thought windows-rs has us beat there, but no, they only have 673 features while we have 1660. https://docs.rs/crate/windows/latest/features

PR #4703, opened 2025-10-01
Fix `TextDe/Encoder` detection for audio worklets

daxpedda:
We do still want to be able to call `TextDe/Encoder` APIs from the main thread while spawning audio worklets.
To that end we just detect if atomics is enabled in the first place before adding the detection code.

Follow-up from https://github.com/wasm-bindgen/wasm-bindgen/pull/4670#issuecomment-3348818733.
Cc @kettle11.


RReverser (approved):

guybedford (approved):
I posted an HTML issue for prosperity https://github.com/whatwg/html/issues/11736.

kettle11:
@daxpedda Thanks for so promptly addressing this!

A thought I had is that `wasm-bindgen` could be useful with AudioWorklets even if atomics are disabled.

A rust program could run on an AudioWorklet in a totally separate instance without SharedMemory and use postMessage to communicate with other threads.

daxpedda:
> A thought I had is that `wasm-bindgen` could be useful with AudioWorklets even if atomics are disabled.
> 
> A rust program could run on an AudioWorklet in a totally separate instance without SharedMemory and use postMessage to communicate with other threads.

I believe this is covered by just not using `TextDe/Encoder` when you are making a module exclusively targeting audio worklets. But let us know if that doesn't work out for you!

PR #4706, opened 2025-10-01
Added CookieStore API

### LINKED ISSUE #4704: Add option to hook into CookieStore
Not sure if this is the right place to post this, but I saw this API on MDN:

https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/change_event

Couldn't find this on the web-sys crate. I could technically use bindgen and make me own wrapper, but I was wondering if this is something that we should include in web-sys?
I can confirm that I have tried this on the browser console (added an event that console logs) and when a cookie is set I get an event fired. Even though it says it's available to service workers, it seems to work on normal windows as well

rakshith-ravi:
Based off of the reference from https://cookiestore.spec.whatwg.org/

Fixes #4704

Please let me know if there's anything else that needs to be updated.


daxpedda (changes requested):
Missing a changelog entry.
Otherwise LGTM!

daxpedda:
Would like some input from more knowledgable maintainers if this API looks stable enough.
My concern with WHATWG APIs (not W3C) is that they sometimes break subtle parts of the interface that we then get stuck with because of Rust SemVer.

If not deemed stable enough, we can just move it to unstable.

rakshith-ravi:
Added changelogs 😄

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 
 ### Added
 
+* Added CookieStore API
```
```suggestion
* Added `CookieStore` API.
```

daxpedda (approved):

PR #4707, opened 2025-10-01
Simplify Clippy CI config

RReverser:
Since https://github.com/wasm-bindgen/wasm-bindgen/pull/4437 was merged, we can just use `--all --all-targets` with Clippy for checking on host.

We also don't need the separate `cargo check` task - `cargo clippy` is just a wrapper that checks both built-in Rust lints as well as its own - and can merge other 2 tasks using a matrix.

This revealed that we were not covering test projects, which Clippy found some issues in, so I fixed those as part of the PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,7 +73,7 @@ jobs:
         -- -D warnings
 
   # Run `cargo clippy` over crates that support `no_std`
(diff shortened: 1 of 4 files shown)


guybedford (changes requested):
Can we update the justfile here as well to reference the new conventions too?

PR #4708, opened 2025-10-01
Test with Wasm MVP

RReverser:
Just using the same trick @daxpedda used for reference tests. Ideally we'd test properly with Rust 1.57 for full MSRV coverage, but we can't do that at the moment as our Cargo workspace is not compatible with Rust that old.

This way we'll at least catch any issues in the generated JS when not targeting externref & multivalue.


guybedford (approved):
Looks like we'll need to update the branch rules with this too.

RReverser:
> Looks like we'll need to update the branch rules with this too.

Yeah I'll update now & merge.

PR #4709, opened 2025-10-01
Move Closure's ownership entirely to JS

RReverser:
This makes its memory management a bit less weird and more in line with normal Rust - there is a single side (JS) owning the actual heap data, and it's the only one responsible for invoking the destructor when refcount reaches zero, doesn't matter how that happens.

That allows to remove another `unsafe` too.


crates/cli-support/src/js/mod.rs
```diff
@@ -2379,15 +2379,17 @@ wasm = wasmInstance.exports;
...
-                            state.dtor(a, state.b);
-                            CLOSURE_DTORS.unregister(state);
-                        }} else {{
-                            state.a = a;
-                        }}
+                        state.a = a;
+                        real._wbg_cb_unref();
+                    }}
+                }};
+                real._wbg_cb_unref = () => {{
```
guybedford, line 2386: I wonder if this could just be`Symbol.dispose` here, effectively making disposal public?

RReverser: Yeah it could, but then we still need custom fallback for environments w/o `Symbol.dispose`. For now I'm just keeping status quo and keeping it private like before this PR.

It wouldn't be too useful until something along lines of https://github.com/whatwg/html/issues/8557#issuecomment-3150359084 is implemented anyway - the persistent `Closure` type is only useful for passing to Web APIs which will invoke it asynchronously, and then it's not up to the user to decide when to dispose them.

RReverser:
> In workers-rs we actually have a closure leak right now - so that a public explicit disposal feature would actually be useful.

Do you mean intentional or accidental? Because in adding this test I noticed that it was broken before (very likely by one of my own previous PRs) and the JS closure wasn't freed, and this PR fixes that.

RReverser:
> This PR is adding a new public method to closures

Also just want to clarify the intent was not so much adding a new method, as hiding `real.original` implementation detail under an API that more clearly conveys it's not meant to be accessed.

RReverser:
> Do you mean intentional or accidental? Because in adding this test I noticed that it was broken before (very likely by one of my own previous PRs) and the JS closure wasn't freed, and this PR fixes that.

Added changelog entry to clarify that this is a fix.

guybedford:
It seems this introduced a bug into Gloo usage in workers-rs. See https://github.com/wasm-bindgen/wasm-bindgen/issues/4732.

PR #4710, opened 2025-10-01
Added library to `wasm_bindgen_cli`.

UebelAndre:


crates/cli/src/wasm_bindgen.rs
```diff
@@ -0,0 +1,179 @@
...
+where
+    I: IntoIterator<Item = T>,
+    T: Into<OsString> + Clone,
+{
+    let args = Args::try_parse_from(args)?;
+    rmain(&args)
+}
+
+fn rmain(args: &Args) -> Result<(), Error> {
+    env_logger::init();
```
RReverser: `env_logger::init()` needs to be left in the binary. In the library environment most likely logger might be already set up globally, or user might decide not to use it, and `env_logger::init()` panics if called twice.

It's really only meant to be called at the top level.

crates/cli/src/bin/wasm-bindgen-test-runner.rs
```diff
@@ -0,0 +1,8 @@
+use std::{env, process};
+
+fn main() -> anyhow::Result<()> {
+    let args: Vec<String> = env::args().collect();
+    let args_ref: Vec<&str> = args.iter().map(|s| s.as_str()).collect();
+    wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args(&args_ref)?;
```
RReverser: I don't think you need to collect it, or even convert to Rust strings.

`wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args(std::env::args_os())` should also just work.

UebelAndre: I initially thought I had to strip out argv0.

crates/cli/src/bin/wasm-bindgen-test-runner.rs
```diff
@@ -1,8 +1,7 @@
 use std::{env, process};
 
 fn main() -> anyhow::Result<()> {
-    let args: Vec<String> = env::args().collect();
-    let args_ref: Vec<&str> = args.iter().map(|s| s.as_str()).collect();
-    wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args(&args_ref)?;
+    env_logger::init();
+    wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args(env::args())?;
```
RReverser: Please use `args_os` as mentioned in the comment, otherwise you're paying for an unnecessary conversion.

crates/cli/src/wasm_bindgen.rs
UebelAndre: Done

crates/cli/src/bin/wasm-bindgen-test-runner.rs
UebelAndre: Oops, Done

crates/cli/src/bin/wasm-bindgen-test-runner.rs
```diff
@@ -0,0 +1,7 @@
+use std::{env, process};
+
+fn main() -> anyhow::Result<()> {
+    env_logger::init();
+    wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args(env::args_os())?;
+    process::exit(0);
```
RReverser, line 8:
```suggestion
    wasm_bindgen_cli::wasm_bindgen_test_runner::run_cli_with_args(env::args_os())
```

Explicit `process::exit` is almost never needed, you can just return the result here.

RReverser: Or I guess you can do what main wasm-bindgen binary is doing and print the error, since returning it would print its debug representation instead.

UebelAndre: That's what my original thought was as well but please see https://github.com/wasm-bindgen/wasm-bindgen/pull/4693#discussion_r2384009456
cc @daxpedda

crates/cli/src/bin/wasm2es6js.rs
```diff
@@ -1,73 +1,5 @@
...
-        let filename = args.input.file_name().unwrap();
-        let dst = p.join(filename).with_extension(extension);
-        fs::write(&dst, contents)
-            .with_context(|| format!("failed to write `{}`", dst.display()))?;
-    } else if print_fallback {
-        println!("{}", String::from_utf8_lossy(contents))
-    }
-
-    Ok(())
+    wasm_bindgen_cli::wasm2es6js::run_cli_with_args(env::args_os())
```
RReverser, line 5: I see we didn't have it before, but we probably should have `env_logger::init()` in this binary as well.

UebelAndre: Done!

crates/cli/src/bin/wasm-bindgen-test-runner.rs
RReverser: Oh wow um that's super weird but okay.

crates/cli/src/wasm_bindgen_test_runner.rs
RReverser: Actually, it should be removed here now instead, right? (maybe move the comment over to the CLI binary)

RReverser:
@UebelAndre Actually, you might want to look a bit more through the library code - looks like there are few more paths invoking `process::exit(...)` that you really don't want to be invoked from a library.

crates/cli/src/wasm_bindgen_test_runner.rs
UebelAndre: Updated!

RReverser:
Since @daxpedda is currently on leave, I'm going to go ahead and merge this, we can always iterate on further improvements separately. Thanks for the patience @UebelAndre!

UebelAndre:
> Since @daxpedda is currently on leave, I'm going to go ahead and merge this, we can always iterate on further improvements separately. Thanks for the patience @UebelAndre!

Thanks! Would it be possible to get a new release for this?

RReverser:
We'll need to wait for @daxpedda to return. You can point your Cargo.toml to `wasm-bindgen-cli = { git = "https://github.com/wasm-bindgen/wasm-bindgen" }` meanwhile though.

PR #4714, opened 2025-10-02
Add missing TryFromJsValue implementations

guybedford:
I'm exploring if the `JsValueCast` generic cast can be replaced with `TryFromJsValue`. This adds some implementations that seem to have been omitted.

In the process, this allows `TryFromJsValue` to become a primary trait for obtaining runtime `JsValue` to arbitrary type conversions, since `From` is restricted from applying to all types since users often have custom implementations (the original motivation for `TryFromJsValue`), and `JsCast` does not convert between arbitrary Rust and JS types in the same way. When combined with `Into<JsValue>`, this can therefore provide full runtime type conversions that can mirror the static semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@
 * Added better support for async stack traces when building in debug mode.
   [#4711](https://github.com/wasm-bindgen/wasm-bindgen/pull/4711)
 
+* Extended support for `TryFromJsValue` trait implementations.
+  [#4714](https://github.com/wasm-bindgen/wasm-bindgen/pull/4714)
+
 ### Fixed
 
 * Fixed multithreading JS output for targets `bundler`, `deno` and `module`.


crates/macro-support/src/codegen.rs
```diff
@@ -1145,6 +1145,16 @@ impl ToTokens for ast::ImportType {
                         unsafe { &*(val as *const JsValue as *const #rust_name) }
                     }
                 }
+
+                #[automatically_derived]
+                impl TryFromJsValue for #rust_name {
```
RReverser: I think this can be done as a blanket implementation in `impls.rs` instead of individual codegens.

guybedford: Good point, added.

guybedford:
Note that regardless of whether we go this direction or `wbg_cast` for the generics PR, this would still be a useful improvement I think.

RReverser:
That trait was really only added as a helper for the `#[wasm_bindgen]` macro, and it was added as a separate internal trait because some users define custom `TryFrom<MyStruct> for JsValue` that we don't want to conflict with.

I'd say for all these primitive types a proper `TryFrom` would be more useful, as macro doesn't need those implementations, but our users might want them, so let's just add that as a proper public API instead? We already implement `TryFrom<JsValue>` for some types, so this would just need to fill the gap for the missing ones.

guybedford:
@RReverser I follow your logic - but we do also need a blanket trait that can be used for dynamic checks in runtime type generics, and having either trait only apply partially to wasm bindgen types wouldn't solve for this use case. Since we made the decision to have an internal `TryFromJsValue` trait, I think we should aim to make it comprehensive. If we want to expose `TryFrom` for primitives we could rather do that in addition for external users - this trait is an internal feature though and that is the use case for this PR.

src/convert/traits.rs
```diff
@@ -342,7 +342,7 @@ impl<T: WasmAbi> WasmRet<T> {
 /// documentation for more details.
 pub trait TryFromJsValue: Sized {
     /// The type returned in the event of a conversion error.
-    type Error;
+    type Error: core::fmt::Debug;
```
Spxg: Is this not necessary?

src/lib.rs
```diff
@@ -927,7 +955,46 @@ impl AsRef<JsValue> for JsValue {
     }
 }
 
-macro_rules! numbers {
+fn to_uint_32(v: JsValue) -> Result<u32, JsValue> {
+    f64::try_from(v).map(|n| {
+        if !n.is_finite() || n == 0.0 {
```
Spxg: It looks like can use `is_infinite`, `n == 0.0` should not be needed

src/lib.rs
```diff
@@ -1011,13 +1109,45 @@ macro_rules! try_from_for_num128 {
...
 }
 
-try_from_for_num128!(i128, i64);
-try_from_for_num128!(u128, u64);
+num128!(i128, i64);
 
-big_numbers! { i64 u64 i128 u128 }
+num128!(u128, u64);
+
+impl TryFromJsValue for () {
```
Spxg, line 1130: Honestly this is a little bit weird, is `undefined` considered to be `()` before?

src/lib.rs
Spxg: If it is infinite, why does it return 0u32? Is there any standard rule?

RReverser (approved):

src/convert/traits.rs
guybedford: Removed.

src/lib.rs
guybedford: The semantics being followed here are https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toint32. The goal is to mirror the static semantics which build on top of the Wasm `ToWebAssemblyValue` semantics so we are building out a runtime dynamic version of the same semantics of the static type system.

guybedford: Removed the unnecessary zero check though from the spec.

src/lib.rs
guybedford: A function with no return in Rust returns the unit type and in JavaScript returns undefined.

Therefore, under the semantics of providing a runtime isomorphism of the static semantics, the unit type can be dynamically converted from undefined to mirror that.

guybedford: I've more clearly documented the semantics in the trait definition to clarify these concepts, further suggestions welcome.

PR #4715, opened 2025-10-03
Fix warning when using `#[wasm_bindgen(wasm_bindgen=xxx)]` on struct

Spxg:
It is already used before being passed to `BindgenedStruct`

https://github.com/wasm-bindgen/wasm-bindgen/blob/04a4f52146f60815cf75a971ea536259625215ae/crates/macro-support/src/lib.rs#L34

### DIFF
--- a/tests/struct_derive.rs
+++ b/tests/struct_derive.rs
@@ -1,0 +1,6 @@
+#![deny(warnings)]
+
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen(wasm_bindgen = wasm_bindgen)]
+struct Foo;
--- a/tests/struct_derive.rs
+++ b/tests/struct_derive.rs
@@ -1,6 +1,0 @@
-#![deny(warnings)]
-
-use wasm_bindgen::prelude::*;
-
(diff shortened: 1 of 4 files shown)


RReverser:
Please add a repro test to ui-tests that would fail before this fix & pass after.

Spxg:
> Please add a repro test to ui-tests that would fail before this fix & pass after.

I don't know if it makes sense since it doesn't fail, just warn. I guess I just need to add a test and deny warning.

RReverser:
It does, that's what unused-attributes.rs already verifies: https://github.com/wasm-bindgen/wasm-bindgen/blob/04a4f52146f60815cf75a971ea536259625215ae/crates/macro/ui-tests/unused-attributes.rs / output in https://github.com/wasm-bindgen/wasm-bindgen/blob/04a4f52146f60815cf75a971ea536259625215ae/crates/macro/ui-tests/unused-attributes.stderr.

It already has relevant `deny()`, you just need to add your test case there.

tests/struct_derive.rs
```diff
@@ -0,0 +1,6 @@
+#![deny(warnings)]
```
RReverser: You can now remove this file.

PR #4717, opened 2025-10-03
Fix atomics test in latest nightly

Spxg:
Related PR: https://github.com/rust-lang/rust/pull/147225

Before new target or thread feature are added, adding a bunch of arguments should be the only way

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
         include:
           - name: "no_std"
           - name: "no_std + atomics"
-            flags: --config "target.wasm32-unknown-unknown.rustflags='-Ctarget-feature=+atomics -Clink-args=--shared-memory -Clink-args=--max-memory=1073741824 -Clink-args=--import-memory -Clink-args=--export=__wasm_init_tls -Clink-args=--export=__tls_size -Clink-args=--export=__tls_align -Clink-args=--export=__tls_base'"
+            flags: --config "target.wasm32-unknown-unknown.rustflags='-Ctarget-feature=+atomics'"
     name: Clippy (Wasm + ${{ matrix.name }})
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
         include:
           - name: "no_std"
           - name: "no_std + atomics"
-            flags: --config "target.wasm32-unknown-unknown.rustflags='-Ctarget-feature=+atomics'"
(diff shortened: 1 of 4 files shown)


RReverser (changes requested):
I think for at least some of these tests just atomics should be enough.

E.g. for the general `clippy_wasm_no_std` or reference tests we don't need actual multithreading.

I'd rather only add those flags where we really need them, which should only be integration tests in examples - and you also need to add those flags to raytrace-parallel.

.github/workflows/main.yml
```diff
@@ -130,9 +130,9 @@ jobs:
       matrix:
         envs:
           - name: mvp
-            rustflags: -C target-cpu=mvp
+            rustflags: -Ctarget-cpu=mvp
           - name: atomics
-            rustflags: -C target-feature=+atomics
+            rustflags: -Ctarget-feature=+atomics -Clink-args=--shared-memory -Clink-args=--max-memory=1073741824 -Clink-args=--import-memory -Clink-args=--export=__wasm_init_tls -Clink-args=--export=__tls_size -Clink-args=--export=__tls_align -Clink-args=--export=__tls_base
```
RReverser, line 135: This can also be reverted, right?

crates/cli/tests/wasm-bindgen/reference.rs
```diff
@@ -143,7 +143,17 @@ fn runtest_targets_atomics() -> Result<()> {
     runtest_with_opts(test, Some("atomics"), |command| {
         command
             .env("RUSTUP_TOOLCHAIN", "nightly")
-            .env("RUSTFLAGS", "-C target-feature=+atomics")
+            .env(
```
RReverser, line 146: And this.

RReverser: Ah hm maybe not if we want to get the same snapshot.

.github/workflows/main.yml
Spxg: https://github.com/wasm-bindgen/wasm-bindgen/actions/runs/18212336137/job/51855210246
maybe not?

crates/cli/tests/wasm-bindgen/reference.rs
Spxg: yeah, this should update reference.js?
https://github.com/wasm-bindgen/wasm-bindgen/actions/runs/18212336137/job/51855210242

.github/workflows/main.yml
RReverser: Oh right, futures test uses multithreading.

Ideally the correct fix here is to switch our checks to "is memory shared" not "does Wasm use atomics in target-features", but we can do that in follow-up PR.

RReverser (approved):
Ok yeah let's do it as-is then, we can update our transformation logic in separate PRs or wait for Rust to gain a new target-feature first.

PR #4720, opened 2025-10-05
Establish clearer ABI trait hierarchy

RReverser:
This is just somewhat cleaning up relations between WasmAbi and other traits, as so far the choosing of one over others has been somewhat random - e.g. `i128` implementing `WasmAbi` that converts to a tuple that is _the actual_ ABI or some `Option<T>` and `Result<T, u32>` implementing `WasmAbi` but other `Option<T>` and `Result<T, E>` implementing `{From, Into, Return}WasmAbi`.

After this PR the trait relationship is a bit more streamlined:

- Any primitive or a tuple of primitives (up to 4 `WasmPrimitive`s) blanket-implements `WasmAbi`.
- Anything that is `WasmAbi` naturally blanket-implements `FromWasmAbi`/`IntoWasmAbi`.

This also opens up a path to using more precise ABIs for slices and such in future PRs.


RReverser:
Converting to draft because this might prevent some simplifications in the generics PR that we discussed with @guybedford, going to look at it again after that one lands.

PR #4724, opened 2025-10-08
feat: chaining setters

guybedford:
As part of working through canonical bindgen, a case that came up recently was supporting a builder pattern for imported JS objects. This implements chaining setters using the existing setter bindgen as the base, then adding a new special case for functions defined of the form:

```rust
extern "C" {
  type ChainingFoo;

  #[wasm_bindgen(method, setter)]
  fn set_name(this: ChainingFoo, name: String) -> ChainingFoo;
}
```

Validation is included that the first argument must be the own type and the return value must also be the own type, then when used via this type-based opt-in, this enables builder patterns on imported JS objects like:

```rust
let foo = foo::new()
  .set_a(a)
  .set_b(b);
```

At a bindgen level, the JS-Wasm bindgen remains equivalent to a normal setter - there is no ownership passing to JS at all, and it is treated like a `set_name(this: &ChainingFoo, name: String)` generation. But, the Rust-based wrapper gets the builder sugar to support more ergonomic usage.


RReverser:
> One nit is that a few cleanups here seem unrelated to the feature, I usually prefer them to be split out into their own PR so that it's easier to review in isolation.

Created a quick https://github.com/wasm-bindgen/wasm-bindgen/pull/4728 which should subsume a few of those.

RReverser:
> Nice, looks pretty clean. I'll just hold off the approval until the CI passes - it's a bit suspicious that it breaks just one particular demo.

Hm I see it fails outside of your PR too, weird but seems unrelated.

guybedford:
I've added a test and docs for exported type setter chaining from the Rust side as well here.

guide/src/reference/attributes/on-rust-exports/getter-and-setter.md
```diff
@@ -62,3 +62,72 @@ The name for a `getter` is by default inferred from the function name it's
 attached to. The default name for a `setter` is the function's name minus the
 `set_` prefix, and if `set_` isn't a prefix of the function it's an error to not
 provide the name explicitly.
+
+## Chaining Setters
+
+Setters can support method chaining by taking `&mut self` and returning `&mut Self`:
```
RReverser, line 68: Is it possible to use by-value returns here too? Both styles are common in Rust, and in theory for Rust setters nothing prevents us from supporting both?

guybedford: I spent some time trying to achieve that but wasn't able to figure out the right Rust hacks to make it work without hitting move bugs.

It would be a great follow-on if it is possible though.

RReverser: Fair enough.

guybedford:
Closing this to rather encourage builder pattern wrappers for wasm bindgen structs, via a multi-field constructor approach.

We should then focus on optimizing the multi-field constructor case as needed by builder pattern wrappers.

PR #4733, opened 2025-10-09
Fix Closure interaction with Pin

RReverser:
This is hell of a bug-at-a-distance, and I almost suspect it might be a Rust bug since for all other intents and purposes `Closure<T>` was still considered `Sized` even without this fix.

I'll look into it a bit more, but for now... Fixes #4732.

PR #4735, opened 2025-10-10
Add CI check for subtle semver breakages

RReverser:


RReverser:
Hm unfortunately not as useful as I had hoped, looks like rustdoc-json doesn't catch the subtle `Unpin` requirement change for `Closure` from #4733.

PR #4738, opened 2025-10-14
Sort custom TypeScript sections

nickbabcock:
Custom TypeScript sections defined via `#[wasm_bindgen(typescript_custom_section)]`
are now sorted alphabetically before being emitted. This ensures consistent output
across different compilation runs regardless of codegen units or other compiler
options that may affect the order in which sections are encountered.

**breaking change** for users that rely upon custom section declaration order.
For instance if the original types generated were:

```ts
type EntityId = string;
type EntityId = number;
```

This commit will change the order to be:

```ts
type EntityId = number;
type EntityId = string;
```

And now typescript will resolve `EntityId` to be a `string` instead of a number
and could cause the typescript compiler to fail typechecking.
    
Redeclaration of types is not a good practice and it's dubious if relying on
this behavior is something that could/should be guaranteed.


crates/cli-support/src/js/mod.rs
```diff
@@ -2781,6 +2793,16 @@ wasm = wasmInstance.exports;
             self.generate_adapter(id, adapter, instrs, kind)?;
         }
 
+        // Sort and emit TypeScript function exports for deterministic output.
+        // This ensures that the TypeScript declarations are consistent regardless
+        // of compiler settings (like codegen units) while preserving the JS export
+        // order as defined in the source code.
+        self.typescript_function_exports
+            .sort_by(|a, b| a.name.cmp(&b.name));
```
Spxg: I think unstable sort can be used

```rust
sort_unstable_by(|a, b| a.name.cmp(&b.name));
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2796,7 +2818,14 @@ wasm = wasmInstance.exports;
             self.generate_struct(s)?;
         }
 
-        self.typescript.push_str(&self.aux.extra_typescript);
+        // Emit custom TypeScript sections, sorted for deterministic output
+        // across different compilation units
+        let mut custom_sections: Vec<_> = self.aux.extra_typescript.iter().collect();
+        custom_sections.sort();
```
Spxg:
```rust
custom_sections.sort_unstable();
```

crates/cli-support/src/js/mod.rs
Spxg: I think unstable sort can be used

```rust
sort_unstable_by(|a, b| a.name.cmp(&b.name))
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2796,7 +2818,14 @@ wasm = wasmInstance.exports;
...
+        for section in custom_sections {
+            self.typescript.push_str(section);
+            self.typescript.push_str("\n\n");
+        }
```
Spxg:
```rust
let mut custom_sections: Vec<_> = self
    .aux.extra_typescript
    .iter()
    .map(|s| s.as_str())
    .collect();
custom_sections.sort_unstable();
self.typescript.push_str(&custom_sections.join("\n\n"));
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2968,11 +2997,17 @@ wasm = wasmInstance.exports;
...
-                            self.typescript.push_str(name);
-                            self.typescript.push_str(ts_sig);
-                            self.typescript.push_str(";\n");
+                            // Collect TypeScript function declarations for sorting
+                            let mut ts_decl = String::new();
+                            ts_decl.push_str(&ts_docs);
+                            ts_decl.push_str("export function ");
+                            ts_decl.push_str(name);
+                            ts_decl.push_str(ts_sig);
+                            ts_decl.push_str(";\n");
```
Spxg: It might be better to use the `format!` macro

Spxg (changes requested):
LGTM, I have some nitpicky suggestions. 
And I'm not very familiar with TypeScript and still need approve from other reviewers.

crates/cli-support/src/js/mod.rs
nickbabcock: I think you're right that an unstable sort can be used here. I'm happy to switch to it, though it looks like the wasm-bindgen codebase has shied away from unstable sorts so far (idk if there's reason, just an observation).

crates/cli-support/src/js/mod.rs
nickbabcock: `join` would create a potentially large and ephemeral string, right?

I figured I'd keep the style of `push_str` relatively the same, but I'm happy to update as well 🙂

crates/cli-support/src/js/mod.rs
```diff
@@ -3114,15 +3143,21 @@ wasm = wasmInstance.exports;
                 match &export.kind {
                     AuxExportKind::Function(name) | AuxExportKind::FunctionThis(name) => {
                         if let Some(ts_sig) = ts_sig {
-                            self.typescript.push_str(&ts_docs);
+                            // Collect TypeScript function declarations for sorting
+                            let mut ts_decl = String::new();
+                            ts_decl.push_str(&ts_docs);
                             if export.js_namespace.is_none() {
```
guybedford: Is it maybe time to just consider outputting all functions in sort-order? This is something that is consistently an issue with toolchain changes causing output ordering changes.

Would value feedback from others here?

nickbabcock: I didn't see additional commentary about avoiding sorting in https://github.com/wasm-bindgen/wasm-bindgen/pull/4190 , but I preserved it if the author felt strongly enough to add a comment.

I'm in favor of the sort everything effort (and can close this PR in favor of it, or have this PR be the first step towards sorted output).

guybedford: I think we almost have consensus for a sort everything effort? How would you feel about taking on a wider scope here? We can make sure to get full core team support first to avoid wasted effort if so.

nickbabcock: Sounds good, I'd be happy to do it

RReverser: +1 from me, I assume it won't add any noticeable overhead so sounds like a cheap way to ensure our output is deterministic in the future.

You'll want to remove some of the normalisation from the reference tests if you add it to the main codegen path though, to avoid duplicate logic.

nickbabcock:
https://github.com/wasm-bindgen/wasm-bindgen/pull/4800 expands the scope of this PR to cover js, d.ts, and wasm files. I'll keep this PR open in case there is a change of heart while the other PR is in review.

nickbabcock:
@Spxg thanks for the review, this PR is in a state to be reviewed again.

Thanks to the work of others in the last month, most of the typescript output for global exports is now deterministic already. Just custom sections remain, so this PR now targets custom sections.

This PR does not touch internal exports (like bindgen generated closures) (and if these have official terminology, let me know).

I don't want perfect to be the enemy of good so this PR can be merged independently of PR #4800 which has a more lofty goal.

New PR description copied below:

---

Custom TypeScript sections defined via `#[wasm_bindgen(typescript_custom_section)]`
are now sorted alphabetically before being emitted. This ensures consistent output
across different compilation runs regardless of codegen units or other compiler
options that may affect the order in which sections are encountered.

**breaking change** for users that rely upon custom section declaration order.
For instance if the original types generated were:

```ts
type EntityId = string;
type EntityId = number;
```

This commit will change the order to be:

```ts
type EntityId = number;
type EntityId = string;
```

And now typescript will resolve `EntityId` to be a `string` instead of a number
and could cause the typescript compiler to fail typechecking.
    
Redeclaration of types is not a good practice and it's dubious if relying on
this behavior is something that could/should be guaranteed.

Spxg (approved):
Thanks

Issue #4742, opened 2025-10-15
`instantiateStreaming` used inefficiently?


kylebarron:
I was reading the generated binding code, and I was confused about some parts of the WebAssembly module instantiation.

Reading MDN docs:

> The quickest, most efficient way to fetch a Wasm module is using the newer [`WebAssembly.instantiateStreaming()`](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/instantiateStreaming_static) method, which can take a fetch() call as its first argument, and will handle fetching, compiling, and instantiating the module in one step, accessing the raw byte code as it streams from the server:

As an example it gives:

> ```js
> WebAssembly.instantiateStreaming(fetch("simple.wasm"), importObject).then(
>  (results) => {
>    // Do something with the results!
>  },
>);
> ```

But by my reading, wasm-bindgen does **not** use this effectively. In these lines:

https://github.com/wasm-bindgen/wasm-bindgen/blob/ac51055a4c39fa0affe02f7b63fb1d4c9b3ddfaf/crates/cli-support/src/js/mod.rs#L966-L970

wasm-bindgen will **`await`** the fetch before passing it to `__wbg_load` (which internally calls `WebAssembly.instantiateStreaming`). I'm not a web expert but doesn't this remove the primary benefit of `instantiateStreaming`, to be able to compile as the fetch comes in?

guybedford:
Fetch returns as soon as the response headers are received, so streaming will still be in progress on the returned response when this happens. So this should actually be an unnecessary optimization to add, since no body is available to parse early until the fetch promise resolves anyway.

kylebarron:
Oh right, that makes sense. Happy to close if you'd like

guybedford:
Sure, yeah I think if it's not broken there is no need to fix it, but thank you for the PR anyway.

PR #4743, opened 2025-10-15
Don't await fetch before passing to `instantiateStreaming`

kylebarron:
Closes https://github.com/wasm-bindgen/wasm-bindgen/issues/4742

This is as-yet untested, because I don't know how to test it 😄


guybedford:
Closing per discussion in #4742.

PR #4744, opened 2025-10-16
feat: support js_namespace on exported types

guybedford:
This adds support for the existing `js_namespace` attributes that works for imported types, to also work for exported types using the same corresponding semantics.

This allows defining arbitrarily nested interfaces instead of just being able to define top-level functions and classes.

Careful support is also added for the `"default"` case, so that it is possible to write a default export library like:

```
#[wasm_bindgen(js_namespace = "default")]
pub struct Counter {
    value: i32,
}
```

corresponding to:

```
class Counter { ... }

export default {
  Counter
}
```

TypeScript declaration output is also updated to support the new generation cases.

While this refactors the output generation a bit, output is kept fully backwards compatible with changes only visible when using the feature. The refactoring may even form a core refactoring of export collection in general for future versions though.

### DIFF
--- a/crates/cli/tests/reference/js-namespace-export.bg.js
+++ b/crates/cli/tests/reference/js-namespace-export.bg.js
@@ -417,7 +417,7 @@ export const utils = {
     },
 };
 
-export function __wbg___wbindgen_throw_451ec1a8469d7eb6(arg0, arg1) {
+export function __wbg___wbindgen_throw_b855445ff6a94295(arg0, arg1) {
     throw new Error(getStringFromWasm0(arg0, arg1));
 };
 
--- a/crates/cli/tests/reference/js-namespace-export.wat
+++ b/crates/cli/tests/reference/js-namespace-export.wat
@@ -12,7 +12,7 @@
   (type (;10;) (func (param i32 i32 i32 i32) (result i32 i32)))
   (type (;11;) (func (param i32 f64)))
   (type (;12;) (func (param f64 f64) (result f64)))
-  (import "./reference_test_bg.js" "__wbg___wbindgen_throw_451ec1a8469d7eb6" (func (;0;) (type 5)))
+  (import "./reference_test_bg.js" "__wbg___wbindgen_throw_b855445ff6a94295" (func (;0;) (type 5)))
   (import "./reference_test_bg.js" "__wbindgen_init_externref_table" (func (;1;) (type 0)))
   (func $add (;2;) (type 6) (param i32 i32) (result i32))
   (func $multiply (;3;) (type 6) (param i32 i32) (result i32))


justfile
hoodmane, line 1: Should changes to justfile be split into a separate PR? They seem unrelated.

guybedford: Thanks, I've separated these and included in the separate PR https://github.com/wasm-bindgen/wasm-bindgen/pull/4753.

crates/macro-support/src/parser.rs
```diff
@@ -487,14 +488,18 @@ impl Parse for BindgenAttr {
...
-                    Err(_) => {
-                        let ident = input.parse::<AnyIdent>()?.0;
-                        (vec![ident.to_string()], vec![ident.span()])
+                    // Try parsing as a string literal, then fall back to identifier
+                    Err(_) => match input.parse::<syn::LitStr>() {
+                        Ok(str) => (vec![str.value()], vec![str.span()]),
+                        Err(_) => {
+                            let ident = input.parse::<AnyIdent>()?.0;
+                            (vec![ident.to_string()], vec![ident.span()])
+                        }
```
hoodmane, line 498: So it looks like `js_namespace = "$$$"` wasn't supported previously and it is now?

crates/macro-support/src/parser.rs
```diff
@@ -1568,10 +1577,22 @@ impl MacroParse<&ClassMarker> for &mut syn::ImplItemFn {
             let kind = operation_kind(&opts);
             ast::MethodKind::Operation(ast::Operation { is_static, kind })
         };
+
+        // Validate that js_namespace is not used on methods
```
hoodmane, line 1598: How do I tell that we are working on methods here? Is it:
```
impl MacroParse<&ClassMarker> for &mut syn::ImplItemFn {
```
and `syn::ImplItemFn` is a `fn` node?

crates/macro-support/src/parser.rs
```diff
@@ -1568,10 +1577,22 @@ impl MacroParse<&ClassMarker> for &mut syn::ImplItemFn {
...
+        if let Some((_, span)) = opts.js_namespace() {
+            return Err(Diagnostic::span_error(
+                span[0],
+                "`js_namespace` cannot be used on methods, getters, setters, or static methods. \
+                Use `js_namespace` on the exported struct definition instead to put the entire class in a namespace.",
+            ));
+        }
+
+        let js_namespace = opts.js_namespace().map(|(ns, _)| ns.0);
```
hoodmane: Didn't we just check that `opts.namespace()` is `None`? So won't we always set `js_namespace` to `None` here?

crates/macro-support/src/parser.rs
```diff
@@ -1568,10 +1577,22 @@ impl MacroParse<&ClassMarker> for &mut syn::ImplItemFn {
...
         program.exports.push(ast::Export {
             comments,
             function,
             js_class: Some(js_class.to_string()),
+            js_namespace,
```
hoodmane: Should it be:
```suggestion
            js_namespace: None,
```

crates/macro-support/src/parser.rs
guybedford: Yes, exactly.

crates/macro-support/src/parser.rs
guybedford: Yes, because `impl` blocks are specifically for struct functions which are by definition methods or static methods.

crates/macro-support/src/parser.rs
guybedford: Added, thanks.

crates/macro-support/src/parser.rs
guybedford: Yes exactly, added per below.

PR #4749, opened 2025-10-17
 Mention wasm-stack-trace as an option for better stack traces

juancampa:
We just published a library that hooks up to `prepareStackTrace` to enhance JavaScript `Error` objects with DWARF data (using gimli compiled to wasm). I thought it was worth mentioning in the wasm-bindgen guide for folks looking to improve their debugging experience.

It's pretty easy to use (Add a `<script>` tag to your HTML). This must run _before_ Module compilation so it can't be done in a rust crate.

To illustrate the difference:

Screenshot 2025-10-17 at 12 43 26

### DIFF
--- a/guide/src/reference/debug-info.md
+++ b/guide/src/reference/debug-info.md
@@ -3,9 +3,13 @@
 Currently, debug information in the form of DWARF, is stripped away from the output module.
 To keep it, use [`--keep-debug`](cli.html#--keep-debug) with the CLI.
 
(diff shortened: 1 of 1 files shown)


guybedford:
I recently added a community projects page to the guide in https://github.com/wasm-bindgen/wasm-bindgen/pull/4781.

Perhaps update the PR to include it there instead, as a more suitable location for it.

PR #4750, opened 2025-10-17
Ref variation for TryFromJsValue

guybedford:
Currently `TryFromJsValue` will always consume the `JsValue` like a typical `TryFrom` implementation.

But by extending it to have a ref variation we can support ref conversions as well, which is useful to avoid unnecessary `JsValue::clone()` operations when treating `TryFromJsValue` as a runtime conversion primitive (the decision to use `TryFromJsValue` itself as a runtime type conversion primitive based on not wanting to reintroduce another trait into the existing systems if we could use an existing trait).

This implements the ref variant and generalizes on that type default where possible.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -889,7 +889,7 @@ impl TryFromJsValue for bool {
 
 impl TryFromJsValue for char {
     fn try_from_js_value_ref(value: &JsValue) -> Option<Self> {
-        let s= value.as_string()?;
+        let s = value.as_string()?;
         if s.len() == 1 {
             Some(s.chars().nth(0).unwrap())
         } else {
@@ -1086,7 +1086,7 @@ macro_rules! num128 {
             fn try_from_js_value_ref(v: &JsValue) -> Option<$ty> {
                 // Truncate the bigint to 64 bits, this will give us the lower part.
                 // The lower part must be interpreted as unsigned in both i128 and u128.
-                let lo = __wbindgen_bigint_get_as_i64(&v)? as u64;                
+                let lo = __wbindgen_bigint_get_as_i64(&v)? as u64;
                 // Now we know it's a bigint, so we can safely use `>> 64n` without
                 // worrying about a JS exception on type mismatch.
                 let hi = v >> JsValue::from(64_u64);


guybedford:
> This adds an extra clone and drop to every try_from_js_value. What's the performance of clone and drop like? Ideally they shouldn't involve a call out to JS. But maybe optimizing that is a separate issue.

I don't believe this does add an extra clone and drop in the general case - are you referring to the clone applied for the exported Rust struct code generation?  That is the one case that is affected here, for the saving that the ref form otherwise avoids the need for cloning by consumers of the trait when possible.

I could possibly look at avoiding that case for structs though as well by implementing both trait functions so the clone is only necessary in the ref case.

> Tests?

Will add tests further, thanks for noting.

guybedford:
I've pushed up a basic test here, further feedback welcome on the clone case.

hoodmane:
Personally, I'm fine with merging as is and improving clone later. But I'm not a very big stakeholder in the project.

src/convert/traits.rs
```diff
@@ -350,9 +350,11 @@ impl<T: WasmAbi> WasmRet<T> {
...
-    /// The type returned in the event of a conversion error.
-    type Error;
+    /// Performs the conversion.
+    fn try_from_js_value(value: JsValue) -> Result<Self, JsValue> {
+        Self::try_from_js_value_ref(&value).ok_or(value)
+    }
 
     /// Performs the conversion.
-    fn try_from_js_value(value: JsValue) -> Result<Self, Self::Error>;
+    fn try_from_js_value_ref(value: &JsValue) -> Option<Self>;
```
Spxg, line 359: Why is `Option<T>` returned instead of `Result<T>`?

src/convert/traits.rs
```diff
@@ -350,9 +350,11 @@ impl<T: WasmAbi> WasmRet<T> {
 /// stability guarantees** are provided. Use at your own risk. See its
 /// documentation for more details.
 pub trait TryFromJsValue: Sized {
-    /// The type returned in the event of a conversion error.
-    type Error;
```
Spxg, line 354: Maybe this can be retained. Although it is an internal API, we should consider the difficulty of future maintenance. Can we assume that the JsValue Error will be returned in the future?

Spxg (changes requested):

src/convert/traits.rs
guybedford: This is the same pattern as `dyn_ref` for JS cast - the object itself can be carried in the error as an own type for own casts, but reference casts can't do that without clone so option is used instead.

src/convert/traits.rs
guybedford: In all cases, the internal error is `JsValue` so this seems like it's ended up being an unnecessary abstraction to me. It also hinders usage having to have this defined in all cases for derived traits, but if there are important cases we can potentially figure out a path to retaining. I was just hitting a bunch of issues with it and it seemed like it's proven to be an unnecessary abstraction given it's unused for the internal type.

crates/macro-support/src/codegen.rs
```diff
@@ -412,6 +409,9 @@ impl ToTokens for ast::Struct {
                         }
                     }
                 }
+                fn try_from_js_value_ref(value: &#wasm_bindgen::JsValue) -> #wasm_bindgen::__rt::core::option::Option<Self> {
+                    Self::try_from_js_value(value.clone()).ok()
```
Spxg, line 413: Maybe we should put the implementation of `try_from_js_value` here to avoid unnecessary clone

src/lib.rs
```diff
@@ -1100,29 +1081,26 @@ macro_rules! num128 {
...
             fn try_from(v: JsValue) -> Result<Self, JsValue> {
+                Self::try_from_js_value(v)
+            }
+        }
+
+        impl TryFromJsValue for $ty {
+            // This is a non-standard Wasm bindgen conversion, supported equally
+            fn try_from_js_value_ref(v: &JsValue) -> Option<$ty> {
                 // Truncate the bigint to 64 bits, this will give us the lower part.
                 let lo = match __wbindgen_bigint_get_as_i64(&v) {
```
Spxg:
```rust
let lo = __wbindgen_bigint_get_as_i64(&v)? as u64;
```

src/lib.rs
```diff
@@ -887,29 +882,22 @@ impl From<bool> for JsValue {
...
+        value.as_bool()
     }
 }
 
 impl TryFromJsValue for char {
-    type Error = JsValue;
-
-    fn try_from_js_value(value: JsValue) -> Result<Self, Self::Error> {
+    fn try_from_js_value_ref(value: &JsValue) -> Option<Self> {
         match value.as_string() {
```
Spxg:
```rust
let s = value.as_string()?;
```
Let's reduce a branch.

tests/wasm/try_from_js_value.rs
```diff
@@ -70,7 +70,7 @@ fn u64_try_from_js_value() {
     assert!(u64::try_from_js_value(JsValue::TRUE).is_err());
     assert!(u64::try_from_js_value(JsValue::from_str("45")).is_err());
     assert!(u64::try_from_js_value(JsValue::NULL).is_err());
-    assert_eq!(u64::try_from_js_value(JsValue::from(-1_i64)), Ok(u64::MAX));
+    assert!(u64::try_from_js_value(JsValue::from(-1_i64)).is_err());
```
Spxg, line 73: This looks like a breaking change

src/convert/traits.rs
Spxg: Well, fortunately it is an internal trait, LGTM

tests/wasm/try_from_js_value.rs
Spxg: It looks more like a bugfix, but it may affect many crates...

Spxg: It's fine, introduced by https://github.com/wasm-bindgen/wasm-bindgen/pull/4714, not released yet

guybedford: Yes exactly, it's actually correcting the semantics.

crates/macro-support/src/codegen.rs
guybedford: I think this would be a good refactoring to have, but it would need to be a very very careful refactoring of the current code as the current codepath is entirely based on value semantics with the memory handling.

I think this would be best to treat as a careful follow-on to avoid the risk of introducing any bugs.

src/lib.rs
guybedford: Much nicer indeed.

crates/macro-support/src/codegen.rs
Spxg: Sorry, I don't understand what the problem is in moving the implementation of `try_from_js_value` down. The only place where value is used is 
```rust
let idx = #wasm_bindgen::convert::IntoWasmAbi::into_abi(&value);
```
```rust
fn try_from_js_value(value: #wasm_bindgen::JsValue) -> #wasm_bindgen::__rt::core::result::Result<Self> {
    Self::try_from_js_value_ref(&value).ok_or(value)
}

fn try_from_js_value_ref(value: &#wasm_bindgen::JsValue) -> #wasm_bindgen::__rt::core::option::Option<Self> {
    xxx
}
```

Spxg: Oh.. `mem::forget` is used

PR #4751, opened 2025-10-18
feat: JsValue.is_null_or_undefined()

guybedford:
Adds a new `.is_null_or_undefined()` method to `JsValue`, along with an optimized intrinsic.

`TryFromJsValue` for `Option<JsValue>` is then updated to use this intrinsic so that this is a single call and not two in the ABI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
 * Extended support for `TryFromJsValue` trait implementations.
   [#4714](https://github.com/wasm-bindgen/wasm-bindgen/pull/4714)
 
+* New `JsValue.is_null_or_undefined()` method and intrinsic.
+  [#4751](https://github.com/wasm-bindgen/wasm-bindgen/pull/4751)
+
 ### Fixed
 
 * Fixed multithreading JS output for targets `bundler`, `deno` and `module`.


evilpie:
Do note that `window.all == null` is true. Is that what you want?

daxpedda:
> Do note that `window.all == null` is true. Is that what you want?

If I'm understanding this right, `window.all` is an undefined property. If so, that would indeed be the intention here.

evilpie:
Sorry, I should have said [document.all](https://developer.mozilla.org/en-US/docs/Web/API/Document/all) of course.

Let me give an example why I think your implementation is problematic:
```javascript
console.log(document.all == null) // true
console.log(document.all === null || document.all === undefined) // -> false
```

That means changing
```rust
        if value.is_undefined() || value.is_null() {
```
to
```rust
        if value.is_null_or_undefined() {
```
just subtly changed the behavior.

daxpedda:
Ah, thank you!

@guybedford I think that's a valid concern. This should be an easy fix, right?

guybedford:
I honestly don't think supporting `document.all` is a requirement for Wasm Bindgen.

daxpedda:
> I honestly don't think supporting `document.all` is a requirement for Wasm Bindgen.

My understanding is that this isn't about `document.all`, its just an example.
Apparently something can be `true` as `== null` but be false with `=== null`. So using `== null` to account for `null` and `undefined` seems to be inconsistent in that case.

Though I'm a bit surprised why `document.all == null` is `true` in the first place. The variable is defined and contains an object.

But please correct me if I'm wrong.

guybedford:
`document.all` the only case of this happening, so only in the `document.all` case, `is_null_or_undefined(document.all)` would return true. The fix would be to change to `=== null || === undefined`. I'm open to taking that fix as well here if someone wants to create the PR though.

daxpedda:
> `document.all` the only case of this happening, [..]

Do you by any chance know why? So we can evaluate what else might be affected by this.
@evilpie.

If it really only affects `document.all`, I agree its not worth the code bloat, but should be worth a doc comment.

PR #4752, opened 2025-10-18
feat: support OptionIntoWasmAbi and OptionFromWasmAbi for JsValue

guybedford:
Splitting out another feature here from the generics work for individual review, this implements `Option<JsValue>` as an ABI input and output type, where `null` or `undefined` is treated as the input None case, and `undefined` is treated as the output case.

This provides static behaviour that matches the runtime behaviour of `TryFromJsValue` / `Into`.

This is based to https://github.com/wasm-bindgen/wasm-bindgen/pull/4751 so should land after that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,8 @@
 * New `JsValue.is_null_or_undefined()` method and intrinsic.
   [#4751](https://github.com/wasm-bindgen/wasm-bindgen/pull/4751)
 
-* Support for `Option<JsValue>` in ABI conversions.
+* Support for `Option<JsValue>` in function arguments and return.
+  [#4752](https://github.com/wasm-bindgen/wasm-bindgen/pull/4752)
 
 ### Fixed
 


hoodmane:
It would be worth considering converting undefined to None and distinguishing null. I don't have a strong opinion here but I recently made changes to Pyodide because people were mad that we converterted null and undefined both to None. OTOH if people care about the difference they can just use `JsValue` and inspect it directly, so maybe prioritizing the ergonomics here is better.

guybedford:
Thanks for the review @hoodmane and excellent point about null here actually. I've updated the ABI for both this and the dynamic falliable API to be based on not accepting null. This is valid since neither has been released yet. PTAL.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@
 * New `JsValue.is_null_or_undefined()` method and intrinsic.
   [#4751](https://github.com/wasm-bindgen/wasm-bindgen/pull/4751)
 
+* Support for `Option<JsValue>` in ABI conversions.
```
Spxg: Missing PR link, I think we need a unified format. And honestly the changelog is not very user friendly.

Spxg (changes requested):

CHANGELOG.md
guybedford: Thanks, I've updated this.

PR #4758, opened 2025-10-22
test runner retain fixtures environment variable

guybedford:
This one is quite useful for debugging - with this `KEEP_TEST_BUILD=1` when running the tests will retain the temporary fixtures folder, allowing to inspect errors for test generations, as was possible previously.

### DIFF
--- a/crates/cli/src/wasm_bindgen_test_runner.rs
+++ b/crates/cli/src/wasm_bindgen_test_runner.rs
@@ -211,7 +211,7 @@ fn rmain(cli: Cli) -> anyhow::Result<()> {
     // Support a WASM_BINDGEN_KEEP_TEST_BUILD=1 env var for debugging test files
     let tmpdir_path = if env::var("WASM_BINDGEN_KEEP_TEST_BUILD").is_ok() {
         let path = tmpdir.keep();
-        eprintln!(
+        println!(
             "Retaining temporary build output folder: {}",
             path.to_string_lossy()
         );


crates/cli/src/wasm_bindgen_test_runner.rs
```diff
@@ -208,6 +208,18 @@ fn rmain(cli: Cli) -> anyhow::Result<()> {
 
     let tmpdir = tempfile::tempdir()?;
 
+    // Support a KEEP_TEST_BUILD=1 env var for debugging test files
+    let tmpdir_path = if env::var("KEEP_TEST_BUILD").is_ok() {
+        let path = tmpdir.keep();
+        eprintln!(
```
Spxg: Why `eprintln!`?

crates/cli/src/wasm_bindgen_test_runner.rs
```diff
@@ -208,6 +208,18 @@ fn rmain(cli: Cli) -> anyhow::Result<()> {
 
     let tmpdir = tempfile::tempdir()?;
 
+    // Support a KEEP_TEST_BUILD=1 env var for debugging test files
+    let tmpdir_path = if env::var("KEEP_TEST_BUILD").is_ok() {
```
Spxg: `KEEP_TEST_BUILD` => `WASM_BINDGEN_KEEP_TEST_BUILD`

crates/cli/src/wasm_bindgen_test_runner.rs
guybedford: That is where debugging output always goes per posix conventions.

crates/cli/src/wasm_bindgen_test_runner.rs
guybedford: Added.

crates/cli/src/wasm_bindgen_test_runner.rs
Spxg: `println!` is used everywhere in this file, although I don't know if POSIX has this convention, it is not a problem to output debug separately to stderr for me.

Spxg (approved):

PR #4759, opened 2025-10-22
feat: reexport attribute

guybedford:
This adds support to all suitable import types the ability to declare a reexport name for that import, resulting in the export interface also including that export, matching ES module semantics for the same feature.

Optionally, a reexport name may be provided to support the `export { foo as bar }` style workflows. It also works with the existing import namespace feature so that it is possible to reexport deep values.

I understand wrapping is typically desired in these cases, but I've found it could be useful in certain scenarios, and it is a fairly trivial PR for a convenience feature.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -698,13 +698,10 @@ wasm = wasmInstance.exports;
 
         // Generate reexport statements
         for (adapter_id, export_name) in &self.aux.reexports {
-            let import_name = self.import_name(match self.aux.import_map.get(adapter_id) {
-                Some(
-                    AuxImport::Value(AuxValue::Bare(js))
-                    | AuxImport::Instanceof(js)
-                    | AuxImport::Static { js, .. },
-                ) => js,
-                Some(_) | None => continue,
+            let import_name = self.import_name(match &self.aux.import_map[adapter_id] {
+                AuxImport::Value(AuxValue::Bare(js)) => js,
+                AuxImport::Static { js, .. } => js,
+                _ => bail!("Unsupported re-export"),
             })?;
             let export_name = export_name.as_ref().unwrap_or(&import_name);
             self.export(export_name, None, ExportJs::Expression(&import_name), None)?;
@@ -2941,13 +2938,10 @@ wasm = wasmInstance.exports;
 
         // Ensure all imports for reexports are defined
         for adapter_id in self.aux.reexports.keys() {
-            self.import_name(match self.aux.import_map.get(adapter_id) {
-                Some(
-                    AuxImport::Value(AuxValue::Bare(js))
-                    | AuxImport::Instanceof(js)
-                    | AuxImport::Static { js, .. },
-                ) => js,
-                Some(_) | None => continue,
+            self.import_name(match &self.aux.import_map[adapter_id] {
+                AuxImport::Value(AuxValue::Bare(js)) => js,
+                AuxImport::Static { js, .. } => js,
+                _ => bail!("Unsupported re-export"),
             })?;
         }
 
--- a/crates/macro-support/src/parser.rs
+++ b/crates/macro-support/src/parser.rs
@@ -1160,7 +1160,7 @@ fn function_from_decl(
             syn::FnArg::Typed(mut c) => {
                 // typical arguments like foo: u32
                 replace_colliding_arg(&mut c);
-                c.ty = Box::new(replace_self(*c.ty));
+                *c.ty = replace_self(*c.ty);
                 arguments.push(c);
             }
             syn::FnArg::Receiver(r) => {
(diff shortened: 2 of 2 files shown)


crates/cli-support/src/wit/nonstandard.rs
```diff
@@ -34,6 +34,10 @@ pub struct WasmBindgenAux {
     /// A map from imported function id to what it's expected to import.
     pub import_map: HashMap<AdapterId, AuxImport>,
 
+    /// A map from AdapterId to the reexport name (if the import should be re-exported).
+    /// An empty string means reexport with the same name as the import.
```
hoodmane: Why the edge case?

crates/macro-support/src/parser.rs
```diff
@@ -1914,11 +1925,13 @@ impl MacroParse<ForeignItemCtx> for syn::ForeignItem {
                     let mut item: syn::ItemStatic =
                         syn::parse(v.into()).expect("only foreign functions/types allowed for now");
                     let item_opts = BindgenAttrs::find(&mut item.attrs)?;
+                    let reexport = item_opts.reexport().map(|s| s.clone().unwrap_or_default());
```
hoodmane: So we allow a raw `rexport` with no `=` and in that case `reexport` is set to `Some("")`.

crates/cli-support/src/wit/nonstandard.rs
hoodmane: Okay I see why. It would simplify things to get rid of this but perhaps it's not worth the effort...

guybedford: It might be cleaner to just use the actual string here, will update that.

crates/macro-support/src/parser.rs
guybedford: Actually sorry this is correct yes - the empty string case corresponds to no value, so it is like `#[wasm_bindgen(reexport)]` without the `reexport = ...`. The empty string value is the same as what HTML does for empty attributes.

guybedford:
On further thought the empty string behaviour is actually correct as the empty string means an empty attribute.

Spxg:
https://github.com/wasm-bindgen/wasm-bindgen/pull/4769
I submitted a PR to fix the cli MSRV.

crates/macro-support/src/parser.rs
Spxg:
```rust
let reexport = item_opts.reexport().cloned();
```

I don't know how HTML represents an empty string, but I think using `Option<Option<String>>` here would fully represent:

1. No `reexport` used

2. Only `reexport` used

3. `reexport = "xxx"` used

Using `Some("")` to "represent only `reexport` used" is a bit of a hack.

guybedford: For HTML like `<foo bar>`, `document.querySelector('foo').getAttribute('bar')` returns the empty string.

Spxg: > For HTML like `<foo bar>`, `document.querySelector('foo').getAttribute('bar')` returns the empty string.

Thanks! But I don't think it should affect our implementation of functionality using `Option<Option<String>>` or other methods, right? For me, `Some("")` is really not a good idea.

guybedford: Ok, I've added `Option<Option<String>>` here now.

crates/cli-support/src/js/mod.rs
```diff
@@ -2924,6 +2939,18 @@ wasm = wasmInstance.exports;
...
 
+        // Ensure all imports for reexports are defined
+        for adapter_id in self.aux.reexports.keys() {
+            self.import_name(match self.aux.import_map.get(adapter_id) {
+                Some(
+                    AuxImport::Value(AuxValue::Bare(js))
+                    | AuxImport::Instanceof(js)
+                    | AuxImport::Static { js, .. },
+                ) => js,
+                Some(_) | None => continue,
```
Spxg: When will this happen? `continue` or `return Err`?

crates/cli-support/src/js/mod.rs
```diff
@@ -695,6 +695,21 @@ wasm = wasmInstance.exports;
...
+
+        // Generate reexport statements
+        for (adapter_id, export_name) in &self.aux.reexports {
+            let import_name = self.import_name(match self.aux.import_map.get(adapter_id) {
+                Some(
+                    AuxImport::Value(AuxValue::Bare(js))
+                    | AuxImport::Instanceof(js)
+                    | AuxImport::Static { js, .. },
+                ) => js,
+                Some(_) | None => continue,
```
Spxg: Same

Spxg (changes requested):
I have some nit questions

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@
...
+  a new `#[wasm_bindgen(this)]` attribute.
+  [#4757](https://github.com/wasm-bindgen/wasm-bindgen/pull/4757)
+
+* Added `reexport` attribute for imports to support re-exporting imported types,
+  with optional renaming.
+  [#4759](https://github.com/wasm-bindgen/wasm-bindgen/pull/4759)
+
+* Added `js_namespace` attribute on exported types, mirroring the import
+  semantics to enable arbitrarily nested exported interface objects.
+  [#4744](https://github.com/wasm-bindgen/wasm-bindgen/pull/4744)
```
Spxg, line 18: I'd like a PR template to help with forgetting to write a changelog. https://github.com/wasm-bindgen/wasm-bindgen/pull/4773

crates/cli-support/src/js/mod.rs
guybedford: Thanks for pointing these out - these are validated ahead of time by the bindgen macro, but I've added an extra error here anyway.

PR #4760, opened 2025-10-22
deps: update to Walrus@0.24

guybedford:
Updates to Walrus 24, including support for const initializer expressions and exception handling.

### DIFF
--- a/crates/msrv/resolver/Cargo.toml
+++ b/crates/msrv/resolver/Cargo.toml
@@ -23,5 +23,5 @@ web-sys = { path = "../../web-sys", default-features = false }
 
 # Pinned sub-dependencies for MSRV
 bumpalo = "=3.12.0"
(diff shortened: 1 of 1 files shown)


crates/macro-support/Cargo.toml
```diff
@@ -17,7 +17,7 @@ strict-macro = []
 
 [dependencies]
 bumpalo = "3.0.0"
-proc-macro2 = "1.0"
+proc-macro2 = "=1.0.101"
```
Spxg: I think you are having problems with MSRV. 
To pin the version, just add it to `crates/msrv/lib/Cargo.toml`.

guybedford: Thanks for the tip, added.

Spxg: And this line can be revert

PR #4763, opened 2025-10-24
fix proc-macro msrv test

guybedford:
Seems this was a general MSRV CI regression.


guybedford:
Actually just rolled this in to https://github.com/wasm-bindgen/wasm-bindgen/pull/4758, since that's landing next anyway.

PR #4770, opened 2025-10-30
feat: Support unsafe-single-threaded-traits configuration

guybedford:
Adds an `unsafe-single-threaded-traits` configuration that auto-implements send and sync for JS types under the single threaded assumption.

This has come up for Cloudflare workers when we want to run Axum or other http router systems that require these traits and we end up doing a whole bunch of custom wrapping code, when really we just require this "hack".

Carefully noted safety and included a compiler error if building with threads.

### DIFF
--- a/tests/wasm/unsafe_send_sync.rs
+++ b/tests/wasm/unsafe_send_sync.rs
@@ -1,9 +1,9 @@
-//! Tests for the unsafe-single-threaded-traits feature
+//! Tests for the unsafe_single_threaded_traits cfg
 //!
-//! This test file verifies that when the feature is enabled, JS types
+//! This test file verifies that when the cfg is enabled, JS types
 //! become Send and Sync, allowing futures to be Send.
 
-#![cfg(feature = "unsafe-single-threaded-traits")]
+#![cfg(unsafe_single_threaded_traits)]
 
 use wasm_bindgen::prelude::*;
 use wasm_bindgen_futures::JsFuture;
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,7 +112,7 @@ jobs:
           - name: enable-interning
             run: cargo test --target wasm32-unknown-unknown --features enable-interning
           - name: unsafe-single-threaded-traits
-            run: cargo test --target wasm32-unknown-unknown --features unsafe-single-threaded-traits
+            run: RUSTFLAGS="--cfg unsafe_single_threaded_traits" cargo test --target wasm32-unknown-unknown
     name: "Run wasm-bindgen crate tests (${{ matrix.runs.name }})"
     runs-on: ubuntu-latest
     env:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,12 +28,6 @@ serde-serialize = ["serde", "serde_json", "std"]
 spans = []
 std = []
 
-# UNSAFE: Enables Send and Sync implementations for JS types (JsValue, Closure, etc.)
-# This is only safe in single-threaded WebAssembly environments where JavaScript
-# values cannot actually be accessed from multiple threads. Using this with
-# threading features (atomics, shared memory) is not supported and will fail to compile.
-unsafe-single-threaded-traits = ["wasm-bindgen-futures/unsafe-single-threaded-traits"]
-
 # Whether or not the `#[wasm_bindgen]` macro is strict and generates an error on
 # all unused attributes
 strict-macro = ["wasm-bindgen-macro/strict-macro"]
@@ -78,6 +72,7 @@ workspace = true
 unexpected_cfgs = { level = "warn", check-cfg = [
   'cfg(wasm_bindgen_unstable_test_coverage)',
   'cfg(xxx_debug_only_print_generated_code)',
+  'cfg(unsafe_single_threaded_traits)',
 ] }
 
 [workspace.lints.clippy]
--- a/crates/futures/Cargo.toml
+++ b/crates/futures/Cargo.toml
@@ -27,7 +27,6 @@ wasm-bindgen = { path = "../..", version = '=0.2.105', default-features = false 
 default = ["std"]
 futures-core-03-stream = ['futures-core']
 std = ["wasm-bindgen/std", "js-sys/std", "web-sys/std"]
-unsafe-single-threaded-traits = ["wasm-bindgen/unsafe-single-threaded-traits"]
 
 [target.'cfg(target_feature = "atomics")'.dependencies]
 web-sys = { path = "../web-sys", version = "=0.3.82", default-features = false, features = [
--- a/crates/futures/src/lib.rs
+++ b/crates/futures/src/lib.rs
@@ -114,11 +114,11 @@ pub struct JsFuture {
 // However, in single-threaded WebAssembly environments, this is safe because
 // the Rc cannot actually be accessed from multiple threads. The Inner struct
 // contains JsValue and Closure, which are both made Send by the
-// unsafe-single-threaded-traits feature.
-#[cfg(feature = "unsafe-single-threaded-traits")]
+// unsafe_single_threaded_traits cfg.
+#[cfg(unsafe_single_threaded_traits)]
 unsafe impl Send for JsFuture {}
 
-#[cfg(feature = "unsafe-single-threaded-traits")]
+#[cfg(unsafe_single_threaded_traits)]
 unsafe impl Sync for JsFuture {}
 
 impl fmt::Debug for JsFuture {
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -257,11 +257,11 @@ pub struct Closure<T: ?Sized> {
 // be accessed from multiple threads. The Send/Sync bounds on T ensure that the
 // closure's captured data would be safe to send in a truly multi-threaded context,
 // maintaining Rust's safety guarantees. The JsClosure reference itself is made
-// Send/Sync through the unsafe-single-threaded-traits feature.
-#[cfg(feature = "unsafe-single-threaded-traits")]
+// Send/Sync through the unsafe_single_threaded_traits cfg.
+#[cfg(unsafe_single_threaded_traits)]
 unsafe impl<T: ?Sized> Send for Closure<T> {}
 
-#[cfg(feature = "unsafe-single-threaded-traits")]
+#[cfg(unsafe_single_threaded_traits)]
 unsafe impl<T: ?Sized> Sync for Closure<T> {}
 
 fn _assert_compiles<T>(mut pin: core::pin::Pin<&mut Closure<T>>) {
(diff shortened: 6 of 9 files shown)


Spxg (changes requested):
This does improve the user experience! But I prefer to use cfg instead of feature to prevent it from being directly used by other third-party crates, which cannot be overridden to disable.

guybedford:
Thanks for the review, this can work for us - I've updated this to use a configuration instead.

tests/wasm/unsafe_send_sync.rs
```diff
@@ -0,0 +1,99 @@
+//! Tests for the unsafe_single_threaded_traits cfg
+//!
+//! This test file verifies that when the cfg is enabled, JS types
+//! become Send and Sync, allowing futures to be Send.
+
+#![cfg(unsafe_single_threaded_traits)]
```
Spxg, line 6: This shouldn't be necessary, because
```rust
#[cfg(unsafe_single_threaded_traits)]
pub mod unsafe_send_sync;
```

daxpedda (changes requested):
@guybedford would it not be possible to do this hack on your end instead? It seems to me like it would be the right place.

(sorry I didn't think about that during our meeting)

guybedford:
@daxpedda it requires us to write wrappers around all types, and then add send and sync to all those types. See e.g. https://github.com/search?q=repo%3Acloudflare%2Fworkers-rs%20unsafe%20impl%20Sync&type=code.

So this small diff here, avoids every consumer having to deal with this problem with a custom wrapper.

I think if arguing against this quality of life improvement, it would be important to provide a substantial counter argument that outweighs the benefits.

daxpedda:
> @daxpedda it requires us to write wrappers around all types, and then add send and sync to all those types. See e.g. https://github.com/search?q=repo%3Acloudflare%2Fworkers-rs%20unsafe%20impl%20Sync&type=code.

You could instead use `struct SendSyncWrapper<T>(T);`. A single wrapper that would work for all of them.

guybedford:
> You could instead use struct SendSyncWrapper<T>(T);. A single wrapper that would work for all of them.

What I'm hearing here is alternatives, what I don't hear is counter arguments.

daxpedda:
> > You could instead use struct SendSyncWrapper(T);. A single wrapper that would work for all of them.
> 
> What I'm hearing here is alternatives, what I don't hear is counter arguments.

The problem with this feature is that it encourages libraries to rely on it for convenience. Which suddenly makes them incompatible with the atomic target feature even when they might not have been. The fact that `JsValue` isn't `Send + Sync` gives users pause to evaluate what they are doing and what kind of repercussions this has. Presenting an easy way out will discourage that.

Which is why the `cfg` is important, because this makes it much harder to rely on. Though still not ideal.

In general though, I don't know why this specific lever should live in `wasm-bindgen` in the first place. Not implementing `Send + Sync` is the correct thing here after all. It seems to me that the issue is local to `workers-rs`, so it should be addressed there. OP seems to suggest that doing it here is "simpler", but that's where I was suggesting the wrapper, which seemingly addresses the issue. Though I'm not convinced that adding this lever in `wasm-bindgen` is really worth saving 27 lines of code in the first place.

Maybe I'm missing some context why this should be in `wasm-bindgen` instead?

guybedford:
> The problem with this feature is that it encourages libraries to rely on it for convenience. Which suddenly makes them incompatible with the atomic target feature even when they might not have been. The fact that JsValue isn't Send + Sync gives users pause to evaluate what they are doing and what kind of repercussions this has. Presenting an easy way out will discourage that.

> Which is why the cfg is important, because this makes it much harder to rely on. Though still not ideal.

If the feature is not recommended in documentation, clearly marked as unsafe, and requires a compiler flag to activate, then it seems to me that libraries simply won't rely on this feature. Therefore I don't understand how that problem surfaces at all?

It is a feature for platforms like Cloudflare, not libraries. Platforms that know their architecture will support it.

> In general though, I don't know why this specific lever should live in wasm-bindgen in the first place.

Any future that captures a JsValue has to deal with this problem, it appears everywhere and requires all Wasm Bindgen types to be wrapped. This is not a minor detail.

daxpedda:
> If the feature is not recommended in documentation, clearly marked as unsafe, and requires a compiler flag to activate, then it seems to me that libraries simply won't rely on this feature. Therefore I don't understand how that problem surfaces at all?
> 
> It is a feature for platforms like Cloudflare, not libraries. Platforms that know their architecture will support it.

I agree.

My argument was just to highlight that this feature isn't free, it comes with all that you just pointed out. It will only ever be an argument because I don't understand exactly the arguments in favor.

> > In general though, I don't know why this specific lever should live in wasm-bindgen in the first place.
> 
> Any future that captures a JsValue has to deal with this problem, it appears everywhere and requires all Wasm Bindgen types to be wrapped. This is not a minor detail.

I don't really particularly understand the issue, can you provide a more detailed example? It seems to me that a simple `SendSyncWrapper` in the right place should really address the issue exactly where its needed and nowhere else. Which is what it looks like in the `workers-rs` search link you have sent.

I can safely say that the Rust ecosystem is very often dealing with non-`Send/Sync` `Future`s, successfully so. I would like to understand the exact hurdle you are hitting there.

guybedford:
A large portion of Cloudflare Workers users use Axum, which assumes threading support for router futures, with Send and Sync as requirements. JsValue is very common in wasm bindgen apps, providing a strong restriction to usage here. As a result all uses of JsValue require wrapping.

The mismatch of expectations here is that we are bringing a GC type (`JsValue`) into a context in which thread-safe send is assumed. The model mismatch results in the requirement for hacks. Facilitating the hack with the least fuss, by appreciating that when JS GC objects meet threads in a single threading environment, we shouldn't practically need to jump through compiler hoops, helps by choosing usability over dogma for users. Certainly it's not a best practice by any means, but allowing it for platforms is a quality of life improvement.

daxpedda (approved):
> Facilitating the hack with the least fuss, by appreciating that when JS GC objects meet threads in a single threading environment, we shouldn't practically need to jump through compiler hoops, helps by choosing usability over dogma for users. Certainly it's not a best practice by any means, but allowing it for platforms is a quality of life improvement.

This change is very much tailored for `workers-rs`. So I don't think this is a *general* user improvement.
- I believe the use-case is extremely limited. Happy to be proven otherwise.
- If this is supposed to be a UX improvement, we can't use a `cfg`. This only works out because `workers-rs` has its own "compiler". This might be a general theme for platforms, so let me know.
- Using a generic wrapper seems like very little work to me. I would say that adding a `RUSTFLAG` in your compiler is actually more difficult work (passing `RUSTFLAG`s correctly has notoriously been challenging in certain scenarios).

I was assuming that there is a bigger issue that would require `workers-rs` users to use a custom wrapper, which would be a perfect argument for this feature. But otherwise it seems to me like we are just shifting work to the wrong place here.

If this turns out to be a *general* UX improvement, most of my remaining concerns would go away if instead of a `cfg` we go back to a crate feature. However, in that case I'm quite sure it will be abused by libraries and make them `atomics` incompatible, negatively affecting the Rust ecosystem. It can also be argued that if users ignore documentation its out of our hands, but the trade-off seems really small to me.

---

If none of this changed your opinion on this feature: feel free to go ahead and merge it, I've approved it for you.

guybedford:
> This might be a general theme for platforms, so let me know.

Yes, as the platform we control the toolchain for the most part, and use a wasm pack fork to run the build where we can apply these configurations.

> Using a generic wrapper seems like very little work to me.

The problem is JsValue can be used in turn by `n` different Rust types for any `n`, so while individually each wrapper is little work, `n` can be arbitrarily large. That is, there isn't a single wrapper around JsValue we can rely on - we don't try to hide wasm bindgen internals from our users. JsValue is a first-class type they can use and need to use with web-sys etc. So the problem extends to all derivative types and type wrappers here. The argument for generics in core is effectively a similar one.

daxpedda:
> > This might be a general theme for platforms, so let me know.
> 
> Yes, as the platform we control the toolchain for the most part, and use a wasm pack fork to run the build where we can apply these configurations.

Doesn't that make it less general? It seems to me that this whole feature is very tailored to `workers-rs` and has no general alignment with e.g. other platforms trying to use `wasm-bindgen`.

> > Using a generic wrapper seems like very little work to me.
> 
> The problem is JsValue can be used in turn by `n` different Rust types for any `n`, so while individually each wrapper is little work, `n` can be arbitrarily large. That is, there isn't a single wrapper around JsValue we can rely on - we don't try to hide wasm bindgen internals from our users. JsValue is a first-class type they can use and need to use with web-sys etc. So the problem extends to all derivative types and type wrappers here. The argument for generics in core is effectively a similar one.

Okay, so I spent some time looking at `workers-rs` to understand the issue a bit more. This is definitely a user-facing issue in `workers-rs` and puts a type overhead *on users* not just `workers-rs`. So the problem is not actually addressable by `workers-rs` which in my eyes makes this a valid feature to add in `wasm-bindgen`.

However, I should note that while compilation will work because you have your custom toolchain, users will still see errors in `rust-analyzer`. They would need to add some configuration to make that work. Is that an acceptable tradeoff for `workers-rs`? I think it could be solved by a simple `.cargo/config.toml`?

If not, we should discuss alternatives.

guybedford:
> However, I should note that while compilation will work because you have your custom toolchain, users will still see errors in rust-analyzer. They would need to add some configuration to make that work. Is that an acceptable tradeoff for workers-rs? I think it could be solved by a simple .cargo/config.toml?

Great point, yes this likely an issue here for us. Expecting extra config could be onerous, especially on backwards compatible upgrade paths. Would be good to discuss further.

parzivale:
Would it make sense to expose entirely different types ie UnsafeSendJsValue under a feature flag thus it becomes extra hard for library developers to rely on this feature while also preserving the DX of not needing a .cargo/config.toml?

guybedford:
@parzivale the problem with that approach is that web-sys and js-sys use `JsValue` so they'd need to then also have a feature / config that switches over to this new type in turn. Unless we were to fork web-sys and js-sys, which generally isn't advisable either.

HighEntropyUsername:
@guybedford 

[This is an alternative solution](https://docs.rs/js_arc/0.1.0/js_arc/) that solves the Send + Sync problem for JsValue without marking JsValue. 

> It makes it possible to wrap a JsValue in an Arc and use it with other wrapped JsValues. It also supports async closures and covers some trait operations. **It accomplishes this by keeping JsValues on one thread and sending closures from other threads to be executed on it**.

**I don't know if it is relevant to your issue**. I like it because if proper multi-threading existed for WASM, it would allow a worker thread and the main thread to share access safely to a JsValue while still using normal rust code in closures. As an example, a lib like [webtorrent](https://webtorrent.io/) could be loaded, then another thread could safely interact with the active torrents managed by the lib through closure sending.

PR #4771, opened 2025-10-30
CI update for latest Rust stable

guybedford:
Seems many of our outputs were bumped by the Rust stable release today.

This should address these CI failures.


hoodmane (approved):

guybedford:
There seems to be some non-determinism here unfortunately.

PR #4781, opened 2025-11-03
add new community projects page to guide

guybedford:
### Description

Adds a new community projects page to the guide for both libraries and tools for Wasm Bindgen as well as platforms built on top of wasm bindgen.

Initial list is kept relatively small for now to only include major items.

### Checklist


guybedford:
Good points, added.

PR #4782, opened 2025-11-03
Interpreter: Support try blocks

hoodmane:
Part of unwinding support.


guybedford:
Bypassing the changelog requirement here as these are low-level fundamentals for an _eventual_ exception handling changelog, but don't represent user-facing progress yet.

PR #4786, opened 2025-11-03
Fix bug where TryFromJsValue failure for structs would invalidate JsValue

guybedford:
### Description

Resolves https://github.com/wasm-bindgen/wasm-bindgen/issues/4289.

The problem was that the `__unwrap` function takes a `JsValue` by value and therefore invalidates the passed JsValue, even in the failure case when we want to be able to get the ownership back. In particular the externref stack was actually being corrupted in these cases.

The fix was to make `__unwrap` take a reference to an externref instead. The drop is then left to normal implicit semantics without requiring any further tricks.

### Checklist

### DIFF
--- a/tests/wasm/try_from_js_value.rs
+++ b/tests/wasm/try_from_js_value.rs
@@ -385,7 +385,7 @@ fn try_from_js_value_invalidates_cloned_value() {
     let value = JsValue::from(circle);
 
     // First check fails (value is CircleShape, not SquareShape)
-    let _ = SquareShape::try_from_js_value(value.clone());
+    assert!(SquareShape::try_from_js_value(value.clone()).is_err());
 
     // Second check should succeed, but the first check invalidated the cloned value
     let result = CircleShape::try_from_js_value(value.clone());


tests/wasm/try_from_js_value.rs
```diff
@@ -364,3 +364,33 @@ fn numeric_types_reject_invalid_js_values() {
...
+    side_length: f64,
+}
+
+#[wasm_bindgen]
+pub struct CircleShape {
+    #[allow(dead_code)]
+    radius: f64,
+}
+
+#[wasm_bindgen_test]
```
Spxg, line 382: This test also runs success in the release version.

Spxg (changes requested):
To use `try_from_js_value` multiple times, must clone the value, so the value will not become invalid?

tests/wasm/try_from_js_value.rs
guybedford: The test works, but it causes invalid externref heap that will cause panic on unrelated tests...!

guybedford:
> To use try_from_js_value multiple times, must clone the value, so the value will not become invalid?

The value no longer becomes invalid because we are passing `&JsValue` to unwrap now, not `JsValue`.

tests/wasm/try_from_js_value.rs
guybedford: To replicate - run the full test suite - `just test-wasm-bindgen`.

Spxg: > The test works, but it causes invalid externref heap that will cause panic on unrelated tests...!

indeed

tests/wasm/try_from_js_value.rs
```diff
@@ -364,3 +364,33 @@ fn numeric_types_reject_invalid_js_values() {
...
+fn try_from_js_value_invalidates_cloned_value() {
+    let circle = CircleShape { radius: 5.0 };
+    let value = JsValue::from(circle);
+
+    // First check fails (value is CircleShape, not SquareShape)
+    let _ = SquareShape::try_from_js_value(value.clone());
```
Spxg: I think we can make a assertion instead of ignoring the result.

```suggestion
    assert!(SquareShape::try_from_js_value(value.clone()).is_err());
```

Spxg (approved):

PR #4788, opened 2025-11-04
support "private" attribute on exports

guybedford:
### Description

Adds support for a `private` attribute on exported structs and enums.

Currently today, all Rust-defined enums and classes must be explicitly exported in the public interface of the module - `import * as ns from './foo'` must have a `ns.ClassName` and `ns.EnumName` for all enums and classes.

When there are a lot of these types, this ends up bloating the interface of the module, even when the intention is only to have a few explicitly exported classes and enums for JS consumers, and the expectation is these are rather the Rust types of arguments on exported functions.

The hide attribute can be used on exports to effectively make these exported Rust types semi-private, cleaning up and reducing the surface area of the exported API, while they can still be returned from exported functions and used in exported function arguments, and are still reflected as a type export from TypeScript.

### Checklist


guybedford:
We do align the typescript with this so the distinction is one of public versus private on the module interface.

Perhaps `skip_export` would most closely align with existing terminology in using skip? Clarifying correctly then that what is being skipped is not JS or TS, but the actual public export of the thing as separate to its binding generation.

guybedford:
I've gone ahead and renamed to `skip_export`. Happy to discuss further.

crates/shared/src/lib.rs
daxpedda: We should bump the schema version in this case.

guybedford:
Yeah that makes sense - okay `private` per the branch name is now implemented.

guybedford:
This is now rebased to tbe exports refactoring and ready for final review.

daxpedda (approved):

PR #4790, opened 2025-11-07
If built w/ unwind and std, wrap JS->Rust calls in catch_unwind

hoodmane:
We introduce a new catch-unwind feature, and a new `PanicError` class. When built with
unwinding and std support, we catch panics and raise into JS as `PanicError`s. If the
entrypoint is async, we reject the promise with the `PanicError`.

This can only work if built with `panic=unwind`, which also requires `-Zbuild-std`, so
it only works on a nightly compiler.
    
The main limitation of this is that it implies that all arguments to functions must be
unwind safe.

### Checklist
- [x] Add changelog entry
- [x] Add tests


guide/src/reference/catch-unwind.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+## Building
+
+Build your project with the required flags:
+
+```bash
+cargo +nightly build --target wasm32-unknown-unknown \
+    -Zbuild-std=std,panic_unwind \
+    -Zbuild-std-features=panic_immediate_abort
```
hoodmane:
```suggestion
cargo +nightly build --target wasm32-unknown-unknown -Zbuild-std
```

guide/src/reference/catch-unwind.md
```diff
@@ -0,0 +1,174 @@
+# Catching Panics
+
+By default, when a Rust function exported to JavaScript panics, the WebAssembly
+module will abort and any allocated resources will be leaked. The `catch-unwind`
```
hoodmane: I think "abort" is not a well defined thing for a WebAssembly module to do. Maybe we should say that "Rust execution will abort" or something like that? And of course on the `wasm32-unknown-unknown` platform "abort" just means "execute the unreachable instruction".

guide/src/reference/catch-unwind.md
```diff
@@ -0,0 +1,174 @@
...
+
+## The PanicError Class
+
+When a panic occurs, a `PanicError` JavaScript exception is created with:
+
+- `name` property set to `"PanicError"`
+- `message` property containing the panic message (if the panic was created with
+  a string message)
+
+If the panic payload is not a string (e.g., `panic_any(42)`), the message will
```
hoodmane:
```suggestion
If the panic payload is not a `String` or `&str` (e.g., `panic_any(42)`), the message will
```

guide/src/reference/catch-unwind.md
```diff
@@ -0,0 +1,174 @@
...
+
+All function arguments must satisfy Rust's `UnwindSafe` trait. This is
+automatically handled by wrapping arguments in `AssertUnwindSafe`, but be aware
+that this assumes your code handles potential inconsistent state after a panic.
+
+### Mutable Slice Arguments
+
+Functions with `&mut [T]` slice arguments cannot be used with `catch-unwind`
+because mutable slices are not `UnwindSafe`. Consider using owned types like
+`Vec<T>` instead.
```
hoodmane: Would be nice if we could do `AssertUnwindSafe<&mut [T]>` or some variation of that but it really does not work.

guide/src/reference/catch-unwind.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Nightly Only
+
+This feature requires a nightly Rust compiler and will not work on stable Rust.
+
+### Panic Message Extraction
+
+Only panics created with string messages (via `panic!("message")` or
+`panic!("{}", formatted)`) will have their message preserved. Panics using
+`std::panic::panic_any` with non-string types will show a generic message.
```
hoodmane: These are super rare anyways.

guide/src/reference/catch-unwind.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Comparison with Default Behavior
+
+| Aspect | Default (panic=abort) | With catch-unwind |
+|--------|----------------------|-------------------|
+| Panic behavior | Module aborts | Exception thrown to JS |
+| Destructors | Do not run | Run properly |
+| Memory | May leak | Properly cleaned up |
+| Error handling | Cannot catch | Can catch in JS |
+| Compiler | Stable | Nightly only |
```
hoodmane: Delete this.
```suggestion
```

src/lib.rs
```diff
@@ -1399,6 +1399,10 @@ pub fn throw(s: &str) -> ! {
 /// function, unlike `panic!` on other platforms, **will not run destructors**.
 /// It's recommended to return a `Result` where possible to avoid the worry of
 /// leaks.
+///
+/// If you need destructors to run, consider using `panic!` with the
+/// `catch-unwind` feature enabled, which catches panics and converts them to
+/// JavaScript exceptions while properly running destructors.
```
hoodmane: Actually this will run destructors when built with `-Cpanic=unwind`. Though it's unclear if this is the intended behavior by the Rust compiler. I don't imagine it will change.

src/lib.rs
```diff
@@ -1416,6 +1420,10 @@ pub fn throw_str(s: &str) -> ! {
...
```
hoodmane: Same thing, if you build with `-Cpanic=unwind` then this will run destructors.

tests/wasm/unwind.rs
```diff
@@ -0,0 +1,126 @@
...
+        predicate: &mut dyn FnMut(JsValue, u32, Array) -> Result<bool, JsValue>,
+    ) -> bool;
+
+    // Predicate panic will panic, but be caught as Result::Err
+    // Unless a WebAssembly.RuntimeError, in which case, abort propagates
+    #[wasm_bindgen(method, catch, js_class = Array, js_name = every)]
+    pub fn try_every(
+        this: &ArrayUnwind,
+        predicate: &mut dyn FnMut(JsValue, u32, Array) -> bool,
+    ) -> Result<bool, JsValue>;
```
hoodmane: Should we remove these other variants? They aren't currently used so we should either add more tests that use them or delete them. I'm not clear that tests using these variants would be all that valuable though.

src/lib.rs
guybedford: Good point, corrected.

tests/wasm/unwind.rs
guybedford: I still feel like it's quite nice to have the coverage, for checking these cases or expanding the test with follow-ons.

tests/wasm/unwind.rs
```diff
@@ -0,0 +1,102 @@
...
+        resolve_
+            .call1(&JsValue::undefined(), &JsValue::undefined())
+            .unwrap();
+    });
+    set_timeout(&closure2);
+    let result = future.await;
+    assert!(result.is_err());
+    // let err = result.err().unwrap();
+    // let msg = Reflect::get(&err, &"message".into()).unwrap();
+    // console_log!("Panic error: {}", &msg.as_string().unwrap());
```
guybedford: @hoodmane I just had a look at this case again, and the error here is `JsValue(undefined)`.

Trying to think this one through:

* The future for promise2 is rejecting undefined
* promise2 is the promise.then(() => panic!)
* when JS calls into this closure back into Wasm and gets the panic, it sounds like it is therefore detecting an unwound error, but not treating that PanicError as the actual error.

I think we do need to try and ensure continuity of the PanicError object here.

PR #4792, opened 2025-11-09
Added CSS Custom Highlight API

feathecutie:
EDIT: Currently blocked by https://github.com/wasm-bindgen/wasm-bindgen/issues/4865

### Description
Adds the IDL of the CSS Custom Highlight API.
- MDN: https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API
- Spec: https://drafts.csswg.org/css-highlight-api

### Checklist


feathecutie:
Nevermind, it seems like this didn't add the `highlights` field to the CSS namespace? Not sure why that is, but this is unusable in its current form

guybedford:
Please just update here when ready for review.

guybedford:
I've added support for namespace attributes as you mentioned in https://github.com/wasm-bindgen/wasm-bindgen/pull/4930. PTAL and let me know if that resolves the blocker here.

PR #4796, opened 2025-11-11
Support building with -Cpanic=unwind

hoodmane:
I'm trying to add a mode where we enable panic unwinding and raise panics as JavaScript exceptions. See #4790. This makes the test suite pass when we compile with `-Cpanic=unwind`.

The first thing we need is to switch all exported JS functions to `extern "C"` instead of `extern "C-unwind"` because `extern "C"` functions when compiled with `-Cpanic=unwind` catch all exceptions (Rust panics and foreign exceptions) and abort. This is only necessary on Rust version >= 1.71, in older Rust versions `extern "C-unwind"` does not exist and `extern "C"` does what we want. We should ask the Rust maintainers whether this behavior is intended -- I understand the reason to catch and abort when a Rust panic unwinds across `extern "C"`, but why JavaScript exceptions are treated in the same way is unclear to me.

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -148,6 +148,23 @@ jobs:
         node-version: '20'
     - run: cargo test --target wasm32-unknown-unknown -Zbuild-std=std,panic_abort --config="target.wasm32-unknown-unknown.rustflags='${{ matrix.envs.RUSTFLAGS }}'"
 
+  test_wasm_bindgen_unwind:
+    name: "Run wasm-bindgen crate tests with panic=unwind"
+    runs-on: ubuntu-latest
+    env:
+      WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
+      RUSTFLAGS: -Cpanic=unwind
+    steps:
+    - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -22,6 +22,7 @@ features = ["serde-serialize"]
 test = false
 
 [features]
+_panic-unwind = ["std", "wasm-bindgen-macro/panic-unwind"]
```
Spxg: Why was the `_` prefix added?

hoodmane: The idea is that it means it's unstable. If we bump the msrv to 1.60 we could remove it entirely.

Spxg: Once a feature is added and released, it's very difficult to remove because it's considered a breaking change, so I suggest giving it a more meaningful name. Maybe `unstable_panic_unwind`?

Spxg: or just `panic_unwind`

hoodmane: If we merge #4801 first we can get rid of this.

hoodmane: Removed the feature.

guybedford (approved):
Thanks for using the exception handling feature gate here. Will land this now.

PR #4799, opened 2025-11-12
Fix `--nocapture` not working when `run_in_browser` was set.

Spxg:
### Description

`--nocapture` not working as expected when `wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);` and `NO_HEADLESS` were set

### Checklist


crates/cli/src/wasm_bindgen_test_runner/server.rs
```diff
@@ -260,6 +260,29 @@ pub(crate) fn spawn(
...
+                    }}
+                    og(args[0]);
+                }};
+            }};
+
+            wrap("debug");
+            wrap("log");
+            wrap("info");
+            wrap("warn");
+            wrap("error");
```
guybedford, line 344: I really don't like mutating globals, even in a test environment. We control bindings, so we should ideally give Wasm a wrapped console, not alter the global console for everyone.

Separately while on the topic of console, it is really painful in wasm bindgen not having proper stdio and logging integrations that could in turn support nocapture. Perhaps time to reopen that discussion to some extent?

crates/cli/src/wasm_bindgen_test_runner/server.rs
```diff
@@ -260,6 +260,29 @@ pub(crate) fn spawn(
...
+
+            window.__wbg_test_output = "";
+            window.__wbg_test_output_writeln = function (line) {{
+                window.__wbg_test_output += line + "\n";
+                document.getElementById("output").textContent = window.__wbg_test_output;
+            }}
```
guybedford, line 350: Again, adding new globals, even in a test environment should be avoided if we can, and I think we can avoid it?

That is, we could check the value of `nocapture` just as well surely? Can't we pass it into initialization somehow?

crates/cli/src/wasm_bindgen_test_runner/server.rs
Spxg: > We control bindings, so we should ideally give Wasm a wrapped console, not alter the global console for everyone.

The intention here is to modify everyone's console, just like native test controls stdio. It seems that a wrapped console cannot do this and requires everyone to call it manually?

crates/cli/src/wasm_bindgen_test_runner/server.rs
Spxg: > Can't we pass it into initialization somehow?

It can, however, passing it in is unnecessary, the `nocapture` option applies to console hook. Do you mean creating a console hook in Rust?

crates/cli/src/wasm_bindgen_test_runner/server.rs
guybedford: Right but native streams are defined natively - stdio as the abstraction. The global console being mutated can lead to subtle bugs. Mutating globals is almost always a bad idea. This is why I say perhaps we should talk about supporting stdio or better logging primitives instead?

crates/cli/src/wasm_bindgen_test_runner/server.rs
guybedford: Happy to start with the console discussion and come back to this one.

crates/cli/src/wasm_bindgen_test_runner/server.rs
guybedford: I realise when I originally wrote this I didn't fully follow the code in our runner to know we are already wrapping console APIs.

Given this my points above are moot, happy to re-review with a rebase if you like.

crates/cli/src/wasm_bindgen_test_runner/server.rs
```diff
@@ -326,6 +326,29 @@ pub(crate) fn spawn(
...
+            }};
+
+            wrap("debug");
+            wrap("log");
+            wrap("info");
+            wrap("warn");
+            wrap("error");
+
+            window.__wbg_test_output = "";
+            window.__wbg_test_output_writeln = function (line) {{
```
guybedford, line 347: I believe this has now changed with some of the more recent PRs - so we should probably update this pattern to match.

guybedford:
@Spxg @drewcrawford is this PR still required?

drewcrawford:
IMO the problem does still exist, and the solution is directionally right and has a good shape. I think it's more a practical problem of whether we take the bird in hand or I get around to the rest of the bush.

ex, here we have:

```
 const nocapture = {nocapture};
  const wrap = method => {
    const og = console[method];
    console[method] = function (...args) {
      if (nocapture) {
        window.__wbg_test_output_writeln(args);
      }
      og(args[0]);
    };
  };
```

In my tree I have for both modes

```
  const nocapture = /* stay tuned! */
  const wrap = method => {
      const og = console[method];
      const on_method = `on_console_${method}`;
      console[method] = function (...args) {
          if (nocapture) {
              appendTo("output").apply(this, args);
          } else {
              appendTo("console_output").apply(this, args);
          }
          /* other stuff here... stay tuned! */
      };
  };
```

PR #4803, opened 2025-11-13
Fix `wasm-bindgen-test-runner` outputting empty line

Spxg:
### Description

Fix `wasm-bindgen-test-runner` outputting empty line when using the `--list` option.

In particular, `cargo-nextest` now works correctly.

resolve https://github.com/wasm-bindgen/wasm-bindgen/issues/4802

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,20 +44,6 @@ jobs:
           tool: taplo-cli
       - run: taplo fmt --check
 
-  # nextest check
-  nextest:
-    name: Nextest
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v5
-      - uses: dtolnay/rust-toolchain@1.90
-        with:
-          targets: wasm32-unknown-unknown
-      - uses: taiki-e/install-action@v2
-        with:
-          tool: nextest
-      - run: cargo nextest run --target wasm32-unknown-unknown --no-run -p wasm-bindgen
-
   # Run `cargo clippy` over everything
   clippy:
     name: Clippy (native)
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -44,6 +44,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dtolnay/rust-toolchain@1.90
+        with:
+          targets: wasm32-unknown-unknown
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+      - run: cargo nextest run --target wasm32-unknown-unknown --no-run -p wasm-bindgen
```
guybedford: IT seems like a little bit of overkill to be explicitly running this in CI. A test fixture for the empty line on its own to avoid a regression would be sufficient and avoid adding extra CI work.

Spxg: `--no-run`, no test will run.

guybedford: Right I just mean from a maintenance point of view that every extra dependency we add is maintenance, and I'm not clear on why we specifically need this very specific coverage. If I fix an issue for a library I don't add the library to my CI I add a unit test for the functionality, unless it's a really important library.

Spxg: Similar issues have actually been fixed before: https://github.com/wasm-bindgen/wasm-bindgen/issues/4517

`cargo-nextest` can make running tests faster (perhaps we should try it).

guybedford: Right, but this is what tests are for. I'd rather have something running `wasm-bindgen-test` with the same args and verifying properties of stdout output, than hard-coding external runners is what I'm saying here though. Is coverage very tricky to do stdio here? Surely it's as simple as `wasm-bindgen-test` with a validator process?

RReverser: I agree with @guybedford, we should strive to add a unit test unless the integration is very complex - which I believe it isn't for this bug.

PR #4804, opened 2025-11-13
Add bench support

Spxg:
### Description

There is still a lot to do.
### Checklist


GuillaumeLagrange:
Hey @Spxg,

I'm currently working on CodSpeed, and I'm both happy and impressed by the work here!
I see you've already managed to run one benchmark successfully [here](https://github.com/wasm-bindgen/wasm-bindgen/blob/w/bench/crates/js-sys/benches/array_interaction.rs)

I notice that you rely a lot on duplicated code or internal undocumented APIs, some of them may be subject to breaking changes in the future. Is there anything we can do to improve the experience? Or anything lacking in our implementations or documentation that you would need in order reduce the reliance on duplicated/internal stuff on our end?

It would both help us understand the downstream needs for custom integrations, as well as provide a more stable experience for you in the long run.

Spxg:
> It would both help us understand the downstream needs for custom integrations, as well as provide a more stable experience for you in the long run.

Thank you for your attention. I'm happy to share my process. 

First, `wasm32-unknown-unknown` is a target with limited capabilities, which means it cannot use file system functions directly. I forked criterion and modified it to store test baseline data in memory, which is eventually written to disk by node.js.

For codspeed, I faced the same issue. Therefore, I wrapped codspeed's collect function using N-API to handle data writing and complete walltime collection. For codspeed's profiling functionality, I created additional wrapper functions to facilitate this process. 

I believe using N-API is unavoidable in this case, especially for profiling scenarios that require certain event notifications to operate properly. Actually, using codspeed with this target is relatively uncommon, but it would be great if there were more convenient ways to use codspeed with custom test runners!

Additionally, I've encountered crashes when using the with `codspeed-macro` runner, it runs without issues on both `ubuntu-latest` and `ubuntu-arm` github runners.
https://github.com/wasm-bindgen/wasm-bindgen/actions/runs/19405856771/job/55520365895

GuillaumeLagrange:
Thank you very much for the answer, these are indeed quite custom needs.

Long term, we really want to provide guides on how to do your own integration, building on top of https://github.com/CodSpeedHQ/instrument-hooks.

In the meantime, I think the custom integration you have spinned up here will hold for a while. Do not hesitate to reach out, through pings on PRs and issues if you have any difficulties encountered with your custom integrations.

Spxg:
https://github.com/wasm-bindgen/wasm-bindgen/pull/4812

PR #4807, opened 2025-11-18
Check output

Spxg:
### Description

test only
### Checklist


Spxg:
Found the regressed commit: https://github.com/rust-lang/rust/commit/350d0ef0ec0493e6d21cfb265cb8211a0e74d766

PR #4808, opened 2025-11-18
Switch to core::error::Error to support no_std

BroderickCarlin:
### Description
This PR changes the usage of `std::error::Error` to use `core::error::Error`, further supporting `no_std` environments 

### Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1796,7 +1796,7 @@ where
     E: core::error::Error,
 {
     fn from(error: E) -> Self {
-        use std::string::ToString;
+        use alloc::string::ToString;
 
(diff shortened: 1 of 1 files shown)


BroderickCarlin:
Given the needs to support version 1.57, and `core::error` having been added only in `1.81.0`, and options such as `cfg(version())` still not stabilized ([#64796](https://github.com/rust-lang/rust/issues/64796)), I'm unsure there is actually a good path forward for this PR at this time. I will leave it open for the time being to see if there are any other's thoughts

hoodmane:
Apparently we can use this to gate on rust version:
https://docs.rs/rustversion/latest/rustversion/

PR #4813, opened 2025-11-19
Refactor export generation with sorted ordering

guybedford:
### Description

This refactors export generation, as a follow-on to the reexport and namespace work that was done and hinted at in that work. By now collecting all exports into a BTree export table, we then obtain deterministic export output ordering.

This refactoring was done carefully by gradually migrating the current structures into the desired single map approach using a new `ExportDefinition` struct. It is effectively "just" a refactoring from immediately outputting exports to instead collecting exports and then outputting them during a dedicated `write_exports()` phase at finalization.

In the process it also unifies renaming paths and enables support for non-identifier export names, as well as reserved word export names like `default` now being supported for classes and functions fully as well. This was done by updating the valid identifier conversion to check against JS reserved words, so that names like `true` or `await` are now turned into `_true` and `_await` while still being exported as the expected name. Some bugs in reexport handling were also fixed in the process, including a new reference test for reexports.

Another benefit of this refactoring is that typescript and JS now get generated together in the same locations as they are both stored on the new `ExportDefinition` that is collected and then output together once ordering has been achieved.

I've also included the lock removal from https://github.com/wasm-bindgen/wasm-bindgen/pull/4791 in this PR, this now resolves the deterministic ordering issue.

### Checklist

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1473,12 +1473,8 @@ wasm = wasmInstance.exports;
             match entry {
                 ExportEntry::Namespace(ns) => {
                     output.push_str(&format!(
-                        "{full_name} = {}{{}};\n",
-                        if existing {
-                            format!("{full_name} || ")
-                        } else {
-                            "".to_string()
-                        }
+                        "{full_name} {}= {{}};\n",
+                        if existing { "||" } else { "" }
                     ));
                     output.push_str(&self.write_namespace(&full_name, &ns.ns, existing)?);
                 }
--- a/crates/cli/tests/reference/js-namespace-export.bg.js
+++ b/crates/cli/tests/reference/js-namespace-export.bg.js
@@ -212,7 +212,7 @@ function concat(a, b) {
  * @param {string} s
  * @returns {string}
  */
-function uppercase(s) {
+function uppercase2(s) {
     let deferred2_0;
     let deferred2_1;
     try {
@@ -231,7 +231,7 @@ const _default = {};
 _default.Counter = Counter;
 _default.concat = concat;
 _default.uppercase = {};
-_default.uppercase.uppercase = uppercase;
+_default.uppercase.uppercase = uppercase2;
 export default _default;
 
 /**
@@ -402,7 +402,7 @@ types.http.HttpStatus = HttpStatus;
  * @param {string} s
  * @returns {string}
  */
-function uppercase2(s) {
+function uppercase(s) {
     let deferred2_0;
     let deferred2_1;
     try {
@@ -419,7 +419,7 @@ function uppercase2(s) {
 
 export const utils = {};
 utils.string = {};
-utils.string.uppercase = uppercase2;
+utils.string.uppercase = uppercase;
 
 export function __wbg___wbindgen_throw_b855445ff6a94295(arg0, arg1) {
     throw new Error(getStringFromWasm0(arg0, arg1));
--- a/crates/cli/tests/reference/js-namespace-export.d.ts
+++ b/crates/cli/tests/reference/js-namespace-export.d.ts
@@ -25,13 +25,13 @@ declare class Counter {
 
 declare function concat(a: string, b: string): string;
 
-declare function uppercase(s: string): string;
+declare function uppercase2(s: string): string;
 
 declare let _default: {
   Counter: typeof Counter,
   concat: typeof concat,
   uppercase: {
-    uppercase: typeof uppercase,
+    uppercase: typeof uppercase2,
   },
 };
 export default _default;
@@ -93,10 +93,10 @@ export let types: {
   },
 };
 
-declare function uppercase2(s: string): string;
+declare function uppercase(s: string): string;
 
 export let utils: {
   string: {
-    uppercase: typeof uppercase2,
+    uppercase: typeof uppercase,
   },
 };


crates/cli-support/src/js/mod.rs
```diff
@@ -481,6 +468,16 @@ impl<'a> Context<'a> {
         let mut js = String::new();
         let mut start = None;
 
+        // take globals
+        let globals = std::mem::take(&mut self.globals);
+
+        // take and emit intrinsics - ensures any new intrinsics added after this will now throw.
+        let intrinsics = self.intrinsics.take().unwrap();
```
Spxg:
```rust
intrinsics: BTreeMap<Cow<'static, str>, Cow<'static, str>>,
let intrinsics = std::mem::take(&mut self.intrinsics);
if intrinsics.is_empty() {
    bail!("xxxx");
}
```
Intrinsics are only `None` in this case; I think it's sufficient to simply check if it's empty and `unwrap()` is not needed elsewhere.

crates/cli-support/src/js/mod.rs
```diff
@@ -4750,6 +4750,19 @@ impl ExportedClass {
...
 }
 
+type Intrinsics = BTreeMap<Cow<'static, str>, Cow<'static, str>>;
+
+fn intrinsic(
+    intrinsics: &mut Option<Intrinsics>,
+    name: Cow<'static, str>,
+    f: impl FnOnce() -> Cow<'static, str>,
+) {
+    if intrinsics.as_ref().unwrap().contains_key(&name) {
```
Spxg, line 4772: why noop? problems will not be easily detected

crates/cli-support/src/js/mod.rs
```diff
@@ -687,30 +684,26 @@ wasm = wasmInstance.exports;
              generated some JS before the imports: {js}"
         );
 
-        let mut push_with_newline = |s| {
-            js.push_str(s);
+        let mut needs_newline = false;
+        let mut push_with_newline = |s: &str| {
```
Spxg: It looks like `join` can be used:

```rust
let mut js = [
    js.as_str(),
    imports.as_str(),
    self.imports_post.as_str(),
    intrinsics.as_str(),
    globals.as_str(),
    init_js.as_str(),
    footer.as_str(),
]
.join("\n");
```

crates/shared/src/identifier.rs
```diff
@@ -55,7 +110,16 @@ pub fn is_valid_ident(name: &str) -> bool {
 /// Converts a string to a valid JavaScript identifier by replacing invalid
 /// characters with underscores.
 pub fn to_valid_ident(name: &str) -> String {
-    maybe_valid_chars(name)
+    let result: String = maybe_valid_chars(name)
         .map(|opt| opt.unwrap_or('_'))
-        .collect()
+        .collect();
+
+    if RESERVED_WORDS.contains(&result.as_str()) {
```
Spxg:
```suggestion
    if is_valid_ident(&result) {
```

crates/shared/src/identifier.rs
```diff
@@ -55,7 +110,16 @@ pub fn is_valid_ident(name: &str) -> bool {
...
+        let mut prefixed = String::with_capacity(result.len() + 1);
+        prefixed.push('_');
+        prefixed.push_str(&result);
+        prefixed
```
Spxg:
```suggestion
        format!("_{result}")
```

crates/cli-support/src/wit/nonstandard.rs
```diff
@@ -34,9 +34,9 @@ pub struct WasmBindgenAux {
     /// A map from imported function id to what it's expected to import.
     pub import_map: HashMap<AdapterId, AuxImport>,
 
-    /// A map from AdapterId to the reexport name (if the import should be re-exported).
-    /// A None value means reexport with the same name as the import.
-    pub reexports: HashMap<AdapterId, Option<String>>,
+    /// List of imports that should be re-exported.
+    /// The Option<String> is the optional renamed export name.
```
Spxg: delete?

crates/cli-support/src/wit/mod.rs
```diff
@@ -620,26 +620,44 @@ impl<'a> Context<'a> {
...
+                    .reexports
+                    .insert(reexport_name.clone(), import.clone());
+            }
+        }
+
+        // Handle reexports without adapters (e.g., types from snippets)
+        if let Some(source) = reexport_source {
             if let Some(reexport_name) = import.reexport {
-                self.aux.reexports.insert(id, reexport_name);
+                self.aux.reexports.insert(reexport_name, source);
```
Spxg: There's also an insert above, will it be overwritten here?

crates/cli-support/src/js/mod.rs
guybedford: This is the design of the function that replaces the former `should_write_global` check. This function is a conditional set into the map if not already set.

crates/cli-support/src/js/mod.rs
guybedford: The newline is only added conditionally between two sections that are both not empty. This avoids double newlines.

crates/shared/src/identifier.rs
guybedford: This is in the shared crate, which is built without `format`.

crates/cli-support/src/wit/nonstandard.rs
guybedford: Thanks.

crates/cli-support/src/wit/mod.rs
guybedford: Thanks, good catch I've refactored this code a bunch more.

crates/shared/src/identifier.rs
guybedford: Added.

crates/shared/src/identifier.rs
Spxg: I understand you're concerned about the no_std environment. But `alloc::format` is available.

guybedford: Added, thanks.

Spxg: get, but seem forgot to push?

guybedford:
To be clear - there is still some non-determinism in various places, this PR doesn't solve all of it, but it does implement ordering which solves our CI problem for now. The primary feature here though is a unified exports collection system that works with namespaces and reexports. This should also be a good base for further work as in https://github.com/wasm-bindgen/wasm-bindgen/pull/4800.

PR #4817, opened 2025-11-21
ci: lock nightly pending closure fix

guybedford:
### Description
Temporarily locks nightly to 11-19, due to the current CI failure on today's nightly.

### Checklist


guybedford:
The nightly lock required both changing toolchain strings internally and also using a custom build since for some reason the dtolnay version with a nightly lock wasn't built with atomics support.

Should be working now, to land and revert!

PR #4825, opened 2025-11-23
Remove unused ABI cfg code

Spxg:
### Description

The Minimum Supported Rust Version (MSRV) has been bumped to 1.71

### Checklist

### DIFF
--- a/crates/cli/tests/reference/js-namespace-export.bg.js
+++ b/crates/cli/tests/reference/js-namespace-export.bg.js
@@ -212,7 +212,7 @@ function concat(a, b) {
  * @param {string} s
  * @returns {string}
  */
-function uppercase(s) {
+function uppercase2(s) {
     let deferred2_0;
     let deferred2_1;
     try {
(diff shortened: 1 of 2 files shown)


Spxg:
It looks like this output is flip-flopping.

guybedford:
Not sure how I feel about this. Would appreciate a review from @hoodmane here. Even though our formal MSRV is 1.71, I do also see no problem with us informally supporting older versions without guarantees?

hoodmane:
I am happy to remove this but we should clean up a bit further I think.

Spxg:
> supporting older versions without guarantees?

But we have specified the MSRV in Cargo.toml, cargo will detect it. Or you mean the forked and modified version?

hoodmane:
Okay I pushed a more thorough cleanup of the changes I made in #4796. Now I think this can merge (assuming I didn't somehow break things...)

Spxg:
> Unknown Walltime execution environment detected

I'm having some trouble with the Codspeed runner, see https://github.com/CodSpeedHQ/action/issues/160

PR #4846, opened 2025-12-08
Call original console methods in worker scripts

drewcrawford:
### Description

Worker console methods (log/debug/info/warn/error) are wrapped to relay output to the main thread, but the original console methods are never called. This means worker logs don't appear in browser DevTools when inspecting the worker context.

This aligns worker behavior with `index-headless.html` which correctly saves and calls the original method via `og.apply(this, args)`.

#### Motivation

When debugging worker-specific issues with `NO_HEADLESS=1`, it's useful to see logs directly in the worker's DevTools console rather than only through the postMessage relay.

### Checklist


guybedford (approved):
Yeah this is always a pain - spotted this with @hoodmane recently while debugging, thanks for the fix.

PR #4848, opened 2025-12-10
When KEEP_TEST_BUILD=1 set, save test to consistent path

hoodmane:
For now, I called the folder `saved_test`.

### Checklist

### DIFF
--- a/crates/cli/src/wasm_bindgen_test_runner.rs
+++ b/crates/cli/src/wasm_bindgen_test_runner.rs
@@ -213,17 +213,18 @@ fn rmain(cli: Cli) -> anyhow::Result<()> {
         return Ok(());
     }
 
-    let tmpdir = tempfile::tempdir()?;
 
     // Support a WASM_BINDGEN_KEEP_TEST_BUILD=1 env var for debugging test files
     let tmpdir_path = if env::var("WASM_BINDGEN_KEEP_TEST_BUILD").is_ok() {
-        let path = tmpdir.keep();
+        let path = Path::new("saved_test").to_path_buf();
         println!(
             "Retaining temporary build output folder: {}",
             path.to_string_lossy()
         );
+        fs::create_dir_all(path.clone())?;
         path
     } else {
+        let tmpdir = tempfile::tempdir()?;
         tmpdir.path().to_path_buf()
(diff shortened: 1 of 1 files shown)


crates/cli/src/wasm_bindgen_test_runner.rs
```diff
@@ -213,17 +213,18 @@ fn rmain(cli: Cli) -> anyhow::Result<()> {
         return Ok(());
     }
 
-    let tmpdir = tempfile::tempdir()?;
 
     // Support a WASM_BINDGEN_KEEP_TEST_BUILD=1 env var for debugging test files
     let tmpdir_path = if env::var("WASM_BINDGEN_KEEP_TEST_BUILD").is_ok() {
-        let path = tmpdir.keep();
+        let path = Path::new("saved_test").to_path_buf();
```
Spxg, line 219: I think this should be configurable, do you think?
```
let path = `WASM_BINDGEN_KEEP_TEST_BUILD_PATH` or_else tmpdir.
```

guybedford: Yes, maybe we remove WASM_BINDGEN_KEEP_TEST_BUILD and replace it with WASM_BINDGEN_TEST_BUILD_PATH say.

hoodmane: Is there any reason it needs to be configurable? For reference, Emscripten's test suite has an unconfigurable output directory and I've never wanted to change it.

Spxg: Using the `saved_test`  will result in git diff, so I prefer `target/saved_test`. 
However, in workspace projects, typically only the root directory's target is gitignored. Therefore, if testing other crates, there will still be diffs. So here, I suggest using a custom PATH env determined by the user, If not provided, just `tmpdir.keep()`.

guybedford: @hoodmane note here that our users also use this test runner for their tests, this isn't just an internal runner and changes have to align with project compatibility. Perhaps that adds context to why we're having such an involved discussion on this.

Will await your update further.

PR #4853, opened 2025-12-10
Fixes incorrect stringification of multi-argument console calls

drewcrawford:
# Fix array-to-string coercion in worker console output

## Summary

Fixes incorrect stringification of multi-argument console calls in worker mode when `nocapture=true`.

## Problem

When `console.log("a", "b")` is called in a worker with nocapture enabled, the output was `"a,b"` instead of `"a b"`. This happened because `args` (an array) was passed directly to `__wbg_test_output_writeln`, which concatenates with `+`, triggering JavaScript's implicit array-to-string coercion (which joins with commas).

## Fix

Change `args` to `args.map(String).join(' ')` to match the expected console output format where arguments are space-separated.

## Files Changed

- `crates/cli/src/wasm_bindgen_test_runner/server.rs` - Line 146

### DIFF
--- a/crates/cli/src/wasm_bindgen_test_runner/server.rs
+++ b/crates/cli/src/wasm_bindgen_test_runner/server.rs
@@ -145,7 +145,7 @@ pub(crate) fn spawn(
                 self.console[method] = function (...args) {{
                     og.apply(this, args);
                     if (nocapture) {{
-                        self.__wbg_test_output_writeln(args);
+                        self.__wbg_test_output_writeln(...args);
                     }}
                     if (self[on_method]) {{
                         self[on_method](args);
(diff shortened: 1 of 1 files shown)


crates/cli/src/wasm_bindgen_test_runner/server.rs
```diff
@@ -143,7 +143,7 @@ pub(crate) fn spawn(
                 const on_method = `on_console_${{method}}`;
                 self.console[method] = function (...args) {{
                     if (nocapture) {{
-                        self.__wbg_test_output_writeln(args);
+                        self.__wbg_test_output_writeln(args.map(String).join(' '));
```
guybedford: What about making the original function variadic rather?

drewcrawford: Updated to variadic function

PR #4866, opened 2025-12-15
Add codspeed and codecov cache

Spxg:
### Description

It will save some time.

### Checklist


Spxg:
> Do we need to hash the engine-related source files in the key here?

Done

PR #4867, opened 2025-12-16
Perf: remove redundant demangling string formatting

nickbabcock:
I ran wasm-bindgen through some input where it took a couple seconds to process and noticed demangle had some extra `to_string()` calls, so this commit reduces the number of times we're calling to_string() on a Demangle to 1.

Results of `demangle` as a portion of total sampled time

Before: 12%
After: 7.8%

The single display impl for rustc_demangle::Demangle now dominates, but it's a small improvement!

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -562,8 +562,7 @@ fn demangle(module: &mut Module) {
         let demangled = sym.to_string();
         match counter.entry(demangled) {
             Entry::Occupied(mut entry) => {
-                let count = *entry.get();
-                func.name = Some(format!("{}[{count}]", entry.key()));
+                func.name = Some(format!("{}[{}]", entry.key(), entry.get()));
                 *entry.get_mut() += 1;
             }
             Entry::Vacant(entry) => {


Spxg:
Thanks for the report! What is your benchmark environment? Because I got the opposite result:
```shell
     Running benches/my_benchmark.rs (target/release/deps/my_benchmark-7c43246d7ad57ed3)
Gnuplot not found, using plotters backend
to_string               time:   [6.2022 ns 6.2100 ns 6.2191 ns]
                        change: [−0.6304% −0.2131% +0.2755%] (p = 0.35 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe

string_clone            time:   [10.740 ns 10.747 ns 10.754 ns]
                        change: [−0.3133% −0.0118% +0.2994%] (p = 0.94 > 0.05)
                        No change in performance detected.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe

before                  time:   [37.131 ns 37.155 ns 37.179 ns]
                        change: [−0.4945% −0.1886% +0.1322%] (p = 0.22 > 0.05)
                        No change in performance detected.
Found 13 outliers among 100 measurements (13.00%)
  13 (13.00%) high severe

after                   time:   [42.946 ns 43.084 ns 43.239 ns]
                        change: [−0.5040% −0.1621% +0.1757%] (p = 0.38 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
```
```rust
use std::collections::{HashMap, hash_map::Entry};

pub fn foo(counter: &mut HashMap<String, i32>, sym: &str) -> String {
    let count = counter.entry(sym.to_string()).or_insert(0);

    if *count > 0 {
        format!("{sym}[{count}]")
    } else {
        sym.to_string()
    }
}

pub fn bar(counter: &mut HashMap<String, i32>, sym: &str) -> String {
    let demangled = sym.to_string();
    match counter.entry(demangled) {
        Entry::Occupied(mut entry) => {
            let s = format!("{}[{}]", entry.key(), entry.get());
            *entry.get_mut() += 1;
            s
        }
        Entry::Vacant(entry) => {
            let s = entry.key().clone();
            entry.insert(1);
            s
        }
    }
}
```

```rust
use std::collections::HashMap;

use benchxx::{bar, foo};
use criterion::{Criterion, criterion_group, criterion_main};

fn to_string(s: &str) -> String {
    s.to_string()
}

fn string_clone(s: &String) -> String {
    s.clone()
}

fn criterion_str(c: &mut Criterion) {
    let s1 = "hello world";
    let s2 = String::from("hello world");
    c.bench_function("to_string", |b| b.iter(|| to_string(s1)))
        .bench_function("string_clone", |b| {
            b.iter(|| string_clone(&s2));
        });
}

fn criterion_counter(c: &mut Criterion) {
    c.bench_function("before", |b| {
        b.iter(|| foo(&mut HashMap::new(), "hello world"))
    })
    .bench_function("after", |b| {
        b.iter(|| bar(&mut HashMap::new(), "hello world"))
    });
}

criterion_group!(benches, criterion_str, criterion_counter);
criterion_main!(benches);
```

nickbabcock:
Good point, I omitted some important details:

- Benchmarks were executed with `samply record --rate 10000 wasm-bindgen --target web` over a large internal wasm bundle (762 MB). 
- We should expect clone to be faster than to_string as the `impl Display for Demangle` is nontrivial ([source](https://github.com/rust-lang/rustc-demangle/blob/c5688cfec32d2bd00701836f12beb3560ee015b8/src/lib.rs#L313-L347)) (and the benchmark posted uses `&str` instead of Demangle)

Have you looked into what might be causing `to_string` to be faster than `string_clone`? I imagine they should be largely equivalent as`(&String).to_string()` will be flagged by clippy as an [implicit_clone](https://rust-lang.github.io/rust-clippy/master/index.html#implicit_clone)

Spxg:
> We should expect clone to be faster than to_string as the impl Display for Demangle is nontrivial ([source](https://github.com/rust-lang/rustc-demangle/blob/c5688cfec32d2bd00701836f12beb3560ee015b8/src/lib.rs#L313-L347))

oh.. you are right

crates/cli-support/src/lib.rs
```diff
@@ -559,15 +559,18 @@ fn demangle(module: &mut Module) {
...
-        } else {
-            sym.to_string()
-        });
-
-        *count += 1;
+        let demangled = sym.to_string();
+        match counter.entry(demangled) {
+            Entry::Occupied(mut entry) => {
+                let count = *entry.get();
+                func.name = Some(format!("{}[{count}]", entry.key()));
```
Spxg:
```suggestion
                func.name = Some(format!("{}[{}]", entry.key(), entry.get()));
```

let's keep same style.

Spxg:
> Have you looked into what might be causing `to_string` to be faster than `string_clone`?

https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#ptr_arg

https://godbolt.org/z/c75To8nvT `to_string`/`to_owned` seem to inline, which might result in better performance, may be you can try with your  large internal wasm bundle.

nickbabcock:
Yeah that godbolt clearly shows that `to_string()` inlines and I can see the call to clone when I dump the asm:

```
$ cargo asm -p wasm-bindgen-cli-support 'wasm_bindgen_cli_support::demangle'
... bl <alloc::string::String as core::clone::Clone>::clone
```

Amusingly the to_string and clone versions produce roughly the same number of lines assembly (593 lines vs 591) despite the inlining differences. 

When benchmarked internally, any performance difference are within margin of error.

Spxg (approved):

PR #4872, opened 2025-12-18
feat: support &[JsValue] passing to JS

guybedford:
Adds support for passing slices of JS value to JavaScript, while ensuring ownership is not lost.

### DIFF
--- a/guide/src/reference/types/jsvalue.md
+++ b/guide/src/reference/types/jsvalue.md
@@ -7,9 +7,7 @@
 ## `&[JsValue]` slices
 
 You can pass `&[JsValue]` slices from Rust to JavaScript. The slice is passed as
-a JavaScript `Array`. Since this is a borrowed slice, JavaScript receives a
-temporary view of the values - the array is only valid for the duration of the
-function call.
+a JavaScript `Array`.
 
 ```rust
 #[wasm_bindgen]


guybedford:
This was brought up in the demo I gave with arguments like `Box<[JsValue]>` being used currently for variadics where `&[JsValue]` would be simpler, so I was doing the follow-up from that feedback here.

In what way does it constitute a footgun do you think?

daxpedda:
No I remember and I think we should still implement this.
But I think we should be careful about giving out views to the user. The default implementation should not have lifetime issues like that.

This is why `Clamped` exists, it still makes those performance gains possible but users have to opt-in to the footgun.

To be completely honest I don't know if we have been entirely consistent with this view so far, at least that's what *I* think should be happening.

I should also say that this might not be ideal to begin with, maybe the default should be the other way around, but what it should be is consistent. Maybe something to re-think for the next version.

guybedford:
Note that the behaviour of passing an externref slice is a GC convention of creating a GC clone, even though a ref is passed. So the ref only affects the Rust side in that the Rust object is not consumed. From the JS perspective, it is a GC reference being passed with no lifetime associations.

Clamped only applies to the primitive array types like Uint8ClampedArray etc which are needed for certain APIs. This slice form is very different from that in being a bulk externref push into the externref slab though.

Personally the main thing I would do differently is avoid functions having to distinguish between &JsValue and JsValue and ensuring consistent conventions on that aspect for the Rust convention side of things.

guybedford:
I've uncommented the corresponding variadic test case to show how this enables the variadic usage as well.

guide/src/reference/types/jsvalue.md
```diff
@@ -4,6 +4,25 @@
 |:---:|:---:|:---:|:---:|:---:|:---:|:---:|
 | Yes | Yes | No | Yes | No | No | Any JavaScript value |
 
+## `&[JsValue]` slices
+
+You can pass `&[JsValue]` slices from Rust to JavaScript. The slice is passed as
+a JavaScript `Array`. Since this is a borrowed slice, JavaScript receives a
+temporary view of the values - the array is only valid for the duration of the
+function call.
```
daxpedda: So, apologies for the confusion.
I was under the impression that you are returning a view into the Wasm memory here, which is what `Clamped` is for.
(now that I'm thinking about it this wouldn't even work properly, because the `externref`s aren't in the linear memory)

However, looking at the code you clearly already do what I was suggesting: creating a JS `Array` and cloning every `externref`.

Why would JS receive only a temporary view here, this implies that there is some lifetime associated which it is not.


```suggestion
You can pass `&[JsValue]` slices from Rust to JavaScript. The slice is passed as
a JavaScript `Array`.
```

Am I missing something here?

guybedford: Yes you're right, this is misleading, thanks. There's nothing temporary about it from a JS perspective.

daxpedda (approved):
LGTM!

Sorry for the earlier confusion, after initially seeing the documentation I didn't actually check the code.

A reference test would be nice, but wouldn't hold the PR up for that.

guybedford:
Thanks, this will be really nice to use this in the variadic cases for the generics PR so I will land this one quickly to get that moving for the upcoming PR.

PR #4873, opened 2025-12-18
feat: allow Option<&T> and &Option<T> for exported Rust structs

guybedford:
Resolves https://github.com/wasm-bindgen/wasm-bindgen/issues/2370.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


guybedford:
We need test coverage for this that validates the borrow only lives as long as the call to JS has not returned, and that after that the pointer is zeroed out, perhaps via a separate invalidate call to set the pointer to zero like consuming supports.

PR #4874, opened 2025-12-18
vscode hint for editing tests

guybedford:
This adds a local `.vscode` settings.json file that contains the `--target wasm32-unknown-unknown`.

This is useful for editing tests, which don't automatically get picked up by Rust analyzer otherwise int he correct wasm32 target.

With this I'm able to debug js-sys tests with code completions.


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
+  "rust-analyzer.cargo.features": ["catch-unwind"],
```
guybedford: This is obviously not yet ready of course.

```suggestion
```

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
```
Spxg, line 2: I'm curious what the behavior of  wasm-bindgen-cli crate would be in this configuration.

guybedford: I can confirm code completions all work correctly in the CLI folder with this configuration, while the converse isn't true for the default target applying to Wasm.

PR #4876, opened 2025-12-19
feat: Erasable Generics

guybedford:
## Erasable Generic Type Parameters                                                            
                                                                                              
This PR adds support for erasable generic type parameters on imported JavaScript types, enabling usage of generic JavaScript APIs. In addition it includes a full backwards-compatible typing of js-sys to support generic core JS types like Array<T>, Promise<T>, Map<K, V>, and Iterator<T> with full inference performed by the Rust type system, while retaining sound type erasure at the JS bindgen boundary.

### ErasableGeneric Trait

The benefit of erasure generics is that it is actually quite simple and there is zero runtime overhead to the approach. The usage of erasable generics is designed based on a minimal trait that in theory should not inhibit any non-erasable variants in future. The changes to codegen have been refined multiple times to ensure the simplest possible bounds and transforms added.

Generic type erasure is implemented via the ErasableGeneric trait with an associated `Repr` type that specifies the concrete ABI representation:

```rust
trait ErasableGeneric {
  type Repr;  // The erased concrete type (default: JsValue)
}
```
                                                                                              
At compile time, a generic type like `Promise<Array<T>>` in a parameter or return type position is transmuted into `Promise<Array<JsValue>::Repr>` for ABI binding generation, based on substituting the known generic params as concrete params into the type expression. The soundness property that the type system then validates at compile time based on the specific types for erasure soundness is that `Promise<Array<T>>::Repr` equals `Promise<Array<JsValue>::Repr>::Repr`.

A useful shorthand for generics implementers wanting to constrain `ErasableGeneric<Repr = JsValue>` is provided with the new `JsGeneric` trait.

Bounds and diagnostics are carefully written to give good error messages when soundness checks fail, and the UI tests include tests for these cases.
                                            
While most generic types erase to JsValue, the erasure mechanism is more flexible than this, supporting:
* `Option<T>` and `Result<T, E>` implementing ErasableGeneric with appropriate representations
* `&[T]` slices when `T` has `ErasableGeneric<Repr = JsValue>`
* `&mut dyn FnMut(T) -> T` is erasable for arbitrary T that is itself ErasableGeneric, so far as there is repr alignment, enabling higher-order functions like `Array::forEach` to work with typed callbacks.

### Macro support

Macro support is extended to generate the new `Upcast` and `Promising` traits described in further detail below. These can be disabled with manual implementations (necessary to define important edge cases) via the new `no_upcast` and `no_promising` attributes.

Generics macro expansion support is handled as generically as possible, with the minimal intervention. The main complexity is impl block hoisting, which is carefully implemented to ensure correctness.
            
## js-sys Updates

All changes to js-sys are designed to be **fully backwards compatible**. Default generic parameters as `JsValue` provides the backwards compatibility path. All existing tests are retained the same.

Comprehensive generic implementations have been added across all of js-sys, forming a comprehensive update to the core JS utilities, including:

* `Array<T>`: Typed Array builtin.
* `ArrayTuple<(T1, ..., T8)>`: JS Array builtin variant with typing per item up to 9 items.
* `Function<fn(Arg) -> Ret>`: JS function external type, with a typed return and up to 9 arguments.
* `Generator<T>` and `AsyncGenerator<T>`: Typed generator builtin.
* `Map<K, V>`: Typed map builtin.
* `Iterator<T>`: Typed iterators.
* `AsyncIterator<T>`: Typed async iterators.
* `Object<T>`: Typed object records.
* `Reflect`: Provides generic operators.
* `Set<T>`: Typed sets.
* `WeakMap<K, V>`: Weak maps.
* `WeakSet<T>`: Typed weak sets.
* `WeakRef<T>`: Typed weak refs.
* `Promise<T>`: Typed promises.

In addition extra traits and types are added for finer-grained type support, including:

* `Iterable<T>`, `AsyncIterable<T>`: An object with a `[Symbol.async?iterator]()` function field implementing the iterator protocol.
* `Promising<T>`: Either `T` or `Promise<T>`.
* `PropertyDescriptor`
* `ArrayBufferOptions`
                                                                                              
Test coverage is fairly comprehensive over all changed files. Documentation is also fully updated to reflect the new generics support.

`Function::call` and `Function::bindn` are both defined to work now taking a tuple of `(&Arg, &Arg, ...)`. Then automatically applying the correct type-inferred arity call on the function types.

Temporal and Intl have been updated to latest generics support features.

WebAssembly has been typed on the next upgrade path to support genreics.

### js-sys next

While adding the generics, we've completely overhauled js-sys. Some functions get variations with generics (for example, try_for_each versus for_each on Array), some get generics functions with a deprecation that will merge into the original (for example, `Array::new_typed` or `Promise::then_map`, to be renamed back into `Array::new` and `Promise::then` in the next major).

To start to lay the groundwork for this upgrade path, js_sys_next_unstable configuration branches have been used to enable this and a test mode for both `just test-js-sys` and `just test-js-sys-next` are available to test both paths.

In the process Intl and Temporal have also been entirely refreshed with full typing. Since Temporal is already behind js_sys_next_unstable, we could make these changes directly. Intl has been guarded with the new typed paths, although brand new APIs that matured more recently that are added are now fully typed.

By having js-sys-next fully ready now, we can test it out and gain feedback on the design before finalizing the generics release decisions, which are now all fully worked through.

## Covariance

With straight generics, unsafe casting can be used to convert any generic type into any other generic type, but it turned out in practical usage that of course we want covariant casting here.

A new trait, `Upcast` is supported which is used to perform upcasting between generic types, using an `upcast()` method that is fully type safe.

For example, an `ArrayTuple<Promise<Function<Number>>, Set<Promise>>` can be upcast into `ArrayTuple<Promise<JsValue>, Set<Promise>>` or any other intermediate covariant variant safely using this trait.

### Deriving Covariance

Instead of adding yet another trait to the codegen, I took a different path here and added `Upcast` as a derive trait that can be custom placed on types. The default derive is full covariance, then functions that need custom upcast implementations like Function and Closure can still have their correct arg contravariance and return covariance applied outside of the generated derive. The trait also auto-generates upcast to jsvalue for all types.

### Closures into Function

When converting closures into functions, we previously had full generic loss on this. We now support a new `Function::from_closure_upcast` method for creating a typed `Function` from a typed closure.

It would be nice to expose the current unstable `Function::from_closure` as well, which does not require explcit inference unlike the upcast form, but this would involve solving the problem of mapping Rust types to their corresponding JS types, which could be a straightforward follow-on to this PR.

### JsOption type

One problem when casting a Function type is functions that take `Option` aren't supported since Rust's Option is not an erasable JsValue (externref) generic. The new `JsOption` is fully supported in function upcasting rules, allowing for example, functions of 3 arguments to be extended to functions of 4 arguments with the 4th argument a `JsOption` type. Similarly, for shrinking trailing `JsOption` arguments.

## web-sys Updates

To support generating generics WebIDL generation is updated to support a new experimental generics bindgen output mode. To ensure backwards-compatibility, all output changes not backwards compatible (pretty much most) are gated behind ensuring the new flag `next_unstable`.

When building under this mode, we now support typed iterators, typed records, typed setlike, typed maplike, iterable, array, and promise all using generics. In addition callbacks are now also typed using `Function<fn (Arg1, ...) -> Ret>` typing so that callbacks taking `Function` can now perform full type validation.

### DIFF
--- a/crates/macro/ui-tests/closure-variance.stderr
+++ b/crates/macro/ui-tests/closure-variance.stderr
@@ -1,29 +1,23 @@
-error[E0277]: the trait bound `i32: UpcastFrom<Number>` is not satisfied
+error[E0599]: no method named `upcast_ref` found for struct `ScopedClosure` in the current scope
  --> ui-tests/closure-variance.rs:7:51
   |
 7 |     let _bad: &Closure<dyn Fn() -> i32> = closure.upcast_ref();
-  |                                                   ^^^^^^^^^^ the trait `UpcastFrom<Number>` is not implemented for `i32`
+  |                                                   ^^^^^^^^^^
   |
-  = help: the following other types implement trait `UpcastFrom<S>`:
-            `i32` implements `UpcastFrom<i16>`
-            `i32` implements `UpcastFrom<i32>`
-            `i32` implements `UpcastFrom<i8>`
-  = note: required for `dyn Fn() -> i32` to implement `UpcastFrom<dyn Fn() -> Number>`
-  = note: 1 redundant requirement hidden
-  = note: required for `ScopedClosure<'_, dyn Fn() -> i32>` to implement `UpcastFrom<ScopedClosure<'_, dyn Fn() -> Number>>`
-  = note: required for `ScopedClosure<'_, dyn Fn() -> Number>` to implement `wasm_bindgen::prelude::Upcast<ScopedClosure<'_, dyn Fn() -> i32>>`
+help: there is a method `as_ref` with a similar name
+  |
+7 -     let _bad: &Closure<dyn Fn() -> i32> = closure.upcast_ref();
+7 +     let _bad: &Closure<dyn Fn() -> i32> = closure.as_ref();
+  |
 
-error[E0277]: the trait bound `i32: UpcastFrom<Number>` is not satisfied
+error[E0599]: no method named `upcast_ref` found for struct `ScopedClosure` in the current scope
   --> ui-tests/closure-variance.rs:11:52
    |
 11 |     let _bad2: &Closure<dyn Fn(Number)> = closure2.upcast_ref();
-   |                                                    ^^^^^^^^^^ the trait `UpcastFrom<Number>` is not implemented for `i32`
+   |                                                    ^^^^^^^^^^
    |
-   = help: the following other types implement trait `UpcastFrom<S>`:
-             `i32` implements `UpcastFrom<i16>`
-             `i32` implements `UpcastFrom<i32>`
-             `i32` implements `UpcastFrom<i8>`
-   = note: required for `dyn Fn(Number)` to implement `UpcastFrom<dyn Fn(i32)>`
-   = note: 1 redundant requirement hidden
-   = note: required for `ScopedClosure<'_, dyn Fn(Number)>` to implement `UpcastFrom<ScopedClosure<'_, dyn Fn(i32)>>`
-   = note: required for `ScopedClosure<'_, dyn Fn(i32)>` to implement `wasm_bindgen::prelude::Upcast<ScopedClosure<'_, dyn Fn(Number)>>`
+help: there is a method `as_ref` with a similar name
+   |
+11 -     let _bad2: &Closure<dyn Fn(Number)> = closure2.upcast_ref();
+11 +     let _bad2: &Closure<dyn Fn(Number)> = closure2.as_ref();
+   |
--- a/crates/macro/ui-tests/immediate-closure-variance.stderr
+++ b/crates/macro/ui-tests/immediate-closure-variance.stderr
@@ -1,33 +1,23 @@
-error[E0271]: type mismatch resolving `<ImmediateClosure<'_, dyn FnMut() -> Number> as ErasableGeneric>::Repr == &ImmediateClosure<'static, dyn FnMut() -> u32>`
+error[E0599]: no method named `upcast_ref` found for struct `wasm_bindgen::prelude::ImmediateClosure` in the current scope
   --> ui-tests/immediate-closure-variance.rs:11:68
    |
 11 |     let _bad: &ImmediateClosure<dyn FnMut() -> Number> = immediate.upcast_ref();
-   |                                                                    ^^^^^^^^^^ expected `u32`, found `JsValue`
+   |                                                                    ^^^^^^^^^^
    |
-   = note: expected reference `&'static wasm_bindgen::prelude::ImmediateClosure<'static, (dyn FnMut() -> u32 + 'static)>`
-              found reference `&'static wasm_bindgen::prelude::ImmediateClosure<'static, (dyn FnMut() -> JsValue + 'static)>`
-note: required by a bound in `upcast_ref`
-  --> $WORKSPACE/src/convert/traits.rs
+help: there is a method `upcast` with a similar name
    |
-   |     fn upcast_ref(&self) -> &T
-   |        ---------- required by a bound in this associated function
-...
-   |         T: Sized + ErasableGeneric<Repr = <Self as ErasableGeneric>::Repr>,
-   |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Upcast::upcast_ref`
+11 -     let _bad: &ImmediateClosure<dyn FnMut() -> Number> = immediate.upcast_ref();
+11 +     let _bad: &ImmediateClosure<dyn FnMut() -> Number> = immediate.upcast();
+   |
 
-error[E0271]: type mismatch resolving `<ImmediateClosure<'_, dyn FnMut(u32)> as ErasableGeneric>::Repr == &ImmediateClosure<'static, dyn FnMut(JsValue)>`
+error[E0599]: no method named `upcast_ref` found for struct `wasm_bindgen::prelude::ImmediateClosure` in the current scope
   --> ui-tests/immediate-closure-variance.rs:17:63
    |
 17 |     let _bad2: &ImmediateClosure<dyn FnMut(u32)> = immediate2.upcast_ref();
-   |                                                               ^^^^^^^^^^ expected `&ImmediateClosure<'_, dyn FnMut(...)>`, found `&ImmediateClosure<'_, dyn FnMut(u32)>`
+   |                                                               ^^^^^^^^^^
    |
-   = note: expected reference `&'static wasm_bindgen::prelude::ImmediateClosure<'static, (dyn FnMut(JsValue) + 'static)>`
-              found reference `&'static wasm_bindgen::prelude::ImmediateClosure<'static, (dyn FnMut(u32) + 'static)>`
-note: required by a bound in `upcast_ref`
-  --> $WORKSPACE/src/convert/traits.rs
+help: there is a method `upcast` with a similar name
    |
-   |     fn upcast_ref(&self) -> &T
-   |        ---------- required by a bound in this associated function
-...
-   |         T: Sized + ErasableGeneric<Repr = <Self as ErasableGeneric>::Repr>,
-   |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Upcast::upcast_ref`
+17 -     let _bad2: &ImmediateClosure<dyn FnMut(u32)> = immediate2.upcast_ref();
+17 +     let _bad2: &ImmediateClosure<dyn FnMut(u32)> = immediate2.upcast();
+   |
--- a/crates/webidl/src/wbg_type.rs
+++ b/crates/webidl/src/wbg_type.rs
@@ -8,8 +8,8 @@ use weedle::types::*;
 
 use crate::first_pass::FirstPassRecord;
 use crate::util::{
-    array, camel_case_ident, generic_ty, js_option_ty, option_ty, shared_ref, slice_ty,
-    snake_case_ident, Direction, TypePosition,
+    array, camel_case_ident, generic_ty, js_option_ty, option_ty, shared_ref, snake_case_ident,
+    Direction, TypePosition,
 };
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Debug)]
--- a/tests/wasm/unwind.rs
+++ b/tests/wasm/unwind.rs
@@ -239,9 +239,9 @@ fn drop_throw_str() {
             foo.foo();
             Ok(true)
         };
-        let closure = ScopedClosure::borrow(&mut func);
+        let closure = ScopedClosure::borrow_mut(&mut func);
         assert!(js_array![0]
-            .try_every_result_closure_borrow(closure.as_mut())
+            .try_every_result_closure_borrow(&closure)
             .is_err());
     }
     assert!(!Reflect::get(&global(), &"food_throw_str".into())


crates/js-sys/src/lib.rs
```diff
@@ -302,149 +303,192 @@ extern "C" {
...
+        predicate: &mut dyn FnMut(T, u32, Array<T>) -> bool,
     ) -> i32;
 
     /// The `flat()` method creates a new array with all sub-array elements concatenated into it
     /// recursively up to the specified depth.
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)
     #[wasm_bindgen(method)]
-    pub fn flat(this: &Array, depth: i32) -> Array;
+    pub fn flat<T>(this: &Array<T>, depth: i32) -> Array<T>;
```
RReverser: I think this should be similar to `flat_map` in that it returns a different type. After all, you do something like `[[1,2],[[3,4],5]] -> [1,2,3,4,5]`.

We probably can't properly type it with arbitrary depth, but at least having separate `<T, U>` sounds right.

crates/js-sys/src/lib.rs
```diff
@@ -302,149 +303,192 @@ extern "C" {
...
+        predicate: &mut dyn FnMut(T, u32, Array<T>) -> bool,
+    ) -> Array<T>;
 
     /// The `find()` method returns the value of the first element in the array that satisfies
     ///  the provided testing function. Otherwise `undefined` is returned.
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)
     #[wasm_bindgen(method)]
-    pub fn find(this: &Array, predicate: &mut dyn FnMut(JsValue, u32, Array) -> bool) -> JsValue;
+    pub fn find<T>(this: &Array<T>, predicate: &mut dyn FnMut(T, u32, Array<T>) -> bool) -> T;
```
RReverser, line 681: I wonder, here and elsewhere, is it possible to type it as

```rust
pub fn find<T>(this: &Array<T>, predicate: &mut dyn FnMut(T, u32, Self) -> bool) -> T;
```

or does macro not support that?

Rust impl blocks allow that, and it would make these callback signatures tiny bit less verbose (more so in types with more than 1 generic param).

crates/js-sys/src/lib.rs
```diff
@@ -514,58 +559,84 @@ extern "C" {
     ///
     /// There are a few bindings to `of` in `js-sys`: `of1`, `of2`, etc...
     /// with different arities.
+    #[wasm_bindgen(static_method_of = Array, js_name = of, variadic)]
+    pub fn of<T>(values: &[T]) -> Array<T>;
+
+    // Next major: Deprecate
```
RReverser: I think there's no harm in adding the `#[deprecated]` attribute with a note right now - adding one is not a breaking change, it's a hint for users to start migrating away from these methods, which is precisely what we want.

crates/js-sys/src/lib.rs
```diff
@@ -514,58 +559,84 @@ extern "C" {
...
     /// the array (from left to right) to reduce it to a single value.
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)
     #[wasm_bindgen(method)]
-    pub fn reduce(
-        this: &Array,
-        predicate: &mut dyn FnMut(JsValue, JsValue, u32, Array) -> JsValue,
+    pub fn reduce<T>(
+        this: &Array<T>,
+        predicate: &mut dyn FnMut(JsValue, T, u32, Array<T>) -> JsValue,
```
RReverser, line 1005: Is it possible to mark accumulator as 2nd generic param or do we only support one param at the moment?

crates/js-sys/src/lib.rs
```diff
@@ -607,47 +696,436 @@ extern "C" {
...
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/with)
+    #[wasm_bindgen(method, js_name = with)]
+    pub fn with<T>(this: &Array<T>, index: u32, value: &T) -> Array<T>;
+}
+
+/// Marker type for an unused generic argument
+/// This allows restricting implementations to a specified number of non-default args only.
+/// For example, for `tuple: ArrayTuple<JsString, Number>`, it will not support `tuple.get2`,
+/// because NeverArg does not support ErasableGeneric.
+pub struct NeverArg;
```
RReverser: The naming is confusing. If it's meant to be truly unused / non-constructible, it's better to use proper "never" / uninhabited type. I'd also make it private, users shouldn't access it.

```suggestion
enum NeverArg {}
```

crates/js-sys/src/lib.rs
```diff
@@ -1925,21 +2643,215 @@ extern "C" {
...
+    /// The `Function` constructor creates a new `Function` object. Calling the
+    /// constructor directly can create functions dynamically, but suffers from
+    /// security and similar (but far less significant) performance issues
+    /// similar to `eval`. However, unlike `eval`, the `Function` constructor
+    /// allows executing code in the global scope, prompting better programming
+    /// habits and allowing for more efficient code minification.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function)
+    #[wasm_bindgen(constructor)]
+    pub fn new3<R: ErasableGeneric, A1: ErasableGeneric, A2: ErasableGeneric, A3: ErasableGeneric>(
```
RReverser: Could we merge all of these via your Iterable interface + variadic?

crates/js-sys/src/lib.rs
```diff
@@ -2287,52 +3397,176 @@ impl Default for Function {
...
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/next)
+    #[wasm_bindgen(method, catch)]
+    pub fn next<T>(this: &Generator<T>, value: &T) -> Result<JsValue, JsValue>;
+
+    /// The `next()` method returns an object with two properties done and value.
+    /// You can also provide a parameter to the next method to send a value to the generator.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/next)
+    #[wasm_bindgen(method, catch)]
+    pub fn next_void<T: FromWasmAbi>(this: &Generator<T>) -> Result<IteratorNext<T>, JsValue>;
```
RReverser: Not sure what's better yet, but "next_void" is a bit confusing naming, it makes it sound like the function would return void so got me stuck for a moment.

RReverser: Maybe just `next_without_value` to match the one below?

crates/js-sys/src/lib.rs
```diff
@@ -2578,6 +3840,34 @@ pub fn try_iter(val: &JsValue) -> Result<Option<IntoIter>, JsValue> {
     Ok(Some(it.into_iter()))
 }
 
+/// Trait for types known to implement the iterator protocol on Symbol.iterator
+/// TODO: Add direct typed [Symbol.iterator] iterator function, when we support symbol functions.
```
RReverser: TODOs should be regular comments, not part of doc comment.

crates/js-sys/src/lib.rs
```diff
@@ -3722,45 +5127,58 @@ extern "C" {
...
+    #[wasm_bindgen(static_method_of = Object, catch, js_name = fromEntries)]
+    pub fn from_entries(entries: &JsValue) -> Result<Object, JsValue>;
 
     /// The `Object.fromEntries()` method transforms a list of key-value pairs
     /// into an object.
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries)
     #[wasm_bindgen(static_method_of = Object, catch, js_name = fromEntries)]
-    pub fn from_entries(iterable: &JsValue) -> Result<Object, JsValue>;
+    pub fn from_entries_typed<T, I: Iterable<Item = ArrayTuple<JsString, T>>>(
```
RReverser: I wonder if we could merge methods like this actually. What if you just define `Iterable for JsValue { type Item = JsValue; }`?

Then `from_entries_typed` becomes a generic case of `from_entries, and works on either `&JsValue` or `&SomeProperlyTypedIterable`, returning backward-compatible `Object<JsValue>` in first case and `Object<T>` in 2nd.

We could even mark that  `Iterable for JsValue { type Item = JsValue; }` as insta-`#[deprecated]` so that people can keep using it for backward-compat but are encouraged to migrate to strongly-typed collections.

RReverser: Looking throughout other bindings, looks like that would allow to remove quite a few other duplicated `_typed` variants.

crates/js-sys/src/lib.rs
```diff
@@ -4629,21 +6172,24 @@ pub mod WebAssembly {
...
-        pub fn compile_streaming(response: &Promise) -> Promise;
+        pub fn compile_streaming(response: &Promise) -> Promise<WebAssembly::Module>;
 
         /// The `WebAssembly.instantiate()` function allows you to compile and
         /// instantiate WebAssembly code.
         ///
         /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/instantiate)
         #[wasm_bindgen(js_namespace = WebAssembly, js_name = instantiate)]
-        pub fn instantiate_buffer(buffer: &[u8], imports: &Object) -> Promise;
+        pub fn instantiate_buffer(buffer: &[u8], imports: &Object) -> Promise<WebAssembly::Module>;
```
RReverser: I love this signature obviously, but aren't changes like this a breaking change?

When someone gets the value from such promise either via futures integration, or via `then2` callback, then the value will be `WebAssembly::Module` not `JsValue` which can lead to all sorts of type conflicts.

When using these types by reference, it's not a concern because we have a `Deref<Target = JsValue>` on these types, but when using by value, I think it's breaking.

crates/js-sys/src/lib.rs
```diff
@@ -6253,30 +7957,62 @@ extern "C" {
...
 
+    /// The `Promise.resolve(value)` method returns a `Promise` object that is
+    /// resolved with the given value. If the value is a promise, that promise
+    /// is returned; if the value is a thenable (i.e. has a "then" method), the
+    /// returned promise will "follow" that thenable, adopting its eventual
+    /// state; otherwise the returned promise will be fulfilled with the value.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)
+    #[wasm_bindgen(static_method_of = Promise, js_name = resolve)]
+    pub fn resolve_typed<T>(obj: &T) -> Promise<T>;
```
RReverser: I think this and other similar methods don't need to be duplicated with `_typed` suffixes.

We needed it in the `Array::new` case because, as you pointed out, Rust needs an explicit type there because it doesn't have anything to infer it from and the default doesn't cut it.

For methods like this, however, there is an input value that is enough for the Rust inference to go on, so having a `Promise<T = JsValue>` and a single `pub fn resolve<T>(obj: &T) -> Promise<T>` is enough for both backward-compat declaration and a new typed extension.

crates/macro-support/src/codegen.rs
```diff
@@ -974,13 +976,90 @@ impl ToTokens for ast::ImportType {
             }
         };
 
+        let class_generic_params = generics::generic_params(&self.generics);
+        let (impl_generics, ty_generics, where_clause) = self.generics.split_for_impl();
+
+        let impl_generics_inner: proc_macro2::TokenStream = impl_generics
```
RReverser: Oof, why do we need to stringify and reparse here? Can't we extract what we need straight from the syn AST?

crates/js-sys/src/lib.rs
guybedford: We can have the accumulator as a second generic, but that would create possible inference compilation errors without that generic being explicitly provided in usage, so it might be better done as a next major or as a new function.

crates/js-sys/src/lib.rs
guybedford: I think we might need to add some extra support to the macro parsing to support `Self`. I can certainly look into it though.

guybedford: The first argument would still need to be the explicit form, and then that would infer `Self` basically - it's quite a bit of wiring though to add a visitor replacement for that based on an arbitrary type expression. So this isn't free.

crates/js-sys/src/lib.rs
guybedford: For `let f = Function::new(args, body)`, if that was variadic, then you'd need to annotate the `let f: Function = ...` type to get the default inference, and there would be no link between the number of args and number of required generic params like we have here.

So then for the other cases you could use `let f: Function<Number, JsString, Foo>` to indicate a function `F(JsString, Foo) -> Number`.

Not sure if that would be better?

crates/js-sys/src/lib.rs
guybedford: Yeah will be better to revert this line and come back to it next major.

crates/js-sys/src/lib.rs
guybedford: You may well be right, I thought I was getting some errors in third-party projects, will report back if I still see that with this.

crates/js-sys/src/lib.rs
guybedford: This must be public though since it is obtainable in Rust constraint theory through the public interface for an array tuple (even though not practically).

RReverser: > This must be public though since it is obtainable in Rust constraint theory through the public interface for an array tuple

Yeah but Rust allows private types in public interface for many versions now, and even in the past you could do that through the so-called voldemort types.

It's still a better interface to prevent user from explicitly accessing it, plus makes it easier for us to properly remove it in the future as a non-breaking change.

crates/js-sys/src/lib.rs
RReverser: Yeah leaving first arg as the source of truth is fine, but I don't think you need a visitor replacement for other instances?

If you pass types of those other args verbatim, they'll end up generated as something like

```rust
impl<T> Array<T> {
  fn find(&self, predicate: &mut dyn FnMut(T, u32, Self) -> bool) -> T { ... }
}
```

which is perfectly fine.

crates/js-sys/src/lib.rs
RReverser: Hm yeah I think we can do something nicer here with proper tuples in the future, but we can leave that as a future improvement.

That said, I wonder if we can simply omit those extra impls for now. The number of arguments already doesn't necessarily match number of strings, e.g. I've seen (& done) `new Function("a,b,c", "body")` often enough as a shorthand for `new Function("a", "b", "c", "body")` so the existing binding is kinda already variadic.

crates/js-sys/src/lib.rs
guybedford: Well it doesn't compile if I remove it.

RReverser: What error do you get? It should only emit a warning that is easy to suppress on specific usage.

crates/js-sys/src/lib.rs
guybedford: The problem is that for the erasable generics we still need to internally substitute `Self` with `Array<T>` to then obtain the concrete `Array<JsValue>::Repr`. So making `Self` work is not a simplification, but a complication here.

For this reason, I would rather avoid the unnecessary work unless there is a strong use case.

crates/js-sys/src/lib.rs
guybedford: Sure, I can go through and add explicit deprecated at this point, just wanted to ensure we had support for this first.

crates/js-sys/src/lib.rs
guybedford: The error comes up for:

```rust
impl<K, V> Iterable for Map<K, V> {
    type Item = ArrayTuple<K, V>;
}
```

as:

```
private type `NeverArg` in public interface
can't leak private typerustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///home/gbedford/Projects/wasm-bindgen/crates/js-sys/src/lib.rs)
lib.rs(811, 1): `NeverArg` declared as private
js_sys::Map
type Item = ArrayTuple<K, V>
Implements notable traits: Future<Output = <ArrayTuple<{unknown}, {unknown}, NeverArg, NeverArg, NeverArg, NeverArg, NeverArg, NeverArg, NeverArg> as Future>::Output>, Iterator<Item = <ArrayTuple<{unknown}, {unknown}, NeverArg, NeverArg, NeverArg, NeverArg, NeverArg, NeverArg, NeverArg> as Iterator>::Item>, Read, Write
```

crates/js-sys/src/lib.rs
guybedford: I was actually wondering if we could deprecate `new Function("a, b, c", "body")` now that we've got `new Function("a", "b", "c", "body")` which wasn't previously supported. I think avoiding custom string splitting is better but that's just my preference.

crates/js-sys/src/lib.rs
guybedford: I think passing undefined is maybe not so bad, so have refactored to `next_iterator` instead here. Further thoughts welcome.

crates/js-sys/src/lib.rs
guybedford: This works well for this case, unfortunately I couldn't find other cases to apply it to - `Promise::all` returns a `Promise<JsValue>` not `Promise<Array<JsValue>>` which would be required for backwards compat to work on that. Same for all_settled. `Promise::any` and `Promise::race` iterables do also work actually. `Array::from_iterable` was changed to return a result so also can't align unless we separate the result case as a new function rather and keep the non-result form as non-deprecated.

guybedford: Unfortunately another issue here is the extra abstraction of `I: Iterable<T>` results in Rust not being able to infer `T` despite having all the info it needs, so explicit type hints are still needed in these trait cases which is unfortunate.

guybedford: For `from_entries_typed` itself we get this error:

```
error[E0271]: type mismatch resolving `<Array as Iterable>::Item == ArrayTuple<JsString, _>`
    --> crates/js-sys/tests/wasm/Object.rs:137:39
     |
 137 |     let object = Object::from_entries(&array).unwrap();
     |                  -------------------- ^^^^^^ expected `ArrayTuple<JsString, _>`, found `JsValue`
     |                  |

```

crates/js-sys/src/lib.rs
guybedford: This seems to work. I will report back if I find another issue here.

crates/js-sys/src/lib.rs
RReverser: Fine by me, it's mostly a nice to have thingy for writing own bindings rather than something we have to support.

crates/js-sys/src/lib.rs
RReverser: Yeah that should be the warning I mentioned, I guess in our case warnings are just promoted to errors?

Usually fixed via either `allow(private_bounds)` or `private_interfaces`, the Cargo message instead of RA one should have the lint name, or you could just try either.

If not, I'll take a look.

RReverser: I guess we could also default them to `()` instead - in most (all?) JS usecases treating items beyond the length of the array as `undefined` is fine and maybe even the right thing to do.

crates/macro-support/src/codegen.rs
guybedford: Thanks, corrected.

crates/js-sys/src/lib.rs
RReverser: > I think avoiding custom string splitting

Oh but it's not us, we don't do string splitting, it's just something `new Function` takes directly in ECMAScript spec since its inception.

It always said the arguments in variadic version will be joined by the comma, but otherwise it doesn't care what those strings are and they will be preserved verbatim.

That's why you can pass 1 arg in one string, 2 args in one string and 3 in another, even insert comments etc like `new Function('/* this is a */ a, b /* and this was b */', 'return a+b+c')` and they all will do the same thing and be preserved with comments and whitespace in `func.toString()`.

In a way, it's the variadic version that requires both us to do more work - sending array of individual strings over the JS Wasm boundary - and the engine to do some extra work - joining them via comma before constructing the function.

crates/js-sys/src/lib.rs
RReverser: > Unfortunately another issue here is the extra abstraction of `I: Iterable<T>` results in Rust not being able to infer `T` despite having all the info it needs

Hm I think that actually hints at a potential issue/improvement in Iterable itself.

On Rust side it should use an associated type, since you can only be Iterable over one type, just like Rust's own Iterator and IntoIter traits have associated Item.

With that change, the inference should work out of the box.

guybedford: Iterable does use an associated type and can do this inference. The problem is that `JsValue` is an iterable for `JsValue` not for `ArrayTuple<JsString, T>`.

crates/js-sys/src/lib.rs
guybedford: The benefit of `NeverArg` is that we don't actually even implement `get8` etc, since these only work for inhabitable params. If we use `()` we lose the ability to do uninhabitable definitions and the ability to do the static `len()` because `()` supports ErasableGeneric trait.

An uninhabitable type is usually an important concept in a type system - what's wrong with having it here?

crates/js-sys/src/lib.rs
guybedford: Yes it could be argued the existing form is more performant, but the new form will give generic support across all args in calling, which I think is what we should encourage going forward.

If we want to just strongly type the old form, we can't auto-infer type length so would need to have new1..8 anyway (functions can't have generic defaults). Ideally we'd have uninhabiltable params for the others per `NeverArg`, but we can't do this for functions until the next major.

crates/js-sys/src/lib.rs
RReverser: I was referring to

> Unfortunately another issue here is the extra abstraction of `I: Iterable<T>` results in Rust not being able to infer `T` despite having all the info it needs

But now I'm confused. Are there two different Iterables or did you use `Iterable<T>` in that quote as a shorthand / pseudocode?

crates/js-sys/src/lib.rs
RReverser: > we can't auto-infer type length

TBH given how "weird" `new Function` is and how it's relatively rare, maybe we should just leave it be and require user to explicitly specify the output signature?

There is no way for us to express some of the more exotic signatures like `new Function('a', '...b', 'return a + b.length')` or `new Function('a,b', 'c,d', 'return a+b+c+d')` anyway - I think it's easier both for us and the user if we required an explicit signature passed via type params or an explicit cast on the result of `Function::new()`.

crates/js-sys/src/lib.rs
RReverser: > For `from_entries_typed` itself we get this error:

As for this one, yeah I see, looks like we do need to keep them separate.

guybedford: > But now I'm confused. Are there two different Iterables or did you use Iterable<T> in that quote as a shorthand / pseudocode?

Yes, that was shorthand for `Iterable<Item = T>`.

crates/js-sys/src/lib.rs
RReverser: > An uninhabitable type is usually an important concept in a type system - what's wrong with having it here?

Oh yeah it definitely has its uses, it's just that usually it's designed to declare a value as non-constructible and that propagates upwards making the whole value non-constructible.

Here though we're trying to express eg `[string, number]` as `[string, number, Something, Something, Something]` with padding up to our supported number of args which suggests it should be `[string, number, undefined, undefined, undefined]` not `[string, number, never, never, never]`.

> The benefit of `NeverArg` is that we don't actually even implement `get8` etc, since these only work for inhabitable params. If we use `()` we lose the ability to do uninhabitable definitions and the ability to do the static `len()` because `()` supports ErasableGeneric trait.

This is a strong counterargument from UX point of view though. I agree, let's keep it never arg just hide it so users don't rely on it.

guybedford: Ok, keeping it public, but now defined as an uninhabitable enum at least.

RReverser: > Ok, keeping it public

But why? Did you try the `allow()` suggestions above?

Let's keep this conversation open so that we don't lose it - worst case, as I said I'm happy to address this one on my own when back next week.

guybedford: Sure, I thought we had addressed all points - yes I tried the suggestions but they did not work for me.

RReverser: > Sure, I thought we had addressed all points

Ah, it just looked funny right after my last sentence, it probably got lost:

> I agree, let's keep it never arg just hide it so users don't rely on it.

crates/js-sys/src/lib.rs
guybedford: With some refinement and testing, I've now added just a single `new_with_args_typed<...>` which works well with inference and usage.

For all constructors we now consistently follow the pattern that we provide the exact same original constructors but with a single `_typed` suffix which in the next major can replace the original constructor.

Marking this as resolved, all the `newn` variations are now fully removed, but feel free to reopen discussion.

crates/js-sys/src/lib.rs
guybedford: Typing this properly is actually possible, but would require us to use a concrete union type like the enum approach implemented in https://github.com/wasm-bindgen/wasm-bindgen/pull/4734 to allow each element to be a concrete type that also captures the individual union variations.

I still think it would be beneficial to explore this direction further, but for now I've implemented `flat<T> -> Array<JsValue>` to then allow users to do a custom cast after calling flat. If we made it generic on `U` right now, the required generic inference could possibly be a breaking change otherwise.

crates/js-sys/src/lib.rs
```diff
@@ -1570,6 +2531,14 @@ extern "C" {
     pub fn value_of(this: &Boolean) -> bool;
 }
 
+impl Boolean {
+    /// Typed Boolean true constant.
+    pub const TRUE: Boolean = unsafe { core::mem::transmute(JsValue::TRUE) };
+
+    /// Typed Boolean false constant.
+    pub const FALSE: Boolean = unsafe { core::mem::transmute(JsValue::TRUE) };
```
drewcrawford: Appears to be the wrong constant

crates/js-sys/src/lib.rs
```diff
@@ -2578,6 +3900,32 @@ pub fn try_iter(val: &JsValue) -> Result<Option<IntoIter>, JsValue> {
     Ok(Some(it.into_iter()))
 }
 
+/// Trait for types known to implement the iterator protocol on Symbol.iterator
+pub trait Iterable {
+    type Item;
+}
+
+impl<'a, T: Iterable> Iterable for &'a T {
+    type Item = &'a T::Item;
```
drewcrawford: It is unusual for Item to a be a reference type, not sure if intentional?

crates/macro-support/src/codegen.rs
```diff
@@ -1356,20 +1453,81 @@ impl TryToTokens for ast::ImportFunction {
...
+                        arguments.push(quote! { #name: #ty });
+                        parse_quote! { #inner_ty: #wasm_bindgen::__rt::marker::ErasableGenericOwn<#concrete_ty> }
+                    });
+                }
+                abi_ty = if let Some(mut_) = ref_mut {
+                    quote! { & #mut_ #concrete_ty }
+                } else {
+                    quote! { #concrete_ty }
+                };
+                convert_arg = quote! { unsafe { core::mem::transmute_copy(&core::mem::ManuallyDrop::new(#var)) } };
```
drewcrawford, line 1644: I think this relies on T being memory-layout-compatible with T::Repr.  Which I think holds, but not simple to verify.

One hardening tactic is to do move the transmute into a `ErasableGeneric.into_repr(self) -> Self::Repr` method, which typechecks the input and output type of the transmute.

Another hardening tactic is comparing their layouts with an assert, but may be overkill

crates/js-sys/src/lib.rs
```diff
@@ -607,47 +884,472 @@ extern "C" {
...
+/// For example, for `tuple: ArrayTuple<JsString, Number>`, it will not support `tuple.get2`,
+/// because NeverArg does not support ErasableGeneric.
+pub enum NeverArg {}
+
+// Tuples as a typed array variant
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_name = Array, extends = Array, is_type_of = Array::is_array, typescript_type = "Array<any>")]
+    #[derive(Clone, Debug)]
+    pub type ArrayTuple<
```
drewcrawford: This is just a shower thought I haven't explored in much detail.  But I wonder if it would be cleaner to separate by arity

```rust
  #[wasm_bindgen]
  extern "C" {
      #[wasm_bindgen(js_name = Array, extends = Array)]
      pub type Tuple2<T1, T2>;

      #[wasm_bindgen(js_name = Array, extends = Array)]
      pub type Tuple3<T1, T2, T3>;

      // ... up to Tuple9
  }

//possible extension
  pub struct MapEntry<K, V>(Tuple2<K, V>);
  pub struct ArrayEntry<T>(Tuple2<u32, T>);
```

Pros:
* Tuple2<K, V> is shorter than ArrayTuple<K, V, NeverArg, NeverArg, NeverArg...>
* May be more flexible to higher numbers of variadic parameters?
* No NeverArg in public API at all
* Clear in error messages and IDE tooltips

Cons:
* more macros to write
* ?

crates/js-sys/src/lib.rs
guybedford: Thanks, fixed.

crates/js-sys/src/lib.rs
guybedford: Good catch, thanks.

crates/macro-support/src/codegen.rs
guybedford: The ErasableGeneric trait is unsafe for good reason indeed. I didn't want to expose the erasable public method, to avoid unnecessary API surface, but that is an option.

crates/js-sys/src/lib.rs
guybedford: I guess the question here is if `Tuple2` significantly improves debuggability. I don't personally find `Never` that ugly in debug, but maybe that's just me.

We could probably get away with just implementing `Tuple2`, `Tuple3` and `Tuple4` only, although it would probably be _more_ code to do this.

I'm open to it, `ArrayTuple` was a fun exercise in generics, but I'm not averse to switching.

crates/macro-support/src/codegen.rs
drewcrawford: Those are good points!  What about a plain fn such as

```rust
pub(crate) fn transmute_generic<E:ErasebleGeneric>(input: E) -> E::Repr { 
    let input = ManuallyDrop::new(input);
    //safety: we typechecked it to an unsafe trait
    unsafe { ptr::read((&*input as *const E).cast::<E::Repr>()) }
 }
```

crates/js-sys/src/lib.rs
drewcrawford: Is there much potential here for nested types?  Or is it mostly you have 1 generic type at toplevel?

guybedford: Generics are arbitrarily nested and that is an important property yes.

It's also worth noting that in most debug logs, default generic arguments are left out so you do only see `ArrayTuple<Number, Number>` etc.

guybedford: For example, here's the error I get for a wrong ArrayTuple assignment:

```rust
error[E0308]: mismatched types
    --> crates/js-sys/tests/wasm/ArrayTuple.rs:7:54
     |
   7 | ...<Number> = ArrayTuple::new1(&JsString::from("first"));
     |               ---------------- ^^^^^^^^^^^^^^^^^^^^^^^^ expected `&Number`, found `&JsString`
     |               |
     |               arguments to this function are incorrect
     |
     = note: expected reference `&js_sys::Number`
                found reference `&js_sys::JsString`
note: associated function defined here
    --> /home/gbedford/Projects/wasm-bindgen-generics/crates/js-sys/src/lib.rs:1036:12
     |
1036 |     pub fn new1<T1>(t1: &T1) -> ArrayTuple<T1>;
     |            ^^^^
```

drewcrawford: LGTM

crates/macro-support/src/parser.rs
```diff
@@ -1023,10 +1045,13 @@ fn function_from_decl(
     if sig.variadic.is_some() {
         bail_span!(sig.variadic, "can't #[wasm_bindgen] variadic functions");
     }
-    if !sig.generics.params.is_empty() {
+
+    // For imported functions (Extern position), generics are supported and validated.
+    if matches!(position, FunctionPosition::Extern) {
+    } else if !sig.generics.params.is_empty() {
```
RReverser: Nit, the empty if looks a bit odd to read

```suggestion
    if !matches!(position, FunctionPosition::Extern) && !sig.generics.params.is_empty() {
```

crates/macro-support/src/codegen.rs
guybedford: That's still making it a public function though. I recently just added a type-safe `upcast()` method which is effectively this same thing though. It doesn't work for JsValue -> AnyType though because that is of course not type safe so doesn't satisfy the trait bounds. I think for a method exposed to users, type safety is far more beneficial over a general purpose erasure transmute.

guybedford:
@RReverser thanks for the feedback and figuring out the Never fix.

One issue with making it private is being unable to declare functions with no return - this test is failing:

```rust
error[E0412]: cannot find type `Never` in this scope
   --> crates/js-sys/tests/wasm/Object.rs:239:33
    |
239 |     let setter: BoundedFunction<Never, Number> = Function::new_with_args_typed("x", "");
    |                                 ^^^^^ not found in this scope
    |
help: you might be missing a type parameter
    |
229 | fn property_descriptor_get_set<Never>() {
    |                               +++++++
```

I could instead have a separate type `VoidBoundedFunction` to resolve this though, will add that.

With regards to `upcast()`, the benefit of this is comprehensive type safety in the same way TypeScript would check for JS. For example, you can't provide an array of String to a function expecting an array of Number. Or also that you can't provide a Function that returns an array of Number to an argument that expects a Function that returns a String, but you can provide it as an argument that expects a Function that returns an array of values. Upcast comprehensively solves type safety, in a way that no other construct handles all recursive covariant and contravariant invariants.

RReverser:
> One issue with making it private is being unable to declare functions with no return

Hm I'm pretty sure these functions should return `()`. Functions that return never type in the type system are reserved for infinite loops and panics, since that's the type they "return":

```rust
pub fn foo() -> ! {
    loop {} // ok
}

pub fn bar() -> ! {
    panic!("") // ok
}

pub fn baz() -> ! {
    // error: mismatched types, because we declare return type to be "never" but it does return
}
```

So it makes sense to use never for argument positions that can't be passed in, but the return type in scenario like you've shown must be `()` instead.

guybedford:
In our TypeScript-like type-safe type system overlaying the Rust trait system, `()` implements `ErasableGeneric` and thus blanket implementations for ErasableGeneric (which all bindgen is) cannot exclude it. This is a requirement to be able to provide polymorphic variations of functions with and without return values, which we are able to do with our `Never` type. These are distinct type systems - the overlay type system and the Rust type system. So we can pick the representation of a function that returns no argument versus a function without a return. If `Never` is a confusing name we can call it `Empty` as well. It must be distinct from `()` though because `ErasableGeneric` is supported for closure casting, and closures support `()`, so there is no way around that unless we remove support for having variations of implementations for functions with and without returns.

guybedford:
For example, this is how we vary the implementation of `apply` between functions with and without a return value:

```rust
    #[wasm_bindgen(method, catch)]
    pub fn apply<R: ErasableGeneric, A1, A2, A3, A4, A5, A6, A7, A8, A9>(
        this: &Function<R, A1, A2, A3, A4, A5, A6, A7, A8, A9>,
        context: &JsValue,
        args: &Array,
    ) -> Result<R, JsValue>;

    /// The `apply()` method calls a function with a given this value, and arguments provided as an array
    /// (or an array-like object).
    ///
    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply)
    #[wasm_bindgen(method, catch)]
    pub fn apply<A1, A2, A3, A4, A5, A6, A7, A8, A9>(
        this: &Function<Never, A1, A2, A3, A4, A5, A6, A7, A8, A9>,
        context: &JsValue,
        args: &Array,
    ) -> Result<(), JsValue>;
```

If we made `()` the Never type, then we'd have to explicitly run through the `()` ABI return path instead of having compile-time separation.

Would it help to rename `Never` to `Empty` or `Hole`? Alternatively, we can remove the compile-time polymorphism on the return being defined versus not provided to unify these all as singular `()` FromWasmAbi paths.

guybedford:
Oh of course the real reason we can't unify `()`, is that `()` is not Erasable-compatible with `JsValue`, so we would be required to define `Undefined` as our return type otherwise, not `()`.

RReverser:
I see, it's a bit more complicated than I realised. Are there other overlay types that differ from their Rust representation?

Does this empty/never/void type get exposed to the user in any way - eg if they call such function callback, do they get regular Rust "()" back or our "Never"? Presumably the former since we made our type uninhabited?

guybedford:
> I see, it's a bit more complicated than I realised. Are there other overlay types that differ from their Rust representation?

No, the uninhabitable type is unique in the type system - there is only one. Functions returning or taking it as an argument or return value will fail to compile, it can only be represented abstractly as an unused generic parameter.

RReverser:
> Functions returning or taking it as an argument or return value will fail to compile,

but what about your example above

> ```rust
>     /// The `apply()` method calls a function with a given this value, and arguments provided as an array
>     /// (or an array-like object).
>     ///
>     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply)
>     #[wasm_bindgen(method, catch)]
>     pub fn apply<A1, A2, A3, A4, A5, A6, A7, A8, A9>(
>         this: &Function<Never, A1, A2, A3, A4, A5, A6, A7, A8, A9>,
>         context: &JsValue,
>         args: &Array,
>     ) -> Result<(), JsValue>;
> ```

why does this function return uninhabited type in the signature if it's meant to compile and actually return `()`?

crates/js-sys/src/lib.rs
```diff
@@ -607,47 +886,541 @@ extern "C" {
...
+        t1: &T1,
+        t2: &T2,
+        t3: &T3,
+        t4: &T4,
+        t5: &T5,
+        t6: &T6,
+        t7: &T7,
+        t8: &T8,
+        t9: &T9,
+    ) -> ArrayTuple<T1, T2, T3, T4, T5, T6, T7, T8, T9>;
```
RReverser: Looking at all these makes me really wish that `variadic` attribute could take arbitrary tuples and auto-convert them to rest params via `ArrayTuple`, similarly to how it already handles homogenous `&[T]` params.

Oh well, a future improvement.

crates/js-sys/src/lib.rs
```diff
@@ -2578,15 +4428,41 @@ pub fn try_iter(val: &JsValue) -> Result<Option<IntoIter>, JsValue> {
...
+
+impl<T: Iterable> Iterable for &T {
+    type Item = T::Item;
+}
+
+impl Iterable for JsValue {
+    type Item = JsValue;
+}
+
+/// Trait for types known to implement the iterator protocol on Symbol.iterator
```
RReverser: Comment needs updating for this one.

crates/js-sys/src/lib.rs
```diff
@@ -3642,78 +5528,213 @@ extern "C" {
...
-    pub fn create(prototype: &Object) -> Object;
+    pub fn create<T>(prototype: &Object<T>) -> Object;
+
+    // Next major: deprecate and rename over create
+    /// The `Object.create()` method creates a new object, using an existing
+    /// object to provide the newly created object's prototype.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create)
+    #[wasm_bindgen(static_method_of = Object, catch)]
+    pub fn create_typed<T, U>(prototype: &Object<T>) -> Result<Object<U>, JsValue>;
```
RReverser: I think these should be normally the same (T=U), as usually you'd do that with a class, and any Class.prototype nominally has type Class itself.

crates/js-sys/src/lib.rs
```diff
@@ -6288,7 +8833,16 @@ extern "C" {
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally)
     #[wasm_bindgen(method)]
-    pub fn finally(this: &Promise, cb: &Closure<dyn FnMut()>) -> Promise;
+    pub fn finally_map<T, U: ErasableGeneric, R: Promising<Resolution = U>>(
+        this: &Promise<T>,
+        cb: &Closure<dyn FnMut() -> Result<R, JsValue>>,
+    ) -> Promise<U>;
```
RReverser: This signature is wrong. `finally` can return a promise that will be awaited, but it can't change the return value of the promise, it will stay whatever it was before, aka `Promise<T>`.

crates/js-sys/src/lib.rs
```diff
@@ -6236,46 +8726,101 @@ extern "C" {
...
     #[wasm_bindgen(method)]
-    pub fn catch(this: &Promise, cb: &Closure<dyn FnMut(JsValue)>) -> Promise;
+    pub fn catch<T>(this: &Promise<T>, cb: &Closure<dyn FnMut(JsValue)>) -> Promise<JsValue>;
 
+    /// Same as `catch`, but returning a result to become the new Promise value.
+    #[wasm_bindgen(method, js_name = catch)]
+    pub fn catch_map<T, U: ErasableGeneric, R: Promising<Resolution = U>>(
+        this: &Promise<T>,
+        cb: &Closure<dyn FnMut(T) -> Result<R, JsValue>>,
+    ) -> Promise<U>;
```
RReverser: I think this signature is wrong too. If there was no error at runtime, then promise will stay `Promise<T>` but it's typed as if `promise.catch()` will always change it.

I think this binding shouldn't take `U`, signature of catch result should be `R: Promising<Resolution = T>`.

RReverser:
Looked a bit at the bindings in js-sys too, found a few issues. There might be more of similar ones, but Github is really struggling with the diff there so only did a cursory look.

guybedford:
> why does this function return uninhabited type in the signature if it's meant to compile and actually return ()?

The uninhabited type is in this type system being used as Void, unlike Rust which uses Unit. The benefit of an uninhabitable type being the ability to statically vary on its existence or not. The reason for that being that erasable JsValue cannot be type erased into the unit return.

crates/js-sys/src/lib.rs
guybedford: I'm very much open to future improvements with this work!

crates/js-sys/src/lib.rs
guybedford: Sure, will do thanks.

crates/js-sys/src/lib.rs
guybedford: Good point, while it's less flexible it is maybe better for inference, will change.

crates/js-sys/src/lib.rs
guybedford: I was sure you could. Good "catch", thanks will fix.

crates/js-sys/src/lib.rs
guybedford: Right, the Promise is actually `Promise<T | U>` and we don't currently have unions without the non-discriminanted enum type (which I'm still hopeful we might consider).

So in this case, it is better to type for the success case agreed.

crates/js-sys/src/lib.rs
guybedford: I actually went ahead and updated `Object.create()` to be typed, since if you pass a object<JsValue> to create, getting that back should be backwards compatible.

Then I've added create_with_properties that does the T, U case that takes the properties arg and returns a new U object.

guybedford: Actually `create_with_properties` is very unwieldy so will leave this out and just stick with `create<T> -> Object<T>`.

crates/js-sys/src/lib.rs
```diff
@@ -301,150 +304,322 @@ extern "C" {
...
+    ) -> Array<T>;
+
+    /// The `filter()` method creates a new array with all elements that pass the
+    /// test implemented by the provided function. _(Fallible variation)_
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)
+    #[wasm_bindgen(method, js_name = filter, catch)]
+    pub fn try_filter<T>(
+        this: &Array<T>,
+        predicate: &mut dyn FnMut(T, u32) -> Result<bool, JsValue>,
```
RReverser: Here and elsewhere, I noticed that try_ counterparts don't take the Array argument - is that intentional because we already have that reference?

crates/js-sys/src/lib.rs
```diff
@@ -2346,44 +4163,59 @@ extern "C" {
...
+    /// ```
+    ///
+    /// **Note:** Consider using [`Map::try_for_each`] if the callback might throw an error.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach)
+    #[wasm_bindgen(method, js_name = forEach, catch)]
+    pub fn try_for_each<K, V>(
+        this: &Map<K, V>,
+        callback: &mut dyn FnMut(V, K) -> Result<(), JsValue>,
+    ) -> Result<(), JsValue>;
```
RReverser, line 4945: Should we maybe change all of these callbacks to take `Result<..., JsError>` instead?

Our `JsError` wrapper plays better with the rest of Rust ecosystem because it has the necessary `From<...any error type...>` implementation that enables the common `?` operator like

```rust
map.try_for_each(&mut |item, key| {
	let do_smth_1 = fallible_rust_op(item)?;
	let do_smth_2 = fallible_rust_op(item)?;
	...
	Ok(())
});
```

something that is not possible directly with `JsValue` without boilerplate.

Technically in JS you can throw with primitive values as well, but there is no good reason to encourage that in usual Rust bindings?

crates/js-sys/src/lib.rs
```diff
@@ -301,150 +304,322 @@ extern "C" {
...
     ///
     /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)
     #[wasm_bindgen(method)]
-    pub fn flat(this: &Array, depth: i32) -> Array;
+    pub fn flat<T>(this: &Array<T>, depth: i32) -> Array<JsValue>;
 
     /// The `flatMap()` method first maps each element using a mapping function, then flattens
     /// the result into a new array.
     ///
+    /// **Note:** Consider using [`Array::try_Flat_map`] for safer fallible handling.
```
RReverser:
```suggestion
    /// **Note:** Consider using [`Array::try_flat_map`] for safer fallible handling.
```

RReverser:
From quick look through in local checkout, I personally don't see any other typing issues in `js-sys/src/lib.rs` aside from suggestions mentioned above. The only thing I'd still like to discuss and put a lid on is the never/void/whatever type stuff when you have the time :)

crates/js-sys/src/lib.rs
guybedford: Thanks

crates/js-sys/src/lib.rs
guybedford: Yes, it seemed to me like this was redundant since the self reference is guaranteed to provide that.

But perhaps there are closure usage patterns where binding to that reference introduces friction and it is useful?

I tried to come up with a case where you'd really need this argument without being able to rely on a closure reference but couldn't so went with the removal, but if you can think of any case at all where this would be useful to retain we can reconsider.

crates/js-sys/src/lib.rs
guybedford: A refactoring back towards JsError might well be beneficial, I can do some investigation on this further tomorrow.

crates/js-sys/src/lib.rs
RReverser: Nah you are probably right, just wanted to double check this was not accidental omission.

crates/js-sys/src/lib.rs
guybedford: `JsError` is now used in all these closure positions.

mohe2015:
```diff
diff --git a/crates/webidl/src/main.rs b/crates/webidl/src/main.rs
index 81fa1d44f..94db6671c 100644
--- a/crates/webidl/src/main.rs
+++ b/crates/webidl/src/main.rs
@@ -30,7 +30,7 @@ fn main() -> Result<()> {
         &opt.output_dir,
         wasm_bindgen_webidl::Options {
             features,
-            generics_compat: true,
+            generics_compat: false,
         },
     )?;
```
```
cd crates/web-sys
cargo run --release --package wasm-bindgen-webidl -- webidls src/features ./Cargo.toml
thread 'main' (417819) panicked at crates/webidl/src/first_pass.rs:867:18:
called `Result::unwrap()` on an `Err` value: CannotConvert
```
I would've expected this to work but maybe I'm doing something wrong?

crates/macro-support/src/codegen.rs
```diff
@@ -1543,19 +1874,199 @@ struct DescribeImport<'a> {
...
+                let mut i = 0;
+                while i < fn_bounds.len() {
+                    if generics::generics_predicate_uses(&fn_bounds[i], &class_generic_params_refs)
+                        && !generics::generics_predicate_uses(&fn_bounds[i], &fn_generic_params)
+                    {
+                        class_bounds.push(fn_bounds.remove(i));
+                    } else {
+                        i += 1;
+                    }
+                }
```
RReverser: If I understand correctly what this is doing, [`Vec::extract_if`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extract_if) might be a cleaner and simpler approach - it performs "remove if..." operation and returns an iterator so you could just do `class_bounds.extend(fn_bounds.extract_if(...))`.

RReverser: Or I guess you could just add this condition to `filter` where you construct `fn_generic_params` above?

crates/macro-support/src/codegen.rs
```diff
@@ -1543,19 +1874,199 @@ struct DescribeImport<'a> {
...
+                        &fn_generic_params,
+                        class_generic_params_set,
+                    );
+
+                    // Also find lifetimes used in class generic expressions
+                    let used_lifetimes = generics::used_lifetimes_in_type(ty, &all_lifetime_params);
+                    class_lifetime_params_set.extend(used_lifetimes);
+                }
+
+                class_generic_params = class_generic_params_set.into_iter().collect();
```
RReverser: Why not just keep the set instead of Vec in the returned collection? Then you don't need the extra conversion, variable, etc.

crates/macro-support/src/codegen.rs
```diff
@@ -1543,19 +1874,199 @@ struct DescribeImport<'a> {
...
+                    });
+                    where_predicates.push(Cow::Owned(predicate));
+                }
+            }
+        }
+
+        let mut class_bounds = Vec::new();
+        let mut fn_bounds = generics::generic_bounds(&self.generics);
+        let mut class_generic_params_set = BTreeSet::new();
+        let mut class_lifetime_params_set = BTreeSet::new();
```
RReverser: I don't know if this causes any issues at the moment, but I'm a bit concerned about using BTreeSet. I can imagine it's easy to break codegen in a subtle way if we rely on the iteration order of the collected generics, and the user-defined order (e.g. `<B, A>`) is not the sorted one we get from the BTreeSet (`<A, B>`).

Switching to IndexMap/IndexSet might be a bit safer and future-proof.

crates/macro-support/src/codegen.rs
guybedford: `extract_if` would need an MSRV 1.88 it seems. We can't move to `fn_generic_params` above because this info is only available from the class params determined later on after fn_generic_params.

crates/macro-support/src/codegen.rs
guybedford: We have two sets here - the original generic params, and the hoisted generic params (what is populated into class_generic_params_set). It's maybe bad naming - the name should really read `impl_generic_params`, I just call it class, because we only do this hoisting when we know the class type (which is only true for the constructor and methods).

Hoisting is a reachability process, hoisted generic params, whose bounds capture other generic params cause those generic params to be hoisted in turn. It is thus unordered by default - the order becomes the graph order of the reachability of the hoisting process. The final hoist list could be re-ordered based on the original ordering, but that would have to be a new operation in the algorithm, and entirely regardless of the data structure.

The order in this case doesn't matter though, because the order is just the impl block ordering `impl<...>`, it doesn't affect the order of the class generics themselves which is maintained in a Vec.

crates/macro-support/src/codegen.rs
guybedford: This works fine sure, added. If we want to introduce ordering let's pick that up in the next thread below.

PR #4879, opened 2025-12-21
refactor: Consolidate JavaScript binding code generation

magic-akari:
### Description

This PR refactors wasm-bindgen CLI’s JavaScript codegen pipeline to make outputs more consistent and easier to reason about.

- Centralize JS output assembly in a single finalize phase (exports/imports/wasm loading).
- Make `--target experimental-nodejs-module` emit one JS entrypoint (no separate `_bg.js`).
- Ensure Node (CJS/ESM) and bundler entrypoints only expose public exports (no internal import shims).
- Add `/* @ts-self-types="./<name>.d.ts" */` to JS entrypoints for JSR/Deno resolution.
- Refresh reference test fixtures.

### Checklist


guybedford:
Thanks for the PR. I'm out this week but will review when I'm back next week.

crates/cli/tests/reference/add.bg.js
```diff
@@ -31,4 +30,14 @@ export function __wbindgen_init_externref_table() {
     table.set(offset + 1, null);
     table.set(offset + 2, true);
     table.set(offset + 3, false);
-};
+}
+//#endregion
+
+
```
guybedford: Let's make sure to avoid double spacing in output.

crates/cli-support/src/js/mod.rs
```diff
@@ -328,174 +325,28 @@ impl<'a> Context<'a> {
...
-            .module
-            .imports
-            .iter()
-            .map(|import| &import.module)
-            .filter(|module| module.as_str() != self_module_name);
-        for (i, module) in import_modules.enumerate() {
-            imports.push_str(&format!("import * as import{i} from '{module}'\n"));
-            wasm_import_object.push_str(&format!("  '{module}': import{i},\n"))
+        if !self.js_imports.is_empty() {
+            self.globals.push_str("//#region js imports\n");
```
guybedford: While I appreciate these region comments are useful to understand the output and for debugging it, we shouldn't leave these in for user-generated bindings.

crates/cli/tests/reference/anyref-import-catch.bg.js
```diff
@@ -1,8 +1,32 @@
...
+}
+//#endregion
+
+//#region wasm imports
+export function __wbg___wbindgen_throw_dd24417ed36fc46e(arg0, arg1) {
+    throw new Error(getStringFromWasm0(arg0, arg1));
+}
+export function __wbg_foo_9ceaa09e9e1e3047() { return handleError(function () {
+    foo();
+}, arguments) }
```
guybedford: Removing the trailing semicolon here is correct, but formally speaking we should have always had a semicolon after the return statement.

crates/cli-support/src/js/mod.rs
```diff
@@ -514,227 +365,198 @@ export {{ wasm as __wasm }};
...
-        let (init_js, init_ts) = init;
+        Ok((self.globals.to_owned(), ts, start))
+    }
 
-        ts.push_str(&init_ts);
+    fn generate_esm_cjs_imports(&mut self, module_name: &str, has_memory: bool) -> String {
+        let mut imports = String::new();
+        let init_memory_arg = if has_memory { "memory" } else { "" };
+        let mut fn_def = format!(
+            "function __wbg_get_imports({init_memory_arg}) {{
```
guybedford, line 493: What's the benefit of making this a function over it being statically defined? Are there cases where we want this multi-import-instancing functionality you have in mind? Or is it for the performance / V8 lazy parsing benefits?

Overall I like the code simplification, just trying to understand the need to wrap here.

crates/cli/tests/reference/import-target-nodejs.js
```diff
@@ -48,66 +115,14 @@ function decodeText(ptr, len) {
     return cachedTextDecoder.decode(getUint8ArrayMemory0().subarray(ptr, ptr + len));
 }
 
-function exported() {
-    const ret = wasm.exported();
-    if (ret[1]) {
-        throw takeFromExternrefTable0(ret[0]);
-    }
-}
-exports.exported = exported;
```
guybedford, line 57: This seems to have gone missing?

crates/cli-support/src/js/mod.rs
```diff
@@ -879,94 +701,89 @@ wasm = wasmInstance.exports;
...
+                    start.push_str(last);
+                } else {
+                    start.push_str(&format!(r#""{last}""#));
+                }
+            }
+            start.push('\n');
+            start.push_str(r#"} from "./"#);
+            start.push_str(module_name);
+            start.push_str(r#"_bg.js";"#);
+            start.push('\n');
```
guybedford, line 780: Surely there should be a CommonJS path here? See also the missing `exported` in the reference CJS test.

guybedford (commented):
Thanks for doing this work, it's a huge help. I finally had some time to do an initial review. Any other aspects of this PR you feel are worth looking at more closely please do let me know.

crates/cli-support/src/js/mod.rs
magic-akari: Done.

crates/cli-support/src/js/mod.rs
magic-akari: Thanks for the question! 

The primary reason is maintainability, since `__wbg_get_imports()` is already used in the web target, extending it to all targets makes the code easier to maintain.

This also aligns with the discussion in #4881 about potentially allowing manual WASM initialization. There are two motivating cases mentioned there:
- Manual memory passing
- Custom WASM binary access

In a manual initialization scenario, the behavior would be very similar to the web target, so unifying the handling across all targets makes sense from both a design and maintenance perspective.

crates/cli/tests/reference/import-target-nodejs.js
magic-akari: The code isn't missing. It's just been reorganized. 

The `exported()` function was moved from the end of the file to the beginning, under the new `//#region exports` section.
This is part of the restructuring to place public exports at the top for better clarity.

crates/cli/tests/reference/anyref-import-catch.bg.js
magic-akari: Done.

crates/cli/tests/reference/add.bg.js
magic-akari: Done.

crates/cli-support/src/js/mod.rs
magic-akari: This is a new change in this refactor. Previously, the code used `export * from "./{module_name}_bg.js";` which would export everything from the bg file, including internal functions like `__wbg_set_wasm`.

I've updated it to only re-export the user-defined exports from `export_name_list`. This ensures that for bundler targets, we only expose the public API and not internal implementation details.

CommonJS targets have their own separate code path with `exports.xxx = ...` syntax, so we don't need CommonJS-specific handling here since this function only gets called for bundler targets.

crates/cli-support/src/js/mod.rs
guybedford: Can you explain how `__wbg_get_imports()` would be used in your manual instantiation concept? `Object.assign(__wbg_get_imports(), maybeCustomImports)`?

crates/cli/tests/reference/import-target-nodejs.js
guybedford: Ahh, I missed this, thanks.

crates/cli-support/src/js/mod.rs
guybedford: Got it, thanks.

crates/cli-support/src/js/mod.rs
magic-akari: ## Expected User Usage Patterns

### 1. Custom File System API (e.g., VSCode Extensions, sandboxed environments)

```javascript
import { initSync } from './my_module.js';

const wasmBytes = await vscode.workspace.fs.readFile(vscode.Uri.parse(wasmPath));
initSync({ module: wasmBytes });
```

**Internal flow**:
```javascript
function initSync(module, memory) {
    //...
    const imports = __wbg_get_imports(memory);
    if (!(module instanceof WebAssembly.Module)) {
        module = new WebAssembly.Module(module);
    }
    const instance = new WebAssembly.Instance(module, imports);
    //...
}
```

### 2. Web Worker with Shared Memory (Node.js)

```javascript
// Main thread
import * as wasm from './my_module.js';
import { Worker } from 'worker_threads';

const worker = new Worker('./worker.js', {
  workerData: {
    module: wasm.__wbindgen_wasm_module,
    memory: wasm.memory
  }
});

// Worker thread
import * as wasm from './my_module.js';
wasm.initSync({ module: workerData.module, memory: workerData.memory });
```

---

For context, this PR doesn't add `initSync` or manual initialization APIs to Node.js and other targets yet. 
The `__wbg_get_imports()` function pattern (which previously only existed in the web target) is being applied to all targets to enable these use cases in the future.

guybedford: Thanks for clarifying, I think we're on the same page roughly with regards to the overall directions, but it would also be good to continue discussing the future directions to avoid wasted effort.

If you are interested in discussing in-person, you are welcome to join our next meeting on the 13th of Jan, just let me know if you'd like to join (gbedford@cloudflare.com for an invite), otherwise happy to continue discussion here as well.

guybedford (approved):
I think these refactorings are heading in an excellent direction. Will wait a few days before landing for any feedback from other collaborators.

crates/cli-support/src/js/mod.rs
magic-akari: Thank you for the kind invitation, Guy! I really appreciate the offer. I won't be able to join the meeting, but I'd be very happy to continue the discussion here.
Please feel free to share any feedback or thoughts you'd like to discuss, and I'll respond as soon as I can.

PR #4882, opened 2025-12-24
Nodejs initSync

drewcrawford:
# Motivation

  Suppose you want to use workers in Node.js. For example, you are writing threaded Rust code with doctests (which by default run in Node).

 # Problem

  In order to run workers in Node.js, you need a method that can initialize the WASM module on a worker thread with shared memory. The worker thread receives the compiled `WebAssembly.Module` and `WebAssembly.Memory` from the main thread, and needs to instantiate them.

  The `web` target already has `initSync(opts)` for this purpose:

  ```javascript
  // Worker thread (web target)
  wasm_bindgen.initSync({ module, memory });
```

  However, the Node.js CommonJS target (--target nodejs) lacks this API. Instead, it immediately instantiates the module on require(), with no way to pass in shared memory or a pre-compiled module.

 # Solution

  Implement initSync and __wbg_get_imports for Node.js CommonJS when threads are enabled.

  Generated API:
  - initSync(opts) - Initialize the WASM module with optional { module, memory, thread_stack_size }
  - __wbg_get_imports(memory) - Get the imports object with optional custom memory
  - __wbindgen_wasm_module - The compiled WebAssembly.Module for sharing with workers

  # Backwards compatibility:
  - Main thread auto-initializes on require() (existing behavior preserved)
  - Only generates new API when threads are enabled

  # Usage:
```rust
  // Main thread
  const wasm = require('./pkg/my_module.js');
  const { Worker } = require('worker_threads');

  new Worker('./worker.js', {
    workerData: {
      wasmModule: wasm.__wbindgen_wasm_module,
      memory: wasm.memory
    }
  });

  // Worker thread
  const wasm = require('./pkg/my_module.js');
  wasm.initSync({
    module: workerData.wasmModule,
    memory: workerData.memory
  });
```

  # Applications

  This API is basically assumed by https://github.com/drewcrawford/wasm_safe_thread.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ members = [
   "examples/closures",
   "examples/console_log",
   "examples/nodejs_and_deno",
+  "examples/nodejs-threads",
   "examples/dom",
   "examples/duck-typed-interfaces",
   "examples/explicit-resource-management",
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -545,20 +545,135 @@ export {{ wasm as __wasm }};
             OutputMode::Node { module: false } => {
                 js.push_str(&self.generate_node_imports(module_name));
 
-                for (id, js) in iter_by_import(&self.wasm_import_definitions, self.module) {
+                // Track import names for __wbg_get_imports generation
+                let mut import_names: Vec<(String, bool, String)> = Vec::new(); // (name, is_memory, js_expr)
+
+                for (id, js_expr) in iter_by_import(&self.wasm_import_definitions, self.module) {
+                    let is_memory = matches!(
+                        self.module.imports.get(*id).kind,
+                        walrus::ImportKind::Memory(_)
+                    );
                     let import = self.module.imports.get_mut(*id);
                     import.module = format!("./{module_name}_bg.js");
+                    import_names.push((import.name.clone(), is_memory, js_expr.clone()));
                     footer.push_str("\nexports.");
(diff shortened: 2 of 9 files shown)


magic-akari (commented):
My personal take on this:

1. We should aim for consistency across different export targets; at minimum, Node.js (ESM/CJS) targets should behave consistently.

2. The actual requirement here is to add an `init` export to enable manual initialization (replacing the current auto-initialization behavior). The current approach of conditionally changing output based on the `threads` configuration feels unintuitive and confusing. I believe we should use an explicit flag to control exports rather than relying on implicit configuration-based changes.

3. My PR #4879 unifies initialization behavior across different export targets. If we decide to move forward with this change, it would likely be more appropriate and easier to implement after #4879 is merged, as it provides a cleaner foundation to build upon. I'd suggest waiting for that refactor to land before proceeding with this one.

4. I'd welcome further discussion on the details in #4881.

drewcrawford:
Thanks for alerting me to #4879, after looking it over we're in complete agreement that ideally that gets a speedy review and I basically rewrite this after that lands.  Unfortunate timing but that's life.

Points of disagreement:

1.  I have the corresponding ESM patch. It felt prudent to incorporate the lessons from one backend before submitting a second.  If apparently the lesson is "rewrite the whole thing", doing them one by one has already saved good time and seems rather a good pattern.
2. I believe we should continue to use implicit configuration-based changes and not use an explicit flag.  However that architectural issue seems totally separate from adding an API to NodeJS, so I will move my remarks on this to #4881.

PR #4887, opened 2026-01-01
Improves "Small Wasm files" section

qwertyenthusiast:
### Description

This PR changes the "Small Wasm files" page by detailing how to get to the 1kb Wasm file, as that's not documented at all (or is outdated).

### Checklist

- [ ] ~~Verified changelog requirement~~ Not a new release.


qwertyenthusiast:
i've just noticed that wasm-pack also does this by default. should that also be included or nah?

guybedford:
This is the expected usage of wasm bindgen, so I'm not sure it needs an expected callout, but we could definitely more clearly clarify that it does a separate GC to the normal Rust compilation process.

PR #4892, opened 2026-01-05
Fix test case determinism and spacing

guybedford:
This fixes generated output to

1. sort the JS adapter generation based on adapter signature sorting (finally!)
2. remove some unnecessary spacing cases introduced by https://github.com/wasm-bindgen/wasm-bindgen/pull/4879.
3. also sort the actual generated wasm binary adapters with Walrus to fix determinism in the binary itself
4. adds deterministic sorting for `__wbindgen_cast_` based on signature sort index


guybedford:
This will slow down builds a little with the extra sorting, but determinism is important and we can't keep messing around with these issues.

@RReverser @daxpedda do you know anything upstream that might have caused Rust mangling to start being non-deterministic to begin with? codegen units? I wonder if this is something we should carefully address upstream. But at the very least it seems important to fix here asap since this has caused enough CI trouble.

guybedford:
This now resolves all test non-determinism for me.

PR #4893, opened 2026-01-09
Add unwind safe closures

hoodmane:
This makes it so that `Closure::new()`, `Closure::once()`, etc all require an
`UnwindSafe` bound on the closure. Followup to #4790.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


drewcrawford:
The main premise LGTM. 

discussion points:

1.  CI is flagging one regression, missing an IntoWasmClosure impl for AssertUnwindSafe<F>
2. Technically this is an API break.  IMO it's well-justified and well-scoped to rare cases but a migration example would help, as a mention in changelog, as it will break somebody.
3. Performance impact feels justified to me.  Those affected are probably on panic=abort anyway.  You'd be surprised how far you can get on panic=abort, but no reason we shouldn't also support the full-featured panic for workloads that want it...

hoodmane:
It only changes the API when built with `panic=unwind` and it was not possible to build with `panic=unwind` until #4796 was merged. Unfortunately 0.2.106 was released containing #4796 so it is technically an API break. We should call it out in the release notes but I expect almost nobody is affected.

Issue #4896, opened 2026-01-09
Interpreter assertion fails when compiling project in nightly with `opt-level = 1`


MarinPostma:
While compile my project, I have stumbled upon an assertion error in wasm bindgen:

```
 ~/.cargo/bin/wasm-bindgen --target=web --out-dir=/private/tmp/bug-wasm-bindgen/target/wasm-bindgen/debug --out-name=test-debug /private/tmp/bug-wasm-bindgen/target/wasm32-unknown-unknown/debug/test_debug.wasm --no-typescript

thread 'main' (27521951) panicked at /Users/adhoc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasm-bindgen-cli-support-0.2.106/src/interpreter/mod.rs:209:17:
__rustc[f3d9969b214b7ee8]::rust_begin_unwind: Condition failed: `address % width == 0` (4 vs 0)
```

I have managed to assemble a minimal project that consistently reproduces the issue, that can be found in this repo: https://github.com/MarinPostma/bug-wasm-bindgen

rust version: `rustc 1.94.0-nightly (31cd367b9 2026-01-08)`.

What I have found:
- work when compiling on stable
- it seems to consistently work on nightly if i remove:
```
[profile.dev]
opt-level = 1
```
from Cargo.toml
- although it is triggered with `console_log`, it doesn't seem to be the root cause

drewcrawford:
This is a very interesting find, thanks for high-quality reproducer.

At a high level the assert discovers an unaligned read from i64 in the wasm file.  This is "clever", hence the compiler/optimizer involvement.  But it looks like a legal sequence.

A better question is why this usually works.  One reason is this sequence is in some panic handler, so panic machinery is somehow involved in reproducing it.  Another reason is the interpreter tries to ignore these difficult sequences

```rust
//! This module contains a tiny and incomplete Wasm interpreter built on top of
//! `walrus`'s module structure. Each `Interpreter` contains some state
//! about the execution of a Wasm instance. The "incomplete" part here is
//! related to the fact that this is *only* used to execute the various
//! descriptor functions for wasm-bindgen.
```

If I'm right about that, then this would do the trick:

```diff
diff --git a/crates/cli-support/src/interpreter/mod.rs b/crates/cli-support/src/interpreter/mod.rs
index 1e8812159..44c75c955 100644
--- a/crates/cli-support/src/interpreter/mod.rs
[...]

PR #4898, opened 2026-01-11
Stop interpreting descriptor after __wbindgen_describe_cast

MarinPostma:
### Description

fixes a #4896

all credits go to @drewcrawford. 
### Checklist


crates/cli-support/src/interpreter/mod.rs
```diff
@@ -204,6 +206,7 @@ impl Interpreter {
             frame.locals.insert(*arg, *val);
         }
 
+        dbg!(&module.funcs.get(id).name);
```
guybedford: Debug

MarinPostma: yes sorry pushing

MarinPostma: will update changelog too

MarinPostma: done

PR #4910, opened 2026-01-20
refactor!: remove the new Function fallback for the global object getter

IceTDrinker:
### Description
As proposed in this comment: https://github.com/wasm-bindgen/wasm-bindgen/issues/1647#issuecomment-3740163424

I removed the Function fallback for the global object getter. This means very old browsers won't be supported anymore, however someone really needing the fallback can do it themselves by inspecting the object returned by the global call and running equivalent code as what was previously there.

There are other options but nobody seemed to weigh in and give a definitive answer. This solution seems reasonable given it's the simplest and any browser requiring the fallback likely cannot display modern websites anyways.

This removal makes packages using wasm-bindgen who don't instantiate Functions usable in scenarios where CSP ~~is set to `unsafe-eval`~~ is strict, which disallows `unsafe-eval` which seems to be considered good practice for modern web development. Otherwise the fallback forces the generation of JS glue containing a new Function(...) call which violates the CSP ~~`unsafe-eval`~~ strict requirements.

Edit: as a note the coverage does not change, meaning this fallback behavior was not tested previously
Edit2: I always mix up strict/non-strict

### Checklist

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -11,5 +11,10 @@
   },
   "devDependencies": {
     "@types/node": "^24.3.0"
+  },
+  "pnpm": {
+    "overrides": {
+      "memfs": "4.17.0"
+    }
   }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1668,7 +1668,8 @@ where
 
 /// Returns a handle to this Wasm instance's `WebAssembly.Module`.
 /// This is only available when the final Wasm app is built with
-/// `--target no-modules` or `--target web`.
+/// `--target no-modules`, `--target web`, `--target deno` or `--target nodejs`.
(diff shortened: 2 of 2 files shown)


IceTDrinker:
network error for the Taplo CI fail

guybedford (approved):
Thanks for putting this together.

guybedford:
Please add a changelog entry to the CHANGELOG.md file. Then we can land.

IceTDrinker:
Will do, I just find the process a bit weird for the Changelog, as the PR needs to exist to be able to add an entry with the link.

Thanks a lot for reviewing !

IceTDrinker:
rebased on latest main and updated Changelog

PR #4911, opened 2026-01-20
Fixed flags in 'Catching Panics' docs

logan-gatlin:
### Description
The current flags are insufficient for building with `panic=unwind` enabled. Adding these flags caused it to finally work.

### Checklist


guybedford:
Forgoing changelog for doc-only change. Thanks for posting.

PR #4912, opened 2026-01-20
Abort reinit

logan-gatlin:
### Description
Added a flag `--abort-reinit` which automatically re-initializes the WebAssembly module when an unrecoverable panic occurs (panic=abort). This flag automatically enables `--experimental-reset-state-function`.

When the `PanicError` intrinsic exists, exceptions with this class are not considered critical. This means that unwinding panics will not re-initialize the module.

### Checklist

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -807,7 +807,7 @@ fn instruction(
         Instruction::CallExport(_)
         | Instruction::CallAdapter(_)
         | Instruction::DeferFree { .. } => {
-            let should_check_aborted = js.cx.unwind_enabled
+            let mut should_check_aborted = js.cx.unwind_enabled
                 || js.cx.config.abort_reinit
                     && matches!(
                         instr,
@@ -848,7 +848,7 @@ fn instruction(
             }
 
             // Call the function through an export of the underlying module.
-            let call = invoc.invoke(js.cx, &args, &mut js.prelude, log_error)?;
+            let call = invoc.invoke(js.cx, &args, &mut js.prelude, log_error, &mut should_check_aborted)?;
 
             let wrap_try_catch = |call| {
                 let unwrap_wrapped = if js.cx.aux.wrapped_js_tag.is_some() {
@@ -1735,6 +1735,7 @@ impl Invocation {
         args: &[String],
         prelude: &mut String,
         log_error: &mut bool,
+        handle_error: &mut bool,
     ) -> Result<String, Error> {
         match self {
             Invocation::Core { id, .. } => {
@@ -1753,6 +1754,9 @@ impl Invocation {
                 let variadic = cx.aux.imports_with_variadic.contains(id);
                 if cx.import_never_log_error(import) {
                     *log_error = false;
+                }
+                if cx.import_never_handle_error(import) {
+                    *handle_error = false;
                 }
                 cx.invoke_import(import, kind, args, variadic, prelude)
             }
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -1400,8 +1400,15 @@ if (require('worker_threads').isMainThread) {{
                 return ptr;
             }}
             free() {{
-                const ptr = this.__destroy_into_raw();
-                wasm.{}(ptr, 0);
+                try {{
+                    const ptr = this.__destroy_into_raw();
+                    wasm.{}(ptr, 0);
+                }} catch(e) {{
+                    if (e instanceof WebAssembly.Exception && e.is(__wbindgen_wrapped_jstag)) {{
+                        throw e.getArg(__wbindgen_wrapped_jstag, 0);
+                    }}
+                    console.log('ABORT', e);
+                }}
             }}
             ",
             wasm_bindgen_shared::free_function(name),
@@ -3715,6 +3722,20 @@ if (require('worker_threads').isMainThread) {{
             _ => false,
         }
     }
+
+    fn import_never_handle_error(&self, import: &AuxImport) -> bool {
+        match import {
+            // Some intrinsics are intended to exactly throw errors, and in
+            // general we shouldn't have exceptions in our intrinsics to debug,
+            // so skip these.
+            AuxImport::Intrinsic(Intrinsic::Throw | Intrinsic:: Rethrow) => true,
+
+            // Otherwise assume everything else gets a debug log of errors
+            // thrown in debug mode.
+            _ => false,
+        }
+    }
+
 
     /// Attempts to directly hook up the `id` import in the Wasm module with
     /// the `instrs` specified.


crates/cli-support/src/lib.rs
```diff
@@ -348,7 +356,18 @@ impl Bindgen {
...
-        // Check that reset_state is only used with --target module
+        // The `abort_reinit` flag implicitly enables `generate_reset_state`.
+        // This check provides a more helpful error message for this case
+        if self.abort_reinit
+            && !matches!(
+                self.mode,
+                OutputMode::Module | OutputMode::Web | OutputMode::Node { module: false }
+            )
+        {
+            bail!("--experimental-abort-reinit is only supported for --target module, --target web, or --target nodejs")
```
guybedford:
```suggestion
            bail!("--abort-reinit is only supported for --target module, --target web, or --target nodejs")
```

crates/cli/src/wasm_bindgen.rs
```diff
@@ -93,6 +93,11 @@ struct Args {
         help = "Generate __wbg_reset_state function for WASM reinitialization (experimental)"
     )]
     generate_reset_state: bool,
+    #[arg(
+        long = "experimental-abort-reinit",
```
guybedford:
```suggestion
        long = "abort-reinit",
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2649,10 +2649,12 @@ impl<'a> Context<'a> {
 
     fn generate_reset_state(&mut self) -> Result<(), Error> {
         self.global("let __wbg_instance_id = 0;");
+        self.global("let __wbg_aborted = false;");
 
         let mut reset_statements = Vec::new();
 
         reset_statements.push("__wbg_instance_id++;".to_string());
+        reset_statements.push("__wbg_aborted = false;".to_string());
```
guybedford: This seems to cause a double spacing in the output, perhaps let's just write then together:

```suggestion
        reset_statements.push("__wbg_instance_id++;\n__wbg_aborted = false;".to_string());
```

guybedford (approved):
Great work, let's ensure we have test coverage for this in the tests/wasm folder demonstrating:

1. State being reset in the abort case
2. Panics causing reset state with abort but built with panic unwind not resetting state

crates/cli/tests/reference/targets-target-module-abort-reinit-atomics.js
```diff
@@ -0,0 +1,89 @@
...
+        __proto__: null,
+        __wbg___wbindgen_throw_be289d5034ed271b: function(arg0, arg1) {
+            throw new Error(getStringFromWasm0(arg0, arg1));
+        },
+        __wbg_random_e2b253f0e987bd7c: function() {
+            const ret = Math.random();
+            return ret;
+        },
+        __wbindgen_init_externref_table: function() {
+            const table = wasm.__wbindgen_externrefs;
```
hoodmane, line 56: Surely we should make a new table?

hoodmane: Okay I see, we do make a new table.

hoodmane: It's a bit confusing:
```js
    const wasmInstance = new WebAssembly.Instance(wasmModule, __wbg_get_imports());
    wasm = wasmInstance.exports;
```
We reassign wasm after calling this function? So while in this function `wasm` is the old `wasm` instance. I guess `__wbindgen_init_externref_table` must get called by `__wbindgen_start()` so it's okay.

hoodmane: It might be a good idea to switch to importing the table to simplify this logic.

crates/cli-support/src/js/binding.rs
```diff
@@ -159,12 +159,17 @@ impl<'a, 'b> Builder<'a, 'b> {
         if let Some(consumes_self) = self.method {
             let _ = params.next();
             if js.cx.config.generate_reset_state {
+                let abort_check = if js.cx.config.abort_reinit {
+                    "__wbg_aborted === true || "
```
hoodmane:
```suggestion
                    "__wbg_aborted || "
```

crates/cli/tests/reference/targets-target-module-abort-reinit-atomics.js
hoodmane: No action needed in this PR though.

crates/cli-support/src/js/binding.rs
hoodmane: Or is there a good reason to do `=== true`? Looks weird to me.

crates/cli-support/src/js/binding.rs
```diff
@@ -701,6 +706,14 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
     fn assert_not_moved(&mut self, arg: &str) {
         if self.cx.config.generate_reset_state {
             // Under reset state, we need comprehensive validation
+            if self.cx.config.abort_reinit {
+                self.prelude(
+                    "\
+                    if (__wbg_aborted === true) {
```
hoodmane: Again `=== true`?

crates/cli-support/src/js/binding.rs
```diff
@@ -829,17 +846,69 @@ fn instruction(
...
+                            "
+                            {CHECK_ABORTED}
+                            try {{
+                                {call}
+                            }} {catch_exception}
+                            "
+                        ));
+                    } else {
+                        js.finally(&format!("{call};"));
+                    }
```
hoodmane: It'd be better to make a closure for this above the match block rather than duplicating it 3 times.

hoodmane: The definitions of `CHECK_ABORTED` and `catch_exception` could also go in this closure.

logan-gatlin: My original solution was to use a closure, @guybedford suggested I inline it for performance reasons

logan-gatlin: Actually sorry, do you mean make this Rust code into a closure or generate a JS closure for this error handling? I was thinking of the latter

crates/cli-support/src/js/mod.rs
```diff
@@ -2652,7 +2727,21 @@ impl<'a> Context<'a> {
...
+                .aux
+                .set_abort_flag
+                .ok_or_else(|| anyhow!("failed to find `__wbindgen_set_abort_flag` intrinsic"))?;
+            self.export_name_of(store);
+            self.global("let __wbg_aborted = false;");
+            reset_statements.push("wasm.__wbindgen_set_abort_flag(1);".to_string());
+        }
+
+        if self.config.abort_reinit {
+            reset_statements.push("__wbg_instance_id++;\n__wbg_aborted = false;".to_string());
```
hoodmane, line 3018: Do we need to pass `__wbg_instance_id` into Rust ever? If the only point of these is to compare them to check if the instance matches, it seems like it would be simpler to use the `WebAssembly.Instance`s themselves.

crates/cli-support/src/js/binding.rs
guybedford: I actually recommended the repetition approach - Wasm Bindgen in general outputs bindgen functions as expansions in general over using extra functions to enable maximum performance.

crates/cli-support/src/js/mod.rs
guybedford: This comes from the reset state work originally. When handles are invalidated it's nice to be able to GC the instance as soon as possible I think in that model - handles could be kept around longer than the instance would have to be (or potentially even indefinitely) which would risk higher memory usage.

crates/cli-support/src/js/binding.rs
logan-gatlin: Not sure what the convention on this is in JS land. I have heard "always use === to avoid type coercion" but `__wbg_aborted` should only ever be a boolean. Maybe there is a small performance difference?

crates/cli-support/src/js/binding.rs
logan-gatlin: Fixed in [6f4d0d0](https://github.com/wasm-bindgen/wasm-bindgen/pull/4912/commits/6f4d0d0bb3d678c9d7e3d5b412c4c7aedc3d7c4a)

crates/cli-support/src/js/binding.rs
guybedford: Yes, in JS you should use a boolean directly without `===` (since `true === true` is just `true`).

crates/cli-support/src/js/binding.rs
```diff
@@ -795,6 +808,10 @@ fn instruction(
         Instruction::CallExport(_)
         | Instruction::CallAdapter(_)
         | Instruction::DeferFree { .. } => {
+            let should_check_aborted = matches!(
+                instr,
+                Instruction::CallExport(_) | Instruction::DeferFree { .. }
+            );
```
guybedford: This could simplify to:

```suggestion
            let should_check_aborted = js.cx.config.abort_reinit && matches!(
                instr,
                Instruction::CallExport(_) | Instruction::DeferFree { .. }
            );
```

Then you don't need to re-check the abort-reinit option below as well.

crates/cli/src/wasm_bindgen.rs
```diff
@@ -93,6 +93,11 @@ struct Args {
         help = "Generate __wbg_reset_state function for WASM reinitialization (experimental)"
     )]
     generate_reset_state: bool,
+    #[arg(
+        long = "abort-reinit",
+        help = "Re-initialized the WebAssembly module after a panic (only on panic=abort). Also enables --experimental-reset-state-function"
```
guybedford:
```suggestion
        help = "Re-initializes the WebAssembly instance after an unrecoverable error such as an abort."
```

crates/cli/tests/reference/targets-target-module-abort-reinit-atomics.js
```diff
@@ -0,0 +1,89 @@
+/* @ts-self-types="./reference_test.d.ts" */
+
+export function __wbg_reset_state () {
```
guybedford: This actually doesn't need to be an export anymore when using `--abort-reinit` without `--experimental-reset-state-function`.

Instead of "enabling" `--experimental-reset-state-function`, perhaps we can just have `--abort-reinit` enable this function output, but without making it an explicit export?

crates/cli-support/src/js/mod.rs
hoodmane: It seems to me like the point of the `__wbg_instance_id` is to protect the webassembly exports from the original `WebAssembly.Instance` from being used. If they are capable of being used, someone still has access to them. That is enough to keep the Instance alive by itself I would think.

guybedford: No, instance ID is associated with all JS handles for classes and closures, which can be held after the instance.

crates/cli-support/src/js/binding.rs
logan-gatlin: Fixed with [39dbb9f](https://github.com/wasm-bindgen/wasm-bindgen/pull/4912/commits/39dbb9ff1db328b8c8152b478d1d64319ffb664b)

crates/cli/tests/reference/targets-target-module-abort-reinit-atomics.js
logan-gatlin: Fixed with [1401890](https://github.com/wasm-bindgen/wasm-bindgen/pull/4912/commits/14018909eb3c2a50b27298d227b0e0e234859f80)

crates/cli-support/src/js/mod.rs
```diff
@@ -2652,7 +2749,18 @@ impl<'a> Context<'a> {
...
 
-        reset_statements.push("__wbg_instance_id++;".to_string());
+        if self.config.abort_reinit {
+            let store = self
+                .aux
+                .set_abort_flag
+                .ok_or_else(|| anyhow!("failed to find `__wbindgen_set_abort_flag` intrinsic"))?;
+            self.export_name_of(store);
+            self.global("let __wbg_aborted = false;");
+            reset_statements.push("wasm.__wbindgen_set_abort_flag(1);".to_string());
```
guybedford: Unfortunately setting the aborted flag on reinitialization is too late! I believe this must instead be set immediately whenever `__wbg_aborted = true` is set, in order to handle the reentrancy checking which might be before the next entry point call (which would set this here currently).

crates/cli-support/src/js/binding.rs
```diff
@@ -829,17 +847,66 @@ fn instruction(
             // Call the function through an export of the underlying module.
             let call = invoc.invoke(js.cx, &args, &mut js.prelude, log_error)?;
 
+            let wrap_try_catch = |call| {
+                // If the module is compiled with panic=unwind, panics are not critical errors
+                let catch_exception = if js.cx.has_intrinsic("panic_error") {
+                    "\
+                    catch(e) {
+                        if (!(e instanceof PanicError) && !(e instanceof WebAssembly.Exception)) {
```
guybedford: We finally traced this one down, this can now be just:

```suggestion
                        if (!(e instanceof PanicError)) {
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2504,26 +2504,75 @@ impl<'a> Context<'a> {
         });
     }
 
+    fn get_set_aborted(&self, err_name: &str) -> String {
+        if self.has_intrinsic("PanicError") || true {
+            format!("if (!({err_name} instanceof PanicError) && !({err_name} instanceof WebAssembly.Exception))\n  __wbg_aborted = true;\n")
```
guybedford:
```suggestion
            format!("if (!({err_name} instanceof PanicError))\n  __wbg_aborted = true;\n")
```

crates/cli-support/src/js/mod.rs
logan-gatlin: Oh yeah duh lol. That makes sense why panics seemed to happen distant in time to the test that caused them.

src/convert/closures.rs
```diff
@@ -57,6 +57,10 @@ macro_rules! closures {
             fn into_abi(self) -> WasmSlice {
                 unsafe {
                     let (a, b): (usize, usize) = mem::transmute(self);
+                    // TODO: Support panic unwind
+                    // if unwind_safe {
+                    //     b |= 0x80000000;
+                    // }
```
guybedford: Left this in as informative for the closure fix, but we should remove before landing actually.

```suggestion
```

PR #4914, opened 2026-01-21
refactor!: gate functionnality requiring unsafe-eval behind a feature

IceTDrinker:
closes https://github.com/wasm-bindgen/wasm-bindgen/issues/1647

### Description
As discussed here: https://github.com/wasm-bindgen/wasm-bindgen/issues/1647#issuecomment-3775469177

- the feature has the same name and is related to a CSP setting that allows unsafe evaluation
- good modern web practices push people to not use unsafe-eval
- by default wasm-bindgen now makes the choice of not requiring such a CSP setting, making it opt-in and allowing to more easily reason about a rust package being compatible with a strict CSP setting

Edit: this PR is obvisouly a breaking change

### Checklist

### DIFF
--- a/crates/js-sys/Cargo.toml
+++ b/crates/js-sys/Cargo.toml
@@ -21,8 +21,12 @@ doctest = false
 test = false
 
 [features]
-default = ["std"]
+# TODO: remove unsafe-eval for 1.0 release
+default = ["std", "unsafe-eval"]
 std = ["wasm-bindgen/std"]
+# It has been agreed to gate functionnality requiring CSP unsafe-eval behind this feature
+# See https://github.com/wasm-bindgen/wasm-bindgen/issues/1647#issuecomment-3775469177
+unsafe-eval = []
 
 [dependencies]
 once_cell = { version = "1.12", default-features = false }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -261,9 +261,9 @@ jobs:
             env:
               RUSTFLAGS: --cfg=web_sys_unstable_apis
           - name: "js-sys"
-            run: cargo test -p js-sys --target wasm32-unknown-unknown
+            run: cargo test -p js-sys --features=unsafe-eval --target wasm32-unknown-unknown
           - name: "js-sys (unstable)"
-            run: cargo test -p js-sys --target wasm32-unknown-unknown
+            run: cargo test -p js-sys --features=unsafe-eval --target wasm32-unknown-unknown
             env:
               RUSTFLAGS: --cfg=js_sys_unstable_apis
           - name: "wasm-bindgen-webidl"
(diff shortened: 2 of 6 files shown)


IceTDrinker:
looks like a whole lot of things crashed, but clippy could not see it, will try to fix

Edit: my bad, checked when I was on top of the previous PR, needs #4910 first, tried to push without it, obviously a bad idea, I guess I'll wait for the first PR to go through, otherwise the CI would not trigger if I target the branch corresponding to the first PR

crates/js-sys/Cargo.toml
```diff
@@ -23,6 +23,9 @@ test = false
 [features]
 default = ["std"]
 std = ["wasm-bindgen/std"]
+# It has been agreed to gate functionnality requiring CSP unsafe-eval behind this feature
+# See https://github.com/wasm-bindgen/wasm-bindgen/issues/1647#issuecomment-3775469177
+unsafe-eval = []
```
guybedford, line 29: Because we aren't making a breaking change initially, it could be worth turning this feature on by default rather, then noting that it will be turned off by default in a subsequent major.

guybedford (commented):
LGTM!

crates/js-sys/Cargo.toml
Pauan: *Technically* that's still a breaking change, since somebody could be using `default-features = false`

In practice I don't expect very many people are using eval.

guybedford: I don't think it's a technical breaking change only, if it affects what has been a stable function for many years. @RReverser mentioned before using `Function::new()` variants, so it sounds like it would be surprising to turn these off by default, but I'm open to further opinions.

Pauan: I mean that even if the feature is enabled by default, it's still a breaking change, because a crate might be using `default-features = false` which would then disable eval.

IceTDrinker: If you are referring to wasm-bindgen-rayon I can check. But I think it's not affected, it's in the project I used to verify the first PR I submitted and there was no trace of a new Function in the JS glue once the fallback foe the global getter was updated.

I can make the feature enabled by default, not sure if the comment in the Cargo.toml is useful, let me know

IceTDrinker: side question: I explicitely enabled the unsafe-eval feature wherever it was required to have tests compiling, should I keep it explicit if the feature is to be made a default feature ?

edit: reason is, if it's not kept as a default feature later, the person making the change will have to fix it, here at least it's obvious where it's used (even if it were to be made default)

guybedford: Sure, feel free to keep the explicit form in tests when making it the default. Also we should ensure to add a comment noting that it must be removed from the default in the next major so we don't forget that as well (we're planning a 1.0 in the coming months).

IceTDrinker: alright, on it then!

IceTDrinker: done, feature is part of the default set, added a TODO comment above the default features to remove it before 1.0

I would suggest opening an issue linked to your 1.0 milestone, comments in a Cargo.toml would be easy to miss I would think 🙂

Issue #4834, opened 2025-11-28
Remaining non-determinism


guybedford:
The exports ordering solved a major piece of the non-determinism but there still remains non-determinism in the adapter ordering specifically for closure generations.

PR #4926, opened 2026-01-30
Fix duplicate string enum handling

mate-h:
### Description

Fixes duplicate string enum handling in WIT bindgen.

Previously, when the same string enum appeared multiple times (e.g. from different WIT interfaces or components), bindgen would error with "duplicate string enums". This PR changes that behavior so duplicate string enums are **merged** instead of rejected:

- If the same enum name is seen again with identical variant values, it is a no-op.
- If the variant values differ, the variant lists are merged and deduplicated by value.

WIT packages that reference the same string enum from multiple places can now bindgen successfully.

**Example:** `GpuBlendFactor` was reported twice with different variant lists (one with 17 variants including `src1`, `one-minus-src1`, etc., another with 13). Previously this errored with "duplicate string enums"; now the variant lists are merged and deduplicated.

### Checklist

### DIFF
--- a/crates/cli-support/src/wit/mod.rs
+++ b/crates/cli-support/src/wit/mod.rs
@@ -5,7 +5,7 @@ use crate::intrinsic::Intrinsic;
 use crate::transforms::threads::ThreadCount;
 use crate::{decode, wasm_conventions, Bindgen, PLACEHOLDER_MODULE};
 use anyhow::{anyhow, bail, ensure, Error};
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{hash_map::Entry, BTreeSet, HashMap};
 use std::str;
 use walrus::ir::VisitorMut;
 use walrus::{ConstExpr, ElementItems, ExportId, FunctionId, ImportId, MemoryId, Module};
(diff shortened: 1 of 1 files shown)


crates/cli-support/src/wit/mod.rs
```diff
@@ -1518,9 +1531,9 @@ impl<'a> Context<'a> {
 
         // ... then the returned value being translated back
 
-        let inner_ret_output = if let Some(sig_inner_ret) = &signature.inner_ret {
```
drewcrawford, line 1521: Breaks clippy, is this needed / related?

crates/cli-support/src/wit/mod.rs
```diff
@@ -1083,15 +1083,28 @@ impl<'a> Context<'a> {
...
-            .and_modify(|existing| {
-                result = Err(anyhow!("duplicate string enums:\n{existing:?}\n{aux:?}"));
-            })
-            .or_insert(aux);
-        result
+        match self.aux.string_enums.entry(aux.name.clone()) {
+            Entry::Occupied(mut e) => {
+                let existing = e.get_mut();
+                if existing.variant_values != aux.variant_values {
+                    // Merge variant lists
```
drewcrawford, line 1090: So we should definitely merge these.  The question then becomes what is the right merge strategy:

| field               | Here       | My first guess                    | remark                                                                                                   |
|---------------------|------------|-----------------------------------|----------------------------------------------------------------------------------------------------------|
| js_namespace        | first wins | first wins                        | unused field?                                                                                            |
| variant_values      | union      | union                             | Different versions of web-sys may disagree on the number of GpuBlendFactors, hence the bug                              |
| comments            | first wins | ordered append of unique elements | In most cases `a.comments == b.comments` but in edge cases we should behave deterministically, #4834   |
| generate_typescript | first wins | `a \|\| b`                        | If crate `b` wanted typescript, adding crate `a` shouldn't drop support                                  |

The main thing is, 'first' seems to vary by compiler, so we want to avoid introducing new non-determinism in the process of fixing this bug.

guybedford:
I don't recall when we hit these conflicts, can you share an end to end example? Test case would be great of course and needed to land.

PR #4928, opened 2026-02-03
feat(web-sys): add non-standard createSyncAccessHandle mode option

### LINKED ISSUE #4907: Adding non-standard `sync_access_handle_with_options`
I need to use the [`sync_access_handle_with_options`](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createSyncAccessHandle), but I realize it is non-standard, so when I wanted to add it using [these instructions](https://github.com/wasm-bindgen/wasm-bindgen/tree/main/crates/web-sys), I found that there was no IDL for this. Is this something we may still consider adding as un unstable feature? If so, should we come up with the IDL ourselves?

guybedford:
## Summary

This PR adds support for the non-standard `mode` option on `FileSystemFileHandle.createSyncAccessHandle()`, which allows specifying the locking mode (`read-only`, `readwrite`, `readwrite-unsafe`).

Closes #4907

### WebIDL Generator Enhancement

The WebIDL generator was updated to track stability at the **signature level** rather than just the operation level. This allows stable methods to have unstable overloads when a partial interface in the unstable directory adds a new signature with unstable argument types.

- Added `stability` field to `Signature` struct in `first_pass.rs`
- Changed `create_imports()` to use signature stability instead of operation stability in `util.rs`

### New Unstable APIs

- `FileSystemSyncAccessHandleMode` enum with values: `read-only`, `readwrite`, `readwrite-unsafe`
- `FileSystemSyncAccessHandleOptions` dictionary with `mode` field
- `create_sync_access_handle_with_options()` method on `FileSystemFileHandle`

## Result

```rust
// Stable - no cfg gate required
file_handle.create_sync_access_handle();

// Unstable - requires --cfg=web_sys_unstable_apis
let options = FileSystemSyncAccessHandleOptions::new();
options.set_mode(FileSystemSyncAccessHandleMode::ReadOnly);
file_handle.create_sync_access_handle_with_options(&options);
```

## References

- [MDN: createSyncAccessHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemFileHandle/createSyncAccessHandle) - documents the `mode` option as "Non-standard"
- [WHATWG File System spec](https://fs.spec.whatwg.org/#api-filesystemfilehandle-createsyncaccesshandle) - confirms the option is not in the standard


logan-gatlin:
Added tests for unstable interface overloads

PR #4932, opened 2026-02-03
fix: use aliased class identifier for FinalizationRegistry in constructors

### LINKED ISSUE #4885: ReferenceError: ArrayFinalization is not defined
I've got a custom structure called `Array`:

```rust
#[wasm_bindgen]
pub struct Array {

}

#[wasm_bindgen]
impl Array {
  #[wasm_bindgen(constructor)]
  pub fn new() -> Self {
    Self {}
  }
}
```

When initialising a corresponding class in the JavaScript code via `new mylib.Array()`, it fails with the following error:

```
  ReferenceError: ArrayFinalization is not defined
      at new Array2 (/path/to/mylib.js:444:9)
```

After looking into generated js code, I can notice following thing: since the `Array` is already an existing name of the type, my custom type is aliased into `Array2` and wasm-bindgen generates a registry called `Array2Finalization`:

```js
const Array2Finalization = (typeof FinalizationRegistry === 'undefined')
    : new FinalizationRegistry(ptr => wasm.__wbg_array_free(ptr >>> 0, 1));
```

However the code for the custom Array constructor uses an un-aliased registry name (which doesn't exist):

```js
class Array2 {
    constructor(items) {
        var ptr0 = isLikeNone(items) ? 0 : passArrayJsValueToWasm0(items, wasm.__wbindgen_malloc);
        var len0 = WASM_VECTOR_LEN;
        const ret = wasm.array_new(ptr0, len0);
        this.__wbg_ptr = ret >>> 0;
        ArrayFinalization.register(this, this.__wbg_ptr, this);   // ArrayFinalization doesn't exist, Array2Finalization should be used instead
        return this;
    }
}
```

It seems that alias is not used in the constructor. Unfortunately for compatibility reasons, I cannot use other naming.

guybedford:
## Summary

Fixes #4885

When a Rust struct has a name that conflicts with a JS builtin (e.g., `Array`), wasm-bindgen aliases the class name (e.g., `Array2`). However, the constructor was incorrectly referencing the unaliased FinalizationRegistry name (`ArrayFinalization` instead of `Array2Finalization`), causing a `ReferenceError` at runtime.

## Changes

- Added `require_class_identifier()` helper method in `mod.rs` to get the (potentially aliased) JS identifier for a class
- Modified `binding.rs` to use the aliased identifier when generating the `FinalizationRegistry.register()` call in constructors
- Added test case `reserved-class-name.rs` that reproduces the issue by importing JS's `Array` and defining a Rust struct also named `Array`
- Updated `default-class.bg.js` reference (this was also affected - `_default` class was using `defaultFinalization` instead of `_defaultFinalization`)

## Testing

- All 43 reference tests pass
- All 31 cli-support tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@
 
 ### Fixed
 
+* Fixed `ReferenceError` when using Rust struct names that conflict with JS builtins (e.g., `Array`).
+  The constructor now correctly uses the aliased `FinalizationRegistry` identifier.
+  [#4932](https://github.com/wasm-bindgen/wasm-bindgen/pull/4932)
+
 ### Removed
 
 ## [0.2.108](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.107...0.2.108)


logan-gatlin:
LGTM

PR #4933, opened 2026-02-03
fix(examples): fix wasm-audio-worklet using inline_js for worklet module

### LINKED ISSUE #4883: wasm-audio-worklet example broken since 0.2.103
The wasm-audio-worklet example appears to have recently been broken.

Versions:

* 0.2.103: clean checkout and build of wasm-audio-worklet works great
* 0.2.104: fails with error about `TextDecoder` being undefined, possibly introduced by [this PR](https://github.com/wasm-bindgen/wasm-bindgen/pull/4670)?
* 0.2.105, 0.2.106 and main (dda4821): fails with error about `bindgen` being undefined in the `worklet.js` code

Possibly related, I noticed that 0.2.103 and 0.2.104 included `worklet.js` via:

```
    let mod_url = dependent_module!("worklet.js")?;
```

But in 0.2.105 `dependent_module.rs` was removed and this changed to:

```
    let mod_url = wasm_bindgen::link_to!(module = "/src/worklet.js");
```

But that `link_to` macro doesn't seem to have the same magic.

Sorry, I would do more to try figuring out the root cause, but I'm pretty new both to WASM and Rust.

In case it may be relevant, I'm using Chrome on Mac for testing: `Version 143.0.7499.170 (Official Build) (arm64)`

guybedford:
## Summary

Fixes the wasm-audio-worklet example which has been broken since `dependent_module!` was replaced with `link_to!`.

Fixes #4883

## Problem

The `worklet.js` file references `bindgen` (the wasm-bindgen exports):

```js
bindgen.initSync({ module, memory });
this.processor = bindgen.WasmAudioProcessor.unpack(handle);
```

But there was no way to import `bindgen` properly:
- `link_to!` returns a blob/data URL when not using `--split-linked-modules`
- Blob/data URLs don't support relative imports
- Even with split modules, the AudioWorklet context has different module resolution

## Solution

Use `#[wasm_bindgen(inline_js)]` to create a function that:

1. Computes the absolute URL of the main wasm-bindgen module using `new URL('../../wasm_audio_worklet.js', import.meta.url)`
2. Creates a blob URL containing worklet code that imports from that absolute URL
3. Returns the blob URL for use with `AudioWorklet.addModule()`

```rust
#[wasm_bindgen(inline_js = "
export function createWorkletModuleUrl() {
    const bindgenUrl = new URL('../../wasm_audio_worklet.js', import.meta.url).href;
    const code = `import * as bindgen from '${bindgenUrl}'; ...`;
    return URL.createObjectURL(new Blob([code], { type: 'text/javascript' }));
}
")]
```

This approach works because the inline JS module's `import.meta.url` gives us a hierarchical URL that we can use to resolve relative paths, and the blob URL contains an absolute import.

## Playwright Test Fix

AudioWorklet runs in a separate thread (AudioWorkletGlobalScope) which bypasses Playwright's `context.route()` interception. The existing test used route-based mocking to serve files, but AudioWorklet requests weren't being intercepted.

Fixed by adding a real HTTP server for examples that need it:
- Added `createHttpServer()` helper that serves files with proper CORS headers
- Examples in `EXAMPLES_NEEDING_SERVER` list use real HTTP server instead of route mocking
- Other examples continue using the lightweight route-based approach


logan-gatlin:
This might work when compiling the example independently, but breaks with the way example tests run currently. I'm not quite sure why the current script does not generate a snippets folder (or if it does, where?). Where you ever able to get this to work running `pnpm serve` in the examples directory?

PR #4938, opened 2026-02-03
fix: extra "env" "__cpp_exception" import is inlined if detected

logan-gatlin:
### Description
See [#4931](https://github.com/wasm-bindgen/wasm-bindgen/pull/4931).
If this tag import is found in the binary, inline it. I also export it in case JS wants access to it later.

### Checklist

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -369,12 +369,30 @@ impl Bindgen {
         if !self.keep_lld_exports {
             unexported_unused_lld_things(&mut module);
         }
+        // Quick fix for https://github.com/wasm-bindgen/wasm-bindgen/pull/4931
+        // which is likely a compiler bug
+        {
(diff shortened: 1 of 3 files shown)


guybedford:
Thank you so much for tracking this one down, that is a huge help to get CI running again.

guybedford:
(Will follow-up with the more permanent fix in due course)

crates/cli-support/src/lib.rs
```diff
@@ -369,12 +369,30 @@ impl Bindgen {
...
 
-        // We're making quite a few changes, list ourselves as a producer.
-        module
-            .producers
-            .add_processed_by("wasm-bindgen", &wasm_bindgen_shared::version());
-
+            // We're making quite a few changes, list ourselves as a producer.
+            module
+                .producers
+                .add_processed_by("wasm-bindgen", &wasm_bindgen_shared::version());
```
hoodmane, line 394: Doesn't really look like it belongs in this block.

PR #4939, opened 2026-02-05
feat: ScopedClosure with panic=unwind support

guybedford:
### Description
This extends closures to support lifetimes via a `ScopedClosure` type that takes into account the lifetime of the closure. Then we can now use Closures in place of immediate `dyn fn() -> ()` arguments in Wasm Bindgen, while retaining an unwind safe trait for unwind safe path immediate functions.

`Closure<T>` and `StaticClosure<T>` are now both aliases for `ScopedClosure<'static, T>`. In future we could make lifetimes associated with closures the default.

So the recommendation here is to then deprecate immediate functions in Wasm Bindgen, and migrate js-sys to always using `ScopedClosure` arguments intead of `dyn FnMut` etc arguments. Then the user of the API can decide if they want to pass a `ScopedClosure` or a full `StaticClosure` (`Closure` for now) to which can still be reffed as a ref closure itself.

Because our own test runner uses an immediate callback to run the tests, unwind errors were being treated as aborts, not recoverable panics. With this change, the test runner now has recoverable Closure immediate which fully supports unwinds for failed tests, making it possible to integrate hard aborts with `--abort-reinit` separately without breaking the entire test runner. This gap was discovered while reviewing failures on https://github.com/wasm-bindgen/wasm-bindgen/pull/4912.

Work here was done by @logan-gatlin @hoodmane.

### Checklist

### DIFF
--- a/src/closure.rs
+++ b/src/closure.rs
@@ -402,10 +402,16 @@ fn _assert_compiles<T>(pin: core::pin::Pin<&mut ScopedClosure<'static, T>>) {
 impl<T: ?Sized> Drop for ScopedClosure<'_, T> {
     fn drop(&mut self) {
         // Invalidate the closure on the JS side.
-        // For owned closures, this decreases the refcount and frees the Rust data
-        // when the count reaches zero.
-        // For borrowed closures, this sets state.a = state.b = 0 to prevent
-        // any further calls to the closure.
+        //
+        // The JS bindings distinguish owned vs borrowed closures via the `dtor_idx`
+        // encoded in `WasmDescribe`: owned closures pass a non-zero destructor
+        // function pointer, borrowed closures pass `0`.
+        //
+        // For owned closures (`Closure::new`/`ScopedClosure::own`), this decreases
+        // the refcount and frees the Rust heap data when the count reaches zero.
+        //
+        // For borrowed closures (`ScopedClosure::borrow`/`borrow_mut`), this sets
+        // state.a = state.b = 0 to prevent any further calls to the closure.
         self.js.unchecked_ref::<JsClosure>()._wbg_cb_unref();
     }
 }
--- a/tests/wasm/closures.rs
+++ b/tests/wasm/closures.rs
@@ -170,8 +170,8 @@ fn long_lived() {
 #[wasm_bindgen_test]
 fn many_arity() {
     many_arity_call1(&Closure::new(|| {}));
-    many_arity_call2(&Closure::new(|a| assert_eq!(a, 1)));
-    many_arity_call3(&Closure::new(|a, b| assert_eq!((a, b), (1, 2))));
+    many_arity_call2(&ScopedClosure::new(|a| assert_eq!(a, 1)));
+    many_arity_call3(&StaticClosure::new(|a, b| assert_eq!((a, b), (1, 2))));
     many_arity_call4(&Closure::new(|a, b, c| assert_eq!((a, b, c), (1, 2, 3))));
     many_arity_call5(&Closure::new(|a, b, c, d| {
         assert_eq!((a, b, c, d), (1, 2, 3, 4))


crates/cli-support/src/js/mod.rs
```diff
@@ -2728,14 +2748,16 @@ if (require('worker_threads').isMainThread) {{
...
                     }};
-                    real._wbg_cb_unref = () => {{
-                        if (--state.cnt === 0) {{
-                            state.dtor(state.a, state.b);
-                            state.a = 0;
-                            CLOSURE_DTORS.unregister(state);
-                        }}
-                    }};
-                    CLOSURE_DTORS.register(real, state, state);
+                    if (dtor) {{
```
hoodmane: Are we sure we ever take the false branch on this?

guide/src/examples/closures.md
```diff
@@ -5,9 +5,27 @@
...
+
+- **`Closure::new`** — For long-lived closures like event handlers or timers.
+  The closure must be `'static` and you must manage its lifetime (store it
+  somewhere or call `forget()`).
+
+- **`Closure::once`** / **`Closure::once_into_js`** — For one-shot callbacks
+  that will only be called once.
+
+All `Closure` variants are unwind safe — panics are caught and converted to
+JavaScript exceptions when built with `panic=unwind`.
```
hoodmane: In particular they also require the function to be unwind safe. Alternatives are using `AssertUnwindSafe` (hopefully it works) or using the aborting variants.

.github/workflows/main.yml
```diff
@@ -111,12 +111,18 @@ jobs:
             run: cargo test --target wasm32-unknown-unknown --features serde-serialize
           - name: enable-interning
             run: cargo test --target wasm32-unknown-unknown --features enable-interning
+          - name: "panic=unwind"
+            run: RUSTFLAGS="-Cpanic=unwind" RUSTDOCFLAGS="-Cpanic=unwind" cargo +nightly test -Zbuild-std=std,panic_unwind --target wasm32-unknown-unknown
```
guybedford: We already have this test actually

```suggestion
```

src/closure.rs
```diff
@@ -327,6 +337,119 @@ where
         Self::_wrap(data.unsize(), true)
     }
 
+    /// Creates a `Closure` from borrowed data. Used internally by `with` and `with_aborting`.
+    ///
+    /// The closure stores a raw pointer to the borrowed data and is only valid
+    /// for as long as that data lives. When dropped, the closure is invalidated
+    /// on the JS side by setting `state.a = state.b = 0`.
+    fn from_borrowed<F>(t: &mut F, unwind_safe: bool) -> Closure<T>
```
hoodmane: This function should probably be marked `unsafe`.

src/closure.rs
```diff
@@ -327,6 +337,119 @@ where
...
+    /// does not store it. For event listeners, timers, or any API where
+    /// JavaScript retains the callback, use [`Closure::new`] instead.
+    ///
+    /// # Choosing a `Closure` API
+    ///
+    /// | Use case | Recommended API |
+    /// |----------|----------------|
+    /// | Immediate/synchronous callbacks | `Closure::with` |
+    /// | Event listeners, timers | [`Closure::new`] + store handle |
+    /// | One-shot callbacks (e.g., Promise) | [`Closure::once`] or [`Closure::once_into_js`] |
```
hoodmane: This table doesn't really belong in the docs for `Closure::with` I don't think?

guide/src/reference/passing-rust-closures-to-js.md
```diff
@@ -28,39 +37,38 @@ use wasm_bindgen::prelude::*;
...
-    let closure = RefClosure::new_mut(&mut func);
+    let closure = ScopedClosure::borrow_mut(&mut func);
     call_with_value(&closure, 21);
 }
 
 assert_eq!(result, 42);
 ```
 
-Benefits of borrowed closures:
+Benefits of scoped closures:
```
hoodmane: This is a bit weird, is there an alternative?

src/closure.rs
```diff
@@ -259,119 +275,172 @@ extern "C" {
...
 
-impl<T: ?Sized> AsRef<OwnClosure<T>> for RefClosure<'_, T> {
-    fn as_ref(&self) -> &OwnClosure<T> {
-        self
+impl<T: ?Sized> Drop for ScopedClosure<'_, T> {
+    fn drop(&mut self) {
+        // Invalidate the closure on the JS side.
+        // For owned closures, this decreases the refcount and frees the Rust data
+        // when the count reaches zero.
+        // For borrowed closures, this sets state.a = state.b = 0 to prevent
```
hoodmane: We don't have owned and borrowed closures anymore? Also I don't see that there is a flag for this, how do the bidnings distinguish between these cases?

src/closure.rs
```diff
@@ -521,6 +594,19 @@ where
         Self::_wrap(Box::new(t).unsize(), true)
     }
 
+    /// Alias for [`new`](Self::new) — creates an owned `'static` closure.
+    ///
+    /// This name is symmetric with [`borrow`](ScopedClosure::borrow) and
+    /// [`borrow_mut`](ScopedClosure::borrow_mut) for borrowed closures.
+    ///
+    /// See [`new`](Self::new) for full documentation.
+    pub fn own<F>(t: F) -> Self
```
hoodmane, line 609: Will someone be able to do `ScopedClosure::Own` or do they have to do `StaticClosure::own` to see this method?

PR #4944, opened 2026-02-09
no-modules mode URL detection logic for WebWorkers

### LINKED ISSUE #4943: URL detection logic in `no-modules` boilerplate does not support WebWorker
Hello,

At the top of the JavaScript generated by `wasm-bindgen` in `no-modules` mode,   I see: 

```
let wasm_bindgen = (function(exports) {
    let script_src;
    if (typeof document !== 'undefined' && document.currentScript !== null) {
        script_src = new URL(document.currentScript.src, location.href).toString();
    }

    ...
```

If I load this into a WebWorker and then call `wasm_bindgen()` I will get an error,  `module_or_path` will never get set because `script_src` is never set, see:

```
        if (module_or_path === undefined && script_src !== undefined) {
            module_or_path = script_src.replace(/\.js$/, "_bg.wasm");
        }
```

I was able to fix this by adding the following `if/else` branch to support WebWorker:

```
let wasm_bindgen = (function(exports) {
    let script_src;
    if (typeof document !== 'undefined' && document.currentScript !== null) {
        script_src = new URL(document.currentScript.src, location.href).toString();
    }
    // webworkers dont have `document`, use `self` instead.
    else if(self && self.location && self.location.href) {
        script_src = self.location.href;
    }
    
    ...
```

ForestJohnson:
### Description
adds the following code to the boilerplate JS generated for `no-modules` mode in order to allow that code to run inside a WebWorker, fixes #4943 

```
    else if(self && self.location && self.location.href) {
        script_src = new URL(self.location.href).toString();
    }
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 
 ## Unreleased
 
+* Added logic to the `no-modules` JavaScript boilerplate in order to support usage within WebWorkers. [#4943](https://github.com/wasm-bindgen/wasm-bindgen/issues/4943)
+
 ### Added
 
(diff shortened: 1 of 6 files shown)


guybedford:
The problem with this approach is WebAssembly loaded via `importScripts` would not be correctly referenced.

I think because of this failure path the fallback in these scenarios usually also involves supporting a global constant lookup or something horrid like that.

PR #4947, opened 2026-02-09
Fix aliased exports

### LINKED ISSUE #4946: Incorrect JS export names when LLVM merges identical functions (opt-level >= 2)
### Summary

When multiple `#[wasm_bindgen]` functions have identical implementations and are compiled with `opt-level >= 2`, LLVM merges them into a single function with multiple exports (ICMF - Identical Code Merging/Folding). The generated JavaScript incorrectly calls the first export name for all wrappers instead of preserving each function's original export name.

**Note:** This does not cause undefined behavior or incorrect results (since the merged functions are identical), but it makes debugging and call tracing significantly harder — stack traces and profilers will show the wrong function names.

### Minimal Reproduction

```rust
  use wasm_bindgen::prelude::*;

  #[wasm_bindgen]
  pub fn get_a() -> u32 { 42 }

  #[wasm_bindgen]
  pub fn get_b() -> u32 { 42 }
```

```toml
# Bug triggers at opt-level >= 2
[profile.release]
opt-level = 3
```

Expected JS Output
```js
  export function get_b() {
      const ret = wasm.get_b();
      return ret >>> 0;
  }
```
Actual JS Output
```js
  export function get_b() {
      const ret = wasm.get_a();  // BUG: should be get_b
      return ret >>> 0;
  }
```

### Root Cause

In [`crates/cli-support/src/js/binding.rs`](https://github.com/rustwasm/wasm-bindgen/blob/main/crates/cli-support/src/js/binding.rs#L1605-L1608), the `Invocation::Core` only stores the `FunctionId`, [discarding the original `ExportId`](https://github.com/rustwasm/wasm-bindgen/blob/main/crates/cli-support/src/js/binding.rs#L1619-L1622). When generating the JS call, [`export_name_of(function_id)`](https://github.com/rustwasm/wasm-bindgen/blob/main/crates/cli-support/src/js/binding.rs#L1652-L1655) returns the first matching export, not the correct one.

### Fix

I have a working fix that preserves the `ExportId` in the `Invocation` enum and uses it when available. 

Note: The existing CLI reference tests build without optimization, so they don't trigger ICMF. Adding an optimized test variant would require new test infrastructure (no existing tests use --release). I can add this if maintainers think it's necessary. I tested the fix using the minimal crate above instead.

This is my first contribution & PR — feedback welcome!

koziolm:
### Description
Fix incorrect JavaScript export names when LLVM merges identical functions at `opt-level >= 2` (ICMF - Identical Code Merging/Folding).

**Problem:** When multiple `#[wasm_bindgen]` functions have identical implementations, LLVM may merge them into a single function with multiple exports pointing to the same index. The JS glue generator was discarding the original `ExportId`, causing all merged functions to incorrectly call the first export name.

**Solution:** Preserve the `ExportId` in the `Invocation` enum and use it directly when generating JS calls.

**Test Plan:** Tested manually with a minimal crate built at `opt-level=3`:
- Before: `get_b()` incorrectly calls `wasm.get_a()`
- After: `get_b()` correctly calls `wasm.get_b()`

All existing CLI tests pass. No automated test added—the bug only triggers at `opt-level >= 2`, and existing reference tests build without optimization. Adding an optimized test variant would require new infrastructure (no existing tests use `--release`). Happy to add one if maintainers prefer.

Fixes #4946

### Checklist


crates/cli-support/src/js/binding.rs
```diff
@@ -1646,8 +1659,11 @@ impl Invocation {
...
-            Invocation::Core { id, .. } => {
-                let name = cx.export_name_of(*id);
+            Invocation::Core { id, export_id, .. } => {
+                let name = match export_id {
+                    Some(eid) => cx.module.exports.get(*eid).name.clone(),
+                    None => cx.export_name_of(*id),
+                };
                 Ok(format!("wasm.{name}({})", args.join(", ")))
             }
             Invocation::Adapter(id) => {
```
guybedford, line 1653: Are there any other `Invocation::Core` matches that would also benefit from this?

guybedford (approved):
It would also be nice to have the new test run with a release build flag, I think that's good coverage to have.

guybedford:
Please let me know when this is ready for final review.

PR #4954, opened 2026-02-12
Fix incorrect Closure adapter deduplication after wasm-ld ICF

### LINKED ISSUE #4953: Closure<dyn Fn(u32) -> u32> breaks Closure<dyn Fn(Array) -> JsValue>
When two `Closure` types have inner invoke functions with the same wasm signature (e.g., `Closure<dyn Fn(u32) -> u32>` and `Closure<dyn Fn(Array) -> JsValue>` which are both `(i32, i32, i32) -> i32` after lowering), the compiler generates byte-identical code for both, and linkers ICF merges them into one function.

This is a problem since `export_table_element()` then returns the same `ExportId` for both, and `export_adapter()`  deduplicates by `ExportId` alone, dropping the second type's signature. I think the second closure then gets the first closure's adapter with wrong argument conversions.

This can be demonstrated by uncommenting the line `let c2 = ...` which includes both closures in the binary and triggers a crash.
```rust
use js_sys::{Array, Function};
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
extern "C" {
    fn log(s: &str);
}

pub fn make_c1() -> JsValue {
    Closure::<dyn Fn(u32) -> u32>::new(|x: u32| x * 2).into_js_value()
}

pub fn make_c2() -> JsValue {
    Closure::<dyn Fn(Array) -> JsValue>::new(|arr: Array| arr.get(0)).into_js_value()
}

#[wasm_bindgen(start)]
pub fn main() {
    console_error_panic_hook::set_once();

    let c1: Function = make_c1().unchecked_into();
    //let c2: Function = make_c2().unchecked_into();

    // Test 1: Fn(u32) -> u32
    let r1 = c1.call1(&JsValue::UNDEFINED, &JsValue::from(21u32)).unwrap();
    log(&format!("r1 = {} (expected 42)", r1.as_f64().unwrap_or(f64::NAN)));
    
    // Test 2: Fn(Array) -> JsValue
    // let arr = Array::of1(&42.into());
    // let r2 = c2.call1(&JsValue::UNDEFINED, &arr).unwrap();
    // log(&format!("r2 = {} (expected 42)", r2.as_f64().unwrap_or(f64::NAN)));
}
```

allsey87:
`export_adapter()` deduplicates adapters by `ExportId` alone, assuming that the same export always needs the same argument conversions. This breaks when wasm-ld's Identical Code Folding merges invoke functions for different Closure types (e.g., `Fn(u32) -> u32` and `Fn(Array) -> JsValue`) into a single function, giving them the same `ExportId`. This causes the second closure to silently get the first closure's adapter, with the wrong argument conversions. Fix by comparing the descriptor signature `(arguments, ret, inner_ret)` when deduplicating. When the same `ExportId` is reused with a different signature, create a new export with a unique name so each adapter gets its own JS function. Fixes #4953.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,10 @@
   errors in applications with deep call stacks or many concurrent async operations.
   [#4951](https://github.com/wasm-bindgen/wasm-bindgen/pull/4951)
 
+* Fixed incorrect `Closure` adapter deduplication when wasm-ld's Identical Code Folding merges
+  invoke functions for different closure types into the same export.
+  [#4953](https://github.com/wasm-bindgen/wasm-bindgen/issues/4953)
+
 * Fixed `ReferenceError` when using Rust struct names that conflict with JS builtins (e.g., `Array`).
   The constructor now correctly uses the aliased `FinalizationRegistry` identifier.
   [#4932](https://github.com/wasm-bindgen/wasm-bindgen/pull/4932)


allsey87:
Looks like the installation of Firefox failed in CI

guybedford:
I'd be curious to know here if https://github.com/wasm-bindgen/wasm-bindgen/pull/4947 solves the same issue or if both approaches are needed here.

allsey87:
@koziolm could you try to reproduce the bug from #4946 with these changes?

Issue #4920, opened 2026-01-24
Logging is slow


drewcrawford:
Opening a generic tracking issue for many underlying reasons that logging is slow.

It's slow:
* incrementally, logging one single line is too slow
* algorithmically, logging one single line gets slower every time you do it
* once you get out to a few MB or so, it is slow enough to appear hung at first glance

Old writeup of some subset of the issues: #4860 
Current benchmark in CI: #4905

PR #4905, opened 2026-01-15
Add a benchmark for logging performance

drewcrawford:
### Description

Adds a benchmark to track logging performance.  No functional changes.

## test plan

* ensure codspeed picks it up

PR #4875, opened 2025-12-19
Fix devtools output in non-headless mode

Spxg:
### Checklist

- [ ] Verified changelog requirement

PR #4845, opened 2025-12-08
Add real-time output streaming in headless browser mode.

drewcrawford:
### Description

Add real-time output streaming in headless browser mode.

#### Motivation

Suppose my code has a bug - a deadlock, an infinite loop, some browser-specific issue. My test hangs. The normal approach is printf debugging: add some web_sys statements, run again, see where it gets to.

In headless mode, when my test hangs, this doesn't work:

```
Waiting for test to finish...
# ... 60 seconds of silence ...
Failed to detect test as having been run. It might have timed out.
```

Output is buffered until completion. If the test never completes, I never see my prints. I don't know which test hung, where it hung, or whether my debug statements even executed.

With this PR, output from the main browser thread streams as it appears:

```
running 3 tests
test foo ... ok
test bar ... ok
test hanging_test ...
```

Now printf debugging works. I can add prints, run, and see where execution stops.

### Checklist

- [x] Verified changelog requirement

PR #4960, opened 2026-02-13
New headless pipeline

drewcrawford:
## Summary

This is a refactor of the wasm-bindgen-test-runner browser headless logging subsystem that rolls up multiple very tightly-integrated improvements.

Notably, this PR makes the runner logging >400x faster on chrome, >10x faster on firefox and almost 5x faster on safari on our single benchmark, and is substantially faster than even those numbers on many of my workloads.  It's much faster than main on every workload and configuration I've tried.

Once we get the details finalized, I will figure out how to break this up into changelog entries.

Related: #4845, #4860, #4920, #4895

## Motivation
Sometimes I need to debug my tests.  Unfortunately for me, interactively debugging my compiled wasm in terms of its underlying rust source (line number, variable value, memory layout) is rather difficult.

A more practical strategy is to log enough trace output that my program's execution flow can be reconstructed in sufficient detail offline.  For example, if I log at every branch point and I dump variables, structs, etc.  Then I can figure out why my code failed.

However, this involves producing rather large log volumes.  Accordingly we need to handle large volumes, correctly, consistently, performantly, and be able to maintain them going forward.

This rolls up a closely-related set of changes from my tree that substantially advance these goals.

## Problem
The current headless runner has many structural issues:

- Logging a line is a multi-millisecond operation, and the more lines you log, the worse it gets (#4920, #4860). It's clearly a runaway polynomial-time function.
- Output collection paths are not aligned across `server.rs`, `headless.rs`, and `index-headless.html`, so fixes to one place break something somewhere else (#4868, #4845, #4875, etc)
    - Related, lack of sufficient test coverage to make incremental smaller changes than this one
- Large outputs cause oversize WebDriver responses which are too slow, right up until they crash completely (#4860)
    - Related, when runs fail/timed out, diagnostics can be incomplete or too large unless logs are scraped incrementally
- Our 1MB benchmark (#4905) is slow, and a 10MB version crashes completely
- We mostly use append semantics (#4852), but we also clear the `#output` div, which can cause test outputs to be lost in some cases
- In non-TTY environments, we emit the `\r` sequence, which pollutes captured logs in CI or automated environments, and complicates unit tests that check for specific known outputs
- On test failures we print only some log outputs and we print them in the wrong order

## Solution
Update the headless pipeline end-to-end, with minimal compatibility changes to other pipelines.

### 1. Incremental streaming and scraping
- `crates/cli/src/wasm_bindgen_test_runner/headless.rs`
  - Switched to incremental DOM scraping via `execute/sync` + `textContent.slice(offset)`.
  - Streams `#output` incrementally while tests run.
  - Captures tail output after completion.
  - added env var `WASM_BINDGEN_TEST_NO_STREAM=1` to use a very simple non-streaming output.  This is useful for diagnosing regressions in the more complex flow.

### 2. Keep runner/server/template contract synchronized

- `crates/cli/src/wasm_bindgen_test_runner/server.rs`
  - Removed clear-on-first-append behavior that could conflict with incremental scrape offsets.
  - Output is append-only once streaming begins; no mid-stream resets.
  - Kept startup output initialization consistent with append-only output assumptions.

### 3.  Merge all log levels on failure

On test failure we expect to print a captured log output to help with debugging.  However, we only print 2 log levels out of the 5, and we capture them separately so they are not intermingled among levels in logging order.

This is inconsistent both with other platforms, and with ourselves in Node for example, and poses several practical problems for performance and maintainability.  I opened #4895 to discuss this change in more detail.  Finding no objection, here the change is.

This PR captures log output in a unified div. In addition to being nicer output to work with.  Simplifying to a single div gives the JS engines fewer codepaths to JIT and the CPU fewer branches to predict, which creates a measureable performance uplift as demonstrated by the results.

### 4. Performance improvements to logging

Logging a line in native environments is generally a <100us operation.  [But in our benchmark it's 16ms](https://github.com/wasm-bindgen/wasm-bindgen/actions/runs/21970238367/job/63469864126), which is **over 160x slower.**  And *that's* just the benchmark that we *can* run.  The ones we can't run are far worse.

Logging is a performance-critical hot-code path, so let's aggressively optimize it (#4920).

In #4860 I discovered four main causes for poor performance:

1.  output divs are visible, causing tons of work to render invisible log output in browsers, including headless browsers
2.  webdriver patterns, leading to performance overhead and 10MB limits
3.  quadratic append, leading to runaway logging performance as sizes grow
4.  pipe backpressure, causing logs to go missing if pipes can't keep up

This PR rolls up all 4, but the implementation behind 1 is surprising and deserves an extended explanation below.

#### Making output divs invisible

At a high level, our output divs function as a buffer. Logging is implemented by inserting into the div; printing to console is implemented by reading out of the div.

Inserting text into a visible div is very expensive.  To simplify, the sources of the expense are

1.  Layout.  We need to consult the font (a computer program unto itself), and kerning rules, right-to-left language support, line wrapping, hyphenation, and so forth, with some very complex text engine like harfbuzz, DirectWrite, CoreText, that has difficult-to-reason-about performance.
2.  Rendering.  We allocate a bitmap texture somewhere and paint its pixels some color, potentially with an anti-aliasing pass or some way to align the text characters to pixel boundaries.  This tends to be GPU-accelerated in interactive cases but in headless/CI cases tends to fall back to a softwrae path, meaning it's also difficult to reason about.

In theory there are at least two ways to do it, that are functionally equivalent for us but may have different performance:

* `display:none` skips both steps and ought to be basically a drop-in free win
* `visibility:hidden` skips only the second step and ought to provide a more moderate perf uplift

However, in benchmarks:

* chrome strongly follows this model, hard preference for`display:none` and can get us all the way down below 2us.
* firefox strongly dislikes visible divs, but has no preference between `display:none` vs `visibility:hidden`, so dealer's choice
* safari for mysterious reasons shows a strong preference for `visibility:hidden`, opposite chrome's preferred strategy

Accordingly, this PR follows the benchmarks to use `display:none` by default and `visibility:hidden` on safari.  

I suspect some deeper explanation may eventually present itself, but in the meantime this strategy works super well.

## Test plan

  - These interrelated changes cause these tests in my internal suite to turn green:
    - `test_no_carriage_return_in_output`
    - `test_default_no_carriage_return_in_output`
    - `test_browser_no_carriage_return_in_output`
    - `test_shared_worker_no_carriage_return_in_output`
    - `test_service_worker_no_carriage_return_in_output`

Accordingly I am upstreaming them along with this PR.

  - Added `bench_console_log_10mb` alongside `bench_console_log_1mb`.  
      - Going forward this is a benchmark, but here it's more of a functional test; main branch cannot pass this benchmark in any length of time.

- `crates/cli/src/wasm_bindgen_test_runner/shell.rs`
  - TTY-gated status/clear to avoid control-character artifacts in non-interactive output.  This is partly what unblocks us to write reliable test coverage for the runner output.

- `crates/test/src/rt/browser.rs`
  - Removed unconditional `#output` clear to preserve append-only streaming semantics.
  - Added a newline separator in `Browser::new()` so pre-existing bootstrap text and test output stay delimited.

- `crates/cli/tests/wasm-bindgen-test-runner/headless_streaming_tests.rs`

#### 1MB (`bench_console_log_1mb`): current vs `wasm-bindgen/main`
| Browser | Current Median | Upstream Median | Speedup (Upstream / Current) |
|---|---:|---:|---:|
| Chrome | `1.3402 µs` | `641.59 µs` | `478.72x` |
| Firefox | `188.34 µs` | `2.0771 ms` | `11.03x` |
| Safari | `178.17 µs` | `824.10 µs` | `4.63x` |

> [!NOTE]
> These numbers are very substantially faster across the board, not only vs main, but also against earlier attempts like #4860

#### 10MB (`bench_console_log_10mb`): current only (`wasm-bindgen/main` cannot support this volume)
| Browser | Current Median | Multiplier vs 1MB (1.0 = O(1)) |
|---|---:|---:|
| Chrome | `1.4518 µs` | `1.08x` |
| Firefox | `312.93 µs` | `1.66x` |
| Safari | `518.64 µs` | `2.91x` |

As you can see, we are not quite O(1), but we are orders of magnitude better.  Chrome, in particular, might be O(1).  So maybe we can pass the buck to the other browsers to find the same optimization as chrome right here, or maybe some other pattern will present itself in the future.

### Github Actions runner

(our CI only benches firefox)

| benchmark | wasm-bindgen/main         | this PR   | speedup vs main | Multiplier vs 1MB (1.0 = O(1)) |
|-----------|---------------------------|-----------|-----------------|-------------------------|
| 1mb       | 16.194 ms                 | 491.25 µs | 33x             | N/A                     |
| 10mb      | not available; cannot run | 1.8081 ms | N/A             | 3.7                     |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,43 @@
 
 ### Added
 
+* Added `bench_console_log_10mb` benchmark alongside the existing 1MB benchmark for the
+  headless test runner. The main branch cannot complete this benchmark at any volume.
+  [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
+
 ### Changed
 
 * Replaced per-closure generic destructors with a single `__wbindgen_destroy_closure`
   export.
   [#5019](https://github.com/wasm-bindgen/wasm-bindgen/pull/5019)
 
+* Refactored the headless browser test runner logging pipeline for dramatically improved
+  performance (>400x faster on Chrome, >10x on Firefox, ~5x on Safari). Switched to
+  incremental DOM scraping with `textContent.slice(offset)`, append-only output semantics,
+  unified log capture across all log levels on failure, and browser-specific invisible-div
+  optimizations (`display:none` for Chrome/Firefox, `visibility:hidden` for Safari).
+  [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
+
+* TTY-gated status/clear output in the test runner shell to avoid `\r` control-character
+  artifacts in non-interactive (CI) environments.
+  [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
+
 ### Fixed
 
 * Fixed `JsOption::new()` to use `undefined` instead of `null`, to be compatible with `Option::None` and JS default parameters.
   [#5023](https://github.com/wasm-bindgen/wasm-bindgen/pull/5023)
 
+* Fixed headless test runner emitting `\r` carriage-return sequences in non-TTY environments,
+  which polluted captured logs in CI and complicated output-matching tests.
+  [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
+
+* Fixed headless test runner printing incomplete and out-of-order log output on test failures
+  by merging all five log levels into a single unified output div.
+  [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
+
+* Fixed large test outputs (10MB+) causing oversized WebDriver responses that were either
+  extremely slow or crashed completely, by switching to incremental streaming output collection.
+  [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
 ### Removed
 
 ## [0.2.114](https://github.com/wasm-bindgen/wasm-bindgen/compare/0.2.113...0.2.114)


guybedford:
Apologies for my delayed review on this - I've been focused entirely on getting the current release with generics and IDL upgrades out.

I will review this next week though.

crates/cli/src/wasm_bindgen_test_runner/headless.rs
```diff
@@ -183,42 +192,82 @@ pub fn run(
     shell.status("Waiting for test to finish...");
     let start = Instant::now();
     let max = Duration::new(test_timeout, 0);
+    let no_stream_scrape = env::var_os("WASM_BINDGEN_TEST_NO_STREAM").is_some();
```
Spxg, line 195: With `WASM_BINDGEN_TEST_NO_STREAM` enabled, running the tests in the browser produces no log output.

```
➜  playground git:(master) ✗ WASM_BINDGEN_TEST_NO_STREAM=1 cargo test --target wasm32-unknown-unknown
   Compiling playground v0.1.0 (/home/uharsn/repo/playground)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running unittests src/lib.rs (target/wasm32-unknown-unknown/debug/deps/playground-e1c94b1316bc46b7.wasm)
Running headless tests in Firefox on `http://127.0.0.1:41895/`
Not found
   Doc-tests playground                           

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

drewcrawford: fixed + test coverage

Spxg (approved):

crates/cli/src/wasm_bindgen_test_runner/headless.rs
Spxg: Do we need to add a changelog for `WASM_BINDGEN_TEST_NO_STREAM`?

guybedford:
@drewcrawford this just needs a CHANGELOG entry now to land.

crates/cli/src/wasm_bindgen_test_runner/headless.rs
Spxg: Is this option really necessary, or could the new implementation cause issues on some platforms?

PR #4962, opened 2026-02-19
feat(web-sys): Add `matrixTransform` method to `DOMPointReadOnly`

Madoshakalaka:
### Description
Add the `matrixTransform()` method from the W3C Geometry Interfaces Module Level 1 specification to `DOMPointReadOnly`, enabling point transformation by a `DOMMatrixInit`.

### Checklist


Madoshakalaka:
We want this because it's the only alternative to the `matrixTransform` method of the deprecated `SVGPoint` class.

guybedford:
I've updated this to add the default and throws attribute.

PR #4963, opened 2026-02-19
Abort reinit wip: Wrap expected errors in custom wasm eh tag

hoodmane:
This way any exception that isn't wrapped in our custom wasm eh tag can be treated as an abort.
Split off from #4912

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -187,7 +187,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
-      WASM_BINDGEN_EXPERIMENTAL_ABORT_REINIT: 1
+      WASM_BINDGEN_ABORT_REINIT: 1
       RUSTFLAGS: -Cpanic=unwind
     steps:
     - uses: actions/checkout@v6
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -788,6 +788,9 @@ fn generate_wasm_catch_wrappers(module: &mut Module, abort_reinit: bool) -> Resu
     log::debug!("Exception handling version: {eh_version:?}");
 
     if eh_version == transforms::ExceptionHandlingVersion::None {
+        if abort_reinit {
+            bail!("abort_reinit requires panic=unwind");
+        }
         return Ok(());
     }
 
--- a/crates/cli-support/src/transforms/catch_handler.rs
+++ b/crates/cli-support/src/transforms/catch_handler.rs
@@ -28,8 +28,9 @@ use super::ExceptionHandlingVersion;
 enum WrapperKind {
     /// A catch wrapper for `#[wasm_bindgen(catch)]` imports.
     CatchWrapper,
-    /// An abort wrapper that rethrows exceptions with a wrapped tag.
-    AbortWrapper { wrapped_js_tag: TagId },
+    /// This rethrows recoverable exceptions with a wrapped tag so that our
+    /// abort machinery will not treat them as fatal errors.
+    NonAbortingWrapper { wrapped_js_tag: TagId },
 }
 
 /// Intrinsics and IDs needed to generate catch wrappers.
@@ -90,7 +91,7 @@ pub fn run(
         let wrapper_kind = if aux.imports_with_catch.contains(adapter_id) {
             WrapperKind::CatchWrapper
         } else if let Some(wrapped_js_tag) = wrapped_js_tag {
-            WrapperKind::AbortWrapper { wrapped_js_tag }
+            WrapperKind::NonAbortingWrapper { wrapped_js_tag }
         } else {
             continue;
         };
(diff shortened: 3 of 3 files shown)


crates/cli-support/src/js/binding.rs
```diff
@@ -824,19 +849,30 @@ fn instruction(
...
-                (false, 0) => js.prelude(&format!("{call};")),
+                (false, 0) => js.prelude(&wrap_try_catch(&call, should_check_aborted)),
                 (false, n) => {
-                    js.prelude(&format!("const ret = {call};"));
+                    js.prelude(&format!(
+                        "\
+                        let ret;
+                        {}",
+                        &wrap_try_catch(&format!("ret = {call};"), should_check_aborted)
+                    ));
```
guybedford: If we do this conditionally on the presence of the try-catch we can avoid the output churn.

crates/cli-support/src/js/binding.rs
```diff
@@ -95,6 +95,25 @@ pub enum TsReference {
     StringEnum(String),
 }
 
+pub fn wrap_try_catch(call: &str, should_check_aborted: bool) -> String {
```
guybedford: Maybe call this `maybe_wrap_try_catch`?

justfile
```diff
@@ -44,6 +44,19 @@ test-wasm-bindgen *ARGS="":
...
+    RUST_BACKTRACE=1 \
+    WASM_BINDGEN_TEST_ONLY_NODE=1 \
+    WASM_BINDGEN_SPLIT_LINKED_MODULES=1 \
+    WASM_BINDGEN_EXPERIMENTAL_ABORT_REINIT=1 \
+    cargo +nightly test \
+        -Zbuild-std=std,panic_unwind \
+        --target wasm32-unknown-unknown \
+        {{ARGS}}
+
+test-wasm-bindgen-unwind-reinit *ARGS="":
```
guybedford, line 58: These seem to be swapped around?

crates/cli-support/src/lib.rs
```diff
@@ -89,6 +90,7 @@ impl Bindgen {
         let externref =
             env::var("WASM_BINDGEN_ANYREF").is_ok() || env::var("WASM_BINDGEN_EXTERNREF").is_ok();
         let multi_value = env::var("WASM_BINDGEN_MULTI_VALUE").is_ok();
+        let abort_reinit = env::var("WASM_BINDGEN_EXPERIMENTAL_ABORT_REINIT").is_ok();
```
guybedford: Fine to just call `ABORT_REINIT`.

.github/workflows/main.yml
```diff
@@ -182,6 +182,24 @@ jobs:
         node-version: '25'
     - run: cargo test --target wasm32-unknown-unknown -Zbuild-std --lib --bins --tests
     - run: cargo test --target wasm32-unknown-unknown -Zbuild-std --test wasm --features std,wasm-bindgen-futures/std
+  test_wasm_bindgen_unwind_legacy_eh_reinit:
+    name: "Run wasm-bindgen crate tests with panic=unwind with legacy eh"
+    runs-on: ubuntu-latest
+    env:
+      WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
+      WASM_BINDGEN_EXPERIMENTAL_ABORT_REINIT: 1
```
guybedford: Seems you forgot to update this one:

```suggestion
      WASM_BINDGEN_ABORT_REINIT: 1
```

crates/cli-support/src/transforms/catch_handler.rs
```diff
@@ -23,11 +23,21 @@ use walrus::{FunctionBuilder, FunctionId, LocalId, Module, RefType, TableId, Tag
 
 use super::ExceptionHandlingVersion;
 
+/// The kind of wrapper being generated.
+#[derive(Clone, Copy)]
+enum WrapperKind {
+    /// A catch wrapper for `#[wasm_bindgen(catch)]` imports.
+    CatchWrapper,
+    /// An abort wrapper that rethrows exceptions with a wrapped tag.
+    AbortWrapper { wrapped_js_tag: TagId },
```
guybedford: Formally this is still performing an Rust stack unwind though right, it's just not a panic unwind and can't be caught via catch_unwind.

Can we please instead note this in the comment here - "foreign exceptions will still unwind the stack and are therefore recoverable, but cannot be caught". And then call this `NonPanicUnwind` or something that makes this super clear as it's very unintuitive behaviour.

crates/cli-support/src/lib.rs
guybedford, line 790: If `abort_reinit` is enabled and there are no exception handling instructions, let's rather just throw an explicit error and make this case entirely invalid now.

This way, we can be very clear about the assumptions of abort reinit.

crates/cli-support/src/transforms/catch_handler.rs
hoodmane: `NonAbortingUnwind` maybe

hoodmane:
Okay I applied the latest review comments.

PR #4965, opened 2026-02-19
fix: web-sys tests broken on Safari

logan-gatlin:
### Description
Safari does not consider an element to be "editable" unless it is attached to the DOM. This leads to the very confusing situation where `editable == true` and `isEditable() == false`. The fix is just to attach the element before calling `isEditable()`.

### Checklist


guybedford:
Skipping changelog for a test only change.

PR #4020, opened 2024-07-28
Update `AudioBufferSourceNode` WebIDL

daxpedda:
Deprecate `AudioBufferSourceNode.onended` and `AudioBufferSourceNode.stop()`.
This required coaxing the generator a little.

Replaces #3978.
Cc @pablosichert.

PR #4970, opened 2026-02-20
Remove incorrect deprecation from AudioBufferSourceNode::stop() and onended

### LINKED ISSUE #4546: AudioBufferSourceNode::stop() is marked as deprecated
As per title. Clippy flags usage of the deprecated method.

Per MDN (and also the `web-sys` source code), [`AudioBufferSourceNode` extends `AudioScheduledSourceNode`](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode), and the latter _does_ have the `stop()` method.

FWIW, it was marked as deprecated in MR #4020.

guybedford:
## Summary

- Remove spurious `#[deprecated]` from `AudioBufferSourceNode::stop()`, `stop_with_when()`, `onended()`, and `set_onended()`
- Use the `rustAudioScheduledSourceNode` mixin include pattern (matching `OscillatorNode` and `ConstantSourceNode`) instead of re-declaring these methods with `[RustDeprecated]`

## Details

`stop()` and `onended` are inherited from `AudioScheduledSourceNode` and are **not** deprecated in the Web API spec. PR #4020 re-declared them directly on `AudioBufferSourceNode` with `[RustDeprecated]`, causing Clippy to flag any usage as deprecated.

The sibling types `OscillatorNode` and `ConstantSourceNode` correctly use:
```
OscillatorNode includes rustAudioScheduledSourceNode;
ConstantSourceNode includes rustAudioScheduledSourceNode;
```

This PR applies the same pattern to `AudioBufferSourceNode`. The `start()` method remains directly on `AudioBufferSourceNode` because it has a different signature (extra `offset` and `duration` parameters).

Fixes #4546


guybedford:
Closing as the correct approach is to just let the deprecation happen, and ref the parent method in the mean time.

PR #4975, opened 2026-02-20
fix: refactor closure abort handlers

guybedford:
### Description
In 0.2.109, the new closures work added `UnwindSafe` trait bound to `once`. When passing a `dyn FnMut()` directly, this works fine as it can just be wrapped with `Closure::once(AssertUnwindSafe(...))` to ensure the trait or the `once_aborting` variant can be used, but it turns out that a common pattern is to pass `Box<dyn FnMut>` directly both for `once` and for `wrap`. The problem here is that `AssertUnwindSafe<Box<dyn FnMut>>` can't work and also `Box<AssertUnwindSafe<dyn FnMut>>` also can't work. So there's no way to "assert unwind safety".

The refactoring here ensures that all closure constructor functions have the variants `Closure::foo()`, `Closure::foo_aborting()` and `Closure::foo_assert_unwind_safe()` this then fully allows switching from the UnwindSafe bound now being applies on `foo()` to use one of the alternatives, given these limitations of `AssertUnwindSafe`.

The same applies to `ImmediateClosure` as well, moving away from the `wrap` prefix to unify on the `new` constructor with the same pattern. New tests are also added for reentrancy of mutable immediate closures, with a fix for the mutable guard.

In addition, to simplify lifetime management, since ImmediateClosure is just a borrow type, they now work based on pass-by-value instead of by ref.

If anyone comes up with a smarter idea, that would very much be welcome too, but for now this implements the full matrix to ensure compatibility paths.

At the same time, some naming unifications are applied to ensure consistency of the model. Since 0.2.109 has only been out for a few hours, we can still get away with these naming changes on these new methods.

### Checklist

### DIFF
--- a/crates/cli/tests/reference/async-number.bg.js
+++ b/crates/cli/tests/reference/async-number.bg.js
@@ -34,10 +34,18 @@ export function __wbg_createTask_b3c71c63b67bb6ef() { return handleError(functio
     const ret = console.createTask(getStringFromWasm0(arg0, arg1));
     return ret;
 }, arguments); }
-export function __wbg_new_typed_ba4e97687832c766(arg0, arg1) {
+export function __wbg_new_typed_41a2f1c19fbf9990(arg0, arg1) {
     try {
         var state0 = {a: arg0, b: arg1};
-        var cb0 = (arg0, arg1) => wasm_bindgen__convert__closures_____invoke__h0000000000000002(state0.a, state0.b, arg0, arg1);
+        var cb0 = (arg0, arg1) => {
+            const a = state0.a;
+            state0.a = 0;
+            try {
+                return wasm_bindgen__convert__closures_____invoke__h0000000000000002(a, state0.b, arg0, arg1);
+            } finally {
+                state0.a = a;
+            }
+        };
         const ret = new Promise(cb0);
         return ret;
     } finally {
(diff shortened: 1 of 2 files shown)


src/convert/closures.rs
```diff
@@ -291,7 +291,8 @@ macro_rules! closures {
                 let rc1 = Rc::new(WasmRefCell::new(None));
                 let rc2 = rc1.clone();
 
-                let closure = Closure::once_wrap(closures!(@closure $FnArgs $($var)* {
+                // TODO: Unwind safety for FnOnce
+                let closure = Closure::once_aborting(closures!(@closure $FnArgs $($var)* {
```
guybedford, line 295: Discovered this during this refactoring - `FnOnce` in its current form doesn't support unwind safety.

src/closure.rs
```diff
@@ -633,13 +680,40 @@ where
...
+    {
+        Closure::_wrap(fn_once.into_fn_mut(), true)
+    }
+
+    // TODO: Update once closures to be generated on construction as once
+    // closures instead of using wrap(). Then we can share the same into_js()
+    // function between all closures, and deprecate this method.
     /// Convert a `FnOnce(A...) -> R` into a JavaScript `Function` object.
     ///
     /// If the JavaScript function is invoked more than once, it will throw an
```
guybedford, line 719: I left `once_into_js` out of all this handling because I'm not sure it's the right design (and doesn't support unwind safety at all currently, since the once wrapper doesn't branch on the once behaviour).

I think if we instead eagerly set the once metadata in the initial closure construction as part of the Closure ABI itself, then we don't need the second wrapper and can remove this approach entirely, while leaning into the existing `into_js()` path without needing a separate one. //cc @RReverser

PR #4977, opened 2026-02-21
fix: allow same js_name in different js_namespace for exports

ZhaoXiangXML:
### Description
Previously, two exported structs (or enums/functions) using the same js_name but different js_namespace values would produce symbol collisions at compile time, because internal wasm symbols were derived from js_name alone.

This fix introduces a 'qualified_name' concept (namespace_jsname when a namespace is present, otherwise just js_name) used for internal wasm symbol generation, while keeping js_name for user-facing JS/TS output.

Key changes:
- Add qualified_name() helper to wasm-bindgen-shared
- Add rust_name to the shared schema so the CLI can key exported_classes by the unique Rust identifier
- Use qualified_name for all wasm symbol generation (free, new, unwrap, field getters/setters) and WasmDescribe encoding
- Add qualified_to_rust_name and qualified_to_js_name maps in the CLI for resolving descriptor names to class keys and TS type names
- Pre-populate js_name on exported classes so identifier generation uses the correct name before generate_struct runs
- Resolve both constructor and descriptor class names through the same mapping for correct constructor path matching

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@
 
 ### Fixed
 
+* Fixed exported structs/enums/functions with the same `js_name` but different
+  `js_namespace` values producing symbol collisions at compile time, by deriving
+  internal wasm symbols from a qualified name that includes the namespace.
+  [#4977](https://github.com/wasm-bindgen/wasm-bindgen/pull/4977)
+
 * Fixed soundness hole in `ScopedClosure`'s `UpcastFrom` that allowed to extend the lifetime after the original `ScopedClosure` was dropped.
   [#5006](https://github.com/wasm-bindgen/wasm-bindgen/pull/5006)
 


ZhaoXiangXML:
I'm not sure what does "Verified changelog requirement" mean, besides that I think this PR is ready for review.

crates/shared/src/lib.rs
```diff
@@ -191,6 +192,26 @@ macro_rules! shared_api {
     }; // end of mac case
 } // end of mac definition
 
+/// Compute a "qualified name" by prepending the namespace (joined with `_`) to the js_name.
```
drewcrawford: So I think this is a good solution.  However, it does rename quite a few internals (`__wbg_counter_free → __wbg_default_counter_free`, `__wbg_point3d_free → __wbg_models_3d_point3d_free`) etc.

In theory they're internal, in practice they get used to work around various things.  IMO this warrants a highlight in the changelog.

crates/cli-support/src/js/mod.rs
```diff
@@ -3159,6 +3200,26 @@ if (require('worker_threads').isMainThread) {{
     pub fn generate(&mut self) -> Result<(), Error> {
         self.prestore_global_import_identifiers()?;
 
+        // Set up qualified_name → rust_name and qualified_name → js_name mappings
+        // before processing adapters, so that WasmDescribe lookups (which use
+        // qualified_name) can resolve to the correct exported class.
+        for s in self.aux.structs.iter() {
+            self.qualified_to_rust_name
```
drewcrawford, line 3211: Seems to overlap with below?

crates/cli-support/src/js/mod.rs
```diff
@@ -4492,12 +4553,23 @@ if (require('worker_threads').isMainThread) {{
...
         class.comments = format_doc_comments(&struct_.comments, None);
         class.is_inspectable = struct_.is_inspectable;
         class.generate_typescript = struct_.generate_typescript;
         class.private = struct_.private;
         class.js_namespace = struct_.js_namespace.as_ref().map(|ns| ns.to_vec());
+        class.js_name = Some(struct_.name.clone());
+        class.qualified_name = Some(struct_.qualified_name.clone());
+        // Build mapping from qualified_name to rust_name so that
+        // WasmDescribe lookups (which use qualified_name) can find the right class.
+        self.qualified_to_rust_name
```
drewcrawford: Seems to overlap with above?

crates/shared/src/lib.rs
```diff
@@ -191,6 +192,26 @@ macro_rules! shared_api {
...
+/// This is used to disambiguate internal wasm symbols when the same js_name
+/// appears in different namespaces.
+pub fn qualified_name(js_namespace: Option<&[impl AsRef<str>]>, js_name: &str) -> String {
+    match js_namespace {
+        Some(ns) if !ns.is_empty() => {
+            let mut name = ns
+                .iter()
+                .map(|s| s.as_ref())
+                .collect::<alloc::vec::Vec<_>>()
+                .join("_");
```
drewcrawford: I think `_` opens the door to collisions, is there a non-identifier pattern we could use here (i.e. `::` or so?)

crates/cli-support/src/js/mod.rs
```diff
@@ -1233,19 +1245,44 @@ if (require('worker_threads').isMainThread) {{
         }
     }
 
+    /// Resolve a class name to the key used in `exported_classes`.
+    /// The name could be a `rust_name` (direct key) or a `qualified_name`
+    /// (from WasmDescribe), which needs to be mapped to the `rust_name`.
+    pub(crate) fn resolve_class_name(&self, name: &str) -> String {
```
drewcrawford: My hunch is this is hot code, mostly allocation, and the allocation is mostly unnecessary

crates/shared/src/lib.rs
ZhaoXiangXML: > I think `_` opens the door to collisions, is there a non-identifier pattern we could use here (i.e. `::` or so?)

This need to be valid identifier in both js and rust, so I've changed it to `__`, which can lower the change.

ZhaoXiangXML:
@drewcrawford I've updated the PR according to your review, please check again.

drewcrawford (approved):

ZhaoXiangXML:
@drewcrawford @guybedford Thank you for your review. I've rebased from main, solved conflicts, and modified `generate_identifier` to avoid one collision. Please kindly check again.

PR #4981, opened 2026-02-21
Experiment with implementing generic wasm_bindgen

rollo-b2c2:
So I was wondering what the difficulty was with implementing generics in wasm-bindgen actually was. It's not too difficult but it's not particularly useful either

As this is just an expirement it only supports structs and getting generic methods to play nice might be a bit of a pain in practice.

The downside to this approach is that you have to declare each generic upfront, kinda hamstringing their usefulness.  I appreciate that anything automatic would be nightmare fuel for the rust compiler as it'd require wasm-bindgen knowing about the instantiations of generic types requiring a monophisation step _after_ linking.

Here's an example API that creates multiple generic instantiations of a structure.

```
use wasm_bindgen::prelude::*;

#[wasm_bindgen(
    generic = (
        new(
            generics(usize, f64), 
            type_alias_name=MyTestStructGen1
        ),
        new(
            generics(f64, f64), 
            type_alias_name=MyTestStructGen2
        ),
    )
)]
pub struct MyTestStruct<T, R> {
    pub the_prop: T,
    pub other_prop: R,
}


type MyTestStructGen1 = MyTestStruct<usize, f64>;
type MyTestStructGen2 = MyTestStruct<f64, f64>;

#[wasm_bindgen]
pub fn example_use() -> MyTestStruct<f64, f64> {
    MyTestStruct { 
        the_prop: 0.1,
        other_prop: 5.5
    }
}
```

Not 100% sure if I'd use such an approach.  Generics aren't _hugely_ useful at the type boundary anyway, my _particular_ usecase is for interop with already existing code where destructing and recreating structures into new type would be a massive pain in already generic laden code.

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -9,9 +9,9 @@ use walrus::Module;
 
 pub(crate) const PLACEHOLDER_MODULE: &str = "__wbindgen_placeholder__";
 
-mod decode;
-mod descriptor;
-mod descriptors;
+pub mod decode;
+pub mod descriptor;
+pub mod descriptors;
 mod externref;
 mod interpreter;
 mod intrinsic;
--- a/crates/macro-support/Cargo.toml
+++ b/crates/macro-support/Cargo.toml
@@ -14,6 +14,7 @@ version = "0.2.112"
 [features]
 extra-traits = ["syn/extra-traits"]
 strict-macro = []
+default = ["extra-traits"]
 
 [dependencies]
 bumpalo = "3.0.0"
--- a/crates/macro-support/src/ast.rs
+++ b/crates/macro-support/src/ast.rs
@@ -425,6 +425,7 @@ pub struct FunctionArgumentData {
 }
 
 /// Information about a Struct being exported
+#[allow(dead_code)]
 #[cfg_attr(feature = "extra-traits", derive(Debug))]
 #[derive(Clone)]
 pub struct Struct {
@@ -446,6 +447,8 @@ pub struct Struct {
     pub js_namespace: Option<Vec<String>>,
     /// Path to wasm_bindgen
     pub wasm_bindgen: Path,
+    // TODO generiate the type lias
+    pub generate_type_alias: Option<Ident>,
 }
 
 /// The field of a struct
--- a/crates/macro-support/src/codegen.rs
+++ b/crates/macro-support/src/codegen.rs
@@ -1,5 +1,6 @@
-use crate::ast;
-use crate::encode;
+use crate::ast::Import;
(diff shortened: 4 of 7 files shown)


guybedford:
Agreed, a concrete generic implementation would require hinting the available type combinations. The other way to do this might be with some compiler magic when generics affect concrete implementations.

rollo-b2c2:
@guybedford I took a bit of a further look, the foreign type implementation is perfectly possible.

I could rewrite the output so that all of the traits implemented on a wasm generic were generic on `T: JsValue`, the only thing that isn't is the WasmDescribe trait, which is annoying since that actually need some metadata.

You could _possibly_ include that metadata by way of some build.rs between dependant packages but it feels fragile to do so.

rollo-b2c2:
https://github.com/wasm-bindgen/wasm-bindgen/issues/5045

Hey @guybedford, I've taken this stuff a bit further, I'm working an internal project that handles generics more completely.  It's internal, but I can try to speak to people at my firm to get it open sourced.  It actually works quite well and I think it's a viable path.

I managed to do it without _meaingfully_ forking wasm-bingden, but for my own sanity I had to make the encoder public.  It'd be really handy if you could consider making that modules public so I didn't need to run a fork to develop this further internally?  Running a fork is something I need to justify to my own CoC.

I've taken two approaches that both work.  One is a tagged vtable implementation that runs a shim in javascript to link itself back.  The weird parts are using a hash of TypeId to generate unique descriptors.  It's not the most optimal thing due to the vtable, but it serves my general needs, JavaScript is just a little slow anyway and what is one bit of indirection in the service of ergonomics?

The other is a little more sketchy and a WIP: the concrete implementation.  I didn't know that Rust symbols aren't really mangled, they're really encoded into stable format.  So I wrote a prototype wasm linker to emit JS code based on this. It looks a little like this


```

struct GenericType<T> where T:  From<JsValue> {
   inner: T
}

generic_link_symbol! {
   GenericType<AbiAny> 
}


#[wasm_bindgen]
pub fn a() -> GenericType<i32> { todo!() }
#[wasm_bindgen]
pub fn b() -> GenericType<usize> { todo!() }

```

Where this emits a special link symbol for the generic type.  `AbiCompat` is basically the "any" type.  Because GenericType is a public symbol (as wasm_bindgen exports) it, I got wasm_bindgen to look into a custom link section and pattern match on link symbol for `GenericType<AbiCompat>`, which contains the metadata needed to know that `GenericType<i32>` are instantiations of of `GenericType<i32>`.

guybedford:
We've since implemented real generics in https://github.com/wasm-bindgen/wasm-bindgen/pull/5230. Do try it out and share feedback further.

PR #4983, opened 2026-02-22
fix(web-sys): Update RTCPeerConnection.addTrack() bindings to match spec

Anillc:
### Description

This PR updates the `RTCPeerConnection::add_track` bindings to match the current WebIDL specification.

https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface-extensions

image

### Checklist

### DIFF
--- a/crates/web-sys/webidls/enabled/RTCPeerConnection.webidl
+++ b/crates/web-sys/webidls/enabled/RTCPeerConnection.webidl
@@ -121,8 +121,7 @@ interface RTCPeerConnection : EventTarget  {
   // indicate which particular streams should be referenced in signaling
 
   RTCRtpSender addTrack(MediaStreamTrack track,
-                        MediaStream stream,
-                        MediaStream... moreStreams);
+                        MediaStream... streams);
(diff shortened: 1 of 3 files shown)


crates/web-sys/webidls/enabled/RTCPeerConnection.webidl
```diff
@@ -121,7 +121,6 @@ interface RTCPeerConnection : EventTarget  {
   // indicate which particular streams should be referenced in signaling
 
   RTCRtpSender addTrack(MediaStreamTrack track,
-                        MediaStream stream,
                         MediaStream... moreStreams);
```
evilpie:
```suggestion
                        MediaStream... streams);
```

guybedford:
What is the issue here of this fix? Is it just a correctness issue? Or are there cases when you want to add with zero arguments? Note this is not backwards compatible so can't land unfortunately. Also when updating variadic methods with backwards compatibility going forward we would like to update them to use the new `&[Foo]` variadic slice form as well.

PR #4986, opened 2026-02-23
revert: remove immediate closures

guybedford:
### Description
`ImmediateClosures` were added in 0.2.109, and while they extend the concept of unwind safety to immediately optimized closure calling conventions, they are quite simply a pain to use.

This removes them while we are still within the release cycle window, based on the fact that we believe there may be a future approach to implement `ScopedClosure::immediate` which could provide the unified closure type while still supporting lazy / immediate closure calling conventions.

Then the `dyn FnMut` and `dyn Fn` argument forms are no longer deprecated here and instead continued to be supported, while being treated as unwind safe by default. The onus is then on the user to verify the unwind safety property for this form specifically, as the caveat of their usage, without an abort safe variant being possible until we have the unified `ScopedClosure::immediate`.

If the `ScopedClosure::immediate` work is successful and we are able to come up with an ergonomic construction for this (say with a macro or otherwise), then we could revisit deprecating these "legacy closures", but for now these are treated as full supported and no longer legacy.

Hoping for a quick turnaround on this one, as a last patch for the current release cycle.

### Checklist


logan-gatlin (approved):
Looks like old UI tests may remain but can fix those later. LGTM

PR #4993, opened 2026-02-24
Reduce unsafe

RReverser:
Just went through the instances of `unsafe` in code and cleaned some up (both new and old) where safe code suffices.

### DIFF
--- a/crates/cli-support/src/intrinsic.rs
+++ b/crates/cli-support/src/intrinsic.rs
@@ -42,7 +42,10 @@ intrinsics! {
         IsFunction = "__wbindgen_is_function",
         IsUndefined = "__wbindgen_is_undefined",
         IsNull = "__wbindgen_is_null",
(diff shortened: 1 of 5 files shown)


crates/cli-support/src/intrinsic.rs
```diff
@@ -42,8 +42,7 @@ intrinsics! {
         IsFunction = "__wbindgen_is_function",
         IsUndefined = "__wbindgen_is_undefined",
         IsNull = "__wbindgen_is_null",
-        ObjectIsNullOrUndefined = "__wbindgen_object_is_null_or_undefined",
-        ObjectIsUndefined = "__wbindgen_object_is_undefined",
+        IsNullOrUndefined = "__wbindgen_is_null_or_undefined",
```
guybedford: I agree the naming is better, but this would imply a schema change right? Better to avoid?

guybedford: I've added the backwards compat for schema here.

PR #4994, opened 2026-02-24
Schema cleanup: rename and remove legacy intrinsics

guybedford:
## Summary

- Renames `__wbindgen_object_is_null_or_undefined` to `__wbindgen_is_null_or_undefined`, moving it from the raw `u32`-based `externs!` block to the safe `&JsValue`-based `wasm_bindgen` extern block.
- Removes `__wbindgen_object_is_undefined` entirely, replacing its only usage in `OptionFromWasmAbi` for `JsValue` with a safe Rust-side `.is_undefined()` call.
- Removes the corresponding manual import mappings in `wit/mod.rs` since the intrinsics system handles matching by function name automatically.

**Note:** This is a schema-breaking change and should be landed as part of a schema version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
   `__wbindgen_object_is_undefined` intrinsic in favor of a safe Rust-side equivalent.
   [#4993](https://github.com/wasm-bindgen/wasm-bindgen/pull/4993)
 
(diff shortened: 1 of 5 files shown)


guybedford:
//cc @RReverser

logan-gatlin (approved):

PR #4998, opened 2026-02-25
Migrate from once_cell to native LazyCell

RReverser:
Since we bumped MSRV, we no longer need an extra dependency.

Note that this only applies to `LazyCell` - `LazyLock` has higher MSRV than we declare, but we only use it in webidl generator and tests, both of which are private usages.


RReverser:
Huh I misread the official docs. I thought I saw that it was added in 1.70, but apparently it was only 1.80.

PR #4999, opened 2026-02-25
Wrap core RefCell instead of reimplementing via `unsafe`

RReverser:
### Description

We can still achieve goals stated in the docs of `WasmRefCell` by simply using non-panicking versions of `RefCell` methods and throwing our own error instead.

This removes couple more usages of `unsafe` and simplifies the implementation, win-win.

### Checklist

- [x] Verified changelog requirement - not required

### DIFF
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -4,6 +4,8 @@ use crate::JsValue;
 #[cfg(all(target_arch = "wasm32", feature = "std", panic = "unwind"))]
 use core::any::Any;
 use core::borrow::{Borrow, BorrowMut};
+#[cfg(target_feature = "atomics")]
+use core::cell::UnsafeCell;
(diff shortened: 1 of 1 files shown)


RReverser:
> Isn't code size the argument though? What effect does this have on final binary sizes?

That's what I meant by

> We can still achieve goals stated in the docs of `WasmRefCell`

the concern about code size, as the comment says, is due to `RefCell::borrow{_mut}` pulling it heavy panic runtime, which in turn pulls formatting and so on.

Whereas `try_` versions of these methods doesn't do any of that - they are almost identical to what we were doing in our own unsafe code, and they still leave error throwing to us.

RReverser:
> (Separately we need to start tracking binary size in CI still)

This is definitely something that would be useful. I think Dan was working on it?

PR #5004, opened 2026-02-26
Add wasm64/memory64 support for `wasm64-unknown-unknown` target

### LINKED ISSUE #4436: official `wasm64-unknown-unknown` support
Given that the [Memory64 proposal](https://github.com/WebAssembly/memory64) has moved into the standardization phase and most major runtime environments already support 64 bit memory addresses either via a feature flag or nightly builds, are there official plans for wasm-bindgen to support Memory64?

### LINKED ISSUE #4499: Create bindings for wasm64-unknown-unknown
### Summary

Is it possible to run wasm-bindgen/walrus or even wasm-pack somehow to compile & create correctly the bindings between rust code compiled to wasm64-unknown-unknown and javascript? If not, would you happen to know whether there is currently an alternative way of creating these bindings between js & rust compiled to wasm64?

### Additional Details

First of all, I just started using wasm with rust since 1-2 weeks ago, forgive my ignorance and whether this is the best place to ask.

I'm working on a research project where we'd like to build our code to wasm64-unknown-unknown (as it will need in most of the cases more than 4GB of memory) and put the binary in a website done with javascript. I managed to successfully compile it to that target using cargo, but I soon realised (as shown also in these two issues I found https://github.com/rustwasm/wasm-pack/issues/1464 , https://github.com/rustwasm/wasm-bindgen/issues/4436 ) that apparently wasm-pack does not support the already merged and approved memory64 proposal ( https://github.com/WebAssembly/memory64 ).

However, it seems that walrus (used by wasm-bindgen if I understood it correctly) does have some memory64 related stuff ( https://github.com/search?q=repo%3Arustwasm%2Fwalrus%20memory64&type=code ). Thus, is it possible to run wasm-bindgen/walrus or even wasm-pack (in a way that I did not find) somehow to compile & create correctly the links between the rust compiled to wasm64 and javascript? If not, would you happen to know whether there is currently an alternative way of creating these bindings between js & rust compiled to wasm64?

Thanks!

audreyt:
## Summary

Add full `wasm64-unknown-unknown` (memory64) support to wasm-bindgen, enabling Rust-to-JS bindings for WebAssembly modules that use 64-bit linear memory addressing (>4 GB).

Closes #4436. Closes #4499.

## Motivation

The [WebAssembly memory64 proposal](https://github.com/WebAssembly/memory64) has reached standardization phase, and Rust nightly ships a `wasm64-unknown-unknown` target. However, wasm-bindgen has been the major blocker for actually using it — the CLI generates JS glue that assumes 32-bit pointers throughout (`>>> 0`, `getUint32`, 4-byte strides), and the runtime library hardcodes `u32` for all pointer/length ABI types. This PR closes that gap.

## Changes

**23 files changed** (+703 −282) across three layers:

### Runtime library (`src/`, `crates/`)

- Widen all `target_arch = "wasm32"` cfg gates to include `wasm64` (~49 occurrences across 11 files)
- `describe.rs` — `usize`/`isize` emit `I64` descriptors on wasm64; pointer-carrying descriptors use the correct width
- `convert/impls.rs`, `slices.rs`, `traits.rs` — ABI conversions use `u64`/`i64` for pointer and length parameters on wasm64
- `closure.rs`, `convert/closures.rs` — Closure data pointers use 64-bit ABI types on wasm64
- `externref.rs` — Table indices remain `u32` (always i32 per the Wasm spec); pointer operations use the correct width
- `lib.rs`, `rt/mod.rs`, `rt/marker.rs` — Use unified `core::arch::wasm::unreachable()` instead of wasm32-specific intrinsic

### CLI codegen (`crates/cli-support/`)

- `js/mod.rs` — Auto-detect `memory64` from the walrus module; generate `BigInt()`/`Number()` conversions for malloc, realloc, free, retptr, string/vector passing, and array getters; adjust strides (4→8) and read methods (`getUint32`→`getBigInt64`) for externref tables on wasm64
- `js/binding.rs` — New `coerce_ptr()`/`to_wasm_ptr()` helpers; `BigInt` wrapping for closure destructors and `__wbindgen_add_to_stack_pointer`
- `interpreter/mod.rs` — Handle `I64` constants (truncated to i32 for the descriptor protocol), `I64Add`/`I64Sub`/`I64And` binary ops, and `I32WrapI64`/`I64ExtendUI32`/`I64ExtendSI32` unary ops
- `wasm_conventions.rs` — Accept `I64` globals for stack pointer and TLS base
- `wit/incoming.rs`, `wit/outgoing.rs`, `wit/mod.rs` — Parameterize `AdapterType` — `I32` on wasm32, `I64` on wasm64 — threaded through all pointer/length positions in the WIT adapter layer

### Build / macro

- `build.rs` — Detect `wasm64` target for conditional compilation
- `crates/macro-support/src/codegen.rs` — Widen `cfg(target_arch)` gates to include wasm64

## Design decisions

1. **Detection** — memory64 is detected at the CLI level by inspecting the walrus module's first memory (`m.memory64`), not via flags. It just works.
2. **JS glue strategy** — Rather than a blanket `BigInt()` wrapper on every wasm call, conversions are inserted surgically where JS needs a Number (array indices, subarray offsets) or where wasm needs a BigInt (malloc/realloc/free arguments, retptr). This keeps the wasm32 codegen path completely unchanged.
3. **Interpreter** — The descriptor-extraction interpreter operates in i32 space (descriptor values are small). wasm64 `i64` constants are truncated, and wrap/extend ops are no-ops in this representation.
4. **Externref tables** — Table indices are always `i32` per the Wasm spec, even with memory64. Only pointer/length slots widen to `i64`.

## Testing

- Verified existing wasm32 codegen is byte-for-byte identical (zero regressions)
- Validated end-to-end with a real-world project ([web-rwkv](https://github.com/audreyt/web-rwkv)) compiled to `wasm64-unknown-unknown` — a 1.7B parameter LLM inference engine that requires >4 GB memory. The generated JS glue correctly handles model loading, tensor operations, and inference across the 64-bit boundary.


Trivo25:
i am crying (tears of joy)

guybedford (commented):
Wow, very nice work. Can we please add a CI run here, and include a test suite specifically for the 64 bit architecture that tests some of these new paths? Happy to land once we have that.

audreyt:
> Wow, very nice work. Can we please add a CI run here, and include a test suite specifically for the 64 bit architecture that tests some of these new paths? Happy to land once we have that.

Done! 🖖

crates/cli-support/src/js/binding.rs
```diff
@@ -1688,7 +1754,8 @@ fn instruction(
 
         Instruction::OptionNonNullFromI32 => {
             let val = js.pop();
-            js.push(format!("{val} === 0 ? undefined : {val} >>> 0"));
+            let coerced = js.coerce_ptr(&val);
+            js.push(format!("{val} === 0 ? undefined : {coerced}"));
         }
     }
```
guybedford, line 1741: **Bug (Critical):** On wasm64, `val` is a BigInt (from an i64 wasm return). In JavaScript, `0n === 0` is **`false`** — BigInt and Number are never strictly equal. This means `Option<NonNull<T>>` returning `None` (represented as `0n`) will never produce `undefined`. Instead it returns `Number(0n)` = `0`.

```js
// JS behavior:
0n === 0   // false
0n === 0n  // true
```

Suggested fix — compare against `0n` on memory64:
```rust
if js.cx.memory64 {
    js.push(format!("{val} === 0n ? undefined : Number({val})"));
} else {
    js.push(format!("{val} === 0 ? undefined : {val} >>> 0"));
}
```

crates/cli-support/src/js/mod.rs
```diff
@@ -2220,21 +2347,36 @@ if (require('worker_threads').isMainThread) {{
             name: "getArrayJsValueFromWasm".into(),
             num: mem.num,
         };
+        // For wasm64, externref table indices are stored as 64-bit pointers
+        // so we read with getBigInt64 and stride is 8. For wasm32 it's getUint32 / stride 4.
+        let (stride, get_method, ptr_fixup) = if self.memory64 {
+            (8, "getBigInt64", "ptr = Number(ptr); len = Number(len);")
```
guybedford: **Bug (Critical):** This comment says "externref table indices are stored as 64-bit pointers" — but that's incorrect. Externref table indices are always `u32` per the Wasm spec (as correctly noted elsewhere in this PR), and `JsValue.idx` is `u32` (unchanged). The **write side** (`passArrayJsValueToWasm`) correctly uses `setUint32` with stride 4.

Using `getBigInt64` with stride 8 on the read side will read 8 bytes where only 4 were written, producing garbage indices.

The stride/method should remain `(4, "getUint32", ...)` for JsValue arrays on both wasm32 and wasm64. Only the `ptr`/`len` parameters (the array base pointer and length) need BigInt→Number conversion; the element access itself should stay 32-bit.

The same issue applies to `getArrayJsValueViewFromWasm` below.

src/convert/impls.rs
```diff
@@ -460,7 +495,7 @@ impl<T> IntoWasmAbi for Option<*const T> {
 
     #[inline]
     fn into_abi(self) -> f64 {
-        self.map(|ptr| ptr as u32 as f64)
+        self.map(|ptr| ptr as usize as f64)
```
guybedford, line 466: **Bug (Critical):** On wasm64, `usize` is 64-bit but the ABI type is still `f64`. An `f64` has only 53 bits of mantissa precision, so any pointer value > 2^53 will be **silently rounded** during the round-trip, producing a different pointer.

Additionally, `F64_ABI_OPTION_SENTINEL` (which is `4294967297.0` = 2^32 + 1) is now a valid pointer value on wasm64, so `Some(0x1_0000_0001 as *const T)` would be misinterpreted as `None`.

For wasm64, `Option<*const T>` / `Option<*mut T>` should use a multi-value ABI (e.g., a discriminant `u32` + pointer `usize`) instead of the f64 sentinel trick. This may need a more significant refactor but is necessary for correctness.

crates/cli-support/src/js/binding.rs
```diff
@@ -1479,9 +1521,15 @@ fn instruction(
...
             if *owned {
                 let free = js.cx.export_name_of(*free);
-                js.prelude(&format!(
-                    "if ({ptr} !== 0) {{ wasm.{free}({ptr}, {len}, 1); }}",
-                ));
+                if js.cx.memory64 {
+                    js.prelude(&format!(
+                        "if (Number({ptr}) !== 0) {{ wasm.{free}(BigInt(Number({ptr})), BigInt(Number({len})), 1n); }}",
+                    ));
+                } else {
```
guybedford: **Issue (Medium):** The `BigInt(Number({ptr}))` pattern is a lossy round-trip for large values. If `{ptr}` is already a BigInt from a wasm i64 return, then `Number(BigInt_value)` loses precision for values > 2^53, and `BigInt(Number_value)` can't recover it.

Since `{ptr}` here comes from a wasm i64 return (already a BigInt on memory64), you could compare and pass directly:
```js
if ({ptr} !== 0n) {{ wasm.{free}({ptr}, {len}, 1n); }}
```

Or use `BigInt({ptr})` directly (which is a no-op if already BigInt, and works if it's a Number):
```js
if (Number({ptr}) !== 0) {{ wasm.{free}(BigInt({ptr}), BigInt({len}), 1n); }}
```

The same pattern appears in `VectorLoad` and `OptionVectorLoad` below.

src/externref.rs
```diff
@@ -35,7 +37,7 @@ impl Slab {
             let curr_len = self.data.len();
             if curr_len == self.data.capacity() {
                 let extra = max(128, curr_len);
-                let r = unsafe { __wbindgen_externref_table_grow(extra) };
+                let r = unsafe { __wbindgen_externref_table_grow(extra as i32) };
```
guybedford, line 41: **Issue (Medium):** `extra` is `max(128, curr_len)` where `curr_len` is `self.data.len()` — a `usize`, which is 64-bit on wasm64. If the slab ever grows beyond ~2 billion entries, `extra as i32` will silently truncate/wrap, potentially requesting a negative delta from `table.grow`.

While unlikely in practice, a bounds check would be safer:
```rust
let extra_i32: i32 = extra.try_into().unwrap_or_else(|_| internal_error("table grow delta overflow"));
```

guybedford (changes requested):
Thorough and well-structured PR — the detection strategy (inspecting `m.memory64` from walrus) and the surgical BigInt/Number insertion approach are both solid design choices. The wasm32 codegen path is preserved unchanged, which is great for zero-regression risk.

However, I found three critical runtime bugs that will affect wasm64 targets, plus several medium-severity issues. See inline comments for details.

**Summary:**

| # | Issue | Severity |
|---|-------|----------|
| 1 | `OptionNonNullFromI32`: `BigInt === 0` is always false in JS | Critical |
| 2 | `getArrayJsValueFromWasm`: stride 8 / `getBigInt64` but `JsValue.idx` is `u32` | Critical |
| 3 | `Option<*const T>` f64 ABI lossy for 64-bit pointers + sentinel collision | Critical |
| 4 | `BigInt(Number(ptr))` double conversion is lossy for large pointers | Medium |
| 5 | `extra as i32` truncation in table grow | Medium |

guybedford:
Added an AI review - let me know if it is relevant.

guybedford (approved):
I'd like to have another @wasm-bindgen/collaborators member review this as well, but overall looks good to land to me.

guybedford:
Note we need to add a CHANGELOG entry as well.

RReverser (changes requested):
Thanks! That's a pretty large PR, so just marking with "request changes" to have time to review it.

I'm also tempted to say that, unless it's urgent, we should postpone it for after the Emscripten one, since both touch a lot of places at the same time and I know they went through rebasing pain on top of our refactorings couple of times now.

guybedford:
@RReverser just a ping for review here - would be nice to include this for a release soon!

RReverser:
@guybedford As I said, I do want us to merge Emscripten PR first - these PRs would have a lot of conflicts, especially around cfg changes - but yeah, I will leave some comments meanwhile.

Knowing from Emscripten's wasm64 experience, this is not as trivial change as one might expect, so needs some extra care.

crates/cli-support/src/js/binding.rs
```diff
@@ -661,6 +661,24 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
...
+            format!("Number({val})")
+        } else {
+            format!("{val} >>> 0")
+        }
+    }
+
+    /// Convert a JS number to a wasm pointer argument.
+    /// For wasm32: identity (already a number).
+    /// For wasm64: identity (invoke handles BigInt wrapping at the wasm boundary).
+    pub fn to_wasm_ptr(&self, val: &str) -> String {
```
RReverser: Since this is no-op, I'd just remove it and revert relevant callsites to make the diff smaller. The comment about the opposite conversion being unnnecessary/identity can be added to `coerce_ptr` above.

crates/cli-support/src/js/binding.rs
```diff
@@ -1074,10 +1104,19 @@ fn instruction(
             // Note that we always assume the return pointer is argument 0,
             // which is currently the case for LLVM.
             let val = js.pop();
-            let expr = format!(
-                "{mem}().{method}({} + {size} * {offset}, {val}, true);",
-                js.arg(0),
-            );
+            // On memory64, the retptr (arg0) is a BigInt. DataView offsets
+            // must be Numbers, and setBigInt64 values must be BigInts.
+            let arg0 = if js.cx.memory64 {
```
RReverser: This should probably reuse `coerce_ptr`. I know it didn't have `>>> 0` before, but that seems like an overlooked bug - this is a pointer after all.

crates/cli-support/src/js/binding.rs
```diff
@@ -1138,14 +1179,14 @@ fn instruction(
             js.assert_not_moved(&val);
             let i = js.tmp();
             js.prelude(&format!("var ptr{i} = {val}.__destroy_into_raw();"));
-            js.push(format!("ptr{i}"));
+            js.push(js.to_wasm_ptr(&format!("ptr{i}")));
```
RReverser: These will be automatically reverted if `to_wasm_ptr` is removed.

crates/cli-support/src/js/binding.rs
```diff
@@ -1479,9 +1521,15 @@ fn instruction(
 
             if *owned {
                 let free = js.cx.export_name_of(*free);
-                js.prelude(&format!(
-                    "if ({ptr} !== 0) {{ wasm.{free}({ptr}, {len}, 1); }}",
-                ));
+                if js.cx.memory64 {
+                    js.prelude(&format!(
+                        "if ({ptr} !== 0n) {{ wasm.{free}({ptr}, {len}, 1n); }}",
```
RReverser: Let's just change the condition to `if (ptr)`, that works for both wasm64 and wasm32 and avoids branching.

crates/cli-support/src/js/binding.rs
```diff
@@ -1055,10 +1077,18 @@ fn instruction(
 
         Instruction::Retptr { size } => {
             js.cx.inject_stack_pointer_shim()?;
-            js.prelude(&format!(
-                "const retptr = wasm.__wbindgen_add_to_stack_pointer(-{size});"
-            ));
-            js.finally(&format!("wasm.__wbindgen_add_to_stack_pointer({size});"));
+            let memory64 = js.cx.memory64;
+            if memory64 {
```
RReverser: I'm not a fan of us having lots of similar `if`s just to conditionally add `n`.

Let's add a centralised helper like `size_to_string` or something that returns a pre-formatted `42` or `42n` correspondingly.

crates/cli-support/src/js/binding.rs
```diff
@@ -1579,10 +1627,17 @@ fn instruction(
             let i = js.tmp();
             let free = js.cx.export_name_of(*free);
             js.prelude(&format!("var v{i} = {f}({ptr}, {len}).slice();"));
-            js.prelude(&format!(
-                "wasm.{free}({ptr}, {len} * {size}, {size});",
-                size = kind.size()
-            ));
+            if js.cx.memory64 {
+                js.prelude(&format!(
+                    "wasm.{free}({ptr}, {len} * {size}n, {size}n);",
```
RReverser: Same here and elsewhere re:a single size formatting helper instead of repetitive branching.

crates/cli-support/src/js/binding.rs
```diff
@@ -1688,7 +1754,9 @@ fn instruction(
 
         Instruction::OptionNonNullFromI32 => {
             let val = js.pop();
-            js.push(format!("{val} === 0 ? undefined : {val} >>> 0"));
+            let coerced = js.coerce_ptr(&val);
+            let zero = if js.cx.memory64 { "0n" } else { "0" };
+            js.push(format!("{val} === {zero} ? undefined : {coerced}"));
```
RReverser: Same here about simplifying the condition instead of generating different literals.

```suggestion
            js.push(format!("{val} ? {coerced} : undefined"));
```

crates/cli-support/src/js/binding.rs
```diff
@@ -1757,7 +1825,25 @@ impl Invocation {
                     Some(eid) => cx.module.exports.get(*eid).name.clone(),
                     None => cx.export_name_of(*id),
                 };
-                Ok(format!("wasm.{name}({})", args.join(", ")))
+                if cx.memory64 {
+                    // On memory64, i64 parameters need BigInt wrapping
```
RReverser: When would they not be already bigints at this point?

crates/cli-support/src/js/mod.rs
```diff
@@ -1330,10 +1380,15 @@ if (require('worker_threads').isMainThread) {{
                 ("obj.__wbg_ptr = ptr;", "obj.__wbg_ptr")
             };
 
+            let ptr_coerce = if self.memory64 {
```
RReverser: This seems like repetition of what we are already doing in js/binding, the helpers should be merged.

crates/cli-support/src/js/mod.rs
```diff
@@ -1818,13 +1891,28 @@ if (require('worker_threads').isMainThread) {{
...
-            let debug = if self.config.debug {
+            let debug = if debug_mode {
                 "if (typeof(arg) !== 'string') throw new Error(`expected a string argument, found ${typeof(arg)}`);\n"
             } else {
                 ""
             };
 
+            // For memory64, malloc/realloc args need BigInt wrapping and
+            // return values need Number() conversion for use as array indices.
+            let malloc_buf = if memory64 {
```
RReverser: These should also reuse the helpers.

crates/macro-support/src/codegen.rs
```diff
@@ -863,7 +868,7 @@ impl TryToTokens for ast::Export {
                 #wasm_bindgen::__wbindgen_coverage! {
                 #(#attrs)*
                 #[cfg_attr(
-                    all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")),
+                    all(any(target_arch = "wasm32", target_arch = "wasm64"), any(target_os = "unknown", target_os = "none")),
```
RReverser: Please see #4443 - we decided to use `cfg(target_family)` there, so it's worth modifying this PR to use the same. That will lead to much simpler conditions as well as fewer merge conflicts once it's in.

src/convert/impls.rs
```diff
@@ -224,6 +227,38 @@ type_wasm_native_f64_option!(
...
 );
 
+#[cfg(target_pointer_width = "64")]
+type_wasm_native_f64_option!(
+    i32 as i32
+    u32 as u32
+    f32 as f32
+);
+
+#[cfg(target_pointer_width = "64")]
```
RReverser: Instead of this branching, we should implement the WasmAbi trait for `usize`/`isize` directly.

src/convert/impls.rs
```diff
@@ -455,23 +490,38 @@ unsafe impl<T: ErasableGeneric> ErasableGeneric for *const T {
...
     fn into_abi(self) -> f64 {
-        self.map(|ptr| ptr as u32 as f64)
+        self.map(|ptr| ptr as usize as f64)
             .unwrap_or(F64_ABI_OPTION_SENTINEL)
     }
 }
 
+#[cfg(target_pointer_width = "64")]
+impl<T> IntoWasmAbi for Option<*const T> {
+    type Abi = Option<usize>;
```
RReverser: Instead of branching and using a large ABI, you can just change the sentinel to `2^53-1 = 9007199254740991` aka `Number.MAX_SAFE_INTEGER`.

That number is about 9 petabytes, and I'm fairly sure nobody would allocate that much even in wasm64 in the foreseeable future.

RReverser (commented):
The primary concern I had about pointer handling and compatibility is actually already resolved - thanks - so most of review will be about cleaning the PR up and reducing the diff somewhat.

Like with Emscripten PR, I'd prefer a few clear-cut small helpers that are reused everywhere we need to do something different on wasm32 vs wasm64, instead of having lots of branching like

```rs
if memory64 {
  ...long expression...
} else {
  ...very similar long expression that might accidentally get out of sync in the future...
}
```

I went through the codebase and left some comments, but this will likely need another couple of reviews before landing as it's a pretty large PR. Thanks for working on it!

RReverser:
One other thing in addition to my reviews above: let's try and add ABI trait implementations for `u64`, `i64`, `*const T` and `*mut T` to use `f64` on wasm64 target.

This should give the same representation as we get with this PR, but without generating extra JS code - all conversions will be done purely on the Wasm side - which should leave only a few places where custom JS `Number(...)` wrapping is still needed.

guybedford:
With the Emscripten PR now landed, it would be great to get moving on this. @RReverser do you have any remaining concerns?

@audreyt thanks for your patience and persistence, I hope we can land this next week for the next release after this current one.

RReverser:
> @RReverser do you have any remaining concerns?

Yeah see the open comments above.

crates/cli-support/src/js/binding.rs
```diff
@@ -1054,10 +1083,19 @@ fn instruction(
...
+            let size = js.size_literal(usize::try_from(*size).unwrap());
+            let memory64 = js.cx.memory64;
+            if memory64 {
+                js.prelude(&format!(
+                    "const retptr = Number(wasm.__wbindgen_add_to_stack_pointer(-{size}));"
+                ));
+                js.finally(&format!("wasm.__wbindgen_add_to_stack_pointer({size});"));
+            } else {
+                js.prelude(&format!(
+                    "const retptr = wasm.__wbindgen_add_to_stack_pointer(-{size});"
```
RReverser: Please carefully review all the code, there is clearly still a bunch of branching duplication that should instead use the shared helpers as suggested above.

crates/cli-support/src/js/binding.rs
```diff
@@ -117,6 +117,8 @@ pub fn maybe_wrap_try_catch(call: &str, should_check_aborted: bool) -> String {
     }
 }
 
+const F64_OPTION_SENTINEL: &str = "9007199254740991";
```
RReverser: This should use `Number.MAX_SAFE_INTEGER` not inlined magic number.

crates/cli-support/src/js/binding.rs
RReverser, line 1457: Shouldn't this and other similar branches change since we're no longer doing just "from i32" conversions?

crates/cli-support/src/js/mod.rs
```diff
@@ -1578,30 +1699,56 @@ if (require('worker_threads').isMainThread) {{
         }
 
         if class.unwrap_needed {
+            let unwrap_null = format!("0{}", self.wasm_bigint_suffix());
```
RReverser: The diff of this logic seems wrong, in JS API pointers should always be numbers, so why are we doing conversions to BigInt here?

crates/cli-support/src/transforms/multi_value/mod.rs
```diff
@@ -179,18 +181,33 @@ fn xform_one(
...
             "can only multi-value transform functions that don't return any \
-             results (since they should be returned on the stack via a pointer)"
+             results (since they should be returned on the stack via a \
+             pointer): `{func_name}` has params {ty_params:?} and results \
+             {ty_results:?}"
         );
     }
 
+    let ptr_ty = if memory64 {
+        walrus::ValType::I64
```
RReverser, line 204: Why is this not F64 if we try to use it as the ABI?

audreyt: This transform runs on the core Wasm function after adapter lowering, so the return-pointer local/global still uses the actual memory64 pointer type (`i64`) here. The JS-facing `f64` ABI only exists in the adapter layer, which is why I added a clarifying comment and left the multivalue shim itself on `i64`.

crates/cli-support/src/interpreter/mod.rs
```diff
@@ -233,7 +233,11 @@ impl Frame<'_> {
             match instr {
                 Instr::Const(c) => match c.value {
                     Value::I32(n) => stack.push(n),
-                    _ => bail!("non-i32 constant"),
+                    // For wasm64, truncate i64 constants to i32. The descriptor
+                    // protocol only uses small values, and pointer arithmetic
+                    // in our tiny interpreter fits in i32.
+                    Value::I64(n) => stack.push(n as i32),
```
guybedford, line 282: We should probably explicitly bail for out of range here then?

crates/cli-support/src/js/mod.rs
```diff
@@ -295,6 +304,117 @@ impl<'a> Context<'a> {
...
+            "n"
+        } else {
+            ""
+        }
+    }
+
+    /// Formats a pointer-sized literal for the current target ABI.
+    fn usize_literal(&self, value: usize) -> String {
+        format!("{value}{}", self.wasm_bigint_suffix())
+    }
```
guybedford: This seems a little unnecessarily convoluted. Why not just:

```suggestion
    /// Formats a pointer-sized literal for the current target ABI.
    fn usize_literal(&self, value: usize) -> String {
        if self.memory64 {
          format!("{value}n")
        } else {
          value.to_string()
        }
    }
```

and the same for other callers of wasm_bigint_suffix.

crates/cli-support/src/js/binding.rs
```diff
@@ -1678,19 +1718,26 @@ fn instruction(
...
             let val = js.pop();
             js.cx.expose_is_like_none();
-            js.assert_optional_number(&val);
-            js.push(format!("isLikeNone({val}) ? 0 : {val}"));
+            if js.cx.memory64 {
+                js.assert_optional_bigint(&val);
+            } else {
+                js.assert_optional_number(&val);
+            }
+            let zero = format!("0{}", js.cx.wasm_bigint_suffix());
```
guybedford: Everywhere we create zero, why not just do `js.cx.usize_literal(0)` instead?

crates/cli-support/src/js/mod.rs
```diff
@@ -295,6 +304,117 @@ impl<'a> Context<'a> {
...
+
+    /// Coerce a wasm pointer-sized expression to an unsigned JS pointer expression.
+    fn coerce_ptr_expr(&self, expr: &str) -> String {
+        format!("{}{}", self.to_number(expr), self.ptr_coerce())
+    }
+
+    /// Coerce a wasm pointer-sized value to an unsigned JS pointer.
+    fn to_js_ptr(&self, expr: &str) -> String {
+        format!("({})", self.coerce_ptr_expr(expr))
+    }
```
guybedford: Again, this is a little convoluted and hard to follow. Why not just:

```suggestion
    /// Coerce a wasm pointer-sized value to an unsigned JS pointer.
    fn to_js_ptr(&self, expr: &str) -> String {
        if self.memory64 {
          format!("Number({expr})")
        } else {
          format!("({expr} >>> 0)")
        }
    }
```

Which also results in better output in the memory64 case as it removes the unnecessary `()` wrapper.

guybedford (commented):
Did another pass, in preparation for the next release. Because this touches such important internals simplicity and clarity are so so important.

Myself and @RReverser are happy to continue reviews until it is ready.

crates/cli-support/src/js/binding.rs
audreyt: I fixed the comments in `ac5cb77c8` — please let me know if you'd like a full rename, e.g.:

I32FromExternrefRustOwned -> WasmPtrFromExternrefRustOwned
I32FromExternrefRustBorrow -> WasmPtrFromExternrefRustBorrow
I32FromOptionRust -> WasmPtrFromOptionRust
RustFromI32 -> RustFromWasmPtr
OptionRustFromI32 -> OptionRustFromWasmPtr

guybedford (commented):
This is looking great.

One last request from me -

I think the main argument that allows me to get on board with the I64AsF64 approach is that we already have precision loss in the isize and usize definitions on 64 bit machines in Wasm Bindgen.

I think this gives a great backwards-compat story for the js-sys APIs exactly as we want and is a neat approach and works great.

But, while I'm on board with this for `isize` and `usize`, I think we must ensure that `i64` and `u64` still use `BigInt` with proper precision.

That is, instead of having `isize` and `usize` go through descriptor `Descriptor::I64` which then goes to `Descriptor::I64asF64`, I think we should rather:

1. Add `I64_AS_F64` and `U64_AS_F64` constants to `crates/shared/src/tys.rs`
2. Add corresponding decode handling in `crates/cli-control/src/descriptor.rs`
3. Then `describe.rs` can map `isize` / `usize` directly to these new wire type constants under memory64

RReverser (changes requested):
For me one other thing is that JS changes are still pretty extensive where a Wasm-side change would suffice.

We have many purely internal Wasm helpers exposed to JS for pointer manipulation (for tasks like "add to stack pointer", "free", "allocate", callback pointers etc).

We should be able to just change signatures of such functions to some custom type like

```rust
#[repr(transparent)]
struct Pointer<T>(*const T);

#[cfg(target_arch = "wasm64")]
#[repr(transparent)]
struct Pointer<T>(f64);
```

with conversion methods that do no-op on wasm32 and f64<>pointer conversion on wasm64.

This would both improve typechecking - everything validated statically using the more rigorous Rust typesystem than JS runtime errors - and significantly reduce complexity of JS diff, since classes, callbacks, etc etc will be able to store numeric pointers like before and don't even need to know that anything changed on the Wasm side.

Basically same benefit as for external consumers already applied via adapters, but we can extend it to our own intrinsics as well.

audreyt:
Thanks again for the careful review and patience here. I’ve addressed the latest feedback, fixed the wasm64 follow-up regressions, and refreshed the affected references. When you have time, I’d really appreciate another look.

RReverser:
@audreyt Can you please go over js/mod.rs diff and revert changes that are no longer necessary now that intrinsics expose / accept f64 directly on the Wasm side? That was the main benefit of my suggestion above but the diff in js/mod.rs still looks huge with lots of pointer coercion that should be unnecessary now. That makes it hard to review - I think I can spot some genuine regressions but difficult to say because of the "noise".

It might be best to do this by hand using hunk feature (eg "revert selected range" in VSCode), as I find AIs still pretty bad at trying to minimise diffs - they only keep adding more additive changes on top of existing ones when working on large features, which goes against the goal of making PRs easily reviewable.

src/rt/marker.rs
```diff
@@ -37,16 +37,28 @@ pub struct CheckSupportsInstanceProperty<T: SupportsInstanceProperty>(T);
 pub trait SupportsStaticProperty {}
 pub struct CheckSupportsStaticProperty<T: SupportsStaticProperty>(T);
 
-#[cfg(all(feature = "std", target_arch = "wasm32", panic = "unwind"))]
+#[cfg(all(
+    feature = "std",
+    any(target_arch = "wasm32", target_arch = "wasm64"),
```
RReverser: Any such blocks should use the simpler `target_family` check instead.

src/closure.rs
```diff
@@ -730,10 +736,38 @@ impl<T: ?Sized> AsRef<JsValue> for ScopedClosure<'_, T> {
 #[repr(transparent)]
 struct OwnedClosure<T: ?Sized, const UNWIND_SAFE: bool>(Box<T>);
 
+#[repr(C)]
+#[derive(Clone, Copy)]
+#[doc(hidden)]
+pub struct WasmClosureSlice {
```
RReverser: Why is this new type necessary? The `WasmWord` changes should be applied to the regular `WasmSlice` since it always represents a pair of words.

src/closure.rs
```diff
@@ -746,8 +780,11 @@ struct BorrowedClosure<T: ?Sized, const UNWIND_SAFE: bool> {
 /// `Box<dyn FnOnce/FnMut/Fn>` closure, or `a` must be zero (in which case this
 /// is a no-op).
 #[no_mangle]
-pub unsafe extern "C" fn __wbindgen_destroy_closure(a: usize, b: usize) {
-    if a == 0 {
+pub unsafe extern "C" fn __wbindgen_destroy_closure(
+    a: crate::__rt::WasmWord,
```
RReverser: Would be cleaner to add `WasmWord` once via `use ...` than repeat full path everywhere.

src/closure.rs
RReverser: (which, in turn, means that slice and string conversions should be able to go via the same wrapper)

src/lib.rs
```diff
@@ -1040,7 +1042,14 @@ macro_rules! big_integers {
         impl From<$n> for JsValue {
             #[inline]
             fn from(arg: $n) -> JsValue {
-                wbg_cast(arg)
+                #[cfg(target_arch = "wasm64")]
+                {
+                    JsValue::bigint_from_str(&arg.to_string())
```
RReverser: This looks suspicious. We shouldn't need to do anything different for i64 / u64 here.

src/lib.rs
```diff
@@ -1056,16 +1065,26 @@ macro_rules! big_integers {
...
+                    let as_self = as_i64 as $n;
+                    // Double-check that we didn't truncate the bigint to 64 bits.
+                    if val == &as_self {
+                        return Some(as_self);
+                    }
                 }
+
+                #[cfg(target_arch = "wasm64")]
+                {
+                    let number = val.as_f64()?;
```
RReverser: Same here, I don't see why this change is necessary. Regular i64/u64 should remain as-is at the ABI level, so high-level code shouldn't worry about wasm32/wasm64.

tests/wasm/bigint.js
```diff
@@ -1,43 +1,61 @@
...
 
 exports.js_works = () => {
-    assert.strictEqual(wasm.zero(), BigInt('0'));
-    assert.strictEqual(wasm.one(), BigInt('1'));
-    assert.strictEqual(wasm.neg_one(), BigInt('-1'));
-    assert.strictEqual(wasm.i32_min(), BigInt('-2147483648'));
-    assert.strictEqual(wasm.u32_max(), BigInt('4294967295'));
-    assert.strictEqual(wasm.i64_min(), BigInt('-9223372036854775808'));
-    assert.strictEqual(wasm.u64_max(), BigInt('18446744073709551615'));
+    assert.strictEqual(wasm.zero(), scalar64(0n));
```
RReverser: These changes to tests should also be reverted, the representation of these numbers wasn't supposed to change so editing tests might be hiding bugs in actual source.

tests/wasm/bigint.rs
```diff
@@ -100,7 +100,13 @@ mod try_from_works {
                 fn $ty() {
                     // Not a bigint.
                     assert!($ty::try_from(JsValue::NULL).is_err());
-                    assert!($ty::try_from(JsValue::from_f64(0.0)).is_err());
+                    if cfg!(target_arch = "wasm64")
+                        && (stringify!($ty) == "i64" || stringify!($ty) == "u64")
```
RReverser: Same.

RReverser:
Tried to read a few more files, and overarching concern remains the same - there seems to be too many changes that shouldn't be necessary for this PR. Some of them might be simple noise, and others might be genuine regressions.

@audreyt Please walk carefully over the PR diff and make sure that every change block is justified or revert if not.

audreyt:
Thanks for the review! I did one more careful pass over `mod.rs`, `binding.rs`, `impls.rs` and `lib.rs`. The memory64 normalization in `wit/mod.rs` is narrowly scoped to `Option<i64/u64>` descriptors whose core wasm ABI is actually `f64`, the JS glue now keeps public 64-bit values on the precise `BigInt` path while only decoding internal struct/vector/free pointers back into JS number space where the runtime expects numeric offsets, and the updated wasm64/reference tests cover those paths.

src/lib.rs
RReverser: @audreyt Please check the open comments from earlier review round too. Looks like this one and some others might have gotten lost?

It's within the same topic of either unnecessary changes or regressions.

I'd suggest explicitly marking comments as reserved when they are or leaving a comment if there are follow ups, so we have a clearer view of what's left.

crates/cli-support/src/js/binding.rs
```diff
@@ -1581,10 +1640,14 @@ fn instruction(
             let i = js.tmp();
             let free = js.cx.export_name_of(*free);
             js.prelude(&format!("var v{i} = {f}({ptr}, {len}).slice();"));
-            js.prelude(&format!(
-                "wasm.{free}({ptr}, {len} * {size}, {size});",
-                size = kind.size()
-            ));
+            let size = js.size_literal(kind.size());
```
RReverser: @audreyt Why do we still need these changes to JS?

crates/cli-support/src/js/mod.rs
```diff
@@ -1564,10 +1669,11 @@ if (require('worker_threads').isMainThread) {{
                 ("obj.__wbg_ptr = ptr;", "obj.__wbg_ptr")
             };
 
+            let ptr_fixup = format!("ptr = {};", self.to_js_ptr_inline("ptr"));
```
RReverser: Same here, why is this change necessary if Wasm already returns a number now?

crates/cli-support/src/js/mod.rs
```diff
@@ -2100,48 +2202,36 @@ if (require('worker_threads').isMainThread) {{
...
-        // expensive in mainstream engines than staying in the JS, and
-        // charCodeAt on ASCII strings is usually optimised to raw bytes.
-        let encode_as_ascii = format!(
-            "\
+            let encode_as_ascii = format!(
+                "\
                 if (realloc === undefined) {{
                     const buf = cachedTextEncoder.encode(arg);
-                    const ptr = malloc(buf.length, 1) >>> 0;
+                    const ptr = {malloc_buf};
```
RReverser: Same for malloc/realloc/free usages.

audreyt:
Thanks again for all the careful review here, and for pushing on keeping this diff tight.

I went back through the remaining feedback, trimmed the JS glue changes further, and addressed the last open allocator-callsite concern in `js/mod.rs`.

Whenever you have a chance, I’d be very grateful for another pass.

src/rt/mod.rs
```diff
@@ -277,6 +282,165 @@ pub fn assert_not_null<T>(s: *mut T) {
     }
 }
 
+#[cfg(target_arch = "wasm64")]
+type WasmWordRepr = f64;
```
guybedford: This results in precision loss, we need to ensure that we properly represent these pointers.

guybedford (changes requested):
Perhaps my last comment was misleading with regards to precision loss - let me clarify what I meant by that. The precision loss I was referring to being okay is that `usize` when used as an argument or return value in imported and exports **wasm bindgen** ABIs is fine. But that we must still properly represent Rust internal data structures that `usize` and `isize` as not truncating but correct. That is, the imprecision should only defined to those few js-sys functions that use `usize` and `isize` as argument types, but everywhere else internally we need to do these conversions we maintain full parity.

This was why I suggested defining the types I64asF64 as the way this conversion gets done as distinct from all the other places we need to do proper parity conversions.

In short, my suggestion further here is that `WasmWord` and related usage uses either the BigInt / Number path as I believe it did before so that it is always accurate.

@RReverser we may also be talking over eachother here as your recent comments seem to suggest you're on board with precision loss in general for the design. It sounds like we may need to discuss this further as well to ensure we're not talking across eachother on this concern.

audreyt:
Thanks again for the careful review here.

I pushed one more pass to make the wasm64 split narrower: public `usize` / `isize` stays on the JS-number ABI, while hidden internal exported-class pointers stay exact on the wasm64 raw-pointer path. I also made the hidden raw field non-enumerable again, added regression coverage for the tricky ABI boundaries, and reran the relevant local CLI + wasm64 checks in OrbStack.

Would appreciate another look when you have time.

RReverser:
> you're on board with precision loss in general for the design

Yeah, I believe that's a much cleaner change:

1. We don't really lose any precision as long as we use this only for actual pointers and lengths which will most definitely not exceed 2^53 bytes in the next decade or few. 
2. Our intrinsics shouldn't need special treatment in terms of ABI - if it's good enough for users, it's good enough for us. In fact, I'm planning to do something similar to https://github.com/wasm-bindgen/wasm-bindgen/pull/4636 but for exported intrinsics after this PR to make it even easier.
3. The change is much smaller if we have consistent Rust representation at the boundary. 
4. It's more difficult for runtime bugs to slip by (number and bigint getting mixed up and throwing on arithmetic ops) in both this and future PRs if we always deal with number type on the JS side for pointers, which makes maintenance a lot easier.

I'm happy to discuss further next week if this is not compelling enough :)

crates/cli-support/src/js/binding.rs
```diff
@@ -197,9 +206,18 @@ impl<'a, 'b> Builder<'a, 'b> {
...
+                    js.args.push("ptr".into());
+                }
             } else {
-                js.args.push("this.__wbg_ptr".into());
+                let ptr = js.class_raw_ptr("this");
+                if js.cx.memory64 {
+                    js.args.push(raw_wasm_ptr_expr(&ptr));
+                } else {
+                    js.args.push(ptr);
+                }
```
guybedford: This seems a bit needlessly complicated? What happened to just having a well-defined pointer concept here? I'd be really interested to understand exactly the set of design decisions that resulted in the need for both __wbg_ptr and __wbg_ptr_raw to coexist - I'd have thought we'd have one Wasm pointer type which would be a BigInt in mem64 and a normal number in 32 bit. @RReverser does that match your expectations?

audreyt: OK. I hid the exact wasm64 pointer slot behind an internal symbol so the public JS-visible shape is back to just `__wbg_ptr` as the numeric view.

Semantically the split is still there: the numeric view is what JS sees, while the exact raw pointer is kept only for the internal wasm64 round-trips that need full precision (`&self` / `self`, `__unwrap`, `free`, finalization). This felt like the smallest change that keeps the JS surface simpler without reintroducing precision loss?

guybedford:
@audreyt just to update you here - we discussed this at the wasm-bindgen core meeting today and were able to come to a shared view on the behaviour. My sincere apologies to you for not syncing on this with @RReverser sooner in the process, it appears we very much have been pushing different design goals.

While I was suggesting BigInt, @RReverser was suggesting simply using number. The PR therefore went back and fourth between these designs!

We came to an agreement that `number` is the correct representation since Wasm Memory 64 is bounded by an upper pointer size limit anyway. Therefore the last few commits can be removed and reverted back to the state where everything used is just `number` for the Wasm pointers.

Thanks for your patience on this and apologies for the confusion. We may also be able to help contribute some work here to help get this over the line.

audreyt:
Hi @guybedford, thanks so much for syncing on this and for the clear write-up, and no worries at all about the back-and-forth. I really appreciate you and @RReverser taking the time to get to a shared view here.

I’ve updated the branch to follow that direction: wasm pointers are back on the plain `number` path, and the extra raw/BigInt scaffolding from the last few commits has been removed. 

Really appreciate the guidance and patience here.

crates/cli-support/src/js/binding.rs
RReverser: @audreyt Looks like these and other pointer changes came back in the last round, can you clean them up again? Since we agreed on pointers being numbers, most of these JS changes should be unnecessary after all.

audreyt: Fixed!

RReverser:
I spent ~2 hours yesterday going over the JS codegen changes myself, reverting chunk by chunk as described above, and reduced its diff from 307/131 (438 lines) to 106/34 (140 lines) — a 68% reduction — while preserving wasm64 functionality.

Most conversions now happen only in Wasm as requested, with a few `>>> 0` happening on the JS side exclusively when we pass the pointer to TypedArray or DataView methods - the only ones that care about unsigned values.

RReverser (approved):

RReverser:
@guybedford would love your thorough review on this too before landing too.

src/convert/impls.rs
```diff
@@ -143,26 +143,122 @@ macro_rules! type_wasm_native {
 }
 
 type_wasm_native!(
-    i64 as i64
-    u64 as u64
```
guybedford, line 147: Why do these need to be changed?

src/convert/traits.rs
```diff
@@ -186,6 +186,7 @@ unsafe impl WasmPrimitive for u64 {}
 unsafe impl WasmPrimitive for i64 {}
 unsafe impl WasmPrimitive for f32 {}
 unsafe impl WasmPrimitive for f64 {}
+unsafe impl WasmPrimitive for usize {}
```
guybedford: Why not `isize`? Are we sure we need this?

src/describe.rs
```diff
@@ -50,15 +50,28 @@ simple! {
...
     f32 => F32
     f64 => F64
     bool => BOOLEAN
     char => CHAR
     JsValue => EXTERNREF
 }
 
+// isize/usize map to I32/U32 on wasm32 and direct *_AS_F64 descriptors on wasm64
+cfg_if! {
+    if #[cfg(target_pointer_width = "64")] {
```
guybedford: How come here we use this attribute, but in other places we use the wasm64 target check? Should we consistently pick one or the other?

crates/cli-support/src/js/mod.rs
```diff
@@ -1618,13 +1642,13 @@ if (require('worker_threads').isMainThread) {{
...
                 "({{ ptr, instance }}) => {{
-                if (instance === __wbg_instance_id) wasm.{}(ptr >>> 0, 1);
+                if (instance === __wbg_instance_id) wasm.{}(ptr, 1);
             }}",
                 wasm_bindgen_shared::free_function(qualified)
             )
         } else {
             format!(
-                "ptr => wasm.{}(ptr >>> 0, 1)",
+                "ptr => wasm.{}(ptr, 1)",
```
guybedford, line 1651: Can you explain why these fixups are no longer needed?

crates/cli-support/src/js/mod.rs
```diff
@@ -1592,7 +1617,6 @@ if (require('worker_threads').isMainThread) {{
             dst.push_str(&format!(
                 "\
                 static __wrap(ptr) {{
-                    ptr = ptr >>> 0;
```
guybedford, line 1595: Can you explain why this can be removed?

guybedford (approved):
Thanks for streamlining the work here @RReverser. Just a few questions but otherwise good to land.

src/convert/traits.rs
RReverser: Good catch, probably not, another revert opportunity.

src/convert/impls.rs
RReverser: I guess another Claude artifact too that can be reverted.

crates/cli-support/src/js/mod.rs
RReverser: So, as mentioned above, in adding the pointer adjustments I tried to isolate them to the places where we pass pointers / sizes to typed arrays and dataviews.

This makes it a bit easier to reason about them than having pointer adjustment scattered on every malloc/realloc/free call even when we treat it as an opaque handle to be stored and passed back to Wasm later.

Note that `>>> 0` was also problematic for wasm64, since it would truncate integers above 32 bits, so the alternative fix would be to add the `{coerce_ptr}` helper usage to all these places. That would be slightly bigger change for no particular benefit.

crates/cli-support/src/js/mod.rs
RReverser: Answered above, I'll mark this thread as resolved to keep discussion in one place.

src/convert/impls.rs
guybedford: Reverted `i64`/`u64` back into the `type_wasm_native!` macro alongside `i128`/`u128`/`f64`. The wasm64 split was unnecessary since the macro produces identical codegen on both targets. The pointer / `usize` / `isize` impls now use `Abi = u32`/`i32` on wasm32 (matching upstream convention) and `Abi = f64` on wasm64.

src/convert/traits.rs
guybedford: Removed `unsafe impl WasmPrimitive for usize {}`. The reason it was needed (and not for `isize`) was that both `usize` and `isize` had `type Abi = usize` on wasm32. I changed the wasm32 ABI for `usize`/`isize`/`*const T`/`*mut T`/`NonNull<T>` and `WasmWordRepr` to use `u32`/`i32` instead, which both matches the upstream pre-PR convention (`type_wasm_native_f64_option!(isize as i32, usize as u32, ...)`) and removes the need for the new `WasmPrimitive` impl entirely.

src/describe.rs
guybedford: Good catch \u2014 changed `target_pointer_width = "64"` to `target_arch = "wasm64"` for consistency with the rest of the PR. This is the more semantically correct guard since the describe output only matters when targeting wasm specifically (on a native target the describe machinery doesn\'t cross any ABI boundary).

crates/cli-support/src/js/mod.rs
guybedford: They were genuinely needed on wasm32 \u2014 restoring them was a regression from the JS-codegen diff reduction. The `>>> 0` coerces a potentially sign-extended i32 pointer return into an unsigned JS number before storage / use as a `__wbg_ptr`, and matters for any pointer \u2265 2^31. Restored both `__wrap`\'s `ptr = ptr >>> 0;` and the finalization callbacks\' `wasm.\${free}(ptr >>> 0, 1)` using the existing `coerce_ptr_assign` / `coerce_ptr_suffix` helpers, so they only emit on wasm32 (on wasm64 the pointer is already a positive JS number and `>>> 0` would truncate it to 32 bits).

guybedford (approved):
Pushed an update addressing the remaining review feedback and rebased onto latest `main`. Squashed the 64-commit history into a single commit for clarity.

**Changes since last review:**

- **`src/convert/impls.rs`** — Reverted the wasm64 split of `i64`/`u64` back into `type_wasm_native!` alongside `i128`/`u128`/`f64`. The split was unnecessary since the macro produces identical codegen on both targets.
- **`src/convert/traits.rs`** — Removed `WasmPrimitive for usize`. Updated `usize`/`isize`/`*const T`/`*mut T`/`NonNull<T>` and `WasmWordRepr` to use `Abi = u32`/`i32` on wasm32 (matching upstream's pre-PR convention) and `Abi = f64` on wasm64. This both answers "why not isize?" (the answer was that both shared `Abi = usize`) and removes the need for the new `WasmPrimitive` impl entirely.
- **`src/describe.rs`** — Switched `target_pointer_width = "64"` to `target_arch = "wasm64"` for cfg consistency with the rest of the PR.
- **`crates/cli-support/src/js/mod.rs`** — Restored the `>>> 0` coercions in `__wrap` and the finalization callbacks. These were a genuine wasm32 regression for pointers ≥ 2^31. Used the existing `coerce_ptr_assign` / `coerce_ptr_suffix` helpers so they only emit on wasm32.
- **`src/rt/mod.rs`** — Gated the `wasm_word_roundtrips_large_pointer_values` test on `target_arch = "wasm64"` (was previously `target_pointer_width = "64"`, which caused failures on native 64-bit hosts after `WasmWordRepr` changed to `u32` on wasm32).

`just test-cli-overwrite` passes (72/72 + 2/2 runner tests). Reference snapshots updated to reflect the restored `>>> 0` coercions.

Marking my open review threads as resolved. Approving — ready to land once green.

guybedford:
Correction on the `>>> 0` discussion: @RReverser is right that the coercions in `__wrap` and the finalization callback aren't needed even on wasm32. JS-to-wasm i32 calls already go through `ToInt32`, and `__wbg_ptr` is only ever passed back to wasm or compared against zero — never used in JS arithmetic where sign would matter. The original PR removal was correct; my restoration was the regression. Reverted in the latest force-push.

The remaining `>>> 0` callsites elsewhere in the codegen (TypedArray/DataView indices) stay as-is since those genuinely need the unsigned interpretation.

src/convert/impls.rs
```diff
@@ -218,12 +217,140 @@ macro_rules! type_wasm_native_f64_option {
 
 type_wasm_native_f64_option!(
     i32 as i32
-    isize as i32
     u32 as u32
-    usize as u32
     f32 as f32
 );
 
+impl IntoWasmAbi for usize {
```
RReverser: All of these changes are overly verbose.

We already have `WasmWord`/`WasmWordRepr` for this, pretty sure if you just changed the macro usage above from `isize as i32` to `isize as WasmWordRepr` that would be enough to eliminate most of these manual impls?

Same for other `#[cfg(target_arch = "wasm64")]` usages in this file (aside from declaration of `WasmWordRepr` itself, it should be the only one that has `cfg` to switch between representations).

guybedford: Good call \u2014 done. Made `WasmWordRepr` `pub` and threaded it through:

- `type_wasm_native_f64_option!(isize as WasmWordRepr, usize as WasmWordRepr)` replaces \~120 lines of manual cfg-gated impls
- `*const T` / `*mut T` / `NonNull<T>` now use `Abi = WasmWordRepr` directly
- All `target_arch = "wasm64"` cfgs in `src/convert/impls.rs` are gone (only the declaration of `WasmWordRepr` itself remains, in `src/rt/mod.rs`)

Net: -196/+16 lines in `impls.rs`. Reference snapshots unchanged \u2014 confirming the macro produces identical codegen on both targets.

Left `Option<*const T>` / `Option<*mut T>` as `Abi = f64` matching the existing convention for other pointer-sized Options (`Option<u32>` etc. already use the f64 sentinel).

src/lib.rs
```diff
@@ -1139,34 +1139,61 @@ impl<T: TryFromJsValue> TryFromJsValue for Option<T> {
...
+            *self == (*other as u32)
+        }
     }
 }
 
 impl From<usize> for JsValue {
     #[inline]
     fn from(n: usize) -> Self {
-        Self::from(n as u32)
+        #[cfg(target_arch = "wasm64")]
```
RReverser: We could probably reuse `Self::from(n as WasmWordRepr)` here too instead of cfg branches.

guybedford: Done. Both `From` and `PartialEq` impls for `usize` / `isize` collapse to single non-cfg bodies via `Self::from(n as WasmWordRepr)` (and `WasmSignedWordRepr` for `isize` so negative values sign-extend correctly when widening to `f64` on wasm64).

Note: I also caught a related bug along the way \u2014 my first attempt had `isize as WasmWordRepr` (which is unsigned u32 on wasm32), causing `-1isize` to round-trip through `f64` as `4294967295.0` instead of `-1.0` and breaking `optional_primitives::works` in CI. Added `WasmSignedWordRepr` (`i32` on wasm32, `f64` on wasm64) for the signed lowering path.

PR #5005, opened 2026-02-26
Add termination detection for panic=unwind builds

hoodmane:
I exported a `__instance_terminated` global from the Rust runtime. We set it if we catch an exception thrown out of wasm which is not wrapped in `__wbindgen_wrapped_jstag`.

To check entering into wasm from JS we add `__wbg_termination_guard` JS function that checks the `__instance_terminated` flag and if it is set, throw a "Module terminated" error.

To check returning from JS to wasm, I updated catch_handler.rs. It now checks the termination guard on all three backwards edges: if the import returned normally, if it threw a JS exception, or if it threw some other error.

One unfortunate aspect of this is that is loses the original exception and replaces it with a `WebAssembly.RuntimeError: unreachable`. As an improvement we could store the original exception as context and retrieve it in `__wbg_handle_catch`.

On top of #5001.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 
 ### Added
 
+* Added termination detection for `panic=unwind` builds. When a non-JS exception (e.g. a Rust
+  panic) escapes from Wasm, the instance is marked as terminated and subsequent calls from JS
+  into Wasm will throw a `Module terminated` error instead of re-entering corrupted state.
+  [#5005](https://github.com/wasm-bindgen/wasm-bindgen/pull/5005)
+
 * Added `unchecked_optional_param_type` attribute for marking exported function parameters as
   optional in TypeScript (`?:`) and JSDoc (`[paramName]`) output. Mutually exclusive with
   `unchecked_param_type`. Required parameters after optional parameters are rejected at compile time.


crates/cli-support/src/js/mod.rs
```diff
@@ -3275,6 +3275,30 @@ if (require('worker_threads').isMainThread) {{
...
+        ));
+
+        // Create a helper function to unwrap and rethrow wrapped JS exceptions
+        self.global(&format!(
+            "\
+function __wbg_handle_catch(e) {{
+    if (e instanceof WebAssembly.Exception && e.is(__wbindgen_wrapped_jstag)) {{
+        throw e.getArg(__wbindgen_wrapped_jstag, 0);
+    }}
+    {mem_view}()[wasm.__instance_terminated/4] = 1;
```
guybedford: `__instance_terminated.value`?

```suggestion
    {mem_view}()[wasm.__instance_terminated.value / 4] = 1;
```

crates/cli-support/src/js/mod.rs
```diff
@@ -3275,6 +3275,30 @@ if (require('worker_threads').isMainThread) {{
             "const __wbindgen_wrapped_jstag = new WebAssembly.Tag({ parameters: ['externref'] });",
         );
 
+        let memory = wasm_conventions::get_memory(self.module).unwrap();
+        let mem_view = self.expose_int32_memory(memory);
+
+        self.global(&format!(
+            "\
+function __wbg_termination_guard() {{
+    if ({mem_view}()[wasm.__instance_terminated/4]) {{
```
guybedford: `__instance_terminated.value`?


```suggestion
    if ({mem_view}()[wasm.__instance_terminated.value / 4]) {{
```

guybedford: Also would it make sense to capture the global as a local to avoid the extra object lookup cost?

guybedford (commented):
This seems correct to me, apart from the one small issue with the global.value check missing. I'm a little suprised if tests were passing with the missing `.value` check, would be nice to know there's coverage of that case.

hoodmane:
> I'm a little suprised if tests were passing with the missing .value check, would be nice to know there's coverage of that case.

Globals coerce to number:
```js
new WebAssembly.Global({value: "i32"}, 7) + 1 // 8
```
Added tests definitely cover this.

guybedford (approved):

PR #5014, opened 2026-02-28
feat: add WebDriver screenshot support for browser tests

Madoshakalaka:
### Description

Add a `screenshot()` API that lets `#[wasm_bindgen_test]` browser tests request the headless runner to take a WebDriver screenshot and save it as a PNG file.

### Background

I'm a maintainer of [Yew](https://github.com/yewstack/yew), where we have ~300 integration tests that run via `wasm-bindgen-test` in headless Chrome and Firefox. All of these tests verify rendering through `inner_html()` string comparisons, there is currently no way to verify *visual* output.

This became a real problem when we hit SVG rendering bugs ([yewstack/yew#3875](https://github.com/yewstack/yew/issues/3875),
  [yewstack/yew#3892](https://github.com/yewstack/yew/issues/3892)) where the DOM markup looked correct but the browser rendered a blank screen because element names like `feDropShadow` were silently lowercased to `fedropshadow`. An `inner_html()` test cannot catch this. Only a screenshot can confirm the SVG actually renders.

We've explored other e2e options like cypress and playwright and multiple maintainers don't entertain the thought of adding Javascript dependencies.

The root issue is that screenshots must be taken from *outside* the browser (via WebDriver), but `wasm-bindgen-test-runner` only uses WebDriver for navigation and text polling. This PR adds the missing plumbing.

### Design

The WASM test and the headless runner communicate through a hidden DOM element (`#__wbgtest_screenshot`):

1. Test calls `wasm_bindgen_test::screenshot("path.png").await`
2. This writes `"path.png"` into the hidden DOM element and polls until it's cleared
3. The runner's existing polling loop detects the non-empty element via `execute_script`
4. Runner calls `GET /session/{id}/screenshot` (standard WebDriver endpoint), decodes the base64 PNG, writes the file, and clears the element
5. The WASM side sees the element is cleared and returns

No new runtime dependencies are added to `wasm-bindgen-test`. The only new dependency is `base64` in the CLI crate (for decoding the WebDriver screenshot response).

### Checklist


crates/test/src/rt/browser.rs
```diff
@@ -35,6 +35,46 @@ extern "C" {
...
+pub async fn screenshot(path: &str) {
+    let el = DOCUMENT.with(|doc| doc.getElementById("__wbgtest_screenshot"));
+    el.set_text_content(path);
+
+    loop {
+        wasm_bindgen_futures::JsFuture::from(delay_promise(50))
+            .await
+            .unwrap_throw();
+
+        if el.text_content().is_empty() {
```
drewcrawford: I suspect there is a race condition in here.  When running two tests simultaneously, they compete over the slot?

crates/test/src/rt/browser.rs
```diff
@@ -35,6 +35,46 @@ extern "C" {
...
+/// This works by writing the requested filename to a hidden DOM element
+/// (`#__wbgtest_screenshot`). The headless test runner detects this, takes a
+/// screenshot via the WebDriver protocol, saves it, and clears the element.
+/// This function polls until the element is cleared, then returns.
+///
+/// The path is relative to the crate root (where `cargo test` is run).
+///
+/// # Panics
+///
+/// Panics if the `#__wbgtest_screenshot` element is not present in the page
```
drewcrawford: My suspicion is, when the element is not present, getElementById returns a JSNull wrapped as Element.  Then `set_text_content` throws a JS exception.  Would it be better to use `web_sys::Document::get_element_by_id` here?

crates/cli/src/wasm_bindgen_test_runner/headless.rs
```diff
@@ -187,6 +188,40 @@ pub fn run(
...
+                    Ok(png_data) => {
+                        let path = Path::new(&ss_text);
+                        if let Some(parent) = path.parent() {
+                            if !parent.as_os_str().is_empty() {
+                                fs::create_dir_all(parent).ok();
+                            }
+                        }
+                        match fs::write(path, &png_data) {
+                            Ok(()) => {
+                                println!("Screenshot saved: {ss_text} ({} bytes)", png_data.len())
```
drewcrawford: Should these use shell.status?

crates/test/src/rt/browser.rs
```diff
@@ -35,6 +35,46 @@ extern "C" {
...
+/// (i.e. when not running under the headless test runner).
+pub async fn screenshot(path: &str) {
+    let el = DOCUMENT.with(|doc| doc.getElementById("__wbgtest_screenshot"));
+    el.set_text_content(path);
+
+    loop {
+        wasm_bindgen_futures::JsFuture::from(delay_promise(50))
```
drewcrawford, line 101: I guess in the worst case we spend 150ms waiting for polling (this 50, plus 100 on the runner).  Mostly the runner poll time hasn't been a big problem, but I wonder if taking a lot of screenshots will add up.

crates/test/src/rt/browser.rs
Madoshakalaka: I added a wait for any in-progress screenshot request to complete before claiming the slot

crates/test/src/rt/browser.rs
Madoshakalaka: Good point. The WASM-side poll is just reading `textContent` on a hidden element, so it's essentially free. I reduced it from 50ms to 10ms. so it's somewhat reduced to 110ms now in the worst case.

I also think in practice, the WebDriver screenshot call itself (network round-trip, rendering, base64 encoding) likely dominates, maybe 50ms? 200ms? not sure. so the polling overhead should be a small fraction of the total per-screenshot cost.

crates/cli/src/wasm_bindgen_test_runner/headless.rs
Madoshakalaka: It's intentionally a persistent message here instead of shell.status() for transient progress messages

line 202 already uses `shell.status()` for the "Taking screenshot: ..." indicator, which I think is the right split.

crates/test/src/rt/browser.rs
Madoshakalaka: Good catch. I changed the local `getElementById` to return `Option<Element>` and added `.expect()` with a descriptive message at both its call sites. Now a missing element produces a clear Rust panic instead of a cryptic JS exception.

(web-sys is not available, only a dev-dependency)

crates/cli/src/wasm_bindgen_test_runner/headless.rs
WorldSEnder, line 193: Though not in scope of this PR strictly, I remember WebDriver BiDi being a thing that could enable better handling here.

[W3C spec](https://www.w3.org/TR/webdriver-bidi/) and [browser compat data](https://developer.mozilla.org/en-US/docs/Web/WebDriver/Reference/BiDi/Modules)

Specifically, this could offer a better way to establish a channel from browser to agent without polling overhead and incidentally offers [screenshots of individual elements](https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot).

Madoshakalaka: wow :thinking: 
maybe I'll investigate this in a different PR

Madoshakalaka: Safari doesn't support it at all
https://wpt.fyi/results/webdriver/tests/bidi

crates/test/src/rt/browser.rs
drewcrawford: Could we be missing a push here?  Not seeing this change in the PR.

crates/cli/src/wasm_bindgen_test_runner/headless.rs
drewcrawford: My information is that bidi is not on safari's roadmap, so we're probably stuck with webdriver classic for awhile, at least as a baseline.

Once in awhile I look into using bidi as an optimization on browsers that support it, but so far haven't seen the uplift to justify it.  It's conceivable to me if we get this in and that's the bottleneck, somebody will come along to PR bidi on a browser-by-browser basis.

crates/test/src/rt/browser.rs
Madoshakalaka: accidentally overwritten the fix in my rebase :skull: 

fixed now

crates/test/src/rt/browser.rs
drewcrawford: I think this fixes some races but not others, as clearing the div currently means 'request consumed' as opposed to 'request complete'.  Consider this test shape

```rust
#[wasm_bindgen_test] async fn mytest() {
          setup_state_for_a();
          screenshot("a.png").await;
          setup_state_for_b();
          screenshot("b.png").await;
}
```

A possible execution flow is:

1.  `setup_state_for_a()`
2. `screenshot()` finds `__wbgtest_screenshot` is clear, so writes `a.png` in the slot
3. Runner sees `__wbgtest_screenshot` is `a.png`, so clears the slot first, then begins taking a screenshot
4. `screenshot()` now sees `__wbg_test_screenshot` is clear again, so returns
5. test proceeds through `setup_state_for_b()`
6. Meanwhile, the screenshot completes and we screenshot B's dom into `a.png`

I think it fixes this race is to move the clear in step 3, to after step 6.  But then there is this flow:

1.  `setup_state_for_a()`
2. `screenshot()` finds `__wbgtest_screenshot` is clear, so writes `a.png` in the slot
3. Runner sees `__wbgtest_screenshot` is `a.png`, then begins taking a screenshot
4. The screenshot ultimately fails somehow
5. The test succeeds?

This raises a design question: are these screenshots advisory, or are they something that should fail the test in step 5?  Because we treat them as test failures sometimes (e.g. panic in the headless runner) but differently here.

Failing consistently is the option that makes more intuitive sense to me, but that won't fit into a 1-bit empty/clear protocol, we'd need more like ready/progress/failed.  So that would require a slightly different solution to the race.

crates/test/src/rt/browser.rs
```diff
@@ -35,11 +35,55 @@ extern "C" {
...
+/// The path is relative to the crate root (where `cargo test` is run).
+///
+/// # Panics
+///
+/// Panics if the `#__wbgtest_screenshot` element is not present in the page
+/// (i.e. when not running under the headless test runner).
+pub async fn screenshot(path: &str) {
+    let el = DOCUMENT
+        .with(|doc| doc.getElementById("__wbgtest_screenshot"))
+        .expect("#__wbgtest_screenshot element not found; are you running under the headless test runner?");
```
drewcrawford: Another thing I'm thinking about is how to make tests portable across headless and non-headless runners.  Working around the panic requires a device like this

```rust
if is_screenshot_available() { //poke around for __wbgtest_screenshot internals?
     screenshot("path").await;
}
```

1.  If screenshots are advisory, we should log and return here – then it's portable by default and this error gets surfaced by `--nocapture`
2. If screenshots are really part of the test, this should be a case on a Result<T,Error>.  Callers can slap an `ok()` on there to be portable, or filter out only the unsupported errors vs file errors, but relying on our detection logic for "supported".  Then if we support more platforms or change mailbox design, updates go smoothly.

crates/test/src/rt/browser.rs
Madoshakalaka: I updated it to a 3-state protocol:

1. Wasm writes the path to the element
2. Runner sees the path, takes the screenshot, saves the file
3. Runner writes `"ok"` or `"err:<message>"` to the element (only after everything is done)
4. Wasm polls until the content changes from the path it wrote, reads the result, clears the element, and returns `Result<(), ScreenshotError>`

The runner's JS filter also now skips `"ok"` and `"err:..."` values so it won't re-process its own responses.

On the design question: screenshots are now treated as fallible operations. The runner signals errors back through the element, and the wasm side surfaces them via `ScreenshotError::Failed(msg)`. So step 5 in your second scenario would propagate the failure to the test (if the caller uses `?` or `.unwrap()`).

crates/test/src/rt/browser.rs
Madoshakalaka: I now made `screenshot()` return `Result<(), ScreenshotError>` instead of panicking. Two variants:

- `ScreenshotError::NotSupported`: the `#__wbgtest_screenshot` element is absent (not running under the headless runner)
- `ScreenshotError::Failed(msg)`: the runner tried but something went wrong (capture failed, file write failed, etc.)

So portable tests can do:

```rust
wasm_bindgen_test::screenshot("snap.png").await.ok();
```

And tests that require the screenshot can do:

```rust
wasm_bindgen_test::screenshot("snap.png").await.unwrap();
```

If we add support for more platforms later, only the internals change; the `Result` API stays the same.

crates/test/src/rt/browser.rs
drewcrawford: I think the race is fixed.  The problem now is, a string like `ok` is also a valid filename.

A simple hack to solve this is to give filenames a well-known prefix like `filename:ok`, freeing up plain `ok` for 3-state protocol.

crates/test/src/rt/browser.rs
drewcrawford: I like this design a lot.  My only comment is it seems more future-proof to have

`-> Result<Screenshot, ScreenshotError>` where `pub struct Screenshot;`.  The reason for this ceremony being, I can foresee wanting to inspect the screenshot data from inside the test, so introducing a newtype now buys us a lot of flexibility to build on this feature down the road.

crates/test/src/rt/browser.rs
```diff
@@ -35,11 +35,89 @@ extern "C" {
...
+}
+
+fn delay_promise(millis: i32) -> js_sys::Promise {
+    js_sys::Promise::new_typed(&mut |resolve, _| {
+        set_timeout(resolve.into(), millis);
+    })
+}
+
+/// Error returned by [`screenshot`].
+#[derive(Debug)]
```
drewcrawford, line 55: Side note, might be best to add `#[non_exhaustive]` here just for future-proofing

crates/test/src/rt/browser.rs
Madoshakalaka: Done. Wasm now writes `filename:{path}` to the element, and the runner's JS filter only acts on values starting with `filename:`. Plain `ok` and `err:...` are reserved for responses.

crates/test/src/rt/browser.rs
Madoshakalaka: Done. `screenshot()` now returns `Result<Screenshot, ScreenshotError>` with `pub struct Screenshot;`. Re-exported from the crate root alongside `ScreenshotError`.

crates/test/src/rt/browser.rs
Madoshakalaka: Done. Added `#[non_exhaustive]` to `ScreenshotError`.

Madoshakalaka:
The `msrv-cli` CI failure is unrelated to this PR. `clap_builder` 4.6.0 was published requiring `edition2024`, which breaks the Rust 1.82 MSRV check. Pinning `clap_builder = "=4.5.60"` in `crates/msrv/cli/Cargo.toml` fixes it. It's for a different PR ig.

Madoshakalaka:
I added some guide in [9ea7967](https://github.com/wasm-bindgen/wasm-bindgen/pull/5014/commits/9ea7967670f36088333091071cbbc8044ad1e1fd)

> you use this in unit tests is to save screenshot files and check if there are diffs from previous screenshots at the binary level?

Our intended usage is very simple. We had multiples bugs that used to cause svgs to be invisible. This is the testing code from a branch of Yew experimenting with screenshots but couldn't eventually do it without wasm-bindgen support (we render a `<rect/>`, sample a grid of pixels and assert anything renders at all by asserting the pixels are not all white):

https://github.com/yewstack/yew/blob/b6de2dd20419942658af542b60f8c49bfebc109e/packages/yew/tests/svg_render.rs#L166-L189

@drewcrawford mentions a normal use case too I think. It's usually very easy to set up screenshot and video recording for e2e tools in CI. This has been a gap in wasm-bindgen-test I think.

guide/src/wasm-bindgen-test/screenshots.md
```diff
@@ -0,0 +1,133 @@
+# Screenshots
```
guybedford: Can we rather just add this as a short section in the existing `browsers.md` section?

guide/src/wasm-bindgen-test/screenshots.md
```diff
@@ -0,0 +1,133 @@
...
+The screenshot feature uses a hidden DOM element (`#__wbgtest_screenshot`) as a
+communication channel between the Wasm test and the headless test runner:
+
+1. The test writes `filename:<path>` to the element.
+2. The test runner polls the element, detects the request, and uses the
+   WebDriver screenshot API to capture the page as a PNG.
+3. The runner saves the PNG to the requested path and writes `ok` (or
+   `err:<message>`) back to the element.
+4. The test polls until it sees the response, then clears the element and
+   returns the result.
```
guybedford: These feel like implementation details that are not relevant to users of the API to me.

Madoshakalaka: makes sense! deleted

guide/src/wasm-bindgen-test/screenshots.md
Madoshakalaka: :+1: 
done

PR #5018, opened 2026-03-03
Delete .vscode/settings.json

RReverser:
### Description
It's simultaneously in gitignore, yet was somehow (force-?)committed in https://github.com/wasm-bindgen/wasm-bindgen/pull/4876 which messes up local configurations and development.

### Checklist

### DIFF
--- a/.vscode/settings.template.json
+++ b/.vscode/settings.template.json
@@ -1,0 +1,12 @@
+// This is a template you can use as a starting point for your own settings.json file.
+// Note that wasm32-unknown-unknown target is not compatible with all crates in the repo,
+// so you might have to adjust these depending on what you are workign on.
+{
+  "rust-analyzer.cargo.target": "wasm32-unknown-unknown",
+  "rust-analyzer.cargo.cfgs": [
+    "panic=unwind"
+  ],
+  "rust-analyzer.cargo.extraEnv": {
+    // "RUSTFLAGS": "--cfg=js_sys_unstable_apis"
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ pnpm-lock.yaml
 yarn.lock
 /publish
 /publish.exe
-.vscode
+/.vscode/settings.json
 webdriver.json
 benchmarks/pkg
 /crates/msrv/*/target


guybedford:
I added this very specifically with the comment for the unstable types as well. When writing js-sys tests on the new unstable interfaces, it is invaluable to have proper Rust analyzer support.

RReverser:
That comment is fine but could be just in "contributing"?

Or we can leave just that comment, it's the `"rust-analyzer.cargo.target": "wasm32-unknown-unknown",` that is problematic because by default in VSCode Rust-Analyzer tries to build the whole workspace, and then you get tons of noisy errors about tokio and other dependencies of CLI being incompatible with Wasm.

The host target for IDEs is a better / less painful default.

RReverser:
To be clear, I'm talking about this problems view that one sees by default after the config was added:
image

guybedford:
Ideally I think we should fix this configuration then rather. Do we need to set the cfg option to ensure getrandom behaves correctly?

Spxg:
The issue here is that both the native crate and the wasm crate are in the same workspace. Perhaps we could split the workspace and let rust-analyzer use different configurations?

RReverser:
> Ideally I think we should fix this configuration then rather. Do we need to set the cfg option to ensure getrandom behaves correctly?

It's not just cfg, bunch of crates that CLI and proc macro depend on are fundamentally incompatible with Wasm. 

> The issue here is that both the native crate and the wasm crate are in the same workspace. Perhaps we could split the workspace and let rust-analyzer use different configurations?

Yeah we talked about it in the past, refactoring Cargo workspaces would be the only option to get default target - and then we could/should set default target in Cargo config not in VSCode specific settings.

It's a pretty big refactoring though and for now I'd rather fix the immediate problem.

Using the default target has worked fine for IDE development all these years and helps to catch usual syntax and type errors, let's just go back to that. @guybedford you can still have this custom config locally.

RReverser:
@guybedford I've re-added it as `settings.template.json` - this way it doesn't override target by default and doesn't conflict with any custom workspace settings each developer might want to set locally but not push to the repo, yet can be used as a template to copy from.

Does this work as a compromise?

.vscode/settings.template.json
```diff
@@ -0,0 +1,12 @@
+// This is a template you can use as a starting point for your own settings.json file.
+// Note that wasm32-unknown-unknown target is not compatible with all crates in the repo,
+// so you might have to adjust these depending on what you are workign on.
```
guybedford, line 3:
```suggestion
// so you might have to adjust these depending on what you are working on.
```

PR #5024, opened 2026-03-05
Support __instance_terminated on panic=abort

logan-gatlin:
## Summary
Emit `__instance_terminated` on `panic=abort`, previously only supported for `panic=unwind`. Modules with aborting panics are assumed to still be valid even after an abort - it is only after garbage collection we can say they are dead for certain. This PR writes 1 to `__instance_terminated` from a finalization registry.
## Changes
- Removed #[cfg(panic = "unwind")] from __instance_terminated static so it is compiled unconditionally for all panic strategies.
crates/cli-support/src/js/mod.rs
- Added instance_termination_mem: Option<String> field to Context to track whether the registry was generated and store the resolved memory export name.
- Added generate_instance_termination_registry() method that emits a FinalizationRegistry singleton rooted on globalThis (with typeof guard fallback for environments without it) and a __wbg_register_instance(instance) helper. The registry callback writes 1 to the __instance_terminated address in linear memory when the Instance is collected.
- Added __wbg_register_instance(...) calls in all 8 loading modes:
  - Web / NoModules (__wbg_finalize_init)
  - Module (generate_module_wasm_loading)
  - Deno (generate_deno_wasm_loading)
  - Node ESM with and without threads (generate_node_esm_wasm_loading)
  - Node CJS with and without threads (generate_node_cjs_wasm_loading)
  - Bundler (__wbg_set_wasm accepts optional second instance parameter)
- Added registration call in generate_reset_state() after new instance creation.
- Added instance_gc_termination integration test that:
  - Builds a minimal panic=abort fixture (no nightly required)
  - Verifies the generated JS contains the registry and globalThis rooting
  - Directly instantiates the wasm module, confirms __instance_terminated is exported and starts at 0, registers with a FinalizationRegistry, drops the Instance, forces GC via --expose-gc with allocation pressure, and asserts the flag becomes 1 while memory remains accessible
### Checklist

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -103,6 +103,10 @@ pub struct Context<'a> {
     /// If exception handling / unwinding is enabled.
     unwind_enabled: bool,
 
+    /// The resolved memory export name for the instance termination
+    /// FinalizationRegistry. `Some` when the registry has been generated.
+    instance_termination_mem: Option<String>,
+
     /// Mapping from qualified name (used in WasmDescribe) to rust_name (used as exported_classes key).
     pub(crate) qualified_to_rust_name: HashMap<String, String>,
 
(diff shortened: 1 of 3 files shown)


logan-gatlin:
Better to do this in the runtime

PR #5026, opened 2026-03-07
Support per-operation [WbgGeneric] in WebIDL codegen

guybedford:
## Summary

- Add support for `[WbgGeneric]` on individual WebIDL operations (not just interface/dictionary/namespace definitions), so that specific methods can opt into typed generics without affecting all sibling methods.
- Fix return type `WbgType` expansion to respect `wbg_generic` for callback-like return types. Previously `next_unstable` was only set temporarily during argument expansion but not when computing the return type in `create_method`.
- Add `[WbgGeneric]` to the two `createImageBitmap` operations in `WindowOrWorkerGlobalScope`, restoring `Promise<ImageBitmap>` return types on `Window` and `WorkerGlobalScope`.

## Motivation

When `VideoFrame` was stabilized in ce535e1a9, the `createImageBitmap` overloads accepting `VideoFrame` on `Window`/`WorkerGlobalScope` lost their typed generic return type (`Promise<ImageBitmap>` became bare `Promise`). Previously these methods were unstable (gated behind `web_sys_unstable_apis`) which automatically gave them generics. After stabilization, the parent `WindowOrWorkerGlobalScope` mixin has no `[WbgGeneric]`, so generics were lost.

Adding `[WbgGeneric]` to the entire mixin was too broad — it changed `setTimeout`, `setInterval`, `fetch`, etc. Per-operation support allows targeting only the affected methods.

## Changes

**`crates/webidl/src/util.rs`**:
- Check `Signature.attrs` for `[WbgGeneric]` and OR it with the parent definition's `wbg_generic` flag in two places: when setting `next_unstable` for argument type expansion, and when passing `wbg_generic` to `create_method`.
- In `create_method`, temporarily set `next_unstable` around the return type `to_wbg_type` call when `wbg_generic` or `unstable_flag` is true, so that callback-like return types also get typed expansion.

**`crates/web-sys/webidls/enabled/WindowOrWorkerGlobalScope.webidl`**: Add `WbgGeneric` to the two `createImageBitmap` operation attributes.

**`crates/web-sys/src/features/gen_Window.rs`**, **`gen_WorkerGlobalScope.rs`**: Regenerated — `createImageBitmap` overloads now return `Promise<ImageBitmap>`.

### DIFF
--- a/crates/webidl/src/lib.rs
+++ b/crates/webidl/src/lib.rs
@@ -797,7 +797,7 @@ impl<'src> FirstPassRecord<'src> {
                 catch: x.catch,
                 variadic: x.variadic,
                 unstable: false,
-                wbg_generic,
+                wbg_generic: x.wbg_generic,
             });
         }
     }
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -361,8 +361,12 @@ impl<'src> FirstPassRecord<'src> {
         for signature in data.signatures.iter() {
             // Signatures from unstable IDL definitions always use typed generics
             // for WbgType expansion (callbacks become typed, etc.)
-            // [WbgGeneric] on the definition also opts into typed generics.
-            if unstable || wbg_generic || signature.stability.is_unstable() {
+            // [WbgGeneric] on the definition or operation also opts into typed generics.
+            if unstable
+                || wbg_generic
+                || signature.stability.is_unstable()
+                || is_wbg_generic(signature.attrs.as_ref())
+            {
                 self.options.next_unstable.set(true);
             }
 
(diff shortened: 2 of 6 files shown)


guybedford:
//cc @evilpie added a correction to the generation here

crates/webidl/src/lib.rs
drewcrawford: Is this still the older non-specific op flag?

crates/webidl/src/util.rs
```diff
@@ -765,7 +772,7 @@ impl<'src> FirstPassRecord<'src> {
                     container_attrs,
                     unstable_flag,
                     false,
-                    wbg_generic,
+                    op_wbg_generic,
```
drewcrawford, line 782: Does this fully cover callback-like return types? It looks like the temporary next_unstable override is only active during signature expansion, while return WbgType is computed later in create_method, after the flag has been restored.

guybedford: Since no stable methods with `[WbgGeneric]` currently return callbacks this is not yet an issue, but will be if that happens and will need to be refactored then, you're right.

guybedford: Ok added a change for this as well. Not pretty but should do the trick.

crates/webidl/src/lib.rs
guybedford: wbg_generic was added already previously, this just extends it to per-op.

drewcrawford: Right, wbg_generic itself is pre-existing. My question is whether this namespace path now needs to preserve the per-op value coming back from create_imports. x can now differ per signature/op, but here we still assign the enclosing wbg_generic instead of x.wbg_generic. Is that intentional for namespaces?

guybedford: Good point, we should respect the per-op value here yes.

guybedford:
Ping @drewcrawford @RReverser for reviews here.

PR #5029, opened 2026-03-07
Improve ? support in JsError under no_std

### LINKED ISSUE #5025: JsError should make ? work under no_std
There is a [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) trait but `JsError` has an artificial restriction to std:
```
impl<E> From<E> for JsError where E: std::error::Error,
```

burdges:
Fixes https://github.com/wasm-bindgen/wasm-bindgen/issues/5025

### Description
Uses alloc::strict::ToString.  Uses core::error::Error because alloc:error does not exist.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Changed
 
+* Fix ? support for JsError under no_std
+  #5029[](https://github.com/wasm-bindgen/wasm-bindgen/pull/5029)
+
 * Replaced per-closure generic destructors with a single `__wbindgen_destroy_closure`
   export.
   [#5019](https://github.com/wasm-bindgen/wasm-bindgen/pull/5019)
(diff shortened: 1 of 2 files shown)


burdges:
Appears core::error has not become stable yet:
https://github.com/rust-lang/rust/issues/103765 

We could conditionally set error_in_core but I guess that's obnoxious:
https://docs.rs/cfg_rust_features/latest/cfg_rust_features/

jsdw:
`core::error::Error` is stable now [since rust 1.81](https://doc.rust-lang.org/stable/core/error/index.html) so this should be ok!

burdges:
Alright thanks, so that's why it just worked for me.  :) 

I suppose the maintainers will choose this whenever they upgrade their rustc in CI.

guybedford:
See the MSRV policy here. It seems this could land in September though.

guybedford (approved):

PR #5030, opened 2026-03-08
fix: JsOption soundness

guybedford:
Resolves https://github.com/wasm-bindgen/wasm-bindgen/issues/5022.

Just the basic fix for now. Unfortunately it's not possible to implement `UpcastFrom<primitive num> for JsOption<Number>` due to orphan rule shenanigans...

//cc @RReverser

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@
 * Fixed large test outputs (10MB+) causing oversized WebDriver responses that were either
   extremely slow or crashed completely, by switching to incremental streaming output collection.
   [#4960](https://github.com/wasm-bindgen/wasm-bindgen/pull/4960)
+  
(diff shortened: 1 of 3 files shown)


src/sys.rs
RReverser: Could you replace these with Upcast or JsCast methods to make compiler check the soundness through existing bounds?

guybedford: We can't actually use `upcast_into()` because it's not actually correct in the JS type system to upcast a type into Option<T> even if that is a valid transmute.

But we can do this with `JsCast`. `JsGeneric` doesn't currently include the `JsCast` trait, but I've gone ahead and added that to it as well so this now works.

RReverser: Perfect, thanks!

RReverser (approved):

PR #5031, opened 2026-03-08
support Promise<T> union expansions

guybedford:
### Description
This adds support for `Promise<T>` arguments to be expanded as if they were a union `T | Promise<T>`.

This is how WebIDL supports all promise arguments based on their coercion going through Promise resolve functionality.

This expansion is only performed under generic expansion.

Then, we add a new wrapper type `CanonicalValue` which we wrap the Promise expansion with to ensure that we don't call all promise args `_promise` and instead disambiguate the non-promise argument instead during naming.

A function `foo` taking a Promise arg `Promise<string>` thus gets two variants - `foo(Promise)` and `foo_str(string)`, with the canonical naming reserved for the promise form.

### Checklist

### DIFF
--- a/crates/webidl/src/util.rs
+++ b/crates/webidl/src/util.rs
@@ -620,17 +620,28 @@ impl<'src> FirstPassRecord<'src> {
                 // CanonicalValue args are the "primary" overload form and
                 // don't contribute to the method name suffix unless a
                 // shorter signature exists (e.g. from optional params
-                // producing a no-arg variant) that would collide.
-                // When they do need disambiguation, always use the arg name
-                // since their type name is intentionally empty.
+                // producing a no-arg variant) that would collide, OR
+                // another sibling is also a CanonicalValue at the same
+                // position (e.g. Promise<DOMString> and Promise<Blob> both
+                // expand to CanonicalValue branches that would otherwise
+                // produce the same unsuffixed name).
                 if is_canonical {
-                    // Only need suffix if there's a shorter signature that
-                    // would produce the same unsuffixed name.
+                    let inner = match arg {
+                        WbgType::CanonicalValue(inner) => inner.as_ref(),
+                        _ => unreachable!(),
+                    };
+                    // Need a suffix if there's a shorter signature (no-arg
+                    // variant) OR another sibling CanonicalValue at position i.
                     let any_shorter = disambiguate_against.iter().any(|&idx| {
                         let other = &all_signatures[idx];
                         signature != other && other.args.get(i).is_none()
                     });
-                    if !any_shorter {
+                    let any_other_canonical = disambiguate_against.iter().any(|&idx| {
+                        let other = &all_signatures[idx];
+                        signature != other
+                            && matches!(other.args.get(i), Some(Some(WbgType::CanonicalValue(_))))
+                    });
+                    if !any_shorter && !any_other_canonical {
                         continue;
                     }
                     if first {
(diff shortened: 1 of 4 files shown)


crates/webidl/src/util.rs
```diff
@@ -584,6 +588,12 @@ impl<'src> FirstPassRecord<'src> {
                 .enumerate()
                 .filter_map(|(i, ty)| ty.as_ref().map(|ty| (i, ty)))
             {
+                // CanonicalValue args are the "primary" overload form and never
+                // contribute to the method name suffix.
+                if matches!(arg, WbgType::CanonicalValue(_)) {
```
drewcrawford: Suppose I have

```webidl
[WbgGeneric]
interface Example {
  undefined foo(optional Promise<DOMString> p);
};
```

I think I get

```rust
fn foo(&self);
fn foo(&self, p: &js_sys::Promise<js_sys::JsString>); //error[E0428]: the name `foo` is defined multiple times
fn foo_with_str(&self, p: &str);
```

crates/webidl/src/lib.rs
drewcrawford: Nit, but suppose I have

```webidl
[WbgGeneric]
interface Example {
  attribute Promise<DOMString> value;
};
```

I think ext is empty, so

```
set_value_str
set_value_
```
trailing underscore is a bit odd

crates/webidl/src/util.rs
guybedford: Edit: I'll check we've got this tested.

guybedford:
Thanks for the review @drewcrawford! Both issues are addressed in 83510c141:

**Optional Promise\<T\> naming collision (comment 1):** `CanonicalValue` args in `compute_rust_name` now check whether a shorter signature (from optional params) would collide before skipping the suffix. When disambiguation is needed, the arg name is used. For `optional Promise<DOMString> p`, the result is:

- `foo()` — no-arg variant (unsuffixed)
- `foo_with_p(&Promise)` — promise variant (suffixed with arg name)
- `foo_with_str(&str)` — value variant

**Trailing underscore on attribute setters (comment 2):** The attribute setter naming in `lib.rs` now skips appending `_` + suffix when `push_snake_case_name` produces an empty string (as it does for `CanonicalValue`). For `attribute Promise<DOMString> value`, the result is:

- `set_promise_value(&Promise)` — canonical setter (unsuffixed)
- `set_promise_value_str(&str)` — value setter

Added tests for both cases: `optional Promise<DOMString>` method overloads and `Promise<DOMString>` attribute setters.

crates/webidl/src/util.rs
```diff
@@ -610,6 +616,33 @@ impl<'src> FirstPassRecord<'src> {
...
+                // When they do need disambiguation, always use the arg name
+                // since their type name is intentionally empty.
+                if is_canonical {
+                    // Only need suffix if there's a shorter signature that
+                    // would produce the same unsuffixed name.
+                    let any_shorter = disambiguate_against.iter().any(|&idx| {
+                        let other = &all_signatures[idx];
+                        signature != other && other.args.get(i).is_none()
+                    });
+                    if !any_shorter {
```
drewcrawford: In Clipboard.webidl we have this

```webidl
typedef Promise<(DOMString or Blob)> ClipboardItemData;
```

Suppose I use that directly:

```webidl
[WbgGeneric]
undefined foo(ClipboardItemData p);
```

This expands to two canonical promise branches, plus two direct-value branches. As written, the two canonical promise branches still look like they both keep the Rust name `foo(...)`, since this block only disambiguates against shorter optional-arg forms. Should those `Promise<_>` branches be disambiguated from each other as well?

guybedford: GREAT catch - updated to handle this case.

PR #5032, opened 2026-03-08
feat: custom builtin extensions via webidl, with custom thenables

guybedford:
### Description
This allows the creation of WebIDL interfaces that extend JS builtin types to support deref and extends functionality.

In addition, by extending `Promise`, it is possible to define a custom thenable which also then auto-implements the `Promising` trait correctly.

A test is included that demonstrates a custom thenable Promise subclass with custom methods that can be used in `Promising<T>` positions for the return value type of its `then()` method.

### Checklist

### DIFF
--- a/crates/msrv/cli/Cargo.toml
+++ b/crates/msrv/cli/Cargo.toml
@@ -24,6 +24,11 @@ rouille = "=3.6.2"
 # tempfile 3.25+ depends on getrandom 0.4+ which requires edition 2024 (Rust 1.85+)
 tempfile = "=3.24.0"
 
+# clap 4.6.0 raises the MSRV to Rust 1.85, and clap_builder 4.5.61 will
+# otherwise float clap_lex to 1.1.0 which also requires Rust 1.85
+clap = { version = "=4.5.61", features = ["derive"] }
+clap_lex = "=1.0.0"
+
 # enable default feature for icu
 writeable = "0.6.0"
 


src/lib.rs
```diff
@@ -551,6 +551,8 @@ impl PartialEq for JsValue {
     }
 }
 
+impl Eq for JsValue {}
```
drewcrawford: Looks unsound.  `PartialEq` is backed by `===`, leading to `NaN !== NaN`

crates/webidl/src/lib.rs
```diff
@@ -985,12 +992,37 @@ impl<'src> FirstPassRecord<'src> {
...
+        // determine the resolution type from the `then` method's return type.
+        let extends_promise = self.extends_promise(&js_name);
+        let promising_resolution = if extends_promise {
+            let resolution = data
+                .operations
+                .get(&OperationId::Operation(Some("then")))
+                .and_then(|op| op.signatures.first())
+                .map(|sig| sig.ret.to_wbg_type(self))
+                .and_then(|wbg_ty| {
+                    if let WbgType::Promise(inner) = wbg_ty {
```
drewcrawford: Suppose I have

```webidl
callback ParseInt = unsigned long (DOMString text);

interface JsonTextPromise : Promise {
  Promise<unsigned long> then(ParseInt onfulfilled);
}
```

IIUC this algorithm produces

```rust
impl Promising for JsonTextPromise {
    type Resolution = u32;
}
```

But presumably it actually resolves to DOMString or maybe JSValue

crates/webidl/src/lib.rs
```diff
@@ -985,12 +992,37 @@ impl<'src> FirstPassRecord<'src> {
...
         }
 
+        // If this interface extends Promise (directly or transitively),
+        // determine the resolution type from the `then` method's return type.
+        let extends_promise = self.extends_promise(&js_name);
+        let promising_resolution = if extends_promise {
+            let resolution = data
+                .operations
+                .get(&OperationId::Operation(Some("then")))
+                .and_then(|op| op.signatures.first())
```
drewcrawford: Suppose I have

```webidl
callback IdentityText = DOMString (DOMString value);

interface TextPromise : Promise {
  Promise<DOMString> then(IdentityText onfulfilled);
};

interface ChildTextPromise : TextPromise {
  DOMString label();
};
```

So then

```rust
ChildTextPromise.data.operations.get("then") == None
```

therefore

```rust
impl Promising for ChildTextPromise {
    type Resolution = ::wasm_bindgen::JsValue;
}
```

I'm not really sure that's "wrong", but presumably improving it later would be an API break

src/lib.rs
guybedford: Thanks, removed to instead do custom derives.

crates/webidl/src/lib.rs
guybedford: Great catchm I've fixed this to use the argument type from the callback itself.

crates/webidl/src/lib.rs
guybedford: Yeah you're right, corrected.

crates/webidl/src/generator.rs
```diff
@@ -783,6 +788,20 @@ impl Interface<'_> {
         });
         let js_ident = raw_ident(js_name);
 
+        let promising_impl = promising_resolution.as_ref().map(|resolution| {
```
drewcrawford, line 791: Suppose I have

```webidl
// in unstable WebIDL
callback TextThenCallback = DOMString (DOMString value);

interface FancyPromise : Promise {
  Promise<DOMString> then(TextThenCallback onfulfilled);
};
```

then I get

```rust
#[cfg(web_sys_unstable_apis)]
#[wasm_bindgen]
extern "C" {
    pub type FancyPromise;
}

//MISSING: #[cfg(web_sys_unstable_apis)]
impl ::wasm_bindgen::sys::Promising for FancyPromise {
    type Resolution = ::js_sys::JsString;
}
```

crates/webidl/src/first_pass.rs
```diff
@@ -1741,6 +1741,76 @@ impl<'a> FirstPassRecord<'a> {
...
+    /// `then` method, no callback argument, or callback not found).
+    pub fn promise_resolution_type(&self, interface: &str) -> Option<WbgType<'a>> {
+        let then_op = self.find_then_operation(interface)?;
+        let sig = then_op.signatures.first()?;
+        // The first argument of then() is the onfulfilled callback
+        let onfulfilled_arg = sig.args.first()?;
+        // Resolve the argument type to find the callback name
+        let arg_type_name = Self::extract_identifier_name(onfulfilled_arg.ty)?;
+        // Look up the callback definition
+        let callback_data = self.callbacks.get(arg_type_name)?;
```
drewcrawford: Suppose I have

```webidl
callback interface TextThenCallback {
  DOMString call(DOMString value);
};

interface FancyPromise : Promise {
  Promise<DOMString> then(TextThenCallback onfulfilled);
};
```

I get

```rust
impl ::wasm_bindgen::sys::Promising for FancyPromise {
    type Resolution = ::wasm_bindgen::JsValue;
}
```

I think poking around in `callback_interfaces` would help infer `JsString` instead

crates/webidl/src/first_pass.rs
```diff
@@ -1741,6 +1741,76 @@ impl<'a> FirstPassRecord<'a> {
...
+    /// Determines the resolution type of a Promise-extending interface by
+    /// inspecting the first parameter of the `onfulfilled` callback in its
+    /// `then` method. This is the type the promise passes to the callback
+    /// when it resolves.
+    ///
+    /// Returns `None` if the resolution type cannot be determined (e.g., no
+    /// `then` method, no callback argument, or callback not found).
+    pub fn promise_resolution_type(&self, interface: &str) -> Option<WbgType<'a>> {
+        let then_op = self.find_then_operation(interface)?;
+        let sig = then_op.signatures.first()?;
```
drewcrawford: Suppose I have

```
callback TextCb = DOMString (DOMString value);

interface FancyPromise : Promise {
  Promise<DOMString> then();
  Promise<DOMString> then(TextCb onfulfilled);
};
```

The first overload has no `onfulfilled` parameter, so inference gives up and falls back to `JsValue`, even though the second overload is clear enough it's a `DOMString` fulfillment value. So maybe we need to scan.

This raises the possibility that they can actually disagree.  I suppose the conservative option is back to `JsValue` again, though I don't know any real-world code like that so not sure if it's even a thing.

crates/webidl/src/generator.rs
guybedford: Fixed - added `#unstable_attr` to the `Promising` impl so it gets the `#[cfg(web_sys_unstable_apis)]` gate matching the extern block.

crates/webidl/src/first_pass.rs
guybedford: Good catch - now also looks up `callback_interfaces` in addition to plain `callbacks` when resolving the onfulfilled parameter type.

crates/webidl/src/first_pass.rs
guybedford: Fixed - now scans all overloaded signatures of `then()` instead of just the first, skipping any that don't have a resolvable callback argument. If overloads disagree we take the first match which seems reasonable since disagreeing overloads would be unusual in practice.

guybedford: Updated - now collects resolution types from all overloaded signatures and passes them through a new `WbgType::singular_union` helper. This deduplicates first (so identical types just collapse), then upcasts to the best common base type: all strings -> `DomString`/`JsString`, all numbers -> `Double`/`Number`, all interfaces -> `Object`, otherwise -> `Any`/`JsValue`.

crates/webidl/src/lib.rs
```diff
@@ -985,12 +992,36 @@ impl<'src> FirstPassRecord<'src> {
...
 
+        // If this interface extends Promise (directly or transitively),
+        // implement the `Promising` trait. The resolution type is determined
+        // from the first parameter of the `onfulfilled` callback in the
+        // `then()` method — that parameter is the type the promise passes to
+        // the callback when it resolves. We walk up the interface hierarchy
+        // to find the `then` method if it is only defined on a parent.
+        //
+        // Falls back to `JsValue` if no `then` method or callback is found.
+        let extends_promise = self.extends_promise(&js_name);
```
drewcrawford, line 1003: Suppose I have

```webidl
interface ResolvedType {};

callback ResolvedCb = ResolvedType (ResolvedType value);

interface PromiseWithInterfaceResolution : Promise {
  Promise<ResolvedType> then(ResolvedCb onfulfilled);
};
```

We emit these features

```toml
PromiseWithInterfaceResolution = []
ResolvedType = []
```

So a crate can enable

```toml
web-sys = { features = ["PromiseWithInterfaceResolution"] }
```

(which will not compile due to the dependent type)

I think this feature edge should instead be:

```toml
PromiseWithInterfaceResolution = ["ResolvedType"]
```

drewcrawford (commented):
I think this is the only edge case left that I can find.

PR #5033, opened 2026-03-09
Add Float16Array bindings

### LINKED ISSUE #4984: Float16Array bindings
Float16Array has been supported by all major browsers for a while now, but looks like we never added bindings for it in js-sys.

This should be fairly straightforward - just follow existing examples of typed arrays as well as add new DataView methods.

teddygood:
### Description

Closes #4984 

This adds `Float16Array` bindings to `js-sys`, along with `DataView::get_float16` / `set_float16`.

Since Rust does not yet have a stable built-in `f16` type, this avoids adding direct `f16` conversions or a dependency on `half` for now. Instead, it adds the raw bindings, `u16` slice helpers, and explicit `f64` element accessors, leaving room for future `f16` APIs once the type stabilises.

Tests were added for the new `Float16Array` bindings and `DataView` float16 methods, and everything passes with `cargo test -p js-sys --target wasm32-unknown-unknown`.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 
 ## Unreleased
 
+### Added
+
+* Added `js_sys::Float16Array` bindings, `DataView` float16 accessors using
+  `f32`, and raw `[u16]` helper APIs for interoperability with binary16
+  representations such as `half::f16`.
+  [#5033](https://github.com/wasm-bindgen/wasm-bindgen/pull/5033)
+
 ## [0.2.115](https://github.com/rustwasm/wasm-bindgen/compare/0.2.114...0.2.115)
 
 ### Added


crates/js-sys/src/lib.rs
```diff
@@ -13119,6 +13147,174 @@ pub fn global() -> Object {
...
+    pub fn set_index_from_f64(this: &Float16Array, idx: u32, value: f64);
+}
+
+impl Default for Float16Array {
+    fn default() -> Self {
+        Self::new(&JsValue::UNDEFINED.unchecked_into())
+    }
+}
+
+impl TypedArray for Float16Array {}
```
drewcrawford, line 13319: We use TypedArray as implying atomic ops https://github.com/wasm-bindgen/wasm-bindgen/blob/d148c73774249e950bd9e3fe0292e79cc4e0f985/crates/js-sys/src/lib.rs#L2413 but f16 doesn't have those.  A new conformance adds ~18 public atomic APIs and generates all the code for them, all to unconditionally trap in each case.  (And it's in JS where the optimizer can't tell to strip it.)  If we had a `T: AtomicArray` bound it would avoid all that headache plus catch misuse of f16 atomics at compile-time.

That said, we don't have that bound. f32 and f64 also work this way, so it can also make sense to do f16 this way.

I'm not sure if there is a tipping point where we decide to change design on this, or whether if we did later there would be a problem with yanking broken APIs.  @RReverser ?

teddygood: Thanks for the review. I didn't realize that `TypedArray` is also used for the atomic APIs, so I see how this would unintentionally expand the public API surface. Since you've already tagged, I'll wait for their input and update the implementation accordingly.

RReverser: Oops thanks for the ping @guybedford.

Yeah this is unfortunate but I'd say we should keep the consistency with other floats, and fix them all in one go in the futurem Worth raising a separate tracking issue so that we don't forget.

teddygood: In that case, should we wrap up this PR here and open a follow-up tracking issue for the design concern, or would it be better to address it in this PR?

crates/js-sys/src/lib.rs
```diff
@@ -3350,6 +3350,20 @@ extern "C" {
     #[wasm_bindgen(method, js_name = getFloat32)]
     pub fn get_float32_endian(this: &DataView, byte_offset: usize, little_endian: bool) -> f32;
 
+    /// The `getFloat16()` method gets a signed 16-bit float at the specified
+    /// byte offset from the start of the DataView.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getFloat16)
+    #[wasm_bindgen(method, js_name = getFloat16)]
+    pub fn get_float16(this: &DataView, byte_offset: usize) -> f64;
```
RReverser: Come to think of it, f32 is also a native Wasm type and enough for a lossless representation.

We need `_as_f32` suffix and a doc comment on these too, as we want to reserve the unsuffixed version for when f16 is stabilised.

crates/js-sys/src/lib.rs
```diff
@@ -3448,6 +3462,20 @@ extern "C" {
     #[wasm_bindgen(method, js_name = setFloat32)]
     pub fn set_float32_endian(this: &DataView, byte_offset: usize, value: f32, little_endian: bool);
 
+    /// The `setFloat16()` method stores a signed 16-bit float value at the
+    /// specified byte offset from the start of the DataView.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/setFloat16)
+    #[wasm_bindgen(method, js_name = setFloat16)]
+    pub fn set_float16(this: &DataView, byte_offset: usize, value: f64);
```
RReverser: Same here as for getters.

crates/js-sys/src/lib.rs
```diff
@@ -13123,6 +13151,174 @@ pub fn global() -> Object {
...
+        byte_offset: u32,
+        length: u32,
+    ) -> Float16Array;
+
+    /// The `fill()` method fills all elements from a start index to an end
+    /// index with a static value.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/fill)
+    #[wasm_bindgen(method)]
+    pub fn fill(this: &Float16Array, value: f64, start: u32, end: u32) -> Float16Array;
```
RReverser: We probably want suffix for all methods that take f64 (f32 after suggested change) directly and not operate on just JsValue/ArrayBuffer.

crates/js-sys/src/lib.rs
RReverser: Yup, leaving some extra comments but this particular issue is imo best left out of scope.

teddygood:
Thanks for the review! I'm also happy to open a follow-up tracking issue for the broader design concern if that would be helpful.

crates/js-sys/src/lib.rs
```diff
@@ -13123,6 +13172,179 @@ pub fn global() -> Object {
...
+        byte_offset: u32,
+        length: u32,
+    ) -> Float16Array;
+
+    /// The `fill()` method fills all elements from a start index to an end
+    /// index with a static `f32` value.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/fill)
+    #[wasm_bindgen(method)]
+    pub fn fill_from_f32(this: &Float16Array, value: f32, start: u32, end: u32) -> Float16Array;
```
RReverser: Totally nitpicking but

```suggestion
    pub fn fill_with_f32(this: &Float16Array, value: f32, start: u32, end: u32) -> Float16Array;
```

sounds better here perhaps?

RReverser:
Thanks! I think this is in pretty good shape, final note - let's extend doc-comments and tests with usages via `half` - we don't want it as a hard dependency, but given that it's going to be primary usage for the `[u16]` APIs, it's worth highlighting in examples and testing against.

crates/js-sys/src/lib.rs
teddygood: Makes sense, I'll update it.

teddygood:
Added doc-comments and tests using `half` for the `[u16]` helpers. If anything else is needed, feel free to let me know!

crates/js-sys/src/lib.rs
```diff
@@ -13142,6 +13191,205 @@ pub fn global() -> Object {
...
+
+    /// Sets the value at `idx` from an `f32`, equivalent to JavaScript
+    /// `arr[idx] = value`.
+    #[wasm_bindgen(method, indexing_setter)]
+    pub fn set_index_from_f32(this: &Float16Array, idx: u32, value: f32);
+}
+
+impl Default for Float16Array {
+    fn default() -> Self {
+        Self::new(&JsValue::UNDEFINED.unchecked_into())
```
RReverser, line 13315: Not related to this PR but ugh I didn't know we had these. Feels super wrong.

RReverser:
Please also update CHANGELOG following existing style, and we'll merge this.

RReverser:
Thanks again!

Issue #2746, opened 2021-12-23
Accessing 'undefined' value in passStringToWasm -- Chrome autocomplete


martin-kolarik:
When Chrome puts autocomplete value into <input "text">, it creates keydown event with key property undefined. Using KeyboardEvent::key() then asks for string value, passStringToWasm is called, and there arg 'undefined'.length is accessed.

Empty string should be returned to caller in this case.

jonboj:
The source code refered above, is this one of the `wasm-bindgen/examples`, in case which one?

martin-kolarik:
Not know, I run my project. I am using dominator's (https://github.com/Pauan/rust-dominator) event handling, where mentioned call appears> https://github.com/Pauan/rust-dominator/blob/1f541fd7170287e657c4cae1ff418430cfd625ed/src/events.rs#L131.

Dominator itself is only accessing key() method of web-sys::KeboardEvent, then the issue happens in code generated by wasm-bindgen (AFAIK).

jonboj:
Looking at [SO autocomplete Chrome](https://stackoverflow.com/questions/59534586/google-chrome-fires-keydown-event-when-form-autocomplete), running the `JSFiddle` snippet in the question, when filling with autocomplete `console.log (e.key);` logs `undefined` in the JSFiddle console. So to me it would be unexpected if `web-sys::KeboardEvent.key()` didn't return same.

martin-kolarik:
Yes, I confirm that running the JSFiddle reports correctly 'e.key = undefined'. I think this is the point. According to https://www.w3.org/TR/uievents/#idl-keyboardevent, 'key' property must always be defined, but Chrome, when autocompleting, constructs fake keydown event without the property defined:
![image](https://user-images.githubusercontent.com/74831268/147971673-84f74182-7434-4477-aaff-42dfe5f83888.png)

I recognize that wasm-bindgen's IDLs are correct, and that Chrome behaves wrongly (IMO). Chrome injects bad event, wasm-bindgen's generated key accessor processes the event, expecting it cannot be undefined, and as a result wasm-bindgen's passStringToWasm fails.

Formally incorrect (but working) solution would be to change IDL from `readonly attribute DOMString key;` to `readonly attribute DOMString? key;`. Then gen_KeyboardEvent.rs would contain `key(...) -> Option<String>` and JS interop would not fail. I have tested this approach and it works.

So, now it looks to me like that wasm-bindgen does not manage to defend against malformed Chrome event.

[...]

PR #5034, opened 2026-03-09
fix: normalize argument in encode_as_ascii to handle input prediction issues

WenSimEHRP:
### Description

This PR adds an extra step in `passStringToWasm` to normalize the string in case if the string is null or undefined. (e.g., when produced by in-browser IME such as Google Input Tools, or input prediction)

Related:

- https://github.com/emilk/egui/issues/7964
- #2746

### Checklist

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2114,8 +2114,12 @@ if (require('worker_threads').isMainThread) {{
         // This might be not very intuitive, but such calls are usually more
         // expensive in mainstream engines than staying in the JS, and
         // charCodeAt on ASCII strings is usually optimised to raw bytes.
+        //
+        // We also normalize the arg passed in in case tools like Google
+        // Input Tools or input prediction mess up the string.
         let encode_as_ascii = format!(
             "\
+                arg = arg ?? \"\";
                 if (realloc === undefined) {{
                     const buf = cachedTextEncoder.encode(arg);
                     const ptr = malloc(buf.length, 1) >>> 0;


crates/cli-support/src/js/mod.rs
```diff
@@ -1836,8 +1836,12 @@ if (require('worker_threads').isMainThread) {{
             // This might be not very intuitive, but such calls are usually more
             // expensive in mainstream engines than staying in the JS, and
             // charCodeAt on ASCII strings is usually optimised to raw bytes.
+            //
+            // We also normalize the arg passed in in case tools like Google
+            // Input Tools or input prediction mess up the string.
             let encode_as_ascii = format!(
                 "\
+                    arg = arg ?? "";
```
guybedford: Can you explain further what case this is guarding for? I'm not familiar with Google Input Tools....

wensimehrp: I'm not sure about other languages, but Chinese and Japanese input methods generally have a candidate zone where the user first enters some key combinations then choose the character from a panel, like this

image

the problem is that this still counts as input. However the input is blocked at the input tools level, and the input tool won't send any strings until the user have selected their candidate. In this case, `arg` would be null, and checking for `length` on that would crash the program.

drewcrawford: I think this fix only works in release builds.  in debug builds, we have `typeof(arg) !== 'string'` earlier, which I think has the same problem, so we'd need to do this earlier to workaround.

@guybedford I think the problem is [KeyboardEvent.key](https://www.w3.org/TR/uievents/#dom-keyboardevent-key) is allegedly a non-nullable DOMString, but we wound up getting `undefined` for it instead, which might be a browser bug.

I did find a cryptic note in the spec that seems to directly refer to this situation:

> In some implementations or system configurations, some key events, or their values, might be suppressed by the IME in use.

I take this to mean we get `keyCode==""` in this scenario, but I have to wonder if someone on the browser end of things read the note and implemented by `undefined`.

guybedford (changes requested):
I think the fix here is to re-type the failing API to take a JsValue instead of a String type, so that it can be coerced there. When the wasm-bindgen type system is invalidated there is no expectation of correct operation.

Therefore, I would recommend updating the bindings not changing the way all strings are interpreted.

wensimehrp:
I think I'll try to locate and fix the related API in the UI library I'm using instead.

PR #5035, opened 2026-03-10
feat: add StaleObjectError class

logan-gatlin:
### Description
When re-initialization happens, objects created in a previous instance are considered stale, and using them is an error. This error is potentially recoverable. Giving it its own type allows it to be caught in a try-catch block

### Checklist

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -184,7 +184,7 @@ impl<'a, 'b> Builder<'a, 'b> {
                 js.prelude(
                     "
                     if (this.__wbg_inst !== undefined && this.__wbg_inst !== __wbg_instance_id) {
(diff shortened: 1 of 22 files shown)


drewcrawford:
My only question on this is the naming.  `StaleObjectError` is a global that could potentially conflict with another crate's export – but I looked around, and can't find any rust or JS code doing that, so that seems unlikely.

Usually wasm-bindgen exports have `__wbg_` prefix, so from that I'd guess a name like `__wbg_StaleObjectError`.  But the exports we have are quite different from this one, so it might create some precedent for other export names.

logan-gatlin:
I fixed the name to be `__wbg_StaleObjectError`. After considering it more, I think the more general approach would be a re-initialize hook function like in [this PR](https://github.com/wasm-bindgen/wasm-bindgen/pull/5036). If others find this useful feel free to resubmit this

PR #5036, opened 2026-03-11
feat: User reinitialize hook

logan-gatlin:
### Description
Adds `#[wasm_bindgen(pre_reinit_hook)]` and `#[wasm_bindgen(post_reinit_hook)]` attributes that let users define lifecycle functions called during `__wbg_reset_state()`. Any type that implements `serde::Serialize` can be passed between these two functions. State transfer requires the user to import `serde_wasm_bindgen` since re-exporting that crate would create a circular dependency
### Checklist


crates/cli-support/src/js/mod.rs
```diff
@@ -2999,6 +3004,34 @@ if (require('worker_threads').isMainThread) {{
...
+        define_export(
+            &mut self.exports,
+            "__wbg_set_reinit_hook",
+            &[],
+            ExportEntry::Definition(ExportDefinition {
+                comments: None,
+                identifier: hook_identifier,
+                definition: hook_definition,
+                ts_definition:
+                    "function __wbg_set_reinit_hook(callback: ((newInstance: InitOutput, oldInstance: InitOutput) => void) | null): void;\n"
```
drewcrawford: I think this breaks some of the non-web targets.  Reproduction steps:

```bash
mkdir -p /tmp/empty-types
npx --yes -p typescript tsc --noEmit --typeRoots /tmp/empty-types crates/cli/tests/reference/targets-target-module-experimental-reset-state-function.d.ts
```

output

```
crates/cli/tests/reference/targets-target-module-experimental-reset-state-function.d.ts:6:64 - error TS2552: Cannot find name 'InitOutput'. Did you mean 'MIDIOutput'?

6 export function __wbg_set_reinit_hook(callback: ((newInstance: InitOutput, oldInstance: InitOutput) => void) | null): void;
                                                                 ~~~~~~~~~~

  ../../../Users/drew/.npm/_npx/9ca470fa61f45e06/node_modules/typescript/lib/lib.dom.d.ts:19564:13
    19564 declare var MIDIOutput: {
                      ~~~~~~~~~~
    'MIDIOutput' is declared here.

crates/cli/tests/reference/targets-target-module-experimental-reset-state-function.d.ts:6:89 - error TS2552: Cannot find name 'InitOutput'. Did you mean 'MIDIOutput'?

6 export function __wbg_set_reinit_hook(callback: ((newInstance: InitOutput, oldInstance: InitOutput) => void) | null): void;
                                                                                          ~~~~~~~~~~

  ../../../Users/drew/.npm/_npx/9ca470fa61f45e06/node_modules/typescript/lib/lib.dom.d.ts:19564:13
    19564 declare var MIDIOutput: {
                      ~~~~~~~~~~
    'MIDIOutput' is declared here.


Found 2 errors in the same file, starting at: crates/cli/tests/reference/targets-target-module-experimental-reset-state-function.d.ts:6
```

My rough idea of a solution is

```rust
let hook_ts_arg = match self.config.mode {
    OutputMode::Web => "((newInstance: InitOutput, oldInstance: InitOutput) => void) | null",
    OutputMode::Module | OutputMode::Node { module: false } => "((newInstance: any, oldInstance: any) => void) | null",
    _ => unreachable!(),
};

```

logan-gatlin: Good catch, I think `WebAssembly.Exports` would be better than `any` though?

drewcrawford: You're right, `WebAssembly.Exports` would be better

crates/cli/tests/reference/targets-target-web-experimental-reset-state-function-atomics.d.ts
```diff
@@ -3,6 +3,8 @@
 
 export function __wbg_reset_state(): void;
 
+export function __wbg_set_reinit_hook(callback: ((newInstance: InitOutput, oldInstance: InitOutput) => void) | null): void;
```
drewcrawford: So I get the motivation to pass both instances here and try to move state across. The problem is that oldInstance is not safely callable inside this hook. Enumerating its exports, as test coverage does, seems ok, so its not exactly a bug.  But the obvious usecase of migrating state would be.

To migrate state safely, I think it'd require a phased process where oldInstance is only used before the swap, and the new instance is only used afterward.

To illuminate the danger, this hook to migrate some state seems pretty reasonable:

```js
wasm.__wbg_set_reinit_hook((newInstance, oldInstance) => {
	wasm.import_profile_draft(profileDraftState());
	wasm.import_session(snapshotSessionFromPreviousInstance(oldInstance));
});
```

Trick is, `snapshotSessionFromPreviousInstance(oldInstance)` does not actually decode through the old instance anymore. It gets a pointer from the old instance, but wasm-bindgen’s JS glue decodes that pointer against the new instance’s memory instead. 

With a bit of heap shaping, here's how build that into an exploit chain for a hypothetical application.  The key move is that the app migrates trusted and untrusted state across reinit and, at the source level, writes each value to the right destination. But at the memory level, the trusted read is resolved against the new instance’s heap, so it trusts an attacker-controlled `role` instead.

```rust
const REINIT_HOOK_OLD_INSTANCE_LIB_RS: &str = r#"
	use wasm_bindgen::prelude::*;
	use std::cell::RefCell;

	#[wasm_bindgen(inline_js = "
		globalThis.__exported_session = null;
		export function stage_exported_session(value) {
			globalThis.__exported_session = value;
		}
	")]
	extern "C" {
		fn stage_exported_session(value: &str);
	}

	thread_local! {
		// Trusted session bytes owned by the pre-reset instance.
		static OLD_SESSION_BYTES: RefCell<Vec<u8>> = RefCell::new(Vec::new());
		// Trusted session state used by the active post-reset instance.
		static ACTIVE_SESSION: RefCell<String> = RefCell::new(String::from("role=anon"));
		// Ordinary user-controlled profile state in the active post-reset instance.
		static PROFILE_DRAFT: RefCell<Vec<u8>> = RefCell::new(Vec::new());
	}

	#[wasm_bindgen]
	pub fn initialize_trusted_session() {
		OLD_SESSION_BYTES.with(|session| *session.borrow_mut() = b"role=user".to_vec());
	}

	#[wasm_bindgen]
	pub fn import_session(value: String) {
		ACTIVE_SESSION.with(|session| *session.borrow_mut() = value);
	}

	#[wasm_bindgen]
	pub fn import_profile_draft(value: String) {
		PROFILE_DRAFT.with(|draft| {
			let mut draft = draft.borrow_mut();
			let value = value.into_bytes();
			if value.len() == draft.len() {
				draft.clone_from_slice(&value);
			} else {
				*draft = value;
			}
		});
	}

	#[wasm_bindgen]
	pub fn current_session() -> String {
		ACTIVE_SESSION.with(|session| session.borrow().clone())
	}

	#[wasm_bindgen]
	pub fn current_profile_draft() -> String {
		PROFILE_DRAFT.with(|draft| {
			let draft = draft.borrow();
			String::from_utf8(draft.clone()).unwrap()
		})
	}

	#[wasm_bindgen]
	pub fn export_session_for_restore() {
		OLD_SESSION_BYTES.with(|session| {
			let session = session.borrow();
			stage_exported_session(core::str::from_utf8(&session).unwrap());
		});
	}
"#;
#[test]
fn reinit_hook_old_instance_uses_new_memory() {
	let mut project = Project::new("reinit_hook_old_instance_uses_new_memory");
	project.file("src/lib.rs", REINIT_HOOK_OLD_INSTANCE_LIB_RS).file(
		"Cargo.toml",
		&format!(
			"
					[package]
					name = \"reinit_hook_old_instance_uses_new_memory\"
					authors = []
					version = \"1.0.0\"
					edition = '2021'

					[dependencies]
					wasm-bindgen = {{ path = '{}' }}

					[lib]
					crate-type = ['cdylib']

					[workspace]

					[profile.dev]
					codegen-units = 1
				",
			REPO_ROOT.display(),
		),
	);

	let out_dir = project
		.wasm_bindgen("--target nodejs --experimental-reset-state-function")
		.unwrap();

	fs::write(
		out_dir.join("test_old_instance_uses_new_memory.js"),
		r#"
const { describe, it } = require('node:test');
const assert = require('node:assert/strict');

const wasm = require('./reinit_hook_old_instance_uses_new_memory.js');

describe('reinit hook old instance', () => {
	it('restores a forged session during reinit from attacker-controlled new-instance input', () => {
		function profileDraftState() {
			return 'role=root';
		}

		function snapshotSessionFromPreviousInstance(oldInstance) {
			globalThis.__exported_session = null;
			oldInstance.export_session_for_restore();
			return globalThis.__exported_session;
		}

		wasm.initialize_trusted_session();

		wasm.__wbg_set_reinit_hook((newInstance, oldInstance) => {
			// Write untrusted state to the untrusted slot.
			wasm.import_profile_draft(profileDraftState());

			// Write trusted state to the trusted slot.
			wasm.import_session(snapshotSessionFromPreviousInstance(oldInstance));
		});

		wasm.__wbg_reset_state();

		assert.strictEqual(wasm.current_profile_draft(), 'role=root');
		assert.strictEqual(wasm.current_session(), 'role=user');
	});
});
"#,
	)
	.unwrap();

	Command::new("node")
		.arg("--test")
		.arg("test_old_instance_uses_new_memory.js")
		.current_dir(&out_dir)
		.assert()
		.success();
}
```

guybedford:
Another thought here - both reinit and dispose could in theory be written in Rust. When we can it's my preference to write Rust hooks.

drewcrawford:
I think the 2-hook design works, provided the dispose hook runs before rebinding `wasm`, and the initialize hook runs after rebinding and start. The forbidden pattern is having both instances callable simultaneously, since the wasm-bindgen glue is a singleton.

As far as writing a hook in Rust, I like that approach, with two key complications.

The first is that we cannot store the old state in Rust linear memory. We need to move it through a JS-owned slot that survives reinit. An API shape for that could look like:

```js
wasm.__wbg_set_reinit_hooks({
    before(oldInstance) {
        return oldInstance.snapshot(); // self-contained JS value
    },
    after(newInstance, snapshot) {
        newInstance.load_snapshot(snapshot);
    },
});
```

The second complication is that not every `JsValue` is actually self-contained. You can construct a `JsValue` in safe Rust that still encodes a raw pointer, a closure tied to old Rust state, or some other handle whose meaning depends on the old instance still being alive.  At that point we have recreated the same class of bug under a different type.

A minimal answer would be to make the Rust snapshot hook unsafe, so callers at least have to acknowledge that requirement explicitly and go out on their own. A better answer would be to introduce a restricted snapshot type here instead of accepting any arbitrary JsValue, e.g. a newtype thats only safe APIs construct from safe sources like string, number, serializable etc.

logan-gatlin:
I think the pre-reinit hook should be unsafe if we allow it to run after abort. Not only is executing after an abort unsafe, but calling any JS code would fail I think, since the aborted flag is set and any re-entrancy into Rust will cause an exception, or maybe recursively call __wbg_reset_state. Maybe we only allow the re-init hook to run if the instance was manually reset and not aborted? That would still support the hot reload use case I think

guybedford:
> I think the pre-reinit hook should be unsafe if we allow it to run after abort. Not only is executing after an abort unsafe, but calling any JS code would fail I think, since the aborted flag is set and any re-entrancy into Rust will cause an exception, or maybe recursively call __wbg_reset_state. Maybe we only allow the re-init hook to run if the instance was manually reset and not aborted? That would still support the hot reload use case I think

This is correct - you would not want to call the destructor hook on an aborted wasm. But in cases where __wbg_reset_state is being called for a non-aborted Wasm (such as in a hot reloading workflow), it would be useful. The design suggestion is in the context of hot reloading, in the spirit of ensuring a well-defined primitive that can support different use cases.

logan-gatlin:
> A better answer would be to introduce a restricted snapshot type here instead of accepting any arbitrary JsValue, e.g. a newtype thats only safe APIs construct from safe sources like string, number, serializable etc.

I think `impl Serialize` with `serde_wasm_bindgen` is what we really want here. My current idea is:
1. Add a feature flag and gate the Serde dependency on it
2. Give the `pre_reinit_hook` function the signature `Fn() -> impl Serialize`
3. Wrap `pre_reinit_hook` in a function that converts the return type into `JsValue` and returns it - we know this `JsValue` is safe because it was serialized
4. Give the `post_reinit_hook` function the signature `Fn(Option<JsValue>)`, and feed it what was returned by the `pre_reinit_hook` function

logan-gatlin:
Something changed with the name mangling, the hashes on closures in the WAT are now way further off on my ARM mac, and sometimes in a different order too

logan-gatlin:
Unfortunately we cant re-export `serde-wasm-bindgen` from `wasm-bindgen` because it creates a circular dependency. The user will have to import `serde-wasm-bindgen` in their project in order for transferring state to work.

Issue #3687, opened 2023-11-04
aborted modules must be poisoned


workingjubilee:
Using wasm-bindgen, a wasm module can be resumed after abort by calling into it again from JavaScript. This violates Rust's soundness preconditions: abort _must_ terminate forward progress.

The wasm component model prevents wasm modules from being reused in this way by locking out calls into them after a module has trapped.

It is likely that wasm-bindgen should be implementing the check that poisons calls into it, or shuts off access in any other way. Even if this were implemented upstream for all users of wasm32-unknown-unknown, it would be best for wasm-bindgen to deprecate any interfaces that serve primarily to expose this route to unsound interactions to userspace. The ability to throw JavaScript exceptions has been called out as a likely example.

Also see:
- https://github.com/rust-lang/rust/issues/117344

daxpedda:
I left some additional questions/comments in the Zulip discussions, but to summarize: I'm not entirely convinced that this is a problem in `wasm-bindgen` because throwing a JS exception doesn't call `abort()`.

RalfJung:
What is the type of throwing a JS exception in Rust? AFAIK wasm is a no-unwind target. So every function that returns `!` is like abort -- no code must ever be executed in this thread any more.

daxpedda:
> So every function that returns `!` is like abort -- no code must ever be executed in this thread any more.

It is indeed `!`. Is this really required by Rust? It is hard to find any documentation around this.
As far as I can tell this isn't practically unsafe: even though no-unwind Wasm throws, it doesn't clear the stack where it throws. Obviously this can and will cause issues, but it shouldn't be unsafe or cause UB.

If this is just a "semantic" issue, I would like to propose to change that in Rust: not require `!` to be an `abort()`, or at least on some platforms, e.g. Wasm.

RalfJung:
The concern is around optimizations Rust performs that need this rule, so it's not something we can fix by just changing some wording somewhere.

I'll need to better understand what this function does and doesn't do... once I have time I'll catch up on Zulip and have some questions there.

daxpedda:
So conclusion:
This is definitely unsafe and Rust should implement some sort of module poisoning when we abort.

From `wasm-bindgen`s side, we could definitely just deprecate any throwing function, though I'm not sure if this does make sense because it's definitely a widely used and probably necessary future, so alternatively we might want to add poisoning to `wasm-bindgen` as well or when we add poisoning to Rust we should add it in a way that can cover issues like that.

[...]

PR #5037, opened 2026-03-12
feat: forward console output from user-spawned workers

drewcrawford:
# Forward Console Output From User-Spawned Workers

## Summary

Capture and forward the log output in the test runner when a test (or any arbitrary user code) spawns a worker, and then that worker calls console.debug/log/info/warn/error to log.

(Workers created by the runner, such as `wasm_bindgen_test_configure!(run_in_dedicated_worker)` do this already; my focus here is on some other, arbitrary user worker, for example [wasm_thread](https://crates.io/crates/wasm_thread))

This PR implements logging redirection for the following worker types, when they are spawned by a user:

- dedicated workers
- module workers
- URL-based workers
- shared workers

(This PR does not cover panic forwarding from these workers, which is a follow-on problem.)

## Motivation

Consider this typical test shape:

```rust
#[wasm_bindgen_test] fn my_test() {
   maybe_this_closure_runs_on_a_worker(|| {
    console_log!("does this log message appear?");
   }
}
```

Today, that log will appear if-and-only-if the closure runs on the same worker as the test itself.  Deciding whether that property holds in practice requires keeping very careful mental track of how `maybe_this_closure_runs_on_a_worker` is implemented.  But this mental math is unrealistic, because the implementation can change at any time

The goal of this PR is: stop keeping mental track of all this, log fearlessly anywhere in a test and that should just work™️.

A workflow for minimizing bugs I use frequently is as follows:  Write a failing test. `cargo vendor` relevant dependencies, and sprinkle logging in there liberally as needed.  I should not be simultaneously saddled with a close analysis of whether or not I'm still on the test worker in each particular sprinkle; I should just expect that if I'm inserting a console.log, it logs.

## Design

At startup, the test runner patches `Worker` and `SharedWorker` constructors in the browser test page. The patched constructors wrap the target script with a small console shim that forwards `debug/log/info/warn/error` back to the main test page while preserving the original console behavior inside the worker.

This is basically what we do already on the main page; the idea behind this PR is to roll that out to a custom user worker.

The main page then routes those forwarded messages into the DOM elements from #4960, i.e. 

- `#output` for `--nocapture`
- `#console_output` for failure-path capture in headless mode

## Why This Design

Earlier I proposed a different design in #4856.  The idea there was to add an explicit public API that can opt into this.

That design had one clear advantage: it is a lot simpler.  However, the opt-in is problematic in practice:

- users must know the API exists
- ecosystem crates must remember to call it
- logs from workers created in dependencies are still easy to miss
- the debugging experience differs depending on whether or not some downstream crate opts in
- you can't just sprinkle `console.log` throughout your dependency tree and expect that will work, you also need to sprinkle the enablement call at the same time.  This makes debugging that may involve adding hundreds or thousands of logs to instrument, very tedious.

VS, in this PR we get:

- no API to learn
- no user-code changes are required
- works for transitive worker usage
- fearlessly log from anywhere in your dependency tree and that should just work ™️

## Test plan

This PR also adds 6 new tests that logs are forwarded from:

- dedicated workers
- module workers
- URL-backed workers
- shared workers
- browser failure-path capture for dedicated workers
- browser failure-path capture for shared workers

Altogether these cover both live --nocapture output and headless failure-output reporting, and verify that worker-spawned debug/log/info/warn/error messages are captured and forwarded.

### DIFF
--- a/crates/test/Cargo.toml
+++ b/crates/test/Cargo.toml
@@ -51,6 +51,9 @@ web-sys = { path = "../web-sys", features = [
 [[bench]]
 name = "headless_output"
 
+[[bench]]
+name = "headless_worker_output"
+
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(wasm_bindgen_unstable_test_coverage)'] }
 
--- a/crates/test/benches/headless_output.rs
+++ b/crates/test/benches/headless_output.rs
@@ -7,8 +7,6 @@
 ///   cargo bench --bench headless_output --target wasm32-unknown-unknown -p wasm-bindgen-test
 use std::time::Duration;
 
-use wasm_bindgen::closure::Closure;
-use wasm_bindgen::JsCast;
 use wasm_bindgen_test::{wasm_bindgen_bench, wasm_bindgen_test_configure, Criterion};
 
 wasm_bindgen_test_configure!(run_in_browser);
@@ -39,47 +37,6 @@ async fn prewarm(kb: usize) {
 
 fn performance_now() -> f64 {
     web_sys::window().unwrap().performance().unwrap().now()
-}
-
-fn create_dedicated_worker(script: &str) -> (web_sys::Worker, String) {
-    use js_sys::Array;
-    use web_sys::{Blob, BlobPropertyBag, Url, Worker};
-
-    let parts = Array::new();
-    parts.push(&wasm_bindgen::JsValue::from_str(script));
-    let opts = BlobPropertyBag::new();
-    opts.set_type("application/javascript");
-    let blob = Blob::new_with_str_sequence_and_options(&parts, &opts).unwrap();
-    let url = Url::create_object_url_with_blob(&blob).unwrap();
-    let worker = Worker::new(&url).unwrap();
-    (worker, url)
-}
-
-async fn benchmark_worker_log(worker: &web_sys::Worker) {
-    let promise = js_sys::Promise::new(&mut |resolve, reject| {
-        let resolve_clone = resolve.clone();
-        let onmessage = Closure::once_into_js(move |_e: web_sys::MessageEvent| {
-            resolve_clone.call0(&wasm_bindgen::JsValue::NULL).unwrap();
-        });
-        worker.set_onmessage(Some(onmessage.unchecked_ref()));
-
-        let reject_clone = reject.clone();
-        let onerror = Closure::once_into_js(move |e: web_sys::ErrorEvent| {
-            reject_clone
-                .call1(
-                    &wasm_bindgen::JsValue::NULL,
-                    &wasm_bindgen::JsValue::from_str(&e.message()),
-                )
-                .unwrap();
-        });
-        worker.set_onerror(Some(onerror.unchecked_ref()));
-
-        worker
-            .post_message(&wasm_bindgen::JsValue::UNDEFINED)
-            .unwrap();
-    });
-
-    wasm_bindgen_futures::JsFuture::from(promise).await.unwrap();
 }
 
 #[wasm_bindgen_bench]
@@ -144,47 +101,4 @@ async fn bench_console_log_10mb(c: &mut Criterion) {
         })
     })
     .await;
-}
-
-#[wasm_bindgen_bench]
-async fn bench_worker_console_log_batch(c: &mut Criterion) {
-    prewarm(1_000).await;
-
-    let (worker, url) = create_dedicated_worker(
-        r#"
-            onmessage = function() {
-                console.log("worker-bench-log");
-                postMessage("done");
-            };
-        "#,
-    );
-
-    c.bench_async_function("worker_console_log_after_1mb", |b| {
-        let worker = worker.clone();
-        Box::pin(async move {
-            b.iter_custom_future(|iters| {
-                let worker = worker.clone();
-                let mut value = 0;
-                async move {
-                    let mut elapsed = Duration::ZERO;
-                    for _ in 0..iters {
-                        let start = std::hint::black_box(performance_now());
-                        benchmark_worker_log(&worker).await;
-                        let elapsed_ms = std::hint::black_box(performance_now()) - start;
-                        elapsed += Duration::from_secs_f64(elapsed_ms / 1000.0);
-                        value += 1;
-                        if value % 100 == 0 {
-                            sleep_ms(0).await;
-                        }
-                    }
-                    elapsed
-                }
-            })
-            .await;
-        })
-    })
-    .await;
-
-    worker.terminate();
-    web_sys::Url::revoke_object_url(&url).unwrap();
 }
--- a/crates/test/benches/headless_worker_output.rs
+++ b/crates/test/benches/headless_worker_output.rs
@@ -1,0 +1,120 @@
+/// Benchmark for worker log forwarding performance in the headless browser runner.
+///
+/// Run with:
+///   cargo bench --bench headless_worker_output --target wasm32-unknown-unknown -p wasm-bindgen-test
+use std::time::Duration;
+
+use wasm_bindgen::closure::Closure;
+use wasm_bindgen::JsCast;
+use wasm_bindgen_test::{wasm_bindgen_bench, wasm_bindgen_test_configure, Criterion};
+
+wasm_bindgen_test_configure!(run_in_browser);
+
+async fn sleep_ms(ms: i32) {
+    let promise = js_sys::Promise::new(&mut |resolve, _| {
+        web_sys::window()
+            .unwrap()
+            .set_timeout_with_callback_and_timeout_and_arguments_0(&resolve, ms)
+            .unwrap();
+    });
+    wasm_bindgen_futures::JsFuture::from(promise).await.unwrap();
+}
+
+async fn prewarm(kb: usize) {
+    let msg: wasm_bindgen::JsValue = "x".repeat(100).into();
+    let n = kb * 10;
+    for i in 0..n {
+        web_sys::console::log_1(&msg);
+        if i % 100 == 0 {
+            sleep_ms(0).await;
+        }
+    }
+}
+
+fn performance_now() -> f64 {
+    web_sys::window().unwrap().performance().unwrap().now()
+}
+
+fn create_dedicated_worker(script: &str) -> (web_sys::Worker, String) {
+    use js_sys::Array;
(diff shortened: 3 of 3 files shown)


drewcrawford:
tyvm for review, following original numbering:

1. amazing catch!!! The URL wrapper now uses `JSON.stringify(scriptUrl)` for both `Worker` and `SharedWorker`, and there are dedicated regressions for both injection cases.

2. I do not think this reproduces. The existing module-worker `--nocapture` path already exercises exactly this case and passes. In the headless runner, successful output is scraped from `#output`, while `#console_output` is only read on failure, so if the module wrapper really could not see `nocapture`, those module-worker logs would land in the wrong element and that test should fail. My best theory is that the inline page script’s top-level `const nocapture = ...` creates a global lexical binding that is still name-resolvable from the later `run.js` module, even though it is not a `window.nocapture` property. If there is a concrete repro that contradicts that, I’d like to see it.

3. great catch! Non-cloneable console arguments are now handled defensively, with regressions for both `Worker` and `SharedWorker`.

4. great catch! The patched `SharedWorker` path now calls `worker.port.start()`, and the relevant tests were updated so they no longer mask this by calling `start()` themselves.

5. Addressed, with one nuance. Wrapper blob URLs are now revoked, and I switched back to the straightforward revocation approach rather than keeping the more defensive delayed variant. If this exposes a real browser bug later, we can address that with a concrete repro.

6. great catch! The `SharedWorker` shim now buffers pre-connect logs and flushes them when the first port connects. There is now a regression for top-level shared-worker logging before `connect`.

7. I agree this is still worth a code comment, but sync XHR is only used in the blob-rewrite path, where constructor interception needs a synchronous result. A larger architectural redesign could avoid this; In particular #next should have a tracking issue on this topic as I have a ton to say but I'm unaware of one (read: if there's not one, I need to file it, but where should I file that to?)

8. I agree with the concern, but my research suggests there is no realistic passive way to detect a dead `MessagePort` across major browsers. If there is a concrete cleanup strategy in mind here, I’m happy to explore it, please let me know.

9. Correct, but these feel like separate pre-existing bugs on main.  My goal here is to bring worker logging to parity with APIs we have already supported.

10. Fixed. The 200ms waits were replaced with message-based synchronization plus timeout guards.

11. Fair enough. I was still actively adding coverage, so I preferred not to mix a larger test refactor into the functional changes here. I’d rather do that cleanup separately.

12. Coverage is vastly improved:
- module worker via URL: added and passing
- non-cloneable console args: added for `Worker` and `SharedWorker`
- user message-handler coexistence: covered for both `onmessage` and `addEventListener("message", ...)`
- nested workers: added and now passing

I also introduced a new benchmark to track worker logging performance more directly.

drewcrawford:
> 1. nocapture scoping bug — still present (High)

I still think this a false positive, and nocapture scoping is correct. While it's true that const in a classic script doesn't create a window property, ES modules do have access to the global lexical scope. Per the ECMAScript spec, both classic scripts and modules share the same Realm's Global Environment Record, which includes the declarative environment where const/let bindings live. A bare nocapture reference (not window.nocapture) resolves correctly through the scope chain.                     

On the spec side,

 1. Global Environment Record — https://tc39.es/ecma262/#sec-global-environment-records: "A Global Environment Record is logically a single record but it is specified 
  as a composite encapsulating an Object Environment Record and a Declarative Environment Record."                                                                      
 2. Module environment setup — https://tc39.es/ecma262/#sec-source-text-module-record-initialize-environment, step 5: Let env be                                       
  NewModuleEnvironment(realm.[[GlobalEnv]]). This sets the module's OuterEnv to the Global Environment Record, so free variable lookups in module code walk up to that same composite record — including its declarative part where const/let bindings live.  

On the empirical side I've verified this behavior experimentally in chrome and firefox.  An external <script type=module> can read a const declared in a preceding classic <script> block. The existing test_worker_console_no_panic_nocapture test also passes, exercising exactly this code path.  If there is a different reproducer than can catch this, I'd need more info.

guybedford:
Addressed the remaining round-2 feedback in commit 3e0b87319:

**Code duplication (issues #3 from round 2)**

`__wbg_forward_console_message` was defined identically in three places. It cannot be a single runtime definition because the three call sites run in separate JS realms (dedicated worker shim, shared worker shim, and the wasm-bindgen worker script), but the three copies now come from a single Rust source string (`forward_console_message_fn`) that is interpolated into each context, so they are guaranteed to stay in sync.

The `Worker`/`SharedWorker` blob-rewrite logic was copy-pasted verbatim. This is now extracted into a shared `__wbg_wrap_worker_url()` JS helper used by both constructor patches, removing ~60 lines of duplication.

**`data:` URL path (issue #2 from round 2)**

Added an explicit comment in the non-blob branch explaining that `data:` URLs fall through to `importScripts()`, that this is blocked by CSP in most configurations, and that only `blob:` and `http(s):` URLs are fully supported.

**Synchronous XHR (issue #5 from round 2)**

Added a comment at the XHR call site explaining the architectural constraint: the Worker constructor is synchronous and blob contents must be rewritten before a URL is passed to it, so an async fetch cannot be awaited here without restructuring the entire constructor patch. Also noted that a tracking issue for an async alternative should be filed against the headless test pipeline.

**Blob revocation strategy (issue #6 from round 2)**

Added a comment explaining the revocation logic: the blob is only needed during script load, so revoking on the first message/error event is correct. The 5 s `setTimeout` fallback is a conservative safety net against browsers that delay the first message event — no specific known bug, but included as a precaution.

**SharedWorker port accumulation (issue #4 from round 2)**

Added a comment explaining why there is no removal path: there is no `close` event on `MessagePort` itself and no reliable passive way to detect a dead port cross-browser. `postMessage` to a dead port is silently dropped by the browser, so the accumulation is benign in practice.

**Clippy (pre-existing)**

Fixed 59 `clippy::uninlined_format_args` violations in `headless_streaming_tests.rs` that were caught by `just clippy`.

guybedford (approved):

guybedford:
One thing we should talk about here - the codspeed slowdown.

It seems this is due to the overhead of loading the shims, so seems understandable. But this also makes me realise how much stuff we need to do here just to get logs when really this should just be something the webdriver gives us via the log API?

Is the major concern here just that Firefox doesn't support it?

Would be interested to hear from @daxpedda as well if headless browsers have the same issue, or if they can avoid console hooks like these.

I'm happy to land this still, but think it's worth some discussion about better logging APIs in future first.

drewcrawford:
> One thing we should talk about here - the codspeed slowdown.

This is actually just noise from the new benchmark, resolved by splitting up into separate executables

> But this also makes me realise how much stuff we need to do here just to get logs when really this should just be something the webdriver gives us via the log API?  Is the major concern here just that Firefox doesn't support it?

My understanding is that **no browser** supports it. My research for the design turned up [this selenium comment](https://www.selenium.dev/selenium/docs/api/javascript/lib_logging.js.html) where they face a similar issue:

> * **NOTE:** Only a few browsers support the remote logging API (notably Firefox and Chrome). Firefox supports basic logging functionality, while Chrome exposes robust [performance logging](https://chromedriver.chromium.org/logging) options. Remote logging is still considered a non-standard feature,

That comment is a bit inaccurate since not even firefox or chrome supports logging for **workers** over **classic** webdriver.  The situation is a bit better on bidi webdriver (which we don't support at all) but only marginally, it adds support for only some combinations of worker types and browsers.

daxpedda:
In JBG we will attempt to stay as close as possible to what Std does (see [`libtest`](https://github.com/rust-lang/rust/blob/3ea2fbcb2a872b7e1fa3cada256f8e97f9e6636f/library/test/src/lib.rs#L150-L158)). This only works because we will offer a direct Std drop-in replacement.

- Instead of re-directing users to `console.log()`, we will offer `js_bindgen_std::print!` instead.
- `js-bindgen-std` has a hook to intercept `print!` calls, which the test runner can use (see [`std::io::set_output_capture()`](https://github.com/rust-lang/rust/blob/3ea2fbcb2a872b7e1fa3cada256f8e97f9e6636f/library/std/src/io/stdio.rs#L1114-L1119)).
- `js_bindgen_std::thread` has a spawn hook, where the test runner can register which thread was spawned by which test (see [`std::thread::add_spawn_hook()`](https://github.com/rust-lang/rust/blob/3ea2fbcb2a872b7e1fa3cada256f8e97f9e6636f/library/std/src/thread/spawnhook.rs#L92-L106)). `libtest` uses the same mechanism to produce the corresponding output when a test fails, instead of showing `print!` output from other unrelated tests.
- We will need a similar hook for spawning `Promise`s, where a Rust closure could also run `print!`.
- Its unclear how we want to handle [unhandled rejections](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event). We could reasonably consider this issue out-of-scope.
  - When running tests in workers, we want to spawn a worker per test, so the global rejection handler does not overlap between tests and we can clearly detect which test triggered the rejection.
  - When running tests in a `Window` or JS engine with `panic = "abort"` we use a Wasm instance per test (to avoid #3687). We could instead use a separate process per test (which would [more closely mimic `libtest`](https://github.com/rust-lang/rust/issues/67650)), but that would degrade testing performance by a lot.
  - When running tests in a `Window` or JS engine with `panic = "unwind"` we intended to have a single Wasm instance, so its even more unclear.
- Add an additional method to [`core::fmt::Formatter`](https://doc.rust-lang.org/1.94.0/core/fmt/struct.Formatter.html) to print `JsValue`s, so they show up nicely in the console. This is important to avoid giving users a reason to fall back to `console.log()` regardless of `print!`s availability.

While WBG doesn't offer a Std replacement, we do still control `web_sys::console::*`. Which means we can implement the same IO hook for our test runner to use. This means however that we need to manually implement `web_sys::console` instead of generating it via WebIDL. It also means that custom bindings generated by non-`web-sys` users wouldn't work.

Correlating threads to tests would be much more complicated. For `wasm_thread` to use it we would have to come up with a stable API, unless we upstream `wasm_thread`.

---

A quick [GitHub search](https://github.com/search?q=%23%5Bwasm_bindgen%28js_namespace+%3D+console%29%5D+language%3ARust&type=code&p=2&l=Rust) shows there are plenty of users re-implementing `console.log()`, which we can't control for. So I believe that given the current state of the ecosystem around WBG, the overall proposal of this PR is probably the better way forward and personally, I can't come up with a better alternative.

I didn't look too close into the implementation, but AFAICS it doesn't handle correlating threads to tests. On the other hand I don't believe the WBG test runner currently correlates output to a specific test in general anyway.

PR #5039, opened 2026-03-15
web-sys: Add missing `cfg` attrubute to `{Open,Save}FilePickerOptions::start_in()`

Tosainu:
### Description

This PR fixes the issue where `OpenFilePickerOptions` and `SaveFilePickerOptions` cannot be enabled without `WellKnownDirectory`.

```
$ RUSTFLAGS=--cfg=web_sys_unstable_apis \
  cargo check --target wasm32-unknown-unknown -p web-sys --features OpenFilePickerOptions --features SaveFilePickerOptions
    Checking web-sys v0.3.91 (/path/to/wasm-bindgen/wasm-bindgen/crates/web-sys)
error[E0425]: cannot find type `WellKnownDirectory` in this scope
   --> crates/web-sys/src/features/gen_OpenFilePickerOptions.rs:149:37
    |
149 |     pub fn start_in(&mut self, val: WellKnownDirectory) -> &mut Self {
    |                                     ^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find type `WellKnownDirectory` in this scope
   --> crates/web-sys/src/features/gen_SaveFilePickerOptions.rs:149:37
    |
149 |     pub fn start_in(&mut self, val: WellKnownDirectory) -> &mut Self {
    |                                     ^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0599]: no method named `set_start_in` found for mutable reference `&mut OpenFilePickerOptions` in the current scope
   --> crates/web-sys/src/features/gen_OpenFilePickerOptions.rs:150:14
    |
150 |         self.set_start_in(val);
    |              ^^^^^^^^^^^^
    |
help: there is a method `get_start_in` with a similar name, but with different arguments
   --> crates/web-sys/src/features/gen_OpenFilePickerOptions.rs:6:1
    |
  6 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^

error[E0599]: no method named `set_start_in` found for mutable reference `&mut SaveFilePickerOptions` in the current scope
   --> crates/web-sys/src/features/gen_SaveFilePickerOptions.rs:150:14
    |
150 |         self.set_start_in(val);
    |              ^^^^^^^^^^^^
    |
help: there is a method `get_start_in` with a similar name, but with different arguments
   --> crates/web-sys/src/features/gen_SaveFilePickerOptions.rs:6:1
    |
  6 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^

Some errors have detailed explanations: E0425, E0599.
For more information about an error, try `rustc --explain E0425`.
error: could not compile `web-sys` (lib) due to 4 previous errors
```

### Checklist


guybedford:
I added a commit that fixes this at the generator level rather than patching the generated files directly.

The root cause was that `DictionaryField::features()` only collected features from the return type (getter), but the deprecated builder method uses the first setter's type as its parameter. Since the getter returns `JsValue` (a builtin that gets filtered out), no `#[cfg(feature = "...")]` gate was generated — even though the builder takes `WellKnownDirectory`.

The fix also refactors doc generation to split `doc_comment()` (description) from `doc_requirements()` (feature gates + unstable API docs), so each setter now gets its own accurate doc requirements instead of sharing the field-level set.

Tosainu:
Thank you so much! I imagined that the root cause was in either IDL or generator but had no idea to fix the problem.

PR #5044, opened 2026-03-18
ts-gen: Experimental TypeScript Generation

guybedford:
### Description
This adds a new `ts-gen` crate, an experiemental TypeScript generation utility for Wasm Bindgen using the new generics support output.

It works very similarly to the webidl generation - expanding function signature overloads and unions based on name suffixing, while matching core js-sys types and supporting getters and setters for interface objects. Variadics are also fully supported.

It supports constrained TypeScript cases currently, with the goal of supporting arbitrary TypeScript files.

It references js_sys where possible, but does not reference web_sys - the goal here being to replace web_sys itself with TypeScript based generation in future.

It also has some extra features including:

* Auto builder creation: Object style dictionary interfaces generate a builder automatically, allows batching the entire object construction process to JS. The builder also validates required properties are set using a bitmask.
* PromiseExt: Generation of a promise extension that provides an `into_future()` method, which is much easier ergonomically than `JsFuture::from(promise)`
* External typing support: External types can be linked using --external definitions.

Comprehensive snapshot tests are included for the generation patterns, as well as an example demonstrating cloudflare workers bindings being generated with the utility.

Higher order generics support is still a work in progress.

To reserve the name on crates.io, this has already now been published under the `ts-gen` crate.

### Checklist

### DIFF
--- a/crates/ts-gen/src/codegen/signatures.rs
+++ b/crates/ts-gen/src/codegen/signatures.rs
@@ -331,11 +331,15 @@ fn flatten_type(ty: &TypeRef, cgctx: Option<&CodegenContext<'_>>, scope: ScopeId
             vec![ty.clone()]
         }
 
-        // Nullable: flatten inner, wrap each in Nullable
-        TypeRef::Nullable(inner) => flatten_type(inner, cgctx, scope)
-            .into_iter()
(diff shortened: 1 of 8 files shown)


guybedford:
Moving this instead to https://github.com/wasm-bindgen/ts-gen.

PR #5047, opened 2026-03-23
fix: Interpreter pops function arguments in reverse order

logan-gatlin:
### Description
When a function is called in the WASM interpreter, arguments are currently popped from the stack into their registers in order. This is incorrect. The WASM calling convention is to push arguments in order at the call site, meaning they must be popped into their registers in reverse order.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@
 
 ### Fixed
 
+* Fixed argument order when calling multi-parameter functions in the
+  `wasm-bindgen` interpreter by reversing the args collected from the stack.
+  [#5047](https://github.com/wasm-bindgen/wasm-bindgen/pull/5047)
+
 * Added support for per-operation `[WbgGeneric]` in WebIDL, restoring typed
   generic return types (e.g. `Promise<ImageBitmap>`) for `createImageBitmap` on
   `Window` and `WorkerGlobalScope` that were lost after the `VideoFrame`


logan-gatlin:
`ureq` update broke MSRV so I pinned it

crates/cli-support/src/interpreter/smoke_tests.rs
```diff
@@ -317,6 +317,35 @@ fn try_table_block() {
...
+                local.get 0
+                call $__wbindgen_describe
+                local.get 1
+                call $__wbindgen_describe
+            )
+
+            (export "foo" (func $foo))
+        )
+    "#;
+    interpret(wat, "foo", &[1, 2]);
```
guybedford, line 346: Can we make sure there's an assert here that fails if the values are not 1 and then 2? e.g. via an unreachable or otherwise. Then we must verify the test fails without the code change.

logan-gatlin: The [interpret](https://github.com/wasm-bindgen/wasm-bindgen/blob/71022038903eda072418c95cc417adc2cd3dba24/crates/cli-support/src/interpreter/smoke_tests.rs#L4) function asserts that `__wbindgen_describe` gets called with the numbers in the array in order. I added a `#[should_panic]` test that intentionally passes the [wrong order](https://github.com/wasm-bindgen/wasm-bindgen/pull/5047/changes/8b3148ec78352c7466df23766c6b2ababb780390#diff-fd15988c1f78487f6c1f5aa1afe80f9976615461df4d29bcdeaf29b70541adf8R350)

LICENSE-APACHE
```diff
@@ -1,4 +1,4 @@
-                              Apache License
+x                              Apache License
```
guybedford: Seems this got included by mistake.

crates/cli-support/src/interpreter/smoke_tests.rs
```diff
@@ -317,6 +317,65 @@ fn try_table_block() {
...
+                local.get 0
+                call $__wbindgen_describe
+                local.get 1
+                call $__wbindgen_describe
+            )
+
+            (export "foo" (func $foo))
+        )
+    "#;
+    interpret(wat, "foo", &[2, 1]);
```
guybedford, line 376: What I mean specifically, is a test case that fails for `interpret(wat, "foo", &[1, 2])` but passes for `interpret(wat, "foo", &[2, 1])` say. That is, an unreachable instruction if seeing a 1 instead of a 2 after a get to be sure we got the right type. Effectively an assertion in WAT that the number is what is expected for the first arg.

PR #5049, opened 2026-03-24
js-sys: implement IntoFuture for Promise, inline wasm-bindgen-futures

guybedford:
## Summary

Implements first-class `Future` support for `js_sys::Promise` by moving the `wasm-bindgen-futures` implementation into `js-sys` behind an optional `futures` feature.

### What this enables

`js_sys::Promise` now implements `IntoFuture`, so any `Promise` can be `.await`ed directly:

```rust
// Before
let result = wasm_bindgen_futures::JsFuture::from(promise).await?;

// After
let result = promise.await?;
```

Typed promises resolve to the correct type without casting:

```rust
let numbers: Array<Number> = fetch_numbers().await?;
```

### How it works

The orphan rule previously prevented `impl IntoFuture for Promise` because `Promise` lived in `js-sys` and `JsFuture` lived in `wasm-bindgen-futures`, with neither owning both. The fix is to move the `wasm-bindgen-futures` implementation into `js-sys`, where `Promise` is defined.

**Dependency graph (unchanged externally):**
```
wasm-bindgen  (unchanged)
    └── js-sys  (gains futures module behind optional feature)
          └── wasm-bindgen-futures  (now a thin re-export shim)
```

The `web-sys` circular dependency (web-sys → js-sys, so js-sys cannot depend on web-sys) is resolved by inlining the two `web-sys` types used in the atomics polyfill (`Worker`, `MessageEvent`) as a private `#[wasm_bindgen] extern "C"` block.

### Feature activation

The `futures` feature on `js-sys` is **off by default**. It is activated automatically when `wasm-bindgen-futures` is in the dependency tree (via `features = ["futures"]` on the `js-sys` dep in `wasm-bindgen-futures/Cargo.toml`). Users who want it without `wasm-bindgen-futures` can add:

```toml
js-sys = { version = "0.3", features = ["futures"] }
```

### Backwards compatibility

- `wasm-bindgen-futures` remains on crates.io and re-exports everything (`JsFuture`, `spawn_local`, `future_to_promise`, `future_to_promise_typed`, `stream`) from `js_sys::futures` unchanged. All existing import paths compile without modification.
- The `futures-core-03-stream` feature on `wasm-bindgen-futures` propagates to `js-sys/futures-core-03-stream`.

### Changes

- **`crates/js-sys/`**: New `src/futures/` module containing the full executor implementation (queue, singlethread/multithread task, wait_async polyfill, stream). Adds optional `cfg-if`, `futures-core`, `futures-util` deps.
- **`crates/futures/`**: Reduced to a re-export shim (`src/lib.rs` ~20 lines). `Cargo.toml` drops all implementation deps.
- **`guide/`**: Updated `js-promises-and-rust-futures.md`, `types/js-sys.md`, and `asynchronous-tests.md` with direct `.await` examples.


logan-gatlin:
Amazing!!

PR #5055, opened 2026-03-25
Emscripten bug fixes

logan-gatlin:
I have been working through [issues](https://github.com/wasm-bindgen/wasm-bindgen/issues/5053) as I encounter them while [porting workers-rs](https://github.com/cloudflare/workers-rs/pull/951) to emscripten. Here are 4 issues I ran into and the patches I made to fix them.

### Change 1
Running `emcc --post-link` on the wasm-bindgen output initially produced the error:
```
let heap_next = heap.length;
                ^

ReferenceError: heap is not defined
```
In `crates/cli-support/js/mod.rs` I see that `expose_global_heap` has an early exit in emscripten mode, while `expose_global_heap_next` does not. I also see that `heap` is pushed to `imports` in `generate_emscritpen_imports`, while `heap_next` is not. When I changed `heap_next` to match `heap` in both of these places (using a placeholder `0` value for the import), this resolved.

### Change 2
```
ReferenceError: __indirect_function_table is not defined
```
This symbol was getting removed in the `unexported_unused_lld_things()` function. My quick fix was just skipping this function on the emscripten target. Maybe we should pass an `is_emscripten` flag here to conditionally keep some symbols?

### Change 3
```
aborted(missing function: $passStringToWasm0)
```
Grepped where that name shows up in code and added a `$` prefix to the `addToLibrary` call. It could also be that the references need to have their `$` prefix removed, but this change resolved the issue so I moved on.

### Change 4
```
warning: undefined symbol: $textEncoder
```
I see in the `expose_text_encoder()` function we check if we are on the emscripten target, and if so, call `self.intrinsic` with an empty string. I am not sure what the intent here is, should an empty string parameter mean nothing is generated? I tried 2 different things (commits `fix4?` and `Fix 4 ammended`), both solved the error but I am not sure which one if either is the correct fix:
1. Early return instead of passing an empty string
2. Check for an empty string inside of `intrinsic()` and skip inserting into `adapter_deps` and exporting to emscripten (the two things that the emscripten path does but other targets don't).

## Remaining Issue
The current error I'm seeing is:
```
HEAP_DATA_VIEW is not defined
```
@google-yfyang mentioned this in the emscripten PR discussion:
> `HEAP_DATA_VIEW` will be actually available when SUPPORT_BIG_ENDIAN or WASM_BINDGEN is set. The relevant change to the Emscripten upstream will land once we can merge this PR. As the result, I don't think we will need this change.

So I am leaving this alone for now.

### Checklist

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -279,9 +279,15 @@ impl<'a> Context<'a> {
 
         if matches!(self.config.mode, OutputMode::Emscripten) {
             let actual_js_name: &str = js_name.unwrap_or(&name);
-            self.adapter_deps.insert(actual_js_name.to_string());
 
-            self.export_to_emscripten(actual_js_name, &val);
+            // Only add as a dependency and export if there's actual content.
+            // Empty content means the intrinsic is handled elsewhere (e.g. via
+            // emscripten_global_deps), so registering a dep here would create a
+            // reference to a non-existent library variable.
+            if !val.trim().is_empty() {
+                self.adapter_deps.insert(actual_js_name.to_string());
+                self.export_to_emscripten(actual_js_name, &val);
+            }
 
             // Register empty string so standard generation skips this key
             self.intrinsics.as_mut().unwrap().insert(name, "".into());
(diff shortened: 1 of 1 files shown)


guybedford:
@google-yfyang @walkingeyerobot PTAL. This gets things running for us on workers - would be good to get these fixes in for the release this week if possible!

PR #5056, opened 2026-03-25
macro: auto-inject MaybeUnwindSafe bound for &dyn Fn / &mut dyn FnMut import args

guybedford:
Implements automatic `MaybeUnwindSafe` enforcement for raw `&dyn Fn` / `&mut dyn FnMut` closure arguments in `#[wasm_bindgen]` imports, ensuring unwind safety when building with `panic=unwind` while remaining fully backward compatible on all other targets (where `MaybeUnwindSafe` has a blanket impl for all types).

When a JS import takes a stack-borrowed closure, the macro previously erased directly to the `dyn` trait object with no unwind safety guarantee. This change makes the generated wrapper enforce the bound at the call site using `impl Trait` syntax, so the constraint is visible directly in the argument position in rustdoc and IDE hovers.

* `detect_raw_fn_trait_obj` identifies `&dyn Fn` / `&mut dyn FnMut` import arguments
* the generated wrapper rewrites the argument to use `impl Trait` with `MaybeUnwindSafe` inline
* the ABI and wire format are unchanged — the argument is still coerced to `&mut dyn FnMut(...)` for `into_abi`
* all affected call sites in js-sys, the test harness, and the wasm test suite wrap non-`UnwindSafe` captured values in `AssertUnwindSafe` before capture

```rs
pub fn try_every_result_aborting(&self, predicate: &mut dyn FnMut(JsValue, u32, Array) -> Result<bool, JsValue>) -> Result<bool, JsValue>
```

```rs
pub fn try_every_result_aborting(
    &self,
    predicate: &mut (impl FnMut(JsValue, u32, Array) -> Result<bool, JsValue>
        + wasm_bindgen::__rt::marker::MaybeUnwindSafe),
) -> Result<bool, JsValue>
```

Tested with `just test-wasm-bindgen-unwind` (351 passed) and the standard `panic=abort` suite (338 passed).

Call sites that capture non-`UnwindSafe` values (e.g. `&mut T`, `Cell<T>`) under `panic=unwind` must wrap them in `AssertUnwindSafe` before capture — the guide has been updated to document this pattern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@
   representations such as `half::f16`.
   [#5033](https://github.com/wasm-bindgen/wasm-bindgen/pull/5033)
 
+### Changed
+
+* The `#[wasm_bindgen]` macro now emits `&mut (impl FnMut(...) + MaybeUnwindSafe)`
+  / `&(impl Fn(...) + MaybeUnwindSafe)` for raw `&mut dyn FnMut` / `&dyn Fn`
+  import arguments instead of a hidden generic parameter and where-clause. The
+  generated signature is cleaner and the `MaybeUnwindSafe` bound is visible
+  directly in the argument position. The ABI and wire format are unchanged.
+  When building with `panic=unwind`, closures that capture non-`UnwindSafe`
+  values (e.g. `&mut T`, `Cell<T>`) must wrap them in `AssertUnwindSafe` before
+  capture; on all other targets `MaybeUnwindSafe` is a no-op blanket impl.
+  [#5056](https://github.com/wasm-bindgen/wasm-bindgen/pull/5056)
+
 ## [0.2.115](https://github.com/rustwasm/wasm-bindgen/compare/0.2.114...0.2.115)
 
 ### Added


guybedford:
//cc @RReverser I think this is finally the right approach for this, would appreciate your thoughts.

guide/src/reference/passing-rust-closures-to-js.md
```diff
@@ -62,7 +61,43 @@ extern "C" {
...
 
-For longer-lived closures, or for closures that support checking unwind safety, use `ScopedClosure` instead.
+### Unwind Safety
+
+The `#[wasm_bindgen]` macro automatically injects a `MaybeUnwindSafe` bound on
+the closure argument for each `&dyn Fn` and `&mut dyn FnMut` parameter. The
+generated signature is equivalent to:
+
+```rust
+fn takes_mut_closure<__WbgCb0: MaybeUnwindSafe + FnMut()>(f: &mut __WbgCb0);
```
RReverser, line 71: Why is the change to generics necessary - adding the bound so that it's just `&mut (dyn FnMut() + MaybeUnwindSafe)` seems simpler implementation wise and less invasive?

guybedford: I did first try that but it gives this error:

```
error[E0225]: only auto traits can be used as additional traits in a trait object
```

Further explanation from Claude:

> The issue is straightforward: `MaybeUnwindSafe` is not an auto trait, so Rust forbids it as an additional bound on a trait object.

> In Rust, a `dyn` type can only have one non-auto trait, plus any number of auto traits (like Send, Sync, Unpin). Auto traits are special compiler-intrinsic traits. `MaybeUnwindSafe` is a regular user-defined trait in `wasm_bindgen::__rt::marker`, not an auto trait.

> So when the macro generates `fn takes_mut_closure(f: &mut (dyn FnMut() + MaybeUnwindSafe))`, the compiler sees two non-auto traits `(FnMut() and MaybeUnwindSafe)` in the same dyn object, and rejects it.

> This is a fundamental language constraint - not something we can work around. The only way to add a `MaybeUnwindSafe` requirement on the closure is to do what the PR already does: introduce a generic type parameter with both bounds: `fn takes_mut_closure<__WbgCb0: MaybeUnwindSafe + FnMut()>(f: &mut __WbgCb0)`.

> This works because generic bounds don't have the "one non-auto trait" restriction that dyn objects do. The generic is then coerced to `&mut dyn FnMut()` internally for the ABI, so the wire format is unchanged.

> Even `std::panic::UnwindSafe` itself is not an auto trait (it just has blanket impls), which is why you can't write `&dyn FnMut() + UnwindSafe` either. The generic-parameter approach is the standard pattern the Rust ecosystem uses when you need to combine `Fn*` bounds with non-auto marker traits.

RReverser: Ah fair enough. Although it's wrong on the last count:

```rust
use std::panic::UnwindSafe;

pub fn foo(f: &mut (dyn FnMut() + UnwindSafe)) {

}
```

compiles just fine. I remember I tried it last time we talked about this stuff.

Does that help us here? As in, can we use `UnwindSafe` instead of `MaybeUnwindSafe`?

RReverser: If not, another also slightly cleaner alternative could be to just use

```rust
use std::panic::UnwindSafe;

pub fn foo(f: &mut (impl FnMut() + MaybeUnwindSafe)) {

}
```

(`impl` instead of `dyn`, which is essentially generics but without changing function signature)

But if `UnwindSafe` is enough, that would be still nicer to avoid generics in favour of `dyn` altogether, since each usage of generics requires monomorphisation and increases Wasm size for no reason - our closures are already `dyn`.

guybedford: In the macro expansion we don't know if we're building with panic=unwind though? So we'd need to branch all generations to have both an unwind and non-unwind variation which seems costly.

The backwards compatibility concern is the other constraint of course, given how simple this syntax is it is hard to replace.

Remember that the current version assumes unwind safety incorrectly so this PR is fixing a genuine soundness issue...

RReverser: > In the macro expansion we don't know if we're building with panic=unwind though? So we'd need to branch all generations to have both an unwind and non-unwind variation which seems costly.

Ah yeah we can't blanket-implement it because it's a built-in trait, and we can't make ours behave like a "marker". Sigh.

> If not, another also slightly cleaner alternative could be to just use

Let's try this one then? At least it will be more (backward-)compatible, eg not interfering with the list of real generics.

Otherwise, IIUC, with current approach + generics we've added recently something like `fn foo<T>(value: T, cb: &dyn Fn());` would now become `fn foo<T, __WbgCb0: ...>(value: T, cb: &dyn  __WbgCb0)` which breaks `foo::<MyType>()` syntax due to generics list mismatch which would be a pretty unintuitive error.

Using `impl` syntax like `fn foo<T>(value: T, cb: &(impl Fn() + MaybeUnwindSafe))` is essentially same as your approach with generics, but without this compatibility pitfall.

guybedford: Yeah this is a **lot** simpler, thanks for the suggestion. Resulted in a lot of code removed!

RReverser: Since you merged this already, please fix the outdated doc and comments in a followup PR.

guybedford: Thanks for spotting that, including this in the release PR at https://github.com/wasm-bindgen/wasm-bindgen/pull/5070 now.

PR #5058, opened 2026-03-26
Add AGENTS.md

guybedford:
Adds an `AGENTS.md` file to the repo root.

This file is picked up by AI coding tools (like OpenCode, Claude Code, etc.) to give them project-specific context — things like the crate structure, how to run tests, code style rules, and the release process. The idea is to reduce the chance of AI tools doing something obviously wrong (wrong test command, wrong format style, etc.) when working in this repo.

The content is mostly drawn from `CONTRIBUTING.md` and the `justfile`, so nothing new is being documented here — it's just collected in one place that these tools know to look for.

Happy to adjust the content or drop sections if anything looks off.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -53,6 +53,11 @@ Set `WASM_BINDGEN_KEEP_TEST_BUILD=1` to retain the temp build folder on test fai
 
 Add an entry to `CHANGELOG.md` for any user-facing change.
 
+## PR Process
+
+- PRs must come with sufficient tests
+- After posting the PR, add a CHANGELOG.md entry, this is a PR requirement
+
 ## Release PR
 
 1. Bump all crate versions:


AGENTS.md
```diff
@@ -0,0 +1,80 @@
...
+
+Set `WASM_BINDGEN_KEEP_TEST_BUILD=1` to retain the temp build folder on test failure.
+
+## Code Style
+
+- Format: `cargo fmt --all` (enforced in CI)
+- Lint: `cargo clippy --all-features --workspace -- -D warnings`
+- Use inlined format args: `format!("{x}")` not `format!("{}", x)`
+- TOML: formatted with `taplo fmt` (see `taplo.toml`)
+- Edition: 2021. MSRV: 1.71 (libraries), 1.82 (CLI)
```
logan-gatlin: The edition and MSRV will become outdated. Agents typically read the `Cargo.toml` first anyways, so I would suggest dropping this

drewcrawford:
If there's broader interest, I have docs I can contribute on this for common workflows like bless, debugging and triage steps for specific kinds of bugs, architectural overviews of some components, etc.

As a general rule I've found it **highly** advantageous to take specific workflows (like releasing in this case) and pull them into their own doc which is linked from AGENTS.md with a good description (this is effectively just the skills pattern).  This enables writing highly detailed workflow instructions (sometimes with bespoke scripts etc) without polluting the context window since the agent decides what to read.

AGENTS.md
guybedford: Removed.

guybedford:
@drewcrawford context windows have gotten pretty large recently so I don't think it's a problem to extend this doc with sections as needed.

PR #5059, opened 2026-03-26
Add wasm_bindgen::handler — hard-abort callbacks and reinit support

guybedford:
Adds `wasm_bindgen::handler` — two experimental hooks for `panic=unwind` builds.

**`set_on_abort(fn()) -> Option<fn()>`**

Registers a callback fired from `__wbg_handle_catch` in the generated JS glue immediately after the instance's terminated flag is set on a hard abort (`unreachable`, OOM, stack overflow). Returns the previously registered handler (`None` if unset), mirroring `std::panic::set_hook`.

The terminated flag is set before the handler fires so any re-entrant export call is immediately blocked. The handler is wrapped in `try/catch` so a throwing or panicking handler cannot suppress the original error. Stored as a Wasm indirect-function-table index (`u32` in a `#[no_mangle] static`) — the table lives outside linear memory, safe to dispatch through even when the heap is corrupt. The JS glue reads the address via the exported global, dereferences into linear memory to get the index, and calls through the table via a shared `__wbg_invoke_handler(addr)` JS helper.

**`reinit() / set_on_reinit(fn()) -> Option<fn()>`**

`reinit()` writes a sentinel (`0xFFFFFFFF`) into `__instance_terminated`. The next export call detects this in `__wbg_termination_guard()`, calls `__wbg_reset_state()` which creates a fresh `WebAssembly.Instance`, then dispatches the reinit callback through the same `__wbg_invoke_handler` helper on the new instance.

Requires `--experimental-reset-state-function`. Because a new instance resets all Rust statics, the callback should be re-registered via `#[wasm_bindgen(start)]`.

Both features are `#[doc(hidden)]` as experimental.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,14 @@
 * Added `VideoFrame`, `VideoColorSpace`, and related WebCodecs dictionaries/enums to `web-sys`.
   [#5008](https://github.com/wasm-bindgen/wasm-bindgen/pull/5008)
 
+* Added `wasm_bindgen::handler` module with `set_on_abort` and `set_on_reinit`
+  hooks for `panic=unwind` builds. `set_on_abort` registers a callback invoked
+  after the instance is terminated (hard abort, OOM, stack overflow).
+  `set_on_reinit` registers a callback invoked after `reinit()` resets the
+  WebAssembly instance via `--experimental-reset-state-function`. Handlers are
+  stored as Wasm indirect-function-table indices so dispatch is safe even when
+  linear memory is corrupt.
+
 ### Changed
 
 * Replaced per-closure generic destructors with a single `__wbindgen_destroy_closure`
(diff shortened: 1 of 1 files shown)


logan-gatlin:
Why does it assign to a global variable instead of using an attribute macro on a function, like `#[wasm_bindgen(on_abort)]`? With this method if you set the callback twice, the first instance gets silently dropped, but with a macro we could give a compiler error

guybedford:
I've updated the handler to return the old handler when registering one. I've also significantly simplified the implementation by restricting it to panic=unwind only for now. In a future change we can extend this experimental work to panic=abort rather. The reason for using the table is explained in the comments - basically we are more likely to be able to call it even if all other state is borked.

Finally, I've significantly simplified the PR to ensure this is minimal, focused and correct (diff down from 2000+ to 600 now).

logan-gatlin:
Since the hook takes an `fn` instead of `impl Fn` it will be difficult to sequence / chain multiple callbacks like what panic hooks allow for. I think this way is still reasonable though. I can make this work for my original use case in workers-rs.

Issue #5065, opened 2026-03-28
0.2.115: Failing tests due to ordering issues on Alpine


kpcyrd:
Hello!

When updating the wasm-bindgen Alpine package (https://gitlab.alpinelinux.org/kpcyrd/aports/-/jobs/2277385) I noticed some tests are failing because the generated web assembly is not identical with what was expected by the test.

As far as I can tell the generated web assembly is semantically correct, but in one case `__wbindgen_malloc` was put into slot 0 and `__wbindgen_realloc` in slot 1, in the other case they are backwards (but the only place it's used on is using the correct slot index, therefore it's semantically correctly, but not deterministic across upstream's Rust compiler and Alpine's Rust compiler.

```diff
---- reference::runtest::test_06_async_void_rs stdout ----
thread 'reference::runtest::test_06_async_void_rs' (7267) panicked at crates/cli/tests/wasm-bindgen/reference.rs:121:13:
assertion failed: `(left == right)`
Diff < left / right > :
 (module $async_void_reftest.wasm
<  (type (;0;) (func (param i32 i32) (result i32)))
<  (type (;1;) (func (param i32 i32 i32 i32) (result i32)))
>  (type (;0;) (func (param i32 i32 i32 i32) (result i32)))
>  (type (;1;) (func (param i32 i32) (result i32)))
   (type (;2;) (func (result i32)))
   (type (;3;) (func (param i32)))
   (type (;4;) (func (param i32 i32)))
   (type (;5;) (func))
   (type (;6;) (func (param i32 i32) (result externref)))
   (type (;7;) (func (param externref externref externref) (result externref)))
   (type (;8;) (func (param externref) (result externref)))
   (type (;9;) (func (param externref externref) (result externref)))
   (type (;10;) (func (param externref i32 i32) (result i32)))
   (type (;11;) (func (param externref)))
   (type (;12;) (func (param externref) (result i32)))
   (type (;13;) (func (param i32 externref)))
   (type (;14;) (func (result externref)))
   (type (;15;) (func (param i32 i32 externref externref)))
   (type (;16;) (func (param i32 i32 externref) (result i32 i32)))
   (import "./reference_test_bg.js" "__wbg___wbindgen_debug_string_d89627202d0155b7" (func (;0;) (type 13)))
   (import "./reference_test_bg.js" "__wbg___wbindgen_is_function_2a95406423ea8626" (func (;1;) (type 12)))
[...]

PR #5068, opened 2026-03-30
deps: revert walrus to 0.25.2

guybedford:
## Summary

Reverts the walrus upgrade from #5057 due to a regression introduced in walrus 0.26 tracked in #5065, which was caused by https://github.com/wasm-bindgen/walrus/pull/304.

Also reverts the associated wasmparser/wasmprinter version bumps back to 0.240.


guybedford:
This fixes #5065. The root cause is a non-deterministic type ordering regression introduced in walrus 0.26 via https://github.com/wasm-bindgen/walrus/pull/304. Walrus began using a HashMap for type deduplication which produces different orderings depending on hash randomization — the generated wasm is semantically correct but not deterministic across compiler invocations, causing reference test failures on Alpine Linux.

guybedford:
Trying the new 0.26.1 first. If we still have issues after this patch, then we can reconsider this revert.

Issue #5076, opened 2026-04-01
CLI panics with "failed to find N in function table" on Leptos WASM builds (rustc 1.94+)


ksr-13:
## The Bug

The `wasm-bindgen` CLI panics when processing WASM files generated by rustc 1.94+ (both stable and nightly) for projects using Leptos 0.8.x. The error occurs in `outgoing.rs:257` when trying to look up a function table entry that doesn't exist.

`thread 'main' panicked at crates/cli-support/src/wit/outgoing.rs:257:61: called Result::unwrap() on an Err value: failed to find 33345 in function table`

## Steps to Reproduce

1. Create a Leptos 0.8.x project with wasm-bindgen
2. Build with `cargo build --target wasm32-unknown-unknown --release`
3. Run `wasm-bindgen --target=web --out-dir=out app.wasm`

## Expected Behavior

wasm-bindgen should successfully generate JS bindings.

## Actual Behavior

Panics with "failed to find N in function table" where N is a large index (e.g., 32752, 33309, 33345).

## Environment

- **wasm-bindgen version**: 0.2.115, 0.2.116, 0.2.117 (all affected)
- **rustc version**: 1.94.1 (stable), nightly-2026-03-20
- **Leptos version**: 0.8.17
- **OS**: macOS aarch64
- **Trunk version**: 0.21.14

## Additional Context

- The error persists across wasm-bindgen CLI versions 0.2.114-0.2.117
- A minimal wasm-bindgen project works fine - the issue is specific to larger projects like Leptos
- The WASM file is ~2.2MB with many closures (Leptos uses many reactive closures)
- Related to PR #4681 which changed closure table element handling
- No existing GitHub issues match this exact error pattern

## Reproduction

Key dependencies:
```toml
leptos = { version = "0.8.17", features = ["csr"] }
wasm-bindgen = "0.2.117"

Build command:
cargo +stable build --target wasm32-unknown-unknown --release
wasm-bindgen --target=web --out-dir=dist target/wasm32-unknown-unknown/release/web-app.wasm

The error occurs in export_table_element() → get_function_table_entry() which uses walrus to find the function ID for a given table index, then looks for an export matching that function. The function table entry doesn't exist at the expected index.

guybedford:
Thanks for sharing - supporting full const expr analysis in the table lookup per #5077 should resolve this, going out in the next release. May yet take a couple of weeks or so for the next release to roll out though.

PR #5077, opened 2026-04-01
fix: handle element segment offsets and interpreter global tracking for rustc 1.94+

guybedford:
This fixes two CLI issues affecting WASM modules built with rustc 1.94+.

## Element segment offset evaluation (#5076)

rustc 1.94+ / lld changed how it emits element segment offsets for large function tables. Instead of a plain `i32.const N`, lld now emits `global.get $__table_base` or `global.get $__table_base; i32.const K; i32.add` (extended const expressions). The previous `get_function_table_entry` only matched `ConstExpr::Value(Value::I32(n))` and skipped everything else, so the loop found nothing and panicked with `failed to find N in function table`.

* Inlines `evaluate_const_expr`, a small stack-machine evaluator covering `Value`, `Global` (resolved via the module's local globals), and the arithmetic subset of `Extended`. Returns `None` for anything that can't be statically resolved.
* Replaces `(idx - offset) as usize` with `idx.checked_sub(offset)` to guard against integer underflow in multi-segment tables.
* Adds 9 unit tests covering all three offset variants, the error path, and the multi-segment underflow case.

## Interpreter global tracking (#5080)

The descriptor interpreter previously treated all `global.get` / `global.set` instructions as stack pointer accesses. This produced incorrect descriptor results for modules containing additional globals (e.g. `GOT.func.internal.*`).

* Resolves the `__stack_pointer` export during initialization and tracks other globals independently in a `HashMap`.
* Falls back to the old "all globals are the stack pointer" behavior for modules without the `__stack_pointer` export, maintaining backward compatibility.
* Updates existing smoke tests to use named globals and `__stack_pointer` exports.
* Adds a `multiple_globals_with_stack_pointer` test verifying the interpreter reads and writes non-stack-pointer globals correctly.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -14,10 +14,15 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CARGO_TERM_COLOR: always
+      # Pinned due to __llvm_profile_runtime regression in newer nightlies
+      RUSTUP_TOOLCHAIN: nightly-2026-02-18
     steps:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@nightly
         with:
+          # Pinned due to https://github.com/nickel-lang/minicov/issues/5
+          # __llvm_profile_runtime symbol removal in newer nightlies
+          toolchain: nightly-2026-02-18
           targets: wasm32-unknown-unknown
       - uses: taiki-e/install-action@just
       - uses: taiki-e/install-action@cargo-llvm-cov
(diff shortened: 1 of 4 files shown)


hoodmane:
> `GOT.func.internal`

Are we trying to add support for dynamic linking here?

guybedford:
> Are we trying to add support for dynamic linking here?

Yeah the comment seems unnecessary. Not specifically, just to resolve the regressions in using global initializers in the latest Rust toolchain.

PR #5081, opened 2026-04-02
feat: support multiple and private `#[wasm_bindgen(start)]` functions

guybedford:
Adds support for multiple start functions by chaining them together when registered, including an end-to-end CLI test to verify execution order.

Also adds support for `#[wasm_bindgen(start, private)]` to declare private start functions which are not exported by the Wasm exports. This one implies a schema bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,25 @@
 ### Added
 
 * Added support for multiple `#[wasm_bindgen(start)]` functions, which are
-  chained together at initialization.
+  chained together at initialization, as well as a new
+  `#[wasm_bindgen(start, private)]` to register a start function without
+  exporting it as a public export.
   [#5081](https://github.com/wasm-bindgen/wasm-bindgen/pull/5081)
 
-* Added `#[wasm_bindgen(start, private)]` to register a start function without
-  exporting it to JS.
-  [#5081](https://github.com/wasm-bindgen/wasm-bindgen/pull/5081)
+### Fixed
+
+* Fixed two CLI issues affecting WASM modules built by rustc 1.94+. First,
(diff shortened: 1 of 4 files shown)


hoodmane:
How do you tell what order multiple start functions should run in?

guybedford:
We basically have zero ordering guarantees.

hoodmane:
Is it fair to call it a start function anymore then? I think that is more like a C constructor. Does Rust allow similar things on other platforms?

guybedford:
The use case is more about the JS embedding - JavaScript modules participate in initialization steps that are able to interact with the host environment. This gives that capacity to separate Rust code, without requiring a singular coordination point, just like modules in JS don't need to register their top-level initializations to access things like globals.

guybedford:
I'm open to giving it a new alias, but this does match the `WebAssembly` `start` function semantics in the JS ESM integration.

guide/src/reference/attributes/on-rust-exports/start.md
```diff
@@ -17,13 +17,39 @@ limitations today the start section of the executable may not literally point to
 `start`, but the `start` function here should be started up automatically when the
 wasm module is loaded.
 
-There's a few caveats to be aware of when using the `start` attribute:
+## Multiple start functions
+
+Multiple `#[wasm_bindgen(start)]` functions can be specified across a module and
+its dependencies. They will be chained together and all executed during
+initialization.
```
hoodmane: Mention that execution order is arbitrary?

hoodmane (approved):

guide/src/reference/attributes/on-rust-exports/start.md
guybedford: Added.

PR #5083, opened 2026-04-03
fix: auto-detect schedule_reinit and tighten termination guard

guybedford:
This disables the auto-reinit functionality of aborts with `--experimental-reset-state-function` to instead require an explicit call to the `schedule_reinit()` function. In addition, any use of `schedule_reinit()` will automatically bring in the reinit machinery by treating it as an intrinsic.

`schedule_reinit()` sets a private JS-side flag instead of writing to a `__instance_terminated` sentinel value. This means:

* `schedule_reinit()` can be called during normal execution without triggering a trap — the current call completes normally and the reset happens on the next export call
* the abort-then-reinit path still works: the abort hook fires, calls `schedule_reinit()`, and the guard resets instead of throwing
* `__instance_terminated` is a clean boolean (`0` = live, `1` = terminated) with no internal sentinel values leaked into the public spec

Usage for abort on reinit:

```rs
fn on_abort() {
    wasm_bindgen::handler::schedule_reinit();
}

#[wasm_bindgen(start)]
fn start() {
    wasm_bindgen::handler::set_on_abort(on_abort);
    wasm_bindgen::handler::set_on_reinit(on_reinit);
}
```

A new guide section is added "Handling Aborts" covering these behaviours as a separate guide page to "Catching Panics".

What changed:

* renamed `reinit()` to `schedule_reinit()` to reflect that it schedules reinitialization for the next export call
* `__wbindgen_reinit` intrinsic now sets a JS variable (`__wbg_reinit_scheduled`) instead of writing `-1` to linear memory
* `__wbg_handle_catch` unconditionally writes `1` (no conditional to preserve a sentinel)
* `__wbg_reset_state` only references `__wbg_called_abort` / `__wbg_reinit_scheduled` when the catch handler machinery is active, fixing a `ReferenceError` on `panic=abort` builds with `--experimental-reset-state-function`
* auto-detection of `schedule_reinit()` usage emits private reinit machinery without `--experimental-reset-state-function`
* updated guide: `__instance_terminated` documented as boolean, `schedule_reinit` docs note it works outside abort context
* remove `on_reinit` for now, now that we support multiple start functions instead

Test coverage: `termination`, `termination_abort_handler`, `termination_reset_state`, `termination_reinit`, and `termination_reinit_auto_detect` all pass.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -20,18 +20,13 @@
 //!
 //! [`schedule_reinit()`] writes a sentinel value into the termination flag.
 //! The next call to any export detects this, creates a fresh
-//! `WebAssembly.Instance` from the same module, and then calls
-//! [`set_on_reinit`]'s callback (if any) on the new instance.
+//! `WebAssembly.Instance` from the same module.
 //!
 //! The reinit machinery is automatically emitted when [`schedule_reinit()`] is
 //! used — no CLI flag is required. `--experimental-reset-state-function` is
 //! only needed for the public `__wbg_reset_state()` export.
 //!
-//! Use [`set_on_reinit`] to register a reinit callback; it likewise returns
-//! the previously registered handler.
-//!
-//! [`schedule_reinit()`] and [`set_on_reinit`] are no-ops on `panic=abort`
-//! builds.
+//! [`schedule_reinit()`] is a no-op on `panic=abort` builds.
 #[doc(hidden)]
 pub use crate::__rt::schedule_reinit;
 #[doc(hidden)]


crates/cli/src/wasm_bindgen.rs
```diff
@@ -93,6 +93,7 @@ struct Args {
         help = "Generate __wbg_reset_state function for Wasm reinitialization (experimental)"
     )]
     generate_reset_state: bool,
+
```
hoodmane: Stray white space change

PR #5099, opened 2026-04-08
feat: support schedule_reinit() under panic=abort builds

guybedford:
## Summary

Previously `handler::schedule_reinit()` was a no-op under `panic=abort`. This extends it to work with both panic strategies.

Under `panic=abort` there's no `__instance_terminated` flag or wasm-level catch wrappers, so the full abort detection machinery doesn't apply. But `schedule_reinit()` is independent of abort detection — it just sets a JS-side flag (`__wbg_reinit_scheduled`) that's checked before each export call. This PR adds that minimal path.

* Remove `#[cfg(panic = "unwind")]` gate on `schedule_reinit()` so it calls `__wbindgen_reinit()` in all builds
* Add `generate_reinit_guard()` — emits a simple `__wbg_termination_guard` that only checks `__wbg_reinit_scheduled` (no try/catch, no `__instance_terminated`)
* Refactor `maybe_wrap_try_catch` into `maybe_wrap_export_call` with separate `check_aborted` / `check_reinit` flags — exports under `panic=abort` get the guard prepended without try/catch wrapping
* Fix `__wbg_reset_state` to reset `__wbg_reinit_scheduled` in all builds (not just when `wrapped_js_tag` is present)
* Add `reinit_panic_abort` integration test: builds with `-Cpanic=abort -Zbuild-std=std,panic_abort`, verifies new instance creation, static reset, persistence without signal, and multiple reinit cycles

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -96,10 +96,21 @@ pub enum TsReference {
     StringEnum(String),
 }
 
-pub fn wrap_try_catch(call: &str) -> String {
+/// Whether and how to guard an export call against instance termination.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ExportGuard {
+    /// No guard wrapping needed.
+    None,
+    /// Wrap with termination guard and try-catch for panic=abort with error handling.
+    Abort,
+    /// Wrap with just termination guard for schedule_reinit() support.
+    GuardOnly,
+}
+
+fn wrap_try_catch(call: &str) -> String {
     format!(
         "\
-        __wbg_termination_guard();
+        __wbg_call_guard();
         try {{
             {call};
         }} catch(e) {{
@@ -109,11 +120,20 @@ pub fn wrap_try_catch(call: &str) -> String {
     )
 }
 
-pub fn maybe_wrap_try_catch(call: &str, should_check_aborted: bool) -> String {
-    if should_check_aborted {
-        wrap_try_catch(call)
-    } else {
-        format!("{call};")
+fn wrap_call_guard(call: &str) -> String {
+    format!(
+        "\
+        __wbg_call_guard();
+        {call};
+        "
+    )
+}
+
+pub fn maybe_wrap_export_call(call: &str, guard: ExportGuard) -> String {
+    match guard {
+        ExportGuard::Abort => wrap_try_catch(call),
+        ExportGuard::GuardOnly => wrap_call_guard(call),
+        ExportGuard::None => format!("{call};"),
     }
 }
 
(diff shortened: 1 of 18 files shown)


crates/cli-support/src/js/binding.rs
```diff
@@ -109,9 +109,20 @@ pub fn wrap_try_catch(call: &str) -> String {
...
+fn wrap_reinit_guard(call: &str) -> String {
+    format!(
+        "\
+        __wbg_termination_guard();
+        {call};
+        "
+    )
+}
+
+pub fn maybe_wrap_export_call(call: &str, check_aborted: bool, check_reinit: bool) -> String {
```
hoodmane: Should we combine `check_aborted` and `check_reinit` into an enum? What does the both true case mean?

hoodmane: Oh yeah it's not allowed. Probably would be good to change it to an enum.

crates/cli-support/src/js/binding.rs
```diff
@@ -924,6 +945,12 @@ fn instruction(
                             {}",
                             &wrap_try_catch(&format!("ret = {call};"))
                         )
+                    } else if should_check_reinit {
+                        format!(
+                            "\
+                            __wbg_termination_guard();
+                            const ret = {call};"
+                        )
```
hoodmane, line 946: The duplication here is not great since it could easily drift. How about just replacing this whole thing with:
```rust
format!(
    "\
    let ret;
    {}",
    &maybe_wrap_export_call(&format!("ret = {call};", should_check_aborted, should_check_reinit))
)
```

crates/cli-support/src/js/binding.rs
```diff
@@ -1429,7 +1456,7 @@ fn instruction(
                             format!(
                                 "\
                                 this.__wbg_ptr = {val} >>> 0;
-                                this.__wbg_inst = __wbg_instance_id;
+                                Object.defineProperty(this, '__wbg_inst', {{ value: __wbg_instance_id, writable: true }});
```
hoodmane, line 1453: So the goal here is to make `__wbg_inst` not enumerable and not configurable? Put a comment saying that?

crates/cli/tests/wasm-bindgen/main.rs
```diff
@@ -1628,6 +1628,125 @@ describe('reinit auto-detection (no --experimental-reset-state-function)', () =>
         .success();
 }
 
+/// Tests that schedule_reinit() works under panic=abort builds.
+/// No abort handler or __instance_terminated — just the reinit flag + guard.
```
hoodmane:
```suggestion
```

hoodmane (approved):
LGTM aside from a few suggestions.

crates/cli-support/src/js/binding.rs
guybedford: We do want to support the union case in future here - https://github.com/wasm-bindgen/wasm-bindgen/issues/5090 exactly tracks that.

guybedford: Actually I've simplified this to unify all the paths without a bool function like this.

crates/cli-support/src/js/binding.rs
guybedford: The problem is that changes the output in the non-wrapped case. I think better to keep it the same there and keep the branch.

crates/cli-support/src/js/binding.rs
guybedford: Added thanks.

PR #5103, opened 2026-04-09
fix: hoist ES module import statements to top of generated JS

guybedford:
The generated JS output previously had `import` declarations scattered across three locations: the JS import header near the top, wasm `import * as` statements mid-file, and wasm loading imports (like `import source` or `import { readFileSync }`) at the bottom.

While ES engines hoist imports semantically, having them scattered makes output harder to read and can trip up tooling that expects imports at the top of the file.

This adds a `hoist_es_imports()` post-processing step in `finalize()` that collects all `import` lines and emits them together at the top of the file, right after the `@ts-self-types` directive when present. It runs for all ES module output modes (Module, Web, Deno, Node ESM, Bundler).

Before:
```js
/* @ts-self-types="./foo.d.ts" */
import { default as _default } from 'import_class.js';

export function exported() { ... }
import * as import1 from "imports.js"

function __wbg_get_imports() { ... }
...
import source wasmModule from "./foo_bg.wasm";
const wasmInstance = new WebAssembly.Instance(wasmModule, __wbg_get_imports());
```

After:
```js
/* @ts-self-types="./foo.d.ts" */
import { default as _default } from 'import_class.js';
import * as import1 from "imports.js"
import source wasmModule from "./foo_bg.wasm";

export function exported() { ... }

function __wbg_get_imports() { ... }
...
const wasmInstance = new WebAssembly.Instance(wasmModule, __wbg_get_imports());
```

All 45 reference tests pass, plus the cli-support unit tests. Snapshot updates are included via `BLESS=1`.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -41,6 +41,10 @@ macro_rules! region {
 
 pub struct Context<'a> {
     globals: String,
+    /// ES module `import` statements collected during codegen, emitted at the
+    /// top of the output in `finalize()`. In bundler mode these go into the
+    /// `_bg.js` file; the bundler entry file has its own hardcoded imports.
+    es_module_imports: String,
     intrinsics: Option<BTreeMap<Cow<'static, str>, Cow<'static, str>>>,
     emscripten_library: String,
     imports_post: String,
@@ -224,6 +228,7 @@ impl<'a> Context<'a> {
     ) -> Result<Context<'a>, Error> {
         Ok(Context {
             globals: String::new(),
+            es_module_imports: String::new(),
             intrinsics: Some(Default::default()),
             imports_post: String::new(),
             export_name_list: Vec::new(),
@@ -465,10 +470,7 @@ impl<'a> Context<'a> {
         }
 
         if !self.js_imports.is_empty() {
-            region!(self, "js imports", {
-                let imports = self.js_import_header()?;
-                self.globals.push_str(&imports);
-            });
+            self.js_import_header()?;
         }
 
         if !self.exports.is_empty() {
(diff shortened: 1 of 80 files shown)


crates/cli-support/src/js/mod.rs
```diff
@@ -5890,3 +5894,54 @@ impl fmt::Display for MemView {
...
+
+/// Hoists all ES module `import` declarations to the top of the generated JS,
+/// placed right after the `@ts-self-types` directive if present.
+fn hoist_es_imports(js: &str) -> String {
+    let lines: Vec<&str> = js.lines().collect();
+    let mut imports = Vec::new();
+    let mut rest = Vec::new();
+
+    for &line in &lines {
+        if line.trim_start().starts_with("import ") {
```
hoodmane: It's in principle possible for the line to be a dynamic import eg `import ("something").then(...)`

crates/cli-support/src/js/mod.rs
```diff
@@ -5890,3 +5894,54 @@ impl fmt::Display for MemView {
...
+        return js.to_string();
+    }
+
+    let mut result = String::with_capacity(js.len());
+
+    // Only the @ts-self-types directive should precede imports
+    let has_directive = rest
+        .first()
+        .is_some_and(|l| l.trim().starts_with("/* @ts-self-types"));
+    let directive_count = usize::from(has_directive);
```
hoodmane: Is this 0 or 1? `count` is a funny way to call a variable that's either 0 or 1.

crates/cli-support/src/js/mod.rs
```diff
@@ -5890,3 +5894,54 @@ impl fmt::Display for MemView {
         write!(f, "{}{}", self.name, self.num)
     }
 }
+
+/// Hoists all ES module `import` declarations to the top of the generated JS,
+/// placed right after the `@ts-self-types` directive if present.
+fn hoist_es_imports(js: &str) -> String {
```
hoodmane: This seems a bit brittle.

hoodmane:
Code seems very brittle. Shouldn't we keep a separate list of imports we need and emit them at the end?

crates/cli-support/src/js/mod.rs
guybedford: Agreed, I've refactored to instead use a proper collection model.

hoodmane (approved):

PR #5106, opened 2026-04-11
fix: use qualified TS names for namespaced exports

ZhaoXiangXML:
### Description

This is a follow-up to #4977.

Fix TypeScript declaration naming for namespaced exports so generated `.d.ts` files use stable namespace-qualified identifiers derived from the existing qualified-name logic.

This is a breaking change for consumers of generated JS/TS output that referenced the previous bare or order-dependent declaration names.

### Problem

When exports use `js_namespace`, wasm-bindgen already computes qualified names for symbol disambiguation, but TypeScript declaration generation did not apply that naming consistently.

As a result:

- namespaced structs and enums in signatures could collapse to the same bare declaration names, for example both `foo.Point` and `bar.Point` appearing as `Point`
- namespaced free functions could use order-dependent aliases like `greet2`, making declaration names unstable as additional exports were added
- array and collection paths could use different naming paths from single-value references, for example `bar_point: Point` but `duplicate_bar_points(points: bar__Point[]): bar__Point[]`
- same-name exports across namespaces were therefore disambiguated inconsistently across `.d.ts` output

In other words, the problem was inconsistent declaration naming across code paths, not a consistent qualified naming scheme.

### Breaking Change

This PR changes the generated TypeScript declaration API for namespaced exports.

JavaScript namespace access remains the same, for example `foo.Point`, `bar.Point`, `foo.Status`, and `bar.Status` are still exposed through the same namespace objects. The breaking change is in the generated `.d.ts` declarations: namespaced exports now use stable namespace-qualified identifiers consistently, such as `foo__Point`, `bar__Point`, `foo__Status`, `bar__Status`, and `foo__greet`.

This also changes how collisions with real top-level exports are represented. If a top-level export and a namespaced export previously shared the same bare declaration name, the new output now disambiguates them instead of collapsing onto the same identifier.

As a result, TypeScript consumers may need to update code that depended on the previous generated declaration names, especially explicit type annotations, signatures inferred from generated declarations, and tests or tooling that depend on exact `.d.ts` output.

### Changes

- add a TS-specific qualified-name map for exported structs and enums
- use namespace-qualified declaration identifiers for namespaced exported structs, enums, and free functions
- use TS declaration names, not JS export names, when rendering TypeScript signatures
- resolve `AdapterType::Enum` through the TS name map
- make namespaced free-function declarations use qualified identifiers instead of order-based names
- keep private-class `export type { ... }` declarations aligned with the actual TS declaration identifier
- keep Emscripten TypeScript export entries aligned with the actual TS declaration identifier
- encode namespace-qualified enum names in `WasmDescribe`, matching the qualified-name handling already used for structs
- extend the CLI reference test coverage for same-name namespaced exports

### Tests

Expanded `crates/cli/tests/reference/js-namespace-export-same-name.rs` to cover:

- same-name namespaced structs and enums used in constructor args, methods, getters, and setters
- arrays of same-name namespaced exported classes
- same-name namespaced free functions
- cross-namespace references:
  - `foo` exports referencing `bar.Point` / `bar.Status`
  - `bar` exports referencing `foo.Point` / `foo.Status`
- same-name exports in nested namespaces
- same-name exports reused across different namespace depths
- same-name collisions across different namespace depths for different exported kinds

Added direct CLI regression tests in `crates/cli/tests/wasm-bindgen/main.rs` for:

- private namespaced classes exporting the actual declaration identifier in `.d.ts`
- Emscripten TypeScript output referencing the actual declaration identifier for namespaced classes

### Example

Before:

```ts
constructor(foo_point: Point, bar_point: Point, foo_status: Status, bar_status: Status);
duplicate_bar_points(points: bar__Point[]): bar__Point[];
declare function greet2(): string;
```

After:

```ts
constructor(foo_point: foo__Point, bar_point: bar__Point, foo_status: foo__Status, bar_status: bar__Status);
duplicate_bar_points(points: bar__Point[]): bar__Point[];
declare function foo__greet(): string;
```

### Checklist


guybedford (dismissed):
Well-structured fix for a real bug — the previous `.d.ts` output was simply wrong when namespaced exports shared the same `js_name`. Types that didn't match the generated JS code aren't a breaking change, they're a bug fix.

## Architecture

The `qualified_to_ts_name` map alongside the existing `qualified_to_js_name` is a clean separation — JS names stay unchanged for runtime behavior, TS declarations get deterministic qualified identifiers. Pre-populating in `generate()` for both structs and enums before adapter processing is the right call.

## Positive observations

- **Well-scoped**: Only TS declaration naming changes; runtime JS namespace access (`foo.Point`, `bar.Point`) is untouched.
- **Consistent naming**: `__` separator matches the existing `qualified_name` function in `shared`, so naming is uniform across wasm symbols and TS declarations.
- **Good test coverage**: Reference test expanded significantly — cross-namespace references, nested namespaces, same-name collisions at different depths, top-level collisions, arrays of namespaced types. Two new CLI integration tests for private class and Emscripten scenarios.
- **`AdapterType::Enum` fix** in `binding.rs`: Enum types were not resolved through the name map unlike `Struct` — a clear bug regardless of this PR.
- **`generate_identifier_with` refactor**: Clean extraction into a static-like method without behavior change.

## Minor items

1. **Enum identifier fallback** (`mod.rs:5198-5202`): In `generate_enum`, the fallback `self.generate_identifier(&enum_.name)` would generate from `name` instead of `qualified_name`, producing an inconsistent identifier if the `qualified_to_ts_name` lookup misses. This should never happen since the map is pre-populated in `generate()`, but consider making it a panic/unreachable or at least a debug warning rather than a silent fallback.

2. **Cosmetic**: A blank line between `identifier` and `comments` in `ExportDefinition` was removed — unrelated to the functional change.

3. **`generate_identifier_with`**: Takes `&mut HashMap` directly rather than `&self`. It's essentially a free function. Could be a standalone `fn` but not a big deal either way.

4. **Changelog entry**: The checklist item for changelog is unchecked.

crates/cli-support/src/js/mod.rs
```diff
@@ -5173,8 +5195,11 @@ addToLibrary({
     }
 
     fn generate_enum(&mut self, enum_: &AuxEnum) -> Result<(), Error> {
-        let identifier = self.generate_identifier(&enum_.name);
-
+        let identifier = self
+            .qualified_to_ts_name
+            .get(&enum_.qualified_name)
+            .cloned()
+            .unwrap_or_else(|| self.generate_identifier(&enum_.name));
```
guybedford: The fallback `self.generate_identifier(&enum_.name)` would generate from `name` (bare) instead of `qualified_name`, producing an inconsistent identifier if the lookup misses. Since `generate()` pre-populates `qualified_to_ts_name` for all enums, this path should be unreachable. A silent fallback here masks potential bugs — consider panicking or at minimum logging a warning:

```rust
let identifier = self
    .qualified_to_ts_name
    .get(&enum_.qualified_name)
    .cloned()
    .unwrap_or_else(|| {
        log::warn!("enum {} missing from qualified_to_ts_name", enum_.qualified_name);
        self.generate_identifier(&enum_.qualified_name)
    });
```

Note if you do keep a fallback, it should use `qualified_name` not `name` for consistency.

crates/cli-support/src/js/mod.rs
```diff
@@ -146,7 +148,6 @@ struct ExportDefinition {
     /// The identifier for the declaration, if distinct from the export name
     /// This allows invalid identifier export names (like "default").
     identifier: String,
-
     comments: Option<String>,
```
guybedford, line 149: Unrelated cosmetic change — the blank line between `identifier` and `comments` was removed. Prefer keeping the diff focused on functional changes.

crates/cli-support/src/js/mod.rs
```diff
@@ -5475,25 +5500,26 @@ addToLibrary({
     }
 
     fn generate_identifier(&mut self, name: &str) -> String {
+        Self::generate_identifier_with(&mut self.defined_identifiers, name)
+    }
+
+    fn generate_identifier_with(identifiers: &mut HashMap<String, usize>, name: &str) -> String {
```
guybedford, line 5502: This is effectively a free function — it doesn't reference `Self` in any way. Consider making it a standalone `fn` outside the `impl` block, or at minimum document that it's intentionally stateless. Minor nit.

guybedford (dismissed):
Good bug fix — the previous `.d.ts` output was wrong when namespaced exports shared the same `js_name`. The `qualified_to_ts_name` map alongside the existing `qualified_to_js_name` is a clean separation. Pre-populating in `generate()` for both structs and enums before adapter processing is the right approach.

The `AdapterType::Enum` fix in `binding.rs` is a standalone bug fix worth calling out — enum types were not resolved through the name map unlike `Struct`.

Test coverage is thorough: cross-namespace references, nested namespaces, same-name collisions at different depths, top-level collisions, arrays of namespaced types, private class and Emscripten scenarios.

A few items to address below.

guybedford (approved):

guybedford (approved):

PR #5112, opened 2026-04-15
Use js_sys::futures instead of wasm_bindgen_futures in codegen

### LINKED ISSUE #5104: use of unresolved module or unlinked crate `wasm_bindgen_futures`
Following #5049 I tried to remove the dependency on `wasm-bindgen-futures` and add the new optional `"futures"` feature on the `js-sys` crate. However, the `#[wasm_bindgen]` macro is now producing an error when building:

```toml
[package]
name = "microgpt-rs"
version = "0.1.0"
edition = "2024"

[dependencies]
rand = { version = "0.10.0" }
rand_distr = "0.6.0"
getrandom = { version = "0.4.2", features = ["wasm_js"] }
wasm-bindgen = { version = "0.2.117" }
# wasm-bindgen-futures = "0.4.67"
web-sys = { version = "0.3.94", features = ["console", 'Response', 'Window'] }
js-sys = { version = "0.3.94", features = ["futures"] }

[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
reqwest = { version = "0.13.2", features = ["blocking"] }

[lib]
crate-type = ["cdylib", "rlib"]
```

```rust
...
#[wasm_bindgen]
pub async fn main_() -> Result<(), JsValue> {
    let mut rng = rand::rng();
    // Let there be a Dataset `docs`: list[str] of documents (e.g. a list of names)
    let window: web_sys::Window = web_sys::window().unwrap();
    // let resp_value: JsValue = JsFuture::from(window.fetch_with_str("https://raw.githubusercontent.com/karpathy/makemore/988aa59/names.txt")).await?;
    // let names = JsFuture::from(resp.text()?).await?.as_string().unwrap();
    let resp_value = window
        .fetch_with_str("https://raw.githubusercontent.com/karpathy/makemore/988aa59/names.txt")
        .await?;
    let resp: Response = resp_value.dyn_into().unwrap();
    let names: String = resp.text()?.await?.as_string().unwrap();
    ...
}
```

```rust
➜ microgpt git:(master) ✗ wasm-pack build --target web --no-typescript --no-pack
...
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `wasm_bindgen_futures`                
   --> src/lib.rs:287:1
    |
287 | #[wasm_bindgen]
    | ^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `wasm_bindgen_futures`
    |
    = help: if you wanted to use a crate named `wasm_bindgen_futures`, use `cargo add wasm_bindgen_futures` to add it to your `Cargo.toml`
...
```

guybedford:
Resolves #5104.

This adds a new `js-sys` feature to wasm-bindgen, which when set, the wasm bindgen macro will emit js_sys::futures bindings instead of wasm_bindgen_futures bindings.

This is necessary, because the macro assumes wasm_bindgen_futures but not necessarily js_sys currently, so would otherwise be a breaking change if we made the hard switch to expecting js_sys for the macro.

To use the new codegen path:

```toml
[dependencies]
js-sys = "..."
wasm-bindgen = { version = "...", features = ["js-sys"] }
```

Then normal async bindings generate through `js_sys::futures`:

```rust
#[wasm_bindgen]
pub async fn exported() {
    // generated glue uses js_sys::futures
}

#[wasm_bindgen]
extern "C" {
    async fn imported();
}
```

Test coverage includes targeted macro expansion/checks for async import/export codegen using `::js_sys::futures`, plus existing compatibility coverage for the `wasm-bindgen-futures` re-export path.

### DIFF
--- a/crates/macro-support/src/ast.rs
+++ b/crates/macro-support/src/ast.rs
@@ -8,6 +8,10 @@ use std::hash::{Hash, Hasher};
 use syn::Path;
 use wasm_bindgen_shared as shared;
 
+pub fn use_js_sys_futures() -> bool {
+    cfg!(feature = "js-sys")
+}
+
 /// Whether a function is a start function, and if so, whether it
 /// should be exported to JS.
 #[cfg_attr(feature = "extra-traits", derive(Debug))]
@@ -115,6 +119,8 @@ pub struct Export {
     pub wasm_bindgen: Path,
     /// Path to wasm_bindgen_futures
     pub wasm_bindgen_futures: Path,
+    /// Path to js_sys
+    pub js_sys: Path,
 }
 
 /// The 3 types variations of `self`.
@@ -212,6 +218,8 @@ pub struct ImportFunction {
     pub wasm_bindgen: Path,
     /// Path to wasm_bindgen_futures
     pub wasm_bindgen_futures: Path,
+    /// Path to js_sys
+    pub js_sys: Path,
     /// Generic parameters as validated simple type parameters for this function
     pub generics: syn::Generics,
 }
--- a/guide/src/reference/js-promises-and-rust-futures.md
+++ b/guide/src/reference/js-promises-and-rust-futures.md
@@ -22,8 +22,6 @@ async fn get_from_js() -> Result<JsValue, JsValue> {
 }
 ```
 
-This works out of the box — the futures support is built into `js-sys` and
-is always available.
 
 For typed promises the awaited value matches the type parameter directly:
 
--- a/guide/src/reference/js-promises-and-rust-futures.md
+++ b/guide/src/reference/js-promises-and-rust-futures.md
@@ -22,14 +22,8 @@ async fn get_from_js() -> Result<JsValue, JsValue> {
 }
 ```
 
-This works out of the box when `wasm-bindgen-futures` is in your dependency
-tree, which activates the `futures` feature on `js-sys`. You can also enable
-it explicitly without `wasm-bindgen-futures`:
-
-```toml
-[dependencies]
-js-sys = { version = "0.3", features = ["futures"] }
-```
+This works out of the box — the futures support is built into `js-sys` and
+is always available.
 
 For typed promises the awaited value matches the type parameter directly:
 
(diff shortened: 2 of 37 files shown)


crates/macro-support/src/ast.rs
```diff
@@ -45,8 +45,8 @@ pub struct Program {
     pub inline_js: Vec<String>,
     /// Path to wasm_bindgen
     pub wasm_bindgen: Path,
-    /// Path to js_sys
-    pub js_sys: Path,
+    /// When set, use `js_sys::futures` instead of `wasm_bindgen_futures` for async codegen.
```
hoodmane: Are you thinking about switching the default in the future? From the comment in `lib.rs`, it sounds like `wasm_bindgen_futures` is now a reexport of `js_sys::futures` in which case it'd make sense just to always switch. Is that not true?

crates/macro-support/src/ast.rs
```diff
@@ -115,6 +115,8 @@ pub struct Export {
     pub wasm_bindgen: Path,
     /// Path to wasm_bindgen_futures
     pub wasm_bindgen_futures: Path,
+    /// When set, use `js_sys::futures` instead of `wasm_bindgen_futures` for async codegen.
+    pub js_sys: Option<Path>,
```
hoodmane: Shouldn't we unify these into:
```
   pub futures: Path
```
and then set it appropriately?

crates/macro-support/src/ast.rs
guybedford: Yes that would be the ideal, the concern is that this is a breaking change since some modules might only have wasm_bindgen_futures dependency without js_sys dependency in scope.

crates/macro-support/src/ast.rs
guybedford: The `js_sys` attribute is an existing attribute that allows setting a path to JS sys. The change here makes the value part of the key value attribute here optional. The convention then is that when js_sys attribute is provided OR set, we use it. So we do still need to have the js_sys path for the previous use case of customizing the js_sys location.

guide/src/reference/js-promises-and-rust-futures.md
```diff
@@ -22,14 +22,8 @@ async fn get_from_js() -> Result<JsValue, JsValue> {
...
-This works out of the box when `wasm-bindgen-futures` is in your dependency
-tree, which activates the `futures` feature on `js-sys`. You can also enable
-it explicitly without `wasm-bindgen-futures`:
-
-```toml
-[dependencies]
-js-sys = { version = "0.3", features = ["futures"] }
-```
+This works out of the box — the futures support is built into `js-sys` and
+is always available.
```
hoodmane: I think people will assume this if we don't say anything here.
```suggestion
```

guybedford: Thanks, added.

guide/src/reference/types/js-sys.md
```diff
@@ -175,16 +175,8 @@ async fn example() -> Result<Number, JsValue> {
...
-automatically when `wasm-bindgen-futures` is a dependency. Enable it directly
-if needed:
-
-```toml
-[dependencies]
-js-sys = { version = "0.3", features = ["futures"] }
-```
-
-If you need a named [`JsFuture`] value, use `JsFuture::from`:
+Futures support is enabled by default. If you need a named [`JsFuture`]
```
hoodmane:
```suggestion
If you need a named [`JsFuture`]
```

hoodmane (approved):

PR #5113, opened 2026-04-15
feat: add Promise combinator utilities to js_sys::futures

guybedford:
`futures_util::future::join_all` polls futures cooperatively within the WASM executor, which serializes JS promise resolution — the JS event loop never gets a chance to drive concurrent I/O between polls. This adds native Promise combinators to `js_sys::futures` that delegate to `Promise.all`, `Promise.race`, and `Promise.allSettled`, enabling true concurrent I/O for JS-backed operations.

Benchmarking 10 parallel 1MB streaming fetches in workers-rs shows a **2.3x speedup** (31ms → 14ms) when using `Promise.all` vs `futures::future::join_all`.

## What's added

`js_sys::futures::utils` module (`pub use`d from `js_sys::futures`) with:

**Traits:**
* `IntoPromise` — accepts both `Promise<T>` (identity) and `Future<Output = Result<T, JsValue>>` (via `future_to_promise_typed`), fully generic over the resolution type `T`
* `PromiseTuple` — maps tuples of promises `(Promise<T1>, ..., Promise<Tn>)` to their resolution shapes (`Joined` and `Settled` associated types), enabling fully inferred heterogeneous combinators

**Homogeneous functions** (take `IntoIterator<Item: IntoPromise>`):
* `join_all` — `Promise.all`, returns `Array<T>`
* `all_settled` — `Promise.allSettled`, returns `Array<PromiseState<T>>`
* `select` — `Promise.race`, returns first settled `T` (named to match the Rust `futures` convention)

**Heterogeneous macros** (accept `Promise<T>` or `Future` args of different types):
* `join!(...)` — `Promise.all`, returns `Promise<ArrayTuple<(T1, T2, ...)>>` with full type inference via `PromiseTuple::Joined`
* `all_settled!(...)` — `Promise.allSettled`, returns `Promise<ArrayTuple<(PromiseState<T1>, ...)>>` via `PromiseTuple::Settled`

Both macros accept a mix of `Promise<T>` and `Future<Output = Result<T, JsValue>>` in any position — `IntoPromise` lifts both uniformly. Result types are fully inferred with no caller annotations needed:

```rust
let (response, buffer) = join!(fetch_promise, array_buffer_promise).await?.into_parts();
```

Guide documentation updated with usage examples for all combinators.


guybedford:
Closing this out because we should encourage using Rust idioms not drop-in alternatives that have two ways of doing the same thing.

The benchmarks here did reveal a slight improvement when working with very large numbers of Promises in that we avoid the future conversion, but if this is a problem, users should reach for `Promise::all` directly instead of something like this.

PR #5116, opened 2026-04-18
Use Crates.io Trusted Publishing

daxpedda:
Use [Crates.io trust publishing](https://crates.io/docs/trusted-publishing) instead of using a secret key.

I've already added the required configuration to Crates.io. After merging we should enable the "Require trusted publishing for all new versions" checkbox for all Crates and delete the secret from GitHub.


RReverser (approved):
Love this!

guybedford (approved):

PR #5118, opened 2026-04-18
feat: expose `WebAssembly.Instance`

CPunisher:
### Description

Just like `wasm_bindgen::module`, `wasm_bindgen::memory`, `wasm_bindgen::exports`, this pr adds `wasm_bindgen::instance` to get current wasm instance.

### Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1678,6 +1678,7 @@ pub fn instance() -> JsValue {
     __wbindgen_instance()
 }
 
+// TODO: deprecate next major
 /// Returns a handle to this Wasm instance's `WebAssembly.Instance.prototype.exports`
 pub fn exports() -> JsValue {
     __wbindgen_exports()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1272,6 +1272,7 @@ extern "C" {
     fn __wbindgen_exports() -> JsValue;
     fn __wbindgen_memory() -> JsValue;
     fn __wbindgen_module() -> JsValue;
+    fn __wbindgen_instance() -> JsValue;
     fn __wbindgen_function_table() -> JsValue;
 
     fn __wbindgen_reinit();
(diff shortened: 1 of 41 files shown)


guybedford (approved):
This seems fine to me.

src/lib.rs
```diff
@@ -1669,6 +1670,14 @@ pub fn module() -> JsValue {
...
+/// Returns a handle to this Wasm instance's `WebAssembly.Instance`.
+/// This is only available when the final Wasm app is built with
+/// `--target no-modules`, `--target web`, `--target deno` or `--target nodejs`.
+/// It is unavailable for `--target bundler`.
+pub fn instance() -> JsValue {
+    __wbindgen_instance()
+}
+
 /// Returns a handle to this Wasm instance's `WebAssembly.Instance.prototype.exports`
 pub fn exports() -> JsValue {
```
guybedford, line 1683: Let's perhaps add a comment that this can be now deprecated in the next release:


```suggestion
// TODO: deprecate next major
/// Returns a handle to this Wasm instance's `WebAssembly.Instance.prototype.exports`
pub fn exports() -> JsValue {
```

src/lib.rs
```diff
@@ -1669,6 +1670,14 @@ pub fn module() -> JsValue {
     __wbindgen_module()
 }
 
+/// Returns a handle to this Wasm instance's `WebAssembly.Instance`.
+/// This is only available when the final Wasm app is built with
+/// `--target no-modules`, `--target web`, `--target deno` or `--target nodejs`.
+/// It is unavailable for `--target bundler`.
+pub fn instance() -> JsValue {
+    __wbindgen_instance()
+}
```
guybedford, line 1679: It could be nice to actually type this properly, effectively moving the js-sys type for WebAssembly.Instance into core, and then referencing that from js-sys instead.

guybedford: I guess we'd need to migrate the constructor as well to do that, which would then pull in all of js-sys, so this doesn't work.

One alternative to allow better typing could be to have a js_sys version of this function that is typed?

CPunisher: > have a js_sys version of this function that is typed

I agree with this approach. We can do that in another PR to add types for all intrinsic.

guybedford:
This can land now, but just needs a changelog entry to do so.

CPunisher:
Done

CPunisher:
> It could be nice to actually type this properly, effectively moving the js-sys type for WebAssembly.Instance into core, and then referencing that from js-sys instead.

@guybedford After thinking about this more, I'm considering not to add these APIs to `js-sys`.

1. `js-sys` describes itself as "Bindings for all JS global objects and functions in all JS environments". These wasm-bindgen runtime / intrinsics do not really fit that scope.

2. Even if we only expose typed wrappers like the following code, the wrapper is so small that it does not seem to add much practical convenience for users. They can already write the conversion directly when they need it.

```rust
pub fn module() -> crate::WebAssembly::Module {
    wasm_bindgen::module().unchecked_into()
}
```

So overall, I think it is better to keep things simple.

Issue #5042, opened 2026-03-17
Breaking change from generics


guybedford:
Just noticed this one from https://github.com/fzyzcjy/flutter_rust_bridge/issues/3009 -

```rs
Error[E0283]: type annotations needed
   --> /Users/nay/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/flutter_rust_bridge-2.11.1/src/third_party/wasm_bindgen/worker_pool.rs:140:30
    |
140 |         worker.post_message(&Array::from_iter([wasm_init_object]))?;
    |                              ^^^^^ cannot infer type for type parameter `T` declared on the struct `Array`
    |
    = note: cannot satisfy `_: JsGeneric`
    = note: required for `js_sys::Array<_>` to implement `FromIterator<Object>`
```

It could make sense to reconsider this specific method's generics approach. Will investigate further soon.

fzyzcjy:
That would be great and looking forward to it!

PR #5121, opened 2026-04-22
feat: Promise ergonomics for typed concurrent awaits

guybedford:
This is a follow-on to the work originally attempted in #5113 to provide better ergonomics for working with multiple combinators on `Promise<T>` generics.

Previously, that PR attempted to provide `js_sys::futures::utils` with the standard Rust futures utils variants in a way that could work with typed Promises.

Instead, this PR leans into the existing `Promise::` functions to ensure they fully handle the ergonomic edge cases required for the type inference through the standard combinators.

In particular:

- To construct an `Array<Promise<T>>` to use `Promise::all_iterable` required multiple steps. Instead, by resolving the `FromIterator` trait issues from before for the generic `Array<T>`, we can now support `rust_iter.collect::<Array<_>>` as providing an ergonomic route without an intermediate `Vec` having to be constructed. The same applies to `all_settled_iterable` and `race_iterable`.

- For the heterogeneous case previously covered by the `join!` macro, this PR adds `Promise::all_tuple((p1, p2, ...))` as an inherent method. It takes a Rust tuple of `Promise<T_i>` with different types and returns a typed `Promise<ArrayTuple<(T_1, ..., T_n)>>` that destructures via `.into_tuple()` into a native Rust tuple. Implemented for arity 1 through 8. In the process conversion from `ArrayTuple -> ()` has been renamed to be `into_tuple()` instead of the former `into_parts()` which is now deprecated from the original generics version.

- For the heterogeneous case previously covered by the `all_settled!` macro, `Promise::all_settled_tuple` follows the same shape and returns `Promise<ArrayTuple<(PromiseState<T_1>, ..., PromiseState<T_n>)>>`. Same arity range.

- For mixing Rust futures with JS promises in these combinators, users should just lift futures with `future_to_promise_typed(fut)` at the call site. This replaces the implicit `IntoPromise` trait from the previous PR and keeps the spawning visible in user code. This way we don't need to worry about mixing or supporting a shared trait.

- There is no `race_tuple` because `Promise.race` returns a value of type `T_1 | T_2 | ... | T_n`, which Rust can't express. For heterogeneous race, users should rather collect into an `Array<JsValue>` and use `Promise::race_iterable`, narrowing with `dyn_into` at the inspection site.

There are also some minor trait cleanups where this work touches them.

The async guide is updated based on the above guidance now.

### FromIterator

To support `FromIterator` for `Array<T>` without creating a breaking change (as previously encountered and fixed in https://github.com/wasm-bindgen/wasm-bindgen/pull/5052), this problem is now resolved with a new `wasm_bindgen::IntoJsGeneric` trait in the core crate. It has an associated type `JsCanon: JsGeneric` that pins the target element type from the input alone, which avoids the `AsRef<T>` inference ambiguity that was the root cause of the #5042 regression. The codegen emits an identity impl for every `#[wasm_bindgen]` type. The impl takes `self` by value and reinterprets the transparent JS handle wrapper into its canonical type, so it applies uniformly to types that aren't Rust-level `Clone`, including generic types whose parameters aren't `Clone`. Types whose wrapper enforces owned-once destruction semantics opt out with `#[wasm_bindgen(no_into_js_generic)]`, which is currently only `JsClosure`. A reference blanket covers `&T` for cloneable `T`, so borrowed iteration works for cloneable JS wrapper types.

That we now support `iter.collect::<Array>()` with typed items is formally breaking - but only against code that is already using generics. That is, it does not affect historical usage. Following up and fixing the former issue now that we have the technique to do that makes sense, and we are within the two months window of generics shipping still to do this therefore I think the benefit of supporting `FromIterator` outweighs the cost here. And if there are adoption issues we can always revert if necessary.

### Promise Tuple Trait

To support the heterogenous `Promise::all_tuple` and `Promise::all_settled_tuple`, these new tuple methods are implemented on a wrapper using an internal doc-hidden `PromiseTuple` trait with one impl per arity. Each impl builds an `ArrayTuple<(Promise<T_1>, ..., Promise<T_n>)>` from the input tuple using the existing `From<(T_1, ..., T_n)>` conversion, hands it to `Promise::all_iterable` or `Promise::all_settled_iterable`, and reinterprets the result. The single `unchecked_into` that reinterprets the returned `Array<JsValue>` as the typed `ArrayTuple<(T_1, ..., T_n)>` is encapsulated inside the trait, so no unchecked casts leak to the call site. Soundness rests on `Promise.all` and `Promise.allSettled` preserving input order and arity, which they do by spec.

### DIFF
--- a/crates/js-sys/src/futures/mod.rs
+++ b/crates/js-sys/src/futures/mod.rs
@@ -46,6 +46,7 @@ use wasm_bindgen::__rt::marker::ErasableGeneric;
 #[cfg(all(target_arch = "wasm32", feature = "std", panic = "unwind"))]
 use wasm_bindgen::__rt::panic_to_panic_error;
 use wasm_bindgen::convert::{FromWasmAbi, Upcast};
+use wasm_bindgen::sys::Promising;
 use wasm_bindgen::{prelude::*, JsError, JsGeneric};
 
 #[cfg_attr(docsrs, doc(cfg(feature = "futures-core-03-stream")))]
@@ -126,7 +127,7 @@ impl<T> fmt::Debug for JsFuture<T> {
     }
 }
 
-impl<T: 'static + FromWasmAbi> From<Promise<T>> for JsFuture<T> {
+impl<T: JsGeneric + FromWasmAbi> From<Promise<T>> for JsFuture<T> {
     fn from(js: Promise<T>) -> JsFuture<T> {
         // Use the `then` method to schedule two callbacks, one for the
         // resolved value and one for the rejected value. We're currently
@@ -216,7 +217,7 @@ impl<T> Future for JsFuture<T> {
     }
 }
 
-impl<T: 'static + FromWasmAbi> IntoFuture for Promise<T> {
+impl<T: JsGeneric + FromWasmAbi> IntoFuture for Promise<T> {
     type Output = Result<T, JsValue>;
     type IntoFuture = JsFuture<T>;
 
@@ -254,10 +255,10 @@ where
         spawn_local(async move {
             match future.await {
                 Ok(val) => {
-                    resolve.call(&JsValue::undefined(), (&val,)).unwrap_throw();
+                    resolve.call(&JsValue::UNDEFINED, (&val,)).unwrap_throw();
                 }
                 Err(val) => {
-                    reject.call(&JsValue::undefined(), (&val,)).unwrap_throw();
+                    reject.call(&JsValue::UNDEFINED, (&val,)).unwrap_throw();
                 }
             }
         });
@@ -295,14 +296,14 @@ where
             let res = future.catch_unwind().await;
             match res {
                 Ok(Ok(val)) => {
-                    resolve.call(&JsValue::undefined(), (&val,)).unwrap_throw();
+                    resolve.call(&JsValue::UNDEFINED, (&val,)).unwrap_throw();
                 }
                 Ok(Err(val)) => {
-                    reject.call(&JsValue::undefined(), (&val,)).unwrap_throw();
+                    reject.call(&JsValue::UNDEFINED, (&val,)).unwrap_throw();
                 }
                 Err(val) => {
                     reject
-                        .call(&JsValue::undefined(), (&panic_to_panic_error(val),))
+                        .call(&JsValue::UNDEFINED, (&panic_to_panic_error(val),))
                         .unwrap_throw();
                 }
             }
@@ -331,10 +332,11 @@ where
 /// If the `future` provided panics then the returned `Promise` **will not
 /// resolve**. Instead it will be a leaked promise. This is an unfortunate
 /// limitation of Wasm currently that's hoped to be fixed one day!
-pub fn future_to_promise_typed<F, T>(future: F) -> Promise<T>
+pub fn future_to_promise_typed<T, F>(future: F) -> Promise<<T as Promising>::Resolution>
 where
     F: Future<Output = Result<T, JsValue>> + 'static,
-    T: FromWasmAbi + JsGeneric + Upcast<T> + 'static,
+    T: Promising + FromWasmAbi + JsGeneric,
+    <T as Promising>::Resolution: JsGeneric,
 {
     let mut future = Some(future);
 
@@ -343,10 +345,10 @@ where
         spawn_local(async move {
             match future.await {
                 Ok(val) => {
-                    resolve.call(&JsValue::undefined(), (&val,)).unwrap_throw();
+                    resolve.call(&JsValue::UNDEFINED, (&val,)).unwrap_throw();
                 }
                 Err(val) => {
-                    reject.call(&JsValue::undefined(), (&val,)).unwrap_throw();
+                    reject.call(&JsValue::UNDEFINED, (&val,)).unwrap_throw();
                 }
             }
         });
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -56,7 +56,7 @@ use wasm_bindgen::JsError;
 
 // Re-export sys types as js-sys types
 pub use wasm_bindgen::sys::{JsOption, Null, Promising, Undefined};
-pub use wasm_bindgen::JsGeneric;
+pub use wasm_bindgen::{IntoJsGeneric, JsGeneric};
 
 // When adding new imports:
 //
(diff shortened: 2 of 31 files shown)


crates/js-sys/src/futures/mod.rs
```diff
@@ -331,10 +332,11 @@ where
 /// If the `future` provided panics then the returned `Promise` **will not
 /// resolve**. Instead it will be a leaked promise. This is an unfortunate
 /// limitation of Wasm currently that's hoped to be fixed one day!
-pub fn future_to_promise_typed<F, T>(future: F) -> Promise<T>
+pub fn future_to_promise_typed<T, F>(future: F) -> Promise<<T as Promising>::Resolution>
```
hoodmane, line 335: Is swapping the type argument order backwards compatible?

crates/js-sys/src/futures/mod.rs
```diff
@@ -126,7 +127,7 @@ impl<T> fmt::Debug for JsFuture<T> {
     }
 }
 
-impl<T: 'static + FromWasmAbi> From<Promise<T>> for JsFuture<T> {
+impl<T: JsGeneric + FromWasmAbi> From<Promise<T>> for JsFuture<T> {
```
hoodmane, line 130: Why remove `'static`? Why add `JsGeneric`? I would appreciate if you could explain what is happening with this bound.

crates/js-sys/src/lib.rs
```diff
@@ -1944,73 +1950,44 @@ impl<T: JsGeneric> core::iter::IntoIterator for Array<T> {
...
-{
-    fn from_iter<I>(iter: I) -> Array
+// `FromIterator` / `Extend` for `Array<T>` project every input value through
+// its canonical [`JsGeneric`] via [`IntoJsGeneric`].
+//
+// Using an associated type (`A::JsCanon`) rather than a free type parameter
+// bound by `AsRef<T>` is what lets inference pick `T` uniquely from the
+// iterator item — no turbofish, no intermediate `Vec`, no `AsRef<T>` ambiguity
+// across the auto-generated `AsRef<JsValue>` / `AsRef<Self>` /
+// `AsRef<Superclass>` impls on `#[wasm_bindgen]` types. This fixes the
```
hoodmane:
```suggestion
// iterator item. This fixes the
```

crates/js-sys/src/lib.rs
```diff
@@ -12810,6 +12789,19 @@ impl<T> PromiseState<T> {
     }
 }
 
+/// Converts a `PromiseState<T>` into a `Result<T, JsValue>`, matching the
+/// spec invariant that exactly one of the fulfilled value or the rejection
```
hoodmane, line 12791: >  exactly one

Can't a `PromiseState` be pending?

src/convert/impls.rs
```diff
@@ -788,6 +788,22 @@ impl IntoWasmAbi for JsError {
     }
 }
 
+// `JsError` is `#[repr(transparent)]` over `JsValue`, so this is the same
+// wire format as `JsValue` — we just rewrap the resulting `JsValue` into
+// `JsError` on the way out. Required so `JsError` can be used in positions
+// that cross the ABI (e.g. `Promising`'s supertrait bound, promise
+// resolutions, async return types).
```
hoodmane:
```suggestion
// `JsError` is `#[repr(transparent)]` over `JsValue`
```

src/convert/slices.rs
```diff
@@ -479,7 +479,10 @@ impl<T: ErasableGeneric<Repr = JsValue> + WasmDescribe> VectorIntoWasmAbi for T
 // JsValue-like slice support (Rust-to-JS only)
 // JsValue-like are repr(transparent) over u32, so &[JsValue] is a contiguous array of heap indices
 
-unsafe impl<T: ErasableGeneric> ErasableGeneric for &[T] {
+unsafe impl<T: ErasableGeneric> ErasableGeneric for &[T]
+where
+    <T as ErasableGeneric>::Repr: Sized,
```
hoodmane: What does this change mean? Is it backwards compatible?

src/convert/traits.rs
```diff
@@ -543,3 +543,77 @@ impl<T: ErasableGeneric<Repr = JsValue> + UpcastFrom<T> + Upcast<JsValue> + JsCa
     JsGeneric for T
 {
 }
+
+/// Value conversion from a type into its canonical [`JsGeneric`] form.
+///
+/// This trait lets APIs that operate over a generic JavaScript element type
+/// (for example [`js_sys::Array<T>`], [`js_sys::Set<T>`], combinators like
+/// `Promise.all`, etc.) project every input value through a single,
+/// functionally-determined target [`JsGeneric`].
```
hoodmane: What does "functionally-determined" mean?

src/convert/traits.rs
```diff
@@ -543,3 +543,77 @@ impl<T: ErasableGeneric<Repr = JsValue> + UpcastFrom<T> + Upcast<JsValue> + JsCa
...
+/// types: each implementor explicitly opts in, which leaves room for future
+/// wrapper types to pick a non-identity [`Self::JsCanon`].
+///
+/// # Example
+///
+/// ```ignore
+/// use js_sys::{Array, Number};
+///
+/// // No intermediate `Vec`, no annotations: the iterator's item type pins
+/// // `Array<A::JsCanon>` uniquely.
```
hoodmane:
```suggestion
```

src/convert/traits.rs
```diff
@@ -543,3 +543,77 @@ impl<T: ErasableGeneric<Repr = JsValue> + UpcastFrom<T> + Upcast<JsValue> + JsCa
...
+
+    /// Produce the canonical [`JsGeneric`] value for `self`.
+    ///
+    /// Takes `&self` because crossing into a JS-side handle is already a
+    /// refcount-level operation (every [`JsGeneric`] is an `Rc`-backed
+    /// externref handle). The "clone" that happens inside each impl body is
+    /// the boundary-crossing refcount, not an extra copy — this is the
+    /// `to_js` (borrow → owned) shape rather than `into_js` (consume → owned),
+    /// matching Rust's naming conventions for value conversion from a
+    /// reference.
```
hoodmane:
```suggestion
```

src/convert/traits.rs
```diff
@@ -543,3 +543,77 @@ impl<T: ErasableGeneric<Repr = JsValue> + UpcastFrom<T> + Upcast<JsValue> + JsCa
...
+// requires.
+impl<T: IntoJsGeneric + ?Sized> IntoJsGeneric for &T {
+    type JsCanon = T::JsCanon;
+    #[inline]
+    fn to_js(&self) -> T::JsCanon {
+        (**self).to_js()
+    }
+}
+
+// Intentionally not a blanket `impl<T: JsGeneric> IntoJsGeneric for T`:
```
hoodmane, line 604: Why is it okay to implement it for `&T` but not for `T`?

hoodmane (approved):
Generally looks very nice.

crates/js-sys/src/futures/mod.rs
guybedford: No it isn't, but generics itself can still be considered under "experimental" given it has only been shipped for 2 months now. Of course we shouldn't do many more of these types of breaks, but I think ergonomically it makes sense to add these kinds of improvements now while we still can.

crates/js-sys/src/futures/mod.rs
guybedford: `JsGeneric` includes `'static`.

crates/js-sys/src/lib.rs
guybedford: No, in JS `PromiseState` is always fulfilled or rejected!

src/convert/slices.rs
guybedford: This was vestigeal from a previous edit where I tried adding `?Sized` to ErasableGeenric for various reasons, but then backed away from that - removed.

src/convert/traits.rs
guybedford: Yeah that's weird, I've fixed up the description.

src/convert/traits.rs
guybedford: We implement it individually and concretely for various `T`, then blanket implement the `&T` implementations off of the `T` implementations. So the blanket is the convenience.

PR #5122, opened 2026-04-25
Simplify web-sys generated bindings

guybedford:
This simplifies the generated `web-sys` bindings by omitting redundant `#[wasm_bindgen]` attributes when they match wasm-bindgen defaults, and improves whitespace formatting of the generated attributes by feeding them through `rustfmt` cleanly.

`rustfmt` only formats unknown macro attribute args of the form `name = <literal>` — bare idents and paths defeat it and the macro tokens are emitted verbatim with stray spaces (e.g. `# [wasm_bindgen (method , js_class = "X" , js_name = abort)]`). Switching the generated `js_name`, `extends`, `static_method_of`, and `vendor_prefix` arguments to string-literal form lets `rustfmt` format every line uniformly.

Implemented:

* remove `structural` / `final` emission from generated WebIDL attributes and methods (structural is already the default)
* remove the dead `is_structural` and `has_ident_attribute` helper paths
* skip redundant `js_name` emission where the generated Rust name already matches
* extend the `#[wasm_bindgen]` attribute parser to accept string-literal forms for the `extends`, `static_method_of`, and `vendor_prefix` arguments alongside the existing bare-path/ident forms
* emit the generated `js_name`, `extends`, `static_method_of`, and `vendor_prefix` arguments as string literals from the WebIDL generator
* regenerate `web-sys` bindings (now overwhelmingly formatted as `#[wasm_bindgen(method, js_class = "X", js_name = "y")]`) and add an Unreleased changelog entry

Before:

```rs
# [wasm_bindgen (method , structural , js_class = "AbortController" , js_name = abort)]
pub fn abort(this: &AbortController);
```

After:

```rs
#[wasm_bindgen(method, js_class = "AbortController", js_name = "abort")]
pub fn abort(this: &AbortController);
```

The string-literal form is purely additive on the macro side; existing user code using `extends = ::js_sys::Object` etc. continues to work unchanged. The remaining unformatted lines in `web-sys` are the small set of type-definition headers that use `is_type_of = |_| false` (a closure expression, not a `name = <literal>` arm) — that's a follow-up.

Test coverage:

* `cargo test -p wasm-bindgen-webidl`
* `cargo check -p web-sys`


crates/web-sys/src/features/gen_AbortController.rs
```diff
@@ -27,14 +27,14 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API requires the following crate features to be activated: `AbortController`*"]
     pub fn new() -> Result<AbortController, JsValue>;
-    # [wasm_bindgen (method , structural , js_class = "AbortController" , js_name = abort)]
+    #[wasm_bindgen(method, js_class = "AbortController")]
```
hoodmane, line 34: Nit: What's the deal with the white space? Here it looks nice now with `method,` but in most places it's still `method ,`.

CHANGELOG.md
```diff
@@ -23,6 +23,12 @@
   instantiated `WebAssembly.Instance`.
   [#5118](https://github.com/wasm-bindgen/wasm-bindgen/pull/5118)
 
+### Changed
+
+* Simplified generated `web-sys` bindings by omitting redundant
+  `#[wasm_bindgen]` attributes when they match wasm-bindgen defaults, including
+  structural method annotations and matching `js_name` entries.
```
hoodmane: Does this need a changelog? Presumably it's transparent to the user?

guybedford: Yeah, this is entirely non-semantic to end users apart from being visible in the code.

hoodmane (approved):

crates/web-sys/src/features/gen_AbortController.rs
guybedford: It's really weird - rustfmt basically can't handle non-quoted string paths, so bails on those and we get the ugly output.

If we quote all names, we can get proper formatting of the attribute. Then we're just left with the `extends = ::js_sys::Object` case that isn't formatible.

Pushed a commit to do that so we get nicer output rather.

PR #5128, opened 2026-04-27
Enforce unwind safety on #[wasm_bindgen] exports under panic=unwind

guybedford:
This implements compile-time unwind-safety validation for `#[wasm_bindgen]`-exported methods and free functions when building with `panic = "unwind"`.

The macro previously wrapped every exported sync body in `maybe_catch_unwind(|| { ... })` but emitted all argument/receiver `from_abi` conversions *inside* the closure. The closure therefore only captured the raw ABI integer parameters — trivially `UnwindSafe` — so the `UnwindSafe` bound on `maybe_catch_unwind` was effectively a no-op for the user's types. `&self` / `&mut self` methods on structs containing `Cell`, `RefCell`, `Mutex`, etc. compiled cleanly, even though a mid-method panic could leave those types in a torn state observable on the next call from JS.

Empirically, the following all compiled fine under `RUSTFLAGS=-Cpanic=unwind` before this change:

```rs
#[wasm_bindgen]
pub struct Holder { counter: Cell<u32>, inner: RefCell<Vec<u32>> }

#[wasm_bindgen]
impl Holder {
    pub fn bump_then_panic(&mut self) {
        self.counter.set(self.counter.get() + 1);
        panic!();                 // counter half-updated, observable next call
    }
    pub fn shared_borrow_panic(&self) {
        let _g = self.inner.borrow_mut();
        panic!();                 // RefCell borrow leaked permanently
    }
}
```

This adds explicit type-level bounds in the export codegen that fire under `panic = "unwind"`:

* `&self` / `&mut self` &rarr; `Self: RefUnwindSafe`
* `&T` / `&mut T` args  &rarr; `T: RefUnwindSafe`
* owned arg / `self`     &rarr; `T: UnwindSafe`

backed by two no-op helpers `__rt::ensure_ref_unwind_safe` / `__rt::ensure_unwind_safe` that only carry the trait bound under `cfg(all(target_family = "wasm", feature = "std", panic = "unwind"))`. For `&self` / `&T` and owned arguments these match exactly what stdlib's closure-capture bound would require naturally — they just produce a clearer error message pointing at the bound rather than at the capture failure. For `&mut self` / `&mut T`, stdlib's blanket `impl<T> !UnwindSafe for &mut T` would otherwise reject every `&mut` method unconditionally, so the bound is deliberately weakened to the same `RefUnwindSafe` rule (the *logical* unwind-safety property on the underlying type), separate from the closure's own `UnwindSafe` bound.

The helpers are named `ensure_*` rather than `assert_*` to avoid confusion with `std::panic`'s `*_assert_unwind_safe` family (which means the *opposite* — wrap in `AssertUnwindSafe`) and to reflect that they are compile-time bounds, not runtime assertions.

```rs
// before — no compile error under panic=unwind
#[wasm_bindgen]
pub fn touch(_c: &Cell<u32>) {}

// after — produces:
//   error[E0277]: the type `UnsafeCell<u32>` may contain interior
//   mutability and a reference may not be safely transferable across
//   a catch_unwind boundary
//     |
//     | required by a bound in `wasm_bindgen::__rt::ensure_ref_unwind_safe`
```

Users whose type is genuinely safe to observe after a caught panic can opt in via either:

* `impl RefUnwindSafe for MyType {}`, or
* wrapping interior-mutable fields in `std::panic::AssertUnwindSafe`.

The in-tree examples that needed updating:

* `tests/wasm/getters_and_setters.rs::SetterCompute(Rc<Cell<u32>>)` — the existing setter is a single primitive `Cell::set` call which cannot panic mid-mutation, so the field is wrapped in `AssertUnwindSafe`. Demonstrates the recommended user pattern.
* `crates/test/src/rt/mod.rs::WasmBindgenTestContext` — the test harness intentionally drives state through interior mutability while orchestrating panicking test execution, so it gets a manual `impl RefUnwindSafe`. Demonstrates the alternative opt-in.

Test coverage:

* New runtime test `tests/wasm/unwind.rs::ref_unwind_safe_method_runtime_behavior` exercises an exported `&mut self` method that panics mid-mutation on a `RefUnwindSafe` struct, verifying both that the panic is caught and that pre-panic mutations remain observable on subsequent calls.
* Inline-documented negative cases (commented-out structs whose uncommenting fails to compile under `panic = "unwind"`).
* All 354 existing wasm tests pass under `RUSTFLAGS=-Cpanic=unwind` with the new check enabled.

The check is a complete no-op outside `panic = "unwind"` builds. `panic = "unwind"` is nightly-only and explicitly experimental, so the slightly-breaking surface (existing crates with `&self`/`&mut self` methods on interior-mutable structs that build with `panic=unwind` will now require an explicit opt-in) is in scope for the feature's correctness goals.

#### Alternative considered: per-struct bound on the exported class itself

An alternative explored was emitting a single `ensure_ref_unwind_safe::<Self>()` (and `ensure_unwind_safe::<Self>()`) per exported struct in its codegen, rather than per call site, leaning on `RefUnwindSafe`'s auto-trait propagation through fields. This is appealingly simpler but is strictly over-tightening: it fires on exported handle structs that hold `!UnwindSafe` data (e.g. `pub struct ClosureHandle { _closure: Closure<dyn FnMut()> }`) but never expose a `&self` / `&mut self` / owned-`self` path — so JS can never re-enter them after a caught panic and there is no torn-state risk. The per-call-site approach correctly fires only where re-entrancy is actually possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,17 @@
 
 ### Changed
 
+* `#[wasm_bindgen]` exports now enforce unwind-safety at compile time under
+  `panic = "unwind"`. The generated wrapper for every `&self` / `&mut self`
+  method asserts `Self: RefUnwindSafe`, and reference arguments `&T` / `&mut T`
+  assert `T: RefUnwindSafe`, before the body runs inside `catch_unwind`. This
+  catches at compile time methods whose interior-mutable state could be
+  silently torn by a caught panic (e.g. a `RefCell` left borrowed, a `Cell`
+  updated halfway through restoring an invariant). The check is a no-op
+  outside `panic = "unwind"` builds. Users whose type is genuinely safe can
+  opt in with `impl RefUnwindSafe for MyType {}` or by wrapping interior-
+  mutable fields in `std::panic::AssertUnwindSafe`.
+
 * Simplified generated `web-sys` bindings by omitting redundant
   `#[wasm_bindgen]` attributes when they match wasm-bindgen defaults, including
   structural method annotations and matching `js_name` entries. The
--- a/crates/macro-support/src/codegen.rs
+++ b/crates/macro-support/src/codegen.rs
@@ -688,7 +688,7 @@ impl TryToTokens for ast::Export {
                     // Owned `self` is consumed inside the catch-unwind closure;
                     // assert it's `UnwindSafe` so a panic mid-method doesn't
                     // surface a half-modified observable value to the caller.
-                    #wasm_bindgen::__rt::assert_unwind_safe::<#class>();
+                    #wasm_bindgen::__rt::ensure_unwind_safe::<#class>();
                     let me = unsafe {
                         <#class as #wasm_bindgen::convert::FromWasmAbi>::from_abi(me)
                     };
@@ -708,7 +708,7 @@ impl TryToTokens for ast::Export {
                     // `&mut self` method, so we use a separate type-level
                     // assertion rather than relying on closure capture
                     // inference.
-                    #wasm_bindgen::__rt::assert_ref_unwind_safe::<#class>();
+                    #wasm_bindgen::__rt::ensure_ref_unwind_safe::<#class>();
                     let mut me = unsafe {
                         <#class as #wasm_bindgen::convert::RefMutFromWasmAbi>
                             ::ref_mut_from_abi(me)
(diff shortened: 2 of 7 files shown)


hoodmane (approved):
It needs some ui tests for the compile errors but otherwise looks good to me.

hoodmane:
Also, the name `assert_unwind_safe` is not great because here it means `compile_error_if_not_unwind_safe`

guybedford:
@RReverser I tried applying this at the class-level, but that actually resulted in an unnecessarily strict semantic since `&self`, `&mut self`, `self` and `mut self` all have different unwind safety properties, which gets discovered at the function-level rather. For example a class with only `&self` references only needs `RefUnwindSafe`. Sticking with the current per-function approach for this reason.

guybedford:
Renamed to `ensure_unwind_safe` `ensure_ref_unwind_safe`.

PR #5138, opened 2026-05-03
Update experimental ViewTransition interface

jhgarner:
### Description
A level 2 version of ViewTransitions has been published and has been implemented by more browsers than level 1. This PR updates the WebIdl to match version 2. The MDN docs for the latest version are here: https://developer.mozilla.org/en-US/docs/Web/API/Document/startViewTransition

### Checklist


guybedford:
I pushed an update here to use `DOMString` instead of `CSSOMString` to ensure better output.

guybedford (approved):

PR #5139, opened 2026-05-04
Add js_sys::AggregateError and typed ErrorOptions bindings

guybedford:
This adds `js_sys::AggregateError` bindings and a typed `js_sys::ErrorOptions` dictionary used by the built-in error constructors.

`AggregateError` is a subclass of `Error` that wraps multiple unrelated errors in a single error — most commonly produced by `Promise.any` when all input promises reject.

`ErrorOptions` is the typed counterpart to the untyped options dictionary previously accepted as `&Object` by `Error::new_with_options`. Now that the spec defines a single property (`cause`) shared by every built-in error constructor, an opaque typed dictionary lets callers construct the options without `Reflect::set` boilerplate, and lets every error type expose a `new_with_options` overload with a uniform signature.

Added:

* `AggregateError`, extending `Error` / `Object`, with three constructor overloads:
  * `AggregateError::new(errors: &[JsValue])`
  * `AggregateError::new_with_message(errors: &[JsValue], message: &str)`
  * `AggregateError::new_with_options(errors: &[JsValue], message: &str, options: &ErrorOptions)`
  and an `errors()` getter returning `Array`.
* `ErrorOptions`, extending `Object`, with `get_cause` / `set_cause` and a Rust-side `ErrorOptions::new(cause: &JsValue)` constructor.
* `new_with_options(message: &str, options: &ErrorOptions)` on every built-in error type that previously only had `new(message)`: `EvalError`, `RangeError`, `ReferenceError`, `SyntaxError`, `TypeError`, `URIError`, `WebAssembly.CompileError`, `WebAssembly.LinkError`, `WebAssembly.RuntimeError`.
* `Error::new_with_error_options(message: &str, options: &ErrorOptions)` alongside the existing untyped `Error::new_with_options(&str, &Object)` (kept for back-compat).

`AggregateError` takes errors as `&[JsValue]` rather than `&[&Error]` because the spec doesn't constrain element type — `Promise.any` itself can reject with non-`Error` values.

Before:

```rs
let cause = Error::new("the cause");
let options = Object::new();
Reflect::set(&options, &"cause".into(), &cause).unwrap();
let error = AggregateError::new_with_options(&[inner.into()], "Hello", &options);
```

After:

```rs
let cause = Error::new("the cause");
let options = ErrorOptions::new(&cause.into());
let error = AggregateError::new_with_options(&[inner.into()], "Hello", &options);
```

## Tests

`crates/js-sys/tests/wasm/AggregateError.rs` covers basic construction and inheritance (`AggregateError`, `Error`, `Object`), `new_with_message` setting `message` and inheriting the `AggregateError` `name`, `new_with_options` propagating `cause` through `Error.cause` (and asserting `ErrorOptions::new` round-trips `get_cause`), the `errors` getter round-tripping multiple errors, the empty-errors case, and the inherited `name` property.

## Changelog

Entries under Unreleased > Added.


crates/js-sys/src/lib.rs
```diff
@@ -438,6 +438,51 @@ extern "C" {
...
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError/AggregateError)
+    #[wasm_bindgen(constructor)]
+    pub fn new_with_options(errors: &[JsValue], message: &str, options: &Object) -> AggregateError;
+
+    /// The `errors` property of an `AggregateError` instance is an array
+    /// representing the errors that were aggregated.
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError/errors)
+    #[wasm_bindgen(method, getter)]
+    pub fn errors(this: &AggregateError) -> Array;
```
hoodmane, line 487: Could this be typed as `Array<Error>`?

hoodmane (approved):

crates/js-sys/src/lib.rs
guybedford: It's defined in ecma-262 as an any type!

PR #5140, opened 2026-05-05
Add js_sys::FinalizationRegistry bindings

guybedford:
## Summary

Adds `js_sys::FinalizationRegistry` bindings, completing the pair with the existing `WeakRef` bindings. `FinalizationRegistry` (ES2021, all browsers / Node 14+) lets JS request a callback when a registered object is garbage-collected.

The bindings:

- `FinalizationRegistry` type, extending `Object`.
- `FinalizationRegistry::new(cleanup_callback: &Function<fn(JsValue) -> Undefined>) -> FinalizationRegistry`.
- `register(this, target: &JsValue, held_value: &JsValue)`.
- `register_with_token(this, target, held_value, unregister_token)` (overload of `register` with the optional third arg).
- `unregister(this, unregister_token: &JsValue) -> bool`.

### Callback type

The cleanup callback is typed as `&Function<fn(JsValue) -> Undefined>`, consistent with the unstable `Promise::new` constructor. Users construct one from a Rust closure via:

- `Function::from_closure(Closure::new(...))` for owned, JS-retained callbacks (the common case for `FinalizationRegistry`, since the callback may fire long after construction).
- `Function::closure_ref(&scoped_closure)` for borrowed scoped closures.

### Argument types

`target`, `held_value`, and `unregister_token` are all `&JsValue`. The spec requires `target` and `unregister_token` to be objects (or non-registered symbols), and permits `held_value` to be any JS value; modeling all three as `&JsValue` matches that flexibility (mirroring the upstream WeakRef-style signature) without forcing users to wrap symbols in `Object`.

## Tests

`crates/js-sys/tests/wasm/FinalizationRegistry.rs` — five tests covering:

- Construction and inheritance (`FinalizationRegistry`, `Object`).
- `register_with_token` followed by `unregister(token)` returning `true`, and a second `unregister(token)` returning `false`.
- `unregister` with an unknown token returning `false`.
- `register` (no token overload) not throwing.
- `held_value` accepting null, strings, and numbers.

GC firing is intentionally not exercised — finalization timing is implementation-defined and not portable to test. All 5 tests pass under `wasm-bindgen-test` on Node.

## Changelog

Added an entry under Unreleased > Added.


crates/js-sys/src/lib.rs
```diff
@@ -4834,6 +4834,74 @@ impl Default for Function {
...
+    ///
+    /// [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry)
+    #[wasm_bindgen(extends = Object, typescript_type = "FinalizationRegistry<any>")]
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    pub type FinalizationRegistry;
+
+    /// Creates a new `FinalizationRegistry` with the given cleanup callback.
+    ///
+    /// The cleanup callback is invoked, at some point after a registered
+    /// target is garbage-collected, with the `heldValue` that was passed to
```
hoodmane:
```suggestion
    /// target is garbage-collected, with the `held_value` that was passed to
```

hoodmane (approved):

PR #5142, opened 2026-05-06
Fix panic unwind bundler js

alexjg:
### Description

This PR fixes a bug encountered when using the `-C panic=unwind` feature in Automerge [here](https://github.com/automerge/automerge/pull/1363). The problem appears to be that when building for the bundler target with `-C panic=unwind` wasm-bindgen produces a javascript wrapper file with a line like this in it:

    export const __wbindgen_wrapped_jstag = __wbindgen_wrapped_jstag;

This is not valid javascript and is rejected by most bundlers and runtimes as it's a redeclaration of a constant.

I've implemented a fix in two commits. The first commit introduces a reference test in `crates/cli/tests/reference/panic-unwind.bg.js` which captures the current state (the invalid JS above). Then a second commit which fixes the issue. 

### Checklist


crates/cli/tests/reference/panic-unwind.bg.js
```diff
@@ -0,0 +1,119 @@
...
+    const table = wasm.__wbindgen_externrefs;
+    const offset = table.grow(4);
+    table.set(0, undefined);
+    table.set(offset + 0, undefined);
+    table.set(offset + 1, null);
+    table.set(offset + 2, true);
+    table.set(offset + 3, false);
+}
+
+export const __wbindgen_jstag = WebAssembly.JSTag;
```
guybedford, line 40: I wonder if these even need to be exported at all here?

guybedford (commented):
Thanks @alexjg for digging into this one, this is a great refactoring.

crates/cli/tests/reference/panic-unwind.bg.js
guybedford: Of course they do, since this satisfies the imports in bundler mode. PR is good to land.

guybedford (approved):

PR #5143, opened 2026-05-06
Add &Vec<T> outgoing argument support

guybedford:
This implements `&Vec<T>` as an argument type for imported JS functions called from Rust, which previously failed to compile.

Today calling an imported JS function with a vector requires giving up ownership: you have to pass `Vec<T>`, after which Rust no longer holds it. That's surprising, because the data crosses the boundary by copy for every element type except primitive typed arrays — JS receives a fresh allocation in every case. There's no Rust-side ownership reason to require the move.

`&Vec<T>` lets the caller hand a vector to JS and keep using it afterwards. The wire format mirrors owned `Vec<T>`: Rust clones the elements into a freshly-allocated buffer that JS owns and frees. The caller's `Vec<T>` is left untouched. As a side effect, the JS-visible type for primitive element kinds becomes a plain `Array` rather than a typed array (because the wrapped value is now semantically "an array of values", not a transferred buffer).

Before:

```rs
let v = vec!["a".to_string(), "b".to_string()];
js_func(v);     // ownership lost
let _ = v;      // compile error
```

After:

```rs
let v = vec!["a".to_string(), "b".to_string()];
js_func(&v);    // borrowed
assert_eq!(v.len(), 2);   // still ours
```

Implemented:

* `IntoWasmAbi` / `OptionIntoWasmAbi` for `&Vec<T>` cloning into `Box<[T]>` and reusing the owned-vector ABI. Requires `T: Clone`.
* New CLI instructions `VectorLoadAsArray` / `OptionVectorLoadAsArray` emitted for `Descriptor::Vector(_)` in ref position, wrapping the resulting buffer in `Array.from(...)` for primitive element kinds.

Test coverage: wasm runtime tests in `tests/wasm/vec_ref.{rs,js}` for `&Vec<u8|u16|i32|f64|String>` and `Option<&Vec<u16>>`, plus a reference test in `crates/cli/tests/reference/vec-ref.*` locking in the codegen. A regression test confirms exported `Vec<u16>` still accepts plain `Array<number>` from JS callers.

No existing types change behavior; `&Vec<T>` was a compile error before, so nothing regresses. Owned `Vec<T>` keeps its current semantics for backward compatibility.


hoodmane:
> As a side effect, the JS-visible type for primitive element kinds becomes a plain Array rather than a typed array (because the wrapped value is now semantically "an array of values", not a transferred buffer).

So `Vec<i32>` turns into a `TypedArray` but `&Vec<i32>` turns into an `Array`? That seems really confusing.

guybedford:
> So Vec<i32> turns into a TypedArray but &Vec<i32> turns into an Array? That seems really confusing.

I think it's beneficial to have a unified bindgen type where `&Vec<T> -> Array<T>` from a generics perspective. This would be far more beneficial than slice special casing we have today.

guybedford:
Per @hoodmane's feedback here, I'm closing in favour of https://github.com/wasm-bindgen/wasm-bindgen/pull/5145.

PR #5145, opened 2026-05-06
Add slice_to_array attribute for plain-Array slice imports

guybedford:
This adds the `slice_to_array` attribute for imported JS functions, an opt-in per-fn or per-`extern "C"`-block that lets idiomatic `&[T]` arguments be used when binding JS APIs that take a plain `Array<T>` rather than a typed array.

Idiomatic Rust uses `&[T]` for borrowed sequences. When binding a JS function that takes an array of values, the natural Rust signature is `fn foo(items: &[T])`. The default `&[T]` binding is split by element kind today: primitive `T` arrives as a typed-array view (`Uint32Array` etc.) because of a zero-copy buffer optimisation, while `String` and JS-imported types arrive as a plain `Array`. There's no way to ask for the plain-`Array` form for primitive `T` without abandoning `&[T]` for `Vec<JsValue>` or hand-rolled plumbing.

`slice_to_array` makes `&[T]` (and `Option<&[T]>`) arrive as a plain `Array` regardless of element kind, while keeping the user-facing `&[T]` signature unchanged:

```rust
#[wasm_bindgen]
extern "C" {
    // JS receives Array<number>, signature stays idiomatic.
    #[wasm_bindgen(slice_to_array)]
    fn set_indices(values: &[u16]);
}
```

Block-form for module bindings with a consistent convention:

```rust
#[wasm_bindgen(module = "/lib.js", slice_to_array)]
extern "C" {
    fn take_numbers(v: &[i32]);
    fn take_strings(v: &[String]);
    fn take_optional(v: Option<&[u16]>);
}
```

Implementation:

* `slice_to_array` parsed in the macro at fn-arg, fn, and extern block levels; purely additive opt-in.
* New internal `VectorRefIntoWasmAbi` trait dispatches the per-element ABI conversion. Two impl shapes, neither requiring `T: Clone`:
  - Primitive `T`: zero-copy borrow of the slice memory (no allocation, same wire as plain `&[T]`). The JS-side shim does `Array.from(typedArrayView)` and never frees the buffer.
  - Blanket `for<'a> &'a T: Into<JsValue>` (covers `String`, `JsValue`, and JS-imported types): iterate the slice and build a fresh `[u32]` index buffer, one externref per element. JS reads the indices into a plain `Array` and frees the index buffer.
* New CLI instructions `VectorLoadAsArray` / `OptionVectorLoadAsArray` bind the resulting descriptor; the JS shim picks the right helper (and decides whether to free) based on the element `VectorKind`.
* `&[ExportedRustStruct]` remains unsupported — the existing default `&[T]` doesn't support exported structs either, and a borrow form for them would require either deep cloning the data per element or reworking the exported-struct ownership model. Owned `Vec<T>` continues to work for that case.

Has no effect on exported functions; default `&[T]` (typed-array view / zero-copy memory borrow) and owned `Vec<T>` semantics are unchanged for callers that didn't opt in.

Test coverage:

* wasm runtime tests in `tests/wasm/slice_to_array.{rs,js}` covering `&[u8|u16|i32|f64]`, `&[String]`, `&[ImportedType]`, `Option<&[u16]>`, plus block-level fan-out.
* reference test in `crates/cli/tests/reference/slice-to-array.*` locking in the bg.js / wat output, including the no-free codegen for primitive borrows.
* new guide page at `reference/attributes/on-js-imports/slice_to_array.html`.


crates/cli-support/src/js/binding.rs
```diff
@@ -1832,6 +1832,42 @@ fn instruction(
...
+            //   owns; the helper drops the externrefs as it reads them
+            //   into a plain `Array`, and we free the index buffer.
+            let len = js.pop();
+            let ptr = js.pop();
+            let f = js.cx.expose_get_vector_from_wasm(kind.clone(), *mem);
+            let i = js.tmp();
+            match kind {
+                VectorKind::String | VectorKind::Externref | VectorKind::NamedExternref(_) => {
+                    let free = js.cx.export_name_of(*free);
+                    js.prelude(&format!("var v{i} = {f}({ptr}, {len}).slice();"));
```
hoodmane: From `slice-to-array.bg.js` the `slice()` looks like an unnecessary copy.
```suggestion
                    js.prelude(&format!("var v{i} = {f}({ptr}, {len});"));
```

crates/cli/tests/reference/slice-to-array.bg.js
```diff
@@ -0,0 +1,103 @@
...
+    let v0;
+    if (arg0 !== 0) {
+        v0 = Array.from(getArrayU16FromWasm0(arg0, arg1));
+    }
+    js_slice_optional_u16_as_array(v0);
+}
+export function __wbg_js_slice_string_as_array_fc6d35fea48d3225(arg0, arg1) {
+    var v0 = getArrayJsValueFromWasm0(arg0, arg1).slice();
+    wasm.__wbindgen_free(arg0, arg1 * 4, 4);
+    js_slice_string_as_array(v0);
```
hoodmane, line 21: Where does `js_slice_string_as_array()` get defined?

hoodmane (approved):
Generally looks okay.

crates/cli/tests/reference/slice-to-array.bg.js
guybedford: This is the JS imported function being wrapped for Wasm calling.

hoodmane: Oh right makes sense.

PR #5149, opened 2026-05-07
Loosen JsFuture<T> bound to support Promise<Union>

guybedford:
Minor patch on top of #4734. Imported `async fn` returning a dynamic-union enum (i.e. `Promise<Union>` resolution) failed to compile because `From<Promise<T>> for JsFuture<T>` required `T: JsGeneric`. `JsGeneric` demands `JsCast` + `AsRef<JsValue>` (layout equivalence with `JsValue`), which a tagged-enum union cannot satisfy.

The closure shim that materialises the resolved value calls `T::from_abi`, which is per-type ABI conversion — `T: FromWasmAbi` is the actual minimum needed at this seam. For union types `from_abi` runs the variant dispatcher.

```rs
impl<T: JsGeneric + FromWasmAbi> From<Promise<T>> for JsFuture<T> { ... }
```

becomes:

```rs
impl<T: FromWasmAbi + 'static> From<Promise<T>> for JsFuture<T> { ... }
```

Same change applied to `IntoFuture for Promise<T>`. No behaviour change for existing `JsGeneric` types (they all impl `FromWasmAbi`).

Adding to 0.2.121 as it has not yet gone out.

This intentionally does not address `Array<Union>` / `Map<K, Union>` and similar — those go through the `ErasableGeneric` macro check at a different layer and need a separate fix.

### DIFF
--- a/crates/js-sys/src/futures/mod.rs
+++ b/crates/js-sys/src/futures/mod.rs
@@ -127,7 +127,11 @@ impl<T> fmt::Debug for JsFuture<T> {
     }
 }
 
-impl<T: JsGeneric + FromWasmAbi> From<Promise<T>> for JsFuture<T> {
+// `FromWasmAbi` is what the closure shim invokes on the resolved value;
+// no layout equivalence with `JsValue` is required at this seam — the
+// per-type `from_abi` does the conversion (e.g. for dynamic unions it
+// runs the variant dispatcher).
+impl<T: FromWasmAbi + 'static> From<Promise<T>> for JsFuture<T> {
     fn from(js: Promise<T>) -> JsFuture<T> {
         // Use the `then` method to schedule two callbacks, one for the
         // resolved value and one for the rejected value. We're currently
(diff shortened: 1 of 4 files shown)


crates/js-sys/src/futures/mod.rs
```diff
@@ -127,7 +127,11 @@ impl<T> fmt::Debug for JsFuture<T> {
     }
 }
 
-impl<T: JsGeneric + FromWasmAbi> From<Promise<T>> for JsFuture<T> {
+// `T: FromWasmAbi` is the actual minimum the closure shim needs to
+// materialise the resolved value; the previous `T: JsGeneric` bound also
+// demanded layout equivalence with `JsValue`, which excludes dynamic-union
+// enums (tagged enums whose `from_abi` runs a variant dispatcher).
```
hoodmane: IMO these "previously it did X but I changed it because it didn't work" comments are kind of a comment antipattern unless you think someone is going to read this and think "Why don't we do X".

hoodmane (approved):
Suggest you remove the comment or make it not in terms of the previous state of the code base. Otherwise makes sense to me.

crates/js-sys/src/futures/mod.rs
guybedford: Reduced.

PR #5152, opened 2026-05-08
Fix CI for latest rustc nightly

guybedford:
This fixes CI on current nightly Rust (2026-05-06+) with two recent issues:

1. modern exception handling is now the default as of 
2. atomics-flagged builds fail with `failed to find \`__heap_base\` for injecting thread id`.

### Use Node.js 24.15+ 22.22.3+

Node.js modern exception handling requires 22.22.3+ OR 24.15+.

We change the tests to use this version, and then explicitly turn on legacy exception handling for those tests.

### Add --export=__heap_base for threading after rust-lang/rust#156174

[rust-lang/rust#156174](https://github.com/rust-lang/rust/pull/156174) (merged 2026-05-06) intentionally removed the implicit `__heap_base`/`__data_end` exports on `wasm*-unknown-unknown` and `wasm32v1-none`, requiring threading toolchains like wasm-bindgen to opt in explicitly. The PR description spells out the required link arg directly:

> After this PR the following would be required for multi-threading support in `wasm-bindgen`:
> `... -Clink-arg=--shared-memory ... -Clink-arg=--export=__heap_base -Clink-arg=--export=__wasm_init_tls ...`

`wasm-bindgen`'s threading post-processing needs `__heap_base` (see `crates/cli-support/src/transforms/threads/mod.rs:193`) to set up per-thread stacks, so without the new export it bails immediately.

This PR adds `-Clink-arg=--export=__heap_base` everywhere the existing TLS exports are listed:

* `.github/workflows/main.yml` — atomics CI rustflags
* `crates/cli/tests/wasm-bindgen/reference.rs` — `runtest_targets_atomics` and `no_duplicate_wasm_export_in_node_esm_atomics_debug`
* `examples/nodejs-threads/package.json`
* `examples/raytrace-parallel/package.json`
* `examples/wasm-audio-worklet/package.json`

Verified locally: the fix makes the failing `headless` test pass on `nightly-2026-05-07` (`365c0e1d7`), and remains backward-compatible with older nightlies (`nightly-2026-05-06` / `e95e73209`).

PR #5120, opened 2026-04-20
feat: inheritance for Rust-exported types via #[wasm_bindgen(extends = ...)]

ZhaoXiangXML:
## Summary

Adds inheritance between Rust-exported `#[wasm_bindgen]` structs. Refs #1721 and closed RFC rustwasm/rfcs#11.

- `#[wasm_bindgen(extends = Parent)]` on a Rust struct with a `#[wasm_bindgen(parent)]` field now:
  - Emits `class Child extends Parent { ... }` in the generated JS and `.d.ts`, so `child instanceof Parent` is true.
  - Emits `AsRef<Parent>` / `AsMut<Parent>` / `Deref<Target = Parent>` / `DerefMut` on the Rust side, projecting through the parent field.
  - Threads the relationship through the shared wire format, `AuxStruct`, and `ExportedClass`.
  - Skips getter/setter emission for the parent field (it exists only for projection).
- Subclass constructors emit `super(__wbgSuperSkip)`. Every user-defined constructor starts with `if (arguments[0] === __wbgSuperSkip) return;`, so the parent's JS constructor body is short-circuited when invoked from a wasm-bindgen subclass — the child's own `#[wasm_bindgen(constructor)]` is the single source of truth for the parent value. The sentinel is module-level and inert for non-wasm-bindgen JS subclasses (they never pass it), preserving behaviour introduced by #3562.
- Parent methods inherit in JS via the prototype chain for plain JS methods. For methods exported from Rust, the supported pattern is a one-liner on the child using `Deref` (see guide). Automatic forwarding of Rust-exported parent methods is future work — it needs a design that accounts for the `Rc<WasmRefCell<T>>` wasm ABI used by exported structs.

## Example

```rust
#[wasm_bindgen] pub struct Animal { name: String }

#[wasm_bindgen]
impl Animal {
    #[wasm_bindgen(constructor)]
    pub fn new(name: String) -> Animal { Animal { name } }
    pub fn name(&self) -> String { self.name.clone() }
}

#[wasm_bindgen(extends = Animal)]
pub struct Dog {
    #[wasm_bindgen(parent)] parent: Animal,
    breed: String,
}

#[wasm_bindgen]
impl Dog {
    #[wasm_bindgen(constructor)]
    pub fn new(name: String, breed: String) -> Dog {
        Dog { parent: Animal::new(name), breed }
    }
    pub fn breed(&self) -> String { self.breed.clone() }

    // One-line re-export of the inherited method via Deref:
[...]

PR #5154, opened 2026-05-12
fix: route methods to the right class when js_name + js_namespace + js_class are combined

ZhaoXiangXML:
## Summary

Fixes a codegen issue where a renamed exported struct combined with a namespaced JS class produced a stub class as the namespace export. Methods and the constructor were written onto a fresh, unreachable JS class entry, so:

```rust
#[wasm_bindgen(js_name = "Foo", js_namespace = ns)]
pub struct FooImpl { /* ... */ }

#[wasm_bindgen(js_class = "Foo")]
impl FooImpl {
    #[wasm_bindgen(constructor)]
    pub fn new(/* ... */) -> FooImpl { /* ... */ }
    pub fn bar(&self) { /* ... */ }
}
```

…produced:

```js
const ns = {};
class ns__Foo { /* __wrap / __destroy_into_raw / free — no constructor, no bar() */ }
ns.Foo = ns__Foo;
```

so `new wasm.ns.Foo(args)` threw `Error: cannot invoke 'new' directly`, and any later `obj.bar()` reported `bar is not a function`. The class with the actual constructor + methods existed but was unreferenced.

## Root cause

`Context::require_class(name)` only resolved its input through `qualified_to_rust_name`. When the macro emits `js_class = "Foo"` for the impl block, that string flows in as `name`. It is neither a `rust_name` key nor a qualified-name key (the qualified form is `ns__Foo`), so the lookup fell through and a fresh class entry keyed by `"Foo"` was created. The struct-side registration keyed by `rust_name` (`"FooImpl"`) therefore stayed empty, and the namespace assignment emitted against the empty entry was a stub.

Non-renamed namespaced classes (where the rust ident matches the JS export name — the existing `Counter` test in this file) were not affected because the impl's `Self` ident happens to match the struct's `rust_name`.

## Fix

Key a new `js_name_to_rust_name: HashMap<String, String>` when each struct is registered (in `Context::generate`), and consult it in `require_class` after the existing `qualified_to_rust_name` lookup. Methods, constructors, getters, and setters now route back to the struct's rust-name entry, the namespace export resolves to a real class, and instances are usable.

### Behavior when there is no `js_name` rename

The new map insertion is guarded by `if s.name != s.rust_name`, so a struct without a `js_name` rename never adds an entry. The three lookups in `require_class` then resolve like this for an incoming `name` of `"Foo"`:

| `#[wasm_bindgen(...)]` on struct | `qualified_to_rust_name.get("Foo")` | `js_name_to_rust_name.get("Foo")` | Final `key` | Outcome |
| --- | --- | --- | --- | --- |
| `#[wasm_bindgen]` (no rename, no namespace; `Self::Foo` impl) | None | None (no insert) | `"Foo"` (fallthrough) | `exported_classes["Foo"]` resolves to the struct entry directly — no change vs. before this PR. |
| `#[wasm_bindgen(js_namespace = ns)]` (no rename; `Self::Foo` impl) | None for `"Foo"` (qualified key is `"ns__Foo"`) | None (no insert) | `"Foo"` (fallthrough) | Same as before — the existing `Counter` test exercises this path. |
| `#[wasm_bindgen(js_name = "Foo", js_namespace = ns)]` + `#[wasm_bindgen(js_class = "Foo")] impl FooImpl` (the failing case) | None for `"Foo"` (qualified key is `"ns__Foo"`) | `Some("FooImpl")` (inserted by the new branch) | `"FooImpl"` | Methods land on the existing struct entry keyed by `rust_name`; namespace export wires through. |
| `#[wasm_bindgen(js_name = "Foo")]` + `#[wasm_bindgen(js_class = "Foo")] impl FooImpl` (rename without namespace) | None | `Some("FooImpl")` | `"FooImpl"` | Same fix path; pre-PR this also routed methods to the wrong entry, just with no namespace export to make it visible. |

So the fix is a no-op on non-renamed structs (the `qualified_to_rust_name` lookup or the fallthrough already handled them); it only changes behavior when the user supplies `js_name` and the `js_class` value on the impl block matches that renamed name.

## Test plan

Extends `tests/wasm/js_namespace_exports.{rs,js}` with three new test functions:

- `renamed_namespaced_class_methods` — rename on struct + namespace + `js_class` on impl. Was the failing pattern; now passes.
- `renamed_class_namespace_on_struct_only` — rename + namespace on struct, `js_class` on impl with no namespace. Also failed before the fix; now passes.
- `namespaced_class_methods_same_name` — no rename, namespace on both struct and impl. Already passed before the fix; serves as a positive control for the no-rename branch in the table above.

Pre-fix: 2 of the 3 fail with `cannot invoke 'new' directly` raised by the stub class. Post-fix: all 3 pass.

```
$ cargo test --target wasm32-unknown-unknown -p wasm-bindgen --test wasm 'js_namespace_exports::'
running 4 tests
test js_namespace_exports::test_namespaced_exports ... ok
test js_namespace_exports::renamed_namespaced_class_methods ... ok
test js_namespace_exports::renamed_class_namespace_on_struct_only ... ok
test js_namespace_exports::namespaced_class_methods_same_name ... ok

$ cargo test --target wasm32-unknown-unknown -p wasm-bindgen --test wasm
test result: ok. 372 passed; 0 failed; 3 ignored
```

Also verified end-to-end against a downstream project (an internal Rust→wasm-bindgen codegen pipeline) that hit this bug on every renamed namespaced class (~6 distinct classes). With the fix patched in via `[patch.crates-io]`, the previously-failing vitest cases all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ZhaoXiangXML:
@guybedford could I get a review on this? It's a small fix in the same area as #5106 / #5120 — both of which you merged — and the trigger condition is documented end-to-end in the description (including a truth table for the no-rename baseline). Pre-existing 4 CI failures are all from the rust-lang/rust#156174 `__heap_base` regression that #5152 fixes, not from this change.

ZhaoXiangXML:
The two remaining red checks on the latest push are unrelated to this PR's change. Recording context for reviewers:

1. **`Run wasm-bindgen crate tests with panic=unwind with legacy eh`** — present on upstream `main` itself at the same base SHA (`14b86ad6`). The push CI on `main` for that commit failed on this exact job: [run 25772470078](https://github.com/wasm-bindgen/wasm-bindgen/actions/runs/25772470078). The crashing test is `no_interpret` with `CompileError: ... invalid value type 'noexternref', enable with --experimental-wasm-gc` under Node 20 — environmental to the new exnref-EH-by-default work in #5152, not anything in this PR.

2. **`CodSpeed Performance Analysis`** — flagged `dynamic_union_roundtrip` -14.41% (1.1µs → 1.3µs). The only code change in this PR is a build-time `HashMap` lookup in `Context::require_class` (`crates/cli-support/src/js/mod.rs`), which never executes during a runtime roundtrip benchmark. A 0.2µs swing on a 1.1µs wall-time bench is well within typical noise; happy to acknowledge it in the CodSpeed UI if preferred.

For completeness: the previous force-push had a legitimate failure in `reference::runtest::test_24_js_namespace_export_same_name_rs`. That fixture (`js-namespace-export-same-name`) already exercises the same `rename + js_namespace + js_class = "<renamed>"` pattern this PR fixes, so its generated output legitimately changes — the regenerated `bg.js` / `d.ts` / top-level export list now carry constructors, methods, and getter/setter properties on `foo__RefToBar` / `bar__RefToFoo` instead of the broken phantom-class form the old snapshot encoded. Refreshed via `BLESS=1` and bundled into the same commit.

guybedford:
@ZhaoXiangXML Thanks for the patch. After reviewing, the original symptom turns out to be the tip of a larger architectural inconsistency — cli-support's `exported_classes` was keyed by `rust_name`, which silently breaks for any two structs that share a Rust identifier in different modules with distinct `js_name`s (a legitimate cross-module pattern). The macro-side wasm shim symbols for impl methods also lacked the namespace prefix, so they could collide at wasm-ld in the same scenario.

Pushed an additional commit (4c3ac050ad) on top of yours that completes the fix. It builds on your macro-side namespace plumbing rather than replacing it.

### What changed

**The principle**: `js_class` (qualified by `js_namespace`) is the primary class identity throughout cli-support. `rust_name` becomes a default value for `js_name` when not specified — it's only ever a JS-side name, never a key.

**Concretely**:

* `cli-support` keys `exported_classes` by `qualified_name` (the struct's JS-side identity: `<js_namespace>__<js_name>`). Lookup sites updated throughout (`require_class`, inheritance walk, upcast symbol generation, finalization registry naming).
* Wasm upcast shim symbols use the parent's qualified name on both macro and cli-support sides — they previously disagreed when the parent was renamed/namespaced.
* New `extends_js_class` / `extends_js_namespace` attributes let users declare the parent's JS-side identity on the child struct. Required when the parent has `js_name` / `js_namespace`; default sensibly otherwise so the no-rename case needs no ceremony.
* New `suggest` module: small hand-rolled Levenshtein helper. Drives "did you mean ...?" hints across impl-class and extends-parent resolution failures.
* Targeted error messages: when an impl block's `(js_class, js_namespace)` doesn't match any struct, cli-support tells the user the exact attribute to add ("a struct with the same `js_name` exists in a different namespace; add `js_namespace = ns` to the impl block").

### Why a different approach from this PR's first commit

The first commit added a parallel `js_name_to_rust_name` map to bridge the lookup. That correctly fixed the symptom for the renamed+namespaced single-struct case, but had a latent collision: two structs with the same `js_name` across different namespaces (`P.Foo` + `Q.Foo`) would both insert key `"Foo"` into the map — last writer wins. The keying refactor sidesteps this entirely by making JS identity (not Rust identity) the universal key.

The cli-support code in the new approach is also smaller net — the additional map and lookup branch go away, replaced by direct qualified-name lookup. The two pieces that get added (`suggest` module + validation pass) pay for themselves via the user-facing error quality.

### Breaking changes

Two new requirements:

1. **`js_class` + `js_namespace` on impl blocks must match the struct.** If the struct declares either, every impl block must declare matching values. Previously the impl-side defaults (Self ident, no namespace) silently lined up in some cases via rust_name keying.

2. **`extends_js_class` + `extends_js_namespace` on child structs when the parent uses `js_name` / `js_namespace`.** New attributes, required for the renamed/namespaced extends case; sensible defaults for the no-rename case.

### Why the breaks are okay here

Both surfaces are very recent:

* `js_namespace` on exported structs was the subject of #5106 (merged 2026-04-14) and is what this PR is fixing further.
* `extends = Parent` on Rust-exported structs is #5120 (merged 2026-05-06), shipped less than a week before this PR was opened.

Neither has had time to accumulate broad ecosystem usage. The set of users affected is essentially the set of users whose code is currently broken by the bugs being fixed — same population. Build errors contain precise migration hints ("add `js_namespace = ns` to the impl block" with the exact value), so the migration is mechanical.

Formalizing the model now — while the feature surface is small — is much cheaper than waiting for a major version bump. Doing it later would mean either shipping the partial fix and leaving the architectural inconsistency in place (with users continuing to hit silent edge cases), or accumulating more affected users that would need to migrate later.

### Tests

* All 50 reference tests pass; fixtures updated where they relied on the implicit-namespace bridge.
* 374 wasm-bindgen tests pass, including two new fixtures exercising the previously-broken patterns: two same-Rust-ident structs with distinct `js_name`s, and renamed + namespaced `extends`.
* 7 new unit tests in `cli-support/src/suggest.rs` cover the Levenshtein helper.
* 3 integration diagnostic tests in `crates/cli/tests/wasm-bindgen/diagnostics.rs` assert the hint output for impl-block misses (single namespace, nested namespace, typo / fuzzy suggestion).

### Guide updates

`js_class.md`, `js_namespace.md`, and `extends.md` updated with the new requirement, example error output, and a worked example for the renamed+namespaced extends case.

ZhaoXiangXML:
Thanks for taking this all the way, @guybedford — keying `exported_classes` by qualified JS identity is clearly the right model, and building it on top of the macro-side plumbing rather than replacing it keeps the diff easy to follow.

I've verified `4c3ac050ad` end-to-end: built `wasm-bindgen-cli` from this branch head and ran a full `wasm-pack build` against a non-trivial downstream crate that exercises the affected surfaces together — exported structs with `js_name` + `js_namespace`, `extends = Parent` inheritance (including renamed/namespaced parents), and exported C-style enums. Build is green, no regressions.

The two breaking changes behave exactly as described:
- The `js_class` / `js_namespace`-on-impl and `extends_js_class` / `extends_js_namespace` requirements surfaced cleanly, and each error names the precise attribute (and value) to add — migration was mechanical, no guesswork.
- The "did you mean …?" suggestions caught a typo for me along the way — nice touch.

I think this is ready. Would you be able to merge and cut a release? Having it in a published version rather than a git patch would unblock downstream adoption.

guybedford (approved):

ZhaoXiangXML:
Thanks for the merge! Any sense of when the next release might be? Recent cadence looks like roughly a 10-day rhythm, so I wanted to check in — a downstream project of mine is currently pinning to a git rev for this fix and I'd like to switch to a published version when I can. Not blocking anything urgent.

PR #5155, opened 2026-05-13
ci: pin legacy EH job to nightly-2026-05-12

guybedford:
The legacy EH CI job has started failing because recent nightlies emit wasm-gc value types (e.g. `noexternref`) in the standard library that Node.js 20's V8 11.3 cannot validate, even when explicitly opted into legacy EH via `-Cllvm-args=-wasm-use-legacy-eh`.

The failure surfaces during wasm validation, not the EH wrappers themselves:

```
CompileError: WebAssembly.Module(): Compiling function ... failed:
invalid value type 'noexternref', enable with --experimental-wasm-gc
```

Pin the legacy EH job to `nightly-2026-05-12` (the last nightly without this issue) so we keep Node 20 compatibility coverage. Dated nightlies remain available on the rust-lang dist server indefinitely, so this pin is durable.

The catch-unwind guide and CHANGELOG `Notices` section are updated to call out the same requirement so downstream users targeting Node 20 know to pin their nightly too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -169,23 +169,30 @@ jobs:
     env:
       WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
       # rust-lang/rust#156061 flipped the wasm panic=unwind default to modern
-      # (exnref) EH, so legacy EH now requires an explicit opt-in via the LLVM
-      # flag below. Node 20 only supports legacy EH.
-      RUSTFLAGS: -Cpanic=unwind -Cllvm-args=-wasm-use-legacy-eh
+      # (exnref) EH. To keep legacy EH coverage we pin nightly to before that
+      # change (see toolchain pin below) so plain `-Cpanic=unwind` still
+      # produces legacy EH. `-Cllvm-args=-wasm-use-legacy-eh` cannot override
+      # the new target spec's `-wasm-use-legacy-eh=false` (LLVM is last-wins
+      # and rustc puts target args after user args), so a pinned nightly is
+      # currently the only reliable way to test legacy EH.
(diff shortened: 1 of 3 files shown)


hoodmane (approved):

.github/workflows/main.yml
```diff
@@ -169,18 +169,30 @@ jobs:
...
       WASM_BINDGEN_SPLIT_LINKED_MODULES: 1
       # rust-lang/rust#156061 flipped the wasm panic=unwind default to modern
-      # (exnref) EH, so legacy EH now requires an explicit opt-in via the LLVM
-      # flag below. Node 20 only supports legacy EH.
-      RUSTFLAGS: -Cpanic=unwind -Cllvm-args=-wasm-use-legacy-eh
+      # (exnref) EH. To keep legacy EH coverage we pin nightly to before that
+      # change (see toolchain pin below) so plain `-Cpanic=unwind` still
+      # produces legacy EH. `-Cllvm-args=-wasm-use-legacy-eh` cannot override
+      # the new target spec's `-wasm-use-legacy-eh=false` (LLVM is last-wins
+      # and rustc puts target args after user args), so a pinned nightly is
```
hoodmane, line 176: I don't know why it does this it has caused several different problems for me.

hoodmane (approved):

.github/workflows/main.yml
guybedford: Upstream discussion in https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Order.20of.20.60-Cllvm-arg.60.20and.20target.20spec.27s.20.60llvm_args.60/with/594937994.

PR #5156, opened 2026-05-15
Fix emscripten output: dep graph, HEAP_DATA_VIEW, function declarations

guybedford:
This makes wasm-bindgen's emscripten output work against vanilla emscripten (no fork required) and fixes several latent issues in the emscripten dependency-tracking layer.

The core change is making library dependency tracking explicit and complete. Previously the emscripten output relied on accidents — file-scope function declarations co-existing with `addToLibrary({...})` registrations, intrinsics emitted at one scope and referenced at another, library symbols pulled in only when the *first* caller happened to register them. Several of these patterns were broken under vanilla upstream emscripten and only worked against the wasm-bindgen-specific fork.

The per-adapter `adapter_deps` `BTreeSet<String>` is no longer cleared between adapters. Per-import `__deps` arrays are computed via snapshot/diff at the start and end of each `Import`/`Adapter` generation. The cumulative set then drives `$initBindgen__deps` so emcc's library tree-shaker keeps every symbol the inlined export bodies reference. The dep edge is registered on every caller, not just the first.

`intrinsic()` now takes `deps: &[&str]` so every helper declares the library symbols its body references. All 45 call sites audited and given accurate deps (e.g. `addHeapObject` declares `["heap", "heap_next"]`, `takeObject` declares `["getObject", "dropObject"]`).

`HEAP_DATA_VIEW` is now declared by wasm-bindgen and refreshed via wrapping `updateMemoryViews` in a postset:

```rust
imports.push_str(
    "$HEAP_DATA_VIEW: 'undefined',\n\
     $HEAP_DATA_VIEW__postset: \"var __wbg_origUpdateMemoryViews = updateMemoryViews; updateMemoryViews = function () { __wbg_origUpdateMemoryViews(); HEAP_DATA_VIEW = new DataView(wasmMemory.buffer); };\",\n",
);
```

Upstream emscripten only declares `HEAP_DATA_VIEW` under `SUPPORT_BIG_ENDIAN`, so the common little-endian case previously referenced an undefined variable.

`export_to_emscripten` handles function declarations — converting the wasm-bindgen-emitted `function name(args) { body }` snippets into inline `addToLibrary({ $name: function(args){body}, $name__deps: [...] })` rather than leaving a file-scope copy that captures library symbols out of scope.

Before:
```js
function passStringToWasm0(arg, malloc, realloc) { /* refs $cachedTextEncoder, $heap */ }
addToLibrary({ '$passStringToWasm0': passStringToWasm0 });
```

After:
```js
addToLibrary({
    $passStringToWasm0: function (arg, malloc, realloc) { /* body */ },
    $passStringToWasm0__deps: ['$cachedTextEncoder', '$WASM_VECTOR_LEN'],
});
```

The catch-wrapper transform is skipped on emscripten output — emscripten may link wasm EH instructions from libcxx/embind that have nothing to do with `#[wasm_bindgen(catch)]`, and the wasm-bindgen runtime intrinsics (`__externref_table`, `__externref_table_alloc`, `__wbindgen_exn_store`) may be absent. Previously bailed; now skips gracefully.

A pre-existing `HEAP_DATA_VIEW0()` bug in `getArrayJsValueFromWasm0` is corrected to use the right heap accessor (`HEAP_DATA_VIEW` directly on emscripten vs the `getDataViewMemory0()` cache function on other targets).

PIC module support is added to the interpreter — `__stack_pointer` and `__table_base` are recognized as imported globals; imported i32/i64 globals are seeded so descriptor execution works on emscripten-linked input. This is a small extension to `wasm_conventions.rs` and `interpreter/mod.rs`.

Architecture invariants verified: every library symbol referenced by any generated JS is reachable in emcc's dep graph (via `$initBindgen__deps`, per-import `__deps`, or `extraLibraryFuncs`); per-import `__deps` correctly diff against pre-import snapshot; `adapter_deps` accumulates monotonically so order of adapter generation doesn't matter; all emscripten-specific logic is gated, so non-emscripten output is unaffected; no file-scope function declarations leak into library output — everything is inside `addToLibrary({...})` literals.

---

**Update**: also fixes `#[wasm_bindgen(module = "...")]` ESM imports for emscripten output.

Previously these emitted a top-level `import { foo } from 'node:os';` into `library_bindgen.js`, but emcc's `--js-library` evaluator rejects ESM imports outright (it's a CommonJS-shaped DSL), and emcc's modularize wrapper isn't a legal scope for ESM imports either.

The fix routes ESM imports to a sidecar `library_bindgen.extern-pre.js` that consumers pass to emcc via `--extern-pre-js`. emcc inlines that file verbatim *before* the modularize wrapper, so the imports land at module top-level. Library functions inside the wrapper close over the imported bindings lexically:

```js
// extern-pre-js (module top-level — legal scope for imports)
import { hostname } from 'node:os';

// emcc modularize wrapper
var Module = (() => {
  return async function(moduleArg = {}) {
    // library functions inlined here, closing over `hostname` lexically
    function __wbg_hostname_xxx() { return hostname(); }
  };
})();
```

This unblocks `module = "..."` for any ESM specifier — `node:*` builtins, local snippet files, third-party packages. The same path applies to local-path snippets emitted by `#[wasm_bindgen(inline_js = "...")]` / `module = "/path/to/snippet.js"`.


guybedford:
@walkingeyerobot this incorporates a few changes I needed to get wasm-pack working with emscripten. In particular, the heap reference updates and scoping for intrinsics were quite big changes. The memory view fix also replaces what you needed to do in the upstream Emscripten branch allowing this to work with current Emscripten without even needing that PR currently. Would be good to discuss further at the next meeting.

walkingeyerobot (approved):
This looks good! My demos and such are all still working with this PR. I'm still going to add `-sWASM_BINDGEN` as an option in my Emscripte PR, but it's awesome to see this working with existing upstream Emscripten. I'll update my PR and if I do run into any trouble I'll let you know. Thanks again!

drcarney66 (approved):

PR #5160, opened 2026-05-20
fix: key WrapInExportedClass/UnwrapExportedClass by qualified_name (renamed-struct .into() duplicate class)

ZhaoXiangXML:
## Summary

A struct renamed via `#[wasm_bindgen(js_name = "...")]` (Rust ident ≠ JS class
name) and/or placed in a `js_namespace` emits a **duplicate phantom class** in
the generated JS glue when the struct crosses the boundary as a `JsValue` (e.g.
via `.into()`). The phantom class is keyed by the Rust ident, has a broken
`free()` referencing a nonexistent wasm export, and is what `__wrap` returns —
so instances are the wrong class and trap at GC / fail `instanceof`.

This is a regression from the qualified-name keying refactor in #5154.

## Root cause

After #5154, `cli-support` keys `exported_classes` by `qualified_name`. But the
wrap/unwrap aux imports in `crates/cli-support/src/wit/mod.rs` still passed
`rust_name`:

```rust
let wrap_constructor = wasm_bindgen_shared::new_function(&qualified_name);
self.add_aux_import_to_import_map(
    ..,
    AuxImport::WrapInExportedClass(rust_name.to_string()),   // ← should be qualified_name
)?;
let unwrap_fn = wasm_bindgen_shared::unwrap_function(&qualified_name);
self.add_aux_import_to_import_map(
    ..,
    AuxImport::UnwrapExportedClass(rust_name.to_string()),   // ← should be qualified_name
)?;
```

The import *function name* is correctly derived from `qualified_name`, but the
`WrapInExportedClass` / `UnwrapExportedClass` payload carried `rust_name`.
Downstream, `require_class_wrap(rust_name)` / `require_class_unwrap(rust_name)`
look up by that value; since `exported_classes` is keyed by `qualified_name`,
the lookup misses and a fresh empty entry keyed by the Rust ident is minted —
emitted as a second, phantom class.

The sibling `RustFromI32` path (a struct returned by value) already uses
`qualified_name` via `Descriptor::RustStruct`, which is why returning the
struct directly does **not** duplicate — only the `JsValue` conversion path
does.

## Commits

1. **`test:`** — two reference fixtures, blessed on the current (buggy) codegen
   so the fix commit's diff shows exactly what changes:
   - `exported-class-rename-wrap` — minimal repro (renamed struct, no
     namespace, returned as `JsValue`).
   - `exported-class-wrap-matrix` — the rest of the rename × namespace ×
     inheritance matrix on the `JsValue` wrap/unwrap path: namespace-only, and
     inheritance crossing `JsValue` (base renamed/child not, child renamed/base
     not, namespaced base/plain child, parent and child in different
     namespaces).
2. **`fix:`** — the two `rust_name` → `qualified_name` changes + re-blessed
   fixtures.
3. **`refactor:`** — removes the now-confirmed-dead `Context::qualified_to_rust_name`
   map (see Notes).

### `exported-class-rename-wrap.bg.js` before the fix

```js
// The real, intended class — correct free symbol.
export class Renamed {
    free() { ... wasm.__wbg_renamed_free(ptr, 0); }
    constructor(value) { const ret = wasm.renamed_new(value); ... }
    get value() { /* ... */ }
}

// PHANTOM duplicate, keyed by the Rust ident `RustRenamed`.
export class RustRenamed {
    static __wrap(ptr) { ... RustRenamedFinalization.register(obj, ...); return obj; }
    free() { ... wasm.__wbg_rustrenamed_free(ptr, 0); }   // ← no such wasm export
}

// Import named from qualified_name ("renamed"), but body wraps with the
// Rust-ident class — the mismatch this PR fixes.
export function __wbg_renamed_new(arg0) {
    const ret = RustRenamed.__wrap(arg0);   // ← should be Renamed.__wrap(arg0)
    return ret;
}
```

After the fix the phantom is gone, `Renamed` gains `static __wrap`, and
`__wbg_renamed_new` returns `Renamed.__wrap(arg0)`.

## Coverage

The matrix fixture verifies every cell emits a single, correctly-named class
with no Rust-ident leakage — inheritance chains render as `Dog extends Animal`,
`Car extends Vehicle`, `Reserve extends wild__Habitat`,
`garden__Derived extends zoo__Base`, etc. The existing
`js-namespace-export-same-name` fixture (added by #5154) re-blesses to drop the
`FooPoint`/`BarPoint` phantoms and route **both** the wrap and unwrap imports
through `foo__Point`/`bar__Point` — exercising both fixed lines.

Only JS-glue snapshots change; `.wat` is untouched (the wasm is identical — a
pure glue-generation fix). Fixtures were blessed with the current stable
toolchain to match CI's `dtolnay/rust-toolchain@stable`. Verified against a
downstream crate that renames ~50 exported structs and converts them to
`JsValue` pervasively: the duplicate classes and the `is not a function` GC
traps disappear.

## Notes

- No `js_namespace` is required to trigger this — plain `js_name` rename +
  `.into()` is enough; namespace-only (no rename) triggers it too.
- The `refactor:` commit removes `Context::qualified_to_rust_name`: it was
  never populated (no `.insert`), so its single `.get` in `resolve_class_name`
  always returned `None`. `resolve_class_name` is now a direct passthrough;
  behavior is unchanged (all reference snapshots byte-identical, no re-bless).
- `AuxStruct.rust_name` is likewise unused in `cli-support` after this fix, but
  **left untouched** — it already carries `#[allow(dead_code)]` upstream with a
  rationale (retained for diagnostics / backward compatibility), and
  `rust_name` remains part of the wire schema in `wasm-bindgen-shared`. Happy
  to drop it too if you'd prefer.


ZhaoXiangXML:
@guybedford this is ready for review — CI is fully green. It's a follow-up regression fix to #5154: after the qualified-name keying refactor, the wrap/unwrap aux imports were still keyed by `rust_name`, which minted a duplicate phantom class on the `JsValue` (`.into()`) path for renamed/namespaced structs. The fix is the two `rust_name` → `qualified_name` changes, with reference fixtures blessed on the buggy codegen first so the fix commit's diff shows exactly what changes. Would appreciate your review when you have a chance.

ZhaoXiangXML:
## Self-review — PR #5160 (round 1)

**Overall:** Mergeable, no blocking issues found. The fix is correct and verified against the generated output; the PR is gated only on a required review approval, nothing actionable.

**Findings:** 🔴 0 / 🟡 0 / 🟢 0

### Verified

- ✅ **Existing discussion** — no prior reviews or inline threads on this PR, so nothing to reconcile against.
- ✅ **Conflict check** — merges cleanly into `main`; the `BLOCKED` merge state is purely the required-approval gate, not a conflict or a failing check.
- ✅ **Base correctness** — confirmed #5154 is merged to `main` (`bbcdb9c`) and that `main` now keys `exported_classes` by `qualified_name` (`.entry(s.qualified_name.clone())`), so keying the wrap/unwrap imports by `qualified_name` matches the real key.
- ✅ **Core fix** — the two `AuxImport::{WrapInExportedClass,UnwrapExportedClass}` payloads now carry `qualified_name`; the import function names (`new_function`/`unwrap_function`) already derived from `qualified_name`, so only the payload was out of step.
- ✅ **Fix verified on the minimal repro** — `exported-class-rename-wrap` emits a single `class Renamed`; `__wbg_renamed_new` calls `Renamed.__wrap(...)` and `free()` calls the real `wasm.__wbg_renamed_free`. No phantom `class RustRenamed` and no `__wbg_rustrenamed_free` anywhere.
- ✅ **Fix verified across the matrix** — `exported-class-wrap-matrix` emits 9 classes, 9 `__wrap` wrap imports, and 9 `_free` imports, each referencing a real declared class (`Animal`/`Car`, not `RustAnimal`/`RustCar`) across the rename × namespace × inheritance combinations. No Rust-ident phantom classes remain.
- ✅ **Dead-code removal is behavior-preserving** — on current `main` (post-#5154) `qualified_to_rust_name` is declared and read but never inserted into, so its lookup branch was already dead; removing it changes no output (consistent with the commit's "all snapshots byte-identical, no re-bless"). `resolve_class_name` is kept as a passthrough and still has its two callers in `js/binding.rs`, so it is not itself dead.
- ✅ **CHANGELOG** — new `### Fixed` entry under `## Unreleased`, in correct keepachangelog order, linking #5160 and accurately attributing the regression to #5154.
- ✅ **Commit hygiene** — the `test:` → `fix:` → `refactor:` split is intentional and helpful: the test commit blesses the buggy codegen first so the fix commit's diff shows exactly what changes. Type prefixes and descriptions all match their diffs.
- ✅ **CI** — all checks green.

### Worth a nod

- The bless-buggy-then-fix commit structure makes the fix's effect self-documenting in the diff — easy to review.
- The matrix fixture is thorough: it pins not just rename and namespace but the inheritance-crossing-`JsValue` combinations, which is exactly where a key mismatch would have been easy to miss.

### Note (no action needed)

`resolve_class_name` is now a pure passthrough. The commit message already justifies keeping it as a single named seam in case key normalization is needed again — that rationale is reasonable, so I'm not raising it as a finding; flagging only so a future reader knows it's deliberate.

No changes requested. Ready for maintainer review.

---
Generated by Claude Opus 4.7

ZhaoXiangXML:
## 🟡 Coverage gap — the unwrap direction (and inheritance × unwrap) isn't exercised by the new fixtures

Following up on my self-review conclusion above (this amends its count to 🔴 0 / 🟡 1 / 🟢 0). To be clear up front: **the fix itself is correct** — keying `UnwrapExportedClass` by `qualified_name` makes `__unwrap` check `instanceof <qualified_class>`, and the old `rust_name` keying would have generated a phantom-class check that always returned `0` (silent unwrap failure). So fixing the unwrap line was necessary. This is purely about test coverage.

### What's missing

Both new fixtures generate **zero `__unwrap`**:

- `exported-class-rename-wrap.bg.js` — 0 occurrences of `__unwrap`.
- `exported-class-wrap-matrix.bg.js` — 0 occurrences of `__unwrap`.

They only exercise the **wrap** direction (`make*()` → `JsValue` → `Class.__wrap(...)`) plus the by-value argument path (`_assertClass(d, Dog)`). The `read_*` consume-functions go through `_assertClass`, not the `UnwrapExportedClass` path the fix touches.

The **only** `UnwrapExportedClass` coverage in the PR is the incidental re-bless of the pre-existing `js-namespace-export-same-name.bg.js` (`foo__Point.__unwrap` / `bar__Point.__unwrap`) — and none of that fixture's structs use `extends`.

So the two dimensions never intersect:

| | wrap (`__wrap`) | unwrap (`__unwrap`) |
|---|---|---|
| **with inheritance** (`Dog extends Animal`, …) | ✅ tested (matrix) | ❌ not tested |
| **no inheritance** | ✅ tested | ✅ tested (same-name) |

### Why it matters

1. The `test:` commit message says the matrix covers "the JsValue wrap/**unwrap** path." For the new fixtures that's overstated — only wrap is covered. If the `UnwrapExportedClass` line of the fix regressed, neither new fixture would catch it; only `js-namespace-export-same-name` would.

2. The matrix's stated purpose is "inheritance crossing `JsValue`," but those cases only cross via **wrap**. There's no fixture where a **subclass instance is unwrapped through a base-typed unwrap**. That's a genuinely interesting case to pin down: `Subclass instanceof Base` is `true`, so `Base.__unwrap` would hand the subclass's wasm pointer back to Rust as a base pointer — and exported-struct `extends` is a JS-prototype-chain convenience, not a guarantee of wasm-side layout compatibility. A fixture here would document whether that is intended/correct or a latent type-confusion.

### Suggestion (optional, non-blocking)

Add to `exported-class-wrap-matrix.rs` a function that actually drives the unwrap path on an inherited type — e.g. a method/free-fn that takes a renamed and/or namespaced exported struct **by value as a `JsValue`-converted argument**, for both a base and a subclass — so the matrix's "inheritance crossing JsValue" claim holds in both directions and the unwrap half of the fix is regression-locked by the dedicated fixtures rather than only by the same-name fixture.

---
Generated by Claude Opus 4.7

ZhaoXiangXML:
Addressed the unwrap coverage gap from the comment above in two test commits (no production-code change):

- **`7a17c4d`** — `exported-class-wrap-matrix`: added one `Vec<Struct>` consumer (`read_*s`) per struct, and corrected the header comment.
- **`f41207749`** — `exported-class-rename-wrap`: added a `Vec<RustRenamed>` consumer for the minimal renamed standalone case.

Root cause of the original gap: `__unwrap` is **not** driven by a single by-value struct argument — that lowers through `_assertClass` + the flat pointer ABI. It's driven by a **`Vec<Struct>` argument**, which is marshaled as a JS array of class instances and unwrapped element-wise via `Class.__unwrap`. The fixtures had no `Vec<Struct>` args, so they emitted zero `__unwrap`.

The unwrap import is now exercised across every keying combination, all resolving to the qualified JS identity (no phantom Rust-ident class):

- standalone renamed — `Renamed.__unwrap` (`instanceof Renamed`)
- standalone namespaced — `nsa__Widget.__unwrap` (`instanceof nsa__Widget`)
- inherited base/child across rename × namespace combos — `Animal`/`Dog`/`Vehicle`/`Car`/`wild__Habitat`/`Reserve`/`zoo__Base`/`garden__Derived`

The blessed output also documents the answer to the open question I raised — *does a base's `__unwrap` accept a subclass instance?* For an **inheriting** class, `__unwrap` checks pointer identity (`__wbg_ptr === __wbg_ptr_<Class>`), not `instanceof`, so `Animal.__unwrap(aDog)` returns `0`: it **rejects** the subclass rather than handing a subclass pointer to wasm as a base. A **standalone** class (no subclass) uses the simpler `instanceof` check. Both are correct, and now both are pinned by a reference snapshot.

---
Generated by Claude Opus 4.7

guybedford:
Picked this up for review and pushed two additions plus a rebase onto current `main` (was behind #5158, which made the diff include unrelated interpreter reverts).

**Added**

- `test: rename-collision fixture exercising distinct js_names across modules` — fills a gap surfaced during review: the no-namespace half of the same-Rust-ident multi-module pattern. Pre-#5154 this case hard-errored with `found duplicate constructor for class FooPoint`; post-#5154/post-this-PR both classes emit cleanly. The namespace half is already covered by `js-namespace-export-same-name`.
- `refactor: drop redundant struct keying fields from schema` — riding the wire-format bump #5154 is already forcing on the next release (the new `extends_js_class`/`extends_js_namespace` fields on `Struct`), drops three pieces of dead/redundant state:
  - `Struct.rust_name` (wire schema) + `AuxStruct.rust_name`. Used to be the `exported_classes` key; replaced by `qualified_name` in #5154, no remaining reader in cli-support, not surfaced in any diagnostic. The `#[allow(dead_code)]` and "retained for diagnostics" comment were aspirational — no other Aux type carries a rust_name field, so it was always specifically the keying handle.
  - `AuxStruct.qualified_name` / `AuxEnum.qualified_name`. Derivable on demand from `(name, js_namespace)` via the existing `wasm_bindgen_shared::qualified_name` helper.
  - `ExportedClass.qualified_name`. Duplicates the `exported_classes` map key. Removing it collapses three defensive `qualified_name.or_else(js_name).unwrap_or(key)` fallback chains that only existed to paper over the pre-#5154 keying disagreement.
  - The `ancestors` tuple shrinks from `(identifier, rust_name, qualified)` to `(identifier, qualified)`. The rust_name slot was for the upcast-symbol parent half, which #5154 unified onto `qualified_name` on both sides via `extends_js_class`/`extends_js_namespace`.

  Net: 98 insertions / 116 deletions. Schema hash re-approved.

**Why fold this in rather than as a follow-up**

#5154 is a breaking schema bump regardless (positional decoder + two new fields). Dropping `rust_name` in the same cycle is free; deferring means carrying a dead wire field across a release boundary and paying a second schema break later to remove it.

**Verification**

- All 84 reference tests pass locally.
- Schema hash approval test passes (hash bumped from `10237465989497079908` → `14631892944928291272`).
- `cargo clippy --workspace` clean.
- The fix commits themselves (`e08a6d71`, `924e190f`) are unchanged — the surgical two-line `rust_name → qualified_name` change is intact.

CHANGELOG entry folded into the existing #5160 bullet so the schema cleanup ships under the same fix narrative.

---
Reviewed by Claude Opus 4.7

ZhaoXiangXML:
Replying to https://github.com/wasm-bindgen/wasm-bindgen/pull/5160#issuecomment-4501655926

I tested this downstream with a locally built `--release` `wasm-bindgen-cli`, and it works for me.

guybedford (approved):

PR #5163, opened 2026-05-22
fix: use js_sys::futures instead of wasm_bindgen_futures in macro

guybedford:
### Description
We're starting to give the guidance that wasm-bindgen-futures is an optional dependency since we consolidated into `js_sys::futures`.

The problem is that the wasm-bindgen macro is still emitting for `wasm-bindgen-futures` in its output.

This makes `js_sys::futures` the default, unless an explicit wasm-bindgen-futures attribute is provided.

### Checklist


guybedford:
Forgot we created `cfg=wasm_bindgen_use_js_sys` exactly for this roll-out. We can mitigate this through wasm-pack instead.

PR #5168, opened 2026-05-26
Polyfill WebAssembly.JSTag for legacy EH on Node 20+

### LINKED ISSUE #5151: Supporting panic=unwind without hard aborts

guybedford:
This implements a JS-side polyfill for `WebAssembly.JSTag` when wasm-bindgen emits legacy EH (`try`/`catch`) catch handlers under `panic=unwind`. Resolves #5151.

The generated JS shim now binds `__wbindgen_jstag` to `new WebAssembly.Tag({ parameters: ['externref'] })` and wraps each JS import in a try/catch that re-throws caught exceptions through that polyfilled tag, so legacy-EH builds run on Node.js 20 without needing the host's `WebAssembly.JSTag` (Node 22.22.3+). The Wasm side is unchanged.

Empirically the legacy-EH ↔ no-JSTag correlation is perfect across real engines: anything new enough to ship `try_table` also ships `WebAssembly.JSTag`. So the polyfill is gated purely on the detected `ExceptionHandlingVersion::Legacy` — no user-facing flag.

The issue originally framed this as supporting panic=unwind *without hard abort handling*. That separation turned out to be unnecessary: the hard-abort machinery (`__instance_terminated`, `__abort_handler`, `__wbg_call_guard`, `__wbg_handle_catch`, the wrapped tag) only uses `WebAssembly.Tag`/`WebAssembly.Exception` (Node 18+), so all of it keeps working under the polyfill.

* Auto-gates the polyfill on `ExceptionHandlingVersion::Legacy` in `catch_handler::run` via a new `aux.jstag_polyfill` flag
* `generate_jstag_import` emits `new WebAssembly.Tag(...)` instead of `WebAssembly.JSTag` when the flag is set
* New `expose_jstag_wrap` helper, applied to every import alongside the existing `handleError`/`logError` wrapping
* Already-tagged `WebAssembly.Exception` instances (e.g. from `__wbindgen_throw`/`__wbindgen_rethrow` via `__wbindgen_wrapped_jstag`) pass through untouched, preserving the recoverable-vs-abort discrimination in `__wbg_handle_catch`

Before, generated shim for legacy EH:

```js
export const __wbindgen_jstag = WebAssembly.JSTag;
export function __wbg_foo(arg0) { /* body */ }
```

After:

```js
const __wbindgen_jstag = new WebAssembly.Tag({ parameters: ['externref'] });
function __wbg_jstag_wrap(f, args) {
    try { return f.apply(this, args); }
    catch (e) {
        if (e instanceof WebAssembly.Exception) throw e;
        throw new WebAssembly.Exception(__wbindgen_jstag, [e], { traceStack: true });
    }
}
export function __wbg_foo() { return __wbg_jstag_wrap(function(arg0) { /* body */ }, arguments); }
```

Test coverage:

* New `runtest_panic_unwind_legacy` reference test that builds with `-Cpanic=unwind -Cllvm-args=-wasm-use-legacy-eh` and snapshots the polyfilled shim shape.
* The existing `runtest_panic_unwind` (modern EH) still asserts native `WebAssembly.JSTag` import — no regression on the Node 22+/modern path.
* CI: `test_wasm_bindgen_unwind_legacy_eh` drops from Node 22.22.3 to Node 20 (Node.js 18 has segfaults...) to exercise the polyfill end-to-end.

Verified manually on Node 20/22 with a standalone WAT module that imports a custom Tag, throws through it from JS, and catches via legacy `catch $jstag` — confirming the round-trip works on all three and that bare JS throws (without the wrap) are NOT caught, which is exactly the engine semantic the polyfill restores via the per-import JS wrapper.

Guide and CI comment updated to frame legacy EH as the path for Node 18+ support.


guybedford:
## Open issue: `unwind::` test suite trap on legacy EH

While verifying the polyfill locally I hit a real regression in the legacy-EH test suite that I haven't been able to root-cause yet. Posting full repro details in case a reviewer with deeper EH context can spot it faster than I can.

**Symptom.** Running the unwind tests under `-Cpanic=unwind -Cllvm-args=-wasm-use-legacy-eh` reproduces a `RuntimeError: unreachable` originating from a Wasm catch wrapper's termination guard. The four `unwind::*` tests all report `ok`, then an async task crashes during teardown:

```
test unwind::try_promise_all ... ok
test unwind::try_every ... ok
test unwind::ref_unwind_safe_method_runtime_behavior ... ok
test unwind::macro_injected_maybe_unwind_safe_bound ... ok

RuntimeError: unreachable
    at wasm.<…::ConsoleTask>::run::__wbg_run_322bf7e1760605e9 catch wrapper (wasm-function[32103]:0x5d7533)
    at wasm.<…::ConsoleTask>::run::__wbg_run_322bf7e1760605e9 externref shim
    at wasm.<…::Task>::run
    at wasm.<…::QueueState>::run_all
    at wasm.<…::Queue>::new::{closure#0}
    at wasm…convert::closures::_::invoke<…>::{closure#0}
    at wasm.AssertUnwindSafe…::call_once
    at wasm.std::panicking::catch_unwind::do_call::<…>
    at wasm.__rust_try
```

The trap is the `unreachable` from `emit_termination_guard` inside the catch wrapper for `ConsoleTask::run` (`crates/cli-support/src/transforms/catch_handler.rs:476-491`). That means `__instance_terminated` has been set, the run loop has already moved past the panicking test's catch boundary, and a queued async task tied to that earlier test is being polled afterwards.

**Reproducer:**

```bash
# Node 20 or 22; we have verified the failure is the same on both.
just test-wasm-bindgen-unwind-legacy-eh --test wasm 'unwind::' --features std,wasm-bindgen-futures/std
```

`just test-wasm-bindgen-unwind-legacy-eh` already exists in the repo and sets `RUSTFLAGS="-Cpanic=unwind -Cllvm-args=-wasm-use-legacy-eh"` + `-Zbuild-std=std,panic_unwind` + `WASM_BINDGEN_SPLIT_LINKED_MODULES=1`. The narrowest filter the wasm-bindgen-test runner accepts is `unwind::` (it supports a single filter only). Running any of the four `unwind::*` tests in isolation (e.g. `just test-wasm-bindgen-unwind-legacy-eh --test wasm unwind::try_promise_all …`) passes — only the combined run trips it.

**What I have ruled out so far:**

* Not Node-version-specific: identical failure on Node 20.20.2 and Node 22.22.3.
* Not the polyfill mechanics itself: the standalone JSTag-polyfill round-trip works perfectly (verified with a 30-line WAT module across Node 18/20/22 before posting the PR).
* Not the reference snapshot: the new `panic-unwind-legacy.bg.js` snapshot matches the intended shape exactly and `runtest_panic_unwind` (modern EH path) still passes unchanged.
* Pre-existing on the modern-EH path: on `main` with native `WebAssembly.JSTag` and Node 22, the same suite passes. So the regression is the polyfill specifically.

**My current theory.** With native JSTag, when a Wasm closure panics inside a JS-called callback, the engine wraps the wasm-thrown `Exception(__wbindgen_wrapped_jstag, …)` and JS-side error pathways propagate it cleanly. With the polyfill, `__wbg_jstag_wrap` catches it, sees it's already a `WebAssembly.Exception`, and rethrows unchanged — but somewhere along the way the exception reaches an export wrapper's `__wbg_handle_catch` as the *non*-wrapped tag and the termination flag gets set. The catch wrapper's first termination guard then traps when `ConsoleTask::run` returns normally on the next poll.

I haven't been able to confirm exactly which path mis-tags the exception. Possible angles:

1. Re-throwing a `WebAssembly.Exception` from JS may not preserve tag identity in the way native JSTag throws do (e.g. catch-and-rethrow might re-tag through the polyfilled tag in some V8 path).
2. `__wbg_jstag_wrap` may need to *also* differentiate between exceptions carrying `__wbindgen_jstag` vs. `__wbindgen_wrapped_jstag` more carefully than the current `instanceof WebAssembly.Exception` check.
3. There may be an additional code path (e.g. the futures task queue polling) that calls back into Wasm without going through a catch-wrapped import, missing the tag-translation step.

Notably the issue ([#5151](https://github.com/wasm-bindgen/wasm-bindgen/issues/5151)) framed this work as supporting `panic=unwind` *without hard aborts* precisely because the hard-abort discrimination relies on JSTag identity. The investigation in this PR found we could keep hard aborts, but this regression suggests there may be a subtle case where the discrimination breaks under the polyfill. Sharing in case someone with deeper EH context (cc folks who worked on #5128 / the hard-abort plumbing) sees it immediately.

For now the PR is otherwise complete: reference snapshot test passes, modern-EH path is untouched, JS-side shape of the polyfill is correct, manual end-to-end smoke test of an isolated panic case works on Node 20.

crates/cli-support/src/js/mod.rs
```diff
@@ -5168,6 +5210,21 @@ addToLibrary({
                     code
                 };
 
+                // JSTag polyfill: when running on engines without
+                // `WebAssembly.JSTag` (e.g. Node 20), JS exceptions thrown
+                // out of imports never tag themselves with the JSTag, so the
+                // Wasm catch wrappers' `catch $jstag` arms never trigger.
+                // Wrap each import in a JS try/catch that re-throws via a
```
```suggestion
                // JSTag polyfill: when running on Node 20, wrap each import in a JS try/catch that re-throws via a
```

crates/cli-support/src/js/mod.rs
```diff
@@ -5168,6 +5210,21 @@ addToLibrary({
...
+                // `new WebAssembly.Exception(__wbindgen_jstag, [e])` against
+                // the polyfilled tag, restoring the engine semantics that
+                // JSTag would otherwise provide.
```
```suggestion
                // the polyfilled tag
```

crates/cli-support/src/transforms/catch_handler.rs
```diff
@@ -132,6 +132,12 @@ pub fn run(
     log::debug!("Catch handler created {} wrappers", wrappers.len());
     aux.js_tag = Some(js_tag);
     aux.wrapped_js_tag = wrapped_js_tag;
+    // Legacy EH support gap: `WebAssembly.JSTag` only exists on Node 22.22.3+,
+    // but the underlying `try`/`catch` instructions work on Node 18/20. When
+    // we're emitting legacy EH wrappers we polyfill JSTag on the JS side with
+    // a custom `WebAssembly.Tag` plus a JS-side try/catch around every import,
+    // so the resulting module runs on engines without `WebAssembly.JSTag`.
```
hoodmane:
```suggestion
```

crates/cli-support/src/wit/nonstandard.rs
```diff
@@ -81,6 +81,14 @@ pub struct WasmBindgenAux {
     /// distinguished from other JS exceptions.
     pub wrapped_js_tag: Option<walrus::TagId>,
 
+    /// When true, `WebAssembly.JSTag` is polyfilled on the JS side with a
+    /// `new WebAssembly.Tag({ parameters: ['externref'] })`, and every import
+    /// is wrapped in a JS try/catch that re-throws caught exceptions through
+    /// that polyfilled tag. Set automatically when legacy EH (`try`/`catch`)
+    /// is detected so that toolchains targeting Node 20/older engines (no
+    /// `WebAssembly.JSTag`) keep working with `panic=unwind`.
```
hoodmane:
```suggestion
    /// that polyfilled tag. Set automatically when legacy EH
    /// is detected. For use when Node 20 is desired.
```

crates/cli/tests/reference/panic-unwind-legacy.bg.js
```diff
@@ -0,0 +1,131 @@
...
+        __wbg_handle_catch(e);
+    }
+    return ret >>> 0;
+}
+export function __wbg___wbindgen_panic_error_c684b5bb231f71fe() { return wrapError(function (arg0) {
+    const ret = new PanicError(arg0);
+    return ret;
+}, arguments); }
+export function __wbg___wbindgen_rethrow_c4d99b4b53265290() { return wrapError(function (arg0) {
+    throw new WebAssembly.Exception(__wbindgen_wrapped_jstag, [arg0]);
```
hoodmane, line 20: Won't this potentially double wrap errors? I guess the tests pass so probably not...

hoodmane: Oh right it wraps the jstag polyfill first and then the wrapped_jstag.

hoodmane (approved):

PR #5175, opened 2026-06-01
Fix unresolved __wbindgen_placeholder__ imports on WASI targets

codeitlikemiley:
The compilation failure on version `0.2.115` is caused by a change in how `wasm-bindgen` gates its target-specific code generation. 

Specifically, target detection was simplified from a strict arch/OS check to a broad `target_family = "wasm"` check. Because **WASI targets belong to the `wasm` target family**, this forces WASI builds down the browser/JS codegen path, emitting unresolved JavaScript imports.

Here are the exact technical details:

### 1. The Gate in Version `0.2.114` (Working)
In `0.2.114`, the procedural macro backend `wasm-bindgen-macro-support` and the core library `wasm-bindgen` restricted Web-specific code emission to targets where the target OS is `unknown` or `none`:

```rust
#[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
```

* **For `wasm32-unknown-unknown`:** Evaluates to `true` -> emits the actual JavaScript bindings and placeholders.
* **For `wasm32-wasip2` / `wasm32-wasi`:** Evaluates to `false` (since `target_os = "wasi"`) -> compiles the target-independent fallback stubs, which are just Rust functions that panic at runtime if called, and **do not** generate external imports.

### 2. The Change in Version `0.2.115` (Breaking)
In `0.2.115`, these guards were simplified to:

```rust
#[cfg(target_family = "wasm")]
```

* **For `wasm32-wasip2`:** Evaluates to `true` because the target family is `wasm`.
* Consequently, `wasm-bindgen` evaluates the build as a browser target and emits the real JS imports linked to the `__wbindgen_placeholder__` module rather than the dummy stubs:

```rust
#[cfg(target_family = "wasm")]
extern "C" {
    fn __wbindgen_object_drop_ref(ptr: u32) -> u32;
}
```

### 3. Why This Fails at the Component Linker
Because `__wbindgen_object_drop_ref` is declared inside a `#[link(wasm_import_module = "__wbindgen_placeholder__")]` block, the Rust compiler leaves `__wbindgen_placeholder__::__wbindgen_object_drop_ref` as an unresolved import in the compiled `.wasm` file.

When translating the core WebAssembly module into a WASIp2 Component, `wasm-component-ld` (or `wit-component`) looks for a WIT interface to resolve the `__wbindgen_placeholder__` import. Since no such WIT interface exists for browser bindings in a WASI execution environment, the linker fails:

```
failed to resolve import __wbindgen_placeholder__::__wbindgen_object_drop_ref
```

### References:
- https://github.com/leptos-rs/leptos/pull/4750
- https://github.com/leptos-rs/leptos_wasi/actions/runs/26672566178/job/78618402860


codeitlikemiley:
@gbj  here is the PR and fix on the issue on https://github.com/leptos-rs/leptos/pull/4750

codeitlikemiley:
@guybedford hope that you would  take a look into this , any reason why we have regression after 0.2.114  ?
This breaks apps that uses leptos and leptos_wasi  

Would love to hear your feedback on this thanks

gbj:
Well, now I remember why this problem sounded familiar :-) 

A little over 3 years ago I made PR #3233, which was merged, and which disabled emitting wasm-bindgen placeholders for WASI, in the same way that they were disabled (at the time) for emscripten. This allowed code that included wasm-bindgen placeholders (including leptos) to be compiled for WASI and run. (The wasm-bindgen functions are never called on the server; if compiled for WASI they could panic just as they would in native code.)

I have not followed the state of things since then, but it looks by the time of #4443 this flag had been adjusted from my `#[cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))]` to `#[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]`. This would still exclude emscripten and WASI, correctly.

For the sake of *enabling* emscripten, #4443 then changed this flag to `#[cfg(target_family = "wasm")]`. But this also turned this codegen on (incorrectly, I think) for WASI.

guybedford:
Thanks for the very clear analysis on this one. To avoid future regressions I've pushed up a test here, and also unified the gates a bit.

guybedford (approved):

PR #5177, opened 2026-06-01
Make UpcastFrom unsafe

### LINKED ISSUE #5176: UpcastFrom is unsound
UpcastFrom should be unsafe because it lets you transmute types. Without the unsafe bound, you can cause segfaults in safe code:

```rust
// [package]
// name = "upcast-soundness-mre"
// version = "0.0.0"
// edition = "2021"
// publish = false
// 
// [dependencies]
// wasm-bindgen = "0.2.122"

use std::hint::black_box;
use wasm_bindgen::convert::{Upcast, UpcastFrom};
use wasm_bindgen::prelude::*;
use wasm_bindgen::JsCast;

#[wasm_bindgen]
extern "C" {
    type Foo;
    #[derive(Clone)]
    type Bar;
}

// The bug, in safe code: a `& &Foo` of *any* lifetime `'a` is reinterpretable as a
// `&'static &'static Bar`. The single free `'a` binds the short-lived source borrow;
// the target is hardcoded `'static` — that is the laundering.
impl<'a> UpcastFrom<&'a &'a Foo> for &'static &'static Bar {}

fn main() {
    const N: usize = 1024;

    let dangling: &'static &'static Bar = {
        let foo: Foo = JsValue::UNDEFINED.unchecked_into();
        let slots: Vec<&Foo> = vec![&foo; N];
        (&slots[N - 1]).upcast_into() // launder `& &foo` to `&'static &'static Bar`
    }; // `slots` freed; `dangling` points into freed memory

    // Reuse that region with zeros, so the slot `dangling` reads is a NULL `&Bar`.
    // `black_box` keeps the allocation (and its zeroing) from being optimized away.
    let zeros = vec![0usize; N];
    black_box(&zeros);

    // `*dangling` = NULL `&Bar`; the second deref reads through address 0 -> SIGSEGV.
    let _ = (**black_box(dangling)).clone();
}
```

ealmloff:
### Description

Marks UpcastFrom as unsafe since that allows you to transmute in safe code. Without the unsafe keyword it is possible to transmute lifetimes incorrectly with safe code using the unstable convert module

Closes #5176 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 ### Changed
 
+* Made `wasm_bindgen::convert::UpcastFrom` an unsafe trait, requiring manual
+  implementations to use `unsafe impl`.
+  [#5177](https://github.com/wasm-bindgen/wasm-bindgen/pull/5177)
+
 * `JsOption<T>` now treats only `undefined` as empty, aligning it with
   TypeScript's strict `T | undefined` semantics and with `Option<T>`'s wire
   shape (`None` ↔ `undefined`). Previously `is_empty`, `as_option`,
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1807,18 +1807,18 @@ macro_rules! impl_tuple_covariance {
         // ArrayTuple -> Array
         // Allows (T1, T2, ...) to be used where (Target) is expected
         // when all T1, T2, ... are covariant to Target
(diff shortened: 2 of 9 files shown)


ealmloff:
requiring unsafe to enable the cast from `Closure<dyn Fn(i16)>` -> `Closure<dyn Fn(i8)>` seems wrong here. I don't think this is the right solution

PR #5178, opened 2026-06-01
Fix UpcastFrom soundness via lifetime-faithful ErasableGeneric repr

### LINKED ISSUE #5176: UpcastFrom is unsound

guybedford:
This makes the generic upcasting machinery sound. Resolves #5176.

`Upcast::upcast`/`upcast_into` perform a transmute that is gated on `T: ErasableGeneric<Repr = Self::Repr>`. The intent is that this repr-equality bound proves the transmute is layout- and lifetime-compatible. But the reference impls erased the borrow lifetime to `'static`:

```rs
unsafe impl<T: ErasableGeneric> ErasableGeneric for &T {
    type Repr = &'static T::Repr;
}
```

so for any source borrow `&'a T` the repr was `&'static …`, and the gate could no longer see the lifetime. That let a safe `UpcastFrom` impl launder a short-lived borrow into a `'static` one — segfaulting in entirely safe code:

```rs
impl<'a> UpcastFrom<&'a &'a Foo> for &'static &'static Bar {}
// `& &foo` of any lifetime is now reinterpretable as `&'static &'static Bar`
```

The `'static` repr was deliberate: at the JS ABI boundary borrows are only ever call-scoped (JS-GC-rooted), so erasing them to `'static` is fine there. The problem is purely that the *same* repr was also trusted by `upcast`, where the value escapes back into Rust.

The fix splits those two safety models. `ErasableGeneric::Repr` now carries the real borrow lifetime, so the repr-equality gate rejects any upcast that would extend a lifetime while still accepting same-lifetime ones. The `'static` erasure the ABI boundary needs stays exactly where it was already sound — the explicit `unsafe` transmute codegen emits at the call site.

* `ErasableGeneric::Repr` drops the `'static` bound; `&T`/`&mut T`/`&[T]`/`&str`/closure reprs capture the actual borrow lifetime instead of `&'static`.
* `ErasableGenericBorrow`/`ErasableGenericBorrowMut` markers compare the borrowed repr at a common lifetime rather than against `&'static ConcreteTarget`.
* `UpcastFrom` stays a safe, user-implementable trait — the laundering impl still compiles, it just can no longer be *called*.

Before, this compiled; now it fails to compile with a lifetime error pointing at the repr-equality bound:

```rs
fn launder_into(foo: &Foo) -> &'static &'static Bar {
    let r: &&Foo = &foo;
    r.upcast_into() // error: `foo` is borrowed for `'static`
}
```

Test coverage: adds `upcast-unsound.rs` as a compile-fail UI test (the #5176 exploit), and updates `generic-import-non-erasable.stderr` for the now lifetime-faithful repr in diagnostics. The full `wasm-bindgen` and `js-sys` wasm suites pass, along with the existing upcast/generics/closure/slice tests, confirming legitimate same-lifetime upcasts are unaffected.

### DIFF
--- a/crates/macro/ui-tests/generic-import-non-erasable.stderr
+++ b/crates/macro/ui-tests/generic-import-non-erasable.stderr
@@ -33,8 +33,8 @@ error[E0271]: type mismatch resolving `<&u32 as ErasableGeneric>::Repr == &JsVal
 24 |     bar::<u32>(&42);
    |           ^^^ expected `&JsValue`, found `&u32`
    |
-   = note: expected reference `&'static wasm_bindgen::JsValue`
-              found reference `&'static u32`
+   = note: expected reference `&wasm_bindgen::JsValue`
+              found reference `&u32`
    = note: required for `u32` to implement `ErasableGenericBorrow<wasm_bindgen::JsValue>`
 note: required by a bound in `bar`
   --> ui-tests/generic-import-non-erasable.rs:3:1
@@ -52,8 +52,8 @@ error[E0271]: type mismatch resolving `<&mut u32 as ErasableGeneric>::Repr == &m
 29 |     baz::<u32>(&mut x);
    |           ^^^ expected `&mut JsValue`, found `&mut u32`
    |
-   = note: expected mutable reference `&'static mut wasm_bindgen::JsValue`
-              found mutable reference `&'static mut u32`
+   = note: expected mutable reference `&mut wasm_bindgen::JsValue`
+              found mutable reference `&mut u32`
    = note: required for `u32` to implement `ErasableGenericBorrowMut<wasm_bindgen::JsValue>`
 note: required by a bound in `baz`
   --> ui-tests/generic-import-non-erasable.rs:3:1
--- a/crates/macro/ui-tests/upcast-unsound.rs
+++ b/crates/macro/ui-tests/upcast-unsound.rs
@@ -1,0 +1,25 @@
+// Regression test for #5176: `UpcastFrom` must not let safe code launder a
+// short-lived borrow into a `'static` one via `upcast_into`/`upcast`.
+//
+// The malicious impl below is still allowed to exist, but calling the upcast
+// must fail to compile because the `ErasableGeneric` repr now carries the
+// borrow lifetime, so the source borrow would have to be `'static`.
+
+use wasm_bindgen::convert::{Upcast, UpcastFrom};
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
+    type Foo;
+    type Bar;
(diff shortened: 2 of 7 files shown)


ealmloff:
Thanks, this looks better. Assuming this lifetime is tied to the lifetime of the data in the closure, it should probably also not be `'static`:

https://github.com/wasm-bindgen/wasm-bindgen/blob/cc60a5104c8ce80128f8ab735834db9df10a8628/src/closure.rs#L997-L1000

After this change, `ErasableGeneric` only remaps structs in the form of `struct A<T> { phantom: PhantomData<T>, ...fields_that_dont_use_t }` from `A<T> -> A<U>` with transmute which seems safe as long as `T: 'static` (doesn't enforce any useful lifetime bound)

guybedford:
So the closure stronger bound actually breaks closure bindgen... but I also just happen to have a real generics PR in progress via https://github.com/wasm-bindgen/wasm-bindgen/pull/5180 that would side-step this entire problem space if I can land it, so will leave this in broken state for now pending the outcome of that work.

PR #5171, opened 2026-05-27
[Draft] Descriptors via custom section: delete the wasm interpreter

guybedford:
# Descriptors via custom section: delete the wasm interpreter

Replaces the long-standing wasm-bindgen-cli descriptor interpreter
with a compile-time custom section transport. The
`crates/cli-support/src/interpreter/` directory is deleted in this
PR.

## What changes

Every `#[wasm_bindgen]` shim's descriptor is now emitted at compile
time as bytes in the new `__wasm_bindgen_descriptors` custom wasm
section. `wasm-bindgen-cli` reads those bytes structurally without
executing any wasm.

Coverage is complete: exports, imports, `ImportStatic`, struct field
getters, dynamic unions, closure-arg wrappers, **and closure-cast
monomorphisations**. The legacy describe-by-execution mechanism —
which involved synthesising `__wbindgen_describe_<name>` exports for
the cli to interpret — is gone end-to-end.

## How the const-time schema works

Every `WasmDescribe` impl carries a pair of associated consts:

```rust
pub trait WasmDescribe {
    const SCHEMA_LEN: usize;
    const SCHEMA_BUF: [u32; SCHEMA_MAX]; // SCHEMA_MAX = 256
}
```

The fixed buffer size (1 KB per impl) sidesteps the
[`generic_const_exprs`](https://github.com/rust-lang/rust/issues/76560)
wall — we don't need generic-length array expressions because every
schema lives in the same-sized buffer with the meaningful prefix
indicated by `SCHEMA_LEN`. Wrapper types (`Option<T>`, `Vec<T>`,
`Result<T, E>`, `&T`, `&mut T`, closure trait objects) compose their
schemas at const time via `const fn`s that read the inner type's
`(SCHEMA_LEN, SCHEMA_BUF)` pair and produce a new pair:

```rust
impl<T: WasmDescribe> WasmDescribe for Option<T> {
    const SCHEMA_LEN: usize = 1 + T::SCHEMA_LEN;
    const SCHEMA_BUF: [u32; SCHEMA_MAX] =
        wrap_schema(&[OPTIONAL], &T::SCHEMA_BUF, T::SCHEMA_LEN);
}
```

For closure trait objects:

```rust
impl<A: FromWasmAbi, R: ReturnWasmAbi> WasmDescribe for dyn Fn(A) -> R + '_ {
    const SCHEMA_LEN: usize = 3 + A::SCHEMA_LEN + 2 * R::SCHEMA_LEN;
[...]

PR #5180, opened 2026-06-02
Generic #[wasm_bindgen] imports via per-monomorphisation

guybedford:
Stacked on #5171.

Implements generic `#[wasm_bindgen]` imports via per-monomorphisation: a generic imported function, method, constructor, getter, setter or static emits a dedicated binding for each concrete instantiation at the call site, marshalling the type parameter by its own ABI.

Generic imports accept the full range of ABI types — primitives, refs/lifetimes, `Option<T>`, `Vec<T>`, `Vec<Option<T>>` (as a `(T | undefined)[]` externref array), generic closures, `async` and `catch`. `&T` arguments also work for copyable value types, so `fn(&U)` imports like `Promise::resolve` accept `Promise<u32>`.

`JsGeneric` is the JS-handle identity bound (`ErasableGeneric<Repr = JsValue>` + upcast/cast); it carries no ABI supertraits, with explicit single-direction bounds at the hand-written js-sys sites that need them.

Draft — opening early for review of the approach.


guybedford:
Superseded by #5230.

PR #5181, opened 2026-06-03
Discover wasm64 main

nickbabcock:
### Description

Fixes the bug where main is always expected to have a `(i32, i32) -> i32` signature, which isn't true for wasm64 as argv should be a pointer width.

For intuition:

```c
int main(int argc, char **argv) {
}
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@
   helpers get instrumented.
   [#5179](https://github.com/wasm-bindgen/wasm-bindgen/pull/5179)
 
+* Fixed `main` silently never running on wasm64 for bin crates.
+
 ### Removed
 
 --------------------------------------------------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
   [#5179](https://github.com/wasm-bindgen/wasm-bindgen/pull/5179)
 
 * Fixed `main` silently never running on wasm64 for bin crates.
+  [#5181](https://github.com/wasm-bindgen/wasm-bindgen/pull/5181)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,8 @@
   helpers get instrumented.
   [#5179](https://github.com/wasm-bindgen/wasm-bindgen/pull/5179)
 
+* Fixed `main` silently never running on wasm64 for bin crates.
```
Spxg, line 48: We also need to paste the PR link here.

crates/cli-support/src/wit/mod.rs
```diff
@@ -2116,3 +2124,117 @@ fn normalize_memory64_descriptor_only_rewrites_explicit_number_abi_options() {
...
+
+/// Don't mistake a `main(i32, i32) -> i32` as the entrypoint on wasm64
+#[test]
+fn discover_main_ignores_wrong_argv_width_on_wasm64() {
+    let (mut module, mem) = make_main_module(true, walrus::ValType::I32);
+    assert!(
+        !run_discover_main(&mut module, mem),
+        "on wasm64 a 32-bit-argv `main` must not be treated as the entry point"
+    );
+}
```
Spxg, line 2242: Put them inside the 
```rust
#[cfg(test)]
mod tests {}
```

Spxg (commented):
Thanks! LGTM, just some nits.

crates/cli-support/src/wit/mod.rs
nickbabcock: Applied. I was keeping with the convention of existing tests like `test_struct_packer`, but based on your suggestion I moved them to be within the tests submodule too 😃

Spxg (approved):

PR #5183, opened 2026-06-04
fix: mutable reference variance

ealmloff:
### Description

The current `UpcastFrom` implementation for `&mut T` models covariance which lets you upcast a type to a more general type. That is correct for `&T`, but leads to incoherent typing with `&mut T`:

```rust
let mut string = JsString::from("valid");
let string_ref: &mut JsString = &mut string;
// Curretly &mut T is modeled as just covariant so you can upcast the type to JsValue
let js_value_ref: &mut JsValue = string_ref.upcast_into();
*js_value_ref = Object::new().into();
// I now hold a `&mut JsString` reference which is actually `&mut Object`
// Converting it to a Rust String throws because the value is not a string.
let _ = String::from(&string);
```

This PR switches the model to make it invariant which only lets you cast between completely equivalent types like casting from `&mut Closure<dyn Fn(Undefined)>` -> `&mut Closure<dyn Fn(())>`.

This is the general version https://github.com/wasm-bindgen/wasm-bindgen/pull/748 which removed upcasting from any specific value -> JsValue for similar reasons.

This fixes the type variance issue in #5176, it does not effect the soundness issue described in the original report

### Checklist

Note this is technically a breaking change. I see some breaking changes for apis where that breaking changed fixed a bug have been released in the past. If that is not the case, feel free to close

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,16 @@
 
 ### Fixed
 
+* Fixed incorrect variance in `&mut` reference upcasting. `&mut T` upcasts
+  were covariant in the pointee, so a `&mut T` could be widened to a `&mut`
+  of a supertype and used to write back a value the original type would not
+  accept, leaving a reference whose static type no longer matches the value
+  it points to. Mutable references are now *invariant* in their pointee:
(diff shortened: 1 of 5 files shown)


guybedford (approved):
Neat, thank you.

PR #5185, opened 2026-06-04
fix: remove Upcast from Array to ArrayTuple

### LINKED ISSUE #5184: Array should not Upcast to ArrayTuple<(T+)>
Currently you can upcast from `Array` into `ArrayTuple<(String)>` infallibly using `.upcast_into()`. `ArrayTuple<(String)>` is a subset of `Array` but not the other way around. This makes it easy to upcast into impossible states without `unchecked` methods:
```rust
let array: Array<Number> = Array::new_typed();
// upcast should only be able to cast concrete values into more
// general values where the conversion cannot fail
let _one: ArrayTuple<(Number,)> = array.upcast_into();
let first: Number = _one.get0();
// this throws because the number is not a number
assert!(first.value_of().is_finite());
```

ealmloff:
### Description

Removes upcasts from `Array<T>` -> `ArrayTuple<(T, *)>`. If the array doesn't have the number of elements the tuple has, this results in invalid values which breaks the type safety guarantee of the `Upcast` trait. The inverse (`ArrayTuple<(T, *)>` -> `Array<T>`) is a superset relationship so upcasting is always safe.

This example no longer compiles:
```rust
let array: Array<Number> = Array::new_typed();
// upcast should only be able to cast concrete values into more
// general values where the conversion cannot fail
let _one: ArrayTuple<(Number,)> = array.upcast_into();
let first: Number = _one.get0();
// this throws because the number is not a number
assert!(first.value_of().is_finite());
```

Fixes #5184

### Checklist

Note this is technically a breaking change. I see some breaking changes for apis where that breaking changed fixed a bug have been released in the past. If that is not the case, feel free to close

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
 
 ### Fixed
 
+* Removed invalid `js_sys::Array<T>` to `js_sys::ArrayTuple<(...)>` upcasts.
+  `ArrayTuple` encodes a fixed tuple arity, while a plain JavaScript array does
+  not prove that arity statically.
+
 * Fixed WASI targets (`wasm32-wasip1`/`wasm32-wasip2`) emitting unresolved
   `__wbindgen_placeholder__` imports, which broke component linking. The
   codegen and runtime gates now exclude `target_os = "wasi"` (restoring the
--- a/crates/js-sys/src/lib.rs
+++ b/crates/js-sys/src/lib.rs
@@ -1803,7 +1803,7 @@ impl_tuple!(8 [JsTuple1 JsTuple2 JsTuple3 JsTuple4 JsTuple5 JsTuple6 JsTuple7 Js
 
 // Macro to generate structural covariance impls for each arity
 macro_rules! impl_tuple_covariance {
-    ([$($T:ident)+] [$($Target:ident)+] [$($Ts:ident)+]) => {
+    ([$($T:ident)+] [$($Target:ident)+]) => {
         // ArrayTuple -> Array
         // Allows (T1, T2, ...) to be used where (Target) is expected
         // when all T1, T2, ... are covariant to Target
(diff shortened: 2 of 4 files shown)


guybedford:
This is an interesting case where TypeScript actually does what we do, see e.g. https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZiEMC8MDaBGANAJiwZgF0BuAKFNEljQEMsYAje4egE3oFNCV5EzSANh1gMaAJwBcMMAFcAtgw5ieHMkA.

So this makes me wonder if TypeScript did this for a good reason, or if it was just a mistake? Would be good to understand that! Noting that with panic=unwind we have separated the concepts of safety and the formal correctness of the JS type system.

ealmloff:
Interesting! That may have more to do with destructuring than upcasting. If you explicitly assign an array tuple, it does fail to type check:

```ts
const foo: number[] = [1,2,3];
// Explicit type array assignment fails
const bar: [number, number, number, number, number] = foo;

// Implicit destructue passes
const [a, b, c, d, e] = foo;

let baz: number = e;
```

https://www.typescriptlang.org/play/?ts=6.0.3#code/MYewdgzgLgBAZiEAuGYCuBbARgUwE4DaAujALwwECMANAEzUDMRA3AFAD07MAogB4AOAGwCWwYbCgBPfjhgBDPHjmT5ECMIDmYDDjCw4c4YIitQkWFgUoC6bPmqpMuPA9vPXT+47t4S5BCBsHFwAkhhCouIwACY40HhowFBosvxyanGm4NAUcg5YDsAO0Q44fvCIQYI4FnIAXihu+GQwOGxAA

If you enable `noUncheckedIndexedAccess` in the config, the original snippet also fails. TS may be modeling destructuring as closer to indexing than upcasting:

```rust
const foo: number[] = [1,2,3];

// Implicit destructue fails with typed indexes
const [a, b, c, d, e] = foo;

let baz: number = e;
```

https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&ts=6.0.3#code/MYewdgzgLgBAZiEAuGYCuBbARgUwE4DaAujALwwECMANAEzUDMRA3AFCsD0HMAkhgA4AbAJbBhsACY5oeNMChoc8AIbDBEGAHdxACxhQAnvxwSYwsFIAe01qEiwCy6jCzPgzic5wlyCEG1ZBHFgsZQAvFHRsfDIYHDYgA

guybedford:
Got it, this is about indexing primarily for TypeScript not the actual tuple conversion process. In that case, let's not make the same mistake.

guybedford (approved):

PR #5189, opened 2026-06-08
Remove cast from fn() to usize in set_on_abort

hoodmane:


guybedford (approved):

src/rt/mod.rs
```diff
@@ -818,20 +824,16 @@ pub static mut __abort_handler: u32 = 0;
...
-        let prev = __abort_handler;
-        __abort_handler = f as usize as u32;
-        if prev != 0 {
-            Some(core::mem::transmute::<usize, fn()>(prev as usize))
-        } else {
-            None
-        }
+        __abort_handler = f;
+    }
+    if !(prev as usize) == (NO_OP_PTR as usize) {
```
Spxg, line 831: This should use `!=`, because `prev` is being bitwise-negated here.

PR #5191, opened 2026-06-08
Make abort handler work when panic=abort if specifically requested

hoodmane:
If --force-enable-abort-handler is passed, abort handler will work even with panic=abort.

Resolves https://github.com/wasm-bindgen/wasm-bindgen/issues/5090.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,16 @@
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


guybedford (approved):
Tested this out in Rust Workers and it finally fixes the sever long-standing issue there.

THANK YOU @hoodmane !

PR #5193, opened 2026-06-09
Jspi

izissise:
### Description

This implements JSPI ([JS Promise Integration](https://github.com/WebAssembly/js-promise-integration)) support in wasm-bindgen, letting plain (non-`async`) Rust suspend the wasm fiber while a JS `Promise` settles, without blocking the event loop.

Two new attributes:

* `#[wasm_bindgen(jspi)]` on exports — wrapped with `WebAssembly.promising`, so JS callers receive a `Promise` (TypeScript reflects `Promise<T>`).
* `#[wasm_bindgen(suspending)]` on imports — wrapped with `new WebAssembly.Suspending(...)`; the declared return type is the type the promise *resolves to*, and `catch` surfaces rejections as `Err(reason)` exactly like it does elsewhere.

The runtime surface is a single new function plus ambient context awareness:

* `js_sys::futures::jspi_block_on_promise` suspends on a single JS `Promise` — the complete primitive, since promises are eager and concurrency composes at the promise level (`Promise.all`/`race`), while Rust `Future`s are awaited via `jspi_block_on_promise(&future_to_promise(fut))` on the ordinary executor.
* There is no executor-side API: `spawn_local` is context-aware. When called from within a JSPI context (a `jspi` export, or transitively a task spawned from one) the task's polls are entered through a `WebAssembly.promising` boundary, so the whole call tree — including sync callees — may suspend; a suspension parks only that task's poll. Library code using plain `spawn_local` gains suspendability automatically when reached from a JSPI context. `#[wasm_bindgen(jspi)]` composes with `async fn` exports on the same mechanism (promising-entered unconditionally, as exports are entered from JS).

The feature is explicitly **experimental**: using either attribute or `jspi_block_on_promise` emits a compiler warning noting the experimental status (so this can ship safely now and stabilize later). JSPI is enabled by default in Chrome 137+, Firefox 153+, and Node.js 25+ (24 behind `--experimental-wasm-jspi`); Safari support has landed in Technology Preview 238.

### Runtime design

The runtime state is minimal by construction — the entire JS-side bridge is a single identity function wrapped in `WebAssembly.Suspending`: JSPI itself awaits the returned promise and resumes the fiber with the settled value as the import's externref return value.

**Return marshalling.** A suspending import's settled value bypasses the JS shim's return conversion (the shim returns the pending promise), so the macro marshals returns post-resume: the value arrives as an externref and is converted through a CLI-generated cast adapter with the standard descriptor-driven ABI semantics — the same conversion any import return gets, just executed after the fiber resumes. All `FromWasmAbi` return types work.

**Rejections.** A rejected promise (or a synchronous throw from the import) is thrown into wasm as a `WebAssembly.JSTag` exception at the resume point. For `catch`-marked suspending imports the in-wasm wrapper catches it and converts it to data: the reason becomes the returned value and a linear-memory flag is stored in-fiber at resume (race-free by construction), which the macro assembles into the `Result`. `js_sys::futures::jspi::block_on_promise` is itself just a `catch + suspending` import with no special-casing. Non-`catch` rejections unwind — running destructors under `panic=unwind` — and reject the promising call's promise with the original reason, integrating with the abort-handler machinery when enabled.

### Shadow-stack approach

JSPI preserves wasm locals across a suspension, but *not* globals — and LLVM's shadow stack lives in linear memory behind the `__stack_pointer` global, so anything that runs while a fiber is suspended would corrupt its stack frames.

This is handled with evacuate-on-suspend instrumentation done entirely inside the wasm module (`cli-support/src/transforms/jspi.rs`):

* `jspi` export wrappers record the shadow-stack watermark at fiber entry (a `__jspi_stack_base` global plus a small node inside the fiber's own stack region, so the tracking stays correct under arbitrary concurrent fiber interleaving);
* `suspending` import wrappers copy the fiber's live region `[SP, base)` out to a heap buffer before the call, and copy it back — restoring `__stack_pointer` from a wasm local — as the very first instructions after resume, inside a `try_table` whose catch-all arm performs the same restore before rethrowing, so exceptions unwind over a restored stack;
* a fiber that suspended completes on an empty stack by definition, so its exit resets the SP to the empty-stack top rather than the entry offset.

All persistent state rides in wasm locals, which JSPI itself preserves, so the save/restore is atomic with fiber execution and needs no JS-side ordering assumptions. Fibers run on the full main shadow stack — no per-fiber stack size to tune, no overflow guard, memory proportional to live stack depth per suspension and returned to the Rust allocator on resume. Reentrancy is fully supported: multiple in-flight suspensions operate on separate stacks without conflict.

All targets are supported, including `--target no-modules` and `--target emscripten` (same instrumentation, no interaction with emscripten's own JSPI machinery). Reference types and an exception-handling-capable engine are required; every JSPI engine ships both. Threads/atomics builds are rejected (JSPI is a single-threaded proposal). The strip signal is attribute-driven: modules that never use the JSPI attributes carry no JSPI instrumentation (the ambient-context probe is rewritten to constant false and the spawn machinery is stubbed out), and keep running on engines without exnref/JSPI.

### Testing

* Reference tests snapshot the generated JS/wat across targets, including the cast-adapter marshalling and `catch` forms.
* Node runtime tests (flagless on Node 25+, `--experimental-wasm-jspi` on Node 24) cover suspend/resume, deep recursion with post-resume heap allocation, concurrent fiber shadow-stack isolation, the full `catch` matrix (plain and suspending imports, under both promising exports and promising-entered polls, on both the handleError and wasm-catch-wrapper paths), non-`catch` rejections unwinding destructors and rejecting the promise in both entry forms, `Result`-returning jspi exports (sync and async), context-aware `spawn_local` (inheritance from jspi exports, transitivity across nested spawns, suspended polls parking only their task, wakes during suspension re-polling), post-resume `String` marshalling, SP reset for fibers entered over live frames, `SuspendError` misuse, abort-handler interplay, and `panic=unwind` destructors across suspensions.
* Emscripten codegen tests verify the library output and wasm instrumentation.
* Browser examples (`jspi`, `jspi-opfs`, `jspi-fetch-streams`) run under Playwright in CI.

### Checklist


crates/js-sys/src/futures/jspi.rs
```diff
@@ -0,0 +1,208 @@
...
+pub fn block_on_promise(promise: &Promise) -> Result<JsValue, JsValue> {
+    let id = alloc_id();
+    jspi_set_pending(id, promise);
+    // #[inline(never)] so the function prologue/epilogue saves/restores
+    // __stack_pointer, which JSPI does not preserve across fiber switches.
+    suspend(id);
+    let result = jspi_get_resolved(id);
+    jspi_cleanup(id);
+    release_id(id);
+    if result.is_instance_of::<crate::Error>() {
```
guybedford: **Blocking — a fulfilled `Error` value is misreported as a rejection.**

The JS bridge stores both resolved and rejected values into the same `_jspiResolved[id]` slot, and this code disambiguates by `instanceof Error`. So a promise that *fulfills* with an `Error` instance (a legitimate value — e.g. an API resolving to an `Error`, or `Promise.resolve(new Error(...))`) is returned as `Err`. This is a real semantic bug.

Use a discriminated result so success and failure are unambiguous, e.g. resolve to `{ok:true,value}` / `{ok:false,error}` on the JS side, or store rejection state in a separate slot/flag, instead of inferring it from the value's type.

crates/js-sys/src/futures/jspi.rs
```diff
@@ -0,0 +1,208 @@
...
+unsafe impl<T> Send for ThreadLocalWrapper<T> {}
+
+// ─── JS bridge ───────────────────────────────────────────────────────────────
+//
+// All six functions share the same inline_js module so they can share the
+// _jspiPending/_jspiResolved arrays and _jspiWakerMap without any extra
+// coupling between Rust and JS.
+
+#[wasm_bindgen(inline_js = "\
+const _jspiPending  = new Array(256).fill(null);\n\
```
guybedford: Hard cap of 256 concurrent suspensions (and a parallel 256 cap for waker ids at line 114). `alloc_id` panics (`expect_throw`) on exhaustion rather than degrading. `block_on` consumes both a suspension id and a waker id per pending poll, so deeply nested/concurrent futures can hit this faster than expected. The limit is arbitrary and the failure mode is a throw — at minimum document it; ideally grow on demand.

crates/js-sys/src/futures/jspi.rs
```diff
@@ -0,0 +1,208 @@
...
+    jspi_cleanup(id);
+    release_id(id);
+    if result.is_instance_of::<crate::Error>() {
+        Err(result)
+    } else {
+        Ok(result)
+    }
+}
+
+#[inline(never)]
```
guybedford, line 78: This `#[inline(never)] fn suspend` is the linchpin of restore-on-resume, but the reasoning needs to be spelled out. After JSPI resumes the fiber, nothing in JS re-establishes `__stack_pointer`; correctness relies on a wasm epilogue restoring the global from a preserved local. A function whose body is just `jspi_do_suspend(id)` won't necessarily allocate a shadow frame / emit such an epilogue, and the first Rust call after resume that *does* allocate a frame (e.g. `release_id` → `Vec::push` → potential malloc) will read whatever `__stack_pointer` the previously-running fiber left.

The concurrent-fiber demos are shallow and may pass by luck. Please document precisely why this is sound for arbitrary call depth and for allocation between resume and return, ideally backed by a deep-recursion + allocating test.

crates/cli-support/src/js/binding.rs
```diff
@@ -1055,6 +1067,30 @@ fn instruction(
...
+                writeln!(js.pre_try, "if (__jspi_sync_sp === undefined) __jspi_sync_sp = wasm.__stack_pointer.value;").unwrap();
+                writeln!(
+                    js.pre_try,
+                    "else wasm.__stack_pointer.value = __jspi_sync_sp;"
+                )
+                .unwrap();
+                writeln!(js.pre_try, "const __jspi_stack = __jspi_stack_alloc();").unwrap();
+                writeln!(
+                    js.pre_try,
+                    "wasm.__stack_pointer.value = __jspi_stack + 65536;"
```
guybedford: Each JSPI export gets a fixed 64 KiB shadow-stack region with no guard page and no reclamation to wasm (freed regions only return to a JS pool, so linear memory never shrinks). The default wasm shadow stack is considerably larger than 64 KiB, so a jspi export with a deep Rust call tree or large stack locals will silently overflow below `__jspi_stack` and corrupt adjacent memory. Please document the limit prominently; ideally make the per-fiber stack size configurable.

crates/cli-support/src/js/mod.rs
```diff
@@ -375,6 +394,61 @@ impl<'a> Context<'a> {
...
+    /// Emits the per-fiber shadow-stack helpers at module scope exactly once:
+    ///
+    /// - `__jspi_sync_sp` — captures the synchronous SP so non-fiber WASM
+    ///   calls between suspensions use the correct shadow stack.
+    /// - `__jspi_stack_alloc()` / `__jspi_stack_free()` — allocate/free 64 KiB
+    ///   fiber stacks via `wasm.memory.grow(1)` + a JS free-list.  Using
+    ///   `memory.grow` avoids any dependency on `__wbindgen_malloc`/
+    ///   `__wbindgen_free`, which may not be exported in modules that have no
+    ///   string or heap-allocation JS glue.
+    pub(crate) fn expose_jspi_stack_setup(&mut self) {
```
guybedford: **Emscripten mode breaks here (silently).** This multi-statement blob (`let ...; const ...; function __jspi_stack_alloc()...; function __jspi_stack_free()...`) is passed through `intrinsic()`, which in `OutputMode::Emscripten` routes to `export_to_emscripten`. That helper only special-cases a snippet starting with `function <js_name>(`; this one starts with `let`, so it falls to the value-expression branch and gets stuffed into a JSON string (`$jspi_sync_sp: "let ..."`). The two helper functions are then never defined, and the `pre_try` code calling bare `__jspi_stack_alloc()`/`__jspi_stack_free()` throws `ReferenceError` (they're also not in any `__deps`).

Separately, `wasm.memory.grow(1)` (line 414) doesn't resolve in emscripten — memory is imported (`wasmMemory`), not on `wasmExports`; cf. the `wasmMemory.buffer` handling elsewhere in this file.

Recommend a codegen guard for the emscripten target (see comment near the `current_adapter_jspi` assignment).

crates/cli-support/src/js/mod.rs
```diff
@@ -5007,6 +5087,12 @@ addToLibrary({
             ContextAdapterKind::Export(e) => format!("`{}`", e.debug_name),
             ContextAdapterKind::Adapter => format!("adapter {}", id.0),
         };
+        // Set JSPI flag on the context so `Invocation::Core::generate()` can
+        // wrap the raw WASM call with `WebAssembly.promising`.
+        builder.cx.current_adapter_jspi = jspi;
```
guybedford: Suggested location for an emscripten guard. Both `jspi` and `is_suspending` are known here, and this catches both a `jspi` export and a `suspending` import regardless of CLI invocation (the macro can't know the output target). Worded as "not yet supported" so future emscripten+JSPI work is just a deletion:

```rust
if matches!(builder.cx.config.mode, OutputMode::Emscripten) && (jspi || is_suspending) {
    bail!(
        "`#[wasm_bindgen(jspi)]` and `#[wasm_bindgen(suspending)]` are not yet \
         supported with the emscripten target"
    );
}
```

`bail!`/`OutputMode` are already in scope; cf. the `--target no-modules` bail earlier in this file.

crates/cli-support/src/js/mod.rs
```diff
@@ -5233,8 +5325,18 @@ addToLibrary({
...
                 }
 
-                self.wasm_import_definitions
-                    .insert(core, ImportDefinition::Function(code));
+                // If this import is marked `#[wasm_bindgen(suspending)]`, wrap
+                // it with `WebAssembly.Suspending` so that calling it from WASM
+                // suspends the current fiber.
+                let import_def = if self.aux.imports_with_suspending.contains(&id) {
+                    ImportDefinition::Expression(format!(
+                        "new WebAssembly.Suspending(function{code})"
```
guybedford: In emscripten mode this `ImportDefinition::Expression` is emitted as an object-valued `addToLibrary` symbol (`jspi_do_suspend: new WebAssembly.Suspending(function(...){...})`). Emscripten's JS library system expects function (or simple value) symbols that it wraps and wires as wasm imports; installing a `WebAssembly.Suspending` *object* this way isn't something that mechanism routes into the import object correctly. Another reason for the emscripten guard.

crates/js-sys/src/lib.rs
```diff
@@ -9028,6 +9028,53 @@ pub mod WebAssembly {
...
+        pub fn new(func: &Function) -> Suspending;
+
+        /// Wraps a WebAssembly exported `Function` so that calling it returns a
+        /// `Promise` and enables JSPI suspension inside.
+        ///
+        /// Use `WebAssembly.promising` to obtain a "promising" wrapper around a
+        /// raw WASM export; any JSPI suspensions that occur while the function
+        /// executes will resolve the returned `Promise` when the fiber finishes.
+        ///
+        /// See [`jspi_exports`][crate::jspi_exports] for a wasm-bindgen
```
guybedford: Broken intra-doc link: `[`jspi_exports`][crate::jspi_exports]` — no `jspi_exports` symbol exists anywhere in the tree. This will trip `broken_intra_doc_links` on the docs build.

crates/macro-support/src/parser.rs
```diff
@@ -1090,6 +1094,15 @@ impl<'a>
             }
         }
         let assert_no_shim = opts.assert_no_shim().is_some();
+        let suspending = opts.suspending().is_some();
```
guybedford, line 1108: Validation is incomplete. `is_jspi` flows through `function_from_decl` for all exports, but nothing restricts `#[wasm_bindgen(jspi)]` to free functions/methods — a constructor or getter/setter would flow through, and a constructor returning a `Promise` can't be `new`'d. Symmetrically, `suspending` on an *export* is silently ignored (only the import path consumes it). Consider rejecting `jspi` on constructors/getters/setters and `suspending` on exports.

examples/jspi-fetch-streams/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+authors = ["The wasm-bindgen Developers"]
+edition = "2024"
```
guybedford: `edition = "2024"` here while the other two new examples use `2021`. Edition 2024 requires Rust 1.85, and the workspace `rust-version` is `1.77`. The dedicated MSRV jobs build `crates/msrv/*` rather than the examples so this won't be caught there, but there's no reason for the divergence — please align on 2021.

guide/src/reference/jspi.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Testing
+
+JSPI exports require a real browser with the flag enabled; Node.js does not
+support `WebAssembly.Suspending` / `WebAssembly.promising`.
+
+### Building the examples
+
+```
+just build-jspi-opfs
```
guybedford: `just build-jspi-opfs` and `just test-jspi-examples` (line 137) don't exist in `justfile`. This section also references a `tests.html` (lines 146, 151) and "QUnit + Playwright" tests (line 116) that aren't part of this PR. Either add the targets/tests or update the docs to match what actually ships.

crates/js-sys/src/futures/mod.rs
```diff
@@ -32,6 +32,8 @@
 
 extern crate alloc;
 
+pub mod jspi;
```
guybedford: `pub mod jspi` is added unconditionally, whereas the `WebAssembly.Suspending`/`WebAssembly.promising` bindings in `lib.rs` are gated behind `js_sys_unstable_apis`. For consistency — and because JSPI is itself still stabilizing across engines — consider gating this new public API behind the unstable flag too.

examples/jspi/index.html
```diff
@@ -0,0 +1,98 @@
...
+    if (elapsed > 400) {
+        err(`✗  concurrent sleeps took ${elapsed.toFixed(0)} ms — expected ~50 ms (parallel)`);
+        return;
+    }
+    log(`         all 3 fibers done in ${elapsed.toFixed(0)} ms (ran concurrently)`);
+    log("         shadow-stack isolation confirmed — no fiber corrupted another");
+    log("");
+}
+
+run().catch(e => err("Fatal: " + e));
```
guybedford, line 165: Note this (and the auto-run `click()` in the opfs/fetch-streams examples) will execute inside the CI Playwright run, since `examples/playwright.spec.ts` auto-discovers every example with a `package.json` and the runner throws on any `console.error`. Please confirm the intended CI behaviour for JSPI examples (run vs. skip) and that headless Chrome in CI has JSPI enabled by default.

guybedford (commented):
> **AI-generated review.** This was produced by an LLM-based code-review agent at a maintainer's request. Treat it as a thorough triage pass, not authoritative judgement — please verify each point yourself, especially the runtime/correctness claims about the shadow-stack mechanism, which are reasoned about statically rather than measured.

Thanks for the substantial work here — full JSPI integration (the `jspi`/`suspending` attributes, `block_on`/`block_on_promise`, per-fiber shadow-stack isolation, examples, and guide) is a genuinely useful feature and the codegen is well commented.

That said, I don't think it's mergeable as-is. The themes below are expanded in inline comments:

**Blocking**
- **No automated tests.** The codegen adds entirely new JS output paths (`binding.rs`, `mod.rs`) with zero reference tests (`crates/cli/tests/reference`) and no js-sys wasm tests. For a feature manipulating `__stack_pointer` this is the biggest gap.
- **Examples vs CI conflict.** `examples/playwright.spec.ts` auto-globs every `*/package.json`, and CI (`main.yml` `build_examples`/`test_examples`) builds all examples and runs the full Playwright suite under `channel: 'chrome'`, failing on any `console.error`. The new examples will run in normal CI even though CONTRIBUTING.md says they must be run manually. Needs deliberate wiring or an explicit skip, plus confirmation CI's Chrome has JSPI on by default.
- **`block_on_promise` conflates a fulfilled `Error` value with a rejection** (see inline).
- **Emscripten target produces silently broken output** — the stack-setup intrinsic is mangled into a string, `wasm.memory.grow` doesn't resolve, and the suspending import is emitted as an object-valued library symbol (see inline). A clean `bail!("... not yet supported with the emscripten target")` guard would convert the miscompile into an actionable error while leaving room for future emscripten+JSPI support.
- Broken intra-doc link, non-existent `just` targets in the guide (see inline).

**Needs explanation / hardening**
- The per-fiber shadow-stack restore-on-resume mechanism is the riskiest part and is under-justified; fixed 64 KiB stacks with no guard or reclamation; hard 256-concurrent limit; incomplete `jspi`/`suspending` attribute validation (see inline).

**Minor**
- Edition 2024 vs MSRV 1.77 inconsistency; unconditional `pub mod jspi` while the `WebAssembly.Suspending`/`promising` bindings are unstable-gated; schema version bump; missing changelog entry.

Happy to expand on any of these.

izissise:
Thanks for the thorough review. Here's what's been addressed since the last push:

  Blocking

  - block_on_promise Error discrimination — fixed. The JS bridge now carries a separate _jspiRejected[id] boolean; jspi_cleanup resets it. Fulfilling a Promise with an Error value no longer produces a false rejection.
  - Emscripten guard — added a bail!("not yet supported with the emscripten target") early in generate_adapter for any export with jspi: true or import with suspending: true. Converts the miscompile into a clear, actionable error.
  - Broken intra-doc link + non-existent just targets in the guide — both removed.
  - pub mod jspi gated — now behind #[cfg(js_sys_unstable_apis)]. Example crates carry #![cfg(js_sys_unstable_apis)] and set RUSTFLAGS=--cfg=js_sys_unstable_apis in their package.json build scripts (matching the webxr pattern for
  web_sys_unstable_apis).
  - CI / Playwright conflict — Chrome has had JSPI enabled by default since v123, so the Playwright suite (channel: 'chrome') should exercise the examples without any extra flag. The demo that intentionally shows stack overflow
  (deep_stack) no longer calls console.error; it logs informational output so CI doesn't fail.

  Hardening

  - Shadow-stack restore-on-resume — the previous #[inline(never)] comment was imprecise. The mechanism is now documented in full: the CLI-generated WebAssembly.Suspending wrapper saves wasm.__stack_pointer.value before await and
  restores it in finally, which runs before the fiber resumes. There is no reliance on LLVM emitting a shadow-stack prologue/epilogue for the trivial suspend shim. The block_on_promise doc comment spells out the numbered execution
  sequence, including why release_id → Vec::push → malloc is safe immediately after resume.
  - Stack size — per-fiber stack is now configurable via --jspi-stack-pages N (default 1, each page = 64 KiB). Documented prominently in the guide with a sizing table, overflow symptom list, and wasm-bindgen CLI usage. The deep_stack
  example demonstrates overflow at 1 page / correctness at 2 pages.
  - Unbounded ID pools — the fixed-256 Option<Vec<u32>> pools are replaced by a growable IdPool { free: Vec<u32>, next: u32 }. No hard limit; IDs are recycled via a free list. JS side uses sparse [] instead of pre-filled Array(256).
  - Attribute validation — #[wasm_bindgen(jspi)] is now rejected on constructors and getter/setter operations with a clear diagnostic.

 Automated tests — reference tests added in crates/cli/tests/reference/jspi.rs, run under three flag variants:
  - --target=bundler — snapshots the full bg.js output: both WebAssembly.Suspending wrappers (with __stack_pointer save/restore), both WebAssembly.promising wrappers (with fiber-stack alloc/free), the
  __jspi_stack_size/__jspi_stack_alloc/__jspi_stack_free intrinsics, and Promise<T> TypeScript signatures.
  - --target=web — same for the web target's combined output.
  - --target=bundler --jspi-stack-pages 4 — verifies the configurable stack flag threads through correctly: __jspi_stack_size = 262144 and memory.grow(4) instead of 65536/memory.grow(1).

  Any future codegen change that touches the JSPI JS output paths will break these snapshots and require an explicit re-bless.

  Two browser-level correctness tests are also wired into the jspi example's Playwright suite (demo 4 and demo 5): deep_alloc(20) recurses 20 levels, suspends at the bottom, then allocates a Vec at every level on the way back up
  (verifiable return value 1210); a second variant runs three concurrent deep_alloc fibers and checks all return the correct value, confirming per-fiber SP isolation. These run in CI via the existing test_examples job.


  Minor

  - edition = "2024" → 2021 (MSRV 1.77).
  - Schema version bump included.

guybedford (commented):
> **AI-assisted review (follow-up).** Same caveats as the first pass: treat the runtime/correctness reasoning as triage to verify rather than authoritative — in particular the JSPI ⇄ exception-handling interaction below is reasoned statically, not measured.

Thanks for the fixes since the last round — the `_jspiRejected` rejection discrimination, the emscripten `bail!`, the `js_sys_unstable_apis` gating, the growable ID pools, the constructor/getter/setter + `async`/`jspi` attribute validation, the doc-link cleanup, and the new bundler/web/`--jspi-stack-pages` reference snapshots all look good. The snapshots in particular are a solid regression net for the codegen paths. I ran the reference suite locally and `jspi` passes alongside the rest.

One substantive item remains around fiber-stack overflow safety, plus two runtime-test gaps I'd like covered before merge. Everything else is minor.

## Fiber-stack overflow safety — please add a guard

I'm happy with the per-fiber-slot design (disjoint slot + swap only `__stack_pointer`, rather than copying stack *contents*). It's clean and avoids a per-suspend memcpy, and the disjoint slots neatly sidestep the cross-fiber spill-stack clobbering that this approach has to deal with explicitly.

The cost of the choice is that it gives up two properties the normal stack has:

1. **Overflow no longer traps.** The normal Rust/wasm stack is "stack-first": it lives at `[0, __heap_base)` and grows *toward 0*, so overflow underflows past address 0 and faults out-of-bounds. I confirmed this on a minimal `wasm32-unknown-unknown` cdylib — `__stack_pointer` inits to `1048576` with data/heap above. The fiber slots are carved from `memory.grow`'d high memory and grow *down into live data* (the previously-grown slot, or the heap), which is always in-bounds, so overflow silently corrupts the neighbour instead of trapping. `deep_stack` is literally a demonstration of that corruption.
2. **Fixed 64 KiB ceiling** (vs ~1 MiB main stack), and slots are pooled but never reclaimed.

In this design, overflowing a slot is the *only* channel by which one fiber can corrupt another, so detection is load-bearing, not a nicety.

**Request:** add a guard before merge. One caveat so nobody chases the wrong thing — a true trapping guard *page* is impossible in wasm: no in-bounds linear-memory address can be made to fault, and only the single region adjacent to address 0 gets the OOB trap. So "guard" here means a sacrificial band plus an explicit check that turns silent corruption into a thrown `RangeError`.

The check is nearly free, because the suspending-import wrapper already reads SP at the fiber's deepest point. Track a per-fiber floor and verify it there, save/restoring the floor alongside `__sp` so it stays correct across interleaved/nested fibers:

```js
// export prelude: reserve a guard band at the bottom of the slot
const __jspi_stack = __jspi_stack_alloc();
__jspi_active_floor = __jspi_stack + __jspi_guard_size;            // usable floor
wasm.__stack_pointer.value = __jspi_stack + __jspi_stack_size;     // top unchanged

// suspending-import wrapper
const __sp = wasm.__stack_pointer.value;
const __floor = __jspi_active_floor;
if (__sp <= __floor) throw new RangeError('JSPI fiber stack overflow');
try { return await __inner(...args); }
finally { wasm.__stack_pointer.value = __sp; __jspi_active_floor = __floor; }
```

The band `[stack, stack + guard_size)` absorbs an overshoot up to `guard_size` *within the fiber's own slot* before it can reach the neighbour, and the check throws (before the `await`, so it propagates as a trap → the export's promise rejects) instead of resuming a corrupted fiber. For `deep_stack` (~96 KiB live at the suspend) `__sp` is well below the floor → it throws rather than corrupts. Optionally also stamp a sentinel across the band and verify it in the export's completion `finally`, to catch transient overflows that unwind before any suspend.

It isn't airtight (a single frame larger than the band that never crosses a checkpoint still slips through), but it converts the common, demonstrated failure mode from silent memory corruption into a clear error — and lets `deep_stack`/the guide say "overflow throws" rather than "overflow silently corrupts", which also removes the risk of that example corrupting the instance in CI.

**Long-term (nice to have, not for this PR):** the shared-stack + content save/restore approach used by Pyodide/CPython is arguably the more *correct* model — it keeps the overflow trap and the full stack size, and the copy cost can be made lazy/COW. See Hood Chatham's writeup, which walks through exactly the reentrancy hazard and the save/restore fix: https://blog.pyodide.org/posts/jspi-with-c-runtime/ (the "Troubles with reentrancy" section). I'm fine shipping the slot-based design now given its simplicity and speed; I'd just like the guard above plus the limitations documented, so the tradeoff is explicit and we can revisit later.

## Two requested runtime test cases (`--force-enable-abort-handler`)

The catch handler wraps *every* import (`wit.implements`), and under `--force-enable-abort-handler` (panic=abort + modern EH) each gets an `Aborting` `try_table` wrapper that catches JS exceptions in-wasm and routes critical ones to the abort handler. That's directly relevant to JSPI and currently untested. Both tests need a JSPI-capable runtime (Chrome ≥137 / Node 24 `--experimental-wasm-jspi`) and a module built with modern exception-handling enabled so the handler is actually generated (otherwise `--force-enable-abort-handler` is a no-op). The natural home is a `--force-enable-abort-handler` build variant of the `jspi` example wired into the existing Playwright suite.

**Test 1 — suspending import called without a `promising` frame aborts cleanly.** Calling a `WebAssembly.Suspending` import with no `promising` frame on the stack throws a JS `SuspendError` at the import boundary. In the default config that's an uncaught foreign exception that tears the instance down without cleanup; under the abort handler the import's `Aborting` wrapper should catch it and route to a clean abort.

Setup:
```rust
#[wasm_bindgen(suspending)]
extern "C" { fn sleep(ms: u32); }

#[wasm_bindgen] // intentionally NOT #[wasm_bindgen(jspi)]
pub fn misuse_suspend() { sleep(0); }
```
Build the module with modern EH, run `wasm-bindgen --force-enable-abort-handler`, then from JS call `misuse_suspend()` and assert it rejects/throws via the abort path (not a silent teardown). Build a second copy *without* `--force-enable-abort-handler` and document/observe the contrasting (corrupting) behaviour if useful. Run on a fresh instance and discard it afterwards — the abort can leave the instance unusable.

**Test 2 — a `jspi` export routes its own faults through the abort handler, and suspend/resume still works with the handler armed.** Two parts:

a) Panic inside a fiber surfaces as a rejected promise, not a hang or silent corruption:
```rust
#[wasm_bindgen(jspi)]
pub fn panics_in_fiber() {
    let p = js_sys::Promise::resolve(&JsValue::UNDEFINED);
    block_on_promise(&p).unwrap_throw();
    panic!("boom after resume");
}
```
`await panics_in_fiber()` should reject.

b) **The important one:** rebuild the existing happy-path demos (`do_sleep`, the concurrent `deep_alloc`) with `--force-enable-abort-handler` and assert they still suspend, resume, and return the correct values. This is the only thing that actually proves a `try_table` EH frame survives a JSPI stack switch — the catch-wrapper transform predates JSPI and assumes imports return synchronously; here it's wrapping a call that suspends mid-flight. The EH landing pad lives on the native stack (which JSPI preserves), so it *should* compose, but that's exactly what needs measuring rather than reasoning, especially since the PR independently juggles `__stack_pointer` on the shadow stack.

**Test 3 — panic=unwind across a suspend (this is the default EH config, not opt-in).** The same `try_table`-around-a-suspending-call interaction as Test 2(b) happens *by default* under panic=unwind: `generate_wasm_catch_wrappers` keys off `detect_exception_handling_version`, and a panic=unwind module already contains EH instructions, so the catch wrappers are generated with no flag. Every suspending import call is then wrapped in an EH frame that's live across the JSPI stack switch on the happy path, every suspend. If that composition is broken, panic=unwind + JSPI is broken out of the box and silently — so this is arguably the most important variant to cover.

Build the `jspi` example `-Cpanic=unwind` (nightly + `-Zbuild-std`, matching the existing `runtest_panic_unwind` setup) and assert:

- **Happy path:** the concurrent `deep_alloc` demo still suspends/resumes and returns correct values — the EH-survives-the-switch proof in the default EH config.
- **Panic in a fiber → rejected promise,** with a `Drop` type held across the `block_on_promise` to confirm destructors actually run on the resumed-then-unwound stack (not just that the promise rejects).
- **Panic *after* resume** specifically (suspend → resume → panic) — exercises unwind starting from a post-switch native stack with the shadow SP freshly restored by the suspending wrapper's `finally`.

Mechanically this should work (unwind rides the native stack JSPI preserves; SP is restored before resume so destructors pop against the right shadow stack), and it's strictly cleaner than abort here — the export's JS `finally` (`SP = sync_sp; __jspi_stack_free(slot)`) reclaims the slot on rejection either way, and panic=unwind additionally runs the Rust destructors that abort drops on the floor. But "should" is the point: none of it is currently exercised.

## Minor

- **Browser-version docs are off.** `js-sys`/guide say "Chrome 117", `CONTRIBUTING.md` says "enabled by default since v123". JSPI was flag/origin-trial-gated through Chrome 136 and only shipped *enabled by default in Chrome 137*. "117 minimum" needs a flag; "v123 by default" is wrong. Worth correcting, since CI's `channel: 'chrome'` relies on default-on JSPI.
- **Document the fixed-size + never-reclaimed behaviour** prominently: each concurrent fiber pins `--jspi-stack-pages × 64 KiB`, freed slots are pooled but `memory.grow` never shrinks, so peak memory is the high-water mark for the instance lifetime.
- **`SuspendError` UX:** a plain export transitively calling a suspending import fails at runtime with an opaque JS error type that doesn't point at the offending Rust call. Can't easily be a compile error (it's a whole-program reachability property), but it should at least be documented; longer term a walrus reachability pass could `bail!` with the function name.
- Changelog entry still missing; checklist item unchecked.
- Branch is a few commits behind `main` — needs a rebase.

guybedford:
I went ahead and had Claude implement the follow-on feedback here. Will land in the next few days unless there is further feedback.

izissise:
Sorry I didn't find time last week to look back on this, thanks for implementing the guard and the other issues

Can I ask you to wait a few days before merging? I want to make sure that the original project this is implementing in still runs correctly when jspi implementation is switch on this branch.

Thanks,
Hugues

guybedford:
Sure, there is no rush here at all, I just wanted to make sure it wasn't being dropped.

izissise:
It started using this branch on my project, currently I have one pain point, I use `wasm-bindgen-cli` with `--target no-modules` and so hit this error `error: importing from ./snippets/js-sys-794b629a11744a30/inline0.js isn't supported with --target no-modules`.
Do you think we would be able to avoid using inline in this pr?

guybedford:
Thanks for spotting that, I've extended these calls to intrinsics. I'd value a final review if you're able to take another look.

izissise:
I've tested more deeply, and everything works correctly, I do have to set `--jspi-stack-pages 64` for debug build not to overflow the stack which prove the guard also works great.

Thanks for the thorough review and help.
Regards,
Hugues

guybedford:
I'm going to hold back on this just a little bit - the reason being that we now have the `wasm32-unknown-emscripten` target for wasm-bindgen as well, and we need to figure out how the JSPI implementations will interoperate between these two systems.

I think it would be worth at least having the discussions with respect to these interactions with other folks including @hoodmane before committing to the path here.

Hopefully it won't be more than a couple of weeks.

izissise:
Hello @guybedford, any news on this?

guybedford:
I've comprehensively reworked the approach here to move the wrappers from JS into Wasm. This is because formally speaking microtask ordering guarantees of stack save/restore were not guaranteed by the JS wrapper approach given microtasks can run before resume.

Secondly, to ensure compatibility I've changed the approach to instead of allocating custom stack pages, maintain the same normal stack space, but save on suspend instead using normal malloc to store the saved stacks. Stable addresses are maintained by storing the offset as well. This does incur a cost of copy on suspend, but the benefit is then removing the need for guard ranges entirely.

guybedford (approved):

guybedford:
Since this ended up being a fairly major reworking, I'm going to delay this one more release. @izissise if you hsve a moment to try it out please do report back.

izissise:
Hello, 

I rebased my project on this new changes, after updating the cli:

```
cargo install wasm-bindgen-cli --git https://github.com/izissise/wasm-bindgen --branch jspi --force
    Updating git repository `https://github.com/izissise/wasm-bindgen`
  Installing wasm-bindgen-cli v0.2.126 (https://github.com/izissise/wasm-bindgen?branch=jspi#2c4a36a1)
    Updating crates.io index
warning: patch `wasm-bindgen-futures v0.4.76 (/home/hugues/.cargo/git/checkouts/wasm-bindgen-9477debe59ea023c/2c4a36a/crates/futures)` was not used in the crate graph
warning: patch `wasm-bindgen-test v0.3.76 (/home/hugues/.cargo/git/checkouts/wasm-bindgen-9477debe59ea023c/2c4a36a/crates/test)` was not used in the crate graph
warning: patch `web-sys v0.3.103 (/home/hugues/.cargo/git/checkouts/wasm-bindgen-9477debe59ea023c/2c4a36a/crates/web-sys)` was not used in the crate graph
...
```

and setting wasm-bindgen to use this branch:
```
[patch.crates-io]
wasm-bindgen = { git = "https://github.com/izissise/wasm-bindgen", branch = "jspi" }
js-sys = { git = "https://github.com/izissise/wasm-bindgen", branch = "jspi" }
web-sys = { git = "https://github.com/izissise/wasm-bindgen", branch = "jspi" }
wasm-bindgen-futures = { git = "https://github.com/izissise/wasm-bindgen", branch = "jspi" }
``` 

The project compiles, but the wasm-bindgen-cli fails with this error:
```bash
$ wasm-bindgen \
    --target no-modules \
    --out-dir "$OUT_DIR"/ \
    target/wasm32-unknown-unknown/"$PROFILE"/"$SW_NAME".wasm
error: unknown intrinsic `__wbindgen_jspi_set_pending`
```

I can find jspi_set_pending in the wasm-bindgen repo but only in the examples.

What am I doing wrong?

guybedford:
@izissise thanks for taking a look. `__wbindgen_jspi_set_pending` is no longer present in the latest version of this branch, so perhaps you need to ensure you're fully rebuilding and running on the latest version of this branch further?

izissise:
@guybedford sorry indeed I previously failed to correctly update Cargo.lock, it now compiles correctly.

When loading the chromium I get the following error
```
CompileError: WebAssembly.instantiate(): Compiling function #30141:"<web_sys[bc4fedeb84ea7518]::features::gen_Heade..." failed: module uses a mix of legacy and new exception handling instructions @+17026209
```
It is un-helpfully truncated, but I don't enable HeadersGuardEnum in Cargo.toml so it must comes from `Headers`.

I'll ask an LLM to understand better why there is a mix

Edit: Updating to clang-20 fixed this issue.

izissise:
After setting export `RUSTFLAGS="--cfg=js_sys_unstable_apis -Clink-args=-zstack-size=4194304"` everything compiles and works correctly. 

Thanks for the amazing work @guybedford !
LGTM

guybedford:
Pushed three commits on top of this PR following a design review of how JSPI interacts with async runtimes:

**`js_sys::futures::jspi::block_on` is removed; `block_on_promise` is the sole primitive.** A runtime-oblivious future executor posing as a generic `block_on` is a semantic trap (runtime-bound futures suspend forever), and it's unnecessary: promises are eager, so concurrency composes at the promise level (start calls, then suspend; `Promise::all`/`race` for combinators), and a Rust `Future` is awaited from sync jspi code as `block_on_promise(&future_to_promise(fut))` on the ordinary executor.

**Added `js_sys::futures::jspi::spawn_local`.** Suspending under the async executor is a property of how a poll is *entered*, not of the model: it runs a future like `spawn_local`, but each poll goes through a `WebAssembly.promising` boundary, so sync code reached from the task may itself suspend. A suspension parks only that task's poll — the queue, other tasks, and the event loop continue — with a poll-in-flight flag so wakes during suspension re-poll instead of reentering. The runtime test proves the semantics with a deadlock detector: two tasks suspended mid-poll in parallel are unblocked *by* an ordinary `spawn_local` future's multi-poll progress on the same queue, resolved out of order.

**`#[wasm_bindgen(jspi)]` now composes with `async fn` exports** on the same mechanism (`jspi::future_to_promise`): the JS contract is identical to a plain async export — the promise is ordinary and the export itself is not promising-wrapped — but the body's whole call tree may suspend, so async code can call sync-style suspending libraries.

Also rebased onto main, reblessed the jspi reference outputs (0.2.127 changed the version-salted shim hashes), and added the missing `wasm-bindgen-futures` dep to the jspi runtime test project.

Remaining follow-on (not in this PR): a monotonic fiber id + `in_fiber()`/`current_fiber_id()` intrinsics for activation-scoped identity (evacuation makes address identity unsound); non-breaking to add later.

izissise:
I've removed the usage of block_on in favor of spawn_local where that make sense, and everything still works as intended

guybedford:
Today I switched the unstable cfg to a new feature on js_sys for this (experimental-jspi). I've also worked through some remaining bugs and also fleshed out coverage for interaction with the catch attribute on imports, as well as panic=unwind tests too.

I think this is close to finally there!

guybedford (approved):

izissise:
It still works for me!

PR #5196, opened 2026-06-11
fix: make __wbindgen_destroy_closure private

ealmloff:
### Description
Currently `__wbindgen_destroy_closure` is publicly documented under the closure module: https://wasm-bindgen.github.io/wasm-bindgen/api/wasm_bindgen/closure/fn.__wbindgen_destroy_closure.html

As far as I can tell, this is not meant for external consumers. It starts with `__wbindgen` which is typically reserved for cli consumed symbols and has no external callers outside of the explicit export search in the cli

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ### Changed
 
+* Made the internal `__wbindgen_destroy_closure` export private in the Rust API.
+
 ### Fixed
 
 ### Removed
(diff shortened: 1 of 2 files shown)


guybedford:
//cc @RReverser there's no reason for this to be public right?

guybedford (approved):

RReverser (approved):
Huh yeah not sure why was it. Good catch!

PR #5201, opened 2026-06-12
Represent descriptors as a tree structure

logan-gatlin:
### Description
In #5171, descriptors are represented as a flat array of bytes. The current design uses an oversized buffer of static size, with the intention to compute the correct size at compile time once generic const expressions (GCEs) are stabilized. This comes with the tradeoffs that there is now a maximum size for descriptors, and intermediate files become more bloated.

I have some concerns with this. GCE's are a very complicated feature that probably won't be here very soon. The feature also sits at the intersection of the three deadly sins of compile time inflation, consteval, trait solving, and macros. I think the approach in this PR could work as a better long term solution. It solves the tradeoffs from the original approach, and doesn't use any unstable features.

Instead of using an array type, the size of which must be known or computed at compile time, this PR uses a tree data structure. 
```rust
#[repr(C)]
pub struct Schema {
    tag: u32,                            // SCHEMA_NODE_LEAF / WRAP / CAT
    // Rust fat array pointers (&[]) are not C ABI stable, 
    // so I used manual pointer + length fields instead
    words: &'static u32,                 // base of a words_len-long run of opcodes
    words_len: usize,
    children: &'static &'static Schema,  // base of a children_len-long run
    children_len: usize,
}
```
Instead of constructing compound schemas by appending to an array until the maximum size is reached, schemas are nested in a tree data structure. The tree can be traversed by the CLI to reconstruct the schema.

### Checklist


guybedford:
Amazing, I will take a look next week. Does this still work correctly against a higher value of `CARGO_PROFILE_RELEASE_CODEGEN_UNITS` as well, or does the referencing require code unit colocation?

logan-gatlin:
> Does this still work correctly against a higher value of `CARGO_PROFILE_RELEASE_CODEGEN_UNITS` as well, or does the referencing require code unit colocation?

I tried running it with this set high and didn't see any issues, but I'm not sure that I am actually testing what I think I am. I would imagine even with this value very high, cargo may decide to use just one codegen unit if the code is small? Linkers are an area I am not super familiar with.

In principle, I'm pretty sure everything I'm doing is well defined Rust so I would be surprised if this caused issues. Has this project seen issues in the past with this flag that would make you suspect it?

guybedford:
> Has this project seen issues in the past with this flag that would make you suspect it?

Yeah this was an issue I hit previously, but it likely doesn't apply here so nothing to worry about.

guybedford:
The tree approach here seems right to me. What would make the schema much easier to review, and also really help ensure we're on the right footing for the project, is to have a maintained schema specification in the repo under the guide itself. This way other projects that want to share the schema would be able to do that as well. For example, @ealmloff's wry-bindgen project would benefit from this I believe (and Evan please let me know if I've got this wrong too).

So I'd suggest adding a `guide/reference/schema.md` or similar that includes a comprehensive technical specification of the types and binary layouts used so that someone given an arbitrary binary would be able to figure out how to interpret it. The benefit of this will in turn be exposing rough edges that we should work on further to ensure the best schema design.

ealmloff:
wry-bindgen reads the type definition at runtime, so either representation of the new descriptor is easy to reach through the unstable trait. After the bytes are collected, either a schema or unstable api for `Descriptor::decode` would be extremely useful

logan-gatlin:
Rebased

PR #5205, opened 2026-06-17
Fix externref/multivalue transforms when `strip = true` drops the `target_features` section

connyay:
### Description

wasm-bindgen only ran the externref/multivalue transforms when the input module advertised the corresponding feature in its `target_features` custom section. A release profile with `strip = true` drops that section entirely, which silently
disabled the externref transform and failed `#[wasm_bindgen(catch)]` builds with "externref table required for catch wrappers".

The transforms now key off the explicit advertisement instead. A new target_feature_state distinguishes an explicit -feature from an absent or stripped section, which the old bool-returning target_feature conflated. An explicit +/-
always wins.

When the section is absent we can't read the advertisement, so we fall back to detecting whether the wasm-bindgen runtime was built with reference-types, via the presence of its externref intrinsics (e.g. `__externref_table_alloc`). `strip
= true` drops only the custom section and keeps those exports, so stripped modern builds still get the transforms and `#[wasm_bindgen(catch)]` keeps working. A `-Ctarget-cpu=mvp` build emits neither the section nor the intrinsics, so it
correctly stays off.

We can't just treat a missing section as "enabled": that turns the transforms on for mvp builds too, which then fail with "failed to find the __wbindgen_externref_table_alloc function". Keying the fallback on the intrinsics avoids that.

### Checklist

### DIFF
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -353,15 +353,18 @@ impl Bindgen {
             self.mode = OutputMode::Emscripten;
         }
 
-        // Enable reference type transformations if the module is already using it.
-        if let Ok(true) = wasm_conventions::target_feature(&module, "reference-types") {
-            self.externref = true;
-        }
-
-        // Enable multivalue transformations if the module is already using it.
-        if let Ok(true) = wasm_conventions::target_feature(&module, "multivalue") {
-            self.multi_value = true;
-        }
+        // `reference-types` and `multivalue` are universally supported by
+        // today's engines (and on by default for `wasm32-unknown-unknown` since
+        // Rust 1.82), so enable the matching transforms unless the module
+        // explicitly disables them via `-Ctarget-feature=-<feature>`. We can't
+        // gate on the `target_features` section being present: `strip = true`
+        // drops it entirely, which would silently disable these transforms and
+        // break `#[wasm_bindgen(catch)]` ("externref table required for catch
+        // wrappers").
+        self.externref =
+            wasm_conventions::target_feature_enabled_by_default(&module, "reference-types")?;
+        self.multi_value =
+            wasm_conventions::target_feature_enabled_by_default(&module, "multivalue")?;
 
         // Check that no exported symbol is called "default" if we target web.
         if matches!(self.mode, OutputMode::Web)
--- a/crates/cli-support/src/lib.rs
+++ b/crates/cli-support/src/lib.rs
@@ -353,18 +353,25 @@ impl Bindgen {
             self.mode = OutputMode::Emscripten;
         }
 
-        // `reference-types` and `multivalue` are universally supported by
-        // today's engines (and on by default for `wasm32-unknown-unknown` since
-        // Rust 1.82), so enable the matching transforms unless the module
-        // explicitly disables them via `-Ctarget-feature=-<feature>`. We can't
-        // gate on the `target_features` section being present: `strip = true`
-        // drops it entirely, which would silently disable these transforms and
-        // break `#[wasm_bindgen(catch)]` ("externref table required for catch
-        // wrappers").
-        self.externref =
-            wasm_conventions::target_feature_enabled_by_default(&module, "reference-types")?;
-        self.multi_value =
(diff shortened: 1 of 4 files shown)


connyay:
Hit this issue with cloudflare/workers-rs + `strip = true` 👉  https://github.com/cloudflare/workers-rs/issues/1014

guybedford:
Thanks for posting this. The CI seems to be failing here for mvp feature set unfortunately.

guybedford:
Is `strip=true` dropping the target features section but not the Wasm bindgen schema section? How does it know that distinction?

connyay:
> Is `strip=true` dropping the target features section but not the Wasm bindgen schema section? How does it know that distinction?

Sorry, I am new to this codebase - by `the Wasm bindgen schema section` are you referring to the `__wasm_bindgen_unstable` bytes?

https://github.com/wasm-bindgen/wasm-bindgen/blob/d37fcefbc4410f4c2f6039d839218f493f54d82d/crates/macro-support/src/codegen.rs#L181-L183

~if so, `target_features` is just metadata while `__wasm_bindgen_unstable` is linked data~

connyay:
> Is `strip=true` dropping the target features section but not the Wasm bindgen schema section?

yes.

> How does it know that distinction?

LLVM strips by known section name, not just blanket dropping all custom sections 👉 
https://github.com/llvm/llvm-project/blob/42743abe0a77cede08b203226a240155e50b4adb/lld/wasm/SyntheticSections.h#L397-L405
https://github.com/llvm/llvm-project/blob/42743abe0a77cede08b203226a240155e50b4adb/lld/wasm/SyntheticSections.h#L354-L359

guybedford:
I'm just thinking some more about this, and the target_features section sometimes applying and sometimes not feels a little brittle to me.

Do you think we could rather just modify the transforms themselves to idempotently only "switch on" when they have work to do? That is the transform itself could detect it doesn't find the things it expects, and early exit without failure instead.

When these features were experimental, it made sense to be stricter, but now that this is the backwards compat path, that feels like a nicer legacy approach that can fall out over time without being too strict about the target_features section.

In short, do we even need to parse the target_features section at all, or can we just make the transforms early bail-out forgiving?

guybedford:
For externref it seems like there is no cost to always running the transform. However for multi-value using target_features as an optimization hint seems helpful to skip running the transform unnecessarily when multi-value is not used.

crates/cli-support/src/lib.rs
```diff
@@ -353,15 +353,25 @@ impl Bindgen {
...
-        }
+        // `reference-types` and `multivalue` are on by default for
+        // `wasm32-unknown-unknown` since Rust 1.82. Honor an explicit
+        // `-Ctarget-feature=+/-<feature>` advertisement when the `target_features`
+        // section is present; when it is absent (e.g. `strip = true` dropped it),
+        // detect support from the runtime's externref intrinsics instead. That
+        // keeps the transforms running on stripped modern builds, so
+        // `#[wasm_bindgen(catch)]` still works, while leaving them off on
+        // `-Ctarget-cpu=mvp` builds.
+        let default_modern_features = wasm_conventions::module_uses_externref_intrinsics(&module);
```
guybedford, line 364: I'm not sure this should determine multi-value. I think in the unknown case we just always want to run the transform. Hence why I'm suggesting we just always run the transforms and have the transform itself early bail instead of this separate step.

guybedford: How about the following defaults:

1. Multi-value transform runs ONLY when target features explicitly excludes it / target features is set and does not include it (i.e. MVP)
2. Externref transform always runs, just bails early if the symbols are not found. If we really want a strict mode we could maintain that when the reference types target feature is enabled.

Let me know if I'm missing anything with the above.

guybedford: Hmm I'm actually not sure there's anyway we can frame this that isn't kind of unsound.

guybedford (changes requested):
Unfortunately I don't think we can land this, because there isn't a clear way to detect target_features without the target features section.

I think the better solution here would instead be to say not to use `strip=true` with wasm-bindgen builds and instead rely on `wasm-opt` to perform this role - which is always required for release builds anyway.

PR #5208, opened 2026-06-17
Emit a wasm-bindgen-exports marker comment for emscripten builds

guybedford:
This makes the clean wasm-bindgen public API discoverable for every emscripten build.

Emscripten output publishes the clean wrappers with `Module.<name> = <name>;`, but there's no machine-readable list of which names those are — tooling that wants the public API has to scrape the generated JS. This adds a marker comment as the first line of `library_bindgen.js` listing those names, comma-separated:

```js
// wasm-bindgen-exports: Counter,add
```

It's emitted on every emscripten build, independent of any other flag. The names are collected at the single point that's already the source of truth for what's public — where the `Module.<name> = <name>;` line is written — so the list can never drift from the actual exports. Internal `__wbg*` names, private classes/enums, and the mangled wasm exports (`_add`) are not included. A `//` line with no braces passes through `reset_indentation` unchanged at module top level.

Test coverage: a new CLI test builds an `add` fn + `Counter` struct, marks the module emscripten, and asserts the first line of `library_bindgen.js` is the `// wasm-bindgen-exports:` marker carrying exactly the public names, with no mangled or `__wbg*` internals.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -79,10 +79,6 @@ pub struct FinalizedOutput {
     /// Content to write alongside the main JS as a sidecar that emcc loads
     /// with `--extern-pre-js`. Empty for non-emscripten output modes.
     pub emscripten_extern_pre_js: String,
-    /// JSON array of the clean wasm-bindgen export names, written alongside the
-    /// main JS as a `library_bindgen.exports.json` sidecar. Empty for
-    /// non-emscripten output modes.
-    pub emscripten_exports_json: String,
 }
 
 pub struct Context<'a> {
@@ -102,12 +98,13 @@ pub struct Context<'a> {
     emscripten_extern_pre_js: String,
     imports_post: String,
     export_name_list: Vec<String>,
-    /// JSON array body of the clean wasm-bindgen export names, accumulated at
-    /// the same point the `Module.<name> = <name>;` lines are emitted (the
-    /// source of truth for what's publicly exported). Written to a
-    /// `library_bindgen.exports.json` sidecar so tooling can discover the
-    /// public API in every emscripten build. Empty for non-emscripten modes.
-    emscripten_exports_json: String,
+    /// Comma-separated list of the clean wasm-bindgen export names,
+    /// accumulated at the same point the `Module.<name> = <name>;` lines are
+    /// emitted (the source of truth for what's publicly exported). Emitted as a
+    /// `// wasm-bindgen-exports:` marker comment at the top of
+    /// `library_bindgen.js` so tooling can discover the public API in every
+    /// emscripten build. Empty for non-emscripten modes.
+    emscripten_exports_list: String,
     typescript: String,
     typescript_emscripten_classes: String,
     config: &'a Bindgen,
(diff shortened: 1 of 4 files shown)


sbc100:
I think it would be nice if whatever files are ingested by emscripten are standard intput file.  i.e. `--pre-js` / `--post-js` / `--library-js` files.  

Or maybe a set of extra command line flags?  

Actually, I wonder if the library JS could encode which symbols were exports somehow?  Maybe it could add to the  could the library JS file?  I think there are some setting that can be appended to in a library js file.. let me check

guybedford:
The problem here is that today, [get_wasm_bindgen_exported_symbols](https://github.com/emscripten-core/emscripten/blob/main/tools/building.py#L280-L294) picks up basically _everything_ so we need a way to introduce this information or otherwise everything from the core Wasm gets exposed by default (separate to the question of final high-level JS exports).

sbc100:
I see.. yes dumping the export list here seems better than using llvm-nm command on the emscripten side.

For simple symbol lists I test to prefer a simple flat file with one synbol per line e.g. `exported_symbols.txt` over something that requires json parsing.

sbc100:
So this direction seems fine to me. 

If we can find way to encode this directly into the library.js file (i.e. avoid the need to 2 files) then that would be even better.

guybedford:
I'll update this to use a `// wasm-bindgen-exports: a,b,c` special comment then. Further feedback welcome.

crates/cli-support/src/js/mod.rs
```diff
@@ -102,12 +98,13 @@ pub struct Context<'a> {
...
     imports_post: String,
     export_name_list: Vec<String>,
-    /// JSON array body of the clean wasm-bindgen export names, accumulated at
-    /// the same point the `Module.<name> = <name>;` lines are emitted (the
-    /// source of truth for what's publicly exported). Written to a
-    /// `library_bindgen.exports.json` sidecar so tooling can discover the
-    /// public API in every emscripten build. Empty for non-emscripten modes.
-    emscripten_exports_json: String,
+    /// Comma-separated list of the clean wasm-bindgen export names,
+    /// accumulated at the same point the `Module.<name> = <name>;` lines are
```
sbc100, line 102: It seems like it would be nice if we had a test/example along side this code so that when we change the code we can see a sample of the how the output would change.  Would that be hard to add here?

Are all the things in this list library elements from the JS file?   If so I wonder if simply doing `EXPORTED_FUNCTIONS.add(xx)` for each of them would be enough to make this work?

guybedford: > Are all the things in this list library elements from the JS file? If so I wonder if simply doing EXPORTED_FUNCTIONS.add(xx) for each of them would be enough to make this work?

Yeah this would be ideal I think and avoid the complexity here. But it means we need to refactor to use `addToLibrary` properly to define these symbols - which it sounds like we should be doing? I'll continue to work on this.

guybedford: I've posted https://github.com/wasm-bindgen/wasm-bindgen/pull/5210 as the follow-on here. This PR can then likely be closed once that lands.

guybedford:
I've verified that #5210 works to apply this instead, properly integrating with Emscripten without requiring this side-channel.

PR #5209, opened 2026-06-18
Use `#[unsafe(export_name)]` instead of `#[no_mangle]` for exports

guybedford:
wasm-bindgen indicates its wasm exports with `#[no_mangle]`, which makes the wasm symbol name implicitly equal to the Rust identifier. This switches every wasm-bindgen-emitted export to `#[unsafe(export_name = "...")]`, making the exported symbol explicit and decoupling it from the Rust identifier. This is also the forward-compatible spelling, since bare `#[no_mangle]` requires the `unsafe(...)` wrapper going forward.

The generated symbol names are unchanged, so consumers and the CLI side are unaffected.

Generated code (`crates/macro-support/src/codegen.rs`):

* struct `free`, `upcast`, field getter (the `cfg_attr` form), field setter, and the `__wbindgen_describe_*` descriptor shim now emit an explicit `export_name` equal to the previous symbol
* the existing main export shim's `export_name` is wrapped in `unsafe(...)` for consistency

Hand-written runtime exports (`src/`): `__externref_table_alloc`/`dealloc`/`drop_slice`, `__wbindgen_skip_interpret_calls`, `__wbindgen_malloc`/`realloc`/`free`/`exn_store`, `__wbindgen_destroy_closure`, and the `__instance_terminated` / `__abort_handler` statics.

One behavioural nuance: `#[no_mangle]` silently suppressed the `non_upper_case_globals` lint, but `#[export_name]` does not. The two lowercase statics are referenced by their Rust identifiers elsewhere, so they keep their idents plus an `#[allow(non_upper_case_globals)]`:

```rs
#[allow(non_upper_case_globals)]
#[unsafe(export_name = "__abort_handler")]
pub static mut __abort_handler: fn() = NO_OP_PTR;
```

`#[no_mangle]` is intentionally left in place where it is not a wasm-bindgen export: test fixtures and the parser logic that strips a *user's* `#[no_mangle]` from an exported function. The contributing design guide examples are updated to the new spelling.

Verified with `cargo build`/`clippy`/`fmt` on `wasm32-unknown-unknown`, the macro-support/shared/cli-support suites, and the CLI pipeline tests (87 pass; the 4 failures all reproduce on the base branch and are environmental). The WASI link command confirms the renamed exports still resolve.

### DIFF
--- a/crates/macro-support/src/codegen.rs
+++ b/crates/macro-support/src/codegen.rs
@@ -223,7 +223,8 @@ impl ToTokens for ast::Struct {
         let name_len = name_str.len() as u32;
         let name_chars: Vec<u32> = name_str.chars().map(|c| c as u32).collect();
         let new_fn = Ident::new(&shared::new_function(&name_str), Span::call_site());
-        let free_fn = Ident::new(&shared::free_function(&name_str), Span::call_site());
+        let free_fn_name = shared::free_function(&name_str);
+        let free_fn = Ident::new(&free_fn_name, Span::call_site());
         let unwrap_fn = Ident::new(&shared::unwrap_function(&name_str), Span::call_site());
         let wasm_bindgen = &self.wasm_bindgen;
         let class_abi = quote! {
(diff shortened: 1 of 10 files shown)


guybedford:
Becasue the `unsafe` pattern only applies since Rust 1.82, we cannot land this until the MSRV bump in October to Rust 1.82.

PR #5210, opened 2026-06-18
Hoist emscripten exports into top-level library symbols

guybedford:
This makes the clean wasm-bindgen API reachable for emscripten builds without any sidecar files, by letting emscripten emit the exports itself. It supersedes the sidecar approaches in #5207 (post-js) and #5208 (marker comment), which can be closed.

Background: under `-sMODULARIZE=instance` the consumer receives ESM named exports + a default `init`, never `Module`, so the `Module.<name> = <name>;` writes wasm-bindgen put inside the `$initBindgen` init closure were dead. Those wrappers couldn't be plain top-level declarations because e.g. `Counter` closes over a `FinalizationRegistry` built at init.

The fix follows how emscripten already exports JS library symbols: a `--js-library` runs in the compiler context, and in instance mode `jsifier` prepends `export` to any `$`-prefixed library symbol whose name is in `EXPORTED_FUNCTIONS` (which also forces its inclusion). So every clean export is hoisted into its own top-level `addToLibrary` symbol and self-registered.

Before, inside `$initBindgen`:

```js
$initBindgen: () => {
  wasm = wasmExports;
  const CounterFinalization = new FinalizationRegistry(...);
  class Counter { ... }
  Module.Counter = Counter;          // dead under instance mode
  function add(a, b) { return wasm.add(a, b); }
  Module.add = add;                  // dead under instance mode
}
```

After, as top-level library symbols:

```js
addToLibrary({
  $add: function add(a, b) { return wasm.add(a, b); },
  $add__deps: ['$initBindgen'],
  $add__postset: "Module['add'] = add;",
});
addToLibrary({
  $CounterFinalization: undefined,
  $CounterFinalization__deps: ['$wasm'],
  $CounterFinalization__postset: "CounterFinalization = (typeof FinalizationRegistry === 'undefined') ? {...} : new FinalizationRegistry(ptr => wasm.__wbg_counter_free(ptr, 1));",
});
addToLibrary({
  $Counter: class Counter { ... },
  $Counter__deps: ['$initBindgen', '$CounterFinalization'],
  $Counter__postset: "if (Symbol.dispose) Counter.prototype[Symbol.dispose] = Counter.prototype.free; Module['Counter'] = Counter;",
});

EXPORTED_FUNCTIONS.add('add');
EXPORTED_FUNCTIONS.add('Counter');
```

giving consumers:

```js
import init, { add, Counter } from './out.mjs';
await init();
add(17, 25);            // 42
new Counter(40).inc();  // 41
```

What's covered:

* **Functions, classes, enums** are hoisted; each depends on `$initBindgen` (which aggregates `$wasm` + all intrinsic deps, so nothing is tree-shaken). Factory-mode `Module.<name>` attachment is done in each symbol's `__postset`; instance-mode named export comes from `EXPORTED_FUNCTIONS`. No separate `extraLibraryFuncs` force-keep is needed — `EXPORTED_FUNCTIONS` membership already forces inclusion.
* **`#[wasm_bindgen(private)]`** classes/enums are hoisted but not attached to `Module` or self-registered (they stay module-internal, matching every other output mode).
* **Source-preservation**: `FinalizationRegistry` instances and `Object.freeze(...)` enums are emitted via `__postset` / string members, because emscripten evaluates the `--js-library` and re-serializes live values (a `FinalizationRegistry` would collapse to `{}`).
* **Namespaced exports** (`js_namespace`): the namespace root (e.g. `app`) is the public export — a `{}` library symbol whose private leaf symbols are listed in its `__deps`, with the nested shape (`app.math.Calc = ...`) assembled in the root's `__postset`.
* **User module / inline-js imports**: these were silently dropped in emscripten (the direct-import elision points a wasm import straight at the user's module, which emcc can't resolve — it only resolves against `env`). They're now wired as `addToLibrary` shims that call the ESM-imported binding, and those bindings are `__wbg_`-prefixed (`import { Module as __wbg_Module } from 'imports'`) so user-chosen names can't collide with emcc runtime globals.
* The `__wbgSuperSkip` subclass sentinel and used string-enum tables are hoisted to library symbols too, since hoisted bodies reference them at module scope.

Testing:

* New CLI codegen tests assert the generated `library_bindgen.js` shape: hoisting + `__postset` Module attachment + `EXPORTED_FUNCTIONS` self-registration + no `$initBindgen` inlining (`emscripten_exports_hoisted_to_library_symbols`); private-class exclusion; the namespaced `.d.ts`/shape (`emscripten_namespaced_exports_valid_ts`, validated with `tsc --strict`); and module + inline-js import prefixing (`emscripten_user_imports_are_prefixed`).
* The wasm-bindgen-test-runner emscripten harness was updated to model the new mechanism (symbols self-register into `EXPORTED_FUNCTIONS` and attach to `Module`), and passes in CI.
* Validated end-to-end against a real `emcc` link (`-sMODULARIZE=instance`): `add(17,25) === 42`, `Counter` exported and usable, finalizers intact.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -2305,9 +2305,18 @@ if (require('worker_threads').isMainThread) {{
                 : new FinalizationRegistry({finalization_callback})"
         );
         if hoist {
-            // The callback closes over the module-scope `wasm` global.
+            // Emit the registry's *source* as a string-valued library member.
+            // emscripten evaluates non-string library values while loading the
+            // library file, then re-serializes them from the live object. A
+            // `FinalizationRegistry` instance has no enumerable own properties,
+            // so that round-trip would collapse it to `{}` and drop
+            // `register`/`unregister`. A string member is emitted verbatim as
+            // `var XFinalization = <source>` (same as intrinsic values like
+            // `$cachedTextDecoder: "new TextDecoder()"`). The callback closes
+            // over the module-scope `wasm` global.
+            let value_lit = format!("{finalization_value:?}");
             self.emscripten_library(&format!(
-                "addToLibrary({{\n    ${identifier}Finalization: {finalization_value},\n    \
+                "addToLibrary({{\n    ${identifier}Finalization: {value_lit},\n    \
                  ${identifier}Finalization__deps: ['$wasm']\n}});"
             ));
         } else {
@@ -6502,7 +6511,12 @@ addToLibrary({
             // ESM exports / `Module.<name>`; namespaced ones are private and
             // reached through their namespace root.
             let is_namespaced = enum_.js_namespace.is_some();
-            let value = format!("Object.freeze({{\n{variants}}})");
+            // String-valued member so the `Object.freeze(...)` source is
+            // emitted verbatim. As a live value it would be evaluated while
+            // loading the library and re-serialized to a plain object, losing
+            // `Object.freeze` (see the finalization-registry note in
+            // `write_class`).
+            let value = format!("{:?}", format!("Object.freeze({{\n{variants}}})"));
             self.hoist_emscripten_export(
                 &identifier,
                 &value,
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -514,6 +514,52 @@ impl<'a> Context<'a> {
         }
     }
 
+    /// Hoist a clean export (free function, class, or enum) to a library symbol
+    /// and, when it's namespaced, register a declaration-free entry in the
+    /// export tree so the namespace root assembles `ns.<name> = <identifier>`
+    /// and the `.d.ts` namespace shape resolves `typeof <identifier>`. The
+    /// hoisted symbol is public (named ESM export / `Module.<name>`) unless it's
+    /// namespaced (reached via the root) or `private`. Shared by the function,
+    /// class, and enum emission paths.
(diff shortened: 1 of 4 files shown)


hoodmane:
>  this can't be exercised end-to-end here (no emsdk in the dev environment) 

Why not install emcc into the github action and actually run the output?

crates/cli-support/src/js/mod.rs
```diff
@@ -195,6 +195,16 @@ pub struct Context<'a> {
...
     emscripten_import_deps: BTreeMap<ImportId, BTreeSet<String>>,
 
+    /// Clean public export identifiers (free functions, classes, enums, and
+    /// namespace roots) that have been hoisted into top-level `addToLibrary`
+    /// symbols in emscripten mode. Used to self-register them at compile time
+    /// (in `generate_emscripten_wasm_loading`) via `extraLibraryFuncs`
+    /// (force-keep the symbol) and `EXPORTED_FUNCTIONS` (the `export` prefix
+    /// under `MODULARIZE=instance`), so emscripten emits the clean named
+    /// exports itself. Factory-mode `Module.<name>` attachment is handled
+    /// separately by each symbol's `__postset`.
```
hoodmane: Pretty wordy comment, maybe cut it down.

hoodmane: I guess this goes for all the comments...

guybedford: Cleaned up.

drcarney66 (approved):

PR #5214, opened 2026-06-23
Fix compilation with (feature = "std", panic = "unwind", target_feature = "atomics") plus prevent a memory leak

### LINKED ISSUE #5213: js-sys can not be built with feature "std" and threading
```text
2026-06-23T17:00:01.413773Z  INFO 🚀 Starting trunk 0.21.14
2026-06-23T17:00:01.413927Z  INFO 📦 starting build
   Compiling js-sys v0.3.102
error[E0277]: the type `UnsafeCell<Option<multithread::Inner>>` may contain interior mutability and a reference may not be safely transferable across a catch_unwind boundary
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:102:26
    |
102 |               Closure::new(move |_| {
    |  _____________------------_^
    | |             |
    | |             required by a bound introduced by this call
103 | |                 // The promise resolution acts like a wake, so ensure the state
104 | |                 // transitions to AWAKE before entering `run`.
105 | |                 this.atomic.wake_by_ref();
106 | |                 this.run();
107 | |             })
    | |_____________^ `UnsafeCell<Option<multithread::Inner>>` may contain interior mutability and a reference may not be safely transferable across a catch_unwind boundary
    |
    = help: within `multithread::Task`, the trait `RefUnwindSafe` is not implemented for `UnsafeCell<Option<multithread::Inner>>`
note: required because it appears within the type `RefCell<Option<multithread::Inner>>`
   --> /root/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:849:12
    |
849 | pub struct RefCell<T: ?Sized> {
    |            ^^^^^^^
note: required because it appears within the type `multithread::Task`
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:83:19
    |
 83 | pub(crate) struct Task {
    |                   ^^^^
    = note: required for `Rc<multithread::Task>` to implement `UnwindSafe`
note: required because it's used within this closure
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:102:26
    |
102 |             Closure::new(move |_| {
    |                          ^^^^^^^^
    = note: required for `{closure@/root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:102:26: 102:34}` to implement `wasm_bindgen::__rt::marker::MaybeUnwindSafe`
note: required by a bound in `ScopedClosure::<'static, T>::new`
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasm-bindgen-0.2.125/src/closure.rs:250:33
    |
248 |     pub fn new<F>(t: F) -> Self
    |            --- required by a bound in this associated function
249 |     where
250 |         F: IntoWasmClosure<T> + MaybeUnwindSafe + 'static,
    |                                 ^^^^^^^^^^^^^^^ required by this bound in `ScopedClosure::<'static, T>::new`

error[E0277]: the type `UnsafeCell<isize>` may contain interior mutability and a reference may not be safely transferable across a catch_unwind boundary
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:102:26
    |
102 |               Closure::new(move |_| {
    |  _____________------------_^
    | |             |
    | |             required by a bound introduced by this call
103 | |                 // The promise resolution acts like a wake, so ensure the state
104 | |                 // transitions to AWAKE before entering `run`.
105 | |                 this.atomic.wake_by_ref();
106 | |                 this.run();
107 | |             })
    | |_____________^ `UnsafeCell<isize>` may contain interior mutability and a reference may not be safely transferable across a catch_unwind boundary
    |
    = help: within `multithread::Task`, the trait `RefUnwindSafe` is not implemented for `UnsafeCell<isize>`
note: required because it appears within the type `Cell<isize>`
   --> /root/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:312:12
    |
312 | pub struct Cell<T: ?Sized> {
    |            ^^^^
note: required because it appears within the type `RefCell<Option<multithread::Inner>>`
   --> /root/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:849:12
    |
849 | pub struct RefCell<T: ?Sized> {
    |            ^^^^^^^
note: required because it appears within the type `multithread::Task`
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:83:19
    |
 83 | pub(crate) struct Task {
    |                   ^^^^
    = note: required for `Rc<multithread::Task>` to implement `UnwindSafe`
note: required because it's used within this closure
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:102:26
    |
102 |             Closure::new(move |_| {
    |                          ^^^^^^^^
    = note: required for `{closure@/root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/js-sys-0.3.102/src/futures/task/multithread.rs:102:26: 102:34}` to implement `wasm_bindgen::__rt::marker::MaybeUnwindSafe`
note: required by a bound in `ScopedClosure::<'static, T>::new`
   --> /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasm-bindgen-0.2.125/src/closure.rs:250:33
    |
248 |     pub fn new<F>(t: F) -> Self
    |            --- required by a bound in this associated function
249 |     where
250 |         F: IntoWasmClosure<T> + MaybeUnwindSafe + 'static,
    |                                 ^^^^^^^^^^^^^^^ required by this bound in `ScopedClosure::<'static, T>::new`

error: could not compile `js-sys` (lib) due to 2 previous errors
2026-06-23T17:00:06.646722Z ERROR ❌ error
error from build pipeline

Caused by:
    0: HTML build pipeline failed (1 errors), showing first
    1: error from asset pipeline
    2: running cargo build
    3: error during cargo build execution
    4: cargo call to executable 'cargo' with args: '["build", "--target=wasm32-unknown-unknown", "--manifest-path", "/workspaces/project/crates/demo/Cargo.toml"]' returned a bad status: exit status: 101
2026-06-23T17:00:06.647072Z ERROR error from build pipeline
2026-06-23T17:00:06.647080Z  INFO   1: HTML build pipeline failed (1 errors), showing first
2026-06-23T17:00:06.647086Z  INFO   2: error from asset pipeline
2026-06-23T17:00:06.647092Z  INFO   3: running cargo build
2026-06-23T17:00:06.647097Z  INFO   4: error during cargo build execution
2026-06-23T17:00:06.647102Z  INFO   5: cargo call to executable 'cargo' with args: '["build", "--target=wasm32-unknown-unknown", "--manifest-path", "/workspaces/project/crates/demo/Cargo.toml"]' returned a bad status: exit status: 101
```

`.cargo/config.toml`:

```toml
[target.wasm32-unknown-unknown]
rustflags = [
    "-C", "target-feature=+atomics",
    "-C", "panic=unwind",
    "-C", "link-arg=--shared-memory",
    "-C", "link-arg=--import-memory",
    "-C", "link-arg=--export=__heap_base",
    "-C", "link-arg=--export=__wasm_init_tls",
    "-C", "link-arg=--export=__tls_size",
    "-C", "link-arg=--export=__tls_align",
    "-C", "link-arg=--export=__tls_base",
]
```

Build command:

```
CARGO_UNSTABLE_BUILD_STD="std,panic_unwind" RUSTUP_TOOLCHAIN="nightly" RUST_LOG="info" trunk build
```

im-0:
### Description

This fixes a compilation failure and also prevents a memory leak after a panic.

I think that the same memory leak exists in singlethread.rs too but I am not entirely sure.

Fixes https://github.com/wasm-bindgen/wasm-bindgen/issues/5213

### Checklist

### DIFF
--- a/crates/cli/tests/wasm-bindgen/main.rs
+++ b/crates/cli/tests/wasm-bindgen/main.rs
@@ -1025,6 +1025,153 @@ describe('nested unreachable', () => {
         .success();
 }
 
+/// Regression test for a `Task` leak in the futures executor: when a spawned
+/// future panics during `poll`, the executor must drop the future (and
+/// everything it owns) rather than leaving it alive via the
+/// `Task -> Inner -> future -> Waker -> Rc<Task>` reference cycle. The panic
+/// propagates out of the microtask that drives the first poll, so the Node
+/// script installs an `uncaughtException` handler to observe it while checking
+/// that the future's `DropGuard` ran.
+#[test]
+fn spawn_local_panic_frees_future() {
+    let mut project = Project::new("spawn_local_panic_frees_future");
+    project
+        .file(
+            "src/lib.rs",
+            r#"
+                use wasm_bindgen::prelude::*;
+                use wasm_bindgen_futures::spawn_local;
+                use std::future::Future;
+                use std::pin::Pin;
+                use std::task::{Context, Poll, Waker};
+
+                #[wasm_bindgen(inline_js = "
+                    export function set_was_dropped(v) { globalThis.was_dropped = v; }
+                ")]
+                extern "C" {
+                    fn set_was_dropped(v: bool);
+                }
+
+                struct DropGuard;
+
+                impl Drop for DropGuard {
+                    fn drop(&mut self) {
+                        set_was_dropped(true);
+                    }
+                }
+
+                // Retains its own waker before panicking, forming the
+                // `future -> Waker -> Rc<Task>` cycle that the single-threaded
+                // executor would otherwise leak on a panicking poll.
+                struct PanicFuture {
+                    _guard: DropGuard,
+                    waker: Option<Waker>,
+                }
+
+                impl Future for PanicFuture {
+                    type Output = ();
+                    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<()> {
+                        self.waker = Some(cx.waker().clone());
+                        panic!("spawned future panic");
+                    }
+                }
+
+                #[wasm_bindgen]
+                pub fn spawn_panicking_future() {
+                    set_was_dropped(false);
+                    spawn_local(PanicFuture {
+                        _guard: DropGuard,
+                        waker: None,
+                    });
+                }
+            "#,
+        )
+        .file(
+            "Cargo.toml",
+            &format!(
+                "
+                    [package]
+                    name = \"spawn_local_panic_frees_future\"
+                    authors = []
+                    version = \"1.0.0\"
+                    edition = '2021'
+
+                    [dependencies]
+                    wasm-bindgen = {{ path = '{root}' }}
+                    wasm-bindgen-futures = {{ path = '{root}/crates/futures' }}
+
+                    [lib]
+                    crate-type = ['cdylib']
+
+                    [workspace]
+
+                    [profile.dev]
+                    codegen-units = 1
+                ",
+                root = REPO_ROOT.display(),
+            ),
+        );
+
+    // The leak fix matters when a poll unwinds, which requires building with
+    // nightly build-std under `panic = "unwind"`.
+    project
+        .cargo_cmd
+        .env("RUSTUP_TOOLCHAIN", "nightly")
+        .env("RUSTFLAGS", "-Cpanic=unwind")
+        .arg("-Zbuild-std=std,panic_unwind");
+
+    let out_dir = project.wasm_bindgen("--target nodejs").unwrap();
+
+    fs::write(
+        out_dir.join("test_spawn_panic.js"),
+        r#"
+const assert = require('node:assert/strict');
+const wasm = require('./spawn_local_panic_frees_future.js');
+
+// The spawned future's panic propagates out of the microtask that drives its
+// first poll, surfacing as an uncaught exception. That's expected here; record
+// it and swallow only that specific panic so the process keeps running.
+let sawPanic = false;
+process.on('uncaughtException', (e) => {
+    const msg = (e && e.message) || String(e);
+    if (/spawned future panic/.test(msg)) {
+        sawPanic = true;
+    } else {
+        throw e;
+    }
+});
+
+wasm.spawn_panicking_future();
+
+// A macrotask runs after the microtask queue, so by now the spawned task has
+// been polled (and panicked).
+setTimeout(() => {
+    try {
+        assert.ok(sawPanic, 'expected the spawned future to panic');
+        assert.strictEqual(
+            globalThis.was_dropped,
+            true,
+            'spawned future must be dropped after a panicking poll, not leaked',
+        );
+        console.log('PASS');
+        process.exit(0);
+    } catch (e) {
+        console.error(e);
+        process.exit(1);
+    }
+}, 50);
+"#,
+    )
+    .unwrap();
+
+    Command::new("node")
+        .arg("test_spawn_panic.js")
+        .current_dir(&out_dir)
+        .assert()
+        .success()
+        .stdout(str::contains("PASS"));
+}
+
 #[test]
 fn termination_reset_state() {
     let mut project = Project::new("termination_reset_state");
--- a/crates/js-sys/src/futures/task/multithread.rs
+++ b/crates/js-sys/src/futures/task/multithread.rs
@@ -100,12 +100,14 @@ impl Task {
         let closure = {
             let this = Rc::clone(&this);
 
+            // `own_assert_unwind_safe` is required because the closure captures
+            // `Rc<Task>`, which isn't `UnwindSafe`. A poll unwinding out of
+            // `run` frees the `Task` via a drop guard to avoid a leak.
             Closure::own_assert_unwind_safe(move |_| {
-                #[cfg(all(feature = "std", panic = "unwind"))]
-                this.catch_unwind_wake_and_run();
-
-                #[cfg(not(all(feature = "std", panic = "unwind")))]
-                this.wake_and_run();
+                // The promise resolution acts like a wake, so ensure the state
+                // transitions to AWAKE before entering `run`.
+                this.atomic.wake_by_ref();
+                this.run();
             })
         };
         *this.inner.borrow_mut() = Some(Inner {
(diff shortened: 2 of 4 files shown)


guybedford (commented):
> _Note: this is an AI-assisted review._

The compile fix is correct and the leak diagnosis is a good catch. One structural concern: the cleanup is installed only in the per-task wakeup closure, but a panic can also escape the **first** poll, which runs under a different driver.

### The first-poll gap

The cycle `Task -> Inner -> closure -> Rc<Task>` exists immediately after `spawn`, before any poll. The first poll runs via the queue closure (`Queue::run_all` -> `task.run()`), **not** the per-task `inner.closure`. So a panic on the first poll bypasses `catch_unwind_wake_and_run` entirely and still leaks the cycle — only subsequent wakeup polls are covered.

Polls absolutely can panic: `inner.future.as_mut().poll()` runs arbitrary user future code (`unwrap`, slice index, `assert!`, explicit `panic!`), which is the whole premise for the `catch_unwind` here.

### Suggestion: catch in `run()` instead of the closure

Every poll — first and wakeup — goes through `run()`, so catching there covers both drivers, and the same pattern drops straight into `singlethread.rs`:

```rust
pub(crate) fn run(&self) {
    #[cfg(all(feature = "std", panic = "unwind"))]
    {
        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| self.run_inner()));
        if let Err(payload) = result {
            // Drop Inner to break the Task -> Inner -> closure -> Rc<Task> cycle.
            *self.inner.borrow_mut() = None;
            std::panic::resume_unwind(payload);
        }
    }
    #[cfg(not(all(feature = "std", panic = "unwind")))]
    self.run_inner();
}

fn run_inner(&self) { /* current run() body */ }
```

The closure then goes back to the simple form (`own_assert_unwind_safe` is still required, since it captures `Rc<Task>` which isn't `UnwindSafe`):

```rust
Closure::own_assert_unwind_safe(move |_| {
    this.atomic.wake_by_ref();
    this.run();
})
```

This lets `catch_unwind_wake_and_run` / `wake_and_run` disappear and closes the first-poll gap.

### singlethread.rs

Re: "I think that the same memory leak exists in singlethread.rs too but I am not entirely sure" — it does, but **conditionally**. `singlethread::run()` likewise skips `*borrow = None` on a panicking poll, and the panic is caught at the queue closure boundary. The difference is the cycle shape:

- multithread: `Task -> Inner -> closure -> Rc<Task>` — **always** present, so it always leaks.
- singlethread: `Task -> Inner -> future -> Waker -> Rc<Task>` — only leaks **if the future retains its own waker** (common, but not guaranteed; otherwise the popped `task` local drops on unwind and the `Task` frees normally).

There's no per-task closure in singlethread, so the only sensible fix location is `run()` — which is the same reason the multithread fix is better placed there too. Applying the `run()` wrapper above to both backends gives one shared cleanup path and fixes both cases.

guybedford:
Pushed a revision to the branch (rebased onto current `main`).

The leak cleanup is now done with an RAII drop guard in `Task::run` rather than `catch_unwind`. The reason: `catch_unwind` only catches **Rust** panics, but a poll can also unwind via a **foreign JS exception** — e.g. a panic that round-trips through `console.run` in debug builds, or a JS import throwing under `panic = "unwind"`. Both kinds of unwind run destructors, so a guard that clears `Inner` on unwind covers both, and also drops the `feature = "std"` requirement. It's placed in `run` so it covers the first poll (driven by the queue) as well as wakeup polls, in both the single- and multi-threaded executors.

The compile fix (`Closure::own_assert_unwind_safe`) is unchanged.

Added a CLI regression test (`spawn_local_panic_frees_future`) that spawns a future which panics on its first poll and asserts its `DropGuard` runs — it fails without the guard and passes with it.

guybedford:
I went ahead and had Claude implement the follow-on feedback here. Will land in the next few days unless there is further feedback.

crates/cli/tests/wasm-bindgen/main.rs
```diff
@@ -1025,6 +1025,153 @@ describe('nested unreachable', () => {
...
 
+/// Regression test for a `Task` leak in the futures executor: when a spawned
+/// future panics during `poll`, the executor must drop the future (and
+/// everything it owns) rather than leaving it alive via the
+/// `Task -> Inner -> future -> Waker -> Rc<Task>` reference cycle. The panic
+/// propagates out of the microtask that drives the first poll, so the Node
+/// script installs an `uncaughtException` handler to observe it while checking
+/// that the future's `DropGuard` ran.
+#[test]
+fn spawn_local_panic_frees_future() {
```
hoodmane, line 1036: Do we need a separate test for the multithreaded case?

crates/js-sys/src/futures/task/multithread.rs
```diff
@@ -99,7 +99,11 @@ impl Task {
 
         let closure = {
             let this = Rc::clone(&this);
-            Closure::new(move |_| {
+
+            // `own_assert_unwind_safe` is required because the closure captures
+            // `Rc<Task>`, which isn't `UnwindSafe`. A poll unwinding out of
+            // `run` frees the `Task` via a drop guard to avoid a leak.
+            Closure::own_assert_unwind_safe(move |_| {
```
hoodmane, line 106: What did this do previously? Did it not compile at all with threading + panic=unwind?

crates/js-sys/src/futures/task/multithread.rs
```diff
@@ -116,13 +120,32 @@ impl Task {
...
 
     pub(crate) fn run(&self) {
+        // A poll can unwind via either a Rust panic or a foreign (JS)
+        // exception. `catch_unwind` only catches the former, but both run
+        // drops, so a drop guard covers both: if a poll unwinds we drop
+        // `Inner`, releasing the future and breaking the
+        // `Inner -> closure -> Rc<Task>` cycle that would otherwise leak the
+        // `Task`. This covers both the first poll (driven by the queue) and
+        // wakeup polls (driven by the per-task closure). The guard is disarmed
+        // (forgotten) on a normal return.
```
hoodmane: Slightly more concise comment:
```suggestion
        // A poll can unwind via either a Rust panic or a JS
        // exception. `catch_unwind` only catches the former, but both run
        // drops, so a drop guard covers both: if a poll unwinds we drop
        // `Inner`, releasing the future and breaking the
        // `Inner -> closure -> Rc<Task>` cycle that would otherwise leak the
        // `Task`. The guard is forgotten on a normal return.
```

hoodmane (approved):
I think this makes sense.

crates/cli/tests/wasm-bindgen/main.rs
guybedford: I don't think we do - the fix is entirely on the `run()` implementation which is identical in both executors.

crates/js-sys/src/futures/task/multithread.rs
guybedford: Exactly, it failed to compile.

PR #5215, opened 2026-06-24
Follow emscripten invoke_* trampolines in the descriptor interpreter

guybedford:
The descriptor interpreter executes wasm-bindgen's generated descriptor functions to recover type information, capturing the sequence of `__wbindgen_describe` calls they make. On emscripten builds with unwinding/longjmp enabled, this fell over: emscripten's exception/longjmp lowering rewrites direct calls into indirect calls through the function table, wrapped in imported `invoke_*(fnptr, ..args)` helpers. The describe helpers a descriptor function must reach are behind these trampolines, so the interpreter never saw the `__wbindgen_describe` calls and produced empty/incorrect descriptors.

Skipping the call (as we do for `__wasm_call_ctors` / `__llvm_profile_*`) isn't viable here: unlike those, `invoke_*` sits on the critical path to the describe calls.

This resolves the trampoline to its real target instead:

* Collect `env::invoke_*` imports and reconstruct the function table from active element segments at interpreter construction.
* On a call to an `invoke_*` helper, resolve the leading `fnptr` against the table and recurse into the real target, forwarding the trailing arguments.
* Evaluate the surrounding "did it throw?" guard emscripten emits around the invoke. This required generalizing the ad-hoc `branch_target` break into a `Flow` enum modelling multi-level branches, plus the comparison binops the guard uses:

```rs
enum Flow {
    Normal,
    Branch(InstrSeqId),
    Return,
}
```

so `block` / `loop` / `if`/`else` / `br` / `br_if` / `br_table` all propagate branches correctly up the evaluation stack.

Verified against real-world emscripten code. Adds interpreter smoke tests covering `invoke_*` redirection, `if`/`else`, `loop` back-edges, and `br_table`; the existing control-flow and descriptor tests continue to pass.

### DIFF
--- a/crates/cli-support/src/interpreter/mod.rs
+++ b/crates/cli-support/src/interpreter/mod.rs
@@ -71,11 +71,16 @@ pub struct Interpreter {
     skip_calls: HashSet<FunctionId>,
     stopped: bool,
 
-    // When a `Br` or `BrIf` (taken) instruction is executed, this holds the
-    // target InstrSeqId. Evaluation breaks out of nested blocks until the
-    // block whose seq matches this target is reached, at which point it is
-    // cleared and normal execution resumes after that block.
-    branch_target: Option<InstrSeqId>,
+    /// Emscripten `env::invoke_*` trampolines. Emscripten rewrites direct calls
+    /// that may unwind/longjmp into indirect calls through the function table,
+    /// wrapped in these imported helpers (`invoke_v(fnptr)`, `invoke_vi(fnptr,
+    /// a)`, ...). The first argument is the table index of the real target; the
+    /// rest are forwarded. Maps the import id to its number of result values.
+    invoke_imports: HashMap<FunctionId, usize>,
+
+    /// Function table contents (index -> function), reconstructed from active
+    /// element segments. Used to resolve the target of an `invoke_*` call.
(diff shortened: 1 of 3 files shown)


crates/cli-support/src/interpreter/mod.rs
```diff
@@ -183,6 +188,40 @@ impl Interpreter {
...
 
+        // Collect emscripten `invoke_*` trampolines so calls to them can be
+        // redirected to their real (table-indexed) targets during descriptor
+        // interpretation.
+        for import in module.imports.iter() {
+            let id = match import.kind {
+                walrus::ImportKind::Function(id) => id,
+                _ => continue,
+            };
+            if import.module == "env" && import.name.starts_with("invoke_") {
```
drcarney66, line 199: do we want to limit this check to emscripten builds? i guess it doesn't matter if it runs across non-emscripten builds that happen to define invoke_whatever

drcarney66 (approved):

crates/cli-support/src/interpreter/mod.rs
guybedford: We could, but because we control these functions ourselves, we don't risk colliding with user conventions here short of things like code coverage injection, so I'd say it's likely not worth it as you point out.

PR #5217, opened 2026-06-30
Reach wasm exports via `wasmExports['name']` in emscripten output

guybedford:
The emscripten output reached wasm exports through a local `wasm` alias (`wasm = wasmExports`, then `wasm.foo`) and `wasmExports.__wbindgen_start` (dot) in the init closure. That works at `-O0`, but emcc only tracks exports written as `wasmExports['name']` (bracket + string literal): that is the form its DCE graph roots, and the form its import/export minifier rewrites across the JS and the wasm in lockstep.

Because the alias/dot forms are invisible to emcc, under `-O3`/`-Os` (where the import/export minifier runs) emcc renamed the wasm export names without updating the JS call sites, so the init glue called stale names and the module failed at startup:

```
TypeError: wasmExports.__wbindgen_start is not a function   // factory
__wbindgen_start is not defined                              // ESM integration
```

This centralizes the export accessor and, for the emscripten mode only, emits bracket access against emscripten's own `wasmExports` object:

```js
// before
wasm = wasmExports;
const ptr0 = passStringToWasm0(name, wasm.__wbindgen_malloc, wasm.__wbindgen_realloc);
wasmExports.__wbindgen_start();

// after
const ptr0 = passStringToWasm0(name, wasmExports['__wbindgen_malloc'], wasmExports['__wbindgen_realloc']);
wasmExports['__wbindgen_start']();
```

* Add `wasm_export_ref(name)` / `wasm_export_of(id)` helpers: emscripten emits `wasmExports['name']`, every other output mode keeps `wasm.name`.
* Route all the emscripten-reachable export accesses through them — the marshalling glue (`binding.rs`), class finalizers/destructors, externref-table/vector helpers, the exn-store/closure-dtor/abort paths, and the `Exports`/`FunctionTable`/externref-init intrinsics.
* Drop the now-unused `$wasm` library variable and the `wasm = wasmExports` alias from the emscripten init closure.

Non-emscripten targets are byte-for-byte unchanged (the helper returns `wasm.name` for them). Verified end-to-end with emcc at `-O3` for the staticlib (emcc-driven) and the cargo/rustc-driven factory (`-sMODULARIZE`) flows; the clean API runs and the wasm-bindgen glue is retained and consistently renamed.

### DIFF
--- a/crates/cli-support/src/js/binding.rs
+++ b/crates/cli-support/src/js/binding.rs
@@ -946,13 +946,12 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
         let val = self.pop();
         let malloc = self.cx.export_name_of(malloc);
         let i = self.tmp();
+        let malloc = self.cx.wasm_export_ref(&malloc);
         let realloc = match realloc {
-            Some(f) => format!(", wasm.{}", self.cx.export_name_of(f)),
+            Some(f) => format!(", {}", self.cx.wasm_export_of(f)),
             None => String::new(),
         };
-        self.prelude(&format!(
-            "const ptr{i} = {pass}({val}, wasm.{malloc}{realloc});",
-        ));
+        self.prelude(&format!("const ptr{i} = {pass}({val}, {malloc}{realloc});",));
         self.prelude(&format!("const len{i} = WASM_VECTOR_LEN;"));
         self.push(format!("ptr{i}"));
         self.push(format!("len{i}"));
@@ -1200,10 +1199,9 @@ fn instruction(
 
         Instruction::Retptr { size } => {
             js.cx.inject_stack_pointer_shim()?;
-            js.prelude(&format!(
-                "const retptr = wasm.__wbindgen_add_to_stack_pointer(-{size});"
-            ));
-            js.finally(&format!("wasm.__wbindgen_add_to_stack_pointer({size});"));
+            let sp = js.cx.wasm_export_ref("__wbindgen_add_to_stack_pointer");
+            js.prelude(&format!("const retptr = {sp}(-{size});"));
+            js.finally(&format!("{sp}({size});"));
             js.stack.push("retptr".to_string());
         }
 
@@ -1460,9 +1458,9 @@ fn instruction(
         Instruction::VectorToMemory { kind, malloc, mem } => {
             let val = js.pop();
             let func = js.cx.pass_to_wasm_function(kind.clone(), *mem);
-            let malloc = js.cx.export_name_of(*malloc);
+            let malloc = js.cx.wasm_export_of(*malloc);
             let i = js.tmp();
-            js.prelude(&format!("const ptr{i} = {func}({val}, wasm.{malloc});",));
+            js.prelude(&format!("const ptr{i} = {func}({val}, {malloc});",));
             js.prelude(&format!("const len{i} = WASM_VECTOR_LEN;"));
             js.push(format!("ptr{i}"));
             js.push(format!("len{i}"));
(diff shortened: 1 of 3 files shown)


crates/cli-support/src/js/binding.rs
```diff
@@ -946,13 +946,12 @@ impl<'a, 'b> JsBuilder<'a, 'b> {
         let val = self.pop();
         let malloc = self.cx.export_name_of(malloc);
         let i = self.tmp();
+        let malloc = self.cx.wasm_export_ref(&malloc);
         let realloc = match realloc {
```
hoodmane: Not really related to this PR but why wouldn't realloc be exported? Seems like it might be an improvement to always export it.

crates/cli-support/src/js/binding.rs
```diff
@@ -1200,10 +1199,9 @@ fn instruction(
 
         Instruction::Retptr { size } => {
             js.cx.inject_stack_pointer_shim()?;
-            js.prelude(&format!(
-                "const retptr = wasm.__wbindgen_add_to_stack_pointer(-{size});"
-            ));
-            js.finally(&format!("wasm.__wbindgen_add_to_stack_pointer({size});"));
+            let sp = js.cx.wasm_export_ref("__wbindgen_add_to_stack_pointer");
```
hoodmane, line 1199: Is this emscripten-only code? Is there a reason we rename things?

crates/cli-support/src/js/binding.rs
```diff
@@ -1530,14 +1528,14 @@ fn instruction(
...
             js.cx.expose_is_like_none();
             let i = js.tmp();
-            let malloc = js.cx.export_name_of(*malloc);
+            let malloc = js.cx.wasm_export_of(*malloc);
             let val = js.pop();
             let realloc = match realloc {
-                Some(f) => format!(", wasm.{}", js.cx.export_name_of(*f)),
+                Some(f) => format!(", {}", js.cx.wasm_export_of(*f)),
                 None => String::new(),
             };
```
hoodmane: Substantial repetition here, maybe we should factor this pattern out?

crates/cli-support/src/js/binding.rs
guybedford: Good point, and I agree. Posted https://github.com/wasm-bindgen/wasm-bindgen/issues/5218.

crates/cli-support/src/js/binding.rs
guybedford: Emscripten's internal compiler expects names like `wasmExports['name']`, while Wasm Bindgen normally outputs `wasm.name`. This just implements that Emscripten-specific branching.

crates/cli-support/src/js/binding.rs
guybedford: Refactored.

crates/cli-support/src/js/binding.rs
hoodmane: I mean `__wbindgen_add_to_stack_pointer` instead of `stackAlloc`?

crates/cli-support/src/js/mod.rs
```diff
@@ -6826,6 +6832,33 @@ addToLibrary({
...
+    }
+
+    fn wasm_export_of(&mut self, id: impl Into<walrus::ExportItem>) -> String {
+        let name = self.export_name_of(id);
+        self.wasm_export_ref(&name)
+    }
+
+    /// JS snippet for an optional trailing function argument, either
+    /// `", <export ref>"` or empty when absent.
+    fn optional_func_arg(&mut self, func: Option<walrus::FunctionId>) -> String {
```
hoodmane, line 6855: Much nicer factored out.

hoodmane (approved):

crates/cli-support/src/js/binding.rs
guybedford: `stackAlloc` is Emscripten-only, we could unify on that for the Emscripten target but it has a different contract to our function which just takes a signed int.

PR #5220, opened 2026-07-01
Drop the emscripten marker static from wasm exports

guybedford:
Emscripten detection relies on a marker static that wasm-bindgen defines in the
guest crate:

```rs
#[used]
#[link_section = "__wasm_bindgen_emscripten_marker"]
static __WASM_BINDGEN_EMSCRIPTEN_MARKER: [u8; 1] = [1];
```

The `#[used]` attribute keeps the static alive through linking so that
`cli-support` can detect emscripten output from its custom section. But
`#[used]` also leaves it behind as a wasm export that nothing references. Under
emscripten's `EMSCRIPTEN_KEEPALIVE` heuristic that stray export gets surfaced as
public API, leaking an internal detection detail into the generated bindings.

This drops the marker export in `cli-support` right after we consume the section
to switch into emscripten mode, so the detection still works but the export no
longer escapes into the public surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,6 @@
   call sites, e.g. `__wbindgen_start is not defined`).
 
 * The emscripten detection marker static is no longer leaked as public API.
-  It is `#[used]` so its custom section survives linking for detection, but that
-  also left it as a stray wasm export that emscripten's `EMSCRIPTEN_KEEPALIVE`
-  heuristic surfaced as public API. `cli-support` now drops the export once it
-  has consumed the section to switch into emscripten mode.
   [#5220](https://github.com/wasm-bindgen/wasm-bindgen/pull/5220)
 
 ### Removed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,13 @@
   `-O3`/`-Os` (previously the export names were minified without updating the JS
   call sites, e.g. `__wbindgen_start is not defined`).
 
+* The emscripten detection marker static is no longer leaked as public API.
+  It is `#[used]` so its custom section survives linking for detection, but that
+  also left it as a stray wasm export that emscripten's `EMSCRIPTEN_KEEPALIVE`
+  heuristic surfaced as public API. `cli-support` now drops the export once it
+  has consumed the section to switch into emscripten mode.
```
hoodmane:
```suggestion
* The emscripten detection marker static is no longer leaked as public API.
```

hoodmane (approved):

Spxg:
I believe that as long as we remove `#[used]`, the static variable won’t be exported, while `custom_section` will still take effect. We can also remove `pub`.

PR #5224, opened 2026-07-04
Mention tarpaulin in the docs

xd009642:
### Description

As of the release 0.37.0, cargo-tarpaulin supports coverage of wasm projects the same as cargo-llvm-cov. As I have gotten issues and comments before about wasm coverage I figured it would be a useful addition to the docs for tarpaulin users who want wasm coverage and don't want to switch to cargo-llvm-cov (or can't, cargo-llvm-cov and cargo-tarpaulin have overlapping but disjoint feature sets).

FYI Currently, this version is on crates.io with github releases uploading.

### Checklist

I'm going to assume this doesn't need a changelog entry.


Spxg (approved):
Thanks

PR #5225, opened 2026-07-06
Reserve thread memory after `memory.initial`

### LINKED ISSUE #5223: Threads bootstrap stalls with large imported shared initial memory
### Problem

With wasm-bindgen threads, imported shared memory, and a large linker-provided initial memory, worker startup can stall at runtime.

In the repro, the browser creates the main worker and the Tokio worker blobs, but the async tasks are never scheduled. The following logs are visible in the browser console.

Observed with `--initial-memory=67108864`:

```text
start task in thread ThreadId(1)
```

Expected:

```text
return from tokio runtime: world d d d d
```

### Reproduction

https://github.com/utooland/tokio-on-browser/tree/repro/wasm-bindgen-initial-memory-tokio-stall

The repro uses wasm-bindgen threads with imported shared memory. The failing configuration includes:

```text
-Clink-arg=--shared-memory
-Clink-arg=--import-memory
-Clink-arg=--initial-memory=67108864
-Clink-arg=--max-memory=4294967296
```

It reproduces with wasm-bindgen `0.2.126`.

### Findings

Bisecting Rust nightlies with the same repro and wasm-bindgen setup found:

```text
nightly-2026-05-05  rustc cb40c25f6  good  std dlmalloc 0.2.11
nightly-2026-05-06  rustc e95e73209  bad   std dlmalloc 0.2.13
```

The relevant Rust std change is that dlmalloc `0.2.13` starts using the linker-provided preexisting heap range, `__heap_base..__heap_end`, before falling back to `memory.grow`. That behavior is valid allocator behavior.

The problem appears to be that wasm-bindgen's threads transform places its thread bookkeeping / temporary stack near the original `__heap_base`, assuming that range will remain unused. With the newer std allocator and a large initial memory, that range can become allocator-visible, so the worker bootstrap state can be overwritten.

Conclusion: wasm-bindgen's threads bootstrap layout should reserve its injected bookkeeping / temporary stack in memory that cannot be reused by the wasm runtime allocator, instead of relying on the original `__heap_base` area remaining unused.

Spxg:
`dlmalloc` introduced a preexisting chunk in 0.2.13, whose range is `[__heap_base, __heap_end)`:

```rust
extern "C" {
    static __heap_base: u8;
    static __heap_end: u8;
}
```

These values are fixed during linking. However, our thread transformation reserves one page starting from `__heap_base`, so this range overlaps with the preexisting chunk.

Therefore, we need to reserve space after `__heap_end`. Using `memory.initial` instead of exporting `__heap_end` might be a good choice, and it would not introduce a breaking change.

resolves #5223


Spxg:
I think we may currently lack real thread tests, so I’m not sure whether this approach is viable. It probably needs some discussion. 
// cc @daxpedda @guybedford @RReverser

crates/cli-support/src/transforms/threads/mod.rs
```diff
@@ -59,6 +59,11 @@ pub fn run(module: &mut Module) -> Result<Option<ThreadCount>, Error> {
 
     let memory = wasm_conventions::get_memory(module)?;
 
+    let mem = module.memories.get_mut(memory);
+    assert!(mem.shared);
```
drewcrawford, line 63: Now that the __heap_base lookup is gone, these asserts are the first thing a misconfigured module hits. If you build with +atomics -Clink-args=--shared-memory but forget --import-memory, wasm-ld emits a defined (non-imported) shared memory, and current toolchains don't export __heap_base by default — so before this PR that case exited with the clean failed to find __heap_base for injecting thread id error, and after it you get a raw assertion failed: mem.import.is_some() panic with a backtrace.

Since these check user input rather than internal invariants, could the import and data-segment checks become bail!s with actionable messages (e.g. "threading requires --import-memory")? The mem.shared one is already guaranteed by is_enabled() above and could just be dropped.

examples/raytrace-parallel/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"scripts": {
-		"build": "wasm-pack build --target no-modules --out-dir ../dist/raytrace-parallel . -Zbuild-std=std,panic_abort --config \"target.wasm32-unknown-unknown.rustflags='-Ctarget-feature=+atomics -Clink-args=--shared-memory -Clink-args=--max-memory=1073741824 -Clink-args=--import-memory -Clink-args=--export=__heap_base -Clink-args=--export=__wasm_init_tls -Clink-args=--export=__tls_size -Clink-args=--export=__tls_align -Clink-args=--export=__tls_base'\"",
+		"build": "wasm-pack build --target no-modules --out-dir ../dist/raytrace-parallel . -Zbuild-std=std,panic_abort --config \"target.wasm32-unknown-unknown.rustflags='-Ctarget-feature=+atomics -Clink-args=--shared-memory -Clink-args=--max-memory=1073741824 -Clink-args=--import-memory -Clink-args=--export=__wasm_init_tls -Clink-args=--export=__tls_size -Clink-args=--export=__tls_align -Clink-args=--export=__tls_base'\"",
```
drewcrawford: Dropping --export=__heap_base from the examples creates a breakage window: the workspace is versioned 0.2.126, which is already published with the old transform that hard-errors without the export. These scripts build via wasm-pack, which resolves the CLI by crate version — i.e. the published 0.2.126 binary — so npm run build on main fails with failed to find __heap_base for injecting thread id until the next release ships. (CI doesn't catch it because it does cargo install --path crates/cli first.)

The new transform just ignores the export, so keeping the flag is free. Maybe keep it here, in nodejs-threads, and in wasm-audio-worklet for now and drop it in a follow-up after the release?

guybedford:
I posted an alternative in https://github.com/wasm-bindgen/wasm-bindgen/pull/5226, which rewrites the dlmalloc inlined value to our bumped value. May be more brittle, but conceptually it's nice to retain contiguous allocator memory ownership too... not sure how to think about the tradeoffs on this, what do you think @Spxg?

Spxg:
> May be more brittle, but conceptually it's nice to retain contiguous allocator memory ownership too

This kind of "contiguous" has almost no impact, because after the allocator exhausts the preexisting chunks, it simply calls `memory.grow`; it doesn’t care whether there are other reserved blocks after `__heap_end`.

I don’t think we should reserve memory starting from `__heap_base`, for a few reasons:

1. We’re assuming that `memory.initial` is exactly `__heap_end`, but that assumption doesn’t necessarily hold. If another toolchain has already reserved a page after `memory.initial`, blindly increasing `__heap_end` could overwrite it.
2. Rewriting the `__heap_base` and `__heap_end` values used by `dlmalloc` or other allocators is risky. Even though we currently only rewrite them when both appear in the same function, we may still end up changing other legitimate values. Also, if some code only references one of them, we won’t be able to rewrite it, which could make the program behave incorrectly.

crates/cli-support/src/transforms/threads/mod.rs
Spxg: Keeping the `mem.shared` assertion doesn’t seem like a bad idea to me, but adding a user-friendly `--import-memory` error is definitely worthwhile. Thanks

examples/raytrace-parallel/package.json
Spxg: Good catch

Spxg: Do we still publish these examples now? Are they built after publishing a new `wasm-bindgen` version, or are they installed first and then built?

Because if each publish uses the old version of `wasm-bindgen`, I think it will consistently fail when the schema changes.

// cc @guybedford

daxpedda:
I agree that we shouldn't rely on any allocator specific behavior here.

In the future we should allocate the necessary stack space on the spawning thread instead of the spawned thread. That would eliminate the temporary stack space we need to allocate in the first place.

examples/raytrace-parallel/package.json
guybedford: The window here will be quite small where the PR is landed and the next release goes out. If we weren't already planning the next release soon I'd be more concerned about this breaking scenario, but given it will likely only be a day or so I think it's fine to land as-is.

guybedford (approved):
@Spxg sure, that makes sense. It does seem like the allocators handle this scenario fine, and if any issues come up in future we can always reconsider.

examples/raytrace-parallel/package.json
Spxg: Thanks for the explain. To ensure stability, I’ve decided to revert these changes for now and make the changes after the release.

PR #5230, opened 2026-07-14
Interpreter-based per-monomorphisation generic imports

logan-gatlin:
# Summary

**Core:** `#[wasm_bindgen(experimental_generic_monomorphization)]` — opt-in per-monomorphisation binding for generic imports (vs. the default `JsValue`-erasure path). Reuses/generalizes the `wbg_cast` discovery mechanism: descriptor interpreter finds the monomorphised marker call, call site gets rewritten to a manufactured import. `GenericImportKey` now covers both callers; marker renamed `__wbindgen_describe_cast` → `__wbindgen_describe_generic_import`, shims `__wbindgen_cast_N` → `__wbindgen_generic_N`. Cast descriptor wire format gains a leading `inform(0)` (zero-length shim key) to mark the degenerate case.

Supported:
- bounds / `where` clauses / HRTB
- associated-type projections
- `async`
- `catch`
- `slice_to_array`
- lifetimes
- block-level inheritance (whole `extern "C"` block opts in)

Rejected:
- const generics
- type-level generics on imported types
- `&mut T`, or `&T` nested inside another type (e.g. `Option<&T>`)
- returning `&_`
- `T` or `&T` as the `variadic` argument
- generic `catch` error type
- generic-element `slice_to_array`
- `reexport`
- `assert_no_shim`
- non-ident argument patterns

Shim-key collisions (two imports differing only in un-hashed attrs like `catch`) are a CLI-side link-time error, unavoidable earlier.

**`SCHEMA_VERSION` bump** — new macro-emitted field (`generic_per_mono`) + cast wire change means old CLI/new runtime (or vice versa) must fail loudly, not silently misbind.

## Future direction
Unify the existing erased generics with per-monomorphization generics, have wasm-bindgen choose between the two appropriately.


guybedford (approved):
This is looking excellent — the discovery-pipeline generalization is clean, and the diagnostics, determinism, and test work are thorough. I did a deep review including some empirical probing of the design limits (argument-position `impl Trait`, class-level generics, and bulk-converting js-sys generic imports as a stress experiment) — the architecture held up in every direction I pushed on it, which gives me a lot of confidence in this as the foundation for the unification direction you mention.

I've pushed a commit adding coverage for two spots the suite didn't pin: one bare-`T` import instantiated across `&str`/`String`/`u32` (the borrowed-vs-owned string wire distinction — no-free vs `__wbindgen_free` — was previously unpinned), and generic-element `&[T]` as both a plain and a variadic argument, which works today but had no coverage.

Requests:

* Add an experimental compiler warning on the attribute, following the JSPI precedent in #5193, so the feature can ship now and stabilize later. Given the supported-shape set is expected to grow, the signal seems at least as valuable here.
* Replace the `Debug`-rendering comparison in `GenericImportMeta::identity()` with derived `PartialEq`. The payload chain (`AuxValue`, `JsImport`, `JsImportName`) is plain data, so it's only a handful of derive lines, and it makes the comparison structural rather than representational — the current form silently weakens if `AuxImport` ever gains a manual `Debug` impl.
* Fix the diagnostic for `impl Trait` in argument position: it currently trips "generic_per_mono requires at least one type parameter", which is wrong-footed — there is one, it's just anonymous. Desugaring APT args to named type parameters in the parser looks like a small, self-contained follow-on; I'll file an issue with details.

Side note on naming: since this plausibly becomes the default generics path eventually, `generic_monomorphization` might read better long-term than `generic_per_mono`. Not a strong opinion, but worth settling before it ships.

guybedford:
It looks like all feedback is addressed now apart from the experimental warning.

Since we do want to test this in workers-rs, rather than an experimental warning let's rename the attribute to `experimental_generic_monomorphization` so it's clear in usage rather than by warning (and since we plan to move to enable it by default, we wouldn't need the flag once that happens).

logan-gatlin:
Changed the name to `experimental_generic_monomorphization`

guybedford:
Since we support `foo: T` across `T` being `&str` or `u32` etc. fine, I've dropped the `IntoWasmAbi` for `&scalar` commits here as that is a major outward change, to instead land that separately if we deem we still want it in future.

PR #5238, opened 2026-07-24
Fix catch imports returning i64 trapping on the exception path

guybedford:
This fixes `#[wasm_bindgen(catch)]` imports returning `Result<i64, JsValue>` / `Result<u64, JsValue>` failing when the JS import throws.

The `handleError` wrapper stores the caught exception and returns `undefined`, which cannot be converted to a Wasm `i64` at the JS -> Wasm boundary (`ToBigInt(undefined)` throws a `TypeError`). Worse than the plain `TypeError`, this conversion error then propagates out into any outer `handleError` (e.g. the test harness invoke), which calls `__wbindgen_exn_store` while the exception slot is still occupied, panicking on the `GLOBAL_EXNDATA` assertion. Other return types are unaffected since `undefined` coerces without throwing for `i32`, `f32`/`f64` and `externref`.

The fix substitutes a `0n` placeholder on the exception path for catch imports whose core signature returns `i64`:

```js
export function __wbg_foo_86c7ea0e4ba8cf8a() { return handleError(function () {
    const ret = foo();
    return ret;
}, arguments) ?? 0n; }
```

The placeholder is discarded, since the generated Rust checks the stored exception before using the return value. Generated output for all other imports is byte-identical.

Test coverage: a new `imports::exn_caught_i64` wasm test covering both `i64` and `u64`, and an `import-catch-i64` reference test locking in the generated glue.

_Made with AI assistance under my review_


Spxg (approved):
I’m not very familiar with JavaScript, so this needs another member to take a look.

guybedford:
@hoodmane would be great to get your approval here as well.

hoodmane:
I'll take a look.

hoodmane:
There's a conflict in the CHANGELOG.

hoodmane (approved):

PR #5243, opened 2026-07-29
cli-support: Add Emscripten shared WebAssembly.Memory fallback expression

ZacCranko:
### Description
When `OutputMode::Emscripten` is active and shared memory is enabled, `wasm_import_definitions` inserts an `ImportDefinition::Expression` for `init_memory` (`new WebAssembly.Memory(...)`).

Previously, this branch defaulted to unconditionally inserting `memory || `, which produces invalid runtime JS bindings under Emscripten (`ReferenceError: memory is not defined`).

This PR updates `crates/cli-support/src/js/mod.rs` to check if `OutputMode::Emscripten` is active:
- If Emscripten, it wraps the memory initialization expression with `(typeof memory !== 'undefined' ? memory : (typeof wasmMemory !== 'undefined' ? wasmMemory : {expr}))`.
- Otherwise, it falls through to the existing `memory || ` prefix.

---
*P.S. Apologies for shamelessly generating this PR with AI! Thank you so much for reviewing it.*


ZacCranko:
cc @walkingeyerobot

guybedford:
Thanks for posting this - I believe the fix is even simpler than this though. Rather than a review cycle I had Claude post up https://github.com/wasm-bindgen/wasm-bindgen/pull/5254 instead for it. Please let me know if that can work for you here instead.

PR #5244, opened 2026-07-31
Restore the stack pointer when an exception leaves an export

### LINKED ISSUE #5245: `panic=unwind`: escaping panics leak the export's stack frame
Under `panic=unwind`, a panic that escapes an exported function never rewinds `__stack_pointer`,
the frame is leaked. Enough panics exhaust the shadow stack and cause a Wasm trap. In Node, the
exhausting call and subsequent calls throw `RuntimeError: memory access out of bounds`; workerd can
poison the instance after the fatal trap and report `Module terminated` on subsequent calls.

### Repro

https://github.com/BernardoPe/wasm-bindgen-panic-unwind-stack-leak-repro

```rust
use std::hint::black_box;
use wasm_bindgen::prelude::*;

#[wasm_bindgen(js_name = "probe")]
pub fn probe() -> String {
    let held = String::from("freed only when unwinding");
    if held.len() > 1 { panic!("probe panic"); }
    held
}

// Same thing with a 4 KB array live across the panic.
#[wasm_bindgen(js_name = "probeBig")]
pub fn probe_big() -> String {
    let mut buf = [0u8; 4096];
    black_box(&mut buf);
    if black_box(buf[0]) == 0 { panic!("big frame panic"); }
    String::from("no")
}

#[wasm_bindgen(js_name = "ok")]
pub fn ok() -> String { String::from("alive") }
```

Exporting `__stack_pointer` lets you watch the leak directly instead of waiting for the crash:

```bash
RUSTFLAGS="-Cpanic=unwind -Clink-arg=--export=__stack_pointer" \
  cargo +nightly build --release --target wasm32-unknown-unknown \
  -Zbuild-std=std,panic_unwind
wasm-bindgen --target web --reference-types --out-dir out --out-name m \
  target/wasm32-unknown-unknown/release/repro.wasm
```

```js
import { readFileSync } from "node:fs";
const mod = await import("./out/m.js");
const ex = mod.initSync({ module: new WebAssembly.Module(readFileSync("./out/m_bg.wasm")) });

const sp = () => ex.__stack_pointer.value;
const before = sp();
for (let i = 0; i < 200; i++) { try { mod.probe(); } catch {} }
console.log((before - sp()) / 200, "bytes leaked per call");
```

### What you see

| export | leaked per call | shadow stack exhausted after |
|---|---|---|
| `probe` | 32 B | 32,768 panics |
| `probeBig` | 4,128 B | 254 panics |
| `ok` (never panics) | 0 B | never |

`probeBig` leaks `probe`'s 32 bytes plus its own 4 KB array, so what is lost is exactly one stack
frame per escaping panic.

BernardoPe:
Fixes #5245

Repeated escaping panics eventually exhaust the shadow stack. In Node, the exhausting call and subsequent calls throw `RuntimeError: memory access out of bounds`; workerd can poison the instance after the fatal trap and report `Module terminated` on subsequent calls.

This adds a `cli-support` transform that wraps the exports an exception may leave as a supported egress. Each wrapper saves `__stack_pointer` on entry, calls the original export inside a try block, then restores the saved value before rethrowing an exception.

The transform supports both exception-handling encodings:

- Modern EH uses `try_table`, `catch_all_ref`, and `throw_ref`.
- Legacy EH uses `try`, `catch_all`, and `rethrow`.

The transform runs after `catch_handler`, while sharing its original EH detection. Detecting again afterward would mistake `catch_handler`'s generated `try_table` instructions for a `panic = "unwind"` module.

The wrapped set is exactly the macro's `extern "C-unwind"` surface: the adapter-backed exports in `wit.exports` (export shims, getters/setters, closure invoke shims), plus the name-resolved class exports `__wbg_*_free` and `__wbg_upcast_*`, which are `C-unwind` but have no adapter -- `free` panics recoverably when the value is still borrowed, so it can unwind repeatedly against a live instance.

Runtime intrinsics such as `__wbindgen_malloc` and the closure dtor are deliberately left unwrapped: they are plain `extern "C"`, so an escaping panic there is already an abort ("panic in a function that cannot unwind") -- the same semantics as a panic during panic unwinding -- and aborts are traps, which no `catch_all` observes. This also keeps the generated JavaScript's `FunctionId`-based intrinsic resolution working. Imported functions, modules without a mutable stack pointer, `panic = "abort"` builds, and Emscripten output are also left unchanged.

### Checklist

_Made with AI assistance_


guybedford (approved):
Thanks for posting this one, I've extended it with two small edge cases for free and upcast calls as well.

PR #5246, opened 2026-08-02
Update WebGPU API to 2026-08-02

beicause:
### Description
Updates the WebGPU bindings to the latest IDL (2026-08-02) at https://gpuweb.github.io/gpuweb/#idl-index

### Checklist

### DIFF
--- a/crates/web-sys/src/features/gen_GpuCommandEncoder.rs
+++ b/crates/web-sys/src/features/gen_GpuCommandEncoder.rs
@@ -212,7 +212,72 @@ extern "C" {
     #[doc = ""]
     #[doc = "*This API is unstable and requires `--cfg=web_sys_unstable_apis` to be activated, as"]
     #[doc = "[described in the `wasm-bindgen` guide](https://wasm-bindgen.github.io/wasm-bindgen/web-sys/unstable-apis.html)*"]
-    pub fn copy_buffer_to_buffer_with_u32_and_u32(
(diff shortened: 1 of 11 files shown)


guybedford (approved):
Thanks for posting, this looks great.

To avoid the churn of the `copy_byffer_to_buffer_with_gpu_buffer_and_i32_...` I added a new simplification pass guarded to `WbgGeneric` in the WebIDL generation so we minimize the diff on that.

Will land tomorrow.

crates/web-sys/src/features/gen_GpuCommandEncoder.rs
beicause, line 309: >To avoid the churn of the copy_byffer_to_buffer_with_gpu_buffer_and_i32_... I added a new simplification pass guarded to WbgGeneric in the WebIDL generation so we minimize the diff on that

I'm not sure why we need to treat it specially. For example, `copy_buffer_to_buffer_with_f64_and_u32` - the actual parameter order is `f64_buffer_source_u32`. In my opinion, it's fine for the naming to be consistent with the parameter order.

guybedford: `with_` variations never include duplicate leading arguments, so this effectively extends the existing eliding to also include subsequent arguments that are the same across all overloads. Yes it does mean that reading a `_with_...` function name now requires not just skipping leading constant args, but also interior ones. But overall I think it does provide greater clarity to the functions? Happy to be convinced otherwise yet.

PR #5248, opened 2026-08-04
Count `char`s, not bytes, in descriptor length prefixes

logan-gatlin:
This fixes descriptor length prefixes for non-ASCII strings.

The descriptor wire format emits one `u32` word per `char`, prefixed by a length, but that length was computed with `str::len()`, which counts UTF-8 bytes. Any non-ASCII string (e.g. a `js_name` or `typescript_type` containing accented characters) made the prefix exceed the number of words actually written, leaving the CLI's decoder consuming subsequent descriptor words as characters — either panicking the build or silently mis-binding everything after the affected item.

* All five char-stream emit sites in `macro-support` codegen (`Struct`, `ImportType`/`typescript_type`, `StringEnum`, `DynamicUnion`, `Enum`) now use `chars().count()` for the length prefix.

Test coverage: the `typescript-type` reference test now includes a non-ASCII `typescript_type`, a non-ASCII string enum `js_name`, and a non-ASCII C-style enum `js_name`, which fail against the previous codegen.


guybedford (commented):
The fix looks right — I verified locally that the reference test fails against the old codegen and passes with this change, and that these five sites are the only char-stream emitters (the remaining string `.len()` uses in codegen are byte-oriented custom-section encodings). A few nits below.

crates/cli/tests/reference/typescript-type.rs
```diff
@@ -4,10 +4,48 @@ use wasm_bindgen::prelude::*;
 extern "C" {
     #[wasm_bindgen(typescript_type = "number | string")]
     type CustomType;
+
+    // A non-ASCII `typescript_type`. The descriptor wire format writes one word
+    // per `char` and prefixes it with a length, so the length must be a `char`
+    // count; a UTF-8 byte count would leave the decoder reading descriptor
+    // payload words as characters and silently mis-bind everything after it.
```
guybedford: This explanation is repeated in the CHANGELOG, here, and twice more below — one short line here (e.g. `// Non-ASCII typescript_type: descriptor length prefixes must count chars, not bytes.`) would cover it, and the "second/third path" comments below can be dropped.

crates/cli/tests/reference/typescript-type.rs
```diff
@@ -4,10 +4,48 @@ use wasm_bindgen::prelude::*;
...
+    Klein = 1,
+    Gross = 2,
+}
+
+#[wasm_bindgen]
+pub fn take_groesse(g: Groesse) -> Groesse {
+    g
+}
+
+
```
guybedford: Trailing blank lines — trim to a single final newline.

crates/macro-support/src/codegen.rs
```diff
@@ -1112,7 +1112,10 @@ impl TryToTokens for ast::ImportType {
         };
 
         let description = if let Some(typescript_type) = &self.typescript_type {
-            let typescript_type_len = typescript_type.len() as u32;
+            // One descriptor word per `char`, so the prefix counts `char`s, not
+            // UTF-8 bytes. `typescript_type` is an arbitrary user string, so a
+            // byte count silently mis-binds for anything non-ASCII.
+            let typescript_type_len = typescript_type.chars().count() as u32;
```
guybedford: Could be a single line, e.g. `// One descriptor word per char, so count chars, not bytes.` — same for the similar comment on the StringEnum site.

CHANGELOG.md
```diff
@@ -9,6 +9,14 @@
...
 ### Fixed
 
+* Length-prefixed strings in the descriptor stream now count `char`s rather than
+  UTF-8 bytes. The descriptor wire format emits one word per `char` and prefixes
+  it with a length, but that length was computed with `str::len()`, so any
+  non-ASCII string made the prefix exceed the number of words actually written
+  and left the CLI's decoder consuming descriptor *payload* words as characters.
+  Depending on what followed, that either aborted the build with an opaque
+  decoder panic or silently mis-bound everything after the affected item.
+
```
guybedford: This is quite long relative to other entries — something like:

```md
* Fixed length prefixes in descriptor strings to count `char`s rather than UTF-8
  bytes, so non-ASCII names in `js_name`/`typescript_type` no longer panic the
  CLI or mis-bind the generated bindings.
```

guybedford (approved):

PR #5249, opened 2026-08-04
Describe `async` imports as returning a `Promise`

logan-gatlin:
This fixes `async` imports whose resolved type is not a JS handle, e.g. `async fn f() -> u32;`, which have silently produced garbage values since 0.2.109.

The erasable generics work (#4876) widened `async` imports to resolve to arbitrary types via typed promises, but the import descriptor was not updated: it still named the *resolved* type rather than the `Promise` handle that actually crosses the ABI. The CLI then marshalled the promise handle as if it were the resolved type — `async fn f() -> u32;` read the handle as the integer, and `async fn f() -> String;` handed the `Promise` object to `passStringToWasm0`. Imports resolving to JS handle types (`JsValue`, `JsString`, custom imported types) marshal identically either way, which is why the existing test suite did not catch it.

The fix is one branch in `DescribeImport`: an `async` import always describes its return as an externref, matching the value-side codegen which already converts the resolved value separately when the returned `JsFuture` is awaited.

Test coverage:

* `crates/cli/tests/reference/async-import.rs` pins the emitted glue: every async import shows `(result externref)` in the `.wat` and the JS shim forwards the promise, across scalar, string, handle, unit and `catch` returns
* `tests/wasm/async_import.rs` adds runtime assertions on concrete resolved values (scalars, string, `Option`, `catch`, and the argument direction), all of which fail against the previous descriptor


crates/cli/tests/reference/async-import.rs
```diff
@@ -0,0 +1,50 @@
+// Pins the descriptor/binding shape of `async` *imports* on the ordinary
+// (non-`generic_per_mono`) path.
```
guybedford: `generic_per_mono` isn't a term that exists anywhere in this codebase — just say the non-generic path. This header also retells the whole story that already lives at the fix site in codegen.rs; a couple of lines pointing there is enough.

crates/macro-support/src/codegen.rs
```diff
@@ -2880,18 +2880,26 @@ impl TryToTokens for DescribeImport<'_> {
...
-            Some(ref t) => {
-                let t = generics::generic_to_concrete(
-                    t.clone(),
-                    &fn_class_generics.concrete_defaults,
-                    &fn_lifetime_params,
-                )?;
-                quote! { <#t as WasmDescribe>::describe(); }
+        // An `async` import's declared return type never reaches the ABI — a
+        // `Promise` handle does — so it is not concretised or described at all.
+        let concrete_ret = match (&f.js_ret, f.function.r#async) {
```
guybedford: `concrete_ret` is always `None` in the async case, so the `(_, true)` arm below re-checks a condition already folded in here. A plain conditional would be more direct: if async describe `JsValue`, else if `Some(t)` concretise and describe, else `()`.

CHANGELOG.md
```diff
@@ -9,6 +9,17 @@
 
 ### Fixed
 
+* `async` imports now describe their return value as the `Promise` handle that
```
guybedford: Quite long relative to the neighbouring entries — a couple of lines covering what was broken, since when (0.2.109, #4876) and what it does now is enough.

tests/wasm/async_import.rs
```diff
@@ -0,0 +1,77 @@
...
+//!
+//! The reference test `crates/cli/tests/reference/async-import.rs` pins the
+//! *shape* of the emitted glue; this pins the values that actually cross.
+
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_test::*;
+
+#[wasm_bindgen(module = "tests/wasm/async_import.js")]
+extern "C" {
+    async fn asyncU32() -> u32;
```
guybedford: Nit: these use raw camelCase Rust names while `async_opt_some` below uses `js_name` — pick one style for the block.

guybedford (commented):
I verified this against main: cherry-picking just the new tests onto main's macro code fails with the expected assertion failures, and passes with the fix. The descriptor/ABI mismatch dates to the erasable generics work (#4876), so this is a released regression affecting 0.2.109+ — please add a PR description saying as much, it's currently empty.

The fix itself is right and minimally scoped: the value side already treats the async return as a typed `Promise<T>` externref handle, and `inner_ret` is unused for import adapters, so describing `JsValue` aligns the descriptor with the ABI and matches the previously-correct no-return case.

A few comments below, mostly around comment volume — the mechanism explanation is currently repeated across the reference test, runtime test, JS fixture and changelog. It belongs once, at the fix site in codegen.rs, with the rest trimmed down.

guybedford (approved):

PR #5250, opened 2026-08-04
Include `js_namespace` in imported function and static shim names

logan-gatlin:
This fixes a silent-miscompile class of bug where two imports with identical Rust signatures that differed only in their `js_namespace` hashed to the same `__wbg_<name>_<hash>` shim symbol, so they were treated as one binding and one of the two call sites silently invoked the wrong JS value. It is the import-side counterpart of #5154.

```rust
mod a { #[wasm_bindgen(js_namespace = ["alpha"])] extern "C" { pub fn log(s: &str); } }
mod b { #[wasm_bindgen(js_namespace = ["beta"])]  extern "C" { pub fn log(s: &str); } }
// previously: one shim, so both `a::log` and `b::log` reached the same object
```

What is implemented:

* The *resolved* `js_namespace` (item-level, or inherited from the enclosing `extern "C"` block) is now part of the shim hash for imported functions.
* Imported statics had the identical collision (`__wbg_static_accessor_*` hashed only `(js_name, module, ident)`), and are fixed the same way.
* Hashing is structural — `ShortHash((data, namespace))` — rather than a hand-rolled separator encoding, so segment contents cannot alias the encoding. It is gated on `Option::None` so shim names for all imports without a namespace are unchanged.

The `js-namespace-shim-key` reference test covers each pair with identical Rust signatures under: block-level namespaces, namespace vs no namespace, item-level namespaces, item-level overriding block-level, nested `["a", "b"]` vs `["a"]`, and namespaced statics. Reference outputs for existing namespaced imports are re-blessed since their shim hashes move; shim names are an internal contract between the macro and CLI of the same version, so this needs no downstream action.


CHANGELOG.md
```diff
@@ -9,6 +9,27 @@
...
+  JS value:
+
+  ```rust
+  mod a { #[wasm_bindgen(js_namespace = ["alpha"])] extern "C" { pub fn log(s: &str); } }
+  mod b { #[wasm_bindgen(js_namespace = ["beta"])]  extern "C" { pub fn log(s: &str); } }
+  // previously: one shim, so both `a::log` and `b::log` reached the same object
+  ```
+
+  This applies whether the namespace is written on the item or inherited from the
+  enclosing `extern "C"` block, and imported statics already accounted for it, so
```
guybedford: This claim is false — imported statics do *not* account for `js_namespace`. The static shim hash is `ShortHash((&js_name, module, &unraw_ident))` (parser.rs:1296) where `module` is the `module = "..."` attr, not the namespace; `js_namespace` is never passed into the static conversion (parser.rs:2662), and the CLI drops the second colliding static the same first-wins way (`crates/cli-support/src/wit/mod.rs:1028`). So two `static FOO: JsValue;` differing only in `js_namespace` silently bind to one value today, exactly like functions did.

I suspect the confusion is with the type *instanceof* shim, which does hash the namespace (parser.rs:1213).

Since it's the identical fix shape, I'd prefer statics were handled in this PR too; otherwise correct this sentence and we should file a follow-up.

crates/cli/tests/reference/js-namespace-shim-key.rs
```diff
@@ -0,0 +1,85 @@
...
+    use wasm_bindgen::prelude::*;
+
+    #[wasm_bindgen(js_namespace = ["gamma"])]
+    extern "C" {
+        pub fn warn(s: &str);
+    }
+}
+
+// Nested block-level namespaces. The separator used when folding the namespace
+// into the shim key must keep `["a", "b"]` + `c` distinct from `["a"]` + `b.c`.
```
guybedford: This case doesn't test what the comment says. The two sides use different Rust idents (`fn c` vs `fn b_c`), and the full signature token stream is already in the hash — so these two produce distinct shims on `main` as well; this pair passes without the fix.

It also can't test separator ambiguity as written: with the length-prefix encoding the aliasing pair would be something like `js_namespace = ["a", "b"]` vs `js_namespace = ["a\u{1}1\u{1}b"]` with *identical* Rust signatures. If that's worth a test at all it needs identical idents on both sides; otherwise I'd drop this pair rather than keep a comment that overstates the coverage.

crates/cli/tests/reference/js-namespace-shim-key.rs
```diff
@@ -0,0 +1,85 @@
+// FLAGS: --target=bundler
+
+// A `js_namespace` is part of a binding's identity: two otherwise identical
+// imports that differ only in their namespace resolve to different JS values
+// and so must not share a shim name. This has to hold whether the namespace is
+// written on the item or inherited from the enclosing `extern "C"` block --
+// the block-level form is the more common spelling.
```
guybedford: The changelog explicitly claims this applies "whether the namespace is written on the item or inherited from the enclosing `extern "C"` block", but every case here is block-level. Please add one pair using item-level `js_namespace` (and ideally one where item-level overrides block-level), since the resolution path being hashed (`item_opts.js_namespace().or(ctx.js_namespace)`) is exactly what that claim rests on.

crates/macro-support/src/parser.rs
```diff
@@ -1056,9 +1058,50 @@ impl<'a>
...
+            // whatever the segments contain.
+            //
+            // This must be the *resolved* namespace — a `js_namespace` written on
+            // the enclosing `extern "C"` block is inherited by every item in it and
+            // is just as much a part of the binding's identity as one written on
+            // the item. The caller performs that resolution
+            // (`item_opts.js_namespace().or(ctx.js_namespace)`) and hands the result
+            // in, so there is only one namespace value in scope here and reading the
+            // item-level attribute directly is not possible.
+            let ns = match js_namespace {
```
guybedford: The length-prefix encoding defends against `\u{1}` inside namespace segments, but folding onto `ns.1` reintroduces the same hole one level up: for methods, `ns.1` is the class name, which can come from `js_class = "..."` — an arbitrary string literal (parser.rs:977). So class `"n\u{1}1\u{1}x"` with no namespace aliases class `"n"` with namespace `["x"]` — precisely the contrivance this comment says is precluded.

Simpler and actually robust: skip manual encoding entirely and hash structurally, gated so no-namespace hashes are preserved:

```rust
let hash = match js_namespace {
    None => ShortHash((ns, sig, module, cfg_attrs)),
    Some(n) => ShortHash((ns, sig, module, cfg_attrs, n)),
};
```

Rust's `Hash` for slices/`Option` is already length-prefixed and discriminated, so there's nothing to hand-roll, `ns` can stay `(i32, &str)`, and the ~40 lines of encoding rationale collapse to a couple of lines.

guybedford (changes requested):
Thanks, the fix mechanism here is correct and this is a real silent-miscompile class of bug worth fixing — it's the import-side counterpart of #5154. I verified the no-namespace hash-compat claim holds (`(0, String)` hashes identically to `(0, &str)`), so the reference churn is confined to namespaced imports as intended.

A few issues before this can land, see the inline comments:

* The changelog claim that imported statics already account for `js_namespace` is false — statics have exactly the same collision and should either be fixed here or tracked in a follow-up.
* The `nested_split`/`nested_joined` test case doesn't exercise what its comment claims — it already passes on `main`.
* There's no item-level `js_namespace` case in the test despite the changelog claiming that coverage.
* The manual length-prefix encoding can be replaced with structural hashing, which is simpler and actually closes the aliasing hole it aims at.

On CI: the Clippy (native) and native test failures are unrelated toolchain drift on the runners (new stable changed closure shim symbol mangling, plus new clippy lints in untouched code) — `main` would fail identically if re-run today, so those need a separate fix on `main` first and a rebase here.

guybedford (approved):

PR #5254, opened 2026-08-05
Leave `env.memory` to the emscripten runtime

guybedford:
This fixes broken emscripten builds with imported (shared) memory, as reported by @ZacCranko in #5243 — thanks for the report!

When a module imports `env.memory` (`-pthread`/`-sIMPORTED_MEMORY` links), wasm-bindgen emitted a JS library entry for it:

```js
addToLibrary({
  memory: memory || new WebAssembly.Memory({initial:17,shared:true}),
});
```

This is broken twice over. The `memory` binding only exists as the `__wbg_get_imports(memory)` parameter in the non-emscripten output modes, so nothing declares it here — and since emcc evaluates `--js-library` values eagerly at link time, the expression fails inside the compiler rather than at runtime. More fundamentally, there is nothing to bind: emscripten's runtime creates `wasmMemory` and supplies `env.memory` to instantiation itself, special-cased outside the JS library mechanism, and any fallback constructing a fresh `WebAssembly.Memory` there could only ever produce a second memory silently desynced from emscripten's HEAP views.

The fix is to skip inserting the memory import definition entirely for `OutputMode::Emscripten`, leaving `env.memory` wiring to the runtime that owns it. `has_memory` correctly becomes `false`; all of its consumers (`__wbg_get_imports` memory parameter, init typings, web loading) are non-emscripten paths. Behavior for all other targets is unchanged, and the existing reference suite (including the atomics imported-memory targets) passes unmodified.

Supersedes the expression-rewrite approach in #5243. Follow-on: a shared-memory (atomics + `-pthread`) case in `tests/wasm32-emscripten` would let the Emscripten CI job cover this path.


ZacCranko:
Sweet, happy to defer to your judgement here. Thanks Guy.

ZacCranko:
cc @walkingeyerobot

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@
 
 ### Fixed
 
+* Emscripten output no longer emits a JS library binding for an imported
+  `env.memory` (shared memory / `-pthread` builds). The emitted
+  `memory || new WebAssembly.Memory(...)` expression referenced a binding that
+  doesn't exist in emscripten output and broke the emcc link; emscripten's own
+  runtime creates `wasmMemory` and supplies `env.memory` to instantiation
+  itself, so no binding is needed.
```
hoodmane: PR link?

crates/cli-support/src/js/mod.rs
```diff
@@ -772,20 +772,27 @@ impl<'a> Context<'a> {
...
-                if let Some(max) = mem.maximum {
-                    init_memory.push_str(&format!(",maximum:{max}"));
-                }
-                if mem.shared {
-                    init_memory.push_str(",shared:true");
+        // Emscripten's runtime owns `env.memory` (`wasmMemory`, created for
+        // `-sIMPORTED_MEMORY`/`-pthread` links) and supplies it to
+        // instantiation itself, so it must never be bound from the bindings
+        // library. `addToLibrary` values are also evaluated eagerly by emcc at
+        // link time, where no runtime binding exists.
```
```suggestion
        // The Emscripten linker automatically handles memory imports.
```

hoodmane (approved):

CHANGELOG.md
hoodmane: Changelog should be more about the problem that was fixed. Maybe:
```suggestion
* When Emscripten output was used with pthreads, wasm instantiation used to fail. It is now fixed.
```

PR #5261, opened 2026-08-06
slice to array fixes

logan-gatlin:
### Description
* `slice_to_array` on a `&mut` slice is now a compile error instead of silently
  discarding JS's writes.

* `slice_to_array` on a slice whose element type is a type parameter is now
  rejected with a single targeted error. Both shape checks now run together in
  the parser, so all violations in an `extern "C"` block are reported in one
  compile.

* Values with string, `JsValue`, or exported-struct elements received by JS
  (e.g. a `Vec<String>` return value, an outgoing `Vec<JsValue>` argument, or
  `Option<&[T]>` under `slice_to_array`) no longer make a redundant copy of the
  fresh string or array handed to JS: the string/externref loaders already
  return freshly-built values, so the `.slice()` in
  `VectorLoad`/`OptionVectorLoad` (needed only for primitive typed-array views
  into linear memory) was a second O(n) copy.

### Checklist


guybedford (approved):
Excellent, thank you. I've also added some further tweaks here.

PR #5264, opened 2026-08-07
Add js_string variants for web-sys string arguments

### LINKED ISSUE #5227: WebIDL - JsString variants for performance for `CanvasRenderingContext2d::set_fill_style()`
[`set_fill_style`](https://docs.rs/web-sys/latest/web_sys/struct.CanvasRenderingContext2d.html#method.set_fill_style) is deprecated. Looks like it was deprecated in 95bf437f35eaef8ee0756dba8685b2ff2278536c, when the `_str` and `_gradient` were added.

`_str()` is right out, for me. Turning `&str` into `JsValue` for `set_fill_style_str()` is about 50% of my UI thread.

I tried to create a gradient instead, and cache that, but that caused main UI thread CPU time to just **explode**.

Draw using
* `set_fill_style()`: about 35ms to draw.
* `set_fill_style_str()`: About 73ms.
* `set_fill_style_canvas_gradient()`: 536ms(!). That is a showstopper, especially in the UI thread.

For `_str()` the time goes to converting to `JsValue`, whereas for `_canvas_gradient()` it's  just an enormous `__wbg_fillRect` / `fillRect`.

[This is my use](https://github.com/ThomasHabets/rustradio/blob/a535027a86c22bb3fe170696fb49350eff98fd2a/rustradio-ui/src/mainthread/spectrum_sink.rs#L625) of `set_fill_style()`. The gradient version is not checked in, but similar. Key part:

```rust
     COLOR_CACHE.with(|slot| {
         slot.borrow_mut()
-            .entry(s.to_owned())
-            .or_insert_with(|| JsValue::from_str(s))
+            .entry(s)
+            .or_insert_with(move || {
+                let gradient = ctx.create_linear_gradient(0.0, 0.0, 1.0, 0.0);
+                let (r, g, b) = s;
+                let color = format!("rgb({r},{g},{b})");
+
+                gradient.add_color_stop(0.0, &color).unwrap();
+                gradient.add_color_stop(1.0, &color).unwrap();
+                gradient
+            })
             .clone()
     })
```

Are the non-deprecated functions really suitable replacements, or is there something I'm missing?

I saw it was linked to https://github.com/rustwasm/wasm-bindgen/pull/4156, and do I understand that this could be because one can only update these from the UI thread, not using shared buffers? Well, I'm doing it from the UI thread.

guybedford:
This adds `&js_sys::JsString` variants for string arguments and setters across web-sys, resolves #5227.

Converting `&str` to a JS string on every call re-encodes the data across the boundary each time, which shows up heavily in hot paths like canvas `fillStyle` — in the linked issue this conversion alone was ~50% of the UI thread. With these variants, callers can cache pre-allocated `JsString` values and pass them by handle with no copy or re-encode.

What was implemented:

* String types (`DOMString`, `ByteString`, `USVString`) at the top level of arguments (including through nullability and unions) now additionally flatten to a `JsString` variant, named with a `js_string` suffix following the existing `js_u8_array` convention
* Existing `&str` signatures keep their exact names: `&str`-only signatures disambiguate only against each other, so the string expansion never perturbs pre-expansion naming
* Methods with multiple string arguments generate only the all-`&str` and all-`JsString` forms, avoiding a combinatorial explosion of mixed variants
* Dictionary fields gain `set_x_js_string` setters, while constructor variant generation excludes the `JsString` expansions so existing constructors are unchanged

```rust
ctx.set_fill_style_str("rgb(255,0,0)");    // re-encodes every call
```

```rust
ctx.set_fill_style_js_string(&cached);     // passes the cached handle
```

The regenerated web-sys diff is verified purely additive: every existing function signature is preserved byte-for-byte, with ~1,950 new functions added. Checked with `--all-features` on `wasm32-unknown-unknown` both with and without `web_sys_unstable_apis`.

Note that enum-like `DOMString` arguments (e.g. `RTCRtpReceiver.getCapabilities(kind)`) also gain variants since the IDL carries no closed-set information; these are harmless. Longer term, generic erasure over string argument types could supersede this expansion with a single polymorphic signature per method.


ThomasHabets:
This seems reasonable to me. I guess the choice to take a ref to JsString instead of a clone is because it's nicer to the caller, not requiring a (cheap, because owned by JS) clone?

guybedford:
Yeah I think this is likely worth the tradeoff but I want to just let this one sit out another release cycle before landing, given the scope.

ThomasHabets:
There's always the option of making these `pub fn set_fill_style_js_string(this: &CanvasRenderingContext2d, value: impl AsRef<::js_sys::JsString>)`, right, thus implementing both.

But only saves the caller like one character, and at runtime one pointer dereference.

guybedford:
This should now be superseded by https://github.com/wasm-bindgen/wasm-bindgen/pull/5317.

PR #5265, opened 2026-08-07
Add `riscv64gc-unknown-linux-gnu` release artifacts

Xeonacid:
### Description

Cross-compile the CLI for RISC-V 64-bit Linux GNU, matching the existing aarch64-linux-gnu release job layout.

### Checklist


guybedford (approved):
Thanks. I've updated this to also add the strip flag for these release binaries.

PR #5266, opened 2026-08-08
Reject nested js_namespace on extern blocks and their items

### LINKED ISSUE #4324: Nested `js_namespace`s ignore the parent `js_namespace`
### Describe the Bug

```rs
#[wasm_bindgen(js_namespace = ["a"])]
extern "C" {
    #[wasm_bindgen(js_namespace = ["b"])]
    fn my_function();
}
```

The above import generates the following JS glue code:

```js
function __wbg_myfunction_8c7b624429f78550() {
    b.my_function();
};
```

Note the function is imported as `b.my_function()` and that `js_namespace = ["a"]` is ignored.

### Expected Behavior

The parent `js_namespace` should not be ignored. The most sane behavior would probably be to import the function as `a.b.my_function()`.

teddytennant:
### Description

Fixes #4324.

Setting `js_namespace` on both an `extern "C"` block and an item inside it used to silently drop the block's value:

```rust
#[wasm_bindgen(js_namespace = ["a"])]
extern "C" {
    #[wasm_bindgen(js_namespace = ["b"])]
    fn my_function();
}
```

generated `b.my_function()`. Nested paths already work via a single attribute (`js_namespace = ["a", "b"]`), so this makes the combination a hard error instead of inventing a second way to spell the same thing.

### Changes

- Error in `MacroParse<ForeignItemCtx>` when both the block and the item set `js_namespace`
- Guide and changelog updated
- UI test for the new diagnostic
- Reference fixtures that used the old override path now set the namespace on only one side

### Verification

- `cargo test -p wasm-bindgen-macro-support --lib`
- `cargo test -p wasm-bindgen-cli --test wasm-bindgen test_26_import`
- `cargo test -p wasm-bindgen-cli --test wasm-bindgen test_31_js`
- UI test `nested-js-namespace` passes (other UI failures match clean `main` on this rustc)
- `cargo fmt --all --check` and clippy on macro-support are clean

### Checklist


guybedford:
Thanks for pushing on this one. Since nesting is already supported via `["a", "b"]` I think we should just double down on this as a the canonical approach, over supporting arbitrary "concatenations".

So my preferred fix here would be to throw under nested attributes and treat this as a hard error instead.

teddytennant:
Good call. Switched this over to a hard error instead of nesting.

`js_namespace = ["a", "b"]` is the one way to write a nested path now. Setting it on both the block and an item inside is rejected, with a note pointing at that form. Guide, changelog, UI test, and the import reference fixtures are updated to match.

crates/macro-support/src/parser.rs
```diff
@@ -2661,6 +2661,20 @@ impl MacroParse<ForeignItemCtx> for syn::ForeignItem {
             BindgenAttrs::find(attrs)?
         };
 
+        // Nested namespaces belong in a single attribute (`js_namespace =
+        // ["a", "b"]`). Setting the attribute on both the block and an item
+        // inside it used to silently drop the block's value; reject it
+        // instead so the full path has to be written in one place.
+        if ctx.js_namespace.is_some() {
```
guybedford, line 2665: Nit: the `syn::ForeignItem::Verbatim` branch above returns early with `js_namespace: None`, so a verbatim static (e.g. a foreign static with an initializer) inside a namespaced block bypasses this check. An item-level `js_namespace` there likely trips `check_used` anyway, so this is fine to leave — just noting the gap for the record.

crates/macro-support/src/parser.rs
```diff
@@ -3165,4 +3179,125 @@ mod tests {
...
+        .unwrap_err();
+        let msg = format!("{err:?}");
+        assert!(
+            msg.contains("cannot be set on both"),
+            "unexpected error: {msg}"
+        );
+    }
+
+    #[test]
+    fn js_namespace_on_block_and_item_is_rejected_for_types_and_statics() {
```
guybedford: This test (and `js_namespace_on_block_and_item_is_rejected`) duplicates exactly what the `nested-js-namespace` UI fixture already asserts. The non-error resolution tests are worth keeping as regression coverage, but I'd trim the rejection ones to keep the suite lean.

crates/cli/tests/reference/js-namespace-shim-key.rs
```diff
@@ -72,19 +72,19 @@ mod item_level_other {
     }
 }
 
-// An item-level `js_namespace` overrides the block-level one, so the resolved
-// namespaces (and shims) differ even though the blocks agree.
-mod override_block {
+// Item-level vs block-level on the same path still need distinct shims.
+// (Nested block+item js_namespace is rejected; spell the full path once.)
+mod item_vs_block {
```
guybedford: After the rewrite this module tests item-only `eta` vs block-only `zeta`, which the neighboring `item_level`/block-level modules already cover — it now mostly exists to keep the reference output stable. Fine either way, but it could just be dropped.

crates/cli/tests/reference/import.rs
```diff
@@ -48,8 +48,8 @@ extern "C" {
 
 #[wasm_bindgen(js_namespace = ["a"])]
 extern "C" {
-    // test that namespaces are overwritten and not inherited/concatenated
-    #[wasm_bindgen(js_namespace = ["b"])]
+    // item-level js_namespace is not allowed when the block already has one;
```
guybedford: Nit: one line is enough here, e.g. `// block + item js_namespace is a hard error; the path goes in one attribute`.

guybedford (commented):
Thanks, this lands the hard-error direction well. The check sits before namespace resolution so it covers fns, types and statics uniformly, `js_namespace = []` is already rejected at attr-parse time so the `spans[0]` index is safe, and the shim-key reference outputs are correctly unchanged since the resolved namespaces didn't move.

Approval pending the nits below — none are structural.

crates/macro-support/src/parser.rs
teddytennant: Good catch. I left this unchanged for now.

crates/macro-support/src/parser.rs
teddytennant: Removed the duplicate tests.

crates/cli/tests/reference/import.rs
teddytennant: Shortened this to one line.

crates/cli/tests/reference/js-namespace-shim-key.rs
teddytennant: Dropped this pair and updated the reference files.

guybedford (approved):

PR #5270, opened 2026-08-11
Reference emcc's top-level export bindings in emscripten glue

guybedford:
This updates the emscripten target output so glue never reads `wasmExports['name']` inline inside inner functions. This avoids a bug where emcc's metadce and import/export minifier only reliably track exports through its own top-level receiving bindings, so name-keyed accesses in inner functions could be stripped at -O2 (e.g. `__wbindgen_start`, the `__wbindgen_externrefs` table) or left referencing pre-minification names, failing at runtime.

* Glue now references the asmjs-mangled identifiers emcc's `assignWasmExports` receiving code binds at the top level for every wasm export (`_add`, `___wbindgen_malloc`, `___wbindgen_externrefs`, ...). These are the canonical DCE-graph pairs: wasm-metadce keeps exactly the exports the included glue uses, and the minifier renames them in lockstep with the wasm.
* `$initBindgen` calls `__initialize` only when the module actually exports `_initialize`, since the receiving binding otherwise doesn't exist.
* Hoisted classes are emitted as `=`-prefixed string value snippets so jsifier declares them as `export var` rather than `export class`, which crashes emcc's acorn-optimizer `ExportNamedDeclaration` handling under `-sMODULARIZE=instance`.

```js
ptr => wasmExports['__wbg_counter_free'](ptr, 1)
```

```js
ptr => ___wbg_counter_free(ptr, 1)
```

Verified end-to-end against emscripten's unified `-sWASM_BINDGEN` flow at `-O2` (`MODULARIZE=instance`, `EXPORT_ES6`, `AUTO_INIT`): internal exports and the externref table survive metadce and are minify-renamed correctly, unused exports are still DCE'd, ESM `module = "..."` imports and exported classes link and run; debug builds unchanged. Emscripten CLI shape tests updated for the class snippet form, and the emscripten test harness now stubs the mangled receiving identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,19 @@
   the block-level namespace.
   [#4324](https://github.com/wasm-bindgen/wasm-bindgen/issues/4324)
 
+* Emscripten glue no longer reads `wasmExports['name']` inline inside inner
+  functions. It now references the asmjs-mangled identifiers emcc's own
+  top-level `assignWasmExports` receiving code binds for every wasm export
+  (e.g. `___wbindgen_malloc`, `___wbindgen_externrefs`), which are the
+  canonical DCE-graph pairs: wasm-metadce keeps exactly the exports the
+  included glue uses (previously internal exports and the externref table
+  could be stripped or left unrenamed by the import/export minifier at `-O2`,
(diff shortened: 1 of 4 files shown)


walkingeyerobot (approved):
looks good, thanks!

logan-gatlin (approved):

PR #5272, opened 2026-08-12
Support `impl Trait` in generic_per_mono argument position

logan-gatlin:
Stacked on #5230.

Support `impl Trait` generics in imports


guybedford:
This looks good, I just pushed the rebase. I'd still like to try this out further before landing. Getting implementation experience first to ensure it's on the right track.

guybedford (approved):

guybedford:
I've updated this branch as a maintainer edit: dropped the scalar `&T` commits and rebased the `impl Trait` work directly onto main, so the PR is now just the `impl Trait` desugaring commit (authorship preserved).

While the scalar `&T` implementation was well done — in particular the two-sided invariant between the runtime impls and the CLI descriptor set — I'd like to hold off on widening the public ABI surface with the `impl IntoWasmAbi for &T` scalar impls for now, since they're insta-stable once released and also observably widen what exported functions can return through type aliases. We can revisit that separately.

The `impl Trait` support needed no changes beyond removing its dependencies on the scalar refs (the `log_ref(&13u32)` reference-test calls and regenerated outputs, plus the CHANGELOG entry). Thanks @logan-gatlin!

PR #5277, opened 2026-08-14
Speed up tests

logan-gatlin:
### Description
Reference and CLI tests currently spawn one `cargo build` per file. Batching these together into workspaces and running `cargo build --workspace` is a modest speedup on my machine. I am guessing `cargo` can better share work between parallel jobs within a single process, and it spends a lot less time fighting over lockfiles.

### Checklist


logan-gatlin:
Looks like this does not improve times in CI, probably because the CI runner is already not very parallel. Will keep looking for speedups

guybedford (approved):

PR #5278, opened 2026-08-14
ci: pin icu_* crates to stop MSRV drift

logan-gatlin:
`icu_collections`/`icu_locale_core`/`icu_normalizer`/`icu_normalizer_data`/`icu_properties`/`icu_properties_data`/`icu_provider` released 2.3.0, which bumps their own MSRV to Rust 1.88.

```
error: rustc 1.86.0 is not supported by the following packages:
  icu_collections@2.3.0 requires rustc 1.88
  ...
```


guybedford (approved):
Thanks!

PR #5281, opened 2026-08-17
generics testing

logan-gatlin:


logan-gatlin:
Closing — this branch predates #5230 (the interpreter-based `generic_per_mono` that actually landed on `main`) and contains an independent, now-superseded implementation of the same attribute. Will re-port the `js-sys` conversion work on top of the merged implementation in a follow-up PR.

PR #5282, opened 2026-08-17
Fix generic_per_mono rejecting default type parameters

logan-gatlin:
Erased generics allowed this. It doesn't have any meaning in per-monomorphization generics, so this PR just drops them instead of erroring. This is a win for backwards-compat with erased generics code, but bad in the Rust philosophy of "ban nonsense code that doesn't do what you think it does". Open to feedback


guybedford:
I think retaining these as errors makes sense actually - Rust doesn't support default generics on functions itself. With erasable generics this was necessary for concretization, but in the real generics case we should give a proper error here instead.

logan-gatlin:
Ok, well if we remove erasible generics in favor of monomorphized ones, this will become a breaking change. Good to keep in mind

guybedford:
Right, if we need to do relaxation once we know the upgrade path that could be an option, but that should be based on knowing what the exact upgrade path looks like first.

PR #5287, opened 2026-08-18
cli-support: --experimental-memory-discard trampoline generation

guybedford:
This adds experimental support for the `memory.discard` instruction from the [memory-control proposal](https://github.com/WebAssembly/memory-control), allowing custom allocators to release physical pages back to the host on wasm32-unknown-unknown.

Wasm linear memory can only grow, so a temporary allocation spike stays resident for the lifetime of the instance. `memory.discard` releases the backing pages with zeroing semantics (equivalent to `madvise(MADV_DONTNEED)`), but LLVM has no way to emit it. Instead, an allocator declares:

```c
__attribute__((import_module("env"), import_name("__wbindgen_memory_discard")))
extern void __wbindgen_memory_discard(void *addr, size_t len);
```

and, behind the new `--experimental-memory-discard` flag, wasm-bindgen replaces the import with a local trampoline containing the instruction, so nothing survives to instantiation and page discard remains a pure wasm operation with no JS involved.

* Gated behind `--experimental-memory-discard` and documented as experimental and subject to change, since the emitted instruction requires an engine supporting the memory-control proposal
* The import without the flag is a hard error pointing at the flag, rather than a dangling `env` import at instantiation time
* Signature-checked as `[i32 i32] -> []`, following the memory's index type (`[i64 i64]` under memory64)
* Bumps walrus to 0.26.5 for `MemoryDiscard` support

Unit tests cover trampoline generation, memory64, flag gating, and signature mismatch errors, and a guide reference page documents the contract. Verified end-to-end against a jemalloc build whose `madvise` purge path forwards to this import ([guybedford/jemallocator memory-discard](https://github.com/guybedford/jemallocator/tree/memory-discard)) - the final bindgen output validates with wasm-tools and contains the discard trampoline with no surviving import.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
   info to a separate `*_bg.debug.wasm` file. Use `--debug-info-url` to set
   the recorded URL for the debug info. [#5279](https://github.com/wasm-bindgen/wasm-bindgen/pull/5279)
 
+* Added an experimental `--experimental-memory-discard` flag which replaces an
+  `env.__wbindgen_memory_discard` function import with a local trampoline
+  containing the `memory.discard` instruction from the memory-control
+  proposal, allowing custom allocators to release physical pages back to the
+  host. Experimental and subject to change.
+
 ### Changed
 
 * Setting `js_namespace` on both an `extern "C"` block and an item inside it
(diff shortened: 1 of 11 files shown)


logan-gatlin:
I added test coverage to make sure the import gets stripped, and that multiple/no memories error.

logan-gatlin (approved):

PR #5289, opened 2026-08-18
Add WebGPU `setImmediates` to immutable slice whitelist

beicause:
### Description
`setImmediates` should take immutable u8 slice.

### Checklist

### DIFF
--- a/crates/web-sys/src/features/gen_GpuComputePassEncoder.rs
+++ b/crates/web-sys/src/features/gen_GpuComputePassEncoder.rs
@@ -326,7 +326,7 @@ extern "C" {
     pub fn set_immediates_with_u8_slice(
         this: &GpuComputePassEncoder,
         range_offset: u32,
-        data: &mut [u8],
+        data: &[u8],
     ) -> Result<(), JsValue>;
     #[cfg(web_sys_unstable_apis)]
     #[wasm_bindgen(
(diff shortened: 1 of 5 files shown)


crates/webidl/src/constants.rs
```diff
@@ -110,6 +110,7 @@ pub(crate) static IMMUTABLE_SLICE_WHITELIST: Lazy<BTreeSet<&'static str>> = Lazy
         "send",
         // WebGPU
         "setBindGroup",
+        "setImmediates",
```
guybedford, line 113: Correct per the WebGPU spec: `setImmediates` only reads `data` to copy immediate values into the encoder, never writes into the caller's buffer, so it belongs here alongside `setBindGroup`/`writeBuffer`/`writeTexture`. Note the whitelist matches by bare operation name globally, but `setImmediates` only appears in WebGPU.webidl, so there's no collateral effect on other interfaces.

crates/web-sys/src/features/gen_GpuComputePassEncoder.rs
```diff
@@ -326,7 +326,7 @@ extern "C" {
     pub fn set_immediates_with_u8_slice(
         this: &GpuComputePassEncoder,
         range_offset: u32,
-        data: &mut [u8],
+        data: &[u8],
```
guybedford, line 329: Verified the generated files exactly match regeneration output (`cargo run --release -p wasm-bindgen-webidl -- webidls src/features ./Cargo.toml` produces a zero diff on this branch), and compile-checked `web-sys` with `--cfg=web_sys_unstable_apis` and the three GPU encoder features on `wasm32-unknown-unknown`.

guybedford (approved):
Verified this against the WebGPU spec and the generator output — details in the line comments. Also confirmed this is source-compatible for existing callers since `&mut [u8]` reborrow-coerces to `&[u8]` at call sites (and these APIs are behind `web_sys_unstable_apis` regardless). LGTM, thanks!

PR #5290, opened 2026-08-18
Support per-mono generics on classes

logan-gatlin:
### Description
Support `generic_per_mono` for classes. This uses mostly the same machinery as erased generics. This means that most of the test coverage added here extends to the erased-generic path as well.

### Checklist


crates/macro-support/src/codegen.rs
```diff
@@ -3325,6 +3401,71 @@ impl<'a> FnClassGenerics<'a> {
...
+            return quote! { impl #class };
+        }
+        // Type lifetimes: declared on the impl (deduplicated, since a header
+        // may bind each name once) AND passed to the type (positionally, as
+        // written — `Holder<'b, 'a>` must not come back out as `Holder<'a, 'b>`,
+        // and `Holder<'a, 'a>` must not collapse to a single argument).
+        //
+        // The two lists can only diverge for a class type carrying more than
+        // one lifetime argument, which no import can express today: the
+        // `type Foo<'a, 'b>;` declaration is itself rejected before any of this
```
guybedford: Nit: `type Foo<'a, 'b>;` isn't rejected anywhere — it fails incidentally with E0261 against the generated `&'a` handle impls, because `impl_generics_with_lifetime_a` never declares the type's own lifetime params. With the ref-impl fix from the review body, multi-lifetime declarations actually start compiling, so this invariant then holds by construction and the positional/dedup split here becomes load-bearing — worth rewording once that lands.

crates/macro-support/src/codegen.rs
```diff
@@ -3539,11 +3694,152 @@ impl ast::ImportFunction {
...
+        // rather than `class_lifetime_params`, but it is declared on the same
+        // impl header and carries the same hazard.
+        //
+        // Like the undeclared-lifetime check above, this is not receiver-only:
+        // the constructor route reaches an identically-shaped `impl`.
+        if !(fn_class_generics.class_lifetime_params.is_empty()
+            && fn_class_generics.class_bound_lifetime_params.is_empty())
+            && !fn_class_generics.class_generic_params.is_empty()
+        {
+            bail_span!(
```
guybedford: I reproduced the E0521 exactly as described here, but the root cause is the `&self` handle impl rather than anything in the hoisting: `impl<'a, T> IntoWasmAbi for &'a Holder<'a, T>` unifies the reference lifetime with the class lifetime, and that unification is where the `'a: 'static` obligation comes from once a hoisted `T` has to resolve through the same impl. With the ref impls generalised (fresh `'__wbg_ref`, and the type's own lifetimes declared — see the review body for the tested patch) this shape compiles fine, so this rejection and its ui-tests can be dropped in favour of pass/runtime coverage.

crates/macro-support/src/codegen.rs
```diff
@@ -3539,11 +3694,152 @@ impl ast::ImportFunction {
...
+        // static method the class comes from the return type, where stripping
+        // the arguments and binding the class's own defaults (`impl Promise`
+        // for a `-> Promise<T::Resolution>`) is the established, working
+        // behaviour shared with the type-erasure path: `class_return_path`
+        // already declines to hoist a non-constraining argument list, and the
+        // resulting inherent method is perfectly valid — it simply hangs off
+        // the defaulted class. Rejecting it here would break real imports, e.g.
+        // `js_sys::Promise::new_typed<T: Promising>(..) -> Promise<<T as
+        // Promising>::Resolution>`.
+        if !is_method {
```
guybedford: This exemption is right for the projection case (`Promise<T::Resolution>` must keep binding the defaults), but it also exempts the all-concrete case: `fn new<T>(v: T) -> Holder<u32>` silently lands in `impl Holder` — the same silent mis-target the receiver check below exists to prevent, since `class_return_path` treats concrete args as (vacuously) constraining and does retarget the impl here. If the `impl Holder<u32>` pass-through from the next comment lands, this resolves naturally; otherwise the concrete-args rejection should apply to the constructor/self-returning route too.

crates/macro-support/src/codegen.rs
```diff
@@ -3539,11 +3694,152 @@ impl ast::ImportFunction {
...
+
+        // Every type argument mentions no generic parameter of the function, so
+        // there is nothing to hoist and the `impl` block would be built for the
+        // class's own parameter defaults instead of the class type as written
+        // (`impl Holder` for a declared `&Holder<u32>`, i.e. a receiver type
+        // mismatch on the generated method).
+        if !fn_class_generics.class_generic_exprs.is_empty()
+            && !fn_class_generics.has_class_generics()
+        {
+            bail_span!(
```
guybedford: Two things here.

First, the mixed form `&Holder<u32, T>` is already accepted and works — `class_generic_exprs` carries every type argument verbatim, so this emits `impl<T> Holder<u32, T>` (compile-verified). That deserves a pass-test line, and both the guide's "each argument must be a generic parameter of the function" and this error's "requires each generic argument … to be a bare generic parameter" overstate the restriction.

Second, the all-concrete case is only unsupported because zero hoisted params falls through to the bare `impl Holder`. Since the exprs are already carried as written, emitting `impl Holder<u32>` when there are exprs but nothing to hoist turns this rejection into support — "this method only exists on `Holder<u32>`" is a legitimate shape. Fine as a follow-up, but then the message should say "not supported yet" rather than implying it can't work.

guybedford (changes requested):
This is very well put together — the shared `class_impl_def` so the erased and per-mono paths can't drift, the lifetime-bucket fixes in `get_fn_generics`, the inline-lifetime-bound and default fixes, and the pass-tests infrastructure are all solid improvements, and the diagnostics are excellent. I verified the E0521 repro, the `generic-import` reference test, and the macro ui/pass suites locally (note CI never triggered on this PR — worth a re-push to kick it).

The main thread of this review: both lifetime restrictions here are downstream of a pre-existing bug in the imported-type codegen, and fixing that root cause makes the feature meaningfully more general — which is where I'd like this to land.

`impl_generics_with_lifetime_a` (codegen.rs ~1179) declares only a fresh `'a` plus the type's *type* params — never the type's own lifetime params — so the `&self` handle impls come out as:

```rs
impl<'a, T> IntoWasmAbi for &'a Holder<'a, T> { .. }
```

Two consequences:

1. A class lifetime not literally named `'a` (`type Tagged<'x>`) fails with E0261 against generated code ("lifetime `'x` is missing in item created through this procedural macro"). Since this PR documents class-level lifetimes as supported, users will hit this immediately.
2. When the lifetime *is* `'a`, the reference lifetime unifies with the class lifetime — and that unification is exactly what forces `'a: 'static` in the mixed class-lifetime + hoisted-type-param shape, i.e. the E0521 this PR rejects as "not supported yet".

I tested this patch:

```rs
let own_lifetimes = generics::lifetime_params(&self.generics);
let impl_generics_with_lifetime_a =
    quote! { <'__wbg_ref, #(#own_lifetimes,)* #(#type_params_with_bounds),*> };
```

with the two `&'a #rust_name` impls and the `<&'a JsValue as IntoWasmAbi>::Abi` projection renamed to `'__wbg_ref`. With that in place both the `type Tagged<'x>` E0261 and the mixed `&'a LtHolder<'a, T>` E0521 disappear (compile-verified), and js-sys builds clean. So: please fix the ref impls in this PR and drop the mixed-shape rejection, moving its ui-tests over to pass/runtime coverage.

Two further generality items are in the line comments: concrete class arguments (`&Holder<u32, T>` already works and should be pinned; `&Holder<u32>` is one small step from working), and the constructor route's exemption from the concrete-args check.

Last: there is no runtime coverage — the reference snapshots pin the generated JS but never execute it. A small `tests/wasm` headless test driving a generic class constructor/method/static across two instantiations would close that gap.

logan-gatlin:
Something I hadn't thought about is multiple extern blocks. For example:
```rust
#[wasm_bindgen]
extern "C" {
    type LifetimeWidget<'a>;
}

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(static_method_of = LifetimeWidget, generic_per_mono)]
    fn utility<T>(value: T);
}
```

The second `extern` block generates an impl without the needed lifetime parameter. It doesn't know about the lifetime parameter on `LifetimeWidget`

```rust
impl LifetimeWidget {
    pub fn utility<T>(value: T) { /* ... */ }
}
```

This is solvable though. Instead of passing just a name to `static_method_of` we should require passing a type. So the second extern block becomes:

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(static_method_of = LifetimeWidget<'a>, generic_per_mono)]
    fn utility<T>(value: T);
}
```
And a missing lifetime error points to user code, not generated code.

logan-gatlin:
accidentally closed

crates/macro-support/src/parser.rs
```diff
@@ -399,6 +399,21 @@ impl Parse for BindgenAttr {
                 return Ok(BindgenAttr::$variant(attr_span, ident))
             });
 
+            (@parser $variant:ident(Span, syn::TypePath)) => ({
```
guybedford, line 402: This syntax extension currently applies to every user of `static_method_of`, not just generic-mono imports — `pass-tests/generic-static-method-class-argument.rs` exercises it on the plain erasure path. With the type-level gate this should only be accepted when the class type is a generic-mono type; a `static_method_of` argument list on anything else should be rejected with a diagnostic.

crates/macro-support/src/parser.rs
```diff
@@ -2731,12 +2756,71 @@ impl MacroParse<BindgenAttrs> for syn::ItemForeignMod {
...
             }
         }
+        add_class_cfg_to_function_shims(&mut program.imports[import_start..]);
         Diagnostic::from_vec(errors)?;
         opts.check_used();
         Ok(())
     }
 }
 
+fn add_class_cfg_to_function_shims(imports: &mut [ast::Import]) {
```
guybedford, line 2788: This runs for every extern block with no generics or attribute check at all: any method whose class carries a `#[cfg]` gets a renamed shim, including entirely non-generic imports. That's a behaviour change on stable paths — please move this (together with `cfg_gate_attrs`, the `ImportType` hash change, and the per-candidate emission in codegen) into a separate base PR.

crates/macro-support/src/codegen.rs
```diff
@@ -189,6 +238,107 @@ impl TryToTokens for ast::Program {
     }
 }
 
+fn imported_class_generics<'a>(
```
guybedford, line 241: The candidate lookup and per-candidate emission run for every import function unconditionally. Under type-level gating this should only consult the types map for generic-mono types, and the cfg candidate machinery goes away with the base-PR split — cfg-duplicated generic-mono declarations can simply be unsupported.

crates/macro-support/src/codegen.rs
```diff
@@ -3302,29 +3661,411 @@ struct FnClassGenerics<'a> {
...
+                        &lifetime_replacements,
+                    ) {
+                        bail_span!(
+                            class,
+                            "wasm-bindgen cannot substitute a projected imported-type default for a non-path class argument"
+                        );
+                    }
+                    type_replacements.insert(parameter.ident.clone(), default);
+                }
+                _ => return Ok(()),
```
guybedford, line 4049: This arm silently stops propagating the remaining declaration bounds on a parameter/argument kind mismatch. rustc catches the arity error on the type itself, so this is unreachable-in-practice, but a `bail_span!` (or at least a comment stating why falling through is safe) would avoid silent bound drops if the lists ever desync.

CHANGELOG.md
```diff
@@ -15,6 +15,19 @@
   proposal, allowing custom allocators to release physical pages back to the
   host. Experimental and subject to change.
 
+* `#[wasm_bindgen(generic_per_mono)]` now supports class-level generic
```
guybedford: These entries need the PR link (#5290), matching the surrounding entries. Note the second entry also describes an erasure-path change, which shouldn't exist once the attribute gating is applied.

crates/macro/ui-tests/generic-per-mono-cross-block-bound-suggestion.stderr
```diff
@@ -0,0 +1,52 @@
...
+   |
+note: required by a bound in `Foo`
+  --> ui-tests/generic-per-mono-cross-block-bound-suggestion.rs:5:21
+   |
+5  |     pub type Foo<T: Clone>;
+   |                     ^^^^^ required by this bound in `Foo`
+   = note: this error originates in the attribute macro `wasm_bindgen` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider further restricting type parameter `T` with trait `Clone`
+   |
+11 |     pub fn value, T: std::clone::Clone<T>(this: &Foo<T>);
```
guybedford, line 35: This pinned suggestion is garbled (`pub fn value, T: std::clone::Clone<T>(...)`) — the wrapper's generics are spanned such that rustc splices the suggestion mid-signature. Fine to keep as the documented cross-block fallback, but worth a look at the spans used for the wrapper's parameter list.

guybedford (commented):
This is strong work — reusing the erasure path's hoisting analysis and unifying the impl assembly in `class_impl_def` is the right structure, and the test coverage is excellent: the positional-vs-deduplicated lifetime lists, the RFC 447 impl-predicate discipline, and the shim `where`-clause restatement are all correctly handled and well pinned. The reference-impl lifetime fixes (fresh `'__wbg_ref`, declaring the type's own lifetime params, the `From<JsValue>` where clause) are genuine bug fixes and are fine to keep unconditional.

Two structural changes requested before this lands:

**1. All new behaviour must be gated by the attribute, on the type declaration.**

The imported *type* should itself be declared `generic_per_mono`, making it a generic-mono type, and every new branch should then be mediated by that: class-level hoisting, `static_method_of = Type<Args>`, and declaration-bound propagation should apply only to such types. Declarations not opted in must produce byte-identical output to main. As written, several behaviours leak onto the stable erasure path unconditionally:

* `static_method_of = TypePath` is accepted everywhere, and the erasure path hoists it (`pass-tests/generic-static-method-class-argument.rs`).
* Concrete receiver class arguments now compile on the erasure path (`&GenericClass<String>` binds `impl GenericClass<String>` — the removed E0308 block in `generics-comprehensive.stderr`).
* `add_class_bounds` propagates declaration bounds on the non-mono branch.
* The `get_fn_generics` widening from `is_static: false` to `Operation(_)`.

**2. The cfg subsystem should be split out into its own base PR.**

`cfg_gate_attrs` / `add_class_cfg_to_function_shims` / the instanceof-shim hash change / per-candidate emission / descriptor cfg unions apply unconditionally to *all* import codegen, including entirely non-generic imports: any method whose class carries a `#[cfg]` gets renamed shims and cfg-wrapped impls. That is a standalone behaviour change to stable paths — and the ungated-method-on-a-cfg'd-out-class shape (the `Absent<T>` test) is a new convenience feature rather than a regression fix, since that shape fails on main too. Please pull all of it into a separate base PR where it can be reviewed on its own terms. This PR can then treat cfg-duplicated generic-mono type declarations as unsupported, which the type-level gate makes an acceptable restriction for an experimental feature.

Smaller items inline. Once gating is in place, the guide additions should live under the `generic_per_mono` page only — `static_method_of.md` shouldn't advertise the arguments syntax as general.

crates/macro/ui-tests/generic-per-mono-cross-block-bound-suggestion.stderr
logan-gatlin: I'm not sure whats happening here. This kept happening in my editor with the "Quick Fix" LSP code action. It would garble or not conditional on where my cursor was. I wonder if its a rust problem

guybedford (approved):
I've reviewed this in depth — the design here is right: reusing the erasure path's `get_fn_generics` hoisting analysis and sharing the `class_impl_def` impl-header assembly between both generic paths keeps them from drifting apart, and the test coverage is excellent (compile-pass pins for the specific rustc errors each shape used to hit, runtime assertions on observable JS state, and the reference outputs).

I've rebased the branch onto main, which needed some real work beyond conflicts:

* Renamed `generic_per_mono` → `experimental_generic_mono` throughout the new tests, docs, diagnostics, and changelog entries, following #5301 (internal field names left as-is, matching main).
* Ported the new code to syn 3.0 (#5297): `TypePath` patterns gained `attrs`, `Type::BareFn` → `Type::FnPtr` in `type_is_constraining_for` and the lifetime visitor, and `TypeParam::default` is now a tuple in `add_class_bounds`.
* Reconciled with #5272 (argument-position `impl Trait`): synthesized parameters are appended to the wrapper's remaining non-hoisted generics — they can never be hoisted, so they keep their inline bounds on the wrapper's own parameter list.
* Regenerated the reference outputs and the ui stderr files under CI's pinned rustc 1.88.

I also pushed a small review-fixes commit on top:

* `impl Trait` in an impl class type's argument list is now rejected up front on both generic paths (`&Holder<impl Bounded>`), with a ui test; previously this fell through to an E0562 reported against macro-expanded code.
* The transitive bound-hoisting loop carried an always-empty lifetime accumulator (`predicate_lifetimes_to_add` was only ever extended after an `is_empty` guard on the same set); the gate now states the invariant directly — a hoisted equality RHS must use no remaining function lifetimes.
* Added a comment on the constructor exemption in `imported_class_generics` candidate matching.

Verified locally: `wasm-bindgen-macro` ui + pass tests under rustc 1.88, macro-support unit tests, and the CLI reference suite. Approving — will land once CI is green.

Issue #2247, opened 2020-07-21
Can't compile structs with the same name in different rust modules


OlegTheCat:
### Describe the Bug

Structs with the same names, that reside in different rust modules and have `wasm_bindgen` attribute applied, cannot be compiled. If I understand correctly this is because `wasm_bindgen` generates JS bindings into a single JS file and doesn't follow rust modules structure.

### Steps to Reproduce

1. Create file `my_mod.rs` with the following content:

```
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub struct MyStruct {
    x: usize
}
```

2. Add the following code to `lib.rs`:
```
use wasm_bindgen::prelude::*;
mod my_mod;

#[wasm_bindgen]
pub struct MyStruct {
    foo: String
}

```
3. Execute `wasm-pack build`.

### Expected Behavior

Rust code is successfully compiled.

### Actual Behavior

Compilation fails with `symbol `__wbg_mystruct_free` is already defined` error.

alexcrichton:
Yes this is an unfortunate limitation of wasm-bindgen because there's no namespacing of exported types. They're simply all exported into the same namespace in JS, so there's no way to export the same name multiple times.

OlegTheCat:
@alexcrichton, I see, thanks for the explanation. Is it something that is going to be addressed in the future? If no, I think, we can close this issue.

alexcrichton:
At this time I don't have any plans to fix this myself, but if someone else is interested in fixing it I can help review! 

(I don't know how best to fix it right now)

hugopeixoto:
As a workaround to avoid having to rename the rust structs, one could use `#[wasm_bindgen(js_name = MyModMyStruct)]` or something like that to avoid the name collision, right?

iacore:
> As a workaround to avoid having to rename the rust structs, one could use `#[wasm_bindgen(js_name = MyModMyStruct)]` or something like that to avoid the name collision, right?

not working

PR #5293, opened 2026-08-21
Mangle export shim symbols with a per-crate hash

guybedford:
This implements per-crate hash mangling for export shim symbols, resolving #2247.

Export shim symbols (`foo`, `classname_method`, `__wbg_classname_free`, field accessors, upcast shims and their `__wbindgen_describe_*` functions) were derived purely from JS-visible names, so two crates exporting the same name — most commonly two semver-incompatible versions of one crate in the same dependency graph — failed the link with wasm-ld duplicate-symbol errors. Import shims have always avoided this by hashing `CARGO_PKG_NAME`/`CARGO_PKG_VERSION` into their symbol names; exports never received the same treatment because cli-support recomputes export symbol names from decoded metadata rather than carrying them in the custom section.

What was implemented:

* The macro now emits every export shim symbol mangled as `{name}_{crate_hash}` (exports, `__wbg_*_free`, field getters/setters, upcast shims, the struct `new`/`unwrap` placeholder imports, and dynamic-union variant describe functions; describe functions follow their shim name automatically). The hash is the `ShortHash` already carried as the suffix of `unique_crate_identifier`.
* cli-support recomputes the mangled names from `unique_crate_identifier` and, once each export is matched, restores its canonical name (including the name-section entry) and re-sorts demangled exports back into wasm-ld's symbol-name order. Generated output is byte-identical for all non-colliding code — every pre-existing reference snapshot is unchanged.
* Same-named `#[wasm_bindgen(private)]` structs/enums from different crates now coexist: an identity pre-pass numbers colliding private definitions (`Status`, `Status2`, ...) and remaps class identities, descriptor type references and derived symbol names per defining crate, so each crate's methods, accessors and free/new/unwrap wiring resolve to its own class.
* Genuinely conflicting public exports now produce a clear, namespace-aware error at processing time instead of a cryptic linker failure.

Before:

```
rust-lld: error: duplicate symbol: __wbg_widget_free
```

After:

```
the name `Widget` is exported multiple times in this build; rename one side
with `js_name`/`js_namespace`, or mark copies that are not part of your
public interface as `#[wasm_bindgen(private)]`
```

Test coverage: a cross-crate runtime test (same-named private struct in `tests/crates/a` and `b`, exercising both classes and their field accessors end-to-end), a reference test pinning the numbered private-dedupe bindings, and diagnostics tests for the public struct and function collision errors. The `private` reference documentation is updated accordingly.

A few nuances: this is a macro/CLI contract change, so `SCHEMA_VERSION` is bumped and version pairing is enforced by the existing schema check. Same-crate duplicate names still fail at link time (same hash), unchanged from today. Structural references by name to a type that exists only as multiple private copies (e.g. from a third crate, or from generic-import monomorphisations which have no defining-crate context) are inherently ambiguous and now error explicitly. String enums and dynamic unions no longer have colliding link symbols, but JS-level dedup of same-named private copies is not extended to them here and could follow up.


logan-gatlin:
LGTM, it would be nice to do this with type descriptors too, but I'll make a different PR for that

logan-gatlin (approved):

PR #5297, opened 2026-08-24
Update syn requirement from 2.0 to 3.0

### LINKED ISSUE #5296: Migrate to syn 3
`serde_derive` 1.0.229 moved to syn 3, while wasm-bindgen still requires `syn ^2.0`. Any downstream crate using both serde and wasm-bindgen (a very common combination for wasm projects) now compiles two major versions of syn, which hurts build times and trips `clippy::multiple_crate_versions` for projects that deny it.

The Dependabot bumps (#5232, #5240) were closed after CI failures, understandably: syn 3 reshaped parts of the syntax tree, so the requirement bump alone does not compile. The actual port is small, though: it only touches `macro-support` and `webidl` (`test-macro` compiles unchanged), and syn 3's MSRV (1.71) is below this repo's 1.77. PR with the port follows.

One consideration for timing rather than for the code: as long as most of the proc-macro ecosystem is still on syn 2, releasing this moves the dual-syn cost to downstreams that don't use syn 3 crates yet. If you'd rather land it but hold the release until more of the ecosystem has switched, the PR can wait on the shelf.

Crovitche-1623:
### Description

Ports `macro-support` and `webidl` to the syn 3 API (`test-macro` needed no code change). Motivation and context in #5296 : the plain requirement bumps (#5232, #5240) could not compile because syn 3 reshaped parts of the syntax tree.

API changes covered:

- Add the new `attrs` field to type-node patterns (ignored) and constructors (empty), and replace the removed `From<TypeX> for Type` impls with explicit `syn::Type::X(...)` wrapping.
- Read `defaultness` from the new `modifiers` bags on `ItemImpl` and `ImplItemFn`; adjust `ItemImpl::trait_` to its new 2-tuple shape (the `!` polarity moved into `modifiers`; trait impls are rejected here anyway).
- Replace `Signature::unsafety` with `Signature::safety` matching (the new `safe fn` classifies as not unsafe, like the previous absence of `unsafe`).
- Rewrite `get_self_method` on top of `ReceiverKind`, since `Receiver` lost its always-populated `ty` field: `&self`/`&mut self` come from `ReceiverKind::Reference`, and the `self: Ty` forms are classified by inspecting the explicit type, as before.
- Detect `?Sized` via `TraitBound::maybe` (`TraitBoundModifier` is gone).
- Follow the `Type::BareFn` -> `Type::FnPtr` rename, and the `Punctuated<Type>` -> `Punctuated<NamedArg>` change in fn-pointer and parenthesized generic arguments.
- `TypeParam::default` now carries its `=` token as `Option<(Eq, Type)>`.

Validation:

- Unit tests pass; clippy is clean on native and Wasm targets.
- Regenerating web-sys with the ported webidl produces byte-identical output.
- The trybuild UI snapshots are unchanged relative to the syn 2 baseline on the same toolchain.
- Full CI matrix ran green on my fork (51 jobs): https://github.com/Crovitche-1623/wasm-bindgen/actions/runs/32736522633 (the Coverage job failure there is the Codecov upload only, which cannot work from a fork).

Closes #5296

### Checklist

### DIFF
--- a/crates/macro-support/Cargo.toml
+++ b/crates/macro-support/Cargo.toml
@@ -19,7 +19,7 @@ strict-macro = []
 bumpalo = "3.0.0"
 proc-macro2 = "1.0"
 quote = '1.0'
-syn = { version = '2.0', features = ['visit', 'visit-mut', 'full', 'extra-traits'] }
+syn = { version = '3.0', features = ['visit', 'visit-mut', 'full', 'extra-traits'] }
 wasm-bindgen-shared = { path = "../shared", version = "=0.2.127" }
 
 [lints]
--- a/crates/macro-support/src/codegen.rs
+++ b/crates/macro-support/src/codegen.rs
@@ -898,6 +898,7 @@ impl TryToTokens for ast::Export {
             converted_arguments.push(quote! { #ident });
         }
         let syn_unit = syn::Type::Tuple(syn::TypeTuple {
+            attrs: Vec::new(),
             elems: Default::default(),
             paren_token: Default::default(),
         });
@@ -2561,6 +2562,7 @@ impl TryToTokens for ast::ImportFunction {
         if let Some((_, class)) = class {
             let mut class = class.clone();
             if let syn::Type::Path(syn::TypePath {
+                attrs: _,
                 qself: None,
                 ref mut path,
             }) = class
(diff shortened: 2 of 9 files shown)


guybedford (approved):
Thanks for posting the PR, I've rebased and fixed it up to the latest changes on main. Will land shortly.

Crovitche-1623:
Thank you @guybedford !

PR #5298, opened 2026-08-24
fix: Incorrect generic bounds suggestion by rustc

logan-gatlin:
### Description
This code fails to compile because the `Clone` trait bound is not met:

```rust
#[wasm_bindgen]
extern "C" {
    pub type Foo<T: Clone>;
}

#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(method)]
    pub fn value<T>(this: &Foo<T>);
}
```

however, the compiler suggested quick fix is wrong:

```
 --> src/lib.rs:8:1
   |
 8 | #[wasm_bindgen]
   | ^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `T`
   |
note: required by a bound in `Foo`
  --> src/lib.rs:5:21
   |
 5 |     pub type Foo<T: Clone>;
   |                     ^^^^^ required by this bound in `Foo`
help: consider restricting type parameter `T` with trait `Clone`
   |
11 |     pub fn value<T>(this: &Foo<T: std::clone::Clone>);
   |                                +++++++++++++++++++
```

This suggestion is wrong, the bound belongs on `value<T: Clone>`

### Checklist

### DIFF
--- a/crates/macro-support/src/generics.rs
+++ b/crates/macro-support/src/generics.rs
@@ -78,6 +78,24 @@ impl<'a, 'b> GenericNameVisitor<'a, 'b> {
             found_set,
         }
     }
+
+    fn generic_param(&self, ident: &Ident) -> Option<&Ident> {
+        self.generic_params
+            .iter()
+            .copied()
+            .find(|param| *param == ident)
+    }
+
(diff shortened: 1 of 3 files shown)


guybedford (approved):
Great work improving these diagnostics!

PR #5299, opened 2026-08-25
Fix cross-crate private type descriptor resolution

logan-gatlin:
Similar to #5293 but for type descriptors. This prevents two types (one marked `#[wasm_bindgen(private)]` and one not) in different crates from colliding. See the regression test for an example: https://github.com/wasm-bindgen/wasm-bindgen/pull/5299/changes#diff-a2169ecc87020b51e343e4974cade0bda6762512b6d8ae0fd279cff089fb9038

### DIFF
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -252,6 +252,10 @@ impl Descriptor {
             | Descriptor::Vector(d)
             | Descriptor::Option(d)
             | Descriptor::Result(d) => d.visit_named_types_mut(f),
+            // An externref has no defining crate, so a JS type whose name
+            // matches an ambiguous all-`private` struct/enum name still hits
+            // the ambiguity check even though it never refers to the Rust
+            // type.
             Descriptor::NamedExternref(name) => f(name, None),
             Descriptor::Enum {
                 name,
--- a/crates/cli-support/src/descriptor.rs
+++ b/crates/cli-support/src/descriptor.rs
@@ -61,6 +61,7 @@ pub enum Descriptor {
     Enum {
         name: String,
         hole: u32,
+        unique_crate_identifier: String,
     },
     StringEnum {
         name: String,
@@ -71,7 +72,10 @@ pub enum Descriptor {
         name: String,
         variant_types: Vec<Descriptor>,
     },
-    RustStruct(String),
+    RustStruct {
+        name: String,
+        unique_crate_identifier: String,
+    },
     Char,
     Option(Box<Descriptor>),
     Result(Box<Descriptor>),
@@ -180,7 +184,12 @@ impl Descriptor {
             ENUM => {
                 let name = get_string(data);
                 let hole = get(data);
-                Descriptor::Enum { name, hole }
+                let unique_crate_identifier = get_string(data);
+                Descriptor::Enum {
+                    name,
+                    hole,
+                    unique_crate_identifier,
+                }
             }
             STRING_ENUM => {
                 let name = get_string(data);
@@ -207,7 +216,11 @@ impl Descriptor {
             }
             RUST_STRUCT => {
                 let name = get_string(data);
-                Descriptor::RustStruct(name)
+                let unique_crate_identifier = get_string(data);
+                Descriptor::RustStruct {
+                    name,
+                    unique_crate_identifier,
+                }
             }
             NAMED_EXTERNREF => {
                 let name = get_string(data);
@@ -224,10 +237,11 @@ impl Descriptor {
 
     /// Visit every struct/enum/externref type name embedded in this
     /// descriptor tree, so callers can rewrite names to their final JS
-    /// identities (see `Context::resolve_descriptor` in `wit`).
+    /// identities (see `Context::resolve_descriptor` in `wit`). Rust types
+    /// include their defining crate identifier; named externrefs do not.
     pub fn visit_named_types_mut<E>(
         &mut self,
-        f: &mut impl FnMut(&mut String) -> Result<(), E>,
+        f: &mut impl FnMut(&mut String, Option<&str>) -> Result<(), E>,
     ) -> Result<(), E> {
         match self {
             Descriptor::Function(function) => function.visit_named_types_mut(f),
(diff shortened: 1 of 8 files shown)


crates/cli/tests/wasm-bindgen/diagnostics.rs
```diff
@@ -348,6 +348,87 @@ fn duplicate_public_class_across_crates_errors() {
     assert_contains!(&err, "#[wasm_bindgen(private)]");
 }
 
+#[test]
+fn external_private_type_ref_binds_to_public_collision() {
```
guybedford, line 355: The strongest behavioral change of this fix isn't covered yet: when *every* colliding definition is private (`public_count == 0`), the old code hit the "exported as `#[wasm_bindgen(private)]` by multiple crates" bail for any cross-crate descriptor reference. With the crate id embedded, every private definition's identity entry is now reachable, so those references resolve deterministically instead of erroring — the ambiguity bail is now unreachable from the descriptor path entirely.

Please add a variant of this test where both dependency crates mark `Widget`/`Mode` private, asserting both numbered identities appear in the output (e.g. one crate's functions bind to `Widget`, the other's to `Widget2`). That pins the error-to-success transition and guards against a future regression reintroducing the bail.

crates/cli/tests/wasm-bindgen/diagnostics.rs
```diff
@@ -348,6 +348,87 @@ fn duplicate_public_class_across_crates_errors() {
...
+        fs::read_to_string(out_dir.join("external_private_type_ref_binds_to_public_collision.js"))
+            .unwrap();
+    let dts = fs::read_to_string(
+        out_dir.join("external_private_type_ref_binds_to_public_collision.d.ts"),
+    )
+    .unwrap();
+
+    assert_contains!(&js, "return Widget2.__wrap(ret);");
+    assert_contains!(&dts, "export function make_private_widget(): Widget2;");
+    assert_contains!(&dts, "export function private_mode(): Mode2;");
```
guybedford, line 432: Also pin the public side so a reversed identity assignment (the public definition getting renamed instead of the private one) can't slip past these assertions:

```suggestion
    assert_contains!(&dts, "export function private_mode(): Mode2;");
    assert_contains!(&dts, "export function make_public_widget(): Widget;");
    assert_contains!(&dts, "export function public_mode(): Mode;");
```

crates/cli-support/src/wit/mod.rs
```diff
@@ -888,14 +888,19 @@ impl<'a> Context<'a> {
...
-    /// outside its defining crate. Same-crate references resolve through the
-    /// identity map; anything else must have an unambiguous canonical owner.
-    fn resolve_item_ref(&self, qualified: &mut String) -> Result<(), Error> {
-        if let Some(identity) = self.item_identities.get(&(
-            self.unique_crate_identifier.to_string(),
-            qualified.to_string(),
-        )) {
+    /// Final JS identity of a struct/enum referenced by name. Rust type
+    /// descriptors carry their defining crate; other named descriptors use
+    /// the current program's crate.
```
guybedford, line 895: Now that struct/enum descriptors self-identify, the comment in `bind_generic_imports` (lines 513–515: "No defining crate context here … so this only rejects references to ambiguous all-`private` names") is stale: in the monomorphisation phase, struct/enum references resolve via their embedded crate id regardless of the cleared `unique_crate_identifier`, and only `NamedExternref` references can still hit the ambiguity bail there. Please update that comment to match.

crates/cli-support/src/wit/mod.rs
```diff
@@ -2095,7 +2104,7 @@ impl<'a> Context<'a> {
                     bare,
                 );
                 let parent_qualified = parent.clone();
-                self.resolve_item_ref(&mut parent)
+                self.resolve_item_ref(&mut parent, None)
```
guybedford, line 2113: Parent-class references still resolve against the current program's crate (`None`), so a cross-crate `extends` whose parent is a renamed private type won't resolve — it keeps the canonical name (binding to the wrong class) or hits the ambiguity bail. Unlike descriptors, the derived crate's macro expansion only has the parent's JS name, not the defining crate's identifier, so the fix isn't the same plumbing.

Fine to leave out of scope here, but please add a brief comment at this call site documenting the limitation so it isn't mistaken for being covered by the descriptor fix. A follow-up issue would also be reasonable.

crates/cli-support/src/descriptor.rs
```diff
@@ -238,9 +252,16 @@ impl Descriptor {
             | Descriptor::Vector(d)
             | Descriptor::Option(d)
             | Descriptor::Result(d) => d.visit_named_types_mut(f),
-            Descriptor::NamedExternref(name)
-            | Descriptor::Enum { name, .. }
-            | Descriptor::RustStruct(name) => f(name),
+            Descriptor::NamedExternref(name) => f(name, None),
```
guybedford, line 259: `NamedExternref` passing `None` means an imported JS type whose name happens to match an all-`private` colliding struct/enum name will spuriously hit the ambiguity bail, even though the externref never refers to the Rust type. Pre-existing behavior rather than a regression from this change, but the crate-id mechanism introduced here is the shape of the eventual fix — worth a note here or a follow-up issue.

guybedford (changes requested):
The mechanism here is right: the `WasmDescribe` impl is expanded in the defining crate, so baking `unique_crate_identifier` into the descriptor stream keys resolution correctly regardless of which crate's signature references the type, and the schema version lock covers the wire format change. Verified decode order matches the `inform` order for both `ENUM` and `RUST_STRUCT`, and that identity resolution runs before all program processing so there's no ordering hazard.

Remaining items, inline below:

* Test coverage for the all-`private` case, which is the strongest behavioral change of this fix (previously a hard error, now deterministic resolution).
* Pin the public side in the existing test's assertions.
* A now-stale comment in `bind_generic_imports`.
* Two adjacent gaps in the same class (parent-class refs and `NamedExternref`) that should at least be documented as out of scope.

On changelog: no entry needed — this fixes behavior introduced by the still-unreleased #5293, so there is no change between released versions to describe.

guybedford:
Pushed a commit addressing the review feedback (rebased to latest main):

* Added `external_all_private_type_collision_resolves` covering the all-`private` collision case: both dependency crates mark `Widget`/`Mode` private, and cross-crate references now resolve to the numbered identities (`Widget`/`Widget2`, `Mode`/`Mode2`) instead of hitting the ambiguity error. Identity assignment sorts by unique crate identifier, so the crate-to-identity mapping is deterministic and pinned in the assertions.
* Pinned the public side in the existing test (`make_public_widget(): Widget`, `public_mode(): Mode`) so a reversed identity assignment can't pass.
* Updated the stale `bind_generic_imports` comment: struct/enum references now resolve via their embedded crate id, only `NamedExternref` can still hit the ambiguity check there.
* Documented the parent-class (`extends`) limitation at its `resolve_item_ref` call site — the macro only records the parent's JS name, so cross-crate extends of a renamed private parent is out of scope for this fix.
* Added a note in `descriptor.rs` on the pre-existing `NamedExternref` behavior where a JS type name matching an all-private collision hits the ambiguity check.

guybedford (approved):

PR #5312, opened 2026-08-28
web-sys: include OffscreenCanvas in WebGL texImage* functions

AdrianEddy:
### Description

The WebGL specification's [`TexImageSource` typedef](https://registry.khronos.org/webgl/specs/latest/1.0/#5.14) includes `OffscreenCanvas`, but the vendored WebIDL predates it, so `web-sys` exposes no `OffscreenCanvas` overloads for `texImage2D` / `texSubImage2D` / `texImage3D` / `texSubImage3D`. Callers currently have to re-wrap an `OffscreenCanvas` as `HtmlCanvasElement` via `unchecked_ref()` to upload it (wgpu's GLES backend, for example, [marks this path `unreachable!()`](https://github.com/gfx-rs/wgpu/blob/d5fde4291f8ecf1c070324a46018a7c2fce09310/wgpu-hal/src/gles/queue.rs#L590)).

This adds the missing union members to `WebGLRenderingContext.webidl` / `WebGL2RenderingContext.webidl` and regenerates the bindings — the same approach as #3077, which added `VideoFrame` to these overloads. Purely additive; the new functions are gated on the existing stable `OffscreenCanvas` feature.

### Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 
 ### Added
 
+* Added `OffscreenCanvas` overloads for the WebGL `texImage2D`, `texSubImage2D`,
+  `texImage3D` and `texSubImage3D` functions, matching the `TexImageSource`
+  typedef in the WebGL specification.
+  [#5312](https://github.com/wasm-bindgen/wasm-bindgen/pull/5312)
+
 * Added `--split-debug-info` to the CLI. This option extracts the DWARF debug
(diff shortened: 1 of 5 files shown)


evilpie:
Would changing those methods to use the  `TexImageSourc` typedef result in changed (Rust) method signatures?

guybedford (approved):

PR #5314, opened 2026-08-31
Per-monomorphization generic callbacks

logan-gatlin:
### Description
Adds support for trait object callbacks (`&dyn Fn(...)` and `&mut dyn FnMut(...)`) with type parameter arguments and return types. Currently arguments and return types cannot be references, or have higher ranked lifetime bounds. So these cases:
- FnMut(&T)
- for<'a> FnMut(&'a T)
- FnMut() -> &'a T

are rejected. These cases should most likely be handled by erasible generics instead (they are already supported in that path).

### Checklist

### DIFF
--- a/crates/macro-support/src/codegen.rs
+++ b/crates/macro-support/src/codegen.rs
@@ -3307,69 +3307,63 @@ impl ast::ImportFunction {
             };
 
             // Raw callback trait objects already have generic ABI support in
-            // `convert::closures`; only their callback signature's bounds need
-            // to be carried onto the monomorphized wrapper.
+            // `convert::closures`. Lower all of them like ordinary imports,
+            // adding bounds when their callback signature is generic.
             if let Some((is_mut, fn_bounds)) = detect_raw_fn_trait_obj(ty) {
                 if let Some((inputs, ret, higher_ranked)) = fn_trait_bound_sig(fn_bounds) {
-                    let uses_generics = inputs
-                        .iter()
-                        .any(|input| generics::uses_generic_params(input, &type_params))
-                        || generics::uses_generic_params(&ret, &type_params);
-                    if uses_generics {
-                        if higher_ranked {
-                            bail_span!(
-                                ty,
-                                "experimental_generic_mono does not support higher-ranked generic callback signatures"
-                            );
-                        }
-                        for input in &inputs {
-                            if generics::uses_generic_params(input, &type_params) {
-                                if generics::references_generic_param(input, &type_params) {
-                                    bail_span!(
-                                        input,
-                                        "experimental_generic_mono does not support borrowed generic callback arguments"
-                                    );
-                                }
-                                where_bounds.push(quote! {
-                                    #input: #wasm_bindgen::convert::FromWasmAbi
-                                });
-                            }
-                        }
-                        if generics::uses_generic_params(&ret, &type_params) {
-                            if generics::references_generic_param(&ret, &type_params) {
+                    if higher_ranked {
+                        bail_span!(
+                            ty,
+                            "experimental_generic_mono does not support higher-ranked generic callback signatures"
+                        );
+                    }
+                    for input in &inputs {
+                        if generics::uses_generic_params(input, &type_params) {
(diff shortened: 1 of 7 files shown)


crates/macro-support/src/codegen.rs
```diff
@@ -3299,6 +3306,73 @@ impl ast::ImportFunction {
...
+                    let uses_generics = inputs
+                        .iter()
+                        .any(|input| generics::uses_generic_params(input, &type_params))
+                        || generics::uses_generic_params(&ret, &type_params);
+                    if uses_generics {
+                        if higher_ranked {
+                            bail_span!(
+                                ty,
+                                "experimental_generic_mono does not support higher-ranked generic callback signatures"
+                            );
```
guybedford: Can we get ui tests for the three new diagnostics (higher-ranked, borrowed argument, borrowed return)? `generic-per-mono-unsupported.rs` is the existing home for this kind of rejection coverage, and the spans here are good — worth locking them in.

crates/macro-support/src/codegen.rs
```diff
@@ -3299,6 +3306,73 @@ impl ast::ImportFunction {
...
+            // Raw callback trait objects already have generic ABI support in
+            // `convert::closures`; only their callback signature's bounds need
+            // to be carried onto the monomorphized wrapper.
+            if let Some((is_mut, fn_bounds)) = detect_raw_fn_trait_obj(ty) {
+                if let Some((inputs, ret, higher_ranked)) = fn_trait_bound_sig(fn_bounds) {
+                    let uses_generics = inputs
+                        .iter()
+                        .any(|input| generics::uses_generic_params(input, &type_params))
+                        || generics::uses_generic_params(&ret, &type_params);
+                    if uses_generics {
```
guybedford: I think we should drop this `uses_generics` gate so *concrete*-signature callbacks on the mono path get the same lowering. As written, `fn with_callback<T>(f: &mut dyn FnMut(u32), times: T)` still binds `f` as a raw dyn wrapper arg with no `MaybeUnwindSafe`, while `FnMut(T)` right next to it gets `impl FnMut + MaybeUnwindSafe` — so panics in the concrete case are still caught under `AssertUnwindSafe` without the caller ever opting in, which is exactly the hole #5056 closed on the ordinary import path.

The where-bound pushes are already gated per-input so removing the outer gate shouldn't introduce trivial-bound errors, and since the attribute is experimental the wrapper signature change is fine. (The erasure path has the same gap, but that's stable surface and out of scope here.)

tests/wasm/generic_import_args.rs
```diff
@@ -30,6 +31,9 @@ extern "C" {
     #[wasm_bindgen(experimental_generic_mono, js_name = withCallback)]
     fn with_callback<T>(f: &mut dyn FnMut(u32), times: T);
 
+    #[wasm_bindgen(experimental_generic_mono)]
+    fn transform<T, U>(value: T, f: &mut dyn FnMut(T) -> U) -> U;
```
guybedford, line 35: Coverage is `&mut dyn FnMut` only — can we add an immutable `&dyn Fn(T) -> U` import alongside so the `is_mut = false` arm is exercised at runtime too?

guybedford (commented):
This looks like the right approach — the runtime closure impls are already fully generic, so per-monomorphization callbacks fall out of the monomorphized describe informing a concrete `invoke` pointer per instantiation, and the wrapper just needs the ordinary path's `impl Fn`/`impl FnMut` lowering with the bounds carried onto the where clause. I verified the guide's `Array::map` shape, the immutable `&dyn Fn` variant, mixed concrete/generic inputs, and all three rejection diagnostics locally.

A few comments inline.

tests/wasm/generic_import_args.rs
logan-gatlin: Added immutable callback test

crates/macro-support/src/codegen.rs
logan-gatlin: Removed this guard and adjusted the tests

guybedford (approved):
Great work!

PR #5316, opened 2026-09-01
Add JsStringLike marker trait

guybedford:
This adds an experimental `JsStringLike` marker trait naming the string shapes that cross the wasm ABI as JavaScript string values.

Generic imports bound with `experimental_generic_mono` can already accept any concrete type satisfying the ABI traits, but there has been no way to name "any string" as a bound. This trait provides that vocabulary, so an import can accept whichever string shape the caller already holds — each crossing at its native wire format (UTF-8 buffers for the Rust strings, a handle for `JsString`, no re-encode) — rather than forcing every caller through one concrete shape.

* `JsStringLike` — implemented for `String`, `&str`, `js_sys::JsString` and `&js_sys::JsString`, with `IntoWasmAbi + OptionIntoWasmAbi` supertraits so both plain and nullable argument positions work under the single bound
* the trait is open: implementing it for a `#[wasm_bindgen]`-imported type is a promise that it marshals as a JS string
* like `experimental_generic_mono` itself, the trait is marked experimental and may change as that feature stabilizes

```rust
#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(experimental_generic_mono)]
    fn set_title<T: JsStringLike>(title: T);
}
```

Runtime tests cover all four argument shapes through a single generic import, and the guide documents the trait under the `experimental_generic_mono` reference.

### DIFF
--- a/src/convert/traits.rs
+++ b/src/convert/traits.rs
@@ -652,16 +652,20 @@ impl<T: IntoJsGeneric + Clone> IntoJsGeneric for &T {
 /// `String` and `&str` cross as UTF-8 buffers, `js_sys::JsString` and
 /// `&js_sys::JsString` as handles — all arriving in JS as a string value.
 ///
-/// Implementing this trait for a type is a promise that its `IntoWasmAbi`
-/// conversion produces a JS string on the other side, as is the case for
-/// `#[wasm_bindgen]`-imported extern types of JS strings.
+/// This trait is sealed: it is implemented only for the string shapes above,
+/// each of which is guaranteed to produce a JS string on the other side.
 ///
 /// `OptionIntoWasmAbi` is a supertrait so that nullable string positions
 /// (`Option<T>`) work under the same bound.
 ///
 /// This trait is experimental, like `experimental_generic_mono` itself, and
 /// may change or be removed as that feature stabilizes.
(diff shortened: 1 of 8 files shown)


src/convert/traits.rs
```diff
@@ -644,3 +644,24 @@ impl<T: IntoJsGeneric + Clone> IntoJsGeneric for &T {
...
+/// Implementing this trait for a type is a promise that its `IntoWasmAbi`
+/// conversion produces a JS string on the other side, as is the case for
+/// `#[wasm_bindgen]`-imported extern types of JS strings.
+///
+/// `OptionIntoWasmAbi` is a supertrait so that nullable string positions
+/// (`Option<T>`) work under the same bound.
+///
+/// This trait is experimental, like `experimental_generic_mono` itself, and
+/// may change or be removed as that feature stabilizes.
+pub trait JsStringLike: IntoWasmAbi + OptionIntoWasmAbi {}
```
logan-gatlin: Should this trait be sealed? I cant imagine any case where user code implementing it would be very good

logan-gatlin (approved):

src/convert/traits.rs
guybedford: Good point, added.

PR #5318, opened 2026-09-01
Add _js_string return variants across web-sys

guybedford:
This adds a `_js_string` return variant for every string-returning web-sys function, completing the string-handle story from #5227 on the return side: #5317 lets arguments cross by handle, and this lets returns come back by handle.

A string return is bound as an owned Rust `String`, which copies the string data into wasm memory on every call — wasted work when the value is large or is only handed back to other JS APIs. Generic returns were considered and rejected: a generic return demands a type annotation at every call site, which the Rust ecosystem reserves for conversion-shaped operations, not accessors. Instead this follows the variant naming approach of #5264 (as `js_u8_array` does), scoped to returns only, where the variant count stays linear — exactly one `_js_string` per string return, no combinatorial expansion.

* every function returning a string gains a `{name}_js_string` variant returning `js_sys::JsString`, across interface methods (including static and indexing getters), attribute getters, namespace functions and attributes, and dictionary field getters
* nullable and throwing returns carry through: `Option<String>` → `Option<JsString>`, `Result<String, JsValue>` → `Result<JsString, JsValue>`
* applies uniformly across generation modes, and composes with #5317's generic arguments (`fn echo_js_string<S0: JsStringLike>(value: S0) -> JsString`)
* the regenerated web-sys diff is purely additive: 1,370 new functions, no existing signature changed

```rust
let name: String = device.product_name();      // copies the string into wasm memory
```

```rust
let name = device.product_name_js_string();    // hands back the JS string by handle
```

Test coverage: the webidl-tests string fixture exercises the variants at runtime in all three generation modes — plain, nullable, throwing, static, attribute getters, dictionary getters, and namespace functions/attributes. web-sys is regenerated and verified with `--all-features` both with and without `web_sys_unstable_apis`.

Based on #5317.


logan-gatlin (approved):

guybedford:
I yanked 0.26.6 to avoid the issue with Walrus.

Issue #5202, opened 2026-06-15
Test deadlock on single-cpu systems


sanvila:
Hello. When I try to build this (version 0.2.108, from Debian) on a system with 1 CPU, the tests timeout:

```
[...]
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running `CARGO=/usr/bin/cargo CARGO_MANIFEST_DIR=/<<PKGBUILDDIR>> CARGO_MANIFEST_PATH=/<<PKGBUILDDIR>>/Cargo.toml
 CARGO_PKG_AUTHORS='The wasm-bindgen Developers' CARGO_PKG_DESCRIPTION='Easy support for interacting between JS and Ru
st.
' CARGO_PKG_HOMEPAGE='https://wasm-bindgen.github.io/wasm-bindgen' CARGO_PKG_LICENSE='MIT OR Apache-2.0' CARGO_PKG_LIC
ENSE_FILE='' CARGO_PKG_NAME=wasm-bindgen CARGO_PKG_README=README.md CARGO_PKG_REPOSITORY='https://github.com/wasm-bind
gen/wasm-bindgen' CARGO_PKG_RUST_VERSION=1.71 CARGO_PKG_VERSION=0.2.108 CARGO_PKG_VERSION_MAJOR=0 CARGO_PKG_VERSION_MI
NOR=2 CARGO_PKG_VERSION_PATCH=108 CARGO_PKG_VERSION_PRE='' LD_LIBRARY_PATH='/<<PKGBUILDDIR>>/target/x86_64-unknown-lin
ux-gnu/release:/<<PKGBUILDDIR>>/target/x86_64-unknown-linux-gnu/release/deps:/usr/lib/rustlib/x86_64-unknown-linux-gnu
/lib' OUT_DIR=/<<PKGBUILDDIR>>/target/x86_64-unknown-linux-gnu/release/build/wasm-bindgen-c0ce613feb8f45be/out /<<PKGB
UILDDIR>>/target/x86_64-unknown-linux-gnu/release/deps/non_wasm_test-5d31c15a506465bf`

running 2 tests
test test ... 
E: Build killed with signal TERM after 60 minutes of inactivity
--------------------------------------------------------------------------------

```

On a machine with 2 CPUs the build takes around 5 minutes to finish, so this looks like a deadlock to me.

[...]

PR #5323, opened 2026-09-06
Fix non_wasm_test deadlock with a single test thread

teddytennant:
### Description

Fixes the `non_wasm_test` deadlock on single-CPU hosts, resolving #5202.

`tests/non_wasm_test.rs` used a condvar handshake between two tests: `test` waited for `test_success` to signal. That only works when cargo runs both at once. On a 1-CPU machine cargo defaults to `--test-threads=1`, so `test` runs first, waits, and never lets `test_success` start.

The condvar was not testing `unsupported = test`. Dropping it leaves a single `#[wasm_bindgen_test(unsupported = test)]` that compiles and runs on native.

Verified:

```
# before (hung after "test test ...")
cargo test --test non_wasm_test -- --test-threads=1

# after
cargo test --test non_wasm_test -- --test-threads=1
# test result: ok. 1 passed
```

### Checklist

### DIFF
--- a/tests/non_wasm_test.rs
+++ b/tests/non_wasm_test.rs
@@ -1,25 +1,6 @@
 #![cfg(not(target_family = "wasm"))]
 
-use std::sync::{Arc, Condvar, Mutex};
-
-use once_cell::sync::Lazy;
 use wasm_bindgen_test::wasm_bindgen_test;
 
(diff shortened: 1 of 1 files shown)


guybedford (approved):
Restored the handshake in an ordering-safe form so `unsupported = test` is still verified to actually run; see the follow-up commit. Thanks for the fix!

PR #5326, opened 2026-09-09
fix: make Emscripten side-module bindings callable

danlapid:
### Description

Fix Emscripten side-module bindings where imported function pointers were treated as zero, causing closure descriptors to select the wrong function.

Descriptor interpretation and binding generation now share one function table, with support for ordinary indirect calls and clear errors for unavailable functions. The change also preserves the function aliases and shared exception state needed by the generated bindings.

Updates Walrus to 0.27.0, which includes the merged [dynamic-link section-order fix](https://github.com/wasm-bindgen/walrus/pull/320).

Verified generated bindings in debug and release, both alone and with another library loaded first to shift the memory and table layout. All 228 unit/CLI tests and strict Clippy pass using the published Walrus crate.

### Checklist


guybedford (commented):
Thanks, this is the right shape: sharing one instantiated table between the interpreter and `export_table_element` means the fabricated indices for GOT-only functions never reach the wasm, and the JS always calls the export by name.

I verified this locally: on `main` the fixture panics with `stack size mismatch for adapter 23` (`js/binding.rs:372`), since every closure `invoke`/`do_call`/`do_catch` shim in the side module is a `GOT.func` import and was resolving to slot 0. With this branch both the `plain` and `padded` layouts pass against emscripten `main`.

A few comments inline, mostly around the test and the scope of the `wasmImports` approach.

crates/cli/tests/emscripten-side-module/run.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python3
```
guybedford, line 1: Nothing runs this. There is already a `test_emscripten` job in `.github/workflows/main.yml` with emsdk, the `wasm32-unknown-emscripten` target and Node set up; could this script be invoked from that job? Otherwise the fixture will rot as soon as the emscripten side changes. (It runs fine on Node 22 here, so the Node 24 requirement in the README doesn't seem necessary.)

crates/cli/tests/emscripten-side-module/run.py
```diff
@@ -0,0 +1,99 @@
...
+    shutil.copy2(args.cli, directory / "bin" / ("wasm-bindgen" + suffix))
+    (directory / "temps").mkdir()
+    env = os.environ.copy()
+    env.pop("CARGO_ENCODED_RUSTFLAGS", None)
+    env.update(
+        RUSTC_WRAPPER="",
+        CARGO_TARGET_DIR=str(directory / "target"),
+        CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_LINKER=args.emcc,
+        RUSTFLAGS="-Clink-arg=-sWASM_BINDGEN=auto -Clink-arg=-sENVIRONMENT=node",
+        EMCC_DEBUG_SAVE="1",
```
guybedford, line 43: Extracting `library_bindgen.js` from the `EMCC_DEBUG_SAVE` temp directory works for the test, but it means there is currently no supported way for a user to get the JS library out of a `SIDE_MODULE` link. Is there a corresponding emscripten change planned so the library is emitted alongside the side module `.wasm`? Without that this isn't usable end-to-end.

crates/cli-support/src/js/mod.rs
```diff
@@ -7066,7 +7076,11 @@ addToLibrary({
     /// wasm export otherwise) and are renamed by the import/export minifier in
     /// lockstep with the wasm. Other modes use the local `wasm` binding.
     fn wasm_export_ref(&self, name: &str) -> String {
-        if matches!(self.config.mode, OutputMode::Emscripten) {
+        if self.emscripten_side_module {
+            // The dynamic loader publishes side-module exports in wasmImports;
+            // they have no receiving variables in the main module's exports.
+            format!("wasmImports[{}]", serde_json::to_string(name).unwrap())
```
guybedford, line 7082: Worth noting the scope this implies. `mergeLibSymbols` in `libdylink.js` is first-definition-wins (`if (!isSymbolDefined(target)) wasmImports[target] = exp`) and only runs for `global` loads. So this supports exactly one statically linked wasm-bindgen side module: a second one would collide on `__wbindgen_start`, `__wbindgen_externrefs`, `__externref_table_alloc`, `__wbindgen_exn_store`, etc. and silently call into the first, and a `dlopen` without `RTLD_GLOBAL` isn't reachable at all. Not blocking, but the changelog / README should state that scope explicitly.

crates/cli-support/src/lib.rs
```diff
@@ -428,14 +428,25 @@ impl Bindgen {
...
-                walrus::ImportKind::Tag(id)
-                    if impt.module == "env" && impt.name == "__cpp_exception" =>
-                {
-                    Some((impt, id))
-                }
-                _ => None,
-            });
+            // Emscripten side modules must use the main module's exception tag
+            // so an exception can cross the dynamic-library boundary.
+            let side_module = self.mode.emscripten()
```
guybedford, line 433: This `dylink` / `dylink.0` check is duplicated in `js/mod.rs` (`emscripten_side_module`). A small `wasm_conventions::is_dynamic_library(module)` helper would keep the two in sync.

crates/cli-support/src/interpreter/mod.rs
```diff
@@ -203,22 +243,83 @@ impl Interpreter {
             }
         }
 
-        // Reconstruct the function table from active element segments so an
-        // `invoke_*(fnptr, ..)` can resolve `fnptr` to a concrete function.
+        // Instantiate the function table in element-segment order. Reserve its
+        // full range, including null entries, before resolving imported symbols.
+        let table = module.tables.main_function_table()?;
+        let mut next_index = table.map_or(0, |id| module.tables.get(id).initial).max(1);
```
guybedford, line 249: Why `.max(1)`? For an imported table `initial` already covers every element slot (69 == elem count in the fixture), so the null slot is never a candidate here. If it is intentional it needs a comment, otherwise drop it.

PR #5331, opened 2026-09-11
cli-support: support Emscripten shared memory / pthreads mode

ZacCranko:
Hello! I am your friendly neighbourhood spaghetti toaster AI.

Emscripten pthreads (`-pthread` / `--shared-memory`) imports `env.memory` (`shared = true`) and emits its own TLS init (`__wasm_init_tls`). Without this change:
1. `transforms::threads::run` sees `shared = true` and deletes `__wasm_init_tls` / overwrites `__tls_base`.
2. `js::Context` sees `mem.import.is_some()` and rewrites `env.memory` to `__wbindgen_placeholder__.memory`, causing a `LinkError` on instantiation.

Skipping both when `OutputMode::Emscripten` is active preserves Emscripten's TLS and `env.memory` import.

### DIFF
--- a/crates/cli-support/src/js/mod.rs
+++ b/crates/cli-support/src/js/mod.rs
@@ -772,8 +772,15 @@ impl<'a> Context<'a> {
             String::new()
         };
 
-        if let Some(mem) = self.module.memories.iter().next() {
-            if let Some(id) = mem.import {
+        // Emscripten pthreads import shared memory as `env.memory` rather than placeholder.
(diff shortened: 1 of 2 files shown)


guybedford:
I believe this was already now fixed in https://github.com/wasm-bindgen/wasm-bindgen/pull/5315 and https://github.com/wasm-bindgen/wasm-bindgen/pull/5254, which should both be available on 0.2.128. Do you still need this on that version?

ZacCranko:
Sorry for the noise. Yep looks good!

PR #5338, opened 2026-09-21
Time out stalled WebDriver commands in the headless test runner

LucaCappelletti94:
**EDIT: this PR is currently fixing a few distinct timeout-related stalling. I will shortly split it apart into smaller modular PRs, one for each failure mode I identified.**

The headless test runner sends every `WebDriver` command through a `ureq` agent configured with no timeouts, so `doit`'s `read_to_string` waits forever whenever a driver accepts a request and never answers. `WASM_BINDGEN_TEST_TIMEOUT` cannot save the run because its elapsed check only runs between polls, and a stalled poll never returns. The runner hangs at `POST /session`, at the `execute/sync` output poll, and at the final `DELETE /session/{id}/window`, and the last case hangs even after a fully green run.

Every request now carries a per-request `timeout_global`, a minute at launch since browser cold start on shared CI runners can exceed the poll budget yet still finish, sized from `WASM_BINDGEN_TEST_TIMEOUT` while polling and a short fixed budget for teardown, and timeouts have an error marker so a stall can be distinguished from an answered failure. A stalled final window close therefore fails an otherwise green run, while answered close errors stay warn-only as before.

Furthermore, `BackgroundChild::drop` kills the driver and then joined reader threads waiting for EOF on its stdout and stderr, but the browser the driver spawned inherits those pipes, so the EOF could wait on a process that outlives the driver entirely. A failing run could wedge in teardown exactly where it used to wedge mid-request. The stdio dump is now collected under a 2 second deadline and skipped with a one line notice when a child still holds the pipes, so the runner always exits and a wedged driver surfaces as a fast, diagnostic failure instead of a silent multi-hour hang.

Unless I am missing anything, `close_window` has no evident reason to take the `id` parameter as it can `read` the `id` from `self.session` directly, which seems less prone to mistakes, but maybe there are use cases where this is needed and at the time of writing I have not identified them.


guybedford (approved):
Thanks, this is a solid fix — I verified the hang on main and the fix with a fake stalled driver locally.

I pushed a small follow-up commit to the branch to trim scope so this can land as-is:

* A stalled `DELETE /window` after a green run stays warn-only (via `Client::drop`) rather than failing the run, since the exit code doesn't change the outcome and would otherwise turn a teardown stall into a red `cargo test`.
* Dropped the two new env vars; startup and page load use the drivers' own defaults (60s / 300s) floored at `WASM_BINDGEN_TEST_TIMEOUT`, so raising the existing var already lifts them.
* Removed the 5s `drop_closes_within_the_cleanup_budget` test, which `failed_close_consumes_the_session` already covers.

No need to split it out further, landing once CI is green.

